PR #179, opened 2023-10-05
feat: add cookie support

hanspagel:
WIP

Doesn’t work yet. Set-Cookie header is present, but I can’t access anything in JavaScript (yet).


hanspagel:
Stale for too long

PR #193, opened 2023-10-08
fix: ensure default tag index is used before adding to tags

marclave:
before we appended to the tags list, but we should be finding the default tag then adding to it.

There might be a more elegant way to do this so im happy to have feedback


hanspagel (approved):
Way better :D

PR #194, opened 2023-10-08
fix: safely return if no example responses

marclave:
if there are no example responses, we can safely return and not cause errors since `Object.keys` on null isn't happy :)


hanspagel (changes requested):
Awesome!

Let’s add a test, though! I’d be happy to add it, but whoever comes first.

hanspagel (approved):

PR #198, opened 2023-10-09
fix: keep swagger editor state

hanspagel:
Previously, the state was kept in storage but only restored onMount. This caused bugs, e.g. resetting the content  when changing tabs.

This PR stores the content from local storage in the the `currentExample` ref which we already introduced anyway. This seems to work better.


marclave (approved):
💵 nice

PR #200, opened 2023-10-09
feat: add proxy url to html api

hanspagel:
This PR adds `data-proxy-url` to the HTML API. 😎


marclave (approved):
so cool 😎 Just tested and works great ✨

PR #201, opened 2023-10-09
Geoff/doc 1138 inject the package version to the api client request proxy

geoffgscott:
Expose the app version for use in deployed health checks


hanspagel:
excerpt from the build:

```
const h = "0.5.2"
// …
{
    app: o,
    listen: (e, s) => o.listen(e, s),
    /** Current proxy version */
    version: h
  };
```

PR #206, opened 2023-10-10
remove grid from mobile reference as it breaks layout

cameronrohani:
Before
https://github.com/scalar/scalar/assets/6201407/f533b53b-b7f4-44b0-a1e7-05b7f866e629

After
image


marclave (approved):
nice work :)

PR #207, opened 2023-10-10
fix: wrap watchEffect onMounted for dark light toggle until we use refs

marclave:
this breaks SSG before, we need to be using a ref.

We also should think about a way to ensure document deploys in internal repo are less fragile


cameronrohani (approved):
Hopefully this bug won't be coming around the MOUNTain again

PR #209, opened 2023-10-10
fix: add SearchModal to exports

marclave:
Now we can use our wonderful SearchModal outside the APIReferences :)


hanspagel (approved):
NICE

PR #211, opened 2023-10-10
fix: active state in sidebar with prefix of httpVerb for all ids

marclave:
before:
image

after:
https://github.com/scalar/scalar/assets/6176314/641883c9-a80b-42a2-b44f-9acc62dd57b7


hanspagel (approved):
Awesome, Sherlock! 🕵️  Let’s get this in.

PR #212, opened 2023-10-10
fix: don’t always show the description loading placeholder

hanspagel:
The description loading placeholder was even shown when a spec file without a description was loaded.

This PR makes sure the description place holder is only shown when there’s no spec file loaded.


marclave (approved):
🔥

PR #217, opened 2023-10-10
fix: fastify plugin does not show content

hanspagel:
With #166 we introduced a regression which made the ApiReference not show the initial content. This PR fixes it by passing the initial content to the parser, not just to the SwaggerEditor CodeMirror instance.


marclave (approved):
tested and works ✨

PR #220, opened 2023-10-10
Fix references breadcrumbs throwing on some specs

hwkr:


marclave (approved):
awesome, let's get this in 💪

PR #221, opened 2023-10-10
Rename old default variables

hwkr:
I _think_ all these should be updated


marclave (approved):
let's go! these defs need to be updated ✨

PR #226, opened 2023-10-11
search modal redesign + sneaky producthunt message

cameronrohani:
Before:
image

After:
image


hanspagel (approved):
That was fast! Really dig it.

Let’s wait for the green ticks to appear. ✅

PR #235, opened 2023-10-12
chore: add turborepo

hanspagel:
This PR adds turborepo. Turborepo tries to run tasks in parallel as much as possible and caches the results. Makes builds and CI extremely fast.

GitHub Actions is down from 5m to 1m. And there’s still a lot more potential, but let’s further improve it in a separate PR.

**Performance**
Full build locally (without Turborepo):
80s

Initial build with Turborepo:
30s

Cached build with Turborepo:
0.2s :exploding_head:

Full build in CI without Turborepo:
95s

Initial build in CI with Turborepo:
35s :exploding_head:

Cached build in CI with Turborepo:
5s :exploding_head: :exploding_head: :exploding_head:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,6 +32,8 @@ jobs:
       - name: Turborepo cache
         uses: dtinth/setup-github-actions-caching-for-turbo@v1
       - name: Build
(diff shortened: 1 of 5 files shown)


marclave (approved):
thursday is no forever turbo day

PR #236, opened 2023-10-12
chore: add new configuration object to the api reference component

hanspagel:
WIP

Marc and I had this idea to pass a configuration object to the <ApiReference /> component instead of using all those props. That universal configuration object can then be used in Vue.js, React, Fastify and with the CDN version.

It also makes it easier to introduce new props/options to other platforms, without adding to every single integration.


marclave (commented):
i love this! let's get this in monday, it is a breaking change tho so how should we handle this with everyones integrations?

hanspagel:
Good point! I’ve added the old props and marked them as deprecated. They are still used as fallbacks for the configuration object, but trigger a console warning now.

marclave (approved):
AWESOME work ✨

PR #237, opened 2023-10-12
feat: add interactive params, headers and variables

marclave:
we now have interactive parameters, variables and headers! Feels more like a rest api client every day now ✨ 

image

image


cameronrohani (approved):
great start, will further tweak and improve some of these states in the coming days

PR #240, opened 2023-10-13
feat: show headings with lower depth in the sidebar

hanspagel:
Let’s say there are only h2 or h3 in the description, let’s find out and just use the highest level that was used and put all headings of that level in the sidebar.

Example: https://github.com/aolabsai/archs/blob/main/core_api.yaml


marclave (approved):
wicked 🆒

PR #243, opened 2023-10-13
chore: make the linting git hook only fix staged files

hanspagel:
The linting Git hook took like 50s or so. 🤡 I’ve changed the hook to only lint staged files, which just takes a few ms (depending on how many files are staged).


marclave (approved):
🤯 awesome fix

PR #244, opened 2023-10-13
wip: add models to api refs

marclave:


hanspagel:
Thanks for kicking this off! I got sidetracked and moved the code to #247, will finish the models in a separate PR.

PR #247, opened 2023-10-13
refactor: new sections & new sidebar

hanspagel:
Initially, I wanted to add a new section with the models from the spec file. But when I wanted to copy the spacing from the other sections I struggled to understand what CSS applies where. That’s why I refactored the references to use `Section` components which come with the proper spacing and borders.

For the new models section I wanted to extend the IntersectionObserver, but that lived in two components already. I merged them and moved them to the new `Section` component. I’ve also added helpers to generate the ids, which (I think) is a huge improvement.

The new model section is hidden in this PR, I’ll finish it in a separate PR. :v:


marclave (approved):
awesome ✨

PR #249, opened 2023-10-13
Fix scrolling for mobile API client

hwkr:
Classic Safari 100vh bug

classic-youre-classic

Fixed:

IMG_9830


marclave (approved):
AWESOME work :)

PR #253, opened 2023-10-16
feat: add analyse helper to return stats about specs

hanspagel:
**analyze()**
This PR introduces a tiny `analyze` helper function which returns a few basic stats about the given spec. The current aim is to make testing of specs easier. I think we’ll extend the anaylze function to return more stats soon.

**preflight()**
This PR adds a `preflight` helper function which analyzes an unparsed spec. It’s faster, but inaccurate. 

**parse instead of parseSwaggerFile**
It also renames `parseSwaggerFile` to just `parse`. It felt wrong to name other helper files `somethingSwaggerFile()`. And it‘s not even expecting a file, just the content.

### DIFF
--- a/.changeset/five-crews-mix.md
+++ b/.changeset/five-crews-mix.md
@@ -1,0 +1,5 @@
+---
+'@scalar/swagger-parser': patch
+---
+
+feat: add analyze helper to get stats for Swagger/OpenAPI specs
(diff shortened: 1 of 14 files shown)


hanspagel:
No harm expected, will merge once it’s green.

marclave (approved):
This is super helpful! Great idea and initiative to get this one ✨

PR #254, opened 2023-10-16
fix: hanging parser

hanspagel:
The recent version of @apidevtools/swagger-parser runs into a timeout with some (not all) large files. As described here:

https://github.com/APIDevTools/swagger-parser/issues/221

The mentioned workaround (pinning @apidevtools/json-schema-ref-parser to 9.0.7) seems to work just fine. I’ve added a test to make sure we don’t reintroduce the regression.


marclave (approved):
🤯 wow, great fix @hanspagel and a regression tests too!

PR #260, opened 2023-10-16
fix: check if the endpoint has example responses

hanspagel:
Before rendering example responses, check whether the operation has example responses. :)


marclave (approved):
safety 👷 !

PR #267, opened 2023-10-16
chore: add image to api reference package readme

hanspagel:
This PR

* adds a link to the demo to the root README image.
* adds the same image to the @scalar/api-reference package README.


marclave (approved):
💅 lets go

PR #270, opened 2023-10-16
feat: add requestbody to api client if json schema available

marclave:
If a user has JSON schema for a request body, we populate it in the api client.

If they dont we set an empty object

image

image

### DIFF
--- a/packages/api-client/src/helpers/generateRequestBody.test.ts
+++ b/packages/api-client/src/helpers/generateRequestBody.test.ts
@@ -1,0 +1,254 @@
+import { generateRequestBody } from 'src/helpers/generateRequestBody'
+import { describe, expect, it } from 'vitest'
+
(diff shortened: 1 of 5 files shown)


packages/api-client/src/helpers/generateRequestBody.test.ts
```diff
@@ -0,0 +1,254 @@
...
+    ).toMatchObject({
+      tags: [
+        {
+          id: 1,
+        },
+      ],
+    })
+  })
+
+  it('uses empty quotes as a fallback for arrays', () => {
```
maplemuse, line 83: empty square brackets

marclave: ah good catch!

hanspagel (approved):
LGTM

PR #271, opened 2023-10-17
fix: default show endpoint when tags have one endpoint

marclave:
before:
image

after:


cameronrohani (approved):
This || changes everything. Thank you!

PR #273, opened 2023-10-17
fix: filter out tags with no operations

marclave:
We can remove tags that have no operations, it's default for what most swagger/oas doc tools use and I think that makes sense.

I couldnt imagine the need for a tag with no endpoints, let me know if you disagree :)

### DIFF
--- a/packages/swagger-parser/src/helpers/analyze.test.ts
+++ b/packages/swagger-parser/src/helpers/analyze.test.ts
@@ -44,7 +44,13 @@ describe('analyze', () => {
     const spec = {
       openapi: '3.1.0',
       info: {},
-      paths: {},
+      paths: {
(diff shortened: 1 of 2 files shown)


marclave:
some tests are failing @hanspagel , not sure what we should do. Should we update this test to just be 1 tag since we only expect 1 tag?

```
  it('counts tags and tags in operations', async () => {
    const spec = {
      openapi: '3.1.0',
      info: {},
      tags: [
        {
          name: 'foo',
        },
      ],
      paths: {
        '/example': {
          get: {
            tags: ['bar'],
          },
        },
      },
    }
    ```

hanspagel (approved):
Oh, yeah. Let’s filter them out!

PR #274, opened 2023-10-17
fix: add null check to tag operations in sidebar

marclave:
handles the edge case where people have no operations for a tag


cameronrohani (approved):
Good catch!

PR #278, opened 2023-10-17
feat: use ‘default’ for example responses

hanspagel:
Some schemas provide a `default` value. With this PR it’s used for the example responses, if it’s available:

Screenshot 2023-10-17 at 15 32 01


marclave (approved):
Love a great default response ✨

PR #279, opened 2023-10-17
fix: deal with empty descriptions

hanspagel:
In #247 I introduced a tiny regression. Specs without a description showed the loading skeleton forever. This PR fixes it.


marclave (approved):
nice fix ✨

PR #283, opened 2023-10-17
refactor: move generate request helper to api reference

hanspagel:
In #270 we duplicated the code to generate JSON from schemas and added it to the client. I think the Swagger business logic should stay in the api-reference package.

This PR changes the following:
* The generateRequestBody is deleted, instead we’re using the existing generateResponseContent (@scalar/api-reference) helper to generate the json
* The generateRequest helper had to be moved to the api-reference

And one feature is added:
The Content-Type header is set to `application/json` when a json body is added to the request.


marclave (approved):
amazing refactor!

PR #285, opened 2023-10-17
chore: make CI even faster

hanspagel:
This PR combines all turbo tasks to a single command, which allows turborepo to further improve the performance.


marclave (approved):
its turbo time!

PR #288, opened 2023-10-17
Marc cam/polish models

marclave:


hanspagel:
Awesome! Why aren’t you using the `SectionColumn` component? 🤔

marclave:
> Awesome! Why aren’t you using the `SectionColumn` component? 🤔

vscode liveshare only had saved our css removed, we refactored the html. good catch! 

we wanted to do a full width version instead

packages/api-reference/src/components/Content/Introduction/Introduction.vue
```diff
@@ -25,12 +25,12 @@ const { state, getClientTitle, getTargetTitle } = useTemplateStore()
...
-        :loading="!info.title"
-        tight>
-        {{ info.title }}
-      </SectionHeader>
       <SectionContent :loading="!info.description && !info.title">
+        <SectionHeader
+          :loading="!info.title"
+          tight>
+          {{ info.title }}
+        </SectionHeader>
```
hanspagel, line 33: This breaks the loading skeleton.

marclave: cam fixed it with css

The motivation is the content is confined to the column and doesnt overflow into the second column

PR #289, opened 2023-10-17
fix: API Client styles

hanspagel:
This PR fixes a few minor CSS issues with the standalone API client.

* Proper box-sizing for the API Client
* Vertically center url


marclave (approved):
awesome fixes ✨

PR #316, opened 2023-10-18
Marc/doc 1223 ensure parent tag gets opened when sidebaritem is changed

marclave:
Now a tag opens if it is closed when trying to scroll to an endpoint ✨ 
https://github.com/scalar/scalar/assets/6176314/c6c44ecd-bd9c-4d47-b4da-c5f4ed0ad17d


cameronrohani (approved):
Much needed, thanks for your hard work on this one!

PR #319, opened 2023-10-18
chore: make release workflow faster

hanspagel:
Let’s add turborepo to the release workflow, too. ⚡


marclave (approved):
gotta go fast! ✨

PR #320, opened 2023-10-18
feat: added sidebar scrolling

amritk:
Added scrolling of the sidebar stripe style

https://stripe.com/docs/api

Also feel free to remove/keep the smooth scroll over the jumpy one stripe has

### DIFF
--- a/packages/api-reference/src/components/Sidebar.vue
+++ b/packages/api-reference/src/components/Sidebar.vue
@@ -179,17 +179,48 @@ const activeSidebarItemId = computed(() => {
 
   return flattenedItems.find((item) => isVisible(item.id))?.id ?? null
 })
+
+// This offset determines how far down the sidebar the items scroll
+const SCROLL_OFFSET = -160
+const scrollerEl = ref<HTMLElement | null>(null)
+
+// Watch for the active item changing so we can scroll the sidebar
+watch(activeSidebarItemId, (activeId) => {
+  const el = document.getElementById(`active-sidebar-element-id-${activeId!}`)
+  if (!el || !scrollerEl.value) return
+
+  let top = SCROLL_OFFSET
+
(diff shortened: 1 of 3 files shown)


hanspagel (approved):
Feels great 🌟

packages/api-reference/src/components/Sidebar.vue
```diff
@@ -179,17 +179,48 @@ const activeSidebarItemId = computed(() => {
...
   return flattenedItems.find((item) => isVisible(item.id))?.id ?? null
 })
+
+// This offset determines how far down the sidebar the items scroll
+const SCROLL_OFFSET = -160
+const scrollerEl = ref<HTMLElement | null>(null)
+
+// Watch for the active item changing so we can scroll the sidebar
+watch(activeSidebarItemId, (activeId) => {
+  const el = document.getElementById(`active-sidebar-element-id-${activeId!}`)
```
marclave, line 189: I think we should do refs here

https://vuejs.org/guide/essentials/template-refs.html#refs-inside-v-for

marclave (changes requested):
left one comment to fix up, but the scrolling feels SO smooth, great work @amritk .

Im happy to merge once that's changed :) 

Specific reasoning for this is SSG would much prefer using refs

packages/api-reference/src/components/Sidebar.vue
amritk: > It should be noted that the ref array does not guarantee the same order as the source array.
So then how do you access the ref after? Loop through the array?

amritk: Alright had to do some vue learning, they made some changes from vue 2. Check it out now

marclave (approved):
AWESOME work, lets get this in

PR #321, opened 2023-10-18
chore: use flex-end instead of end

hanspagel:
[vite:css] end value has mixed support, consider using flex-end instead


cameronrohani (approved):
That's the end of this issue!

Issue #312, opened 2023-10-18
Responses assumed to be from proxy


danp:
### Expected Behavior

Loaded scalar from a CDN and not using a scalar request proxy. Issue a request. See the response.

### Actual Behavior

The network inspector shows the response coming back successfully but the UI doesn't show it.

### Reproduction steps

```bash
1. Find an operation
2. Issue a test request for it (maybe changing headers/etc to make it succeed)
3. See that only response duration is shown
```

### Version

0.7.9

### What browsers are you seeing the problem on?

Chrome

danp:
I was able to get things going (partially, will open another issue) with this:

```diff
diff --git a/packages/api-client/src/helpers/sendRequest.ts b/packages/api-client/src/helpers/sendRequest.ts
index 493c25d4..d130a04b 100644
--- a/packages/api-client/src/helpers/sendRequest.ts
+++ b/packages/api-client/src/helpers/sendRequest.ts
@@ -76,7 +76,8 @@ export async function sendRequest(
     await axios(config)
       .then((res) => {
         return {
-          ...res.data,
+          ...(proxyUrl ? res.data : res),
+          ...(!proxyUrl && {statusCode: res.status}),
           error: false,
         }
       })
```

hanspagel:
Good catch! :)

danp:
👍 

Actually, before I open another issue, I also had to add this patch to get the JSON bodies to show:

```diff
diff --git a/packages/api-client/src/components/ApiClient/Response/Response.vue b/packages/api-client/src/components/ApiClient/Response/Response.vue
index 712a2c8d..924b8c3b 100644
--- a/packages/api-client/src/components/ApiClient/Response/Response.vue
+++ b/packages/api-client/src/components/ApiClient/Response/Response.vue
@@ -59,6 +59,9 @@ const responseData = computed(() => {
   if (value && isJsonString(value)) {
     return JSON.stringify(JSON.parse(value), null, 2)
   }
+  if (value && !isJsonString(value)) {
+    return JSON.stringify(value, null, 2)
+  }
 
   return value
 })
```

Perhaps related? Was getting an error in the dev console otherwise.

marclave:
@danp wouod you be interested in opening that as a PR; @hanspagel maybe we can also describe the interface for the proxy with inline comments maybe?

danp:
[...]

PR #331, opened 2023-10-20
fix: handle both direct and proxied responses

danp:
Direct request responses look different than responses received via a
proxy. Transform direct responses to look more like proxied responses
after they are received.

Updates #312


hanspagel:
AWESOME! Thanks for taking the time to fix this. :v:

hanspagel (approved):

PR #332, opened 2023-10-20
fix: optional chain

hanspagel:
This should fix an error with some specs having tags without operations.


marclave (approved):
optional chaining to the rescue! ✨

PR #334, opened 2023-10-20
feat: use min value as the default for numbers

hanspagel:
With this PR the min value for numbers is used as the default when generating example responses. 🤓


marclave (approved):
ah great find! ✨

PR #335, opened 2023-10-20
feat: generate example objects in arrays

hanspagel:
With this PR generated example responses include example objects *inside* an array.


marclave (approved):
awesome work @hanspagel , we are moving towards full OAS/swagger support 🖖

packages/api-reference/src/helpers/generateResponseContent.ts
```diff
@@ -4,7 +4,7 @@
 export const generateResponseContent = (
   schema: Record<string, any>,
   level: number = 0,
-) => {
+): any => {
```
marclave, line 7: one day we will write an amazing type for this, but until then any is perfect

PR #337, opened 2023-10-20
fix: don’t overwrite the URL on small screens

hanspagel:
We had the idea to shorten the URL on small screens, but what we did was actually overwriting the URL. That doesn’t work. :) I disabled dit for now, we need to find a rendering-based solution.


marclave (approved):
ah great idea, had no idea it was overwriting! good catch ✨

PR #339, opened 2023-10-20
fix: add target blank to markdown links

marclave:
now markdown links open in new tabs ✨


cameronrohani (approved):
I would approve this but I'm on another tab

PR #345, opened 2023-10-23
fix: make inputs more consistent

cameronrohani:
In preparation for auth branch... We have 3 different types of inputs across our app... Now we only have 2.5!

Before:
image

After:
image


marclave (approved):
great work @cameronrohani ! Love seeing our inputs getting more efficient

PR #346, opened 2023-10-23
chore: use different heading levels, fix #310

### LINKED ISSUE #310: SectionHeader uses h1 for all headings, use different levels
The SectionHeader uses h1 everywhere. We should only use h1 for the spec title, and h2 + h3 for the rest.

hanspagel:
With this PR different heading levels are used:

- h1 title
  - h2 tag
    - h3 operation
  - h2 model


marclave (approved):
we have achieved semantic gains

PR #347, opened 2023-10-23
fix: externalize all @scalar/use-codemirror deps, fix #309

### LINKED ISSUE #309: Fresh Vite app build fails with `Cannot read properties of null (reading 'refs')`
I created a new Vite app and added the `ApiClient` component from @scalar/api-client.

The build fails with the following message:
```
index-73bc151c.js:1 Uncaught TypeError: Cannot read properties of null (reading 'refs')
```

hanspagel:
This PR externalizes all deps of `@scalar/use-codemirror`. This should fix #309.


geoffgscott (commented):
These dependency lists are getting very long. If we don't want to bundle I suggest we look at reducing them. We already have utilities for time ago and pretty bytes in the main repo. 

We also need to ensure all of these are listed in the package as required peer dependencies.

hanspagel:
A few of the deps weren’t even used in the package, I removed them. I want to go through all packages, but that’s probaly out of scope for this PR.

marclave (approved):

PR #348, opened 2023-10-23
feat: make servers selectable

hanspagel:
This PR reads the servers list from the spec and makes the server selectable.

**Preview**
https://github.com/scalar/scalar/assets/1577992/263ed7a9-c6c6-498d-86c1-1de565848d7c

**Features**
* Select servers from the list
* Change variables for server environments
* Passes the configured server environment to the example request generator
* Passes the configured server environment to the API client
* New Cookie section in the API client

**Refactoring**
This PR introduces a set of new helpers to generate request data (trying to stay close to the HAR format) and we’re using those new helpers for both, the example requests and the API client.

The `getHarRequest` takes in as many partial HAR requests as needed and merges them to one valid HAR request:

```ts
const request: HarRequest = getHarRequest(
  {
    url: getUrlFromServerState(serverState),
  },
  getRequestFromOperation(operation),
  getRequestFromAuthentication(authenticationState),
)
```


marclave (approved):
AWESOME work @hanspagel

PR #349, opened 2023-10-23
DOC-1162 fix: active state for operations with multiple tags

hanspagel:
This PR prefixes all operation section ids with their parent tag. This makes it possible to distinguish which operation should be active, if the operation has multiple tags and is listed in multiple places.


marclave (approved):
awesome work! ✨

PR #350, opened 2023-10-23
style: use ellipsis for empty strings

hanspagel:
I think it looks nice to have `"…"` instead of just an empty strings in the example responses. But just for the responses, not for the request bodies.

And the helper generates example *request* and *response* bodies. I renamed it: `generateResponseContent` -> `getExampleFromSchema`

Screenshot 2023-10-23 at 15 46 50

### DIFF
--- a/.changeset/spicy-owls-suffer.md
+++ b/.changeset/spicy-owls-suffer.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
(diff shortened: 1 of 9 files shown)


hwkr:
Yo this looks great

marclave (approved):
I had a good look at what else is out there for `"..."` in code examples , it makes sense for the responses.

Nice work @hanspagel

PR #354, opened 2023-10-23
fix: use undefined as the default for deprecated props

hanspagel:
In #236 I’ve added the deprecated props, which by default fallback to false and are then considered to be set. Eventually `showSidebar` was `false`.

With this PR the default for those deprecated boolean props is `undefined`. This should fix the bug and the sidebar should appear again. 🥵


marclave (approved):
great work on the quick fix ✨

PR #365, opened 2023-10-24
show request type in reference sidebar

cameronrohani:
Before:
image

After:
image

### DIFF
--- a/packages/api-reference/src/components/SidebarElement.vue
+++ b/packages/api-reference/src/components/SidebarElement.vue
@@ -96,8 +96,13 @@ defineExpose({ el })
           :src="item.icon.src" />
         <p>{{ item.title }}</p>
         <div
+          v-if="item.httpVerb"
           class="sidebar-heading-type"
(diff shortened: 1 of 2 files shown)


packages/api-reference/src/components/SidebarElement.vue
```diff
@@ -94,6 +95,9 @@ defineExpose({ el })
           class="sidebar-icon"
           :src="item.icon.src" />
         <p>{{ item.title }}</p>
+        <div
+          class="sidebar-heading-type"
+          :class="`sidebar-heading-type--${item.httpVerb}`"></div>
```
hanspagel, line 100: Am on mobile, maybe I’ve overseen it, but should we add this class only when httpVerb is set?

cameronrohani: 💯% 

Will update

thelinuxlich:
nice!

marclave (approved):
great work @cameronrohani

PR #370, opened 2023-10-25
Marc/ai content writer

marclave:
integrating our ai content writer! :)


cameronrohani (approved):
Thankfully I removed 889 lines of code today, so I guess we can add 908. Looks good to me. Let's go!

PR #372, opened 2023-10-25
fix heading and inputs

cameronrohani:
before:
image

After:
image


marclave (approved):
the picasso of headings! great work :)

PR #362, opened 2023-10-24
feat: download spec

hanspagel:
This PR adds a download link to the reference:

Screenshot 2023-10-26 at 11 33 41

It also includes some minor changes:
* adds footerBelowSidebar as a deprecated prop
* Github -> GitHub
* refactors how the data flows between the swagger editor and the reference

PR #376, opened 2023-10-26
hotfix: load swagger editor in readonly mode

hanspagel:
I’ve added a regression in #362. I didn’t want to load the swagger editor if it’s not used, but that also means the spec isn’t parsed … because currently the swagger editor is responsible for the parsing.


marclave (approved):
quick fix! great work @hanspagel

PR #378, opened 2023-10-26
hotfix: import spec from url

hanspagel:
Wow, another regression from #362. The spec URL was fetched and stored in a ref, but the wrong ref was passed to the editor. This is the fix:

```diff
-:value="parsedSpecRef"
+:value="rawSpecRef"
```

Everything else cleans up the messy data flow just a little bit. Hopefully I didn’t introduce a new regression. Looks like it’s time to refactor the reference <> swagger editor <> parser data flow. 😵‍💫


marclave (approved):
nice fix ✨

PR #380, opened 2023-10-26
fix markdown collisions + prefix the card name

cameronrohani:
We're seeing some collisions with markdown + cards on users websites so I unset all markdown and prefixed cards. 

We should come up with something more robust later but this should be good for now


marclave (approved):
huge improvements :) nice work

PR #382, opened 2023-10-27
style: hide empty table headers

hanspagel:
I saw some empty table headers in a spec. This PR adds CSS to hide empty table headers:

**Before**
Screenshot 2023-10-27 at 14 24 55

**After**
Screenshot 2023-10-27 at 14 26 00

In theory, you could have some table headers empty and some not, but this looks still okay-ish and is probably less likely than having all table headers empty.


marclave (approved):
a nice addition! ✨

PR #383, opened 2023-10-27
fix: empty example responses

hanspagel:
I saw some empty example responses. This PR includes two related fixes:

* When there’s an example, make sure to transform it to text before passing it to CodeMirror.
* If there’s an example, don’t show the generated example.

**Before**

**After**
Screenshot 2023-10-27 at 14 29 27


marclave (approved):
ah this is gunna make a lot of docs look ✨ amazing

PR #384, opened 2023-10-27
refactor: test send request with proxy

hanspagel:
This PR improves sendRequest, the heart of the API client.

* I’ve added a test using it with a proxy and made the responses with/without a proxy look similar.
* I’ve also decided to keep the response body as a string. We shouldn’t care whether the response is JSON or not, except when rendering a preview.


marclave (approved):
a total eclipse of the heart (of the client)!

no better feeling that waking up on a friday to more test coverage

PR #385, opened 2023-10-27
fix: merge query strings from URL with given query strings

hanspagel:
With this PR query strings in the URL and given query strings are merged. 💅


marclave (approved):
let's go! this is nice 💯

PR #388, opened 2023-10-27
fix: minor styling issues with native web view

hanspagel:
This PR fixes some minor styling issues I’ve seen in the native web view on macOS.

* The CodeMirror component wasn’t full width.
* The address bar text color was always black.
* The response status code was always blue.


marclave (approved):
nice touch ups ✨

PR #389, opened 2023-10-27
feat: use variable name as placeholder for the server url

hanspagel:
When variables are empty, the server url looks wrong. Let’s use the variable name as the placeholder then.

https://github.com/scalar/scalar/assets/1577992/e3c9ff1d-8b46-4ca3-937a-2bd4bc834542


marclave (approved):
nice quick fix!

PR #396, opened 2023-10-29
Marc/make search hotkey configurable

marclave:
now the search modal hotkey can be configured, a good use case is your adding scalar to an existing site but want two searchbars

image


marclave:
@hanspagel maybe we also want to allow the hotkey to be disabled entirely... not sure

cameronrohani:
Hopefully no one makes it ⌘q 😅

hanspagel (approved):
Awesome! A customizability key feature.

PR #398, opened 2023-10-30
refactor: swagger editor reactivity

hanspagel:
This PR refactors how the data flows between the reference and the editor.

**Before**
We’ve had a handful of props, refs, computed refs and watchers. Debugging what triggered what became so hard, we introduced regressions every time we touched it.

Example: Checkout `main`, scroll down in the swagger editor, start typing. The editor scrolls to the top. 👎  Instead of fixing just this, let’s refactor:

**After**
This PR moves all refs and watchers to `useSpec()` and `useParser()`. Probably still hard to reason what triggers what, but at least it’s mostly only one file that you need to look at.

And this PR adds tests for `useSpec` and `useParser`.

### DIFF
--- a/.changeset/purple-books-impress.md
+++ b/.changeset/purple-books-impress.md
@@ -1,0 +1,8 @@
+---
+'@scalar/swagger-editor': patch
+'@scalar/use-codemirror': patch
+'@scalar/api-reference': patch
+'@scalar/api-client': patch
+---
+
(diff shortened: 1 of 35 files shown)


hwkr:
I'm so excited you're tackling this 🙌

hanspagel:
Sorry, I need to merge this to fix some critical yjs bugs on top of it.

PR #401, opened 2023-10-30
chore: update cdn link

marclave:


cameronrohani (approved):
This is huge (or should I say smol)

PR #402, opened 2023-10-31
feat: split out api references from the view

amritk:
Not quite done but up for comment.

@geoffgscott is this what you were thinking, did you want the editor slotted out this far? OR maybe its slotted inside `ApiReferences` and we just use a boolean to import it or not. 
@hanspagel need to run this by you as well. Also I need to add the editor a few more places where it is used. I think we might need one more wrapper that has the editor inside of it.

### DIFF
--- a/packages/api-reference/src/components/ApiReference.vue
+++ b/packages/api-reference/src/components/ApiReference.vue
@@ -9,10 +9,9 @@ import { deepMerge } from '../helpers'
 import { useParser, useSpec } from '../hooks'
 import { useDarkModeState } from '../hooks/useDarkModeState'
 import type { ReferenceConfiguration, ReferenceProps, Spec } from '../types'
-import ApiReferenceBase from './ApiReferenceBase.vue'
+import ApiReferenceLayout from './ApiReferenceLayout.vue'
 
 const props = withDefaults(defineProps<ReferenceProps>(), {
-  hasEditor: true,
   showSidebar: undefined,
   isEditable: undefined,
   footerBelowSidebar: undefined,
@@ -123,9 +122,8 @@ const { toggleDarkMode, isDark } = useDarkModeState()
 <template>
   <ThemeStyles :id="currentConfiguration?.theme" />
   <FlowToastContainer />
-  <ApiReferenceBase
+  <ApiReferenceLayout
     :currentConfiguration="currentConfiguration"
-    :hasEditor="hasEditor"
     :isDarkMode="isDark"
     :parsedSpec="parsedSpecRef"
     :rawSpec="rawSpecRef"
@@ -160,5 +158,5 @@ const { toggleDarkMode, isDark } = useDarkModeState()
     <template #footer>
       <slot name="footer" />
     </template>
-  </ApiReferenceBase>
+  </ApiReferenceLayout>
 </template>
--- a/packages/api-reference/src/components/ApiReference.vue
+++ b/packages/api-reference/src/components/ApiReference.vue
@@ -1,23 +1,13 @@
 <script setup lang="ts">
-import { useApiClientStore } from '@scalar/api-client'
-import {
-  type SwaggerEditor,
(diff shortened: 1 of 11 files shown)


projects/web/src/pages/ApiReferencePage.vue
```diff
@@ -38,5 +45,9 @@ const configuration = reactive<ReferenceConfiguration>({
   </div> -->
   <ApiReference
     :configuration="configuration"
-    @changeTheme="(theme: ThemeId) => (configuration.theme = theme)" />
+    @changeTheme="(theme: ThemeId) => (configuration.theme = theme)">
+    <template #editor="slotProps">
+      <LazyLoadedSwaggerEditor v-bind="slotProps" />
```
amritk, line 50: Yep think I went too far here, will pull this back in under a boolean prop

hanspagel:
Converted to draft while it’s work in progress. 🚧

hanspagel:
This will probably have a lot of conflicts with #398. I’ll take a closer look once #398 is in (and resolve conflicts), but I think this could be the right direction.

amritk:
Sounds good @hanspagel let me know if you need any help merging.

projects/web/src/pages/ApiReferencePage.vue
```diff
@@ -38,5 +38,6 @@ const configuration = reactive<ReferenceConfiguration>({
   </div> -->
   <ApiReference
     :configuration="configuration"
-    @changeTheme="(theme: ThemeId) => (configuration.theme = theme)" />
+    @changeTheme="(theme: ThemeId) => (configuration.theme = theme)">
```
amritk, line 41:
```suggestion
    :hasEditor="false"
    @changeTheme="(theme: ThemeId) => (configuration.theme = theme)">
```
Try passing `hasEditor`  = false here to see it with no editor.

packages/api-reference/src/components/ApiReference.vue
```diff
@@ -229,91 +222,36 @@ watch(
...
+  <ApiReferenceLayout
+    :currentConfiguration="currentConfiguration"
+    :isDarkMode="isDark"
+    :parsedSpec="parsedSpec"
+    :rawSpec="rawSpecRef"
+    :swaggerEditorRef="swaggerEditorRef"
+    @changeTheme="$emit('changeTheme', $event)"
+    @toggleDarkMode="toggleDarkMode">
+    <template #header>
+      <slot name="header" />
```
geoffgscott, line 244: I don't think we need a slot here (@hanspagel what do you think?) but whether we have a slot or not the content should be provided here (either as slot fallback or as direct slot content) same as it was before.

packages/api-reference/src/components/ApiReference.vue
```diff
@@ -325,309 +263,10 @@ function handleAIWriter(
...
-              @changeExample="swaggerEditorRef?.handleChangeExample"
-              @changeTheme="$emit('changeTheme', $event)"
-              @openSwaggerEditor="swaggerEditorRef?.handleOpenSwaggerEditor" />
-          </template>
-        </Content>
-      </div>
-      <div class="references-footer">
-        <slot name="footer" />
-      </div>
+        @startAIWriter="handleAIWriter" />}
```
geoffgscott, line 266:
```suggestion
        @startAIWriter="handleAIWriter" />
```

packages/api-reference/src/components/ApiReference.vue
amritk: I believe this was just copied over from how it was done before 
image

packages/api-reference/src/components/ApiReference.vue
amritk: strange lol

hanspagel (approved):
successfully untangled 🫡

marclave (approved):
huge improvement!

PR #403, opened 2023-10-31
property overflow fix

cameronrohani:
Before:
image

After: 
image
(scrollable ui)


hanspagel (approved):
You’re setting a good example.

PR #406, opened 2023-10-31
Make theme priority instead of default theme

cameronrohani:
Found some misc places where --default theme was overwriting --theme, this should fix that


marclave (approved):
let's make this a priority to get in!

PR #369, opened 2023-10-25
Remove unused icons

cameronrohani:

PR #407, opened 2023-10-31
fix: re-add missing menu icon

jonataw:
This PR adds the `Menu.svg` icon that was removed in #369 as it is still used: https://github.com/scalar/scalar/blob/9b1e2653f0c46f053681bc126dec5a631e281658/packages/api-reference/src/components/MobileHeader.vue#L15


marclave (approved):
great catch @jonataw !

Appreciate the contribution, will get this in once the build finishes and do a release ✨

PR #408, opened 2023-11-01
fix: don’t show a select for a single auth method

hanspagel:
**Before**
Screenshot 2023-11-01 at 09 40 17

**After**
Screenshot 2023-11-01 at 09 38 56


hwkr:
You're doing gods work

marclave (approved):
✨ ah this is great

PR #410, opened 2023-11-01
users not using themed css fallbacks

cameronrohani:
quick fixes for two common things I've seen 

Making our --theme-background-1 transparent (which makes search + client unusable)

And making darklight display: none;


hwkr (approved):
Might be nice to pull the fallback colors out into variables too

marclave (approved):
amazing work @cameronrohani ! ✨

PR #413, opened 2023-11-02
chore: don’t run ci on push to main

hanspagel:
CI is running on push to main, but we don’t do anything with the output, so we could also not run it on push to main. :)

It’ll still running on all PRs.


marclave (approved):
Fantastic :) our carbon footprint just got a bit smaller 🍃

PR #414, opened 2023-11-02
fix: do not touch the content if nothing is configured

hanspagel:
With #398 I introduced a regression: When a Swagger spec is added, and the config changes, the content is overwritten - even if there’s not configuration for it.

This PR adds a test and a fix.


marclave (approved):
fixes and a test! huge ✨

PR #416, opened 2023-11-02
Marc cam/add oas version spec version to title

marclave:
image


hanspagel (approved):
I like this specific version of your change.

PR #421, opened 2023-11-06
feat: support more mime types for request bodies

hanspagel:
This PR adds support for other MIME types in request bodies:

- [x] application/json
- [x] application/octet-stream
- [x] application/x-www-form-urlencoded ⚠️
- [x] application/xml
- [x] multipart/form-data ⚠️
- [x] text/plain

JSON and XML have good support already. The other formats have better support than before this PR, but they all have their own complexity.

⚠️ Some formats need the response body as params (`foo=bar`), not as an object (`{foo: 'bar'}`) and support custom delimiters ("boundries"). This PR doesn’t include those features, but it’s better than before (showing nothing).

**Preview**
Screenshot 2023-11-10 at 15 37 38
Screenshot 2023-11-10 at 15 37 33
Screenshot 2023-11-10 at 15 37 28
Screenshot 2023-11-10 at 15 37 22


marclave (approved):
amazing work @hanspagel ! This is great, let's create some tickets for the functionality you mentioned we can add in the future:

- application/x-www-form-urlencoded ⚠️
- multipart/form-data ⚠️ 

I know these have cascading effects with the rest api client, so we can talk about this in slack or something ✨

PR #422, opened 2023-11-06
chore: add SSG example

hanspagel:
We break SSG envs too often. This PR adds a SSG demo which is built in CI and should actually fail in this PR so we can test the SSG build here.

I’ll create a separate PR with the fix.

EDIT: Rebased and CI is green, awesome. ✅

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -14,3 +14,4 @@ pnpm-lock.yaml
 cdn
 
 fixtures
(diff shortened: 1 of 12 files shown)


marclave (approved):
thanks @geoffgscott and @hanspagel for this work! glad we have ssg checks now ✨

PR #425, opened 2023-11-06
chore: align tsconfigs

hanspagel:
This PR aligns all tsconfigs. I’ve cherry-picked and cleaned up the related commits from #419. 🙌

PR #427, opened 2023-11-06
chore: custom package json formatting

hanspagel:
Took the custom formatting script from #425 and added it to the repository. 🙌

Changed some minor details and used vite-node instead of bun. I don’t really care, but vite-node is already used 7 other packages in the repository.

This PR doesn’t actually apply the formatting, we’ll do this in a separate PR when the PR list is shorter.

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,6 +18,7 @@
     "syncpack": "11.2.1",
     "turbo": "1.10.15",
     "typescript": "5.2.2",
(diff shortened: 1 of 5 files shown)


geoffgscott (approved):

marclave (approved):
🙌 i just fixed up the conflicts so will merge once build passes

PR #428, opened 2023-11-06
fix: hide version if it’s not set

hanspagel:
The version field is required according to the OpenAPI spec, but the parser won’t complain. So let’s ignore it, if there’s no version defined.

We don’t need this for the OpenAPI version. The parser requires this field.

**Before**
Screenshot 2023-11-06 at 15 05 57

**After**
Screenshot 2023-11-06 at 15 07 06


marclave (approved):
awesome!  ✨

PR #430, opened 2023-11-06
chore: remove deleted icons from type definition

hwkr:


packages/api-reference/src/components/Icon/icons/ExternalLink.svg
hwkr, line 1: Turns out this was being used

hanspagel (approved):
You’re the icon here.

PR #431, opened 2023-11-07
fix: examples in collaborative editing mode

hanspagel:
WIP

We introduced a regression where loading examples created duplicate content. This PR has the following changes to fix this and related bugs.

* Fix: The editor is mounted even when another tab than 'Swagger Editor' is active (otherwise content can’t be synced in collaborative editing mode)
* Fix: Whitespace is trimmed from the content before it’s parsed.
* Refactor: The SwaggerEditor doesn’t “know” about the examples anymore, it just deals with strings. All examples have been moved to the GettingStarted component.
* Bonus: Collaborative editing can be configured in the dev toolbar now
Screenshot 2023-11-07 at 12 20 44

### DIFF
--- a/packages/api-reference/src/hooks/useParser.ts
+++ b/packages/api-reference/src/hooks/useParser.ts
@@ -65,7 +65,9 @@ export function useParser({
       return
     }
 
-    parse(value)
+    const processedValue = value.trim()
+
+    parse(processedValue)
       .then((spec) => {
         errorRef.value = null
 
--- a/packages/swagger-editor/src/components/SwaggerEditor/SwaggerEditor.vue
+++ b/packages/swagger-editor/src/components/SwaggerEditor/SwaggerEditor.vue
@@ -34,14 +34,7 @@ const swaggerEditorHeaderRef = ref<typeof SwaggerEditorHeader | null>(null)
 
 const awarenessStates = ref<StatesArray>([])
 
-const rawContent = ref('')
-
 const handleContentUpdate = (value: string) => {
-  if (value === rawContent.value) {
-    return
-  }
-
-  rawContent.value = value
   emit('contentUpdate', value)
 }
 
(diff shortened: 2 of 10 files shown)


packages/swagger-editor/src/components/SwaggerEditor/SwaggerEditorInput.vue
```diff
@@ -108,7 +110,7 @@ const codeMirrorRef = ref<typeof CodeMirror | null>(null)
   <div class="swagger-editor-input">
     <CodeMirror
       ref="codeMirrorRef"
-      :content="value"
+      :content="!props.hocuspocusConfiguration ? value : undefined"
```
hanspagel, line 113: This one caused duplicate content.

packages/api-reference/src/hooks/useParser.ts
```diff
@@ -65,7 +65,9 @@ export function useParser({
       return
     }
 
-    parse(value)
+    const processedValue = value.trim()
```
hanspagel, line 68: I’ve run into issues with some kind of additional whitespace sometimes. That fixed it.

packages/swagger-editor/src/components/SwaggerEditor/SwaggerEditor.vue
```diff
@@ -164,10 +134,10 @@ defineExpose({
       {{ formattedError }}
     </SwaggerEditorNotification>
     <SwaggerEditorInput
-      v-if="activeTab === 'Swagger Editor'"
+      v-show="activeTab === 'Swagger Editor'"
```
hanspagel, line 137: This caused the duplicate content. 🫠

hanspagel:
When symlinking to our app this throws a duplicate Yjs instance error, but the build doesn’t include Yjs - I’ve double-checked this. I don’t know how to debug what’s not there. My best guess it’s somehow related to symlinking/caching or an issue in our app.

I hope I don’t regret that, but I think we should release and see if the bug still appears then. 👀

PR #432, opened 2023-11-07
feat: Yaml syntax highlighting

hanspagel:
I’m not sure whether we had this already, but this PR adds Yaml syntax highlighting to the Swagger editor. 💅


marclave (approved):
ah this is so much better, nice work ✨

PR #436, opened 2023-11-07
fix: show active state for headings in description

hanspagel:
This PR fixes the active state for headings taken from the description.

With this PR the description is split in multiple sections (based on the headings) and those sections are wrapped in an `IntersectionObserver` component.

**Example**
This is a good file to test it: https://raw.githubusercontent.com/outline/openapi/main/spec3.json

**Preview**
https://github.com/scalar/scalar/assets/1577992/5208affc-2f50-4cba-9f40-a3f2666dd731


marclave (approved):
@hanspagel 👏 this is so amazing, great fix.

PR #437, opened 2023-11-08
chore: store content in local storage

hanspagel:
We used to store the Swagger editor content in local storage but broke that recently. But I miss it, it made the dev cycle so much faster.

I’ve added the code to cache the content to the example and removed it from the package.
And I’ve added a `updateContent` event to the `ApiReference` component.


marclave (approved):
love this ✨ makes testing way better

PR #439, opened 2023-11-09
feat(doc-1284): OSS component library

amritk:
Alright so this is by no means complete, but I think its good enough to get merged and we can just continuously chip away at it. We are testing out tailwind here! But the build system and all that jazz is still #todo so we cant consume it yet. You can however try it out with
```bash
pnpm storybook
```

Also as always open to feedback on all of the decisions!! A lot of the setup files are still the defaults, can definitely change those to the library standard.

Update: Alright its now building and we are consuming on the admin dashboard! 
image

Still a couple more things to do:
- fix test
- remove tests and storybook from build
- host storybook somewhere
- add lint/build etc to ci
- put up npm package


hanspagel (changes requested):
Awesome initiative! There seems to be something wrong with the TypeScript config, I receive an error "tsconfig.app.json can’t be found" and an index.html is needed for the build:

```
 Could not resolve entry module "index.html".
  error during build:
  RollupError: Could not resolve entry module "index.html".
```

We should set the build config to library mode, an index.html isn’t needed then.

amritk:
@hanspagel yea I didnt touch the build side yet, can check that out next 🙏🏾

amritk:
Building now, but not what we want

packages/components/src/components/ScalarTextField/ScalarTextField.stories.ts
```diff
@@ -0,0 +1,57 @@
...
+  }),
+} satisfies Meta<typeof ScalarTextField>
+
+export default meta
+type Story = StoryObj<typeof meta>
+
+export const Base: Story = {
+  args: {
+    modelValue: '',
+    label: 'Scaler Text Field',
```
hwkr, line 28:
```suggestion
    label: 'Scalar Text Field',
```
(This seems to have been copied a bunch of places)

packages/components/src/components/ScalarIconButton/ScalarIconButton.vue
```diff
@@ -0,0 +1,25 @@
+<script setup lang="ts">
+import { type ButtonProps, ScalarButton } from '../ScalarButton'
+import { type Icon, ScalarIcon } from '../ScalarIcon'
+
+withDefaults(
+  defineProps<{
+    label: string
```
hwkr, line 7: I think the label could be a slot here that's `sr-only`

packages/components/src/components/ScalarButton/ScalarButton.vue
```diff
@@ -0,0 +1,114 @@
...
+const label = computed(() =>
+  (slots.default?.()?.[0]?.children as string)?.trim(),
+)
+const isIconOnly = computed(() => Boolean(slots.icon && !slots.default))
+</script>
+
+<template>
+  <button
+    :aria-disabled="isDisabled"
+    :aria-label="label || title"
```
hwkr, line 85: Instead of setting the aria-label you can just add an sr-only span.

Something like:

`<span class="sr-only"><slot /></span>`

amritk: Going to remake this branch due to conflicts, but what is the advantage of `sr-only` over aria-label? I thought `aria-label` would be better no? isn't `sr-only` a hack?

marclave (approved):
let's go!

hanspagel (approved):

PR #441, opened 2023-11-09
feat: CommonJS builds for `@scalar/api-client-proxy` and `@scalar/echo-server`

hanspagel:
This PR fixes wrong paths for the api-client-proxy introduced in #426 and adds CommonJS builds for `@scalar/api-client-proxy` and `@scalar/echo-server`.

I’m a fan of ESM-only packages, but I needed them in another environment and using ESM-only there was just too big of a hurdle. And there are probably other people preferring CommonJS builds, too.

### DIFF
--- a/.changeset/fair-avocados-speak.md
+++ b/.changeset/fair-avocados-speak.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client-proxy': patch
+---
+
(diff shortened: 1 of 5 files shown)


marclave (approved):
✨ stoked to have this helping out for neat things in the future for @scalar

PR #447, opened 2023-11-09
feat: detect variables in urls

hanspagel:
With this PR variables used in server URLs are detected and inputs are rendered for them, even if they are not properly configured in the spec.

This PR also fixes a bug where single URLs didn’t had the variables inputs.


marclave (approved):
nice fix! and tests, the best kind of PR :)

PR #448, opened 2023-11-09
chore: better fallbacks for the name

hanspagel:
It’s somehow confusing how the fallback for the operation.name is chosen. With this PR it’s done in a single place, and the order is:

1. `operationId` or…
2. `summary` or …
3. `path`

In the SearchModal, the `path` is only added if it’s not used for the name fallback already.

Screenshot 2023-11-09 at 16 50 51


marclave (approved):
alignment on the fallbacks, lets go!

PR #450, opened 2023-11-09
fix: add fallback if no summary or key is provided for examples

marclave:
before if you had no key or summary it would be an empty list, now we provide a `scalarExampleName` generated to the object with the key of the example


hanspagel (approved):
You’re making a great example here

PR #468, opened 2023-11-13
fix cards overlapping header

cameronrohani:
Before:
https://github.com/scalar/scalar/assets/6201407/45d88859-fdd9-4cf4-9a81-6b9d93deeb0d

After:
Uploading after.mov…

++ improve card max height

Before:
image

After:
image

### DIFF
--- a/packages/api-reference/src/components/Card/Card.vue
+++ b/packages/api-reference/src/components/Card/Card.vue
@@ -13,7 +13,9 @@
   background: var(--theme-background-2, var(--default-theme-background-2));
   display: flex;
   flex-direction: column;
-  max-height: calc(50vh - 30px);
+  max-height: calc(
+    ((var(--full-height) - var(--theme-header-height, 0px)) - 60px) / 2
+  );
   position: relative;
 }
 </style>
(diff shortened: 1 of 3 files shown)


packages/api-reference/src/components/Card/Card.vue
```diff
@@ -13,7 +13,9 @@
   background: var(--theme-background-2, var(--default-theme-background-2));
   display: flex;
   flex-direction: column;
-  max-height: calc(50vh - 30px);
+  max-height: calc(
+    ((var(--full-height) - var(--theme-header-height, 0px)) - 60px) / 2
```
hwkr, line 17: You should add a comment when you add math like this, also I think you want `--refs-header-height` with no fallback.
```suggestion
    ((var(--full-height) - var(--refs-header-height)) - 60px) / 2
```

packages/api-reference/src/components/Content/ReferenceEndpoint/ReferenceEndpoint.vue
```diff
@@ -45,6 +45,6 @@ defineProps<{
 <style scoped>
 .examples {
   position: sticky;
-  top: 24px;
+  top: calc(var(--theme-header-height, 0px) + 24px);
```
hwkr, line 48:
```suggestion
  top: calc(var(--refs-header-height) + 24px);
```

marclave (approved):
a victory lap! :)

marclave:
also thanks @hwkr for the review!

PR #471, opened 2023-11-14
fix word breaks in .markdown

cameronrohani:
before:
image

After:
image

+ fix line height for headings


marclave (approved):
awesome work :)

PR #472, opened 2023-11-15
feat: only load swagger editor if we are editable

amritk:
Only load the editor if we are editable


amritk:
Closed in favor of https://github.com/scalar/scalar/pull/474

PR #474, opened 2023-11-15
feat: lazyload editor

amritk:
Only load the editor if we are editable. This needs testing on the preview side before merge

### DIFF
--- a/packages/api-reference/src/components/ApiReferenceBase.vue
+++ b/packages/api-reference/src/components/ApiReferenceBase.vue
@@ -31,9 +31,11 @@ const emit = defineEmits<{
 /**
  * The editor component has heavy dependencies (process), let's lazy load it.
  */
-const LazyLoadedSwaggerEditor = defineAsyncComponent(() =>
-  import('@scalar/swagger-editor').then((module) => module.SwaggerEditor),
-)
+const LazyLoadedSwaggerEditor = props.configuration?.isEditable
+  ? defineAsyncComponent(() =>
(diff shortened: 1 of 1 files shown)


hanspagel:
Awesome! This has one side effect, that we might want or might not want:

The SwaggerEditor is responsible to sync changes with the collaborative editing backend. With the current implementation, changes from remote are synced in Preview. So you can edit the reference in one tab, and preview the changes in another tab. 

With this PR, the preview becomes static. What do we prefer, @marclave?

amritk:
Yep we wanted to test that, which led us into the rabbit hole of trying to better link the packages. As much as I don't want it, the only option I see is adding another config option (keepEditorAlive or something). We really don't want to include the editor for users that don't need it as its so heavy! Is there any other way?

hanspagel:
Actually, it’s not as heavy as I thought. We need CodeMirror anyway. And in most cases we need the parser, too. So there’s not much overhead.

amritk:
Wait we use code mirror on standalone references? We got a better solution as well! Will implement it later

hanspagel:
Yeah, we use it for syntax highlighting, but also as an input in the embedded API client.

marclave:
@amritk @hanspagel 

we defs want the preview to be dynamic, I wonder how much we are saving here if we end up removing this.

What do y'all think about it?

marclave:
sorry about the late comment!

amritk:
Yea I got an idea of what we can do, without losing it on the preview side and dropping it on the standalone side. Low priority though can even close this PR and we can come back to it when we do more perf improvements

PR #478, opened 2023-11-15
Amrit/doc 1328 deep linking anchor tags doesnt work in references

amritk:
Basically we need to pop open the sections before we can scroll to the element. Tested on OSS side, needs testing on org.

Also could not for the life of me avoid that setTimeout short of using a ref and watching it (making it worse imo). Open to a better way of doing this. If you switch to nextTick, the page scrolls down so the section is partially showing only.


marclave (approved):
awesome! tested let's get this in :)

PR #479, opened 2023-11-15
fix: remove empty content update

hwkr:
I don't know if this is important but it breaks a lot of stuff...

### DIFF
--- a/.changeset/strange-dogs-hope.md
+++ b/.changeset/strange-dogs-hope.md
@@ -1,0 +1,5 @@
+---
+'@scalar/swagger-editor': patch
+---
+
+fix: remove empty content update deleting rendered refs
--- a/packages/swagger-editor/src/components/SwaggerEditor/SwaggerEditorInput.vue
+++ b/packages/swagger-editor/src/components/SwaggerEditor/SwaggerEditorInput.vue
@@ -39,9 +39,6 @@ watch(
       yCodeMirrorExtension.value = null
     }
 
-    // Reset the content, before loading a new Y.js document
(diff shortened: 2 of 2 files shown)


amritk:
I agree, this thing was breaking standalone cdn references. But I'll leave to @hanspagel since he knows what its for

hanspagel:
Yeah, without it, we’ll have duplicate content when switching between different documents. There’s probably a better way to reset the editor when switching documents, but I failed with a few different attempts (e.g. using Vue’s `key`).

It’s tough … :)

Can you describe the issues you’re seeing so I can test whatever solution we come up with against it?

hwkr:
> Yeah, without it, we’ll have duplicate content when switching between different documents. There’s probably a better way to reset the editor when switching documents, but I failed with a few different attempts (e.g. using Vue’s `key`).
> 
> It’s tough … :)
> 
> Can you describe the issues you’re seeing so I can test whatever solution we come up with against it?

This causes the content to disappear whenever the hocuspocus config changes. For example in the app when you switch between guides and references the hocuspocus token gets updated and we lose the rendered content.

hwkr:
@hanspagel I'm also not seeing any duplication when switching documents on this branch

https://github.com/scalar/scalar/assets/6374090/9651e717-d719-4279-b420-0690f0a31050

hanspagel:
Yeah, let’s get this in! Hasn’t been a proper fixed and it seems it’s not even working. Thanks for finding this! 🤝

hanspagel (approved):

PR #484, opened 2023-11-16
chore: store api reference demo editor content in local storage

hanspagel:
With this PR the editor content of the demo is stored in local storage (again).

Makes development way easier.


marclave (approved):
Great for testing i agree!

PR #486, opened 2023-11-16
feat: add support for oneOf and allOf in the response schema

hanspagel:
This PR adds support for the use of `oneOf` and `allOf` in the schema for responses.

When generating example responses from the schema, the first entry in `oneOf` is used.

For `allOf` we generated example for all given schemas and try to merge them.


marclave (approved):
love this ✨

PR #487, opened 2023-11-16
style: do not show empty example requests

hanspagel:
We can’t generate example requests e. g. if an invalid URL is given. In those (rare) cases, we can hide the example request card, because it’s empty anyway.

**Before**
Screenshot 2023-11-16 at 11 44 40

**After**
Screenshot 2023-11-16 at 11 47 43

### DIFF
--- a/.changeset/wild-taxis-call.md
+++ b/.changeset/wild-taxis-call.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
+chore: don’t show the example request card if it’s empty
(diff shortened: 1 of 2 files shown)


marclave (commented):
@cameronrohani thoughts on this?

@hanspagel I wonder if we should just default to using the URL the docs are hosted on?

hanspagel:
I haven’t seen a case where the window.url was the actual API URL. It’s possible, but it doesn’t happen very often it seems.

IMHO not showing examples is better than having a wrong URL in most cases. 🤷

hanspagel:
Let’s make a decision here: Close or Approve. :v: @cameronrohani @marclave

PR #488, opened 2023-11-16
feat: swagger 2.0 host and schemes

hanspagel:
In Swagger 2.0 files servers are configured like this:

```json
{
  "host": "api.example.com",
  "schemes": [
    "https"
  ]
}
```

This PR adds support for this syntax.

### DIFF
--- a/packages/api-reference/src/components/Content/Content.vue
+++ b/packages/api-reference/src/components/Content/Content.vue
@@ -46,7 +46,11 @@ const fallBackServer = useRefOnMount(() => {
 const localServers = computed(() => {
   if (props.parsedSpec.servers && props.parsedSpec.servers.length > 0) {
     return props.parsedSpec.servers
-  } else if (props.parsedSpec.host && props.parsedSpec.schemes.length) {
+  } else if (
+    props.parsedSpec.host &&
+    props.parsedSpec.schemes &&
+    props.parsedSpec.schemes.length > 0
+  ) {
     return [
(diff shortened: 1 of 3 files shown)


packages/api-reference/src/components/Content/Content.vue
```diff
@@ -46,6 +46,14 @@ const fallBackServer = useRefOnMount(() => {
 const localServers = computed(() => {
   if (props.parsedSpec.servers && props.parsedSpec.servers.length > 0) {
     return props.parsedSpec.servers
+  } else if (
+    props.parsedSpec.host &&
+    props.parsedSpec.schemes &&
+    props.parsedSpec.schemes.length > 0
```
marclave, line 52: could this change to

```
props.parsedSpec.schemes?.length > 0
```

marclave (changes requested):

packages/api-reference/src/components/Content/Content.vue
hanspagel: We could replace it with just `props.parsedSpec.schemes?.length`, but not with `props.parsedSpec.schemes?.length > 0`, because `undefined.length` can’t return a number. 🤡 

I tried to keep it similar to the lines before, where we have a `…length > 0`. Anyway, whatever you think is right.

marclave: ah good catch, i forgot about that!

marclave (approved):

PR #489, opened 2023-11-16
feat: fallback for http auth

hanspagel:
With this PR bearer auth is taken as the default http auth method (that’s what Swagger UI does, too). E.g. this is enough:

```
  "security": [
    {
      "http": []
    }
  ],
```

Source: https://raw.githubusercontent.com/outline/openapi/main/spec3.json

**Before**
Screenshot 2023-11-16 at 12 27 54

**After**
Screenshot 2023-11-16 at 12 25 58

### DIFF
--- a/.changeset/afraid-carpets-watch.md
+++ b/.changeset/afraid-carpets-watch.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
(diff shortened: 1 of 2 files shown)


marclave (approved):
ah this is fantastic!

PR #498, opened 2023-11-16
chore: force loose peer dependencies

hanspagel:
We’re using pinned versions in all package.json. With this PR the peer dependencies are set to allow minor updates (`^`).

I might be wrong, but I think this allows to have new versions of the peer dependencies in userland code bases. And if all packages follow semver, that shouldn’t break anything. And if it breaks something at some point, the solution is to pin a version of a specific dependency.


amritk (approved):
🦺

amritk (approved):
🧷

marclave (approved):
I dig it ✨

PR #501, opened 2023-11-17
style: custom Fastify theme

hanspagel:
With this PR a custom Fastify theme is used for the plugin. You can still use one of the other themes, but the new default is the Fastify theme.

Preview: https://fastify.apidocumentation.com/


marclave (approved):
great idea @hanspagel ! we can do the same with all integrations going forward 👀

PR #502, opened 2023-11-17
feat: add support for models with allOf, anyOf, oneOf …

hanspagel:
I noticed some Models were empty and found out we’re not dealing with schemas that define anyOf, allOf, oneOf … on the top level. This PR fixes it.

**Before**
Screenshot 2023-11-17 at 11 40 55

**After**
Screenshot 2023-11-17 at 11 38 19


marclave (approved):
a small change with big impact, awesome work @hanspagel

PR #504, opened 2023-11-17
chore: remove babel

hanspagel:
This PR removes Babel, because … we don’t use it. 😅


marclave (approved):
good cleanup :sparkles:

PR #507, opened 2023-11-17
chore: do not use the operationId as name

hanspagel:
Okay … that’s kind of a long journey. I experimented with different fallbacks for the name, but based on the user feedback I don’t think we should use the operationId as a potential fallback for the operation name.

With this PR we’re looking for the `summary` and if it’s not available, we just use the path.

That change gives users still control, but avoids confusion and the use of (in most cases technical) operationIds.


marclave (approved):
The journey continues :)

PR #520, opened 2023-11-21
feat: [DOC-1357] allow test button + api client to be disabled, fix #519

### LINKED ISSUE #519: [DOC-1357] Allow test button + api client to be disabled
### Expected Behavior

A user should be able to add or remove testing button + api client :) 

### Actual Behavior

Testing is always enabled so the button + client are always usable

### Reproduction steps

```bash
View any published references or latest main with a test spec
```

### Version

latest

Chrome

<sub>[DOC-1357](https://linear.app/api-docs/issue/DOC-1357/allow-test-button-api-client-to-be-disabled)</sub>

hanspagel:
This PR does two things:

1. The configuration is put into a global store now. So we don't need to do prop drilling all the time.
2. This feature is used to add a `withApiClient` flag to the configuration and show/hide the `Test Request` button.

https://github.com/scalar/scalar/assets/1577992/07e3239d-155f-4fd5-8439-f040c9d5dd52

### DIFF
--- a/.changeset/breezy-jars-buy.md
+++ b/.changeset/breezy-jars-buy.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
+feat: add `withApiClient` flag to the configuration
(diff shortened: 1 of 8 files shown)


marclave (approved):
nice work, glad it can be disabled now ✨

PR #522, opened 2023-11-21
fix: don’t show an empty form if "No Authentication" is selected

hanspagel:
WIP: Need to fix the TS issues.

What I changed in #489 was a mistake. This produced empty forms if "Not Authentication" is selected.

This PR adds an Authentication story 💅 and fixes it. The initial issue was actually caused by a lacking spec.

**Before**
Screenshot 2023-11-21 at 14 43 56

**After**
Screenshot 2023-11-21 at 14 57 33

### DIFF
--- a/.changeset/angry-worms-smell.md
+++ b/.changeset/angry-worms-smell.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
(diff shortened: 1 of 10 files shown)


marclave (approved):
amazing! nice fix, and we it's storytime. win win ✨

PR #526, opened 2023-11-21
fix: set codemirror peer dependency version

hwkr:


marclave (approved):
🙏 awesome work on a speedy fix

PR #529, opened 2023-11-22
feat: export tailwind config from components package

amritk:
Also fixed textField styles and hoisted devTools packages to root. Prettier config still exists in here due to tailwind config


marclave (approved):
awesome :)

PR #534, opened 2023-11-22
chore: change how we deal with dependencies

hanspagel:
This PR changes how we deal with dependencies.

**dependencies**
All `dependencies` will be installed in user land projects. npm is allowed to install different versions though.

**devDependencies**
The `devDependencies` includes dependencies, that are not needed in the user’s project. For example @types/ packages.

**peerDependencies**
Everything under `peerDepeondencies` is a strict requirement. npm isn’t allowed to install something else.

This change is not as big as it seems, but it has the high risk of breaking things. Upside: We just did it wrong and this PR will bring use closer to how it should be. And this will probably cause less issues eventually.

**Tests**
- [x] Examples work
- [x] Build works
- [x] SSG builds
- [x] Tests run
- [x] CI is green
- [x] Symlinking the package works
- [x] Netlify preview works 

Let's relrease it! I just assume there are issues with it and I’m up to fix them.


hwkr:
I feel like I don't really have a good enough understanding of how this works / the motivations to weigh in on it. I'm in favour of anything that reduces our conflict issues though.

PR #536, opened 2023-11-23
fix(getApiClientRequest): added missing path parameter variable and default values.

### LINKED ISSUE #531: Endpoint-level parameters not provided as variables in API client
### Expected Behavior

This is an extension of issue #499. While endpoint-level parameters are now properly documented, they do not show up as variables in the API client like method-level parameters do, even if they are required.

The API client should populate the variables with endpoint-level parameters just like it does for method-level parameters. Ideally, they should also use the default values when provided in the spec.

### Actual Behavior

The variables for endpoint-level parameters must be manually added for every request.

### Reproduction steps

```bash
1. Go to this API spec: https://raw.githubusercontent.com/gravitee-io/gravitee-api-management/master/gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/resources/openapi/openapi-apis.yaml
2. Copy the spec and paste it into the demo: https://docs.scalar.com/swagger-editor
3. Go to preview and go to the first endpoint "List APIs"
4. Select "Test Request"
5. There is no variable for endpoint-level parameter envId
```

### Version

Demo and/or CDN

Chrome, Safari (macOS)

Mohib834:
## Description

The API client should populate the variables with endpoint-level parameters just like it does for method-level parameters. Also, it should use the default values when provided in the spec.

Resolves #531

### DIFF
--- a/packages/api-reference/src/helpers/getApiClientRequest.ts
+++ b/packages/api-reference/src/helpers/getApiClientRequest.ts
@@ -38,8 +38,6 @@ export function getApiClientRequest({
     ...operation.pathParameters,
   ]
 
-  console.log('o', operation)
-
   return {
     id: operation.operationId,
     name: operation.name,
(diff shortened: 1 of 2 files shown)


marclave (approved):
awesome fix! ✨

PR #540, opened 2023-11-24
fix: only activate hotkey if component is active

hwkr:
Prevents search modal hotkey from being active if the component is in a `<keepalive>`  (like our references are) but deactivated.

Before:

https://github.com/scalar/scalar/assets/6374090/41a34bbb-05dc-4c20-8bb2-7a29e6e991b4

After:

https://github.com/scalar/scalar/assets/6374090/556bae47-71f8-4bca-bbf4-4529a8a7f688

### DIFF
--- a/packages/api-reference/src/components/SearchButton.vue
+++ b/packages/api-reference/src/components/SearchButton.vue
@@ -4,6 +4,7 @@ import { useModal } from '@scalar/use-modal'
 import { isMacOS } from '@scalar/use-tooltip'
 import { type Spec } from 'src/types'
 
+import { useActive } from '../hooks/useActive'
 import FlowIcon from './Icon/FlowIcon.vue'
 import SearchModal from './SearchModal.vue'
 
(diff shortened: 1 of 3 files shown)


marclave:
Ah this is so good!!! Great work @hwkr

marclave (approved):

packages/api-reference/src/components/SearchButton.vue
```diff
@@ -17,12 +18,15 @@ const props = withDefaults(
...
 
+const { isActive } = useActive()
+
 const modalState = useModal()
 
 useKeyboardEvent({
   keyList: [props.searchHotKey],
   withCtrlCmd: true,
   handler: () => (modalState.open ? modalState.hide() : modalState.show()),
+  active: () => isActive.value,
```
hwkr, line 29: Maybe keyboard shortcuts should always check this? I dunno @hanspagel ?

PR #561, opened 2023-11-24
fix: remove leaky header css

hwkr:


packages/api-reference/src/components/Sidebar.vue
```diff
@@ -508,12 +508,4 @@ const setRef = (el: SidebarElementType, id: string) => {
     padding-top: 12px;
   }
 }
-
```
hwkr, line 511: I don't think this was doing anything (except breaking stuff)

cameronrohani (approved):
😎

marclave (approved):
awesome ✨

PR #562, opened 2023-11-24
fix: remove editor tabs

marclave:
rolling back the editor interface to simplify things, we can in the future add the AI writer + tabs as slots :) 

image


geoffgscott (commented):
We should move getting started to the references. Might also be good to slot the buttons while we are here. Could be done in a separate PR though

marclave:
I'm happy to fast follow with that :)

cameronrohani:
Awesome work, let's make sure we default to showing the getting started page on mobile:
image

Also let's make sure the buttons are spaced like how they were previously :) 
image

marclave:
> Awesome work, let's make sure we default to showing the getting started page on mobile: image
> 
> Also let's make sure the buttons are spaced like how they were previously :) image

ah good idea, maybe we just remove the ai writer tab and just show the getting started tab on mobile?

marclave:
closing in favor of https://github.com/scalar/scalar/pull/589

PR #563, opened 2023-11-24
feat: Initial vite config cleanup

geoffgscott:
Migrates towards using peer dependencies to define externals.


marclave (approved):
nice cleanup! 🧹

PR #565, opened 2023-11-26
Marc/remove tableau example

marclave:
we can add more smaller examples, but tableau is way too big to be shipping :) 

before:
image

image

after:
image


hanspagel (approved):
🤏

PR #566, opened 2023-11-26
feat: reduce fastify scalar bundle size by 800kb

marclave:
alright, I would LOVE assistance here on vite + rollup. But it appears in library mode vite [can't minify es builds](https://github.com/vitejs/vite/pull/6314), but jsdelivr does this automatically for our cdn.

https://cdn.jsdelivr.net/npm/@scalar/api-reference@1.2.5/+esm

When I added terser to our minification, we shaved off **800KB**!

before (**3.9MB**)[before gzip]:
image

after (**3.1MB**) [before gzip]:
image

### DIFF
--- a/.changeset/swift-points-report.md
+++ b/.changeset/swift-points-report.md
@@ -1,0 +1,11 @@
+---
+'@scalar/fastify-api-reference': minor
+'@scalar/use-keyboard-event': minor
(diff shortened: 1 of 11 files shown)


hanspagel (approved):
Tiny addition, tiny bundle. 🤜 🤛

Issue #470, opened 2023-11-14
[DOC-1330] Allow SEO Customization


marclave:
### Expected Behavior

A user should be able to pass in information to the configuration object to configure meta information

let's expose all the options here

[https://nuxt.com/docs/api/composables/use-seo-meta](https://nuxt.com/docs/api/composables/use-seo-meta)

### Actual Behavior

currently you can't change anything :)

### Reproduction steps

```
1. try to change the title of the API References
2. You can't unless you use external javascript
```

### Version

latest

### What browsers are you seeing the problem on?

*No response*

<sub>From [SyncLinear.com](https://synclinear.com) | [DOC-1330](https://linear.app/api-docs/issue/DOC-1330/allow-seo-customization)</sub>

PR #572, opened 2023-11-27
feat: pass custom CSS to the fastify plugin

hanspagel:
With this PR:

* The fastify plugin uses the browser version from @scalar/api-reference, no extra build step required anymore. :)
* Minfies the @scalar/api-reference browser build (4.4 mb > 2.2 mb)
* The EJS templating for the fastify plugin is removed.
* The `pageTitle` option is removed, because we support more attributes through the ReferenceConfiguration object already #470 
* A fastify theme is added as the default custom CSS. 💅 

And: To align the functionality, the Express middleware supports `content()` callbacks now.


marclave (approved):
AMAZING work @hanspagel ✨

Issue #570, opened 2023-11-27
When switching the spec (or the operation?), the API client response should be reset


hanspagel:
it doesn't clear the response when you switch specs which is a little confusing.

_Originally posted by @hwkr in https://github.com/scalar/scalar/issues/557#issuecomment-1826021461_

PR #573, opened 2023-11-27
fix: URL doesn’t match the active sidebar item (DOC-1361, #296)

### LINKED ISSUE #296: Set the URL to the active sidebar item
Currently, the IntersectionObserver sets the URL. The URL is then set to the section that’s becoming visible, not to the active item (which is the first item in the sidebar, where the section visible).

hanspagel:
This PR fixes #296 and #570. But let’s not make the easy change, but make the change easy. :)

With this PR all the navigation logic is moved to a new `useNavigation` composable, which now is responsible for:

* aggregating sidebar items
* tracking section visibility
* handling the expanded/collapsed sidebar item state
* calculating the active sidebar item

This all used to live in multiple components (IntersectionObserver.vue, Sidebar.vue) and multiple states (useApiClientState, useTemplateState). Still a long and complex file, which can be split up in multiple files, but it’s all in one place now.

**Before**
`IntersectionObserver` sets the URL to the appearing section ID.

**After**
The `useNavigation` composable is watching the active sidebar item and updates the URL accordingly.

Is it perfect? No. Is it better than what we had before? Yes. :)


amritk (approved):
Tested it out, worked great 🎉

PR #575, opened 2023-11-27
fix: don't trim the slash when it’s the only character in the URL (fix #567)

### LINKED ISSUE #567: landing page bug
### Expected Behavior

When requesting the path `/`, Scalar should return the response returned to (that) path at `http://serveraddress/v1/` 
Screenshot (25)

### Actual Behavior

Scalar actually requests a non correct path at `http://serveraddress/`. This is incorrect as that is an invalid path. However, if you use a new path at `http://serveraddress//`, then Scalar queries the correct path. I dunno how to describe it more than that. 
Screenshot (27)

### Reproduction steps

```bash
1. Any OAS/Swagger file with the paths "/" and "//"
2. Any HTTP server that can serve data to those paths.
3. Generate scalar docs at the http server address.
4. Request data from the paths specified.
```

### Version

CDN

Chrome

hanspagel:
We normalize/format URLs and paths before sending the request and we want to trim the first trailing slash from the path when  concatenating the URL, but we don’t want that, if the path is literally just the slash. This PR fixes it.

`https://example.com/v1/` + `/posts` = `https://example.com/v1/posts`
🆕 `https://example.com/v1/` + `/` = `https://example.com/v1/`

### DIFF
--- a/.changeset/witty-masks-wave.md
+++ b/.changeset/witty-masks-wave.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
+
+fix: don’t trim / from the path, if it’s the only character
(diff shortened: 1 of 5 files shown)


marclave (approved):
A fix and lots of tests! a perfect way to start the week ✨

PR #581, opened 2023-11-28
feat: set `customCss` in the configuration

hanspagel:
Currently, we’re adding custom CSS for all our integrations to have a branded default theme for Fastify, Platformatic, Express, Hono … and need to build the functionality for every integration. @marclave had this amazing idea to add a custom CSS attribute to the configuration object. 🤯 

This PR adds `customCss` to the configuration object, so we can add custom CSS to all the integrations easily from now on. :)

### DIFF
--- a/.changeset/ten-points-fix.md
+++ b/.changeset/ten-points-fix.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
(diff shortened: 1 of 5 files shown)


marclave (approved):
🔥 amazing work @hanspagel

Issue #585, opened 2023-11-28
@scalar/hono-api-reference content and preparsedContent not working


IHIutch:
### Expected Behavior

Add swagger json object to either  `spec.content` or `spec.preparsedContent` and see UI full of my related swagger schema content  when navigating to `/reference` path.

### Actual Behavior

An empty Scalar UI with no content

### Reproduction steps

```bash
1. Set up default project
2. Implement basic @scalar/hono-api-reference example
3. Use spec.content instead of spec.url
```

### Link to site / OpenAPI Spec

[...]

PR #586, opened 2023-11-28
fix: @scalar/hono-api-reference doesn’t use content & preparsedContent

IHIutch:
Fixes https://github.com/scalar/scalar/issues/585

Looks like there may be some issues parsing due to the escaped quotes not being unescaped. I also noticed there were different implementations from the `express` implementation, so this makes `hono` consistent with that by adding the swagger json as an embedded `application/json` script.

I also updated some typing to ensure that it's clear that the parsed json is in fact an object (rather than `any`) as it trickles down to deeper components.

### DIFF
--- a/.changeset/ninety-onions-carry.md
+++ b/.changeset/ninety-onions-carry.md
@@ -1,0 +1,6 @@
+---
+'@scalar/hono-api-reference': patch
+'@scalar/api-reference': patch
+---
+
+fix: hono middleware fails to read configuration.spec.content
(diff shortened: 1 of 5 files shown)


IHIutch:
I don't want to go too far, but the fastify plugin may have a [similar issue](https://github.com/scalar/scalar/blob/3edd2cae422826795ad6e0b9bade5c0773f11ab8/packages/fastify-api-reference/src/templates/render.ts#L56C11-L56C11). I think the same fix from `standalone` could apply there as well (and the typings)

hanspagel:
Great PR! I’m working on the fastify plugin right in #572. Once that’s done, I’d like to go through the integrations and align them. I’ll review and incorporate your changes then. :v:

IHIutch:
Awesome. This is such a cool project! Its been fun playing around with it

hanspagel (approved):
Awesome fix! 🌟  I’ve added just a little bit to it.

PR #602, opened 2023-11-30
chore: remove yjs

hanspagel:
WIP

**Problem**
Currently, we’ve integrated support for collaborative editing. But we’re busy with dependency conflicts all the time.

**Explanation**
This happens because multiple versions of yjs are installed and we can’t find a way to solve this.

**Solution**
With this PR everything related to collaborative editing is removed from the packages.

**Tasks**
- [ ] Somehow expose CodeMirror, so people can add this in their code base
  - [ ] add a demo
  - [x] move the logic from CodeMirror.vue to the hook
  - [ ] slot the editor? expose the CodeMirror instance?


hanspagel:
Done in #698

Issue #603, opened 2023-11-30
Access content in a React app


hanspagel:
Hi all,
I am using the scalar/api-reference in my react app, just wanted to find out how I can access the contents of the swagger editor, if I wanted to export the specification text

_Originally posted by @Uday287 in https://github.com/scalar/scalar/discussions/601_

PR #605, opened 2023-11-30
feat: expose spec content updates

marclave:
**Problem**
Currently, we can't get the spec programatically

**Explanation**
This happens because we dont emit any events to listen to 

**Solution**
With this PR we can configure the references to emit up events. we dont want to always emit events cause that can be expensive to bubble up large payloads

### DIFF
--- a/.changeset/big-bulldogs-double.md
+++ b/.changeset/big-bulldogs-double.md
@@ -1,0 +1,9 @@
+---
+'@scalar/express-api-reference': patch
+'@scalar/fastify-api-reference': patch
+'@scalar/hono-api-reference': patch
+'@scalar/api-reference': patch
+'@scalar/api-client': patch
+---
+
+feat: expose spec content updates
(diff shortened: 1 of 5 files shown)


marclave:
this closes https://github.com/scalar/scalar/issues/603

hanspagel (approved):
It’s good. I have an alternative approach for you:

```diff
const configuration = {
-  emitSpecUpdates: true,
+  onUpdate: (value) => { console.log(value) }
}
```

Could be a simpler solution. Wdyt?

hwkr:
> Good job! I have an alternative approach for you:
> 
> 
> 
> ```diff
> 
> const configuration = {
> 
> -  emitSpecUpdates: true,
> 
> +  onUpdate: (value) => { console.log(value) }
> 
> }
> 
> ```
> 
> 
> 
> Could be a simpler solution. Wdyt?

Oh I like this a lot, this is also more of a standard JS flow that fits in both Vue and React

PR #609, opened 2023-12-01
feat: added two POC's for injection by useCodeMirrorHook

amritk:
This PR is just a POC, well 2 POC's. I did not add the config or the style yet, we should decide what we want to do.

1. First I tried rendering the component from inside the hook but it turns out you can't do that anymore (you could in vue 2 😛). 
2. Then I tried the trusty old slot method, but I found it had too much boilerplate. I left it in the code in case we go this way.
3. Finally landed on the render function method where we just need to pass `setCodeMirrorRef` from the hook to the `SwaggerEditor` component and thats it. Feels the cleanest to me. What do you think?

### Slots
```vue
<script setup lang="ts">
const { codeMirrorRef } = useCodeMirror({
  onUpdate: (v) => {
    console.log('change', v.state.doc.toString())
  },
})
</script>
<template>
  <SwaggerEditor>
    <template #swagger-editor><div ref="codeMirrorRef" /></template>
  </SwaggerEditor>
</template>
```

### Function Ref
```vue
<script setup lang="ts">
const { setCodeMirrorRef } = useCodeMirror({
  onUpdate: (v) => {
    console.log('change', v.state.doc.toString())
  },
})
</script>
<template>
  <SwaggerEditor setCodeMirrorRef="setCodeMirrorRef"/>
</template>
```

They end up being mostly the same, except with the function method we have control of the div and can style it accordingly


hanspagel (changes requested):
Oh man, great work! :) I’d be fine with both.

But shouldn’t we call it `codeMirrorRef` instead of `setCodeMirrorRef`? With `set…` I’d expect it to be a function. Or is that just me?

Personally, I’d prefer 1), because it keeps the responsibility for the styling in the OSS package.

amritk:
@hanspagel `setCodeMirrorRef` is a function, I don't think you can pass down refs like that so you need to use a function. I think 2 keeps the responsibility for the styling in the OSS package since its styled inside `SwaggerEditor`

PR #610, opened 2023-12-01
fix: prevent css leakage from references reset

hwkr:
Adds a `<ResetStyles>` component which loads in a scoped css reset using css modules.

We could also do this as a hook instead of a component 🤷 

Also removes all utility class and usage(s) in the OSS.

We should probably move `<ResetStyles>` to it's own package rather than SwaggerEditor but I'm not very good at package configuration, @hanspagel if you wanted to take a swing at it feel free 🙏


hanspagel (approved):
OMG, this is so awesome. Thanks @hwkr!

The examples used to have Tailwind, including their CSS reset. Without the Tailwind reset I saw some issues.

Here is one example:
Screenshot 2023-12-01 at 12 15 49

We ran into those kind of issues multiple times over the last days. I’ve decided to remove Tailwind (and their reset) from the examples 🥲  and fixed the issues I saw.

Issue #604, opened 2023-11-30
Oauth 2.0 is unfinished, PKCE is missing


marclave:
### Expected Behavior

oauth2.0 to be functional

### Actual Behavior

![image](https://uploads.linear.app/06a32804-030f-45a1-98aa-79d02a076af5/52e38c5a-2f4a-4eb4-81bb-d0399a5dea6b/d26cb30f-29af-4949-9708-ebe2f4777acb?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzA2YTMyODA0LTAzMGYtNDVhMS05OGFhLTc5ZDAyYTA3NmFmNS81MmUzOGM1YS0yZjRhLTRlYjQtODFiYi1kMDM5OWE1ZGVhNmIvZDI2Y2IzMGYtMjlhZi00OTQ5LTk3MDgtZWJlMmY0Nzc3YWNiIiwiaWF0IjoxNzAxMzYyNTM5LCJleHAiOjE3MDE0NDg5Mzl9.Svq0h0NPwnE0sq_fkfe0cHKNx_0zuFJ0J0OZejXHc4E)

### Reproduction steps

```
on latest main with oauth2.0 as an auth type, it's missing and has warning
```

### Link to site / OpenAPI Spec

[docs.scalar.com/swagger-editor](http://docs.scalar.com/swagger-editor)

### Version

latest

### What browsers are you seeing the problem on?

*No response*

<sub>From [SyncLinear.com](https://synclinear.com) | [DOC-1382](https://linear.app/api-docs/issue/DOC-1382/oauth-20-is-unfinished)</sub>

hanspagel:
Started to implement in #613 

**Remaining Tasks**
- [ ] Retrieve the token after the redirect, and keep that in the store
- [ ] Show the stored token
- [ ] Use that token for all requests
- [ ] Optional: Set up an Open Auth 2 backend to test the flow

big-kahuna-burger:
Anyone willing to collaborate on this to have it finished?

TBH, I'm not a good UI developer.
But I can provide supporting OP (OIDC Provider that you can also run locally) 
I depend on `@scalar/fastify-api-reference`. The way I use scalar it is to expose a management API for provider.

https://github.com/big-kahuna-burger/f0

What I'm looking for is a similar behaviour as in SwaggerUI, supporting both oauth2/openIdConnect props in swagger definition and token pre-fill for relevant APIs. 

[...]

PR #613, opened 2023-12-01
chore: add an interface to start an OpenAuth2 flow, see #604

hanspagel:
**Problem**
Currently, there’s no support for OpenAuth 2 authentication.

**Explanation**
This flow is way more complex than the others, but let’s start at least with something.

**Solution**
With this PR an basic interface is added to start an Open Auth2 flow, which covers at least the config from the Petstore example.

**Preview**
Screenshot 2023-12-01 at 15 23 34

**Swagger UI**

Screenshot 2023-12-01 at 15 48 41

### DIFF
--- a/packages/api-reference/src/stores/globalStore.ts
+++ b/packages/api-reference/src/stores/globalStore.ts
@@ -18,7 +18,7 @@ export const createEmptyAuthenticationState = (): AuthenticationState => ({
   },
   oAuth2: {
     clientId: '',
-    scopes: {},
+    scopes: [],
   },
 })
 
--- a/packages/api-reference/src/stores/globalStore.ts
+++ b/packages/api-reference/src/stores/globalStore.ts
@@ -16,6 +16,10 @@ export const createEmptyAuthenticationState = (): AuthenticationState => ({
   apiKey: {
     token: '',
   },
+  oAuth2: {
+    clientId: '',
+    scopes: {},
+  },
 })
 
 const authentication = reactive<AuthenticationState>(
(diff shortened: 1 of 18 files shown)


packages/api-reference/src/stores/globalStore.ts
```diff
@@ -16,6 +16,10 @@ export const createEmptyAuthenticationState = (): AuthenticationState => ({
   apiKey: {
     token: '',
   },
+  oAuth2: {
```
hwkr, line 19: Shouldn't this be owned be the ApiClient store?

hanspagel: That’s a really good question! I thought of the auth stuff in the reference as an extra layer that we should keep in a different store. But I guess it would make it simpler to just have one store.

Currently, if I edit the credentials in the API client, it has no impact on the credentials in the reference. If we keep only one store, editing credentials in the API client would also change what’s in the reference. 

Is that what we want? 🤔 Give me some time to think about that.

hanspagel:
I’d say let’s resolve the TS issues and get this in! We can ff with the other features.

packages/api-reference/src/stores/globalStore.ts
hwkr: > If we keep only one store, editing credentials in the API client would also change what’s in the reference.

@marclave @cameronrohani thoughts?

marclave: I think we want a global store, cause when we feed in the auth from outside of the references or even from intro, it should be fed into the client. however, i think when you are editing in the client its kind of fair to think of it as an isolated sandbox environment with seeded information.

PR #616, opened 2023-12-01
refactor: extract layout type into component

hwkr:
Types the layout props and slots using a base component so you get type errors if you don't use the right props / slots.

Example if `ClassicLayout.vue` included an extraneous prop.

Code-2023-12-01-12-47-14@2x
Code-2023-12-01-12-47-06@2x


marclave (approved):
wow, amazing work @hwkr .

Love the safety too with the onMounted!

hanspagel (approved):
This is so good

PR #620, opened 2023-12-04
chore(fastify-api-reference): move terser and @scalar/api-reference to devDependencies

hanspagel:
**Problem**
Currently, we’re forcing npm/pnpm/yarn to download terser and @scalar/api-reference when installing the fastify plugin, but those needed for development only.

**Solution**
This PR moves them to the devDependencies.


marclave (approved):
ah my bad on that, great work @hanspagel

PR #622, opened 2023-12-04
chore: use fastify logging facilities

hanspagel:
Let’s use the fastify logging facilities, instead of just console.warn.


marclave (approved):
nice 😎

PR #623, opened 2023-12-04
chore: add PostCSS to the examples again

hanspagel:
**Problem**
Currently, the style reset doesn’t work in the examples.

**Explanation**
This happens because I’ve just uninstalled Tailwind and PostCSS in the examples. But we’re aliasing the source files from the packages and need PostCSS to compile them. :)

**Solution**
With this PR PostCSS is added to the examples again.


marclave (approved):
nice nice 😎

PR #624, opened 2023-12-04
feat: show references sidebar slot content in api client sidebar

hwkr:
Puts whatever is in the references sidebar into the api client sidebar as well

Google Chrome-2023-12-04-13-17-30@2x
Google Chrome-2023-12-04-13-17-07@2x

Also fixes a subpixel rendering issue with the swagger editor tabs


cameronrohani (approved):
Nice!

marclave (approved):
awesome work! One would say a pixel perfect fix ✨

PR #625, opened 2023-12-04
feat: switch to @scalar/component library

amritk:
**Problem**
Currently, we were still using the older flow* components

**Explanation**
This happens because we still need to move over to the new scalar* components

**Solution**
With this PR we move over to the scalar* components

Remaining work: move useToasts, useTooltip and useModal (partially complete) into the component library as well. This is a bit trickier as we need to deprecate the packages

Ended up scoping the reset but not the tailwind classes. If that becomes a problem we can re-visit.

Don't forget to build! Then check storybook to test


packages/api-reference/src/components/GettingStarted.vue
```diff
@@ -90,13 +90,17 @@ watch(
...
       <div class="start-cta">
-        <FlowButton
-          label="Test Petstore"
-          @click="example = 'Petstore'" />
-        <FlowButton
-          label="Upload File"
+        <ScalarButton
+          isFullWidth
+          @click="example = 'Petstore'">
+          Test Petstore
```
hwkr, line 96: I love slots so much

hwkr:
Google Chrome-2023-12-04-14-30-20@2x

image

amritk:
> I am very stoked on this, small thing it looks like the import url modal is expanding to the full width
> 
> image
> compared to main image

hey @marclave looks okay to me
image
~Two~ Three questions:
- which references is that on?
- did you build? component library needs to be built
- did the style change so the header is the same color?

marclave:
> > I am very stoked on this, small thing it looks like the import url modal is expanding to the full width
> > image
> > compared to main image
> 
> hey @marclave looks okay to me image ~Two~ Three questions:
> 
> * which references is that on?
> * did you build? component library needs to be built (probably should have included that in the PR notes)
> * did the style change so the header is the same color?

@amritk ah yeah its all good now!


@hwkr @amritk what is left in this body of work ✨

PR #628, opened 2023-12-05
chore: add @scalar/fastify-api-reference CommonJS build again

hanspagel:
**Problem**
Currently, there’s no CommonJS build anymore, I removed it in #621. 

**Explanation**
This happens because `fastify-html` was breaking the CommonJS build … and I thought we can just get away with removing the CommonJS build. Turns out, the fastify community still loves CommonJS and we need it. :)

**Solution**
With this PR the CommonJS build is fixed, with a dynamic `import()` instead of a regular import.

This PR also removes some unused dependencies and does some clean up.

### DIFF
--- a/.changeset/fast-crabs-flow.md
+++ b/.changeset/fast-crabs-flow.md
@@ -1,0 +1,5 @@
+---
+'@scalar/fastify-api-reference': patch
+---
+
+fix: common js build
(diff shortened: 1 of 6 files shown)


marclave (approved):
nice fix @hanspagel ✨

PR #630, opened 2023-12-05
chore: prefix fastify assets, make routePrefix optional

hanspagel:
**Problem**
Currently, we’re using an unprefixed URL for the JavaScript asset.

**Explanation**
This happens because it was easier to not pass the routePrefix, but it feels cleaner to prefix the JS asset URL, too.

**Solution**
With this PR the JS URL is prefixed, and I’ve also made the routePrefix optional. (Not passing a routePrefix will add the reference to `/`).


marclave (approved):
WOW, much cleaner 🧼

PR #644, opened 2023-12-07
fix: handle root path in request generation

marclave:
currently if a server is / or // we don't handle it being the location of the docs, throwing an error https://github.com/litestar-org/litestar/pull/2847#issuecomment-1842048832

this fixes that :)


JacobCoffee (approved):

marclave (commented):
@hanspagel reviewed in real life :)

PR #646, opened 2023-12-07
docs: Fix some backlinks in readme files

mouhannad-sh:
Thanks a lot for building this. 
I'm considering to contribute on another issue but I came across some broken links here so I thought I'd help !

**Problem**
Contributors link and fastify-api-reference links are wrong

**Solution**
Change the urls to fix them up !


hanspagel (approved):
Thanks! :)

PR #647, opened 2023-12-07
feat: centralize url state management in oss

amritk:
**Problem**
Currently the state management was intermingled with the sidebar stuff, also navigation didn't work 100% of the time.

**Explanation**
This happens because of scrolling to ID and the intersection observer triggering after the fact

**Solution**
- centralized all navigation logic
- changed the hash to look like `#tag=pet&method=get&path=/pet/findByStatus` can be easily changed
- use window.location.replace to move to section instead of scrollIntoview
- renamed useNavigation to useSidebar since that's all it does now
- moved all navigation logic into useNavigate

PR on pause for some re-configuration


packages/api-reference/src/hooks/useNavigate.ts
```diff
@@ -0,0 +1,88 @@
...
+
+type NavState = {
+  id?: string
+  label?: string
+}
+
+// Enables the interesection observer
+const canIntersect = ref(true)
+
+const navState = ref<NavState>({})
```
geoffgscott, line 15: I think we want this to be reactive not a ref? Depends on the setting logic but refs don't trigger on deep sets.

packages/api-reference/src/hooks/useSidebar.ts
```diff
@@ -66,8 +57,10 @@ const items = computed((): SidebarEntry[] => {
   // Check whether the API client is visible
   const { state } = useApiClientStore()
 
+
   // Introduction
   const headingEntries: SidebarEntry[] = headings.value.map((heading) => {
+  console.log('adss')
```
geoffgscott, line 63: Stray log

packages/api-reference/src/hooks/index.ts
```diff
@@ -1,7 +1,8 @@
 export * from './useDarkModeState'
-export * from './useNavigation'
+export * from './useNavigate'
```
geoffgscott, line 2: @hanspagel @amritk I realise this is an existing pattern but it is one we should be looking to factor out. Usually with hooks the approach would be to have a single function export (like `useNavigate`) and that export would just expose the internal methods of that module. So we would do:
```
const  { getHeadingId, navigate  } = useNavigate()
```
Which is a pretty nice pattern for a few reasons:
- Strong association between a method and it's logical module
- Ability to have only a single import and access methods as needed either namespaced in a single variable or destructured into individual methods
- Ability to instantiate state at a module level or inside a hook instance depending on situation

Alternatively we can use ES module style import and then just import the functions directly from hooks/useNavigate. Exporting all functions from all hooks into a single index.ts file is just a recipe for naming conflicts and confusion of which hook a function belongs to. I would not generally call these hooks as this is really just a semi-isolated module with exports

packages/api-reference/src/hooks/useNavigate.ts
```diff
@@ -0,0 +1,88 @@
+import { slug } from 'github-slugger'
+import { ref } from 'vue'
+
+import { type Heading, sleep } from '../helpers'
+import type { Tag, TransformedOperation } from '../types'
+
+type NavState = {
+  id?: string
```
geoffgscott, line 8: Does this encapsulate the whole state? Can/should we break this down further into the individual elements of tag/path/method/operation?

packages/api-client/src/stores/apiClientStore.ts
```diff
@@ -43,16 +41,10 @@ function setSnippetType(type: string) {
   state.snippetType = type
 }
 
-function setBreadcrumb(item: string) {
-  state.activeBreadcrumb = item
-}
-
 export const useApiClientStore = () => ({
```
geoffgscott, line 44: Was this really just used externally in the ApiReferences package?

packages/api-reference/src/hooks/useNavigate.ts
amritk: In this case we only set the whole object so its the same? Style?

packages/api-reference/src/hooks/useSidebar.ts
amritk: nooooo

hwkr:
This looks sweet, do you think it's possible to use `<a>` tags in the sidebar the was we do in the search?

amritk:
This PR is closed in favour of https://github.com/scalar/scalar/pull/650

PR #651, opened 2023-12-09
fix: remove duplicate slash in previews for slash based servers

marclave:
**Problem**
currently double slashes if server URL is `/`
image

**Solution**
Now this edge case is handled properly ✨
image
image


cameronrohani (approved):
Perfect

PR #652, opened 2023-12-09
fix: don't override Content-Type header when present in the request

mouhannad-sh:
**Problem**
Currently, The `Content-Type` header is always getting overridden 

**Explanation**
This happens because of these lines of code here 
https://github.com/scalar/scalar/blob/00c01b56b37a278083ce1bc3b38714061d154703/packages/api-client/src/helpers/prepareClientRequestConfig.ts#L33-L39

**Solution**
With this PR we only override the `Content-Type` header if it's not present in the request

### DIFF
--- a/packages/api-client/src/helpers/prepareClientRequestConfig.ts
+++ b/packages/api-client/src/helpers/prepareClientRequestConfig.ts
@@ -1,6 +1,14 @@
 import type { AuthState, ClientRequestConfig } from '../types'
 import { isJsonString } from './isJsonString'
 
+/**
+ * Before a request is sent to the server, we’ll do some final preparation.
+ *
+ * - Add authentication headers
+ * - Add Content-Type header if request.body is JSON
+ * - Parse request.body if it’s JSON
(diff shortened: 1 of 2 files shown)


packages/api-client/src/helpers/prepareClientRequestConfig.ts
```diff
@@ -29,14 +29,13 @@ export const prepareClientRequestConfig = (configuration: {
 
   // Check if request.body contains JSON
   if (request.body && isJsonString(request.body)) {
-    // Add Content-Type header
-    request.headers = [
-      ...(request.headers ?? []),
-      {
+    if (!request.headers?.some((header) => header.name === 'Content-Type')) {
```
marclave, line 32: ah small change, we should check lowercase to make sure we checked it all

```
header.name.toLowerCase() === ‘content-type’
```

hanspagel:
Great fix! I’ve added tests and made the header name lowercase before checking. ✌️

hanspagel (approved):

PR #655, opened 2023-12-09
fix: select menu on chrome

cameronrohani:
**Problem**
Some select menu's currently don't have `appearance:none` which create a conflict when your systems preferred colour scheme doesn't match the prefers-color-scheme attribute on .dark-mode / .light-mode. 

Results in this:
image (1)

**Solution**
Add appearance: none; to all select menu's

### DIFF
--- a/packages/api-client/src/components/ApiClient/Request/Request.vue
+++ b/packages/api-client/src/components/ApiClient/Request/Request.vue
@@ -251,10 +251,11 @@ const { activeRequest, readOnly } = useRequestStore()
   background: transparent;
   outline: none;
   border: none;
+  -moz-appearance: none;
   -webkit-appearance: none;
+  appearance: none;
   font-size: var(--theme-micro, var(--default-theme-micro));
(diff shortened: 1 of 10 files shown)


packages/api-client/src/components/ApiClient/Request/Request.vue
```diff
@@ -251,10 +251,11 @@ const { activeRequest, readOnly } = useRequestStore()
   background: transparent;
   outline: none;
   border: none;
+  -moz-appearance: none;
```
hwkr, line 254: I think we might have autoprefixer doing this for us 🤔

marclave: i couldnt see it autoprefixer doing this for us in older browsers.

marclave: maybe it was the with the software i was testing tho, but let's get this in to fix and i'll make a ticket to audit autoprefixer on older browsers for `appearance: none`

marclave (approved):

PR #657, opened 2023-12-11
chore: don’t pass optional parameters to the client

hanspagel:
**Problem**
Currently, the list of parameters becomes really long and combines parameters that are not possible to use together.

**Explanation**
This happens because we pass all known parameters to the API client.

**Solution**
With this PR optional parameters aren’t passed to the API client. The UI only shows required parameters now.


marclave (approved):
awesome work @hanspagel ✨ 

tested and it works great!

PR #662, opened 2023-12-11
fix: dont alias themes source files

hwkr:
I'm not sure if this is the best way to fix this but it fixes an issue with importing theme css files


hanspagel:
Let’s try it!

PR #663, opened 2023-12-12
feat: toggle parameters, fix #653

### LINKED ISSUE #653: unchecking a header from the api client header grid doesn't omit it from the request
### Expected Behavior

When you uncheck a header or any item in the grid editor it should omit that item from the request

### Actual Behavior

The checkbox input doesn't omit the item, I think it doesn't actually do anything (first time looking at Vue code 😅 )

https://github.com/scalar/scalar/blob/00c01b56b37a278083ce1bc3b38714061d154703/packages/api-client/src/components/Grid/Grid.vue#L102-L105

### Reproduction steps
assuming the following schema

```yaml
openapi: 3.1.0
info:
  title: Swagger Petstore - OpenAPI 3.0
  version: 1.0.11
servers:
  - url: https://petstore3.swagger.io/api/v3
tags:
  - name: pet
paths:
  /pet:
    post:
      tags:
        - pet
      operationId: updatePet
      requestBody:
        content:
            schema:
              $ref: '#/components/schemas/Pet'
        required: true 
      responses:
        '200':
          description: Successful operation
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/Pet'
components:
  schemas:
    Pet:
      required:
        - name
        - photoUrls
      type: object
      properties:
        id:
          type: integer
          format: int64
          examples: 
            - 10
        name:
          type: string
          examples: 
            - doggie
        photoUrls:
          type: array
          xml:
            wrapped: true
          items:
            type: string
            xml:
              name: photoUrl
        tags:
          type: array
          xml:
            wrapped: true
        status:
          type: string
          description: pet status in the store
          enum:
            - available
            - pending
            - sold
      xml:
        name: pet
  requestBodies:
    Pet:
      description: Pet object that needs to be added to the store
      content:
        application/json:
          schema:
            $ref: '#/components/schemas/Pet'
        application/xml:
          schema:
            $ref: '#/components/schemas/Pet'
```

- Test the pets request using the UI 
- In the headers section uncheck the header "Content-Type: application/x-www-form-urlencoded" 
- open chrome devtools on Network tab
- Send the request using the UI 
- Inspect the request -> the content type header will still be `"Content-Type: application/x-www-form-urlencoded"` which is incorrect

_No response_

### Version

0.8.7

Chrome

hanspagel:
**Problem**
Currently, there’s a checkbox to enable/disable parameters, but they are sent anyway.

**Explanation**
This happens because we didn’t build the feature, just the UI. :)

**Solution**
With this PR the parameter type is extended to have `enabled: boolean`, by default `enabled` is set to `true` and the client will only sent what’s enabled.

### DIFF
--- a/.changeset/dull-mugs-knock.md
+++ b/.changeset/dull-mugs-knock.md
@@ -1,0 +1,6 @@
+---
+'@scalar/api-reference': patch
+'@scalar/api-client': patch
+---
+
+feat: toggle parameters to include/exclude them in/from the request
--- a/packages/api-client/src/components/ApiClient/Request/RequestCookies.vue
+++ b/packages/api-client/src/components/ApiClient/Request/RequestCookies.vue
@@ -12,7 +12,7 @@ function handleDeleteIndex(index: number) {
 }
 
 function addAnotherHandler() {
-  activeRequest.cookies?.push({ name: '', value: '' })
+  activeRequest.cookies?.push({ name: '', value: '', enabled: true })
 }
 </script>
 <template>
--- a/packages/api-client/src/components/ApiClient/Request/RequestHeaders.vue
+++ b/packages/api-client/src/components/ApiClient/Request/RequestHeaders.vue
@@ -12,7 +12,7 @@ function handleDeleteIndex(index: number) {
 }
 
 function addAnotherHandler() {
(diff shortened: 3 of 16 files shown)


marclave (approved):
awesome work! ✨

PR #664, opened 2023-12-12
feat: centralize URL state management

amritk:
This almost makes the URL the one source of truth, there was one issue where pressing back does not go to the correct section 100% of the time. Instead of wasting any more time searching that down, just booted it into another ticket and used scrollToId for now.

One remaining piece for this PR would be applying these changes to the API client as well but that is out of scope.

- the URL is back to `#tag/pet/put/pet`
- when you click on a sidebar item, it will save it in history so you can hit back
- when you scroll down the page, it will NOT save it to the history stack
- when you scroll down the page, it will open the next section if you scroll to it
- when you scroll to the top, it removes the hash
- switched to using regular anchor links (for the most part)


hwkr (approved):
Looks great!

hanspagel (changes requested):
Awesome, thanks for all your effort! Two things:

If there’s just one tag, we remove the tag level and just list all pages. But they look like they are nested now:

Screenshot 2023-12-13 at 14 14 08

When I click on a sidebar item, sometimes the next one is shown as the active item:

Screenshot 2023-12-13 at 14 14 28

amritk:
hey @hanspagel 

I'm not quite sure what you mean by the first point
>If there’s just one tag, we remove the tag level and just list all pages. But they look like they are nested now:

I cut the petstore example down to one tag and its still nested. Could you give me an example spec?
image

> When I click on a sidebar item, sometimes the next one is shown as the active item:

Yes this is a known issue and it happens on main too, its unrelated to this PR so I created a separate ticket for it. If you think its important I can boost it up in priority let me know!

https://linear.app/api-docs/issue/DOC-1399/unify-oss-interaction-observers-into-one

hanspagel:
Are you sure this is in scalar/scalar main? I fixed this in #573 and can’t reproduce it in main, but in this PR here. 🤔

amritk:
> Are you sure this is in scalar/scalar main? I fixed this in #573 and can’t reproduce it in main, but in this PR here. 🤔

Oh it was fixed on main, okay let me try again

amritk:
hey @hanspagel so I did some digging, and intersection observer works the same way on main and this branch. The difference is on this branch I fixed headings being highlighted so that is why it looks broken on this branch but not on main. 

What is happening is the way we have separate intersection observers for each section, we sometimes get multiple intersecting at once. The proper fix for this would be to have one intersection observer and if we have multiple we take the one which is higher up on the dom. 

Screenshot from main:
image

BUT for a separate issue I need to disable intersection observer on click, so I will just put that hack back in right now and it should solve the problems!

amritk:
Alright @hanspagel give it another go!

hanspagel:
Thanks for fixing this! This is how it works in main right now:

All IntersectionObservers track if a section is visible or not and we keep a list of all sections with the information whether there are visible or not. The active state is then what’s the first of those sections in the list of sidebar items. So basically what you described, but the difference is probably that we just want to keep the URL as a state?

Anyway, the issue is resolved! Great work.

amritk:
@hanspagel you are the best at finding bugs in my code 😅 I will test and fix and get back to you!

amritk:
@hanspagel Alright I fixed one issue, the headings not being highlighted. 

For the rest, now we need to disable intersection observer on back button. I did a bit of testing but its not so straightforward. I will think of something 🤔

amritk:
Alright @hanspagel give it another go!

also @marclave @geoffgscott should we add playwright to the OSS side? it could be beneficial here

marclave:
> Alright @hanspagel give it another go!
> 
> also @marclave @geoffgscott should we add playwright to the OSS side? it could be beneficial here

playwright on the OSS side sounds awesome! Let's cut a ticket and plan it out more after we get more fixes in 🕺

hanspagel:
Man, this is awesome! It’s so much better. 😅 I’ll try to get CI green and merge it.

hanspagel (approved):

PR #665, opened 2023-12-13
fix: add ref back

amritk:
There is something  very strange going on with content. I haven't found the source yet but it looks like its getting overwritten from a ref -> to a string.


packages/api-reference/src/components/ApiReferenceBase.vue
```diff
@@ -109,6 +109,8 @@ const isMobile = useMediaQuery('(max-width: 1000px)')
 
 const swaggerEditorRef = ref<typeof SwaggerEditor | undefined>()
 
+watch(rawSpecRef, (nc) => console.log(nc))
```
hwkr, line 112: Console log?

amritk: good call

PR #669, opened 2023-12-13
fix: mobile sidebar doesnt close on change item

marclave:
**Problem**
When you click the sidebar on mobile, it doesnt close the sidebar

**Solution**
sidebar closes once activeitem changes

### DIFF
--- a/packages/api-reference/src/components/Layouts/ModernLayout.vue
+++ b/packages/api-reference/src/components/Layouts/ModernLayout.vue
@@ -26,8 +26,8 @@ const config = computed(() => {
 
 const { activeItemId } = useNavigation()
 
-watch(activeItemId, () => {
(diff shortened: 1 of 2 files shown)


packages/api-reference/src/components/Layouts/ModernLayout.vue
```diff
@@ -22,6 +23,14 @@ const config = computed(() => {
     : props.configuration?.showSidebar
   return { ...props.configuration, showSidebar }
 })
+
+const { activeItemId } = useNavigation()
+
+watch(activeItemId, () => {
```
hwkr, line 29: Might be could to do a quick comparison too

```suggestion
watch(activeItemId, (n,o) => {
    if (n && n !== o) {
```

PR #672, opened 2023-12-13
fix: temp cap on schema level due to stack overflow

amritk:
**Problem**
Currently we stack overflow on a recursive schema property

**Explanation**
This happens because our parser just loops on it (I think)

**Solution**
With this PR we just put a temp cap on recursion level while we sort out a proper fix

To test, throw startons spec into the editor and open the models section
https://docs.starton.com/api-reference

The proper fix would be to test for recursion in the parser and just set the schema name as the type when referencing itself. I will throw up a ticket for this


marclave (approved):
nice ✨

PR #673, opened 2023-12-13
fix: make layout self contained

cameronrohani:
**Problem**
app not self contained 

**Explanation**
if someone adds padding to the app container it is not containing the apps content 

**Solution**
add position relative to reference content + fix toast layout to left


marclave (approved):
nice fix! ✨

PR #676, opened 2023-12-14
fix: hotfix for position relative mobile menu

amritk:
Quick hotfix for mobile menu


marclave (approved):
Thanks for the quick fix ✨

PR #681, opened 2023-12-14
refactor: use `Schema` component for parameters, fix #598

### LINKED ISSUE #598: Scalar doesn't expand a query parameter as object
### Expected Behavior

When I have a parameter defined as a object, it should be expanded to all its inner properties

### Actual Behavior

A sample of what is being rendered:

image

### Reproduction steps

```bash
Add this sample to a GET endpoint:


"parameters": [
          {
            "schema": {
              "type": "object",
              "properties": {
                "search": {
                  "type": "string"
                },
                "status": {
                  "type": "string",
                  "enum": [
                    "invited",
                    "onboarding",
                    "onboarded"
                  ]
                },
                "role": {
                  "type": "array",
                  "items": {
                    "type": "string",
                    "enum": [
                      "org_admin",
                      "team_admin",
                      "org_finance_admin",
                      "team_finance_admin"
                    ]
                  }
                },
                "team": {
                  "type": "array",
                  "items": {
                    "type": "string",
                    "format": "uuid"
                  }
                }
              },
              "additionalProperties": false
            },
            "in": "query",
            "name": "filter",
            "required": false
          }
        ],
```
```


_No response_

### Version

latest

Chrome

hanspagel:
We’ve built a pretty powerful `Schema` component to render the models a while ago already. The parameters do basically the same, but don’t cover as many use cases as the Schema component.

This PR adds the `Schema` component to render the parameters and the response body. This adds a ton of features to the parameter list and the response body and fixes some bugs.

* The `Schema` component has a `compact` layout now (less borders, less spacing, no heading).
* The `Schema` component can have a toggle to show/hide children.
* `required` is still rendered, but `optional` isn’t rendered anymore
* I’ve changed the oneOf/anyOf/allOf layout (no more orange badge, instead using the `Badge` component).
* I’ve removed the global `<style>` in `RequestBody` to style the parameter list 🤔  and replaced it with scoped styles in the components.
* The `MarkdownRenderer` had CSS just for the parameter description, that’s in the actual component now.

Screenshot 2023-12-14 at 14 50 09
Screenshot 2023-12-14 at 14 50 32
Screenshot 2023-12-14 at 14 51 01
Screenshot 2023-12-14 at 15 17 49


hwkr (changes requested):
This looks awesome! Seems like it broke a bunch of the classic layout though,

Google Chrome-2023-12-14-13-19-27@2x
Google Chrome-2023-12-14-13-19-41@2x

hanspagel:
Oops, I didn’t check the classic layout.

* I’ve added some padding to the card, not sure if this is a proper solution? Feel free to change it. 
* And I’ve set a single font size for the `Schema` component.

What do you think? I struggled a little bit with defining the spacing, because there’s so many places it’s used (models, parameters, request body, classic layout) and so many uses cases (flat, nested, objects, arrays, enums, oneOf, allOf …).

I should have set up a few Storybook stories with all the uses cases, but HMR is broken (again) in the @scalar/api-reference storybook setup. :(

hwkr (approved):
Looks good! We can tidy up the card look in [Cam's other PR](https://github.com/scalar/scalar/pull/685)

PR #683, opened 2023-12-14
fix: does not add cookies

hanspagel:
**Problem**
Currently, you can’t add cookies if the list is empty.

**Explanation**
This happens because in some cases `cookies` is undefined and a `push()` to undefined fails. 

**Solution**
With this PR variables, queryStrings, headers and cookies are initialized before we try to use `push()` on it.


marclave (approved):
ah great catch @hanspagel !

PR #689, opened 2023-12-15
feat: remove yjs + hocus pocus from OSS

amritk:
YJS always gives us issues and is not even used/required on the OSS side. This PR removes it from this repo and there's a matching PR on the closed source side to add it there.

To test, poke around the swagger editor and ensure it still works the same


marclave (approved):
Fantastic work @amritk ✨ 

we might want to make this a minor release is the only thing, but the rest is tested and works amazing ✨

PR #690, opened 2023-12-16
docs: Update Scalar Hono API Reference Plugin README.md on 

sinh117801:
**Problem**
Currently, on Scalar Hono API Reference Plugin README, when I click on link `Zod OpenAPI Hono`, it redirects to 404 page.

**Explanation**
This happens because `https://hono.dev/guides/openapi` not exist anymore

**Solution**
With this PR, I change the URL from `https://hono.dev/guides/openapi` to `https://github.com/honojs/middleware/tree/main/packages/zod-openapi`

### DIFF
--- a/packages/hono-api-reference/README.md
+++ b/packages/hono-api-reference/README.md
@@ -17,7 +17,7 @@ npm install @scalar/hono-api-reference
 
 ## Usage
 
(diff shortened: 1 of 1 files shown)


marclave (approved):
ah great fix! thanks so much ✨

PR #691, opened 2023-12-16
Marc/bug remove caching of configuration

marclave:
**Problem**
Currently when changing the spec passed as a prop, it would merge the spec with the old spec

**Explanation**
This happens because we merge the config objects and some funny business was happening

**Solution**
With this PR i removed the cache of the spec config : )


hwkr (approved):
Tested it out and seems to be working gucci :)

PR #699, opened 2023-12-18
feat: scope all scalar components using a class prefix

### LINKED ISSUE #692: Tailwind CSS conflict
### Expected Behavior
Embedding ApiReference component should not affect the website.

### Actual Behavior

Looks like Scalar's hidden class overrides Tailwind's selector
```
@media (min-width: 1024px)
.lg\:flex {
    display: flex;
}
```

### Reproduction steps

Import ApiReference into tailwind powered Vue.js project with mobile-first approach with `.hidden `class

```html
<script setup lang="ts">
  import { ApiReference } from '@scalar/api-reference'
</script>
```

_No response_

### Version

^1.11.0

_No response_

hwkr:
Prefixes all our classes with [`postcss-prefix-selector`](https://www.npmjs.com/package/postcss-prefix-selector)

Fixes #692 

## Tailwind classes:

Google Chrome-2023-12-18-14-21-50@2x

## Global classes:

Google Chrome-2023-12-18-14-21-59@2x

### DIFF
--- a/packages/api-reference/src/components/GettingStarted.vue
+++ b/packages/api-reference/src/components/GettingStarted.vue
@@ -91,12 +91,12 @@ watch(
       </p>
       <div class="start-cta">
         <ScalarButton
-          isFullWidth
+          fullWidth
           @click="example = 'Petstore'">
           Test Petstore
         </ScalarButton>
         <ScalarButton
-          isFullWidth
+          fullWidth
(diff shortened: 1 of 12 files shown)


marclave (approved):
nice work! seems to be all working well from my testing.

I think we just need to fix the build and we are good to go :)

PR #703, opened 2023-12-19
chore: remove customClass attribute from parameters

hanspagel:
We added support for `required` to parameters, there’s no need to pass a `customClass` anymore. :)


marclave (approved):
nice :)

PR #704, opened 2023-12-19
chore: switch to @readme/openapi-parser

hanspagel:
@apidevtools/swagger-parser seems to be unmaintained and @readme/openapi-parser is the most popular fork.

| Metric           | @apidevtools/swagger-parser | @readme/openapi-parser |
| ---------------- | --------------------------- | ---------------------- |
| Weekly downloads | 1.2M                        | 150K                   |
| Last commit      | 5 months ago                | 2 weeks ago            |
| Last update      | 2 years ago                 | 9 months ago           |

So … both teams aren’t very active, but at least we can remove the override to prevent hanging specs (verified with the regression test).

### DIFF
--- a/.changeset/violet-suits-chew.md
+++ b/.changeset/violet-suits-chew.md
@@ -1,0 +1,5 @@
+---
+'@scalar/swagger-parser': patch
+---
+
+chore: switch to @readme/openapi-parser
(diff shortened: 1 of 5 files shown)


marclave:
@hanspagel so what im seeing is we need a  `@scalar/swagger-parser` or `@scalar/openapi-parser`

hanspagel:
Needs SSG testing

PR #714, opened 2023-12-20
refactor: scope all sidebar styles

hwkr:
I don't think this breaks anything but worth giving a test


marclave (approved):
tested and approved! great work :)

PR #725, opened 2023-12-23
fix: declare theme font for buttons and fix regressed font weight

cameronrohani:
Before:
image

After:
image

Before:
image

After:
image


hanspagel:
Oops, thanks for the fix!

hanspagel (approved):

marclave (approved):
Nice work! let's fix the build and add a change set then get this in :)

Issue #705, opened 2023-12-19
Stripe: all form fields empty (Content-Type: application/x-www-form-urlencoded' )


hanspagel:
I’ve checked the Stripe spec and it seems like all form fields are empty.

Probably the same as #597

[Stripe API.yaml.zip](https://github.com/scalar/scalar/files/13716792/Stripe.API.yaml.zip)

Issue #597, opened 2023-11-29
Body is missing when it has consumes: ['multipart/form-data']


thelinuxlich:
### Expected Behavior

Scalar should show the body fields always

### Actual Behavior

When you add consumes: ['multipart/form-data'] to the endpoint, it doesn't show the body fields

### Reproduction steps

```bash
Add consumes: ['multipart/form-data'] to a POST endpoint with a body definition
```

### Version

latest

### What browsers are you seeing the problem on?

Chrome

hanspagel:
You’re so right, I’ve skipped this. Do you have a good real-world snippet for me?

thelinuxlich:
```json
"/company/": {
      "post": {
        "requestBody": {
          "content": {
            "multipart/form-data": {
              "schema": {
                "type": "object",
                "properties": {
                  "image": {},
                  "name": {
                    "type": "string"
                  },
                  "description": {
                    "type": "string"
                  }
                },
                "required": [
                  "name",
                  "description"
                ],
                "additionalProperties": false
              }
            }
          },
          "required": true
        },
        "security": [
          {
            "bearerAuth": []
          }
        ],
        "responses": {
          "200": {
            "description": "Default Response",
            "content": {
              "application/json": {
                "schema": {
                  "type": "object",
                  "properties": {
                    "success": {
                      "type": "boolean"
                    },
                    "data": {
                      "type": "object",
                      "properties": {
                        "name": {
                          "type": "string",
                          "minLength": 3,
                          "maxLength": 100
                        },
                        "size": {
                          "type": "string",
                          "enum": [
                            "micro_sized",
                            "small_sized",
                            "medium",
                            "large_sized"
                          ],
                          "default": "medium"
                        },
                        "incorporation_state": {
                          "type": "string",
                          "minLength": 1,
                          "maxLength": 50
                        },
                        "incorporation_type": {
                          "type": "string",
                          "enum": [
                            "sole_proprietorship",
                            "private_limited_company",
                            "limited_liability_company",
                            "s-corporation",
                            "c-corporation",
                            "public_limited_company",
                            "limited_partnership",
                            "holding_company",
                            "non_government_organisation",
                            "statutory_company",
                            "subsidiary_company",
                            "unlimited_partnership",
                            "charitable_incorporated_organisation",
                            "chartered_company",
                            "association",
                            "non_profit_organisation",
                            "im_not_sure"
                          ],
                          "default": "sole_proprietorship"
                        },
                        "website": {
[...]

PR #734, opened 2024-01-04
feat: Add dynamic content type selection for request body

Chinlinlee:
**Problem**
Currently, the request body of all other content-types are empty except `application/json`

before:
Screenshot_1
Screenshot_2

after:
image

In addition, I have also added `multipart/form-data` example `uploads an image using form data` in the pet

**Explanation**
Because v-if only specifies that the body section of application/json can be displayed.
In the requestBody.content, there are content-type keys corresponding to the openapi schema, so we can directly use it to generate a select.

**Solution**
- Add dynamic content type selection for request body display
  - Generate select using the key of `requestBody?.content`
- User can view the request body with a specified content-type
- Add an example of `multipart/form-data` in petstore
- Probably related to #705 and #597

### DIFF
--- a/packages/api-reference/src/components/Content/ReferenceEndpoint/RequestBody.vue
+++ b/packages/api-reference/src/components/Content/ReferenceEndpoint/RequestBody.vue
@@ -5,6 +5,7 @@ import type { ContentType, RequestBody } from '../../../types'
 import Schema from '../Schema.vue'
 
 const prop = defineProps<{ requestBody?: RequestBody }>()
+
 const selectedContentType = ref<ContentType>('application/json')
 if (prop.requestBody?.content) {
   const keys = Object.keys(prop?.requestBody?.content)
@@ -17,17 +18,19 @@ if (prop.requestBody?.content) {
   <div v-if="prop?.requestBody">
     <div class="request-body-title">
       <slot name="title" />
-      <select
-        v-if="prop?.requestBody"
-        v-model="selectedContentType"
-        style="display: block">
-        <option
-          v-for="(value, key) in prop.requestBody?.content"
-          :key="key"
-          :value="key">
-          {{ key }}
-        </option>
-      </select>
+      <div class="request-body-title-select">
+        <span>{{ selectedContentType }}</span>
+        <select
+          v-if="prop?.requestBody"
+          v-model="selectedContentType">
+          <option
+            v-for="(value, key) in prop.requestBody?.content"
+            :key="key"
+            :value="key">
+            {{ key }}
+          </option>
+        </select>
+      </div>
     </div>
     <div
       v-if="prop?.requestBody.content?.[selectedContentType]"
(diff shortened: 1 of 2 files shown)


packages/api-reference/src/components/Content/ReferenceEndpoint/RequestBody.vue
```diff
@@ -1,21 +1,39 @@
...
+          v-for="(value, key) in prop.requestBody?.content"
+          :key="key"
+          :value="key">
+          {{ key }}
+        </option>
+      </select>
     </div>
     <div
-      v-if="requestBody.content?.['application/json']"
+      v-if="prop?.requestBody.content?.[selectedContentType]"
```
marclave, line 31: this is causing a build error for undefined being used as an index, we should change the line to this for an undefined check on `selectedContentType`

```
v-if="selectedContentType && prop?.requestBody.content?.[selectedContentType]"
```

```
12:31:40 AM: @scalar/api-reference:build: src/components/Content/ReferenceEndpoint/RequestBody.vue(31,41): error TS2538: Type "undefined" cannot be used as an index type.
12:31:40 AM: @scalar/api-reference:build: src/components/Content/ReferenceEndpoint/RequestBody.vue(36,43): error TS2538: Type "undefined" cannot be used as an index type.
```

packages/api-reference/src/components/Content/ReferenceEndpoint/RequestBody.vue
```diff
@@ -1,21 +1,39 @@
...
+    ? Object.keys(prop?.requestBody?.content)[0]
+    : undefined,
+)
 </script>
 <template>
-  <div v-if="requestBody && requestBody.content?.['application/json']">
+  <div v-if="prop?.requestBody">
     <div class="request-body-title">
       <slot name="title" />
+      <select
```
marclave, line 18: @cameronrohani would you be able to come in and bring in the styles we use for `selects` or guide on where we should be pulling from :)

packages/api-reference/src/petstorev3.json
```diff
@@ -452,6 +452,54 @@
         ]
       }
     },
+    "/pet/{petId}/uploadImageForm": {
```
marclave, line 455: awesome! :)

packages/api-reference/src/components/Content/ReferenceEndpoint/RequestBody.vue
```diff
@@ -1,21 +1,39 @@
...
+import { ref } from 'vue'
+
 import type { RequestBody } from '../../../types'
 import Schema from '../Schema.vue'
 
-defineProps<{ requestBody?: RequestBody }>()
+const prop = defineProps<{ requestBody?: RequestBody }>()
+const selectedContentType = ref(
+  prop?.requestBody?.content
+    ? Object.keys(prop?.requestBody?.content)[0]
```
marclave, line 10: do we need to do an empty check here incase `prop.requestBody.content = {}`? since we check `[0]` would that fail if it exists with nothing?

marclave (changes requested):
hey @Chinlinlee , thank you so much for your contribution!

this has been something on my mind for awhile, so it's great to see this! :) 


I added a review below ✨ we just need to fix the build errors around undefined indexing and @cameronrohani can add some styling.

Then we can get this merged ✨

marclave (approved):
awesome work @Chinlinlee I pushed up some style alignment fixes, excited to get this deployed :)

marclave:
will do a changeset and a release! ✨

PR #749, opened 2024-01-08
fix: request body does not reset, fix #730

### LINKED ISSUE #730: [DOC-1431] request body doesnt reset on active item change
### Expected Behavior

the body to change based on the active request

### Actual Behavior

it stays to the last remaining request with a body

image

image

### Reproduction steps

```
1. load petstore
2. test and endpoint
3. go to a request with a body
4. go to a different request
5. see the body doesnt change
```

*No response*

### Version

latest

*No response*

<sub>From [SyncLinear.com](https://synclinear.com) | [DOC-1431](https://linear.app/api-docs/issue/DOC-1431/body-doesnt-reset-on-active-item-change)</sub>

hanspagel:
**Problem**
Currently, the request body doesn’t reset when you navigate to a request without a request body (#730).

**Explanation**
This happens because I watched for content changes, but ignored `null`/`undefined`/`''`. I can’t remember why I did that, but I think it was just that the `setCodeMirrorContent` helper didn’t support `undefined` or `null`.

**Solution**
With this PR the `setCodeMirrorContent` helper deals with those values and we don’t ignore them anymore.

### DIFF
--- a/.changeset/ninety-pugs-tickle.md
+++ b/.changeset/ninety-pugs-tickle.md
@@ -1,0 +1,6 @@
+---
+'@scalar/use-codemirror': patch
+'@scalar/api-client': patch
+---
+
+fix: request body doesn’t reset on navigating to a request without a body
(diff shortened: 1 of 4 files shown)


marclave (approved):
awesome work @hanspagel , I just tested it out and its all fixed ✨

PR #754, opened 2024-01-10
feat: client ignore list

hwkr:
Allows you to ignore clients by key.

Code-2024-01-10-18-54-01@2x
→
Google Chrome-2024-01-10-18-53-54@2x

## Some questions

- Why do we remove `fetch` and then re-add it? @hanspagel?
- Do we want to target all clients with a key or should it be a target client pair?
  - Should it be a tuple or a string? `['javascript', 'fetch']` vs `'javascript/fetch`?
  - The string is easier to read (imo) but the first one can be typesafe(er)
- Do we want to be able to remove a target entirely? e.g. `javascript/*`


packages/api-reference/src/hooks/useSnippetTargets.ts
```diff
@@ -0,0 +1,67 @@
...
+
+const DEFAULT_EXCLUDED_CLIENTS = ['unirest'] as const
+
+const excludedClients = ref<string[]>([...DEFAULT_EXCLUDED_CLIENTS])
+
+export function useSnippetTargets() {
+  const availableTargets = computed(() =>
+    allTargets()
+      .map((target) => {
+        // Remove fetch because we want a different one
```
hwkr, line 12: @hanspagel I don't understand why we're doing this 😂

marclave (approved):
amazing! im stoked on this :) Definitely the long term solution to letting people opt-out of client libraries to render ✨

Issue #755, opened 2024-01-10
Ctrl K -> select doesn't scroll to appropriate place


pq43:
### Expected Behavior

When I Ctrl K / search and select a route, I expect it to scroll / jump to that route

### Actual Behavior

It doesn't, so I have to manually scrolldown

### Reproduction steps

```bash
1. Ctrl K or click search
2. Click on any endpoint
3. Notice scroll doesn't happen
```

### Version

"@scalar/api-client": "0.8.3"

### What browsers are you seeing the problem on?

Chrome

marclave:
hey @pq43, i noticed this happens when an endpoint is in a tag that is collapsed, is this what you are seeing as well?

pq43:
H, yes I think that's it! Would be really nice to be able to ctrl k any endpoint tho :)

pq43:
Thank you so much! working with you all through open source has been a joy :)

PR #758, opened 2024-01-10
fix(#755): open sections on search click

amritk:
**Problem**
If you click on a search result from a closed section, it will not get scrolled to.

**Explanation**
This happens because the section is not being opened.

**Solution**
With this PR we open the section on click of the search result.


marclave (approved):
awesome work ✨

PR #760, opened 2024-01-10
fix: cleanup safari bugs

hwkr:
I think this should fix everything?

### DIFF
--- a/packages/api-client/src/helpers/findVariables.ts
+++ b/packages/api-client/src/helpers/findVariables.ts
@@ -6,7 +6,7 @@ export const findVariables = (value: string) => {
   // Wrapped in single or double curly braces.
   // Ignores whitespace
   // Works with lowercase, uppercase, numbers, dashes, underscores
-  const regex = /(?<=\{)\s*\w+\s*(?=\})/g
+  const regex = /(?=\{)\s*\w+\s*(?=\})/g
 
   return value.match(regex)?.map((variable) => variable.trim()) || []
 }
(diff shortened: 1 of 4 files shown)


packages/api-client/src/helpers/findVariables.ts
```diff
@@ -6,7 +6,7 @@ export const findVariables = (value: string) => {
   // Wrapped in single or double curly braces.
   // Ignores whitespace
   // Works with lowercase, uppercase, numbers, dashes, underscores
-  const regex = /(?<=\{)\s*\w+\s*(?=\})/g
+  const regex = /(?=\{)\s*\w+\s*(?=\})/g
```
hwkr, line 9: So safari doesn't support look behind until version 16.4 and I was running 16.3, do we need this?

https://caniuse.com/js-regexp-lookbehind

packages/api-reference/src/components/Content/ReferenceEndpoint/ExampleRequest.vue
```diff
@@ -301,7 +301,7 @@ computed(() => {
   padding: 6px;
 }
 .request-editor-section {
-  display: grid;
+  display: flex;
```
hwkr, line 304: I _think_ this fixes DOC-1444?

cameronrohani: it did for me!

marclave (approved):
awesome work @hwkr ✨

PR #763, opened 2024-01-12
fix: classic layout fixes

marclave:
fixed some small classic layout issues around padding and custom themes


cameronrohani (approved):
nice 1

PR #772, opened 2024-01-14
feat: support fastify platform on nestjs

dunklesToast:
**Problem**
Currently, fastify based nestjs apps are throwing a 500 Error since `res.send` does not exist on fastify middlewars. This PR attempts to fix this.

**Explanation**
[NestJS supports fastify via an additional adapter](https://docs.nestjs.com/techniques/performance). fastify handles middlewares a little bit different than express does so the scalar NestJS extension doesn't work out of the box. As fastify's reply is a raw [ServerResponse](https://nodejs.org/api/http.html#class-httpserverresponse), we need to `write` the content ourself. 

**Solution**
I've added a check into the `nestJSApiReference` file to check if `req.send` is a function. If so, we assume that we're running on express and just use the old logic. If `req.send` does not exist, it'll be a fastify-powered NestJS instance which means we need to `write` manually. I've also added a new example for fastify (which is why this PR is 18 files). The example is basically a 1:1 copy with only the steps from NestJS fastify docs added.
I was undecided whether to do this `isExpressResponse` function or let the user toggle a value via the configuration to specify if it is fastify (smth. like a `withFastify` boolean) but imo this is the best approach as it works and doesn't require additional user input. 

Hope this helps, let me know what you think!

To speed review up a bit here's a list of the interesting files:

[`main...dunklesToast:scalar:main`#diff-4fc0122499](https://github.com/scalar/scalar/compare/main...dunklesToast:scalar:main#diff-4fc0122499d9331e9eb4556fce36865875625d279b5e977f24331473f5aad839)
[`main...dunklesToast:scalar:main`#diff-eb3e46efd9](https://github.com/scalar/scalar/compare/main...dunklesToast:scalar:main#diff-eb3e46efd9e16ee1b8b1cbc20aba4d2b09f3b86eca1182e79f23c596a8888412)


marclave:
@dunklesToast thanks so much for the contribution! I'm leaning to the `withFastify` option for the nestJS Scalar integration :) 

i'm excited to test this out and dive into the code, im pretty new to NestJS so this is an incredibly insightful writeup, thank you so much and what an amazing first contribution to Scalar ✨ 

let me get started on this review.

what are your thoughts on the `withFastify` config option for this integration @dunklesToast ? :)

dunklesToast:
hey @marclave,
thanks for the quick reply!

i don't have any objection with the `withFastify` option. I'll push a change with that shortly!

marclave (approved):
@dunklesToast just tested it all and works AMAZING! let me add a changeset ✨ and let's get this released.

Amazing contribution. We do a changelog writeup every week and this will be in it ✨

marclave:
released in `@scalar/nestjs-api-reference@0.1.13` ✨

marclave:
@dunklesToast released in `@scalar/nestjs-api-reference@0.1.13`

PR #776, opened 2024-01-15
fix: reset body margin

cameronrohani:


packages/swagger-editor/src/components/ResetStyles.vue
```diff
@@ -79,4 +80,8 @@ onMounted(() => {
     font-family: var(--theme-font, var(--default-theme-font));
   }
 }
+body.light-mode,
+body.dark-mode {
+  margin: 0;
+}
```
hanspagel, line 86: One of those classes is applied anyway, can’t we just use `body { margin: 0; }` then?

hanspagel:
I was hesitant to add this to the reset and added it to all our integrations, but now that I think about it, it’s way better to add it to the reset! Thanks!

hanspagel (approved):

marclave (approved):

Issue #770, opened 2024-01-13
[DOC-1460] Allow Download of zip response from api client request response 


marclave:
### Expected Behavior

Hey gents, I need some help! 
I have an API the returns a zip file using express response.download. We just changed our API docs from swagger to scalar, but with scalar we can not download the file. Is there a way to do this?
It shows the output correct, but doesn't offer a "blob" to download the file

### Actual Behavior

but doesn't offer a "blob" to download the file

### Reproduction steps

```bash
test an api that returns a blob
response is shown properly
but no download file button
```

### Version

latest

<sub>[DOC-1460](https://linear.app/api-docs/issue/DOC-1460/allow-download-of-zip-response-from-api-client-request-response)</sub>

PR #784, opened 2024-01-15
feat: return zip files from echo server, see #770

hanspagel:
With this PR the echo server responds with an empty zip file for all request URLs ending with *.zip.

Preparation for #770.


marclave (approved):
nice ✨

PR #785, opened 2024-01-15
fix: body and formData parameters don’t show up

hanspagel:
Currently, `body` and `formData` parameters don’t show up.

Back then, in Swagger 2.0, a valid location for parameters was `body` and `formData`, we ignored this until now. With this PR those parameters are rendered, too.

- OpenAPI 3.x: Possible values are “query”, “header”, “path” or “cookie”.
- Swagger 2.0: Possible values are "query", "header", "path", "formData" or "body".

**Tasks**
- [ ] Use body to prefill the API client
- [ ] Use formData to prefill the API client?


marclave (approved):
😍 this is GREAT!!!

PR #793, opened 2024-01-16
feat: show security scheme name and description, fix #792

### LINKED ISSUE #792: securityScheme description field and key aren't displayed
### Expected Behavior

Maybe this should be a feature request as it's something that the package doesn't do already, but it's something that is in the spec which isn't being displayed in the docs, so that feels like a bug to me.

The Open API spec allows [a `description` attribute](https://swagger.io/specification/#security-scheme-object). I would expect this field to be displayed. It should also allow rendering markdown.

Each scheme will also have a unique key which can be useful for differentiating otherwise similar schemes. I would expect this to be displayed somewhere in the UI, perhaps in the dropdown when picking an authentication type.

### Actual Behavior

The description and key aren't visible.

### Reproduction steps

```bash
Create a spec including like so:


{
  [...]
  "components": {
    "securitySchemes": {
      "MyAuthScheme": {
        "type": "apiKey",
        "description": "Here is some *helpful text*.",
        "name": "Authorization",
        "in": "header"
      }
    }
  }
}
```

Observer that `MyAuthScheme` and `Here is some *helpful text*` aren't displayed in the UI.
```


_No response_

### Version

1.12.7

Chrome

hanspagel:
Currently, we don't show the security scheme key and description as pointed out in https://github.com/scalar/scalar/issues/792.

This PR adds the security schema key in quotes and shows the description (if there’s one).

It’s part of the spec, so we should show it somehow, but I’m not happy with the design. Wdyt @cameronrohani?

Screenshot 2024-01-16 at 13 34 16

Screenshot 2024-01-16 at 13 34 35

Screenshot 2024-01-16 at 13 28 05

### DIFF
--- a/.changeset/hot-knives-relax.md
+++ b/.changeset/hot-knives-relax.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
+feat: show security schema key and description
--- a/packages/api-reference/src/components/Content/Authentication/SecurityScheme.vue
+++ b/packages/api-reference/src/components/Content/Authentication/SecurityScheme.vue
@@ -2,9 +2,9 @@
 import { type OpenAPIV3 } from 'openapi-types'
 import { computed } from 'vue'
 
-import { Security } from '../../../../../api-client/dist'
 import { useGlobalStore } from '../../../stores'
 import type { SecurityScheme } from '../../../types'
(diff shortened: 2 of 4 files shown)


marclave:
@hanspagel this is great! I defs lean towards **Alternative: Type in brackets**

marclave (approved):
awesome ✨

PR #806, opened 2024-01-17
fix: package versioning

amritk:
This PR fixes some versioning while also adding a new strategy for testing packages locally.
```bash
sh ./pack-packages.sh
```
will create tarballs of the local package versions. This strategy is closer to an npm prod release than linking packages.

### DIFF
--- a/packages/api-client/package.json
+++ b/packages/api-client/package.json
@@ -61,7 +61,6 @@
   "devDependencies": {
     "@scalar/api-client-proxy": "workspace:*",
     "@scalar/echo-server": "workspace:*",
-    "@scalar/use-codemirror": "^0.7.17",
     "@types/content-type": "^1.1.6",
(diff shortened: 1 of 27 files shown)


packages/api-client/package.json
```diff
@@ -61,7 +61,6 @@
   "devDependencies": {
     "@scalar/api-client-proxy": "workspace:*",
     "@scalar/echo-server": "workspace:*",
-    "@scalar/use-codemirror": "^0.7.17",
```
hanspagel, line 64: Why is that removed? 🤔 We’re importing from it in @scalar/api-client.

amritk: it was duplicated, still in there

hanspagel (approved):
AWESOME, let’s try this

marclave (approved):

PR #808, opened 2024-01-17
fix: actually update fastify-html

mcollina:
There was a double entry of `fastify-html`, and those were not in agreement on which version should be installed.


hanspagel:
Oh yes, I was too fast with that review. Works now:

```bash
→ why -r fastify-html

devDependencies:
fastify-html 0.3.0
```

hanspagel (approved):

mcollina:
Could you ship a new release? Thx!

PR #822, opened 2024-01-18
fix: classic scrollbar

marclave:


cameronrohani (approved):
good temp fix until @hwkr  has time to take a look :)

PR #833, opened 2024-01-19
fix: search keyboard navigation doesn’t work when the mouse hovers the results, fix #827

### LINKED ISSUE #827: search:  pressing and holding the down arrow key, the selection doesn't go down
> when pressing and holding the down arrow key, the selection doesn't go down

https://github.com/scalar/scalar/assets/1577992/c352addf-31d5-452e-9ee3-ca0487a0a5e8

hanspagel:
When hovering the search results and then pressing the arrow keys, the list of results sticks to where it’s at.

This is because there was an @mouseover listener overwriting the selected item.

See #827

### DIFF
--- a/.changeset/popular-papayas-kneel.md
+++ b/.changeset/popular-papayas-kneel.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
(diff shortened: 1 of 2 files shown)


marclave (approved):
Nice work ✨

PR #836, opened 2024-01-19
feat: webhooks

hanspagel:
This PR adds support for webhooks.

The OpenAPI spec allows multiple webhooks with different request methods, that’s why we should consider to keep the request method badges.

The data structure is pretty much the same as for operations, but the logic is different. Webhooks are initiated by the API, that’s why the rendering needs to be slightly different: I suggest to use _Payload_ instead of _Request Body_. Also, it doesn’t make sense to add the request examples or the API client.

<details>
<summary>Example Spec</summary>

```json
openapi: 3.1.0
info:
  title: Webhook Example
  version: 1.0.0
# Since OAS 3.1.0 the paths element isn't necessary. Now a valid OpenAPI Document can describe only paths, webhooks, or even only reusable components
webhooks:
  # Each webhook needs a name
  newPet:
    # This is a Path Item Object, the only difference is that the request is initiated by the API provider
    post:
      summary: newPet
      description: Information about a new pet in the system
      requestBody:
        description: Information about a new pet in the system
        content:
          application/json:
            schema:
              $ref: "#/components/schemas/Pet"
      responses:
        "200":
          description: Return a 200 status to indicate that the data was received successfully
components:
  schemas:
    Pet:
      required:
        - id
        - name
      properties:
        id:
          type: integer
          format: int64
        name:
          type: string
        tag:
          type: string

```

</details>

**Preview**

Screenshot 2024-01-19 at 16 01 56

Screenshot 2024-01-19 at 16 13 48


marclave (approved):
amazing work @hanspagel ✨ you're a rockstar

we can merge once the conflicts are resolved

PR #844, opened 2024-01-22
fix: Restoring missing forward slash between URL and API routes

dotfortun:
**Problem**
Currently, if your `window.location.origin` doesn't end with a trailing slash and your OpenAPI docs' servers have a leading forward slash, the slash between the URL and the route is removed.

**Explanation**
This happens because the fix for doubling up on forward slashes didn't check if there was no forward slash before removing the leading slash from the server.

**Solution**
With this PR there is now a check to see if `window.location.origin` has a trailing slash before removing the leading slash from the server.


hanspagel:
I just ran into this and was prepared this PR, great timing. 😅 Thanks!

dotfortun:
> I just ran into this and was prepared this PR, great timing. 😅 Thanks!

Glad to help!

PR #846, opened 2024-01-22
fix: additional properties false is ignored

hanspagel:
We’ve added support for `additionalProperties`, but a valid value is `false` - which means there are no additionalProperties … This PR doesn’t add additionalProperties if it’s set to `false. :)


marclave (approved):
ah great fix! :sparkles:

PR #848, opened 2024-01-22
fix: deal with incomplete schemas

hanspagel:
All schemas are beautiful, but not all are complete. With this PR we deal better with them. 💅 

**Before**
Screenshot 2024-01-22 at 15 59 25

**After**
Screenshot 2024-01-22 at 15 59 53

**Example Spec**
https://sandbox.scalar.com/e/OlTLp


marclave (approved):
nice fix ✨

PR #849, opened 2024-01-22
feat: operation scoped security, fix #532 fix #843

### LINKED ISSUE #532: Endpoints get rendered with security scheme even if they don't have it
### Expected Behavior

A security scheme defined on the document root should be applied only for endpoints which refer to the scheme through the `security` object.

### Actual Behavior

For example, a Bearer scheme defined on top but a endpoint that doesn't require it: 
image

### Reproduction steps

```bash
Add a security scheme to the root


"components": {
    "securitySchemes": {
      "bearerAuth": {
        "type": "http",
        "scheme": "bearer",
        "bearerFormat": "JWT"
      }
    },
    "schemas": {}
  }


Then add a endpoint without a security object, you should see it requiring the security scheme anyway
```

### Version

latest

Chrome

### LINKED ISSUE #843: don’t make public endpoints look like they require auth
another note, is that exact same schema will look like this in swagger-ui
(notice /auth endpoints are public)

image

but with scalar. all endpoints will show as-if they require auth header (screen shot from https://sandbox.scalar.com/e/Ojam_#tag/auth)

image

_Originally posted by @alzalabany in https://github.com/scalar/scalar/issues/839#issuecomment-1902057936_

hanspagel:
Currently, we ignore whatever is configured in `security` on the operation level. With this PR, we’ll take the configuration into account.

* When there’s nothing configured, we take the globally configured security schemes.
* When there’s an empty array `[]`, auth is optional and won’t be added to the examples.
* When there’s an array containing an empty object `{}`, auth is optional and won’t be added to the examples.

Read the OpenAPI spec here: https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md

I’ve also tested Swagger UI and got the same results.

See #532, #843

### DIFF
--- a/packages/api-reference/src/components/Content/ReferenceEndpoint/ExampleRequest.vue
+++ b/packages/api-reference/src/components/Content/ReferenceEndpoint/ExampleRequest.vue
@@ -12,7 +12,6 @@ import {
   getRequestFromAuthentication,
   getRequestFromOperation,
   getUrlFromServerState,
-  mergeAllObjects,
 } from '../../../helpers'
 import { useSnippetTargets } from '../../../hooks'
 import { useGlobalStore } from '../../../stores'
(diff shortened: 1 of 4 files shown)


marclave:
very excited to get this one in @hanspagel. looks like one of the tests 

```
@scalar/api-reference:test
```

is failing, once fixed up I'll dive into it :)

marclave (approved):

Issue #835, opened 2024-01-19
Error when scrolling inside Next.js


ginkcode:
### Expected Behavior

Integrate scalar with Next.js with [veaury](https://github.com/devilwjp/veaury)
Things should work fine when scrolling.

### Actual Behavior

After click on any specific link and then scrolling, a bunch of errors throw in the console
<img width="751" alt="Screenshot 2024-01-19 at 8 47 35 PM" src="https://github.com/scalar/scalar/assets/9301098/3c504d1b-6dee-4db5-ba23-cab56d6cbac2">

### Reproduction steps

```bash
1. Go to Api Document page (setup inside Next.js with veaury)
2. Click on any link in the sidebar
3. Scroll up / down
4. See error
```

### Version

1.13.10

### What browsers are you seeing the problem on?

Chrome, Safari (macOS), Firefox

amritk:
hey @ginkcode do you have some example code of your setup? Also which version of Next.js were you using?

ginkcode:
@amritk I'm using Next.js 14
Let me extract a sample to reproduce it

ginkcode:
Here it is
https://github.com/ginkcode/reproduce-scalar-issue

amritk:
Thanks for doing that! I will check it out

ginkcode:
Sounds cool. I’m willing to assist.

amritk:
hey @ginkcode I got a PR up if you want to check it out. You can clone the repo and test out the example if you like

https://github.com/scalar/scalar/pull/854

😄

ginkcode:
Thank so much @amritk 
Now we can skip `veaury` and import ApiReference directly.
But there's still some issues.
The new adapter has peer dependencies with next "^13". So when install this package with Next.js 14 we will see this error
```sh
npm ERR! Found: next@14.1.0
npm ERR! node_modules/next
npm ERR!   next@"14.1.0" from the root project
npm ERR! 
npm ERR! Could not resolve dependency:
npm ERR! peer next@"^13" from @scalar/nextjs-api-reference@0.1.0
```

And the exception throwing when scrolling after click any link in the sidebar is still there. I'm not sure this is Scalar or Next issue. Because it seems Next override some behaviors on `window.history` also.

hanspagel:
Oops! Maybe like this: https://github.com/scalar/scalar/pull/865

amritk:
[...]

PR #854, opened 2024-01-23
feat(#835): create nextjs adapter for references (doc-1476)

amritk:
Added a next.js component so we no longer need to use veaury. This works for now, but eventually we should add a proper server side rendered adapter. 

To test you can cd into `./examples/nextjs-api-reference` and `pnpm dev`


hanspagel (approved):
🔥 SO GOOD 🔥

PR #856, opened 2024-01-23
feat: render images from markdown, filter out XSS vulnerable tags #842

SebastianBienert:
**Problem**
Currently we can't render images from markdown content included in OpenAPI spec

**Explanation**
This happens because img tags are filtered out in [MarkdownRenderer.vue](https://github.com/scalar/scalar/blob/d9b6d8104dd0f1997e45a5220c8bbc11ae2062c4/packages/api-reference/src/components/Content/MarkdownRenderer.vue#L25C1-L30C9) component

**Solution**
With this PR I exclude img tags filtering and add filtering out tags vulnerable for XSS attacks (`script`, `iframe`, `object`)

### DIFF
--- a/packages/api-reference/src/components/Content/MarkdownRenderer.vue
+++ b/packages/api-reference/src/components/Content/MarkdownRenderer.vue
@@ -25,7 +25,7 @@ watch(
       .use(rehypeSanitize, {
         ...defaultSchema,
         tagNames: defaultSchema.tagNames?.filter(
-          (tag) => !['img'].includes(tag),
+          (tag) => !['script', 'iframe', 'object'].includes(tag),
         ),
       })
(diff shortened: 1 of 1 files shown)


hanspagel:
Thanks for the PR! `script`, `iframe` and `object` are removed by [rehype-sanitize](https://github.com/rehypejs/rehype-sanitize) anyway and are not part of the allowed `tagNames` in the `defaultSchema` we use here:

https://github.com/syntax-tree/hast-util-sanitize/blob/main/lib/schema.js#L147-L204

I think we can just remove the options. :)

hanspagel:
Wait, I think we want to make this an option. I’ll take your PR as a starting point and add an option to the MarkdownRenderer.

PR #857, opened 2024-01-23
feat: export SidebarGroup, SidebarElement from api-reference #781

SebastianBienert:
**Problem**
Currently, SidebarGroup and SidebarElements are not exported from `api-referance` package. SidebarElement and SidebarGroup components could be used to extend Sidebar with custom static content while preserving current design and behaviour

**Explanation**
This happens because this is how it was designed and implemented.

**Solution**
With this PR I export SidebarGroup and SidebarElement components.


hanspagel (approved):
Great idea! Let’s get this in.

PR #864, opened 2024-01-23
feat: better schema support for parameters, fix #862

### LINKED ISSUE #862: Constraints for params
Currently all these fields show the type and if they are required, Scalar should also display additional constraints like `pattern`, `minLength`, `maxLength`, etc

_Originally posted by @thelinuxlich in https://github.com/scalar/scalar/discussions/860_

hanspagel:
We have this powerful `Schema` component to render schemas, but that wasn’t really used for parameters (except for objects).

With this PR it’s used for the parameters aswell. And I’ve added support for minLength, maxLength and pattern.

Schemas are complex and allow a ton of features, we’re still not 100 % feature complete, but we’re getting there … ✨

Screenshot 2024-01-23 at 16 57 51

See #862

### DIFF
--- a/.changeset/itchy-rivers-occur.md
+++ b/.changeset/itchy-rivers-occur.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
+feat: better schema support for params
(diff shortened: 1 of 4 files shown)


marclave (approved):
we just keep getting better and better ✨

PR #866, opened 2024-01-23
fix: move next.js adapter to api routes

amritk:
It turns out, next.js was highjacking the window.history object so I moved the references to an API route. This prevents any next.js client side interference. 

You can test it the same way, cd into `./examples/nextjs-api-reference` and `pnpm i && pnpm dev`


hanspagel (approved):
Great PR! I’ve just changed the default path to match with the other packages and examples. 💅

PR #872, opened 2024-01-25
style: no text wrap for method badges, fix #871

### LINKED ISSUE #871: patch label in sidebar is wrapper
### What happens?

image

no wrapping

seems like it's reproducible on the default swagger editor setup

### Swagger/OpenAPI Example

https://sandbox.scalar.com/e/N2Uai

hanspagel:
This PR improves the styling of the method badge in the sidebar.

* Text doesn’t wrap anymore.
* The size of the badge (width, height, line-height) is configured in em now, so it’s relative to the font size.
* The margin is removed, it’s using `gap` now.
* The badge is set to `overflow: hidden`.

**One thing changed:**

The link and the badge were set to `flex-start`, so they were attached to the top. For long operation names the badge was still on top, I think that was expected.

Unfortunately, this required also to fix the position of the badge for one-line operation names with a `top: 3.5px;` to make it look centered. Which feels very hacky to me, and doesn’t work well with a dynamic (em) size.

That’s why I actually centered it now:

Screenshot 2024-01-25 at 11 55 04

See #871

### DIFF
--- a/.changeset/angry-crabs-double.md
+++ b/.changeset/angry-crabs-double.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
+fix: text in method badges wraps for PATCH operations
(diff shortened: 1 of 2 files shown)


marclave (approved):
😎

cameronrohani:
I'll come up with a better solution to keep it top aligned soon. I need to spend a few hours on the sidebar in the next week or two :)

PR #873, opened 2024-01-25
chore: improve types

hanspagel:
Fortunately, this Tweet didn’t take off. Anyway, it’s time to improve the types, at least a bit. 🤡 

https://twitter.com/FreckleBytes/status/1744934392588616082

I’ve added better types to our `Spec` type, was able to remove some of our own types, properly type the security schemes in most places.

But I’ve also had to add some additional @ts-ignores, mostly in tests and in some places where I just didn’t know better. Still lots to do …

### DIFF
--- a/.changeset/gentle-buses-lie.md
+++ b/.changeset/gentle-buses-lie.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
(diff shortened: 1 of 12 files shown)


marclave (approved):
✨ some better types! Love it

PR #875, opened 2024-01-25
fix: example spec on mobile

amritk:
Fixes setting the example spec on mobile.

1. To test shrink the screen to mobile size and open the [editable references](http://localhost:5050/editable-api-reference)
2. hit getting started then one of the example specs
3. they should fill the content now

Test this on main and it does not work, because `ModernLayout` doesn't do anything with the `updateContent` event

@marclave double checked and setting the rawSpec triggers a watcher which updatesContent so its safe
```ts
watch(rawSpecRef, () => {
  emit('updateContent', rawSpecRef.value)
})
```

### DIFF
--- a/.changeset/silver-actors-wonder.md
+++ b/.changeset/silver-actors-wonder.md
@@ -1,0 +1,10 @@
+---
+'@scalar/express-api-reference': patch
+'@scalar/fastify-api-reference': patch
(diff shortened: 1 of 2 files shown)


hanspagel (approved):
Oh, great fix!

PR #879, opened 2024-01-26
chore: add the next Townhall to the README

hanspagel:
Let’s add an information about the next townhall to the root README. :)


marclave (approved):
✨

PR #880, opened 2024-01-26
chore: clean up use CodeMirror

hanspagel:
We’ve had some light/dark mode features in CodeMirror, that we didn’t use. We’re actually only using one theme which then uses the CSS variables. So to force CodeMirror to use a specific mode, you just need to wrap it in a container with a `light-mode` or `dark-mode` class.

This PR removes those unused features. Funfact: We even had `const isDark = ref(false)` and used it as a static value. :) (My bad.)


marclave (approved):
wow great work! Lots cleaned up :)

PR #881, opened 2024-01-26
chore: add Relative CI to monitor the bundle size

hanspagel:
Relative CI does “In-depth bundle analysis and monitoring” and could help us to keep an eye on our bundle size.

More and more users install our packages, and their users probably download the bundle probably a million times already. I think we should be careful with adding more dependencies, and should actively look for ways to reduce the bundle size.

Once the integration works, there are a few possibilities. I’m mostly looking forward to an automated PR comments with the bundle size changes. What I’ve configured for testing purposes:

* New dependencies require approval
* Duplicate dependencies, require approval.
* If the bundle grows by more than 10 kb, require approval.
* If the bundle grows by more than 2 %, the GitHub check will fail.
* If the bundle grows by less than 2 %, or even shrinks, the GitHub check will be green.

Probably need to fine tune/deactive some of them eventually.

BTW it’s free for open source projects

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,7 +35,8 @@ jobs:
         run: pnpm turbo types:check
       - name: Run tests
         run: pnpm turbo test
-      - name: Send bundle stats and build information to RelativeCI
+      - if: matrix.node-version == 20
+        name: Send bundle stats and build information to RelativeCI
         uses: relative-ci/agent-action@v2
         with:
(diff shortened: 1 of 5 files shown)


marclave (approved):
I love it! but it looks like it fails on the bundle size output check for adding this package 😆 

we can just merge probably or we can make the limit higher ✨

hanspagel:
Yeah, because the bundle size increased from 0 to 1.8mb. ;) We can ignore it for this PR.

PR #887, opened 2024-01-28
fix: add explicit table alignment rules

ShadiestGoat:
**Problem**
Currently, tables do not respect markdown column alignment.

**Explanation**
This happens because align key is deprecated & some browsers potentially ignore it now. You also use `text-align: left`, but disabling that doesn't fix it so I'm not 100% sure.

**Solution**
With this PR, I add explicit column alignment rules.

Note on implementation - its not a particularly graceful solution, if loops were available with something like sass or less, or if [attr](https://developer.mozilla.org/en-US/docs/Web/CSS/attr) was [supported by browsers](https://caniuse.com/css3-attr), that would be nicer but its not a huge deal.

### DIFF
--- a/.changeset/angry-mayflies-add.md
+++ b/.changeset/angry-mayflies-add.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': minor
+---
+
+Fix markdown table alignment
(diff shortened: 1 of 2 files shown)


marclave (approved):
awesome work @ShadiestGoat ! 

we have much more alignment now! ✨

PR #890, opened 2024-01-28
feat: add default state for api client collapsed panels

marclave:
**Problem**
We always show all collapsed sections even if there's no content, this can add visual clutter

https://github.com/scalar/scalar/assets/6176314/df23e939-74b8-4b9a-ad97-a0d07083105a

**Explanation**
We don't check to see the content in the section

**Solution**
We watch to see the active request, i had to do some magic with the headlessui component. but you can't have too much magic ✨ 

We might want to have a slight placeholder to show that there's nothing in the section (i.e. No Variables)

https://github.com/scalar/scalar/assets/6176314/0ecf52b1-512a-43f7-bb09-019c6689e667

### DIFF
--- a/packages/api-client/src/components/ApiClient/Request/RequestBody.vue
+++ b/packages/api-client/src/components/ApiClient/Request/RequestBody.vue
@@ -1,5 +1,6 @@
 <script setup lang="ts">
 import { CodeMirror } from '@scalar/use-codemirror'
+import { computed } from 'vue'
 
 import { useRequestStore } from '../../../stores/requestStore'
 import { CollapsibleSection } from '../../CollapsibleSection'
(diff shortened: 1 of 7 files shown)


amritk:
Video looks clean

hanspagel (approved):
Looks awesome to me

packages/api-client/src/components/ApiClient/Request/RequestBody.vue
```diff
@@ -1,5 +1,6 @@
 <script setup lang="ts">
 import { CodeMirror } from '@scalar/use-codemirror'
+import { computed } from 'vue'
```
hwkr, line 3: Is this being used in here?

hwkr (approved):
lgtm

packages/api-client/src/components/ApiClient/Request/RequestBody.vue
marclave: no it's not, I can remove it good catch!

PR #895, opened 2024-01-29
fix: response examples with a single value do not render, fix #894

### LINKED ISSUE #894: response examples with a single value trigger error and don't render
### What happens?

Example doesn't render if examples contains only a single value.

Here's a snippet. In this case we have an examples object with only a single option.

```yaml
paths:
  /user:
    post:
      responses:
        '409':
          description: Conflict
          content:
            application/json:
              examples:
                username_exists:
                  summary: "username already taken"
                  value: {
                    "errors":[
                      {
                        "type": "conflict",
                        "message": "username-exists"
                      }
                    ]}
```

The 409 example should render.

I think this code needs to change:

https://github.com/scalar/scalar/blob/c0348981dea70d9276c25e47440e7452a1915ad7/packages/api-reference/src/components/Content/ReferenceEndpoint/PathResponses/ExampleResponse.vue#L46

I believe it should be

```ts
prettyPrintJson(Object.values(response?.examples)[0].value) 
```

1. Go to 
2. Click the "409" example and verify no example is shown
3. Verify in developer console an error is logged: `TypeError: Cannot read properties of undefined (reading 'response')`

### Swagger/OpenAPI Example

https://sandbox.scalar.com/e/930ux

hanspagel:
There was a typo within the example responses. This PR fixes it. :)

For more context, see #894.

### DIFF
--- a/.changeset/slimy-paws-confess.md
+++ b/.changeset/slimy-paws-confess.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
+fix: response examples with a single value don’t render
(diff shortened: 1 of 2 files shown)


marclave (approved):
Nice fix :)

PR #911, opened 2024-02-01
feat: details and summary tags in markdown

hanspagel:
This PR adds support for details and summary tags in the MarkdownRenderer.

* HTML is allowed now, but runs through a sanitizer
* the unified/remark/rehype dependencies are updated
* the second MarkdownRenderer in @scalar/swagger-editor isn’t used anymore, I’ve removed it form the code base
* some other unused deps are removed from @scalar/swagger-editor

Screenshot 2024-02-01 at 12 29 45

See #910.


marclave (approved):
Awesome work @hanspagel !

PR #913, opened 2024-02-01
fix: swagger 2.0 response body isn’t shown

hanspagel:
Currently, we ignore the Swagger 2.0 response body. This PR fixes it.

See #912


marclave (approved):
Ah!!! Amazing work @hanspagel

PR #922, opened 2024-02-01
feat: collapse all subschemas by default

hwkr:


marclave (approved):
performance gains!!! Great work @hwkr

PR #924, opened 2024-02-02
feat: add sticky card headers

hwkr:


marclave (changes requested):
awesome work @hwkr ! I played around with it and it's feeling better so much better with the sticky card headers!

Would be nice to make the show more attributes in the request body not full width as before, after playing around ✨.

hanspagel:
Love it, great PR!

hwkr:
> awesome work @hwkr ! I played around with it and it's feeling better so much better with the sticky card headers!
> 
> Would be nice to make the show more attributes in the request body not full width as before, after playing around ✨.

Updated with narrow button 👍 

Google Chrome-2024-02-02-13-06-40@2x

marclave (approved):
✨ let's go!!!!

PR #927, opened 2024-02-02
fix: css reset on markdown

marclave:


cameronrohani (approved):
we'll need a fix for details but we need this in asap :)

PR #936, opened 2024-02-06
feat: Migrate most logic to useCodemirror

geoffgscott:
Moves most logic to the Codemirror hook for better encapsulation. 

Two way data binding is not currently working. Appears to be an issue where setting the spec does not update the visible codemirror content.


marclave (changes requested):
awesome work! doing some testing and first round I found these 

- [ ] api client URL is missing
image
- [ ] API Client Response is missing 
image
- [ ] API Client Request Body doesnt change when switching endpoints 
- [ ] Show Schema doesnt work 
image
- [ ] operation Show Schema doesnt work 
- [ ] operation Response doesnt work 
image


will do a deeper dive after these meetings ✨

packages/use-codemirror/src/components/CodeMirror/CodeMirror.vue
```diff
@@ -1,44 +1,12 @@
...
 import { useCodeMirror } from '../../hooks'
 import type { CodeMirrorLanguage } from '../../types'
-import { variables } from './extensions/variables'
 
 const props = withDefaults(
   defineProps<{
-    name?: string
-    extensions?: Extension[]
-    content?: string
+    content: string | undefined
```
hanspagel, line 9: I don’t know what the difference is between those two. Can you enlighten me?

hanspagel:
What an effort! Awesome work!

packages/use-codemirror/src/components/CodeMirror/CodeMirror.vue
geoffgscott: The first is an optional property. The second is a required property but the variable provided can be undefined. 

The advantage of the second is a user can't forget to assign a value to it.

marclave (approved):
awesome work! @geoffgscott 

can add a changeset and we can merge it ✨

PR #941, opened 2024-02-06
fix: remove auto open tags on scroll

amritk:
**Problem**
Tags will auto-open as you scroll down

**Solution**
With this PR we remove that auto-open. Maybe we can add it as an option down the road when we fix performance.


marclave (approved):
🔥

PR #945, opened 2024-02-07
fix: shrink label and not actions on mobile

hwkr:


marclave (approved):
shrinking some bugs!

we love to see it, great work @hwkr ✨

PR #950, opened 2024-02-07
fix: expect a blob when fetching spec via proxy

hwkr:
I didn't know this was using the API Client proxy when I changed it to return blobs 🤦


marclave (approved):
nice fix! ✨

PR #955, opened 2024-02-08
fix: parse axios error responses as blobs

### LINKED ISSUE #939: Response body is returning `{}` on requests that don't return `200`
### What happens?

As discussed on [discord](https://discord.com/channels/1135330207960678410/1135330208627564597/1204433174579515443), when creating a request to a any type (such as Post, Patch, Delete, Get) of route through Scalar's API Reference that doesn't return `200`, the response body is `{}` only, the rest of the response (such as headers) appears as expected. 

Additionally, no errors are reported in the console on the client or through the Hono client, where the API Reference/Documentation is running.

The response body should be the response given by the route, as it is listed correctly in the console.
image

https://github.com/wanderer-moe/api/blob/v2/src/index.ts

### Swagger/OpenAPI Example

_No response_

hwkr:
This PR does 3 things:
1. Updates `sendRequest` to still parse the data correctly when Axios throws an error for a non-200 status code
2. Updates the client proxy's `Access-Control-Expose-Headers` header to only include our special scalar proxy headers rather than exposing everything (which was showing a bunch of weird headers).
3. Adds a dev toolbar to the api client dev environment to make it easier to test

fixes #939

# Before (main)

https://github.com/scalar/scalar/assets/6374090/cf90d527-b22f-4d17-8073-52d55677d334

# After (this PR)

https://github.com/scalar/scalar/assets/6374090/4ed54f7f-e3df-487c-96aa-f07a98b0560f

### DIFF
--- a/.changeset/blue-tomatoes-tan.md
+++ b/.changeset/blue-tomatoes-tan.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
+
+fix: parse axios error responses as blobs
--- a/.changeset/rare-moons-run.md
+++ b/.changeset/rare-moons-run.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client-proxy': patch
+---
+
+fix: only expose special proxy headers instead of \*
--- a/examples/web/src/components/DevApiClientOptions.vue
+++ b/examples/web/src/components/DevApiClientOptions.vue
@@ -1,0 +1,61 @@
+<script setup lang="ts">
+import { type ThemeId, availableThemes } from '@scalar/themes'
+import { computed } from 'vue'
+
+type ApiClientConfig = {
+  proxyUrl?: string
+  readOnly?: boolean
+  theme?: ThemeId
+}
+
+const props = defineProps<{
+  modelValue: ApiClientConfig
+}>()
+
+const emit = defineEmits<{
+  (e: 'update:modelValue', v: ApiClientConfig): void
+}>()
+
+/** Computed config proxy for v-model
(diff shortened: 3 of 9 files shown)


amritk (approved):
LGTM.  One thought: we should standardize how we return/handle errors across the board. @geoffgscott

amritk (changes requested):
On second thought, testing on the deployed app.

image


Not sure if this has to do with the deployment or the code

hwkr:
> On second thought, testing on the deployed app.
> 
> 
> 
> image
> 
> 
> 
> 
> 
> Not sure if this has to do with the deployment or the code

Hmm that doesn't seem right, what was happening when you did that?

hwkr:
> Code LGTM. One thought: we should standardize how we return/handle errors across the board. @geoffgscott

Like in terms of throwing errors? Or how they're returned by APIs?

amritk:
> > Code LGTM. One thought: we should standardize how we return/handle errors across the board. @geoffgscott
> 
> Like in terms of throwing errors? Or how they're returned by APIs?

Returned by API's and handled by the FE

hwkr:
> > > Code LGTM. One thought: we should standardize how we return/handle errors across the board. @geoffgscott
> > 
> > 
> > Like in terms of throwing errors? Or how they're returned by APIs?
> 
> Returned by API's and handled by the FE

Yeah it would be nice to type them at very least 😅

amritk (approved):
Turned out to be a preview error

chlorophant:
I would love for this to be fixed. Can we get this merged?

marclave (approved):

marclave:
@chlorophant released in `@scalar/api-reference@1.15.4` ✨

PR #959, opened 2024-02-08
fix: deploy preview workflow triggers

tmastrom:
With this PR the deploy preview workflow will trigger on PR open as well as on push if there are changes to the UI. This PR also clarifies workflow naming, adds a header to the deploy preview comment and explicitly adds permissions although this may be unnecessary for a public repo.


marclave (approved):
awesome :)

PR #960, opened 2024-02-08
feat: Bundle codemirror dependencies

geoffgscott:
Centralizes all the codemirror dependencies into a single bundled hook to eliminate the frequent package update issues. 

Also adds language support for CSS, Rust, and PHP (which is likely most of the bundle size increase)


amritk (approved):
Legendary move

marclave (approved):

PR #967, opened 2024-02-09
feat: allow configuration of request proxy via env variable or dev toolbar

hwkr:
Now we can set the `VITE_REQUEST_PROXY_URL` to be `https://staging.scalar.com/request-proxy` for our deployed previews if we want (or leave it blank and it won't use a proxy at all).

Google Chrome-2024-02-08-17-03-00@2x
Google Chrome-2024-02-08-17-02-23@2x


marclave (approved):

marclave (commented):
it would be nice if all our examples defaulted to the proxy `http://localhost:5051` without the need to change the `.env` manually : )

once that gets changed we can merge it ✨

hwkr:
Okay I switch `.env.example` to `.env.development`, it can still be overridden local by creating a [`.env.development.local` file](https://vitejs.dev/guide/env-and-mode#env-files).
Code-2024-02-13-11-35-42@2x
Google Chrome-2024-02-13-11-35-45@2x

PR #977, opened 2024-02-09
chore: use vitest workspaces

hanspagel:
I just found out that vitest has a “workspaces” feature, which we could use to run the tests. The output is better than the output of `pnpm -r test`.

But one test is failing and I don’t get why. 🤔 I think the expected output isn’t really related to the test, though:

```diff
    expect(wrapper.html()).toBe(
-      `<!--teleport start-->
-<!--teleport end-->
<div style="position: fixed; height: 0px; padding: 0px; overflow: hidden; clip: rect(0px, 0px, 0px, 0px); white-space: nowrap; border-width: 0px; display: none;"></div>`,
    )
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,7 +34,7 @@ jobs:
       - name: Check types
         run: pnpm turbo types:check
       - name: Run tests
-        run: pnpm turbo test
+        run: pnpm test
       - if: matrix.node-version == 20
         name: Send bundle stats and build information to RelativeCI
         uses: relative-ci/agent-action@v2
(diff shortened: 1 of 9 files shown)


hwkr:
To be fair we shouldn't really be comparing html output for our tests, makes them pretty susceptible to small config changes like this. Also I have no idea why there would be a teleport there.

hwkr:
OMG THE VSCODE EXTENSION WORKS (kinda)

Code-2024-02-09-14-17-12@2x

hwkr:
Okay! I updated the test to make it more robust (although it still doesn't really test anything) and pushed the vscode config for vitest that was working well for me

hanspagel:
AWESOME

PR #979, opened 2024-02-09
feat: lazy load api reference

amritk:
THIS PR IS NOT COMPLETE, just for testing

~~Will remove extra specs and debug data before merge~~

Things to test:
- mobile performance to see if we need to add in viritualization
- SSG to make sure it still works, just need to add in the flag really and test deep linking


amritk:
This PR is being closed in favour of a rebased one

PR #980, opened 2024-02-09
fix: remove type="text" from the search button placeholder span

hwkr:
This is a span within a button so it shouldn't really have a type attribute regardless.

I'm guessing it was left there from a time when the search input was an actual input instead of a button that opened the modal.


marclave (approved):
✨

PR #987, opened 2024-02-12
chore: remove use keyboard event

hanspagel:
This PR removes the use-keyboard-event package. Probably a matter of taste, but my arguing is:

* we should focus on API tools (less responsibility, smaller repo)
* vueuse is well tested and maintained

I know that @hwkr fixed some bug with the KeepAlive routing component. Would you mind to check whether I re-introduced that bug?


marclave:
@hanspagel only thing is we do use this in the internal org, so we would need to handle that fix ✨ 

im pretty open to either or, I think as time moves on vue use will update this clone of package with improvements, and we get tree shaking so we dont get too much bloat in it.

hwkr:
I agree with @marclave! I like vueuse.
  
@hanspagel Are we planning on refactoring the internal usages before we merge this? If so I'll test it all together internally.

You can check the old bug by switching from the guide to references, loading the search with the hotkey and then switching back to the guides and trying to load the search with the hotkey again. It shouldn't open when you're on the guide rather than the references.

hanspagel:
I mean the package will still be on npm, so there’s no urgency to do a refactor now, but I think that’s the way forward.

If we’d prefer to not do this, I’d suggest to move `useKeyboardEvent` to the components package.

If we’d like to do this, could you look into this @hwkr?

packages/api-client/src/components/ApiClient/AddressBar.vue
```diff
@@ -20,6 +20,9 @@ const emits = defineEmits<{
   (event: 'onSend'): void
 }>()
 
+const keys = useMagicKeys()
```
hwkr, line 23: Do we use the variable `keys` here?

hwkr: Nvm right below 🤦

marclave (approved):

PR #993, opened 2024-02-13
chore: remove use toasts

hanspagel:
With this PR we’re removing the use-toasts plugin and migrate to vue-sonner. Currently, there’s only one toast notification. It’s displayed whenever you copy something (for example the base URL).


packages/api-reference/src/components/ApiReferenceBase.vue
```diff
@@ -145,7 +145,21 @@ if (props.configuration?.authentication) {
     {{ currentConfiguration.customCss }}
   </component>
   <ThemeStyles :id="currentConfiguration?.theme" />
-  <FlowToastContainer />
+  <Toaster
+    :toastOptions="{
+      style: {
+        'padding': '18px',
```
hwkr, line 151: Can we configure this some other way than a big style block?

hanspagel: Yess! Added a class and made a CustomToaster component to hold the styles.

PR #994, opened 2024-02-13
chore: upgrade to Vite 5

hanspagel:
Let’s upgrade from Vite 4 to 5. I tried it a while ago, but the `@scalar/fastify-api-reference` package build process is kind of custom (it’s using Vite to build a Node package), but I was able to fix this today.

I thought there are some performance benefits, too, but it’s minor.

**Vite 4**
Time: 38.504s

**Vite 5**
Time: 35.981s 

This PR removes `vite-plugin-node-polyfills` from @scalar/api-reference (but not from the standalone/CDN build!). It never did what we wanted it do. We still had to add the same plugin in our projects anyway. BUT, we should double check once released.


hanspagel:
More duplicate packages … 🤔

marclave (approved):
rad! im down to merge but i'll leave it for you @hanspagel to do a release ✨ + test

hanspagel:
Let’s get this in. Very unlikely something breaks, and if it does, we can re-add the polyfills in a heartbeat.

PR #995, opened 2024-02-13
chore: serve OpenAPI file for express example

hanspagel:
I was just looking for TODO comments in the code base and saw that we served a `swagger.json` in the express example, but didn’t even use it.

I consider serving an OpenAPI file best practice, so I did that and removed the TODO comment.


marclave (approved):
great work ✨

PR #1000, opened 2024-02-13
chore: switch tailwind to use px values

### LINKED ISSUE #944: Component Library uses `rem`s but font size isn't set on `<html>`
### What happens?

The component Library uses `rem`s but font size isn't necessary set on the `<html>` element by the consumer of the component library leading to unexpected behaviour.

image

Sizes should be consistent across uses

Create an icon in the `api-references` package and try to use the `size` attribute.

### Swagger/OpenAPI Example

_No response_

hwkr:
fixes #944

### DIFF
--- a/packages/api-reference/src/components/Content/Operation/OperationAccordion.vue
+++ b/packages/api-reference/src/components/Content/Operation/OperationAccordion.vue
@@ -58,7 +58,7 @@ const { getOperationId } = useNavState()
         class="endpoint-copy"
         icon="Clipboard"
         label="Copy endpoint URL"
-        size="full"
+        size="sm"
         variant="ghost"
         @click.stop="copyToClipboard(operation.path)" />
     </template>
@@ -178,7 +178,6 @@ const { getOperationId } = useNavState()
 
 .endpoint-copy {
   opacity: 0.44;
-  height: 14px;
 }
 .endpoint-copy:hover {
   opacity: 1;
--- a/packages/components/env.d.ts
+++ b/packages/components/env.d.ts
@@ -2,3 +2,4 @@
 
 declare module '*.module.css?inline'
 declare module 'vite-plugin-scope-tailwind'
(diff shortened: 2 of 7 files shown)


hanspagel:
I have no idea how that works (I’m too dumb) and the answer is probably yes, but do we want a 2 year old dependency with 2 stars?

If not, we could also just do a search and replace for the values defined in the plugin:
https://github.com/rise8-us/tailwind-pixel-perfect-preset/blob/main/index.js#L1-L213

Anyway, I trust you, add whatever you think is best. 👍

amritk (approved):
Yes so the downside of using that unmaintained package is, if tailwind adds any more options they won't get converted right?

I thought they would just have had config option for this.

Two options:
- keep the plugin, if it ever comes up that we need to add more values, open a PR
- just copy the values into our theme config as stated in the [tailwind recommended](https://github.com/tailwindlabs/tailwindcss/issues/1232#issuecomment-557126355) way

Its up to you!

hwkr:
> Yes so the downside of using that unmaintained package is, if tailwind adds any more options they won't get converted right?
> 
> I thought they would just have had config option for this.
> 
> Two options:
> 
> * keep the plugin, if it ever comes up that we need to add more values, open a PR
> * just copy the values into our theme config as stated in the [tailwind recommended](https://github.com/tailwindlabs/tailwindcss/issues/1232#issuecomment-557126355) way
> 
> Its up to you!

Yeah I'm not sure if it's actually unmaintained or just they haven't needed to update anything in while because its just a fairly simple preset. I figure if we ever need to change it we can just bring the one file over to our repo and start managing it ourselves.

I don't love the way tailwind recommends because it would mean adding a couple hundred lines of (imo) boilerplate for what should probably be a single line config change. Using a preset makes it more or less a single line config change which can then be overridden by our tailwind file.

hanspagel:
Cool! Thanks for the explanation 🙌

PR #1005, opened 2024-02-14
refactor: add new api reference editor package

hanspagel:
Currently, we ship `@scalar/api-reference` *with* `@scalar/swagger-editor`. This is great, but adds to the bundle size. The editor not that much (we need CodeMirror anyway), but the `GettingStarted` page (that is only shown when `isEditable: true`) with the two example specs.

But most people use the reference *without* the editor. So let’s add a separate package for that use case: `@scalar/api-reference-editor`. The idea is that people who like to edit their OpenAPI file can just add `-editor` to the package name and have the same API.

**But I need some help!** I don’t know how to replicate the API in `@scalar/api-reference-editor`. Looks like we need to duplicate a lot of the components: `ApiReference`, `ApiReferenceBase`, `ModernLayout` … 🤔  Any ideas how to reuse the existing components as much as possible? @hwkr @amritk 

The bundle size should decrease by 142kb/6%.

**Before**

```diff
→ api-reference → add-new-api-reference-editor-package pnpm build:default
✓ 1174 modules transformed.
+dist/index-16645a27.js       7.54 kB │ gzip:   2.58 kB
+dist/index.js              159.15 kB │ gzip:  23.48 kB
+dist/index-85009e33.js   2,080.18 kB │ gzip: 542.98 kB
```

**After**

```diff
→ api-reference → main pnpm build:default
✓ 1144 modules transformed.
+dist/index.js            2,105.81 kB │ gzip: 545.53 kB
```


amritk:
Alright 100% on board here. But I think we can save it for a refactor? Or do it now? 

It would be nice that references works great in other frameworks too like Next so we can see what else needs to be done at the same time

hwkr:
I think the way to do this (having not tried to do this) would be to create a slot for the editor in ApiReferenceBase (might have that already) and then consume it in our new package and sync up the editor state stuff in the new package.

hanspagel:
I think we should do it now. :)

hanspagel:
That said, if you think Next.js needs some love before that, work in that first. ✌️

marclave:
this is amazing! we gotta do some nextjs/react work but then let's do this ✨ so hopefully we can tackle this first thing next week

PR #1007, opened 2024-02-14
refactor: use own helper instead of javascript-time-ago

hanspagel:
With this PR `javascript-time-ago` is replaced with a tiny helper function in our repository. Should decrease the bundle size by 60kb/3%. :)


marclave (approved):
LOVE this

PR #1012, opened 2024-02-16
fix: updated scroll region padding to for chrome update

hwkr:


cameronrohani:
Awesome, I found another issue related to the chrome update that we need to fix (maybe we just merge this and fast follow though?)

image

Can't see the active states of the tabs anymore

cameronrohani (approved):
amazing let's get this in ASAP 👯

PR #1013, opened 2024-02-16
feat: add lazy loading to API references

amritk:
This PR will require a lot of testing. Also its definitely not perfect (still too many setTimeouts). And I keep running into edge cases. But it works _most_ of the time! 

In v2 we can definitely smooth things out by detecting codemirror load events to replace the timeouts, and add some height estimation functions. As well as virtual scrolling for those really big specs.

Have at it I'm sure there will be much to fix!

P.s. didn't test org after my latest drastic changes so I'll test it tomorrow 🫡

To fix:
- [ ] ~~don't lazy load until index, this should allow the setTimeouts to just be 0 instead~~ booted to v2 for better rendering control
- [ ] add lazy to show schema button
- [ ] check strange scrolling on initial upload (maybe need to disable some things here)
- [x] error handling bad hash
- [x] strange jump on tag routing


packages/api-reference/src/components/Content/Lazy/Lazy.vue
```diff
@@ -0,0 +1,57 @@
+<script lang="ts">
+import { type EventBusKey, useEventBus } from '@vueuse/core'
+
+/**
+ * Setup an event bus so we can listen for loaded events
+ */
+const lazyEventBusKey: EventBusKey<{ id: string }> = Symbol('lazyEventBusKey')
+export const lazyBus = useEventBus(lazyEventBusKey)
```
geoffgscott, line 8: Lets drop this into a vanilla ts file. In general I find exporting/importing from .vue files generally hard to follow

geoffgscott (commented):
Untested but overall looks great to me. Nice work.

packages/api-reference/src/components/Content/Lazy/Lazy.vue
amritk: cool beans, just saw it elsewhere. Will do 🖖🏾

cameronrohani:
did some more testing tonight, found that the modern theme was faster than the classic with with large specs

marclave (approved):

PR #1018, opened 2024-02-17
feat: switch codemirror to prism

amritk:
Should be lighter, definitely has more control, lets see. 

Comes with json, bash, js, and css pre-loaded. Then the rest of the languages are auto loaded on the fly

@cameronrohani we should probably ~~make our own css file which ingests the theme vars~~ exposed some css classes so we can theme the code as well


hanspagel:
Great work! I’d be tempted to rename the component to `CodeBlock`. In HTML, Markdown and rich text `code` refers to inline code. Wdyt?

amritk:
Sounds like a good plan. Or even code highlight or something

packages/api-reference/src/components/CodeBlock/CodeBlock.vue
```diff
@@ -0,0 +1,187 @@
...
+})
+</script>
+<template>
+  <pre
+    :class="{
+      'line-numbers': lineNumbers,
+    }"><code ref="el" :class="`lang-${language}`">{{content}}</code></pre>
+</template>
+<style>
+code[class*='language-'],
```
amritk, line 59: This was just copy/pasted from one of their themes

packages/api-reference/src/components/CodeBlock/CodeBlock.vue
```diff
@@ -0,0 +1,186 @@
...
+  },
+)
+
+const el = ref(null)
+const language = computed(() => {
+  return props.lang === 'node' ? 'js' : props.lang
+})
+
+plugins.autoloader.languages_path =
+  'https://cdnjs.cloudflare.com/ajax/libs/prism/1.29.0/components/'
```
marclave, line 39: i know we are loading these async but should we have this bundled with the repo? :)

amritk: I always prefer cdn over bundle for better performance! But doesn't matter too much these days

marclave: alright im down, we might need to bundle it for some of our integrations but this is cool : )

marclave (commented):
nice work @amritk ! ✨

marclave (approved):

PR #1022, opened 2024-02-20
fix: set max height on request body codemirror

hwkr:
The code was already there just the class got removed from the codemirror element

Google Chrome-2024-02-20-15-50-01@2x


marclave (approved):
nice! 😎

PR #1031, opened 2024-02-22
fix: basic auth shows encoded colon string, fix #1026

### LINKED ISSUE #1026: Basic Authentication displays Og=='
### What happens?

When specifying Basic Authentication as a security schema in an OpenAPI spec, the string `Og=='` is displayed in the rendered code box and _Test Request_ dialog:

image
image

The string `Og=='` is a base64-encoding of the colon character :
It should not be rendered but `Authorization: Basic` should still appear

Include a basicAuth securitySchemes in the openapi spec, for example:

```cli
openapi: 3.0.0
info:
  title: HTTPS + Basic auth example
  version: 1.0.0

servers:
  - url: https://httpbin.org

components:
  securitySchemes:
    basicAuth:
      type: http
      scheme: basic
      description: Use `user` / `passwd` as the test credentials

paths:
  /basic-auth/user/passwd:
    get:
      security:
        - basicAuth: []
      responses:
        '200':
          description: OK
        '401':
          description: Unauthorized. The username or password provided is incorrect.
```

### Swagger/OpenAPI Example

https://sandbox.scalar.com/e/snZUC

hanspagel:
For basic auth we’re encoding username and password, even if both are empty. Instead of `hans:test123` the string that we encode is then just `:`.

With this PR the token is omitted if we don’t have credentials.

It’s totally valid to send just the username or just the pasword, so we’re still showing a token for `:test123` or `hans:`.

Screenshot 2024-02-22 at 12 09 19

### DIFF
--- a/.changeset/lucky-lobsters-cry.md
+++ b/.changeset/lucky-lobsters-cry.md
@@ -1,0 +1,5 @@
+---
+"@scalar/api-reference": patch
+---
+
+fix: http basic auth empty credentials should not be encoded
--- a/examples/web/src/pages/ApiReferencePage.vue
+++ b/examples/web/src/pages/ApiReferencePage.vue
@@ -18,17 +18,17 @@ const configuration = reactive<ReferenceConfiguration>({
   showSidebar: true,
   layout: 'modern',
   spec: { content },
-  authentication: {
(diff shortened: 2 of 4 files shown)


marclave (approved):
nice fix @hanspagel

PR #1040, opened 2024-02-23
fix: quick fix for supporting application/problem+json

amritk:
**Problem**
if your response is of type `application/problem+json` we return No Body
image

https://discord.com/channels/1135330207960678410/1135330208627564597/1210630137326862378

https://datatracker.ietf.org/doc/html/rfc9457


marclave (approved):
Nice fix @amritk !

amritk:
Tested the API client as well

image

Issue #1042, opened 2024-02-24
Using @scalar/fastify-api-reference not working in ESM


big-kahuna-burger:
### What happens?

https://github.com/big-kahuna-burger/scalar-repro

### What did you expect to happen?

For package to work with ESM.

### How can we reproduce the issue?

1. Clone my repro repository
2. `npm i`
3. `node index.js`
4. Observer err:

```
node index.js
file:///<HOME_DIR>/d/f0/services/scalar-repro/node_modules/@scalar/fastify-api-reference/dist/index.js:5
    path.resolve(`${__dirname}/js/standalone.js`),
                    ^

ReferenceError: __dirname is not defined in ES module scope
This file is being treated as an ES module because it has a '.js' file extension and '/<HOME_DIR>/d/f0/services/scalar-repro/package.json' contains "type": "module". To treat it as a CommonJS script, rename it to use the '.cjs' file extension.
    at getJavaScriptFile (file:///<HOME_DIR>/d/f0/services/scalar-repro/node_modules/@scalar/fastify-api-reference/dist/index.js:5:21)
    at fastifyApiReference (file:///<HOME_DIR>/d/f0/services/scalar-repro/node_modules/@scalar/fastify-api-reference/dist/index.js:150:23)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)

Node.js v20.11.1
```

marclave:
@big-kahuna-burger did you fix this by upgrading to the latest version of scalar? Do you know what version this error happened on? :)

big-kahuna-burger:
Oh, never mind. 
It didn't worked on my f0 project when using `npm` instead of `pnpm`. (Different workspaces setup)
Nothing to do with scalar, just manifested like the issue was in scalar.
My bad :)
That's why I closed.

marclave:
> Oh, never mind. 
> 
> It didn't worked on my f0 project when using `npm` instead of `pnpm`. (Different workspaces setup)
> 
> Nothing to do with scalar, just manifested like the issue was in scalar.
> 
> My bad :)
> 
> That's why I closed.

Cool! Just wanted to double check it wasn't a precious broken version we released :)

markusgeert:
I'm encountering this when upgrading `@scalar/fastify-api-reference` from 1.16 to 1.17. It seems to have to do with the removal of `fastify-api-reference/src/vite-plugins/nodeShims.ts` in #994.

I created a PR that fixes this at #1043.

[...]

PR #1043, opened 2024-02-25
fix: fastify-api-reference 1.17 not working with ESM

markusgeert:
**Problem**
Currently, @scalar/fastify-api-reference breaks with version 1.17.x. This seems to be introduced by PR #994. The problem is described by the issue #1042.

**Explanation**
This happens because the `__dirname` variable is not defined in ES module scope.

**Solution**
With this PR the `__dirname` variable is explicitly set in the `getJavascriptFile` function, fixing the bug.


hanspagel:
Oh, you’re so right! Let’s release this quickly. :) Thanks for creating this PR!

hanspagel:
Unfortuntely, another commit slipped into this PR. I’ll create a new one!

Issue #1053, opened 2024-02-26
generated example for objects is wrong


hanspagel:
## Description

We are using Hono for the API, and [Hono Zod Openapi](https://hono.dev/snippets/zod-openapi) to generate the OpenAPI specs.

When creating a field that is a `Record<string, unknown>`, Scalar generates an example with the shape: `field: {}`.
However, when creating a field that is a `Record<string, string>`, Scalar generates an example with the shape: `field: ""`

![image](https://github.com/scalar/scalar/assets/42624869/08b3c220-4e18-4883-aa4d-eb0299e04445)

In the second case, the **example is not valid** since the field is not a string.

## Expected behavior
It would be useful to have examples of keys and values when an object is defined. Also, maybe an empty object as an example could be confusing, since in most of the cases an empty object would not be a valid input and the field could be nullable.

### Redoc (example)
![image](https://github.com/scalar/scalar/assets/42624869/15236ccb-ce25-45a1-a7d2-9d515165ddfc)

- Does it make sense?

## Steps to reproduce

- Add `@scalar/hono-api-reference` to the project;
- Add OpenAPI fields to the OpenAPI json schema.

### Package version
```json
{
    "@scalar/hono-api-reference": "^0.3.56",
}
```
### OpenAPI fields
```json
{
  "additionalAttributes": {
        "type": "object",
        "nullable": true,
        "additionalProperties": {
            "nullable": true
        },
        "description": "Atributos adicionais do item segurado. **O campo será removido futuramente, utilize o campo metadata**.",
        "deprecated": true
    },
    "metadata": {
        "type": "object",
        "nullable": true,
        "additionalProperties": {
            "type": "string",
            "minLength": 1,
            "maxLength": 500
        },
        "description": "Metadados do item segurado: você pode especificar até 50 chaves, com nomes de até 40 caracteres e valores de até 500 caracteres.",
        "maxItems": 50
    },
}
```

_Originally posted by @Amorim33 in https://github.com/scalar/scalar/discussions/1051_

hanspagel:
@Amorim33 wants to work on this ✌️

PR #1054, opened 2024-02-26
fix: show key-value pair examples for objects

Amorim33:
fix: #1053 

## **Problem**
Currently, the generated example for objects is not consistent.

- It does not generate key-value pair examples
- The object example is always generated with the type from `additionalProperties` (e.g. If the `additionalProperties` field has `type: string` definition, then the generated example will be `myObject: ""`

image

## **Explanation**
This happens because the `getExampleFromSchema` function bubbles up the type from `additionalProperties` when overwriting the response.

## **Solution**
With this PR:
- The `getExampleFromSchema` function is changed to:
  - return null example for nullable properties;
  - add an example of key-value pair examples when the object has additionalProperties;
  - do not overwrite response with additionalSchema.
- The `getExampleFromSchema` test is changed to:
  - assert if a key-value pair example is added;
  - assert if a null example is added.
- The `getRequestBodyFromOperation` test is changed to:
  - assert that examples for object schemas are generated correctly for each OpenAPI type. 

> Note: decided to go with `early return` to avoid making the response type nullable (`Record<string, any> | null`), what would break the property xml tag access (`response[propertyXmlTagName ?? name]`). Please, let me know if it is not aligned with the codebase patterns.

image
image
image

## 🚨Still not working for nested objects with type definitions in the `additionalProperties` field🚨
image
image

Any ideas on how to solve it?

### DIFF
--- a/.changeset/plenty-beds-marry.md
+++ b/.changeset/plenty-beds-marry.md
@@ -1,6 +1,0 @@
----
-"@scalar/api-reference": patch
-"@scalar/api-client": patch
----
-
-fix: allow application/problem+json return type
--- a/.changeset/spotty-colts-jam.md
+++ b/.changeset/spotty-colts-jam.md
@@ -1,0 +1,5 @@
+---
+"@scalar/api-reference": patch
+---
+
+fix: show key-value pair examples for objects
--- a/README.md
+++ b/README.md
@@ -194,7 +194,9 @@ await fastify.register(require('@scalar/fastify-api-reference'), {
 })
 ```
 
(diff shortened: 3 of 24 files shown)


marclave:
amazing contribution! excited to try this out :) Will have a review EOD

> How to generate a changeset?

```
pnpm changeset
```

then it should be an interactive step by step process on whats being changed ✨

hanspagel:
Love it, great contribution! 🤝

hanspagel (approved):

PR #1056, opened 2024-02-26
fix: request type polish

marclave:


cameronrohani:
I vertically align request type with first line of text rather than centre

Before:
image

After:
image

As well as increase font size

cameronrohani (approved):
a l i g n e d

PR #1057, opened 2024-02-27
feat: api client in react

amritk:
Added a react client-only wrapper for the api-client.

@cameronrohani I just copied over the css file you had in the other PR, needs some 🪄 

Still needs:

- [ ] style (@cameronrohani)
- [x] documentation
- [x] typescript


packages/api-client-react/README.md
```diff
@@ -0,0 +1,60 @@
+# Scalar API Client
+
```
amritk, line 2: will whip this up

hanspagel:
Do you need help with the documentation? If so, just ping me. ✌️

packages/api-client-react/README.md
amritk: @hanspagel hows this for docs?

packages/api-reference/src/hooks/useSpec.test.ts
```diff
@@ -101,6 +101,7 @@ describe('useSpec', () => {
 
     await new Promise((resolve) => {
       watch(rawSpecRef, (value) => {
+        if (!value) return
```
amritk, line 104: @hanspagel there was no value in here which was causing an error. Is this value allowed to be falsey?

PR #1065, opened 2024-02-27
chore: remove coinmarketcap example

hanspagel:
Currently, we’re including two examples: Petstore + Coinmarketcap 

This is great for users playing around with the editor, but completely pointless for all the users just seeing a read-only spec.

I think we should remove the Coinmarketcap example. 👀 Saves 30kb only, though.

**Before**
Screenshot 2024-02-27 at 19 40 50

**After**
Screenshot 2024-02-27 at 19 41 22


marclave:
@hanspagel im happy to merge this, I think we should really be doing URL based imports to save on bundle size for openapi examples :) 

so this can be the step in that direction ✨

marclave (approved):

PR #1066, opened 2024-02-27
fix: add back color schemes

cameronrohani:
If your system differs from the  .dark-mode and .light-mode colours the scrollbar colours get messed up.

This fixes that.

+ fix fastify theme issue


marclave (approved):
we are so back

PR #1067, opened 2024-02-27
fix: max width size and client width breakpoint & sizes

cameronrohani:
change max width to 1540px for refs

change breakpoints + sizes of default theme client library icons

After:
image

Before:
image


marclave (approved):
awesome! i'll add a changeset soon ✨

PR #1068, opened 2024-02-27
fix: fixsidebar theme precedence and remove unused OSS variables

cameronrohani:
Fixed sidebar not having `--default` prefix 

Removed unused header css

### DIFF
--- a/packages/themes/src/presets/bluePlanet.css
+++ b/packages/themes/src/presets/bluePlanet.css
@@ -33,24 +33,24 @@
 /* Document Sidebar */
 .light-mode .t-doc__sidebar,
 .dark-mode .t-doc__sidebar {
-  --sidebar-background-1: transparent;
-  --sidebar-color-1: var(--default-theme-color-1);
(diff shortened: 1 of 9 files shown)


packages/themes/src/presets/bluePlanet.css
```diff
@@ -144,7 +137,7 @@
   color: var(--default-theme-background-1) !important;
 }
 .active_page .sidebar-heading-type {
-  color: var(--sidebar-color-1) !important;
+  color: var(--default-sidebar-color-1) !important;
```
hanspagel, line 140: We need to use the unprefixed var as the fallback, don’t we?

```css
 color: var(--default-sidebar-color-1, var(--sidebar-color-1)) !important;
```

cameronrohani: Good catch, this is leftover theme code from closed source repo so I can actually just delete that :)

marclave (approved):

PR #1069, opened 2024-02-27
fix(#1061): api client response body syntax highlighting

### LINKED ISSUE #1061: Responses aren't being highlighted
### What happens?

Responses aren't highlighted with their content-type.

image

To highlight as a request body and request command.

image

https://github.com/kitajs/minimal-example (Uses fastify with default fastify theme). 

Kita registers scalar automatically with the following config:

```ts
if (options.fastifyScalarUi !== false) {
  await fastify.register(
    require('@scalar/fastify-api-reference'),
    Object.assign({ routePrefix: '/reference' }, options.fastifyScalarUi || {})
  );
}
```

### Swagger/OpenAPI Example

https://kitajs-minimal-example.up.railway.app/reference#tag/hello/get

amritk:
fixes #1061 

1. Moved the `CodeBlock` component into the component library as `ScalarCodeBlock`
2. Switch to new component in references
3. Switch to new component in api-client


marclave (approved):
awesome work @amritk ✨

PR #1072, opened 2024-02-28
chore: allow hono@v4 as peerDependency

soulchild:
**Problem**
Only hono@v3 is specified in the peerDependencies of @scalar/hono-api-reference

**Solution**
With this PR, hono@v4 is allowed as well.

### DIFF
--- a/.changeset/lemon-waves-enjoy.md
+++ b/.changeset/lemon-waves-enjoy.md
@@ -1,0 +1,5 @@
+---
+"@scalar/hono-api-reference": patch
+---
+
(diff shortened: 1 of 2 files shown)


soulchild:
I've successfully used @scalar/hono-api-reference with hono@v4, so I'm suggesting including it in the allowed peerDependencies.

marclave:
awesome work @soulchild ! 

are you able to add a changeset :) can do

```
pnpm changeset
```

marclave (approved):
✨

PR #1073, opened 2024-02-28
fix: select all does not work

hanspagel:
Select all doesn’t work in CodeMirror. We’ve had a fix, I removed it in https://github.com/scalar/scalar/commit/3d712d7464894a07ee91f62416ec1ef3251a60ff and this PR adds it again. 🤡


marclave (approved):
we are so back!

PR #1074, opened 2024-02-28
feat: hide secret credentials, fix #1062

### LINKED ISSUE #1062: Authentication keys should be hidden to avoid token leaks
### What happens?

Authentication is entirely visible after writing it, making it impossible to screenshare/stream the scalar UI without leaking secrets.

image

image

Authentication should also be kept on local storage, so it can persist on page reloads and not force the user to paste it _(showing up to everyone looking at the screen)_ every time.

To be hidden under a mask, unless the user clicks on it to edit it.

example:

image

image

just use the authentication feature

### Swagger/OpenAPI Example

https://docs.scalar.com/swagger-editor

hanspagel:
Currently, we’re showing all tokens/passwords/keys in plain text. See #1062.

With this PR:
* the relevant inputs become password fields
* the secret credentials are hidden in the code blocks (but still there if you copy the content)

The current approach has one downside. If someone passes "curl" or just a single letter as a token, all those are hidden in the example. The passed strings are considered secret and whenever they appear in the code block they are hidden.

But they are still visible in the API client. We could hide the `Authorization` header value always, but not everyone uses the `Authorization` header. They are plenty of alternatives (cookies, URL, other header names …).

So we could pass the credentials and just match them, but what if you change the credentials in the API client? Are they shown then?

In an ideal world the API clients would know about the OpenAPI specification and could hide the exact fields, but we’re not there yet.

Screenshot 2024-02-28 at 13 22 58

### DIFF
--- a/.changeset/fuzzy-dolphins-try.md
+++ b/.changeset/fuzzy-dolphins-try.md
@@ -1,0 +1,7 @@
+---
+"@scalar/api-reference": patch
+"@scalar/api-client": patch
+"@scalar/components": patch
+---
+
+feat: hide secret credentials
--- a/packages/api-client/src/components/Grid/Grid.vue
+++ b/packages/api-client/src/components/Grid/Grid.vue
@@ -2,7 +2,7 @@
 import { ScalarIcon } from '@scalar/components'
 import { ref } from 'vue'
 
-const props = defineProps<{
+defineProps<{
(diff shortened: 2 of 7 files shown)


marclave (approved):
im down for this! we can re-think how we do copy pasting from snippets and other things in followup work ✨

PR #1079, opened 2024-02-28
feat: api-client-react wrapper

amritk:
Added a react client-only wrapper for the api-client. It should be good to go!

to test:
- run `pnpm i && pnpm build && make up`
- hit http://localhost:5058/client


hanspagel:
This is 🔥 

Any idea why the bundle size increases? Because we bundle the CSS in the components now? 🤔

amritk:
> This is 🔥
> 
> Any idea why the bundle size increases? Because we bundle the CSS in the components now? 🤔

It says codemirror is the reason the bundle size increases, so I'm not sure if its not de-duping? I did have to delete the pnpm-lock so that might have something to do with it

image

amritk:
Oh I see the standalone JS as well. I can test it out, but yea that would make sense. So there's really no change in the bundle size it just went from css -> js

hanspagel (approved):
A PP RO VED

PR #1080, opened 2024-02-28
feat: Add codemirror syntax highlighting back in and move to a static color

geoffgscott:
For some reason the ::selection does not seem to accept our .dark-mode, .light-mode variables. For now we will hard code a color. 

This also adds syntax highlighting back in.

### DIFF
--- a/packages/use-codemirror/src/hooks/useCodeMirror.ts
+++ b/packages/use-codemirror/src/hooks/useCodeMirror.ts
@@ -1,11 +1,25 @@
+import {
+  autocompletion,
+  closeBrackets,
+  closeBracketsKeymap,
+  completionKeymap,
+} from '@codemirror/autocomplete'
 import { html } from '@codemirror/lang-html'
 import { json } from '@codemirror/lang-json'
-import { type LanguageSupport, StreamLanguage } from '@codemirror/language'
-import * as yamlMode from '@codemirror/legacy-modes/mode/yaml'
+import { yaml } from '@codemirror/lang-yaml'
+import {
+  type LanguageSupport,
+  type StreamLanguage,
+  bracketMatching,
+  defaultHighlightStyle,
+  indentOnInput,
+  syntaxHighlighting,
+} from '@codemirror/language'
 import { type Extension, StateEffect } from '@codemirror/state'
 import {
   EditorView,
   type KeyBinding,
+  highlightSpecialChars,
   keymap,
   lineNumbers as lineNumbersExtension,
 } from '@codemirror/view'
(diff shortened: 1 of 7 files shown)


amritk:
should we just move to prism?

marclave:
> should we just move to prism?

im down for this! I still say let's merge this since it fixes a bunch of bugs and we can do follow up work to remove the couple of lang  highlighting we have : )

packages/use-codemirror/src/hooks/useCodeMirror.ts
```diff
@@ -283,11 +298,25 @@ function getCodeMirrorExtensions({
...
+    extensions.push(
+      indentOnInput(),
+      bracketMatching(),
+      autocompletion(),
+      closeBrackets(),
+      keymap.of([
+        ...completionKeymap,
+        ...closeBracketsKeymap,
+        selectAllKeyBinding,
+      ]),
```
hanspagel, line 313: What are these doing? I can’t see any indent, bracket, autocomplete … magic when I play around with the editor. 🤔

hanspagel (approved):
Awesome PR! I’ve added some minor details that came to mind when reviewing.

Anyway, I’m too stupid to see what the added CodeMirror plugins do. If you could enlighten me, that would be great.

But let’s get this in!

hanspagel:
Ah, and I just saw @marclave had another fix for the selection color here: https://github.com/scalar/scalar/blob/3108786309e4cd6daaa7854a7cc1a5819fa702ed/packages/use-codemirror/src/hooks/useCodeMirror.ts#L260-L263

Should we add this to the PR here instead of hard coding the color? 🤔

packages/use-codemirror/src/hooks/useCodeMirror.ts
marclave: maybe do a fresh build?

marclave:
> Ah, and I just saw @marclave had another fix for the selection color here:
> 
> https://github.com/scalar/scalar/blob/3108786309e4cd6daaa7854a7cc1a5819fa702ed/packages/use-codemirror/src/hooks/useCodeMirror.ts#L260-L263
> 
> Should we add this to the PR here instead of hard coding the color? 🤔

woah i had no idea this was in main! we noticed css variables dont work in selection selectors

geoffgscott:
> Ah, and I just saw @marclave had another fix for the selection color here:
> 
> https://github.com/scalar/scalar/blob/3108786309e4cd6daaa7854a7cc1a5819fa702ed/packages/use-codemirror/src/hooks/useCodeMirror.ts#L260-L263
> 
> Should we add this to the PR here instead of hard coding the color? 🤔

This approach does not work. I think just stale code in Marcs branch

marclave (approved):

PR #1081, opened 2024-02-28
fix: scope scrollbar styles

hwkr:
This is a **BREAKING CHANGE** for `@scalar/themes` and `@scalar/swagger-editor`. I didn't bump the major version because neither are v1 yet but maybe we should?

Breaking interface changes:
- `@scalar/swagger-editor`: the export of `ResetStyles` is removed
- `@scalar/themes`: the export of `scrollbar.css` is removed

Other than that this scopes the scrollbar styles for the OSS the same way we scope the css reset. I thought about combining them into a single component but I couldn't think of a good name 🤷‍♀️.

## Before

Google Chrome-2024-02-28-15-02-23@2x

## After

Google Chrome-2024-02-28-14-57-39@2x


amritk:
Hmm why use a component instead of a css class?

marclave:
@hwkr what are the breaking changes : ) about to dive into the code

hwkr:
> Breaking interface changes:
> 
> * `@scalar/swagger-editor`: the export of `ResetStyles` is removed
> * `@scalar/themes`: the export of `scrollbar.css` is removed

image

hwkr:
> Hmm why use a component instead of a css class? (for scrollbar)

Because we have to scope it, before it was just a global class

packages/api-reference/src/components/ApiReferenceBase.vue
```diff
@@ -161,49 +165,51 @@ if (props.configuration?.authentication) {
...
-          <template #editor-input="editorInputProps">
-            <slot
-              name="editor-input"
-              v-bind="editorInputProps" />
-          </template>
-        </LazyLoadedSwaggerEditor>
-      </template>
-    </ApiReferenceLayout>
+  <ResetStyles v-slot="{ styles: reset }">
+    <ScrollbarStyles v-slot="{ styles: scrollbars }">
```
geoffgscott, line 169: This approach is not really making sense to me. We have an existing `scrollbar.css` in the closed source repo. `ApiReference` base is what we consume there. Should be not be applying these styles in the `ApiReference` component?

marclave:
> @marclave
> 
> image

right i just was curious on where we use these changes if they're in org or if we have some public interface for them being broken :)

hwkr:
> > @marclave
> right i just was curious on where we use these changes if they're in org or if we have some public interface for them being broken :)

Oh! The packages are published publicly so it would be a breaking change for any consumers using those export but I don't know of any projects using them externally.

packages/api-reference/src/components/ApiReferenceBase.vue
hwkr: Followup question is should ThemeStyles be scoped the way we're scoping the reset

amritk (approved):
I will begrudgingly accept this PR even though I believe it should be scoped under a css class and not a component. 

Long story short we tried it and need to add postcss everywhere if we want this to work with HMR and not have crazy css class bloat.

hwkr:
> I will begrudgingly accept this PR even though I believe it should be scoped under a css class and not a component.
> 
> Long story short we tried it and need to add postcss everywhere if we want this to work with HMR and not have crazy css class bloat.

To add, we can all chat about a more cohesive postcss / css / tailwind strategy for this as well as #1103

PR #1082, opened 2024-02-28
style: set scroll and border on response body

amritk:
Noticed a little issue with the api-client since switching to prismjs. This PR sets a max height + little border so it matches the look of codemirror.
Before:
image
After:
image


marclave (approved):
nice work 🖖

PR #1085, opened 2024-02-29
test: api reference cdn

tmastrom:
**Problem**
Currently, we are not validating the api reference package delivered from the CDN. 

**Explanation**
The CDN doesn’t deliver 1:1 what we build so we should continuously validate this. 

**Solution**
With this PR:
* create a small fastify server that serves a  static HTML page with the api reference imported from the CDN. 
* test api reference in the fastify server with Playwright
* run the Playwright test hourly(?) using a GitHub action workflow


marclave (approved):
wow i love this, great work @tmastrom ✨

PR #1088, opened 2024-02-29
chore: remove alias from components package

hanspagel:
Currently, we import the source files for all @scalar dependencies. I mean that’s why we’re doing the monorepo thing, to have all in the same place and have a short dev cycle.

However, we made an exemption for @scalar/components and only included the build. If you want to see changes to @scalar/components in @scalar/api-reference you have to make a build.

And the only reason we did this, was the aliases. We once removed all aliases from all packages to avoid the trouble. With the current setup, if you import a source file from another package, the alias can’t be resolved. Because the Vite config isn’t read then.

This PR removes the alias from the components package and removes the exemption. Means, we have cross-package HMR for the components now. :)

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -7,7 +7,7 @@
   "singleQuote": true,
   "singleAttributePerLine": true,
   "bracketSameLine": true,
-  "importOrder": ["^@/(.*)$", "^[./]"],
+  "importOrder": ["^[./]"],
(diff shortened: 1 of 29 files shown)


amritk (approved):
Sweet @hanspagel, yea I don't even remember why I did that 😄

marclave (approved):
UNREAL

PR #1095, opened 2024-03-01
ci: take screenshots of URLs

hanspagel:
WIP

Goal: Take screenshots in CI

It’s too tough to automatically add the screenshots to the PR as a comment: https://github.com/actions/upload-artifact/issues/50

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -70,4 +70,9 @@ jobs:
         uses: swinton/screenshot-website@v1.x
         with:
           source: ${{ matrix.url }}
-          destination: screenshot-${{ matrix.url }}.png
+          destination: screenshot.png
+      - name: Store screenshots as artifacts
+        uses: actions/upload-artifact@v4
+        with:
+          name: screenshots
+          path: screenshot.png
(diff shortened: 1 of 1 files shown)


amritk:
Perfect!! We definitely need this. 

There's a few tools we can check out. I used chromatic in the past but its pretty expensive. We can find something cheaper, maybe browser stack or something?

hanspagel:
I’m not sure whether visual regression testing (only) is the way, though. I’d still want a screenshot of the regression then. 😬

I don’t want to waste time clicking like 23 times until I get to see the screenshot that was taken, or run the tests locally (which probably takes like 5 minutes until I see the results), just to find out that my change (new feature, bugfix, CSS …) changed … the visual representation. Surprise. I’m pretty sure I’d start to ignore that sooner or later then.

I just want to make sure we don’t just rely on my/our discipline. It should be impossible to ignore when sth. is off. 🤡

amritk:
Yea I don't think you would need to click through 23 screenshots. Ideally this is how it would work:

1. takes a few screenshots of pages and compares with previous screenshots
2. if there is a discrepancy, asks if it is a wanted change 
  2a. if it is, then tests pass
  2b. if it isn't then test fail

PR #1096, opened 2024-03-01
chore: improve the contributor experience

hanspagel:
I’ve received great feedback from @Amorim33 for the contributing experience. This PR tries to address a few of the points.

With this PR:

* The CONTRIBUTING guide mentions the initial build (which is required for some things to work).
* The CONTRIBUTING talks about the tests.
* I’ve changed all used ports, to make sure they are somehow in an order that makes sense.
* The start page has links to all the integration examples now.
* Unused examples are removed.
* The NestJS express example is fixed.
* The NestJS examples are added to the startup script.
* All the integration examples serve the reference on / now, instead of /reference (easier to find).


Amorim33 (approved):
Awesome changes❤️🙏

marclave (approved):

marclave:
Unreal! Great work to both @hanspagel @Amorim33 !!

PR #1098, opened 2024-03-01
fix: mobile header button fix

cameronrohani:
Fix mobile headers icons


marclave (approved):
nice work @cameronrohani ! speedy :)

PR #1100, opened 2024-03-01
fix: revert css injection

amritk:
**Problem**
Currently, …

**Explanation**
This happens because …

**Solution**
With this PR …


packages/swagger-editor/src/components/SwaggerEditor/SwaggerEditor.vue
```diff
@@ -1,4 +1,5 @@
 <script lang="ts" setup>
+import '@scalar/components/style.css'
```
hwkr, line 2: The important line

hwkr (approved):

PR #1103, opened 2024-03-02
fix: revert previously reverted code for testing

amritk:
I put back in some of the code we quickly reverted earlier.

@hanspagel @hwkr this seems to work now, but we still need to build components in development. I believe its because with the vite alias, vite doesn't see the postcss config which is in the components folder. If we wanted to have components work in development without building we can[ link the postcss config into all of the vite configs](https://vitejs.dev/config/shared-options#css-postcss) that consume components. Though not sure if that is ideal, what do you think?

We won't merge this without some extensive regression testing!


hanspagel:
Ahhhh! This makes sense. Great find, let’s properly fix in the upcoming week.

amritk:
Lets do it incrementally, we can merge this then decide what we want to do with the postcss config. We can, and probably should add postcss to all of the packages!

hwkr (approved):
Gave the app a look over and everything looks good to me.

Confirmed new styles are being applied and properly scoped 👍 

Google Chrome-2024-03-04-16-28-39@2x

marclave (dismissed):
we lost line numbers in codemirror
image


still reviewing :)

marclave:
inside of the endpoints list the spacing is different between endpoints
image

image

marclave:
`Body application/json`

is misaligned


image

image

marclave:
in models theres extra spacing between text lines

image

image

hwkr (dismissed):
Okay so looked into these:

CC @marclave 

## Line Numbers

This appears to be an issue with the prisma css not getting injected an dsome of our own css not being applied, maybe a scoping issue?

Google Chrome-2024-03-05-13-30-57@2x

## Endpoint List

I tested this and the spacing seems to be identical.

Google Chrome-2024-03-05-13-28-13

## Typography Alignment Stuff

This appears to be a subpixel rendering issue but I don't think it was introduced my this PR. I was able to replicate similar behaviour switching between two tabs of the preview site.

Google Chrome-2024-03-05-13-41-10

Google Chrome-2024-03-05-13-43-33



@amritk let me know if you want any help debugging the line numbers.

amritk:
nah I should be able to fix it, actually I think I might just need to remove a "fix" I had to apply before. And for the sub pixel thing @marclave 🦅 👁️ 

I will jump back on this PR after the onboarding one

amritk:
Alright I fixed the line numbers, @marclave I think the rest of the discrepancies are due to this change:

https://github.com/scalar/scalar/commit/99bf014d9e6ce72170a169602e3b5afd6cf4970d#diff-172aff7cc13474e64d5be4ce4b03319b281a63f62deada67abeaaeff5863ed18R207

The wider width of the page

cameronrohani (approved):
Looking good 💯

PR #1108, opened 2024-03-03
feat: add proxy as prop to react client

marclave:


packages/api-client-react/src/ApiClientReact.tsx
```diff
@@ -11,14 +11,21 @@ type Props = {
   isOpen: boolean
   // The OpenApi request object
   request: ClientRequestConfig
+
```
amritk, line 14: Can you add a comment description for typescript plz

amritk (approved):
Let's gooo

PR #1113, opened 2024-03-04
chore: remove preparsed content configuration

hanspagel:
Once upon a time, I thought we could provide a variant of @scalar/api-reference without a parser. I thought we could just parse the specification once and pass it, to remove the parser from the package and decrease the bundle size.

Unfortunately, that’s not as easy as a I thought. There’s no way to just JSON.stringify the result of the parser, because circular references can’t be stringified.

I mean, we could kind of come up with our own format and add a light-weight parser to restore from that custom format, but that’s out of scope for now.

I don’t think anyone used the preparsed content, there was no benefit in doing it. So I’ve just removed it (without deprecating it).

### DIFF
--- a/.changeset/friendly-goats-begin.md
+++ b/.changeset/friendly-goats-begin.md
@@ -1,0 +1,5 @@
+---
+"@scalar/api-reference": patch
+---
+
+chore: remove preparsed content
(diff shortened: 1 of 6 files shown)


marclave (approved):
good call! we dont even use it :)

PR #1114, opened 2024-03-04
fix: content not reactive

hanspagel:
I noticed that the content wasn’t restored from local storage in dev. Turns out `configuration.spec` wasn’t reactive anymore, changes to it didn’t have any effect.

I traveled through the Git history to find out when this was introduced and it seems it was introduced with #1079, though I don't see any related change. Maybe a race condition?

Anyway, I think this PR fixes it. I’ve added a watcher for the spec configuration and added a helper method to `useSpec` to manually update the passed configuration.

I feel every watcher is introducing another problem, so I’d be open for an alternative solution that doesn’t involve a watcher. 👍

### DIFF
--- a/.changeset/strong-lobsters-juggle.md
+++ b/.changeset/strong-lobsters-juggle.md
@@ -1,0 +1,5 @@
+---
+"@scalar/api-reference": patch
+---
+
+fix: content not reactive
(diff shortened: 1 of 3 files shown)


marclave (approved):
nice!

geoffgscott:
> I feel every watcher is introducing another problem, so I’d be open for an alternative solution that doesn’t involve a watcher. 👍

Absolutely. The heavy dependence on watchers I have seen is both error prone and difficult to follow.

PR #1115, opened 2024-03-04
fix: custom operation security scheme list is ignored, fix #1112

### LINKED ISSUE #1112: A custom security scheme gets rendered as Authentication: Bearer
### What happens?

Scalar is not honoring the security scheme for custom apiKeys like this example:

```
"components": {
  "securitySchemes": {
    "bearerAuth": {
      "type": "http",
      "scheme": "bearer",
      "bearerFormat": "JWT"
    },
    "v1Auth": {
      "type": "apiKey",
      "name": "X-Auth-Token",
      "in": "header",
      "description": "JWT token"
    }
  },
  "schemas": {}
}
``` 

Where a route uses the custom scheme like this:

```
"security": [
  {
    "v1Auth": []
  }
]
```

image

Scalar should show the correct header name and type (apiKey)

Create a sample OAS definition with a custom security scheme with a apiKey and a route pointing to it

### Swagger/OpenAPI Example

_No response_

hanspagel:
We’re currently only configuring the security schemes on a global level, but OpenAPI allows to overwrite the list of available security schemes on a operation basis. See #1112 

With this PR the list is taken into account. And if the globally selected security scheme isn’t available for the operation, we’re just using the first available security scheme.

We should probably also add a way to selected the security scheme for every operation, but that’s a lot more work I guess. I’ll see this PR as a pragmatic solution though. :)

### DIFF
--- a/.changeset/rare-windows-join.md
+++ b/.changeset/rare-windows-join.md
@@ -1,0 +1,5 @@
+---
+"@scalar/api-reference": patch
+---
+
+fix: custom operation security scheme is ignored
--- a/packages/api-reference/src/helpers/getRequestFromAuthentication.test.ts
+++ b/packages/api-reference/src/helpers/getRequestFromAuthentication.test.ts
@@ -381,4 +381,43 @@ describe('getRequestFromAuthentication', () => {
       headers: [],
     })
   })
+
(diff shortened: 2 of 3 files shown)


marclave (approved):
😎

PR #1123, opened 2024-03-05
feat: API references composition

geoffgscott:
This PR aims to streamline the top level config state management and better separate the two different use cases for API References. 

**Use Case 1: Standalone API References**
- Requires an extensive config object
- Requires the mapping of the config object to internal state and handling two way data binding as needed. 
- Does not utilise an editor (we may consider doing a separate package that exports an editor equipped package)
- Most server packages, standalone references, and CDN deploys will use this

**Use Case 2: Embedded API Reference component**
- Will be embedded into a external web app that may have it own paradigms for state managment, flows for loading swagger files, and theming
- Should be provided _only_ a spec file, a parsed spec file, and few small visual configurations
- State should be managed by the wrapper app and not internally to the API Reference component
- Scalar Docs and other SPA apps should use this component. It allows a lower level of access to the internals of the API References for a cleaner integration.
- Currently we use `ApiReferenceLayout.vue` but we should likely be using the Layout wrappers @hwkr  created

**Changes Made:**
 - `ApiReferenceBase.vue` was dissolved. The configuration to internal state mappings we hoisted to the parent `ApiReference.vue` export
 -  Use dark mode was converted to a hook only used in the top level component. Toggles emit events that are either handled in the top level or by the wrapping SPA.
 - Download spec was streamlined to use an event bus to avoid drilling the raw references downward.

** Further Work **
- There should be a discussion around theme injection but ideally the theme injection would not happen in `ApiReferenceLayout.vue`. Any SPA integration should handle the theme injection in the parent app. 
- How multiple layouts are supported in embedded applications should be considered. It is possible that we should be using the exports of the layout folder and not ApiReferenceLayout as the primary embedded component


packages/api-reference/src/components/ApiReference.vue
```diff
@@ -1,36 +1,102 @@
...
+mapConfigToState('hiddenClients', setExcludedClients)
+
+const { parsedSpec, rawSpec } = useReactiveSpec({
+  proxy: toRef(() => props.configuration?.proxy || ''),
+  specConfig: toRef(() => props.configuration?.spec || {}),
+})
 </script>
 <template>
+  <!-- Inject any custom CSS directly into a style tag -->
+  <component
```
hwkr, line 84: I know this is how we used to do but could we just inject this into the `<head>`?

packages/api-reference/src/components/DarkModeToggle/DarkModeToggle.vue
```diff
@@ -1,18 +1,22 @@
 <script lang="ts" setup>
 import { ScalarIcon } from '@scalar/components'
 
-import { useDarkModeState } from '../../hooks/useDarkModeState'
+defineProps<{
+  isDarkMode: boolean
```
hwkr, line 5: I also learned theres a vue use package that we could use that just handles all the dark mode syncing stuff for you

https://vueuse.org/core/useDark/

packages/swagger-parser/src/helpers/fetch-spec.ts
geoffgscott, line 1: Move this to the parser as it better relates to parsing/loading than it does to editing. It's small anyways.

marclave (changes requested):
starting the review and caught a couple ui issues, look like the big icons are back @geoffgscott @amritk ?

image

we lost code syntax highlighting

image

and models.

doing a deeper dive shortly :) 

image

hwkr:
@marclave @geoffgscott the big icons will have been because the import of the component styles only existed in the swagger editor package and I'm guessing Geoff removed the that package (which is good, it probably shouldn't alway be loaded). Amrits PR will have fixed because now anytime the JS for the components library is loaded the CSS is injected into the page.

packages/api-reference/src/types.ts
```diff
@@ -77,18 +83,6 @@ export type SpecConfiguration = {
   content?: string | Record<string, any> | (() => Record<string, any>)
 }
 
-/** Default reference configuration */
-export const DEFAULT_CONFIG: DeepReadonly<ReferenceConfiguration> = {
```
marclave, line 81: was this not being used? :)

packages/api-reference/src/helpers/specDownloads.ts
```diff
@@ -0,0 +1,36 @@
+import { isJsonString } from '@scalar/api-client'
```
marclave, line 1: for the future we can add this in a different package ✨

packages/api-reference/src/components/DarkModeToggle/DarkModeToggle.vue
marclave: defs we can use this, only thing is it uses localStorage by default so would need something to handle the SSG part of our app ✨

examples/web/src/fixtures/emptySpec.ts
```diff
@@ -0,0 +1,28 @@
+import { type Spec } from "@scalar/api-reference";
+
+// Generate a new empty spec instance
+export const emptySpecGenerator = (): Spec => ({
```
marclave, line 4: niiiice ✨

packages/api-reference/src/components/ApiReference.vue
```diff
@@ -1,36 +1,102 @@
...
+  useSeoMeta(configuration.value.metaData)
 }
 
-function handleUpdateContent(value: string) {
-  content.value = value
-  props.configuration?.onSpecUpdate?.(value)
+// ---------------------------------------------------------------------------/
+// HANDLE MAPPING CONFIGURATION TO INTERNAL REFERENCE STATE
+
+/** Helper utility to map configuration props to the ApiReference internal state */
```
marclave, line 49: should this be moved out into a helper function outside of a vue file @geoffgscott ?

packages/api-reference/src/components/ApiReference.vue
amritk: yea that would be nice, this was just copy pastad but definitely open to a better way

packages/api-reference/src/helpers/specDownloads.ts
amritk, line 1: circular dependency!

packages/api-reference/src/types.ts
amritk: moved to the generator I think

amritk (approved):
LGTM

marclave (approved):
LETS GO ✨

marclave (changes requested):
fastify integration has this bug

image

marclave (approved):

PR #1126, opened 2024-03-07
fix: define color-scheme to improve request client selector readability

Amorim33:
fix  #1125 

## **Problem**

Currently, using the default theme and light mode, the request client selector options are not really readable, since the text color is dark.

image

## **Explanation**
This happens because the request example card always uses the `dark mode`. However, when the light mode is active, its default theme colors override the theme colors from dark mode.

## **Solution**
With this PR, when light mode is active, the `dark color scheme`  is considered for every component that references the dark-mode CSS class.

image

> 🚨 I am not totally sure if this is the best approach. However, I believe that it solves the problem without regressions

### DIFF
--- a/.changeset/unlucky-moles-develop.md
+++ b/.changeset/unlucky-moles-develop.md
@@ -1,0 +1,5 @@
+---
+"@scalar/themes": patch
+---
+
+fix: define color-scheme to improve request client selector readability
--- a/packages/themes/src/base.css
+++ b/packages/themes/src/base.css
@@ -76,6 +76,10 @@
   --default-theme-shadow-2: rgba(0, 0, 0, 0.08) 0px 13px 20px 0px,
     rgba(0, 0, 0, 0.08) 0px 3px 8px 0px, #eeeeed 0px 0 0 1px;
 }
+/* On some browsers, the light color scheme takes precedence when the light mode is active */
+.light-mode .dark-mode {
+  color-scheme: dark !important;
+}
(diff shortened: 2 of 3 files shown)


cameronrohani (changes requested):
Amazing thanks for doing this (this bug isn't on MacOS so wasn't sure if this would work) - I believe we can't do it in a theme file perhaps best to sick it in `base.css` with a comment as to why we do this.

Amorim33:
> Amazing thanks for doing this (this bug isn't on MacOS so wasn't sure if this would work) - I believe we can't do it in a theme file perhaps best to sick it in `base.css` with a comment as to why we do this.

Thanks for the feedback @cameronrohani !

I totally see your point, I initially tried to add it to the `base.css`, but it didn't work :/

A workaround could be adding a `!important` flag.

```css
/* base.css */

/* Light color scheme takes precedence when the light mode is active. */
.light-mode .dark-mode {
  color-scheme: dark !important;
}
```

From my understanding, I guess what happens is that when the light mode class is added to the `<body />` its color scheme takes precedence on the dark mode definition added to the selector component.

I ended up inserting this code in the default CSS, as I hadn't realized that the bug also occurs for other themes, sorry.
I don't know exactly why the theme's CSS precedes the base one.

marclave (changes requested):
awesome work @Amorim33 !! 

want to add a changeset then we can get this merged? :)

marclave (approved):

PR #1127, opened 2024-03-07
chore: update to typescript 5.4

hanspagel:
[TypeScript 5.4 is out](https://devblogs.microsoft.com/typescript/announcing-typescript-5-4/) … so let’s update, we were still on 5.2. :)

Not really necessary I guess, but it TS is failing for me every now and then (Reload VS Code FTW), so I just thought that might help.


marclave (commented):
:)

marclave (approved):
amazing :)

PR #1130, opened 2024-03-07
fix: move vue to devDepenencies in packages

amritk:
**Problem**
We have some double imports of vue going on

**Explanation**
We are packaging vue inside a package somewhere 🔍 

**Solution**
This PR removes one issue with vue packaging - in a package it should always be a devDependency


amritk:
Have this in the other PR

PR #1132, opened 2024-03-08
feat: add new client only api reference wrapper for react

amritk:
This is required for docusaurus


marclave (approved):
WOW great work @amritk

PR #1134, opened 2024-03-08
fix: add deprecated path to classic layout

marclave:
now we have deprecated strike through for classic layout :) 

image


cameronrohani (approved):
quality fix

PR #1135, opened 2024-03-08
fix: duplicate border top

marclave:
image

no more duplicate top bar


cameronrohani (approved):
brynn and I need to make this better when we tackle models but let's get this in for now

PR #1140, opened 2024-03-08
feat: Add OAS utils package and package bootstrapping

geoffgscott:
We use many of these utils in different places. A separate package encapsulates the logic and dependencies in 1 place. This also prevents imports of simple functions from large packages.

### DIFF
--- a/.changeset/new-rockets-stare.md
+++ b/.changeset/new-rockets-stare.md
@@ -1,0 +1,6 @@
+---
+"@scalar/build-tooling": patch
+"@scalar/oas-utils": patch
+---
+
+Add OAS utils package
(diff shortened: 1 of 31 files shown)


hanspagel:
Sorry, I’m on mobile and just skimmed through the changes just briefly, but some thoughts that popped up:

Why is that adding bun? Was that intentional? I think that would make contributing for windows users harder.

Most (or all?) OpenAPI related helper methods could live in the parser package.

Personally, I don’t like the name. We don’t use the OAS abbreviation anywhere else.

Thanks!

geoffgscott:
@hanspagel I can pull bun. I like it for native TS execution but vite-node will work
I also agree we can migrate some more helper methods here
@marclave approved the package name. I am open to alternatives but I also don't think it matters much. `spec-utils` seemed too vague

marclave (approved):
nice :)

amritk:
Once bundows is in we'll convert!!

PR #1151, opened 2024-03-09
test: playwright visual regression testing for cdn

tmastrom:
**Solution**
This PR adds visual regression testing to the api reference being served by the CDN. Playwright built-in tools are used to create screenshots and perform pixel diffing with `pixelmatch`. We can look at using a fully featured visual regression tool like Chromatic in the future if we find this is not sufficient.


tests/cdn.spec.ts-snapshots/cdn-snapshot-Mobile-Chrome-darwin.png
hwkr, line 1: Did this one fail to render completely or something?

tmastrom: This seems to be the size of the phone. I'm not sure why the screenshot is extended but this is the view I see when running `pnpm test:e2e:ui`

hwkr (approved):
Lgtm

amritk:
wait we have visual regression testing already?

hwkr:
> wait we have visual regression testing already?

Sort of but it's a bit difficult to test / update, also the themes one isn't turned on for CI because it's slow.

PR #1165, opened 2024-03-11
fix: quick fix for supporting application/vnd.api+json

tsiwek-piwik:
**Problem**
if your response is of type application/vnd.api+json we return No Body

image

**Explanation**
https://www.iana.org/assignments/media-types/application/vnd.api+json


marclave (approved):
Awesome work @tsiwek-piwik ! Let's add a changeset then we can get this merged :) 

pnpm changeset

In the root

PR #1171, opened 2024-03-13
fix: component cjs export

amritk:
The cjs exports had an incorrect umd.cjs extension in the components package.

image


marclave (approved):
✨

PR #1174, opened 2024-03-13
feat: add x-tagGroups in sidebar

marclave:


packages/api-reference/src/hooks/useSidebar.ts
```diff
@@ -130,6 +131,30 @@ const items = computed(() => {
...
+    ? parsedSpec.value?.['x-tagGroups']?.map((tagGroup) => {
+        const children: SidebarEntry[] = []
+        tagGroup.tags.map((tagName: string) => {
+          const tag = operationEntries?.find(
+            (entry) => entry.title === tagName.toUpperCase(),
+          )
+          if (tag) {
+            children.push(tag)
+          }
+        })
```
amritk, line 146: n^3 loops, I would re-write with one loop if possible

amritk (changes requested):
Good for the most part, I would re-write that triple loop into one is possible. For something the size of the stripe spec it will destroy the cpu.

Also whats the difference between a `tag` and an `x-tagGroup`?

packages/api-reference/src/hooks/useSidebar.ts
```diff
@@ -130,6 +131,30 @@ const items = computed(() => {
...
+        const sidebarTagGroup = {
+          id: tagGroup.name,
+          title: tagGroup.name.toUpperCase(),
+          children,
+          show: true,
+          isGroup: true,
+        }
+        return sidebarTagGroup
+      })
```
amritk, line 155: This business

marclave: @amritk imma be honest, i was cooking this at 1am and i had no idea what i was doing in es6 land 😭

amritk: Alright once I got the spec, this code isn't the worst. BUT I will throw up a ticket for perf improvements on `useSidebar` as a whole as we can definitely combine some loops

amritk (approved):

PR #1182, opened 2024-03-13
fix: add back sidebar item hover color

cameronrohani:
Sidebar item hover colour variable wasn't being applied


marclave (approved):
nice work!

PR #1183, opened 2024-03-13
fix: meta data is not added, fix #709

### LINKED ISSUE #709: page title can’t be changed
I think there’s no way to change the page title, though I assumed you can do it through the `metaData` configuration. 🤔

Would be nice to have a way to set a custom page title, too.

hanspagel:
WIP 
Currently, the meta data (including the page title) isn’t added to the mark up. I think, because the code ended up being in the wrong component.

This PR moves it to the ApiReferenceLayout component, which should be used in all cases.

### DIFF
--- a/.changeset/green-poets-smoke.md
+++ b/.changeset/green-poets-smoke.md
@@ -1,0 +1,5 @@
+---
+"@scalar/api-reference": patch
+---
+
+fix: meta data is not added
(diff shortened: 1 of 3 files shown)


hanspagel:
It’s just something different, closing this here.

joshamaju:
Hi, any progress on this?

PR #1194, opened 2024-03-14
feat: keyboard shortcuts to navigate sections

hanspagel:
In the last townhall we had this idea to add j and k to navigate through the sections in the reference quickly. j and k are used in [Vim](https://vim.rtorr.com/), but also in Google Mail (enable keyboard shortcuts in settings) and others. A lot of developers know them. It has its origin in the ADM-3A keyboard layout, that was lacking dedicated cursor keys:

image

This PR adds those shortcuts. :)

I’ve made sure the keyboard shortcuts are ignored in inputs, textareas and when used in CodeMirror.


packages/api-reference/src/components/Content/Content.vue
```diff
@@ -19,6 +23,8 @@ const props = defineProps<{
   layout?: 'default' | 'accordion'
 }>()
 
+useNavigationKeyboardShortcuts()
```
geoffgscott, line 26: This seems like a strange place to initialize. Should this not be done in either the sidebar or ApiReferenceLayout?

packages/api-reference/src/hooks/useSidebar.ts
```diff
@@ -256,9 +256,48 @@ export function useSidebar(options?: { parsedSpec: Spec }) {
...
+  /**
+   * Flattened list of all entries
+   */
+  const flattenedEntries = computed(() => {
+    return items.value.entries.reduce((acc, entry) => {
+      if (entry.children) {
+        return [...acc, entry, ...entry.children]
+      }
+
+      return [...acc, entry]
```
geoffgscott, line 268: Will improve efficiency considerably with large arrays
```suggestion
      acc.push(entry, ...entry.children || [])
      return acc
```

packages/api-reference/src/components/Content/Content.vue
hanspagel: I moved it around a few times and discussed it with @amritk and we eventually agreed to put it there, because it’s closest to what you’re navigating.

amritk: I don't mind layout either. Not really a right answer here

packages/api-reference/src/hooks/useSidebar.ts
hanspagel: I think this code could add some empty arrays to the flattened list, but I’ll add a variation of it. ✌️

marclave:
hey @hanspagel after some trial and testing out last night, I think we should re-visit hotkeys when we dive into the api client more.

Let's revisit later <3

PR #1197, opened 2024-03-14
chore: remove swagger editor

hanspagel:
I somehow missed the PR, but it seems the @scalar/swagger-editor was removed from @scalar/api-reference and we have a slot there instead.

This PR removes it as a dependency from @scalar/api-reference and also removes the package from the repository.


geoffgscott (approved):
Love it.

marclave (approved):

PR #1198, opened 2024-03-14
feat: add monaco editor to local testing environment

hanspagel:
I was heavily relying on the @scalar/swagger-editor to test our reference in the local setup and miss it already.

This PR adds Monaco Editor just for examples/web, so it’s not part of what we ship to npm, but helps (at least me) to play around with @scalar/api-reference interactively.

Cool thing, Monaco is way more powerful than CodeMirror and we got autocomplete for the OpenAPI specification there.

Screenshot 2024-03-14 at 16 53 03


examples/web/src/components/MonacoEditor.vue
```diff
@@ -0,0 +1,197 @@
...
+  }>(),
+  {
+    darkMode: false,
+  },
+)
+
+// const isDark = useDarkMode(
+const model = defineModel<string>()
+
+const monacoEditorRef = ref<HTMLElement>()
```
geoffgscott, line 21: This needs to be:
```suggestion
const monacoEditorRef = ref<HTMLElement | null >(null)
```

geoffgscott (approved):
Approved for testing. If we want to ship this at all we need to improve the reactivity model of the editor.

hanspagel:
I don’t think we can ever ship this as part of the npm package. users would need to properly set up web workers in their projects (when not using Vite), which is a headache for the users and for us.

PR #1200, opened 2024-03-15
chore: update axios

hanspagel:
Updates axios from 1.6.5 to 1.6.7, which includes an update of follow-redirects to 1.15.6, which patches a security issue. 🤖

Read more: https://hackerone.com/reports/2390009


geoffgscott (approved):
Looks good. For how light our usage is we could probably ditch axios for vanilla fetch

marclave (approved):

PR #1214, opened 2024-03-17
feat: add support of a default scheme if not present in parsedSpec

chauhan17nitin:
**Problem**
The request url getting generated on the basis of current window URL when no schemes present in swagger json

**Explanation**

**Solution**
Add a default HTTP scheme and generate request URL on the basis of host and basepath

### DIFF
--- a/packages/api-reference/src/components/Content/Content.vue
+++ b/packages/api-reference/src/components/Content/Content.vue
@@ -42,7 +42,18 @@ const localServers = computed(() => {
         }`,
       },
     ]
-  } else if (fallBackServer.value) {
+  } else if (
+    props.parsedSpec.host
+  ) {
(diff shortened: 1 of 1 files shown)


hanspagel (changes requested):
Great PR! I think Swagger 2.0 has a field for the scheme. I’ll check this tomorrow and add to the PR so we can get this in.

hanspagel:
Oh, we have kind of the same logic twice: In `Content.vue` and in `getUrlFromServerState.ts`. I think I’ll take the opportunity and try to merge them in a separate PR.

hanspagel:
Thanks for the PR! We fixed this in https://github.com/scalar/scalar/pull/1475 :)

Issue #1180, opened 2024-03-13
When the response contains a field with a array it doesn't render properly if the items are anyOf


thelinuxlich:
### What happens?

Scalar renders an empty array on the UI when the response has an array field with anyOf

![image](https://github.com/scalar/scalar/assets/5043/5b8a1781-5515-4805-9775-b1b3e513fafb)

### What did you expect to happen?

The `data` field rendering the schema properly

### How can we reproduce the issue?

Use this sample response schema:

```json
{
  "schema": {
    "type": "object",
    "properties": {
      "success": {
        "type": "boolean"
      },
      "data": {
        "type": "array",
        "items": {
          "anyOf": [
            {
              "type": "object",
              "properties": {
                "hash": {
                  "type": "string"
                }
              },
              "required": [
                "hash"
              ],
              "additionalProperties": false
            },
            {
              "type": "object",
              "properties": {
                "error": {
                  "type": "string"
                }
              },
              "required": [
                "error"
              ],
              "additionalProperties": false
            }
          ]
        }
      }
    },
    "required": [
      "success",
      "data"
    ],
    "additionalProperties": false
  }
}
```

hanspagel:
Another great find, thanks!

thelinuxlich:
So the behavior now is to display the first option?

PR #1215, opened 2024-03-18
fix: example for object with anyOf, allOf, oneOf schemas doesn’t have a value

lundibundi:
**Problem**
We generate examples from schemas in a lot of places and there was a case we didn’t cover: object schemas with top level anyOf, allOf, oneOf.

**Solution**
With this PR getExampleFromSchema can not handle the above properly.

See https://github.com/scalar/scalar/issues/1180.


hanspagel (approved):
Oh, what a great PR! Great addition, thank you. The change will be part of the next release.

PR #1216, opened 2024-03-18
fix: fastify api reference doesn’t load JS on proxied/prefixed routes

hex0id:
**Problem**
In case Fastify is deployed at a non-root path, the initial HTML will incorrectly link the main JavaScript, causing the application to break.

**Explanation**
The script's src is missing an optional web root variable. I've set up a proxy to Fastify at this URL: http://domain.example/something/. When I deploy Scalar Fastify in a production environment with the "docs" prefix, the main JavaScript fails to load from the correct location, leading to a blank page, even though the Fastify route to "docs" is correct.

HTML at http://domain.example/something/docs/ loads fine with a 200 OK response.
The JS file at http://domain.example/docs/@scalar/fastify-api-reference/js/browser.js is not found, resulting in a 404 error.

**Solution**
Add an optional webRoot variable in the configuration, it will be added to script src.

I'm not completely sold on the name webRoot for this variable. If you have any better suggestions, I'd love to hear them.

### DIFF
--- a/packages/fastify-api-reference/src/fastifyApiReference.test.ts
+++ b/packages/fastify-api-reference/src/fastifyApiReference.test.ts
@@ -72,6 +72,26 @@ describe('fastifyApiReference', () => {
     )
   })
 
+  it('prefixes the JS url', async () => {
+    const fastify = Fastify({
+      logger: false,
+    })
+
+    fastify.register(fastifyApiReference, {
+      routePrefix: '/reference',
+      webRoot: '/foobar',
+      configuration: {
+        spec: { url: '/swagger.json' },
+      },
+    })
+
+    const address = await fastify.listen({ port: 0 })
+    const response = await fetch(`${address}/reference`)
+    expect(await response.text()).toContain(
+      '/foobar/reference/@scalar/fastify-api-reference/js/browser.js',
+    )
+  })
+
   it('has the spec URL', async () => {
     const fastify = Fastify({
       logger: false,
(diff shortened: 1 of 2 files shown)


hanspagel (changes requested):
You’re so right! I think we could also make the JS URL relative. So basically this output:

```diff
-<script src="/@scalar/fastify-api-reference/js/browser.js"></script>
+<script src="@scalar/fastify-api-reference/js/browser.js"></script>
```

That could simplify the whole code. Any chance you could try if that would work in your setup?

hex0id:
> You’re so right! I think we could also make the JS URL relative. So basically this output:
> 
> ```diff
> -<script src="/@scalar/fastify-api-reference/js/browser.js"></script>
> +<script src="@scalar/fastify-api-reference/js/browser.js"></script>
> ```
> 
> That could simplify the whole code. Any chance you could try if that would work in your setup?

It would indeed simplify the code, but in source code getJavaScriptUrl function is used and it complicates everything a bit, since we need to provide a link for router and a link for browser. Also routePrefix is there.

hanspagel:
You’re so right! Let me review this PR here then. ✌️

hanspagel (approved):
Great work! I’ve just added a test, some comments and I renamed it to publicPath. 

“Web root” to me is the absolute path on disk. Public path has been used in Webpack and others for similar use cases.

If someone has a better name, lmk! Happy to rename it again, if we’ll fast-follow with it. 😅

PR #1221, opened 2024-03-18
chore: move some types and helpers into oas-utils

amritk:
We shouldn't have to import a giant package like references just for types, move these commonly used methods and types into the slimmer oas-utils package

Most of the changes here are just types, the only code I changed was in the [header normalization](https://github.com/scalar/scalar/pull/1221/files#diff-6747418ef671bb8914d9dad1ab79ff6eebc0ab4b5a9d5df97d96b9799648e542) to avoid some extra imports. Just need to test that bit.


hanspagel (approved):
Awesome! The headers are still properly normalized.

Let’s get this in.

PR #1226, opened 2024-03-19
feat: make API client url editable, fix #1189

### LINKED ISSUE #1189: Make the URL editable for friendly UX
### What happens?

The URL is not editable and requires to replace a variable through **VARIABLES** section. However, this is user friendly and possibly will make things worse. Please see the attached screen

image

The Address bar or the URL should be editable to make the changes

- Provide a URL with any OpenApi spec encapsulating a variable and try to edit or update it through Address bar.

The issue was already confirmed on the discord channel by the author

https://discord.com/channels/1135330207960678410/1180602650476154950/1217598466457604370

### Swagger/OpenAPI Example

_No response_

hanspagel:
Currently, the URL in the API client isn’t editable, but sometimes people like to just jump into the address bar and change the URL (#1189).

With this PR the URL becomes editable.

We used to pass `readOnly` to the API client, that was easily removed. The address bar did update the value in the state, which isn’t easily possible though. CodeMirror is prefilled with two values (URL + path). We can’t easily update them both from one value. (Sure, we could split the string, but in OpenAPI specifications URLs can include a path … just as an example).

Also, the request name was editable then, but didn’t have any effect, so I just disabled it for now.

### DIFF
--- a/packages/api-client/src/components/ApiClient/Request/Request.vue
+++ b/packages/api-client/src/components/ApiClient/Request/Request.vue
@@ -7,7 +7,13 @@ import RequestHeaders from './RequestHeaders.vue'
 import RequestQuery from './RequestQuery.vue'
 import RequestVariables from './RequestVariables.vue'
 
-const { activeRequest, readOnly } = useRequestStore()
+const { activeRequest } = useRequestStore()
+
+/**
+ * TODO: This is a workaround to make the address bar editable, but not the rest. If we’d really like to have an
+ * API client where everything can be edited, we need to invest more time.
+ */
+const readOnly = true
 </script>
 <template>
   <div class="scalar-api-client__main__left custom-scroll">
--- a/packages/api-reference/src/components/ApiClientModal.vue
+++ b/packages/api-reference/src/components/ApiClientModal.vue
@@ -68,7 +68,6 @@ export { useApiClientStore } from '@scalar/api-client'
         </template>
         <ApiClient
           :proxyUrl="proxyUrl"
-          readOnly
           theme="none"
           @escapeKeyPress="hideApiClient" />
       </div>
--- a/.changeset/heavy-hairs-sin.md
+++ b/.changeset/heavy-hairs-sin.md
@@ -1,0 +1,6 @@
+---
+"@scalar/api-reference": patch
+"@scalar/api-client": patch
+---
+
+feat: make api client url editable
--- a/packages/api-client/src/components/ApiClient/AddressBar.vue
+++ b/packages/api-client/src/components/ApiClient/AddressBar.vue
@@ -87,8 +87,6 @@ const lastRequestTimestamp = computed(() => {
     : 'History'
 })
 
-// TODO we need to not update the active request with these computed properties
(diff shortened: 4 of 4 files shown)


packages/api-client/src/components/ApiClient/Request/Request.vue
```diff
@@ -7,7 +7,13 @@ import RequestHeaders from './RequestHeaders.vue'
 import RequestQuery from './RequestQuery.vue'
 import RequestVariables from './RequestVariables.vue'
 
-const { activeRequest, readOnly } = useRequestStore()
+const { activeRequest } = useRequestStore()
```
marclave, line 10: we should keep `readOnly`

and have a config object at root, then we can do

```
        <ApiClient
          :proxyUrl="proxyUrl"
          :readOnly="readOnly"
          theme="none"
          @escapeKeyPress="hideApiClient" />
```

then inside of 
```
// Whether the request is in read mode or edit mode
const readOnly = ref(true)

export const useRequestStore = () => ({
  authState,
  readOnly,
  activeRequest,
  activeResponse,
  requestHistory,
  requestHistoryOrder,
  activeRequestId,
  setActiveResponse,
  resetActiveResponse,
  addRequestToHistory,
  setActiveRequest,
})

```

we can add a `setReadOnly` :)

hanspagel: Hm … I removed the `readOnly` just from the `Request.vue` file, so the operation name isn’t editable - because it doesn’t have an effect anyway.

I kept the `readOnly` in the store and we still use it to determine whether the address is editable or not.

Does that make sense?

With your proposed change the operation name would become an input again (changes wouldn’t have an effect).

packages/api-reference/src/components/ApiClientModal.vue
```diff
@@ -68,7 +68,6 @@ export { useApiClientStore } from '@scalar/api-client'
         </template>
         <ApiClient
           :proxyUrl="proxyUrl"
-          readOnly
```
hanspagel, line 71: When using the API Client inside the reference, we want it to be editable (at least the address).

packages/api-client/src/components/ApiClient/Request/Request.vue
```diff
@@ -7,7 +7,13 @@ import RequestHeaders from './RequestHeaders.vue'
...
+
+/**
+ * TODO: This is a workaround to make the address bar editable, but not the rest. If we’d really like to have an
+ * API client where everything can be edited, we need to invest more time.
+ */
+const readOnly = true
```
hanspagel, line 16: Currently, we don’t want the operation name to be editable in any case (because it’s not really tied up anyway).

amritk:
Is it possible to remove codemirror from the address bar? What is it being used for?

hanspagel:
Used to highlight variables! Autocomplete For variables would be nice in the future.

marclave (approved):
great work @hanspagel 

sorry it took me so long to get to this, and i'm eager to get this released!

oss/acc

PR #1232, opened 2024-03-19
style: improve text in markdown tables

hanspagel:
The Markdown tables needed some love!

Columns with few content are huge, though. But when we remove `table-layout: fixed;` small columns get crammed. @cameronrohani any CSS magic, that we can apply here?

**Before**

* every word breaks
* font size too high
* line height too small
Screenshot 2024-03-19 at 08 34 35

**After**

* scroll if content is too wide
* smaller font size
* bigger line height

Screenshot 2024-03-19 at 13 53 04

Screenshot 2024-03-19 at 13 50 30

### DIFF
--- a/packages/api-reference/src/components/MarkdownRenderer/MarkdownRenderer.vue
+++ b/packages/api-reference/src/components/MarkdownRenderer/MarkdownRenderer.vue
@@ -210,13 +210,12 @@ watch(
 }
 
 .markdown :deep(table) {
-  display: table;
+  display: block;
+  overflow-x: auto;
   position: relative;
   border-collapse: collapse;
-  table-layout: fixed;
   width: 100%;
   margin: 1em 0;
-  overflow: hidden;
   box-shadow: 0 0 0 1px
     var(--theme-border-color, var(--default-theme-border-color));
   border-radius: var(--theme-radius-lg, var(--default-theme-radius-lg));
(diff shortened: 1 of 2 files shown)


cameronrohani:
If you remove `table-layout: fixed`

and add `word-break: initial;` to  `.markdown td, .markdown th` 

it should work the way you want :) 

Let me know if that works @hanspagel :)

hanspagel:
@cameronrohani Unfortunately not! It doesn’t break long words then.

Screenshot 2024-03-19 at 11 42 29

cameronrohani:
Probably the best course of action is to add a scroll to the table on overflow then?

To do that you'd need to change `.markdown table` to `display: block` and add `overflow-x: auto`

Thoughts @hanspagel

hanspagel:
Looks good!
Screenshot 2024-03-19 at 13 50 30

marclave (approved):
Awesome work everyone :) so stoked for this fix!

packages/api-reference/src/components/MarkdownRenderer/MarkdownRenderer.vue
```diff
@@ -211,13 +210,12 @@ watch(
 }
 
 .markdown :deep(table) {
-  display: table;
+  display: block;
+  overflow-x: auto;
```
hwkr, line 214: Do we want the custom scroll classes here?

PR #1234, opened 2024-03-19
fix: cli does not watch files outside of cwd

hanspagel:
Currently, the CLI watches for changes in process.cwd. If the file is somewhere else, file changes aren’t detected.

With this PR we’re watching the directory where the OpenAPI specification is located instead.


marclave (commented):
🤯 nice fix!

marclave (approved):

PR #1235, opened 2024-03-19
fix: cli watcher exists when specification is invalid

hanspagel:
Currently, the file watch (e.g. `scalar reference openapi.json --watch`) exits when the specification is invalid.

With this PR it’ll show a warning, but keeps running. Easier for local tinkering. :)

Not perfect, need to add this to more commands, but better than what we had and I’m focused on sth. else rn.


marclave (approved):
more like watch this speed! great work @hanspagel 🌸 

loving the PRs from 🇨🇦

PR #1236, opened 2024-03-19
feat: add scalar example specification

hanspagel:
Currently, we’re using the Swagger Petstore as an example. Wouldn‘t it be cool, if we’d have our own example?

This PR adds a _Scalar Galaxy_ example specification. It’s not used anywhere yet, but it’s a start. I switched to YAML, it’s so much easier to write multiline Markdown in YAML.

We could add way more schema variations, but I want to wrap this up before the week ends. You’re all invited to add to it. 👐  You can quickly test it like this:

```bash
pnpm @scalar/cli reference ../../specifications/scalar-galaxy-3.1.yaml --watch
```


cameronrohani:
I would love if we had a handful of different markdown throughout 
i.e 

- tables
- lists (ordered / unordered)
- links
- summary / details
- linebreaks
- block quotes

specifications/scalar-universe-3.1.0.json
```diff
@@ -0,0 +1,651 @@
...
+            "required": true,
+            "example": 1
+          },
+          "name": {
+            "type": "string",
+            "required": true,
+            "example": "Mars"
+          },
+          "description": {
+            "type": ["string", "nullable"],
```
hwkr, line 585: I think we also support `"nullable": true", I'm not sure which way is the right way though

hanspagel: the array is the recommended modern syntax, but we should add the old one, which is still supported too 👍

marclave (approved):
awesome work @hanspagel 

let's fast follow by adding all our examples, we can put this on our CDN and use it EVERYWHERE ✨

PR #1238, opened 2024-03-19
fix: nextjs example port blocked

hanspagel:
I don’t know how this happened, but I used one port twice (proxy + nextjs). This PR moves the Nextjs example to another port. :)


marclave (approved):
we are now unblocked 🖖 nice work :)

PR #1240, opened 2024-03-19
feat: improve nullable types

hanspagel:
In the recent OpenAPI version, you can set `type` to an array (`['string', 'nullable]`). This PR adds support for it. And it also makes sure the old way (`nullable: true`) is rendered exactly the same way.

**Before**
Screenshot 2024-03-19 at 13 26 04

**After**
Screenshot 2024-03-19 at 13 30 18


marclave (approved):
Awesome work @hanspagel !

PR #1243, opened 2024-03-20
fix: use scroll margin instead of padding to align scroll headers

hwkr:
Updates our scroll alignment code to use [`scroll-margin`](https://developer.mozilla.org/en-US/docs/Web/CSS/scroll-margin) to line up the scroll rather than using a negative margin trick. I hadn't used it before when I first did this alignment fix but browser support seems good 🚀 

https://github.com/scalar/scalar/assets/6374090/b418a519-b4d8-4b41-979a-76a818a04f83

### DIFF
--- a/.changeset/lazy-ducks-sniff.md
+++ b/.changeset/lazy-ducks-sniff.md
@@ -1,0 +1,5 @@
+---
+"@scalar/api-reference": patch
+---
+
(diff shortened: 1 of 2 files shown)


marclave:
Unreal work @hwkr !!! Let's go :)

marclave (approved):

amritk:
sick! this was annoying me haha

PR #1244, opened 2024-03-20
fix: move theme flare behind all content

hwkr:
Google Chrome-2024-03-19-18-05-35@2x
Google Chrome-2024-03-19-18-05-19@2x


marclave (approved):
Nice fix @hwkr

PR #1252, opened 2024-03-20
fix: allow api client react to be editable

amritk:
Quick fix for react, but we still need the reactivity fix on the client


hanspagel (approved):
Great fix

PR #1254, opened 2024-03-20
fix: address bar changes do not effect request, fix #1189

### LINKED ISSUE #1189: Make the URL editable for friendly UX

hanspagel:
We made the address bar editable in #1226. The address is actually two values (url + path), so I decided to not update the request state. Which means the updated address isn’t used for the actual request.

This PR fixes that bug, and updates the request state.

We can’t easily update two values with just one value. So I decided to reset the path then and just put everything into the `url` property.

This will bite us if we ever want to persist the api client state and switch between envs (different base urls), but it’s a pragmatic workaround for now. :)

### DIFF
--- a/.changeset/sharp-toes-mate.md
+++ b/.changeset/sharp-toes-mate.md
@@ -1,0 +1,5 @@
+---
+"@scalar/api-client": patch
+---
+
+fix: address bar changes do not have an effect on the actual request
--- a/packages/api-client/src/components/ApiClient/AddressBar.vue
+++ b/packages/api-client/src/components/ApiClient/AddressBar.vue
@@ -96,8 +96,10 @@ const onChange = (value: string) => {
     return
   }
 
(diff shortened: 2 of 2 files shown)


marclave (approved):

marclave:
Quick turnaround! Nice fix Hans :)

amritk (approved):
S00per fast. Yea lets re-think this when we build the app 😄

PR #1255, opened 2024-03-20
fix: react example port blocks echo server

hanspagel:
Just found another port that was used twice (react example + echo server). 🙈 I swear, it’s the last time I change any port.


marclave (approved):
Haha port party!

PR #1258, opened 2024-03-20
feat: hide auth credentials/token in client

hanspagel:
This PR moves the responsibility for Authentication to the API client.

This means:
* The AuthenticationState is in the client now
* The Authentication components are in the client now
* The client exposes the state and the components
* The reference uses the components and writes to the client state
* The api clients shows generated parameters (headers, cookies, and queries).
* If you click on generated parameters, the auth section is scrolled into view.

Breaking changes
* How we open the API client is different now. We need to pass the global security configuration and the transformed operation to the client (to be able to translate the authentication information to the actual request parameters).

Same state, same components (composed differently), same logic:

Screenshot 2024-03-22 at 09 52 39
Screenshot 2024-03-22 at 09 52 07


marclave (approved):
@hanspagel this is UNREAL work ✨ 

long overdue and I really appreciate you going well above what we imagined and making it a great UX.


let's get this in! had a few things to talk about before we do tho
- [ ] resolve conflicts
- [ ] where is the breaking change specifically? should we do a minor bump instead of a patch then?


again, unreal work @hanspagel

amritk (changes requested):
Just saw that there's an emitter here, would the vue-use event bus not work?

amritk (approved):
Switched from mitt to an event bus. The rest looks good! moving things into the client is awesome since its lazy loaded. @marclave wanna test my bus change before merging?

PR #1259, opened 2024-03-20
feat: sidebar border indent colors + theme polish

cameronrohani:
This PR makes it easy to add sidebar indent border colors to a theme with out having to write custom code per theme:
image

In addition it refines the scalar-card to have more consistent font styles in order to prepare for upcoming UX improvements
image

Note we are now using the button colours for the 'test request' buttons which was much needed as the contrast was not working well as we continued to scale our themes.


packages/api-reference/src/components/Icon/icons/ChevronDown.svg
```diff
@@ -1,3 +1,3 @@
 <svg width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
-<path d="M4.5 8.25L12 15.75L19.5 8.25" stroke="currentColor" stroke-width="1.5" stroke-linecap="round" stroke-linejoin="round"/>
+<path d="M4.5 8.25L12 15.75L19.5 8.25" stroke="currentColor" stroke-width="1.5 stroke-linecap="round" stroke-linejoin="round"/>
```
hwkr, line 2: Looks like this is missing a quote and broke the SVG
```suggestion
<path d="M4.5 8.25L12 15.75L19.5 8.25" stroke="currentColor" stroke-width="1.5" stroke-linecap="round" stroke-linejoin="round"/>
```

packages/api-reference/src/components/Sidebar/SidebarElement.vue
```diff
@@ -271,24 +291,18 @@ const handleClick = async () => {
...
-    )
-  );
+}
+.active_page .sidebar-heading-type {
+  background: var(--method-color);
+  color: color-mix(in srgb, var(--method-color), white 85%);
+}
+.dark-mode .active_page .sidebar-heading-type {
+  background: var(--method-color);
+  color: color-mix(in srgb, var(--method-color), black 80%);
```
hwkr, line 305: Yo color-mix let's gooo

packages/themes/src/presets/alternate.css
```diff
@@ -105,7 +109,7 @@
   --default-theme-color-orange: var(--default-theme-color-1);
   --default-theme-color-purple: var(--default-theme-color-1);
 }
-.download-cta a,
+.references-rendered .download-cta .download-button,
 .references-rendered .markdown a {
   color: var(--default-theme-color-1) !important;
```
hwkr, line 114: Can we remove these importants with the added specificity?

packages/themes/src/presets/bluePlanet.css
```diff
@@ -77,7 +89,7 @@
 .show-api-client-button svg {
   color: black !important;
 }
-.download-cta,
+.references-rendered .download-cta .download-button,
 .references-rendered .markdown a {
   text-decoration: underline !important;
```
hwkr, line 94: Another important we could maybe remove

packages/themes/src/presets/alternate.css
cameronrohani: For some reason `markdown a` has an important tags on them. We need to figure out why that is and then we can remove these importants

packages/api-reference/src/components/Icon/icons/ChevronDown.svg
cameronrohani: Great catch, appreciate it!

packages/themes/src/presets/bluePlanet.css
cameronrohani: There's nothing more annoying than adding important for these, let's properly fix them in the coming days / weeks 😓

PR #1260, opened 2024-03-20
feat: URL support for the CLI

hanspagel:
Currently, the CLI works with files only. With this PR you can pass an URL to the mock and reference command. 🤯 

```bash
scalar reference https://raw.githubusercontent.com/outline/openapi/main/spec3.json
```

```bash
scalar mock https://raw.githubusercontent.com/outline/openapi/main/spec3.json
```

You can even pass the `--watch` flag to poll the URL every few seconds. 

To test it locally, use `pnpm @scalar/cli` instead of `scalar`.


marclave (approved):
great work @hanspagel !! we got url support now ✨ ✨

PR #1261, opened 2024-03-20
fix: clear response on close and set request on open

amritk:


marclave (approved):
✨ 

great work @amritk , we can eventually add a "keepRequestState" config option :)

PR #1262, opened 2024-03-20
fix: section flare if position fixed

cameronrohani:
With us adding container queries to replace the old javascript mobile queries we broke the rare case of people having position fixed for section flares. 

This fixes that :)

### DIFF
--- a/packages/api-reference/src/components/Content/Content.vue
+++ b/packages/api-reference/src/components/Content/Content.vue
@@ -66,9 +66,19 @@ const isLazy =
   !window.location.hash.startsWith('#model')
 </script>
 <template>
+  <!-- For adding gradients + animations to introduction of documents that :before / :after won't work for -->
+  <div class="section-flare">
+    <div class="section-flare-item"></div>
+    <div class="section-flare-item"></div>
+    <div class="section-flare-item"></div>
+    <div class="section-flare-item"></div>
+    <div class="section-flare-item"></div>
+    <div class="section-flare-item"></div>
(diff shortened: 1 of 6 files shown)


packages/api-reference/src/components/Content/Content.vue
```diff
@@ -198,4 +208,10 @@ const isLazy =
   ) {
   margin-top: 0;
 }
+.section-flare {
+  position: absolute;
+  top: 0;
+  right: 0;
+  pointer-events: none;
+}
```
hwkr, line 216: Do we not need the z-index anymore?

cameronrohani: nope because the content is inherently above it now

packages/themes/src/presets/kepler.css
```diff
@@ -167,5 +167,5 @@
     rgba(120, 119, 198, 0.3),
     transparent
   );
-  height: 100%;
+  height: 100vh;
```
geoffgscott, line 170: This is probably a breaking change. 100vh is not always the height of references.

cameronrohani: This just controls the size of a gradient, not a breaking change

marclave (approved):
great work @cameronrohani ! let's get this in for our customers :)

packages/api-reference/src/components/Content/Content.vue
hwkr: 🙏

PR #1269, opened 2024-03-21
feat: switched css injection plugin

amritk:
The current css injection plugin injects the style via JS so it doesn't work on SSR.

[This new plugin](https://www.npmjs.com/package/vite-plugin-lib-inject-css) just auto imports the css file in the component so it makes SSR friendly.
```ts
import './style.css';
```

Not sure how to test this locally because of our package aliases so I'm putting up the PR to test on the preview!

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
   "private": true,
   "scripts": {
     "build": "turbo build",
-    "build:packages": "pnpm --filter './packages/**' build",
+    "build:packages": "turbo --filter './packages/**' build",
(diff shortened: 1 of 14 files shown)


geoffgscott (commented):
Is this better than just requiring a second CSS import with the component?

amritk:
It's just quality of life so you don't need to do the import. So we avoid that issue we had a few weeks ago with the missed import

hanspagel:
I’ll test this locally tomorrow ✌️

marclave (approved):
nice work @amritk !

PR #1273, opened 2024-03-21
feat: change svg rendering

amritk:
moved to a more SSR friendly SVG rendering strategy


packages/components/src/components/ScalarIcon/icons/index.ts
```diff
@@ -1,16 +1,21 @@
...
 
-  if (icons[filename] === undefined) {
+  if (!icons[filename]) {
     console.warn(`Could not find icon: ${name}`)
-    return ''
+    return null
   }
 
-  return icons[filename]
+  return defineAsyncComponent(icons[filename]!)
```
hwkr, line 18: So this just transforms the raw SVG files into Vue components? 👀

amritk: yep I'm assuming thats what we wanted? We can keep em as svgs as well

hwkr: This seems cool, only thing we have to watch out for it wether or not the SVGs added to the folder are well formed (e.g. have correctly set viewboxes and use `currentColor` for the color)

amritk: I dumped them in from flowIcon, but only the ones that didn't already exist. So the previous ones should keep their currentColor. The new ones might not have it. Should I change em all?

packages/api-reference/src/components/Content/ClientLibraries/ClientSelector.vue
```diff
@@ -1,89 +1,50 @@
...
-          targetKey: 'python',
-          clientKey: 'python3',
-        },
-        {
-          targetKey: 'c',
-          clientKey: 'libcurl',
-        },
-      ],
-)
+const featuredClients = [
```
amritk, line 15: If we want to keep the smaller set on mobile, we can easily do so using css. Just left it out for now because it doesn't look half bad

packages/components/src/components/ScalarIcon/icons/index.ts
```diff
@@ -1,16 +1,21 @@
+import { defineAsyncComponent } from 'vue'
+
 import { type ICONS } from './icons'
 
-const icons = import.meta.glob('./*.svg', { as: 'raw', eager: true })
+const icons = import.meta.glob<SVGElement>('./*.svg')
```
hwkr, line 5: If this uses the SVG loader could we just do this?
```suggestion
const icons = import.meta.glob<SVGElement>('./*.svg?component')
```
https://www.npmjs.com/package/vite-svg-loader#component

hwkr:
I wonder if we could alternatively set [`defaultImport: 'component'`](https://www.npmjs.com/package/vite-svg-loader#viteconfigjs-1) then import the files directly. It would allow us to do tree shaking better 🤔

packages/components/src/components/ScalarIcon/icons/index.ts
amritk: Would they still be lazy loaded then?

packages/api-reference/src/components/Content/ClientLibraries/ClientSelector.vue
hwkr: I think this overcrowds it on the classic layout 
Google Chrome-2024-03-21-10-23-10@2x

packages/components/src/components/ScalarIcon/icons/index.ts
amritk: Or would that load all of them at once

amritk:
Yep we can definitely improve tree shaking. Especially with the other PR. I left that as out of scope for this pr though.

packages/api-reference/src/components/Content/ClientLibraries/ClientSelector.vue
amritk: definitely, I always forget about classic haha. Will add some css

geoffgscott (commented):
Overall looks good and matched the scalar-org approach. Can we drop XMLDOM from the packages now?

packages/api-reference/src/components/Content/ClientLibraries/ClientSelector.vue
amritk: Done!

hwkr (approved):
Lgtm!

hwkr:
Looks like this changed a few of the unit test and they need to be updated :)

marclave (approved):
unreal work @amritk , cant wait for `prequel-3`

PR #1277, opened 2024-03-21
fix: buffer not available in browser environments, fix #1272

### LINKED ISSUE #1272: Buffer is not defined
### What happens?

After entering credentials for basic auth the http client previews will no longer render

```
api-reference:19 ReferenceError: Buffer is not defined
    at vf (api-reference:26:206867)
    at yf (api-reference:26:206992)
    at api-reference:38:762435
    at r (api-reference:19:2864)
    at rr (api-reference:21:2908)
    at Proxy.<anonymous> (api-reference:38:283486)
    at mn (api-reference:19:3180)
    at le.l [as fn] (api-reference:21:19220)
    at le.run (api-reference:14:1475)
    at api-reference:21:20095
```

This should be following line in code:
https://github.com/scalar/scalar/blob/c4a26a7e9317436ca75c5c837aa40c873637efdf/packages/api-reference/src/helpers/getRequestFromAuthentication.ts#L167

Authentication should be shown correctly in the client code previews

1. Create a page as explained in the "From a CDN" example

2. Add a basic security scheme to the open api spec
```json
    "securitySchemes" : {
      "auth" : {
        "type" : "http",
        "scheme" : "basic"
      }
    }
```

3. Enter credentials in the authentication field

image

### Swagger/OpenAPI Example

_No response_

hanspagel:
We recently remove all Node polyfills. Unfortunately, we still need a polyfill for `Buffer`, which is not available in browser environments. :(

See #1272

### DIFF
--- a/.changeset/stale-carpets-drive.md
+++ b/.changeset/stale-carpets-drive.md
@@ -1,0 +1,6 @@
+---
+"@scalar/api-reference": patch
+"@scalar/api-client": patch
+---
+
+fix: buffer not available in browser environments
--- a/packages/api-client/package.json
+++ b/packages/api-client/package.json
@@ -49,7 +49,6 @@
     "@scalar/use-modal": "workspace:*",
     "@vueuse/core": "^10.4.1",
     "axios": "^1.6.7",
-    "buffer": "^6.0.3",
     "content-type": "^1.0.5",
     "nanoid": "^5.0.1",
     "pretty-bytes": "^6.1.1",
--- a/packages/api-client/package.json
+++ b/packages/api-client/package.json
@@ -49,6 +49,7 @@
     "@scalar/use-modal": "workspace:*",
     "@vueuse/core": "^10.4.1",
     "axios": "^1.6.7",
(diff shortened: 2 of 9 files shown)


amritk (changes requested):
@hanspagel instead of adding a polyfill try btoa instead of buffer in the browser
a little something like this:
```ts
if (typeof window === 'undefined') {
   Buffer.from(`${username}:${password}`).toString('base64') 
}
else {
  btoa(`${username}:${password}`);
}
```
btoa(`${username}:${password}`);

geoffgscott (changes requested):
@hanspagel We can use btoa() for this purpose in Browser environments and not require the polyfill.

amritk:
@geoffgscott did u just copy paste my comment? 😆

hanspagel:
OMG, totally forgot about btoa

hanspagel:
OMG, totally forgot about btoa

amritk (approved):

marclave (approved):
awesome work @hanspagel !

PR #1279, opened 2024-03-21
fix: rebuild schema property details to truncate long strings

hwkr:
- Refactors the property type details into components
- Adds the option to turn on truncation for any of the type values
- Fixes a typo in `$.nullable` so that actually works now
- ~~Moves rule badges and `writeOnly` / `readOnly` labels to be right aligned (I did this by accident but I think it looks better / cleaner, we can move them back to the left if we want)~~ Ignore it being on the right in the screenshots, it's back on the left
- Adjusts the color slightly of the type details prefixes to differentiate them from the detail value (e.g. `max: 100`)

**Note:** There might be slight pixel differences due to moving from an inline layout to a flex layout but we need the flex layout for the truncation.

[Cursed Petstore Test File](https://github.com/scalar/scalar/files/14709743/petstorev3.json) which I'm using in the below screenshots.

Sorry for big before / afters

## Before

Google Chrome-2024-03-21-14-16-31@2x

Google Chrome-2024-03-21-14-16-43@2x

Google Chrome-2024-03-21-14-17-17@2x

## After

Google Chrome-2024-03-21-14-13-56@2x

Google Chrome-2024-03-21-14-14-30@2x

Google Chrome-2024-03-21-14-15-16@2x

### DIFF
--- a/.changeset/heavy-lobsters-play.md
+++ b/.changeset/heavy-lobsters-play.md
@@ -1,0 +1,5 @@
+---
+"@scalar/api-reference": patch
+---
+
+fix: rebuild schema property details to truncate long strings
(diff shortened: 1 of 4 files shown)


hanspagel:
Looks so much better

marclave (approved):
awesome work @hwkr !!!

PR #1280, opened 2024-03-21
fix: use darkmode state hook to manage mode

### LINKED ISSUE #1248: Dark Mode Icon doesn't change
### What happens?

The dark mode icon and label used to change but I think it was broken in #1123

The dark mode icon and label (in both the classic and modern themes) should change to reflect the current dark / light mode.

Click on the dark mode button.

### Swagger/OpenAPI Example

_No response_

hwkr:
I think we can just use the hook now to manage everything since the root component syncs the dark mode config to the app state.

Fixes #1248 

Let me know if I'm missing something 🤔 

### Changing via the app state

https://github.com/scalar/scalar/assets/6374090/a45a3d35-bc6d-43ab-b306-4bbbd5f1c6f4

### Changing via the config object

https://github.com/scalar/scalar/assets/6374090/52765838-0aff-459e-be68-b155d5fdec27

### DIFF
--- a/.changeset/afraid-bugs-cry.md
+++ b/.changeset/afraid-bugs-cry.md
@@ -1,0 +1,5 @@
+---
+"@scalar/api-reference": patch
+---
+
+fix: use darkmode state hook to manage mode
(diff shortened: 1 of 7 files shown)


geoffgscott (changes requested):
@hwkr this hook has side effects that should be applied outside the top level component. This is why the `isDarkMode` value is provided as a prop.

hwkr:
> @hwkr this hook has side effects that should be applied outside the top level component. This is why the `isDarkMode` value is provided as a prop. 

No worries, this was just trying to fix the regression #1123. I'm not quite sure what your vision was for handling the data here so I'll just close this PR and you can do what your original plan was 😊

PR #1285, opened 2024-03-22
fix: operator will always return left

hanspagel:
The build process had this warning for a while:

> The "??" operator here will always return the left operand

This PR drops the right side. :)


marclave (approved):
ah nice fix! ❇️

PR #1286, opened 2024-03-22
fix: circular dependency

hanspagel:
The build process had this warning for a while:

> @scalar/cli:build: src/utils/index.ts -> src/utils/useGivenFileOrConfiguration.ts -> src/utils/index.ts

This PR resolves the circular dependency. :)


marclave (approved):
nice fix @hanspagel 🖖

PR #1289, opened 2024-03-22
fix: Add priority for dark mode state injection

geoffgscott:
The configuration property is used to set the initials state IF there is not an existing localstorage value. After initial setting the state is managed by the ApiReference component.


marclave (approved):
✨ ✨ ✨

PR #1290, opened 2024-03-22
fix: schema show on example responses

amritk:
Previously broke the schema show with the SSR PR. This fixes that issue and also adds a test to prevent it in the future


marclave (approved):
awesome work @amritk AND a test ✨

PR #1292, opened 2024-03-22
fix: does not show schema, fix #1291

### LINKED ISSUE #1291: does not show schema
Basic Auth works for me now. Not sure if this is related to the changes, but schemas aren't showing for me after updating

image

_Originally posted by @gpsfl in https://github.com/scalar/scalar/issues/1272#issuecomment-2014560366_

hanspagel:
Since we merged #1276 the schema is shown as `[Object object]`. This PR makes sure the schema is transformed to a string to fix that.

Actually, that shouldn’t be necessary. The code block checks the format anyway, but for some reason it is! So let’s just roll with it. :)

### DIFF
--- a/.changeset/shiny-shirts-laugh.md
+++ b/.changeset/shiny-shirts-laugh.md
@@ -1,0 +1,5 @@
+---
+"@scalar/api-reference": patch
+---
+
+fix: does not show schema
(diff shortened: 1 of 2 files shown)


amritk (changes requested):
@hanspagel can you confirm if https://github.com/scalar/scalar/pull/1290 fixes it?

PR #1300, opened 2024-03-23
fix: use node watch instead of @parcel/watcher in CLI

marclave:
to run the executable of the cli now you can do

```
bun build src/index.ts --compile --outfile scalar
./scalar <commands>
```


hanspagel:
That’s probably better! Just want to mention that we had node watch and it was really, really slow (like 3s delay) in Node. I don’t know enough about bun: Do we switch to bun completely to publish the CLI? If so, let’s go! Otherwise, we might need to check alternative solutions or use both approaches.

PR #1302, opened 2024-03-23
fix: only lowercase hostname

marclave:
now `https://petstore3.swagger.io/API/v3/pet` will send as `https://petstore3.swagger.io/API/v3/pet` vs `https://petstore3.swagger.io/api/v3/pet`

i also wrote some tests 😏


amritk (approved):

hwkr (approved):
🙌

PR #1306, opened 2024-03-24
feat: tab and shift+tab codemirror, and escape codeblock hotkeys

marclave:
tab works!
shift tab works!
if you escape you can tab normally in the page :) an AMAZING experience boost

https://github.com/scalar/scalar/assets/6176314/1b15f649-a652-4eb4-9141-be6913d232e1


amritk (approved):
wow, so simple

PR #1319, opened 2024-03-26
fix: don't stretch markdown tables

hwkr:
Okay so I did a bunch of research on this and couldn't come up with a CSS way to do what we want to do. This PR mirrors how GitHub handles tables, see [this gist](https://gist.github.com/hwkr/391f2010150c199e195ad42990d15778) as an example.

* Test file: [petstorev3.json](https://github.com/scalar/scalar/files/14764505/petstorev3.json)

Google Chrome-2024-03-26-13-34-25@2x

### Add a fake cell to the right

I figured out a CSS solution to always have them full width but it requires squishing content to the left and setting `white-space: nowrap` to keep them from breaking.

Google Chrome-2024-03-26-13-25-58@2x

### `table-layout: auto`

We can use `table-layout: auto` to fill the space but it breaks out of the scroll container. We would have to add something like [rehype-wrap-all](https://github.com/florentb/rehype-wrap-all) (which doesn't seem very well supported) to add a scroll container.

Google Chrome-2024-03-26-13-40-04@2x

### DIFF
--- a/packages/api-reference/src/components/MarkdownRenderer/MarkdownRenderer.vue
+++ b/packages/api-reference/src/components/MarkdownRenderer/MarkdownRenderer.vue
@@ -220,7 +220,8 @@ watch(
   overflow-x: auto;
   position: relative;
   border-collapse: collapse;
-  width: 100%;
+  width: max-content;
+  max-width: 100%;
   margin: 1em 0;
   box-shadow: 0 0 0 1px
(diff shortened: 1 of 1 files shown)


amritk (approved):
Didn't test, but elegant css only solution 👍🏾 👍🏾 👍🏾

marclave (approved):
great work @hwkr !

PR #1323, opened 2024-03-27
feat: move sidebar to css

amritk:
Our current JS controlled sidebar doesn't work in SSR so this uses css classses (controlled by js) to show/hide the sidebar


marclave (approved):
the saga continues ✨ 

great work @amritk

PR #1324, opened 2024-03-27
fix: theme variable name + remove some importants

cameronrohani:
First time using sidebar indent border hover variable and I misnamed it 😅

+ remove the important's that are no longer needed


marclave (approved):
great work @cameronrohani

PR #1325, opened 2024-03-27
fix: revert major api version

amritk:
We accidentally had a major version upgrade in there, this reverts it.


marclave (approved):
nice work @amritk ! 😅

PR #1327, opened 2024-03-27
fix: typescript in hono example

amritk:
Just some minor changes, fixed types and added the scalar package which was missing.


marclave (approved):
✨ nice work @amritk

PR #1330, opened 2024-03-28
feat: add SSR state to references and client

amritk:
The Final PR  in the SSR Prequel saga. This PR instantiates the client state from the server. This does not need to be checked, but we should just poke around and see if eveyrthing works:
- markdown
- auth
- description


marclave (approved):
awesome work @amritk !!!!

PR #1331, opened 2024-03-28
feat: add configuration to hide models from sidebar, search and content, fix #1312

### LINKED ISSUE #1312: add configuration to hide the models
### What happens?

currently we always show models

we can hide the models from the sidebar, search + docs :) 

have a spec with models

### Swagger/OpenAPI Example

_No response_

hanspagel:
Currently models (components.schemas) are always shown. As requested in #1312 this PR adds an option to hide models from the sidebar, search and content.

By default, models are shown.

This PR hooks into the `hasModels` helper. But it felt wrong to pass the configuration as a prop through all levels, so I did something I dreamed of for a while: I’ve just put the configuration in a globally accessible store (read-only).

Let me know what you think!

### DIFF
--- a/packages/api-reference/src/components/ApiReferenceLayout.vue
+++ b/packages/api-reference/src/components/ApiReferenceLayout.vue
@@ -6,7 +6,7 @@ import {
   ThemeStyles,
 } from '@scalar/themes'
 import { useDebounceFn, useMediaQuery, useResizeObserver } from '@vueuse/core'
-import { computed, onMounted, provide, ref } from 'vue'
+import { computed, onMounted, provide, ref, watch } from 'vue'
 import { toast } from 'vue-sonner'
 
 import {
@@ -16,6 +16,7 @@ import {
 } from '../helpers'
 import { useNavState, useSidebar } from '../hooks'
 import { useToasts } from '../hooks/useToasts'
+import { useConfigurationStore } from '../stores'
 import type {
   ReferenceLayoutProps,
   ReferenceLayoutSlot,
@@ -36,6 +37,17 @@ defineEmits<{
   (e: 'linkSwaggerFile'): void
   (e: 'toggleDarkMode'): void
 }>()
+
+// Put the configuration in a globally accessible store
+const { setConfiguration } = useConfigurationStore()
+
+watch(
+  () => props.configuration,
+  () => {
+    setConfiguration(props.configuration)
+  },
+  { immediate: true, deep: true },
+)
 
 defineOptions({
   inheritAttrs: false,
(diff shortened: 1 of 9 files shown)


packages/api-reference/src/components/ApiReferenceLayout.vue
```diff
@@ -37,6 +38,17 @@ defineEmits<{
   (e: 'toggleDarkMode'): void
 }>()
 
+// Put the configuration in a globally accessible store
+const { setConfiguration } = useConfigurationStore()
```
marclave, line 42: @hanspagel do we need to put the config in a global store?

I think we want to prop drill down but i'm open to doing this but I want to have a conversation before adding it ✨ @amritk @geoffgscott

hanspagel:
Just want to add, that we could use that to remove a lot of the prop drilling. I’d see this as a benefit, because it might be easier to follow and reason about.

packages/api-reference/src/components/ApiReferenceLayout.vue
amritk: So this is the way I think of it

- if data is needed 1-2 children deep, prop drill
- if data is needed deeper than that, but still a descendant, [use provide/inject](https://vuejs.org/guide/components/provide-inject)
- if data is needed in a sibling, or no direct parent/descendant path use a store
- if data need is triggered by an event, use the bus

what you think @geoffgscott, this is more of a style guide thing as they all work

packages/api-reference/src/components/ApiReferenceLayout.vue
```diff
@@ -37,6 +38,17 @@ defineEmits<{
...
+
+watch(
```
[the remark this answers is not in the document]
geoffgscott: I think it should be ok @marclave . Vue will handle the diffing internally. If it becomes an issue we can break out individual refs/computed properties in the store.

packages/api-reference/src/components/ApiReferenceLayout.vue
geoffgscott: Yeah @amritk I agree. I would probably prop drill up to 4 levels but it's preferential. 

While _generally_ I would not agree with this pattern there could be a use case in the future here if we find we are referencing the config object very deeply. I don't think this use case justifies it. we are only drilling from SearchButton through to SearchModal. 

Global stores should really be avoided as much as possible _especially_ when you want to control the store values from external triggers.

geoffgscott (changes requested):
Lets remove the global store for now. I don't think this use case is strong enough to support it.

marclave:
closing in favor of https://github.com/scalar/scalar/pull/1383

PR #1346, opened 2024-03-30
fix: set global security on sidebar navigation

### LINKED ISSUE #1336: Authentication resets on sidebar navigation
### What happens?

[Screencast from 2024-03-28 23-52-28.webm](https://github.com/scalar/scalar/assets/42624869/cdcc004a-947f-46ae-950c-de0e640fbec3)

Currently, when changing the current endpoint via sidebar, the authentication header is cleared.
The UI continues showing it. However, it is not added to the request headers.

Authentication header to be present in the request.

1. Add authentication to the API
2. Click on "Test Request" in any endpoint
3. Navigate through the endpoints via sidebar
4. Try to "Send Request"

### Swagger/OpenAPI Example

_No response_

Amorim33:
fix  #1336  

## **Problem**

Currently, with the api client opened, when navigating through the endpoints via sidebar the global authentication header is not sent in the request.

## **Explanation**
Probably, this pull request introduced the bug https://github.com/scalar/scalar/pull/1188.

This happens because `useSidebar` was not updated, so every time the `openClientFor` function was called, without the global security parameter being passed, the `globalSecurity` of `useOpenApiStore` was set to undefined.

## **Solution**
With this PR, `useSidebar` has been updated so that `openClientFor` calls pass the `globalSecurity` parameter, thus preventing it from being overwritten

## **Follow-up ideas**

- Add e2e test to prevent regressions in this behavior;
- Solve this TODO https://github.com/scalar/scalar/blob/main/packages/api-reference/src/helpers/openClientFor.ts#L39, in order to avoid passing `globalSecurity` as a parameter.

### DIFF
--- a/.changeset/olive-onions-approve.md
+++ b/.changeset/olive-onions-approve.md
@@ -1,0 +1,5 @@
+---
+"@scalar/api-reference": patch
+---
+
+fix: set global security on sidebar navigation
--- a/packages/api-reference/src/hooks/useSidebar.ts
+++ b/packages/api-reference/src/hooks/useSidebar.ts
@@ -1,4 +1,4 @@
-import { useApiClientStore } from '@scalar/api-client'
+import { useApiClientStore, useOpenApiStore } from '@scalar/api-client'
 import { type TransformedOperation, ssrState } from '@scalar/oas-utils'
 import { type OpenAPIV3_1 } from '@scalar/openapi-parser'
 import { computed, reactive, ref, watch } from 'vue'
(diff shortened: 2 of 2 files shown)


marclave:
great work @Amorim33 ! gunna test this with my morning coffee! ☕ 😎

Amorim33:
> great work @Amorim33 ! gunna test this with my morning coffee! ☕ 😎

Good morning @marclave !!

marclave (approved):
wow this is awesome @Amorim33 ! great work ✨ 

let's get this merged 😎

Issue #1339, opened 2024-03-29
URL slug not staying as expected after scroll of @scalar/api-reference documention


Rahul-kalsha:
### What happens?

I integrated scalar API documentation via [CDN link](https://github.com/scalar/scalar?tab=readme-ov-file#from-a-cdn) in angular SPA project. but there is a issue coming on url slug. my existing url slug not staying when scroll down the API documentation. Here i am adding the code. Help me to fix this bug. 

Here is the component file:

```typescript
import { AfterViewInit, Component, OnDestroy, OnInit, Renderer2 } from '@angular/core';

@Component({
  selector: 'app-api-list',
  standalone: true,
  imports: [],
  templateUrl: './api-list.component.html',
  styleUrls: ['./api-list.component.scss'],
})
export class ApiListComponent implements OnInit, AfterViewInit, OnDestroy {
  constructor() { }

  ngOnInit(): void {
    this.loadScript('https://cdn.jsdelivr.net/npm/@scalar/api-reference').then((res: any) => {
      console.log('res:', res);
    }).catch(error => {
      console.error('Error loading script:', error);
    });
  }

  ngOnDestroy(): void {
  }

  ngAfterViewInit(): void {
  }

  private loadScript(url: string): Promise<void> {
    return new Promise<void>((resolve, reject) => {
      const scriptElement = document.createElement('script');
      scriptElement.src = url;
      scriptElement.onload = () => resolve();
      scriptElement.onerror = (error) => reject(error);
      document.body.appendChild(scriptElement);
    });
  }
}
``` 

Here is the html file:
```html
<div
id="api-reference"
data-url="https://petstore3.swagger.io/api/v3/openapi.json"
data-proxy-url="https://api.scalar.com/request-proxy"></div>
``` 

You can check visually here and suggest me proper solution if i am miss any configuration then also please help me.

https://github.com/scalar/scalar/assets/56481473/de108e22-09c8-4322-8d73-19c3837b78b0

### What did you expect to happen?

When we scroll down API documentation Url 

Should be like:

```
http://localhost:4200/public-api-list#tag/pet
http://localhost:4200/public-api-list#tag/store
http://localhost:4200/public-api-list#tag/user
http://localhost:4200/public-api-list#models
```

Not like that:

```
http://localhost:4200/#tag/pet
http://localhost:4200/#tag/store
http://localhost:4200/#tag/user
http://localhost:4200/#models
```

### How can we reproduce the issue?

[...]

PR #1348, opened 2024-03-30
fix(#1339): only update hash value on scroll

amritk:
**Problem**
We remove the rest of the path on scroll (this doesn't seem to happen on local but it has for a user)

**Explanation**
we were just passing the hash into history state

**Solution**
With this PR we create a new URL with the current href and just update the hash. This _should_ work


marclave (approved):
nice fix :)

PR #1351, opened 2024-04-01
fix: descriptions repeat themselves

### LINKED ISSUE #1318: Descriptions repeat themselves
### What happens?

Descriptions of an object repeat twice:

image

Not to repeat descriptions

Go to sandbox

### Swagger/OpenAPI Example

[https://sandbox.scalar.com/e/SpV4O](https://sandbox.scalar.com/e/SpV4O)

---

It gets even weirder when testing against [https://docs.scalar.com/swagger-editor](https://docs.scalar.com/swagger-editor). `a`, `c`, `d` work, but `b` has no description at all.

image

antlio:
**Problem**
as described in #1318, the description of an object is duplicated, as seen below: 

Screenshot 2024-04-01 at 11 25 12

**Solution**
this PR is an attempt to fix #1318 by introducing a function for displaying the description unless the value has properties.

after the change:
Screenshot 2024-04-01 at 11 30 27

### DIFF
--- a/.changeset/fuzzy-wombats-thank.md
+++ b/.changeset/fuzzy-wombats-thank.md
@@ -1,0 +1,5 @@
+---
+"@scalar/api-reference": patch
+---
+
+fix: use function for displaying description in schema property component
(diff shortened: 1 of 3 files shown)


marclave (changes requested):
wow, awesome work @antlio ! what an amazing thing to wake up to!

we just need to change the `major` to a `patch` and we can get this merged 😎

marclave (approved):

Issue #1355, opened 2024-04-02
Anchor tag root path is not kept as is


Rahul-kalsha:
### What happens?

1) I am using CDN link in angular project. Anchor tag route path not coming. you can see below visualization.
https://github.com/scalar/scalar/assets/56481473/1b30a131-6c9b-44b1-bbb0-3c66dfe130bc

### What did you expect to happen?

1) The anchor tag root path should be in the page it's in.

### How can we reproduce the issue?

1) You can check attached video

marclave:
@Rahul-kalsha this was fixed in https://github.com/scalar/scalar/pull/1359

Rahul-kalsha:
Hi @marclave, Thank you for fixed that issue, but I have gotten another issue like when we click on sidebar menu scroll to section not working.

https://github.com/scalar/scalar/assets/56481473/3961113b-427e-4ccc-991a-ecd7de7dd9e6

amritk:
hey @Rahul-kalsha do you think you could share a minimal reproducible example? This one would require some specific debugging since it seems to be working currently in our examples

Rahul-kalsha:
Hi @amritk, You can visit this my stackblitz link, I just uploaded a demo for you to debug
.
https://angular-scalar-api.stackblitz.io/api-list
https://stackblitz.com/edit/angular-scalar-api?file=src/app/components/api-list/api-list.component.ts

krishnamenon22:
Hi @amritk  the issue still persists for Side bar.

amritk:
hey @krishnamenon22 @Rahul-kalsha do you think you could help me add an angular example to the repo? Then we can fix this right quick

hanspagel:
I think that should be fixed, happy to reopen with a reproduction. ✌️

PR #1359, opened 2024-04-02
fix(#1355): root path on sidebar items

amritk:
This will build the URL using the constructor then update the hash. This is only temporary, soon we will switch to vue-router.


marclave (approved):
awesome work @amritk !!!

PR #1365, opened 2024-04-02
feat: improve API Client request parameter tables

hwkr:
This code was a little crazy so I ended up doing a fair bit of refactor into components as well as scoping styles. Feel free to play around with it and make sure everything still works.

That being said I also tried not to refactor too too much since a bunch of the API Client might be changing soon.

I broke the commits up so it might be easier to review commit by commit to see what moved where.

1. [Organized the grid styles and scoped them](https://github.com/scalar/scalar/pull/1365/commits/0f460606867547f275e984f84c7794b834241d99)
2. [Componentized the grid rows](https://github.com/scalar/scalar/pull/1365/commits/93f9fff0e91b83a77be8f4df9a1eb5acb0b62ecb)
3. [Enabled the parameter on field change](https://github.com/scalar/scalar/pull/1365/commits/96f653cec0be1ed38935c62a3e563912ad7b9b68)

### Before

https://github.com/scalar/scalar/assets/6374090/9930db28-27a0-416b-93d3-95e8c510f202

### After

https://github.com/scalar/scalar/assets/6374090/c81cb250-13fa-46d5-b97e-b9b02cd95f87


packages/api-client/src/components/ApiClient/RequestHistory.vue
```diff
@@ -40,8 +40,238 @@ const { requestHistoryOrder } = useRequestStore()
     </div>
   </div>
 </template>
-
 <style>
+.navtable {
```
hwkr, line 44: Some of these un scoped styles I just moved to be closer to the components that consume them.

marclave (approved):
awesome work @hwkr ✨

PR #1366, opened 2024-04-02
feat: add header example

hwkr:
This shows what is could look like if we replaced the `<script>`  tag instead of injecting next to it. I don't _think_ this would have any downstream effects for users. If we want to take a safer route we could also just add a class or id to the div we inject.

Here's what the DOM looks like with this approach:

Google Chrome-2024-04-02-15-07-02@2x

Vs how we do it now:

Google Chrome-2024-04-02-15-13-14@2x

### DIFF
--- a/examples/cdn-api-reference/src/index.ts
+++ b/examples/cdn-api-reference/src/index.ts
@@ -13,6 +13,10 @@ app.get('/live', (_request, reply) => {
   reply.sendFile('api-reference-cdn-live.html', { cacheControl: false }) // overriding the options disabling cache-control headers) // serving path.join(__dirname, 'public', 'myHtml.html') directly
 })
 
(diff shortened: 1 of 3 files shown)


marclave:
nice work @hwkr !

im curious on what your thoughts are for this?
>  I don't think this would have any downstream effects for users.

what do you imagine could be a downstream effect?

hwkr:
> nice work @hwkr !
> 
> im curious on what your thoughts are for this?
> 
> > I don't think this would have any downstream effects for users.
> 
> what do you imagine could be a downstream effect?

Theoretically if someone was either targeting the `<div>` with some weird CSS that relied on the the relative position of the `<div>` to the `<script>` tag we could break their styles. This would definitely make their lives easier going forward though for anyone who wants to target our app with CSS.

PR #1367, opened 2024-04-03
feat: add custom security option

amritk:
This allows users to add basic security when the spec does not contain any. This security should also persist between requests.

Currently this PR works but is missing oAuth 2


marclave (approved):
✨

PR #1383, opened 2024-04-05
feat: add config to hide models

marclave:
you can now hide models which results in not showing it in the Content, Sidebar & search

### DIFF
--- a/packages/api-reference/src/types.ts
+++ b/packages/api-reference/src/types.ts
@@ -33,6 +33,12 @@ export type ReferenceConfiguration = {
   isEditable?: boolean
   /** Whether to show the sidebar */
   showSidebar?: boolean
+  /**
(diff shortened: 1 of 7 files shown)


packages/api-reference/src/types.ts
```diff
@@ -33,6 +33,12 @@ export type ReferenceConfiguration = {
   isEditable?: boolean
   /** Whether to show the sidebar */
   showSidebar?: boolean
+  /**
+   * Whether to show models in the sidebar, search, and content.
```
amritk, line 37: 10 points for jsdoc

amritk: we should probably start adding defaults to all of our configs, then display the types somehow

amritk (approved):
lets goo

Issue #1250, opened 2024-03-20
hiddenClients property doesnt hide clients


Lumca:
### What happens?

It stays as it is, even when empty array is passed, it doesnt show clients that are not shown by default

### What did you expect to happen?

Hide the clients

Or that could be a feature where can I set which clients I want to show, we want to show just a HTTP1.1. 
But the feature for hidding is good enough

### How can we reproduce the issue?

Set for example hiddenClients: [`node`] in ApiReference

Thanks for checking! :)

hanspagel:
Thanks for the report! I’ll fix that.

SlavaSereb:
Hey @hanspagel I can see that you're extensively working on the fix, thank you!

Any estimated time for pushing it?

hanspagel:
Very soon :)

hanspagel:
Should be fixed in #1387 :)

Lumca:
Thanks guys!

Lumca:
So I have tested it and there is still little bug.
I have setted:
`hiddenClients: {
          php: true,
          python: true,
          c: true,
          node: true,
          javascript: true,
          go: true,
          java: true,
          ruby: true,
          shell: true,
          clojure: true,
          csharp: true,
          kotlin: true,
          objc: true,
          swift: true,
          r: true,
          powershell: true,
          ocaml: true,
          curl: true,
        },`

It still makes curl default:
![image](https://github.com/scalar/scalar/assets/4989415/966b9d8c-4c8e-4837-940b-8cd0c67cec4c)

and upon clicking there is only http
![image](https://github.com/scalar/scalar/assets/4989415/795feab7-4b46-4a6f-9e92-e8200451bd11)

should I create new bug? :) 

thank you guys!

PR #1387, opened 2024-04-05
fix: non-working `hiddenClients` configuration option 

HelgeSverre:
**Problem**
Because the "allTargets()..."-code is defined outside of the usehttpClients() hook, it will only run once (which was probably done due to performance benefits), however this breaks this feature as only the DEFAULT_EXCLUDED_CLIENTS will be applied the first time this runs, and the next time this is meant to run to filter the targets and clients based on the provided user configuration, it is not updated, aka, doesnt work, this is references in #1250.

**Solution**

This commit fixes the bug, and retains the performance benefit by caching the target list using watchEffect to only update/recompute the target list when the configuration changes.

**Further improvements** 

I'm working on an additional fix, that will add the ability to specify clients in specific targets to ignore/include, since the behaviour right now will hide all the "curl"-s instead of only hiding the "php"-curl if you wanted to do so. 

Still WIP, but feedback and pointers welcome, struggling a bit with TypeScript errors, even though running the embedding demo with this config works in the browser:

```html
const configuration = reactive<ReferenceConfiguration>({
  proxy: import.meta.env.VITE_REQUEST_PROXY_URL,
  isEditable: false,
  spec: { content },
  hiddenClients: {
    php: true, // Example: Exclude all PHP clients
    python: true, // Example: Exclude all Python clients
    c: ['libcurl'],
    node: ['fetch'], // We expect the (browser) javascript fetch client to still be available
    javascript: ['jquery', 'xhr'], // Exclude specific JavaScript clients
  },
})

```

### DIFF
--- a/packages/api-reference/src/components/Content/ClientLibraries/ClientSelector.vue
+++ b/packages/api-reference/src/components/Content/ClientLibraries/ClientSelector.vue
@@ -43,16 +43,16 @@ const featuredClients = (
       clientKey: 'libcurl',
     },
   ] as const
-).filter((featuredClient) => {
-  return availableTargets.value.find((target: AvailableTarget) => {
+).filter((featuredClient) =>
+  availableTargets.value.find((target: AvailableTarget) => {
     return (
       target.key === featuredClient.targetKey &&
       target.clients.find(
         (client: ClientInfo) => client.key === featuredClient.clientKey,
       )
     )
-  })
-})
+  }),
+)
 
 /**
  * Icons have longer names to appear in icon searches, e.g. "javascript-js" instead of just "javascript". This function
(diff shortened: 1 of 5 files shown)


hanspagel:
Thanks for the awesome PR! The type check seems to fail, I’ll look into that. :)

hanspagel (approved):
Looks 🌟  awesome 🌟 to me!

amritk (approved):
Awesome PR! I left a few comments but mostly nit picks and nothing major. Overall its great!

hanspagel:
I’ll take care of @amritk’s feedback. ✌️

hanspagel:
@amritk Would you mind to do a 2nd review? 🙌

amritk (approved):
Nice, looks clean 👍🏾

hanspagel:
@marclave Final review? Let me know if it needs further changes. 👍

marclave (approved):
wow, awesome work @HelgeSverre and @hanspagel for getting this over the finish line!

tested and works GREAT!!!

PR #1388, opened 2024-04-06
feat: add config to hide download button

marclave:
now you can hide the download button :) 

image


amritk (approved):
Should we just add all config options into a performant global store like zustand?

marclave:
> Should we just add all config options into a performant global store like zustand?

im down! lets do that in a separate body of work :)

PR #1398, opened 2024-04-09
feat: improve tooling and move ssr state to ctx.payload.data

amritk:
SSR prequels are back! This one includes:
- some minor tooling
- make the dev server use turbo
- move creat hash to oas-utils
- move our server state to ctx.payload.data
- fix sonner toast on ssr
- fix scalar codeblock for SSR

Changeset is in the next PR


marclave (approved):
awesome work @amritk !!!

PR #1401, opened 2024-04-09
fix: endpoint navigation

antlio:
**Problem**
when a section is closed and user clicks on an operation, it opens the section and navigates toward the first operation of the list instead of the selected one.

[as seen here →](https://docs.scalar.com/project-default/version-default/reference-YCSmGxwEOVk-0oBSMCo-1#tag/store)

**Solution**
this pr is a first attempt to follow the commented instructions, in order to make sure the endpoint is accessible before navigating towards it by using the [mutation observer api](https://vueuse.org/core/useMutationObserver/) from vueuse.

### DIFF
--- a/packages/api-reference/src/components/Content/Tag/Endpoints.vue
+++ b/packages/api-reference/src/components/Content/Tag/Endpoints.vue
@@ -17,19 +17,21 @@ import {
 
 const props = defineProps<{ id?: string; tag: Tag }>()
 
+const emit = defineEmits<{
+  (event: 'observeAndNavigate', value: string): void
+}>()
+
 const { getOperationId, getTagId } = useNavState()
 const { setCollapsedSidebarItem } = useSidebar()
 
-// We need to make sure the endpoint tag is open before
-// we try to scroll to it
-// we wait for next render after we open the tag
+const navigateToOperation = (operationId: string) => {
+  emit('observeAndNavigate', operationId)
+}
+
 // TODO in V2 we need to do the same loading trick as the initial load
 async function scrollHandler(operation: TransformedOperation) {
+  navigateToOperation(getOperationId(operation, props.tag))
   setCollapsedSidebarItem(getTagId(props.tag), true)
-
-  setTimeout(() => {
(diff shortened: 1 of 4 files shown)


packages/api-reference/src/components/Content/Tag/Endpoints.vue
```diff
@@ -17,19 +17,21 @@ import {
 
 const props = defineProps<{ id?: string; tag: Tag }>()
 
+const emit = defineEmits<{
+  (event: 'observeAndNavigate', value: string): void
+}>()
```
amritk, line 22:
```suggestion
const emit = defineEmits<{
  [observeAndNavigate: [value: string]
}>()
```
While there's absolutely nothing wrong with that syntax, this is the [cool new way of doing emits](https://vuejs.org/guide/typescript/composition-api#typing-component-emits)


image

packages/api-reference/src/components/Content/Tag/Endpoints.vue
```diff
@@ -17,19 +17,21 @@ import {
...
-// We need to make sure the endpoint tag is open before
-// we try to scroll to it
-// we wait for next render after we open the tag
+const navigateToOperation = (operationId: string) => {
+  emit('observeAndNavigate', operationId)
+}
+
 // TODO in V2 we need to do the same loading trick as the initial load
 async function scrollHandler(operation: TransformedOperation) {
+  navigateToOperation(getOperationId(operation, props.tag))
```
amritk, line 33:
```suggestion
async function scrollHandler(operation: TransformedOperation) {
   emit('observeAndNavigate', getOperationId(operation, props.tag))
```

Super nit pick, it can go either way really. We can also get rid of that old comment

amritk (approved):
Just super nitty gritty like we were saying, add some comments and its good to go 👍🏾

packages/api-reference/src/components/Content/Tag/Endpoints.vue
antlio: wow til! thank you

PR #1407, opened 2024-04-09
fix(#1405): decode uri component for hash values

### LINKED ISSUE #1405: Support URL encoded params
### What happens?

When I share a link to our API reference document the link is formatted as follows:
`https://example.com/reference#tag/credential-templates/post/v1/projects/{projectId}/templates/credentials/sd-jwt-vc`

Now when I post think link in e.g. slack and then click on the link, either slack or my browser will automatically encode the URL and thus open it like this:
`https://example.com/reference#tag/credential-templates/post/v1/projects/%7BprojectId%7D/templates/credentials/sd-jwt-vc`

Because `{projectId}` is encoded to `%7BprojectId%7D`, the link now won't link to the specific endpoint anymore.

I would expect event the url encoded link to link to the correct endpoint

1. Go to the sandbox link below
2. Copy the link to the endpoint in there it should have `{projectId}`
3. Paste this in .e.g your notes app on macOS (not sure if windows e.g. also does this) so it is shown as a link and click on the link.

The link you copied should be `https://sandbox.scalar.com/p/Xnomb#tag/default/get/v1/projects/{projectId}/webhooks`, while the link you're now navigated to is `https://sandbox.scalar.com/p/Xnomb#tag/default/get/v1/projects/%7BprojectId%7D/webhooks`.

### Swagger/OpenAPI Example

https://sandbox.scalar.com/e/Xnomb

amritk:
fixes #1405 
Sometimes URL's get encoded, this will ensure selecting the correct endpoint works even if it gets encoded.

### DIFF
--- a/.changeset/large-bags-end.md
+++ b/.changeset/large-bags-end.md
@@ -1,0 +1,5 @@
+---
+"@scalar/api-reference": patch
+---
+
+fix: decode uri component when setting hash values from url
--- a/packages/api-reference/src/hooks/useNavState.ts
+++ b/packages/api-reference/src/hooks/useNavState.ts
@@ -61,7 +61,8 @@ const getSectionId = (hashStr = hash.value) => {
 }
 
 // Update the reactive hash state
-const updateHash = () => (hash.value = window.location.hash.replace(/^#/, ''))
+const updateHash = () =>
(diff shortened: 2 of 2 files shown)


marclave (approved):
awesome fix @amritk :)

PR #1410, opened 2024-04-09
fix: show example picker if the list only has one example

### LINKED ISSUE #1406: Response `examples` not used (only `example`)
### What happens?

When using `examples` in the response content media type examples for OpenAPI, they are not rendered by Scalar.

If you look at the OpenAPI spec, it mentions you can use `examples`, allowing you to add extra metadata like description, etc..: https://spec.openapis.org/oas/v3.1.0#media-type-examples.

I worked around it for now by post processing our generated OpenAPI file and create an `example` field based on the first example from the examples object

At least the first example from examples is rendered, preferably all examples

1. Go to the example sandbox and see the difference between the examples. The `examples` are not rendered, but `example` is.

### Swagger/OpenAPI Example

https://sandbox.scalar.com/e/BpTOG

hwkr:
Fixes #1406

### Test Spec

```json
{
  "openapi": "3.0.0",
  "components": {
    "examples": {},
    "headers": {},
    "parameters": {},
    "requestBodies": {},
    "responses": {}
  },
  "paths": {
    "/example": {
      "get": {
        "summary": "Single example in `examples`",
        "responses": {
          "200": {
            "description": "Ok",
            "content": {
              "application/json": {
                "schema": {
                  "type": "object"
                },
                "examples": {
                  "one": {
                    "value": {
                      "hello": "example nested"
                    }
                  }
                }
              }
            }
          }
        }
      }
    },
    "/example2": {
      "get": {
        "summary": "Multiple examples in `examples`",
        "responses": {
          "200": {
            "description": "Ok",
            "content": {
              "application/json": {
                "schema": {
                  "type": "object"
                },
                "examples": {
                  "one": {
                    "value": {
                      "hello": "example"
                    }
                  },
                  "two": {
                    "value": {
                      "hello": "example"
                    }
                  }
                }
              }
            }
          }
        }
      }
    },
    "/example3": {
      "get": {
        "summary": "Single example",
        "responses": {
          "200": {
            "description": "Ok",
            "content": {
              "application/json": {
                "schema": {
                  "type": "object"
                },
                "example": {
                      "hello": "example"
                }
              }
            }
          }
        }
      }
    }
  }
}
```

## Before (Main)

Google Chrome-2024-04-09-12-51-46@2x

## After (This PR)

Google Chrome-2024-04-09-12-50-48@2x

### DIFF
--- a/.changeset/seven-actors-film.md
+++ b/.changeset/seven-actors-film.md
@@ -1,0 +1,5 @@
+---
+"@scalar/api-reference": patch
+---
+
+fix: show example picker if the list only has one example
--- a/packages/api-reference/src/components/Content/Operation/PathResponses/PathResponses.vue
+++ b/packages/api-reference/src/components/Content/Operation/PathResponses/PathResponses.vue
@@ -41,9 +41,7 @@ const orderedStatusCodes = computed(() => {
 })
 
 const hasMultipleExamples = computed<boolean>(
-  () =>
-    !!currentJsonResponse.value.examples &&
-    Object.keys(currentJsonResponse.value.examples).length > 1,
+  () => !!currentJsonResponse.value.examples,
 )
 
 // Keep track of the current selected tab
--- a/packages/api-reference/src/components/Content/Operation/PathResponses/PathResponses.vue
+++ b/packages/api-reference/src/components/Content/Operation/PathResponses/PathResponses.vue
@@ -69,7 +69,8 @@ const currentResponseWithExample = computed(() => ({
   ...currentJsonResponse.value,
   example:
     hasMultipleExamples.value && selectedExampleKey.value
-      ? currentJsonResponse.value.examples[selectedExampleKey.value]
+      ? currentJsonResponse.value.examples[selectedExampleKey.value].value ??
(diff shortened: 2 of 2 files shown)


hwkr:
I guess we also render the `value` stuff maybe we should sub select that?

TimoGlastra:
> I guess we also render the `value` stuff maybe we should sub select that?

Yes, i think only the value should be rendered. You could add properties like description top-level to add more context to the example

hwkr:
Hey @TimoGlastra! I updated it to use the `$.value` field matching with the OpenAPI spec and fall back to the entire object if not. This PR should be good to merge now.

On another note we should really clean up the types in the responses @marclave, it's like the wild west in there 😅

marclave:
> Hey @TimoGlastra! I updated it to use the `$.value` field matching with the OpenAPI spec and fall back to the entire object if not. This PR should be good to merge now.
> 
> On another note we should really clean up the types in the responses @marclave, it's like the wild west in there 😅

wanna cut a ticket to prune up the responses :) great idea

marclave (approved):
✨ nice work @hwkr :)

PR #1411, opened 2024-04-09
fix: allow json primitives as valid json objects

### LINKED ISSUE #1396: Response message get processed by JSON.stringify() multiple times.
### What happens?

Response message get processed by JSON.stringify() multiple times.

See example
Screenshot 2024-04-08 233406 (2)

Method "ping" actually returns string `"Pong\n"` not the `"\\"Pong\\\\n\\""`.

I want to note that i'm not using data-proxy-url.
It seems that using a proxy may change the visible response value.

Raw data that server returns:
Screenshot 2024-04-08 233406 (3)

The same result can be achieved by using this command:
`console.log(JSON.stringify(JSON.stringify("Pong\n")));`

### Swagger/OpenAPI Example

_No response_

hwkr:
Fixes #1396

### DIFF
--- a/.changeset/happy-moons-draw.md
+++ b/.changeset/happy-moons-draw.md
@@ -1,0 +1,5 @@
+---
+"@scalar/oas-utils": patch
+---
+
+fix: allow json primitives as valid json objects
--- a/packages/oas-utils/src/parse.test.ts
+++ b/packages/oas-utils/src/parse.test.ts
@@ -8,26 +8,74 @@ import {
   yaml,
 } from './parse'
 
-describe('Handles yaml and json parsing', () => {
-  test('Parses basic yaml', () => {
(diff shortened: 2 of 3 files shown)


geoffgscott:
@marclave we may want to consider this but it will have significant downstream impacts as "AnyObject" is not the true valid return type of `Content-Type: 'application/json'`

I am also a little confused about the original issue as it looks like their request was using `Content-Type: 'text/plain'` in which case no parsing should be going on.

hwkr:
Going to close this for now until @geoffgscott gets a chance to chat with @marclave @hanspagel about how we want to proceed

hanspagel:
I think we should focus on the shift to the new proxy first (because it affects basically all users) and get this PR here in after we’ve migrated to the new proxy.

PR #1415, opened 2024-04-10
fix: content modal history style

antlio:
**Problem**
in darkmode when a user open the request history modal, the header displays a background. according to #177 i believe this shouldn't be the case. hope i read you right @cameronrohani :). 

before

**Solution**
use inherit background in modal header for darkmode.

after


marclave (approved):
awesome fix @antlio ✨

PR #1416, opened 2024-04-10
chore: add playwright snapshot tests for all themes and layouts

hwkr:
Okay so this is a little crazy and might not work to run on every build due to time / processing cost but might be useful to just run on demand locally to check for some of those smaller theme / layout bugs (e.g. like the kind of refactor I'm doing right now).


marclave (approved):
awesome work @hwkr !!! will help so much ✨

PR #1418, opened 2024-04-11
feat: new scalarParse translation function

tmastrom:
This PR adds `scalarParse` to `oas-utils` to convert the output of the parser into a more usable data object for rendering. The translation layer dereferences and groups fields for rendering as well as creates new zod schemas for parsing incoming data. We want to allow malformed data as much as possible to render what we can.

The `parse.ts` function is removed from `@scalar/api-reference` and replaced with the new `scalarParse` function

Todo
- [ ] decide on critical error handling 
  - we can implement custom errors, refinements, and preprocessing in zod schemas as well as safe parsing data to give a good sense of what data is available
- [x] don't blow up the bundle size - what to do with testing yaml files? 
- [x] proper yaml import process for testing


packages/oas-utils/src/test-files/bbccouk.yaml
geoffgscott, line 1: We should break all the test files out into a separate repo so it can just be installed as a dev depedency and keep the spec files out of the main repo.

hanspagel:
Great work already! Shouldn’t we keep this in @scalar/api-reference, though? It’s not useful for other people, `transformResult` is just too specific to our needs. 🤔

packages/oas-utils/src/test-files/bbccouk.yaml
tmastrom: Moved these test files to the CDN for now while we figure out the best way to break out the test files

tmastrom:
> Great work already! Shouldn’t we keep this in @scalar/api-reference, though? It’s not useful for other people, `transformResult` is just too specific to our needs. 🤔

The only downside I see to this is adding a dependency to @scalar/api-reference but I would defer to @marclave and @geoffgscott for this decision.

hanspagel:
Dumb question: What dependency do we need to add to @scalar/api-reference? 😳

hanspagel:
I’m sorry to say this, but I think we should close this PR. It got a bunch of conflicts and we might get even more with the upcoming merge of The Secret Project(tm). 😬

marclave:
lets close this, sorry we didnt get this in sooner @tmastrom but we can defs cherry pick a lot of this work post "The Secret Project(tm). " :)

PR #1425, opened 2024-04-12
fix: intersection observer detection

antlio:
**Problem**
when a low height element lies between two others, it is not always detected by the observer as the next one according to scroll speed is already in the viewport and therefore not highlighted in the sidebar.

https://github.com/scalar/scalar/assets/14966155/9e75c2d3-4cf5-460a-af91-3f3086fb618c

**Solution**
this pr attempt adds two methods in order to calculate a margin + set the threshold according to the element height. good scroll yall!

https://github.com/scalar/scalar/assets/14966155/dd1d921b-fdb7-46d1-b321-4833da97ed38

### DIFF
--- a/.changeset/pink-bananas-exist.md
+++ b/.changeset/pink-bananas-exist.md
@@ -1,0 +1,5 @@
+---
+"@scalar/api-reference": patch
+---
+
+fix: intersection observer detection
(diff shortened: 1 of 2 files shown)


marclave (approved):
great fix @antlio ✨

PR #1428, opened 2024-04-12
fix: nuxt module fixes

amritk:
The problem isn't completely solved yet, but this fixes the import issues


marclave (approved):
😮‍💨 😮‍💨 😮‍💨

PR #1429, opened 2024-04-13
fix: sidebar heading type remove hover color

cameronrohani:
fixes new precedence issue and removes unused / code


marclave (approved):
great fix @cameronrohani !

Issue #1395, opened 2024-04-08
Markdown does not render properly for `object` types


databyjp:
### What happens?

The same Markdown that renders fine for `string` or other object types fail to render the property includes `additionalProperties`

Please see:
<img width="1487" alt="image" src="https://github.com/scalar/scalar/assets/61258750/b5b1da19-58fc-4956-bb8a-60f0d98c81f4">

It seems to render as a naked div without rendered <p> tags, unlike the working example above.

### What did you expect to happen?

Consistent markdown rendering behaviour regardless of the property datatype. 

### How can we reproduce the issue?

Add a `additionalProperties` value to the object.

PR #1435, opened 2024-04-15
fix: object description markdown

antlio:
**Problem**
as highlighted by @databyjp in #1395 — schema descriptions do not render markdown.

no markdown support

**Solution**
this PR is a proposal to use the markdown rendering for the schema value description.

markdown support


marclave (approved):
awesome fix @antlio ✨

databyjp:
Omg that's so fast - thank you @antlio and @marclave ❤️🙏

Issue #1440, opened 2024-04-16
Search option anchor tag element not taking relative path of @scalar/api-reference documention #1339


mjethva69:
### What happens?

I integrated scalar API documentation via [CDN link](https://github.com/scalar/scalar?tab=readme-ov-file#from-a-cdn) in angular SPA project. but there is a issue coming on url slug. Whenever I search the API then anchor tag element not taking relative path.
![Swagger](https://github.com/scalar/scalar/assets/83976275/b42d3b9a-b554-4333-8dcb-967e799a6d8d)

### What did you expect to happen?

When we search API documentation Url
Should be like:
```
http://localhost:4200/public-api-list#tag/pet
http://localhost:4200/public-api-list#tag/store
http://localhost:4200/public-api-list#tag/user
http://localhost:4200/public-api-list#models
```
Not like that:
```
http://localhost:4200/#tag/pet
http://localhost:4200/#tag/store
http://localhost:4200/#tag/user
http://localhost:4200/#models
```

### How can we reproduce the issue?

1. Setup CDN in Angular project as i already mentioned code above
2. Setup API documentation page in particular url slug
3. Search API
4. See URL slug

PR #1441, opened 2024-04-16
fix: give explicit URL to href in search modal

marclave:
fixes https://github.com/scalar/scalar/issues/1440


amritk (approved):
Lets goooo

PR #1448, opened 2024-04-16
fix(#1447): ctrl key for search on linux or windows

### LINKED ISSUE #1447: searchHotKey uses WIN (meta) instead of CTRL on Linux / Windows
### What happens?

The search box display shows `^K` so I assumed it to be CTRL  + K. Unfortunately this was not working, it seems to be using WIN + K instead.
I, using Linux, asked my colleague to check on Windows and they came across the same issue. Another colleague using macOS found the CMD+K shortcut to work fine.

Search box to open with CTRL + K, not with WIN + K

1. Use a Windows / Linux machine
2. Visit a page using scalar e.g. https://sandbox.scalar.com
3. Hit CTRL + K (nothing happens)
4. Hit WIN + K (works)

### Swagger/OpenAPI Example

https://sandbox.scalar.com

amritk:
Needs testing on MacOS but basically use CTRL as the search hotkey modifier on non-macos machines. Also prevent the browser default behavior.

closes #1447


marclave (approved):
rad fix @amritk !

works great on macos 😎

amritk:
why does it update all the things!? @geoffgscott

PR #1450, opened 2024-04-16
fix: run synpack to update package versions

amritk:
We had a non synpacked PR merged so our versions are outta whack. Specifically the vite version mismatch was causing issues. ~~This PR needs some more testing so I threw it as draft~~


marclave (approved):
:)

PR #1455, opened 2024-04-18
feat: render images in tag markdown description

MaxBreida:
**Problem**
Currently, I can't embed images in my markdown descriptions for tag descriptions. I have UML Diagrams that describe my workflows and would like to show them there.

**Explanation**
This happens because the markdown renderer component does use the optional prop value at this time.

**Solution**
With this PR, images will get rendered in tag descriptions when using the optional `withImages` prop value.


marclave (approved):
awesome work @MaxBreida !

PR #1457, opened 2024-04-19
fix: nuxt integration

amritk:
This fixes the commonJs issues that the nuxt integration was having. Also updated the theme and made the toaster client only.


marclave (approved):
unreal work @amritk

PR #1466, opened 2024-04-19
fix: hotfix integration fonts

hwkr:


packages/express-api-reference/src/expressApiReference.ts
```diff
@@ -9,9 +9,6 @@ export type ApiReferenceOptions = ReferenceConfiguration & {
  * The custom theme CSS for the API Reference.
  */
 export const customThemeCSS = `
-:root {
-  --scalar-font: 'Inter', var(--system-fonts);
```
hwkr, line 13: This line breaks everything because `var(--system-fonts)` wasn't being applied

marclave (approved):

amritk (approved):

PR #1474, opened 2024-04-22
fix: stop rendering infinite sub-schemas for type object

Amorim33:
see https://github.com/scalar/scalar/discussions/1049
## **Problem**
image

Currently, infinite sub-schemas are rendered for type object.

## **Explanation**
image

This happens because when we define a Record<String, unknown>. The property has type object, and additionalProperties only contains the nullable field.  
The fact that the additionalProperties does not have a "type" field, makes this spread to not overwrite the type object from the original value.
Causing the Schema component to be recursively called.

> Check the "Schema" component from the api reference.

## **Solution**
With this PR the original "additionalAttributes" from the schema is overrided, in order to avoid infinite schemas merge.

## QA
- Added vue test utils to the api reference;
- Configured jsdom in the vitest config file;
- Added tests for many possibilities of "sub-schema" render in the `SchemaProperty`component.

> Followed arrange/act/assert pattern for tests

## 🚨🚨
- Added tests to avoid further regressives;
- This is a hotfix. However, I believe that the logic of merging schemas is becoming too complex to maintain in long-term. Maybe it should be refactored.


Amorim33:
https://github.com/scalar/scalar/pull/1424/

Maybe this pull request solved the problem, will rebase and check if the tests are passing

Amorim33:
This pull request https://github.com/scalar/scalar/pull/1424 solved the infinite sub-schemas behavior. However, for `Record<string, unknown>` "Show child attributes" is still beign rendered without relevant information. Is this the expected behavior?

image

What should I do with this PR?

hwkr:
> This pull request #1424 solved the infinite sub-schemas behavior. However, for `Record<string, unknown>` "Show child attributes" is still being rendered without relevant information. Is this the expected behavior?

Hey! That probably shouldn't be rendering. If you can send me the schema you're using I can dig into it :)

Amorim33:
Hello @hwkr !!

Here is the property schema:
```json
{
     "metadata": {
            "type": "object",
            "nullable": true,
             "additionalProperties": {
                     "type": "string",
                     "minLength": 1,
                     "maxLength": 500
               },
               "description": "Metadados do item segurado: você pode especificar até 50 chaves, com nomes de até 40 caracteres e valores de até 500 caracteres.",
                "maxItems": 50
        },
}
```

hwkr:
@Amorim33, testing it out on main it should look something like this I think (which seems about right)

Google Chrome-2024-04-22-19-29-27@2x

Here's the spec file I'm using:

```json
{
  "openapi": "3.1.0",
  "info": {
    "title": "Hello World",
    "version": "1.0.0"
  },
  "components": {
    "schemas": {
      "metadata": {
        "type": "object",
        "nullable": true,
        "additionalProperties": {
          "type": "string",
          "minLength": 1,
          "maxLength": 500
        },
        "description": "Metadados do item segurado: você pode especificar até 50 chaves, com nomes de até 40 caracteres e valores de até 500 caracteres.",
        "maxItems": 50
      }
    }
  },
  "paths": {
    "/": {
      "get": {
        "requestBody": {
          "content": {
            "application/json": {
              "schema": {
                "$ref": "#/components/schemas/metadata"
              }
            }
          }
        }
      }
    }
  }
}
```

hwkr:
@Amorim33 ahhh I see, opened a fix in #1499

hanspagel (approved):
Oh, man what a great PR! It seems like we fixed the bugs already, so this PR is only the tests.

I love tests, and it would be so great to have tests for more components. The relationship between me and jsdom is complex to say the least, but I’d say let’s just merge it.

If jsdom becomes a problem, we can re-evaluate.

What do you think @amritk?

We should just remove the changeset before/when merging.

amritk (approved):
Changelog removed lets go

marclave (approved):
AWESOME work @Amorim33 ✨

PR #1475, opened 2024-04-22
feat: use http as default scheme for Swagger 2.0 specifications, fix #1218

### LINKED ISSUE #1218: Use default scheme
The project we have is using old spring boot version in which `schemes` is not present, because of which the correct request URL is not getting generated.

Can't we just ignore that and use the default HTTP scheme in that case?

_Originally posted by @chauhan17nitin in https://github.com/scalar/scalar/discussions/1213_

hanspagel:
Swagger 2.0 has a different syntax for server URLs. It’s using a `schemes` array to provide a list of available http schemes (http, https).

Currently, we expect it to be present. Turns out, some files don’t have it.

This PR makes it optional.

**Example (with schemes)**

```json
{
  "swagger": "2.0",
  "info": {
    "version": "1.0.7",
    "title": "Swagger Petstore",
  },
  "host": "petstore.swagger.io",
  "schemes": [
    "https",
    "http"
  ],
  "paths": {}
}
```

Result: `https://petstore.swagger.io`

**Example (without schemes)**

```json
{
  "swagger": "2.0",
  "info": {
    "version": "1.0.7",
    "title": "Swagger Petstore",
  },
  "host": "petstore.swagger.io",
  "paths": {}
}
```

Result: `http://petstore.swagger.io`

### DIFF
--- a/.changeset/thin-knives-relax.md
+++ b/.changeset/thin-knives-relax.md
@@ -1,0 +1,5 @@
+---
+"@scalar/api-reference": patch
+---
+
+feat: use http as default scheme for Swagger 2.0 specifications
--- a/packages/api-reference/src/components/Content/Content.vue
+++ b/packages/api-reference/src/components/Content/Content.vue
@@ -48,14 +48,13 @@ watch(
 
     if (parsedSpec.servers && parsedSpec.servers.length > 0) {
       servers = parsedSpec.servers
-    } else if (
-      props.parsedSpec.host &&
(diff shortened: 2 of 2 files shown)


marclave (approved):
nice fix @hanspagel! ✨ I didnt know 2.0 had it optional :)

hanspagel:
Even if it’s optional, people just have all sorts of specifications so we need to make the best out of it. :)

PR #1477, opened 2024-04-22
feat: configure whether to fetch the default fonts, fix #1471

### LINKED ISSUE #1471: Swagger loads fonts from google
### What happens?

https://github.com/elysiajs/elysia-swagger/issues/104

as said here the swagger interface loads some google fonts

To not load external fonts and remain a completely local interface

See code in linked issue

### Swagger/OpenAPI Example

_No response_

hanspagel:
Currently, we’re fetching Inter and JetBrains Mono from Google fonts - always. Even if different fonts are used or people just don’t want to use Google Fonts (see https://github.com/elysiajs/elysia-swagger/issues/104)

This happens because the CSS imports are just part of our base styles.

This PR separates the fonts from the base styles, and adds a setting to not fetch the default fonts from Google Fonts. 

Great for perf (can save 67kb, if the fonts are not used anyway), great for privacy. Enables users to self-host the fonts, or use something else, or just use the default system font stack.

```vue
<ApiReference :configuration="{
  withoutDefaultFonts: true
} />
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -519,6 +519,8 @@ Overwrite our CSS variables. We won’t judge.
 }
 ```
 
+> Note: By default we’re using Inter and JetBrains Mono, served by Google Fonts. If you use a different font or just don’t want to use Google Fonts, pass `withoutDefaultFonts: true` to the configuration.
+
 You can [use all variables](https://github.com/scalar/scalar/blob/main/packages/themes/src/base.css) available in the base styles as well as overwrite the color theme.
 
 To build your own color themes overwrite the night mode and day mode variables. Here are some basic variables to get you started:
--- a/README.md
+++ b/README.md
@@ -519,7 +519,7 @@ Overwrite our CSS variables. We won’t judge.
 }
 ```
 
-> Note: By default we’re using Inter and JetBrains Mono, served by Google Fonts. If you use a different font or just don’t want to use Google Fonts, pass `withoutDefaultFonts: true` to the configuration.
+> Note: By default we’re using Inter and JetBrains Mono, served by Google Fonts. If you use a different font or just don’t want to use Google Fonts, pass `withDefaultFonts: true` to the configuration.
 
 You can [use all variables](https://github.com/scalar/scalar/blob/main/packages/themes/src/base.css) available in the base styles as well as overwrite the color theme.
 
--- a/packages/api-reference/src/components/ApiClientModal.vue
+++ b/packages/api-reference/src/components/ApiClientModal.vue
@@ -63,9 +63,11 @@ const isMobile = useMediaQuery('(max-width: 1000px)')
             </Sidebar>
           </div>
         </template>
+        <!-- Fonts are fetched by @scalar/api-reference already, we can safely set `withoutDefaultFonts: true` -->
         <ApiClient
           :proxyUrl="proxyUrl"
           theme="none"
+          withoutDefaultFonts
           @escapeKeyPress="hideApiClient" />
       </div>
     </div>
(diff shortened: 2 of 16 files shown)


marclave (approved):
AWESOME work @hanspagel 

lets get @hwkr to review this also, and we can get it merged after they review ✨

packages/api-reference/src/components/ApiReferenceLayout.vue
```diff
@@ -307,9 +309,11 @@ hideModels.value = props.configuration.hideModels ?? false
           </div>
         </template>
         <!-- REST API Client Overlay -->
+        <!-- Fonts are fetched by @scalar/api-reference already, we can safely set `withoutDefaultFonts: true` -->
         <ApiClientModal
           :parsedSpec="parsedSpec"
-          :proxyUrl="configuration?.proxy">
+          :proxyUrl="configuration?.proxy"
+          withoutDefaultFonts>
```
hwkr, line 316: Is this actually a prop on this component?

packages/themes/src/components/ThemeStyles.vue
```diff
@@ -1,12 +1,20 @@
...
+    id?: ThemeId
+    withoutDefaultFonts?: boolean
+  }>(),
+  {
+    withoutDefaultFonts: false,
+  },
+)
 </script>
 <template>
+  <DefaultFonts v-if="!withoutDefaultFonts" />
```
hwkr, line 17: Does this work? I was under the impression CSS imports always get imported if the component is imported anywhere regardless of whether or not it's rendered.

packages/api-reference/src/components/ApiClientModal.vue
```diff
@@ -63,9 +63,11 @@ const isMobile = useMediaQuery('(max-width: 1000px)')
             </Sidebar>
           </div>
         </template>
+        <!-- Fonts are fetched by @scalar/api-reference already, we can safely set `withoutDefaultFonts: true` -->
```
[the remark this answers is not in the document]
hanspagel: But if api reference does not load the fonts they would be loaded through the client otherwise. :)

or we pass down the config, but it’s easier to only load them once in that context.

packages/api-reference/src/components/ApiReferenceLayout.vue
hanspagel: Oops, probably not (yet). Need to take a look

packages/themes/src/components/ThemeStyles.vue
hanspagel: Works locally, but could be true for the build. I’ll check that.

packages/api-reference/src/components/ApiClientModal.vue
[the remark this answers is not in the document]
hwkr: Oh I see, makes sense!

packages/themes/src/components/ThemeStyles.vue
hanspagel: Really good find, it doesn’t work. :)

README.md
```diff
@@ -519,6 +519,8 @@ Overwrite our CSS variables. We won’t judge.
 }
 ```
 
+> Note: By default we’re using Inter and JetBrains Mono, served by Google Fonts. If you use a different font or just don’t want to use Google Fonts, pass `withoutDefaultFonts: true` to the configuration.
```
hwkr, line 522: This still says `withoutDefaultFonts` 😅

hwkr (changes requested):
Just the docs update I think

README.md
hanspagel: Ha, how could I miss that!? 😬

PR #1484, opened 2024-04-22
fix: getting started themes

antlio:
**Problem**
theme from starting section are displaying raw name value

image

**Solution**
this displays theme name right format + fix last theme excess border / hover radiuses

image

### DIFF
--- a/packages/api-reference/src/components/GettingStarted.vue
+++ b/packages/api-reference/src/components/GettingStarted.vue
@@ -329,7 +329,10 @@ function handleEmitPetstore() {
 .start-section-colors .start-item:not(:last-of-type) {
   border-right: none;
 }
-.start-section-colors .start-item:not(:nth-of-type(3n)) {
+.start-section-colors .start-item:last-of-type {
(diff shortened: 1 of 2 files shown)


amritk (approved):
Wicked first official PR!

packages/api-reference/src/components/GettingStarted.vue
```diff
@@ -28,6 +28,20 @@ const themeIds: ThemeId[] = [
   'deepSpace',
 ]
 
+const themeNames: Record<ThemeId, string> = {
```
hwkr, line 31: I would be sweet to actually put these in `@scalar/themes` so we could use them everywhere 😍

hwkr (approved):

packages/api-reference/src/components/GettingStarted.vue
antlio: will do!

marclave (approved):
ah this is much better :) nice fix @antlio

PR #1485, opened 2024-04-22
fix: sidebar models collapse

antlio:
**Problem**
we cannot collapse models section passed the third elements:

https://github.com/scalar/scalar/assets/14966155/94d109d9-a75d-4561-a1c9-22566e87b066

**Solution**
this pr is a proposal to follow a used waiting logic to avoid intersection observer to be active: 

https://github.com/scalar/scalar/assets/14966155/63c8d355-c0f1-4493-baec-87f35cfe164c

### DIFF
--- a/.changeset/early-tomatoes-argue.md
+++ b/.changeset/early-tomatoes-argue.md
@@ -1,0 +1,5 @@
+---
+"@scalar/api-reference": patch
+---
+
+fix: sidebar models collapse


amritk (approved):
Yep we use this log everywhere, nice one!

marclave (approved):
soaring out the gates! nice fix @antlio ✨

PR #1490, opened 2024-04-23
feat: add `@scalar/api-reference-editor`

geoffgscott:
This adds in an editable version of the references (ESM only). 

You can import the vue component OR mountApiReferenceEditable to create an editable version of the references. For CDN delivery we may want to consider bundling Vue with the package. Otherwise import maps would have to be used (which is also well supported). 

PR also includes a very light refactor to the way Layouts are imported into APIReference. This provides a single component that can be used.


packages/api-reference-editor/src/components/ApiReferenceEditor.vue
```diff
@@ -0,0 +1,166 @@
...
+  </component>
+  <Layouts
+    :configuration="configuration"
+    :isDark="isDark"
+    :parsedSpec="parsedSpec"
+    :rawSpec="rawSpec"
+    @toggleDarkMode="() => toggleDarkMode()"
+    @updateContent="handleContentChange">
+    <template #editor>
+      <!--  -->
```
hwkr, line 152: Intentional empty comment?

packages/api-reference-editor/src/components/ApiReferenceEditor.vue
```diff
@@ -0,0 +1,166 @@
...
+    </template>
+    <!-- Expose the content end slot as a slot for the footer -->
+    <template #content-end><slot name="footer" /></template>
+  </Layouts>
+</template>
+<style>
+body {
+  margin: 0;
+}
+</style>
```
hanspagel, line 166: That’s part of the theme anyway, isn’t it? I think we can remove it.

packages/api-reference-editor/test/main.ts
```diff
@@ -0,0 +1,42 @@
...
+  { useExternalState: true },
+  (v: string) => {
+    updateSpecValue(v)
+    console.log('input')
+  },
+)
+
+setTimeout(
+  () =>
+    updateSpecValue(`openapi: 3.0.3
```
hanspagel, line 14: Maybe we can use the new Scalar galaxy example here, usage examples here:

https://github.com/scalar/scalar/pull/1539 (was merged just now)

hanspagel:
Did a rebase, added a demo and some minor fixes. Needs some clean up to get CI green, but it’s pretty close I’d say. ✌️

PR #1491, opened 2024-04-23
style: getting started tables

antlio:
**Problem**
active start item border misalignment, hover wrong radiuses and background lacking accessibility.
image

**Solution**
this pr is a proposal approach on the active start item style + re-use call to action hover background color for start item.
image


marclave (approved):
rad work @antlio !! much more `a l i g n m e n t`

PR #1493, opened 2024-04-23
chore: deploy express example

tmastrom:
Deploy the express api reference example using Docker and CloudBuild. This PR also updates the api reference to fetch the scalar-galaxy spec from the CDN to avoid issues in Docker with relative path imports

### DIFF
--- a/examples/express-api-reference/Dockerfile
+++ b/examples/express-api-reference/Dockerfile
@@ -41,8 +41,4 @@ WORKDIR /app/examples/express-api-reference
 
 USER node
 
-EXPOSE 5050
-ENV PORT 5050
-ENV HOSTNAME "0.0.0.0"
-
 CMD ["dumb-init", "node", "dist/index.js"]
--- a/examples/express-api-reference/Dockerfile
+++ b/examples/express-api-reference/Dockerfile
@@ -1,0 +1,49 @@
+ARG BASE_IMAGE
+FROM ${BASE_IMAGE} AS builder
+WORKDIR /app
+
+COPY pnpm-lock.yaml .
+COPY pnpm-workspace.yaml .
+COPY package.json .
+COPY tsconfig.json .
+COPY packages/api-reference ./packages/api-reference
+COPY packages/components ./packages/components
+COPY packages/themes ./packages/themes
+COPY packages/oas-utils ./packages/oas-utils
+COPY packages/build-tooling ./packages/build-tooling
+COPY packages/express-api-reference ./packages/express-api-reference
+COPY examples/express-api-reference ./examples/express-api-reference
+
+RUN --mount=type=cache,id=pnpm,target=~/.pnpm-store \
+    pnpm --filter @scalar-examples/express-api-reference install --frozen-lockfile && \
+    pnpm --filter @scalar-examples/express-api-reference build
+
+
+FROM node:18-bullseye-slim AS runner
+RUN apt-get update && apt-get install -y --no-install-recommends dumb-init
+
+RUN npm install -g pnpm
+ENV NODE_ENV production
+
+WORKDIR /app
+
+# Copy root node modules and any utilized packages
+COPY --from=builder --chown=node:node /app/node_modules /app/node_modules
+COPY --from=builder --chown=node:node /app/packages/components /app/packages/components
+COPY --from=builder --chown=node:node /app/packages/themes /app/packages/themes
+COPY --from=builder --chown=node:node /app/packages/oas-utils /app/packages/oas-utils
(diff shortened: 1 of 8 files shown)


examples/express-api-reference/Dockerfile
```diff
@@ -0,0 +1,49 @@
...
+FROM node:18-bullseye-slim AS runner
+RUN apt-get update && apt-get install -y --no-install-recommends dumb-init
+
+RUN npm install -g pnpm
+ENV NODE_ENV production
+
+WORKDIR /app
+
+# Copy root node modules and any utilized packages
+COPY --from=builder --chown=node:node /app/node_modules /app/node_modules
```
geoffgscott, line 31: Can we just chown the /app directory here? These copy commands are very verbose.

examples/express-api-reference/src/index.ts
```diff
@@ -49,6 +49,7 @@ app.use(
 )
 
 // Listen
-app.listen(5055, () => {
-  console.log('💻 Express listening on http://localhost:5055')
+const PORT = 8080
```
geoffgscott, line 52: This is a bad port to use. Very common in old webpack apps

examples/express-api-reference/Dockerfile
```diff
@@ -0,0 +1,49 @@
...
+COPY --from=builder --chown=node:node /app/packages/api-reference /app/packages/api-reference
+COPY --from=builder --chown=node:node /app/packages/express-api-reference /app/packages/express-api-reference
+COPY --from=builder --chown=node:node /app/examples/express-api-reference /app/examples/express-api-reference
+
+WORKDIR /app/examples/express-api-reference
+
+USER node
+
+# Cloud build requires port 8080 to be exposed
+EXPOSE 8080
```
geoffgscott, line 45: I don't think this is right.

examples/express-api-reference/Dockerfile
```diff
@@ -0,0 +1,49 @@
...
+ENV PORT 8080
```
geoffgscott, line 46: PORT is generally supplied by cloud run not by the Dockerfile

examples/express-api-reference/src/index.ts
geoffgscott: For deployment this should likely be loaded by a .env file and not hardcoded

geoffgscott (changes requested):

examples/express-api-reference/src/index.ts
tmastrom: Generally I agree but is this really necessary for a small example app?

packages/api-reference/src/components/GettingStarted.vue
```diff
@@ -2,18 +2,19 @@
...
   theme: ThemeId
 }>()
-
 const emits = defineEmits<{
   (e: 'changeTheme', value: ThemeId): void
   (e: 'loadSwaggerFile'): void
   (e: 'linkSwaggerFile'): void
   (e: 'updateContent', value: string): void
 }>()
+const placeholderSpecificationRes = await fetch(
```
tmastrom, line 14: fetch the placeholder from the CDN. I was having issues with this relative import

examples/express-api-reference/Dockerfile
tmastrom: Are you suggesting copy all the files and then `chown` the directory? Setting the owner during the copy seems to be best practice although slightly verbose.

geoffgscott: You should be able to chown the directory then copy the files

examples/express-api-reference/src/index.ts
geoffgscott: Yeah let's just leave the port alone and in the .env file. This is in alignment with all of our other projects in both repos.

examples/express-api-reference/Dockerfile
geoffgscott: Or do it after. I don't see any benefit to this approach from a security standpoint or otherwise. Just very verbose.

packages/api-reference/src/components/GettingStarted.vue
```diff
@@ -2,18 +2,19 @@
...
+  'https://cdn.scalar.com/spec/scalar-galaxy-3.1.yaml',
+)
+const placeholderSpecification = await placeholderSpecificationRes.text()
```
geoffgscott, line 17: Nitpick but a promise chain would be tidier here. 
```ts
await fetch('https://cdn.scalar.com/spec/scalar-galaxy-3.1.yaml').then(r => t.text())
```

geoffgscott (approved):

PR #1494, opened 2024-04-23
chore: Make y-codemirror external and optional

geoffgscott:
Should fully externalize y-codemirror and yjs from the use-codemirror package


marclave (approved):
nice work @geoffgscott !

PR #1501, opened 2024-04-23
refactor: move theme labels to @scalar/themes

hwkr:
Just exposes the theme labels object so we can use it elsewhere


marclave (approved):
nice work @hwkr

PR #1503, opened 2024-04-23
refactor: remove @layer from base theme files

hwkr:
So I wasn't thinking about this at the time but we can't have the themes used in the docs site have the `@layer` declaration (at least in the theme editor). This makes the layer an option that is inject by `getThemeById` and adds some testing for it.


geoffgscott (approved):

marclave (approved):
nice @hwkr ✨

PR #1514, opened 2024-04-24
fix: handle circular references in code blocks, fix #1511

### LINKED ISSUE #1511: Response schema not rendered when it has circular type
### What happens?

When using a circular type in response objects, the `Show Schema [ ]` button on a response will not show the response schema. If the circular type is removed, it will render correctly.

I know it's really pushing the limits of API definitions, however it seems seems to handle the recursion fine for the Models sections. I think the issue here is that it should be rendered to a JSON object, and the recursion means that it will never finish construction the JSoN (as there's always another layer)

This is also given as error when i click "Show Schema":

```
Unhandled Runtime Error
TypeError: Converting circular structure to JSON
    --> starting at object with constructor 'Object'
    |     property 'anyOf' -> object with constructor 'Object'
    |     index 3 -> object with constructor 'Object'
    --- property 'additionalProperties' closes the circle
```

I expect the schema in the response to render, maybe using a `$ref` for circular types, or showing something like an object where the value is a comment noting it's a circular type.

1. Go to https://sandbox.scalar.com/e/ub9yy and look at the response type of example endpiont. Click on `Show Schema [ ]` and see that nothing shows. If you remove the circular type (e.g. remove `JsonObject` from `JsonValue`, it will render correctly. 

### Swagger/OpenAPI Example

https://sandbox.scalar.com/e/ub9yy

hanspagel:
Currently, passing an object with circular references to `ScalarCodeBlock` throws an error as reported in #1511.

This is because we’re just running `JSON.stringify` and it doesn’t know what to do with the circular references.

This PR adds handling for circular references to `ScalarCodeBlock`. I’d love to add a storybook example, but storybook doesn’t deal with JS objects with circular references and throws an error. Which took me way too long to find out. 🤡

### DIFF
--- a/.changeset/lemon-spies-march.md
+++ b/.changeset/lemon-spies-march.md
@@ -1,0 +1,5 @@
+---
+"@scalar/components": patch
+---
+
+fix: circular references in code blocks throw errors
--- a/packages/components/src/components/ScalarCodeBlock/ScalarCodeBlock.vue
+++ b/packages/components/src/components/ScalarCodeBlock/ScalarCodeBlock.vue
@@ -160,6 +160,7 @@ onServerPrefetch(async () => {
 // Here we overwrite the SSR with client rendered syntax highlighting
 onMounted(async () => {
   // This bit async autoloads any syntax we have not pre-loaded
+  // @ts-expect-error This is a dynamic import
   await import('prismjs/plugins/autoloader/prism-autoloader.js')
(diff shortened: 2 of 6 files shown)


marclave (approved):
wow, nice fix @hanspagel this is great! we can add storybook down the line ✨ 

make a ticket in linear? : )

PR #1517, opened 2024-04-24
chore: remove date from townhall event

hanspagel:
Currently, I try to put the next town hall event date in the README and keep it up to date.  But it’s tedious.

Let’s just change the wording. :)


marclave (approved):
we have discord.gg/scalar now :O 

not sure if that helps out at all ✨

PR #1520, opened 2024-04-24
feat: self host api-reference fonts on scalar CDN

tmastrom:
This PR hosts default api-reference fonts on the scalar CDN. This is currently a drop-in replacement but further optimizations can be made by removing unused alphabets and selectively fetching based on the selected language.


hanspagel (changes requested):
Thanks for the PR! Any chance we can make them `.css` files? That’s what they are, also I assume the mimetype will be fixed then (currently text/plain, should be text/css).

tmastrom:
> This looks great! Love how clean the CDN URL is now. Just one thing: Both CSS files still point to Google, for example:
> 
> https://fonts.gstatic.com/s/jetbrainsmono/v18/tDbY2o-flEEny0FZhsfKu5WU4zr3E_BX0PnT8RD8yKxTN1OVk6OThhvAWV8.woff2
> 
> We need to change this to be an URL on our CDN. We don’t want to hit the Google Fonts servers anymore and their URL with the cache hash might change and break in the future.

Good idea. We will now have our own font css files and host the .woff2 font files on the CDN. This should reduce the number of requests necessary 😄

hanspagel (approved):
So clean, so good. Good bye, Google Fonts! :)

amritk (approved):

marclave (approved):

hwkr (changes requested):
Looks like were only hosting and setting up the regular font weight (400) so bold it just being interpolated by the browser (poorly) off of the regular font. 

### This branch (left) vs main (right)

hanspagel:
So great you spotted this! 🙌

hwkr (approved):

marclave:
we are missing some fonts here and weights right @cameronrohani ?

marclave (changes requested):
we are missing some weights right @cameronrohani ?

hwkr:
> we are missing some weights right @cameronrohani ?

Shouldn't be once we get the variable width font uploaded to the CDN, you'll be able to set any font weight from 100 → 900 (including in between weights like 421 or something), we only use 400, 500 and 600 through our CSS vars but this gives us the option to change in the future or for people themeing to use other font weights.

From [Google](https://fonts.google.com/knowledge/glossary/variable_fonts):

> Variable [fonts](https://fonts.google.com/knowledge/glossary/font)—or, more specifically, [OpenType](https://fonts.google.com/knowledge/glossary/open_type) Font Variations—are a relatively new font [format](https://fonts.google.com/knowledge/glossary/font_format) introduced in 2016 that allow one font file to contain multiple stylistic variations, and thus break down the strict delineations of “traditional” (static) fonts. [Weight](https://fonts.google.com/knowledge/glossary/weight), [width](https://fonts.google.com/knowledge/glossary/width), [style](https://fonts.google.com/knowledge/glossary/style), [optical size](https://fonts.google.com/knowledge/glossary/optical_sizes), etc. can then be manipulated by the designer or adjusted based on contextual rules.
> There are three main benefits offered by variable fonts:
> * To compress: the unified delivery of multiple [instances](https://fonts.google.com/knowledge/glossary/instance)—such as a set of different weights—in a single font file that is much smaller than the total file size of a collection of individual font files.
> * To express: the control given to users to select axes values directly, such as a custom weight that is between two named instances of weight. For example, between Regular (400) and Medium (500), a weight of 427 might be just the right one to express a design intent perfectly.
> * To finesse: the control given to programs to select axes values indirectly, based on context. For example, to automatically select an [optical size](https://fonts.google.com/knowledge/glossary/optical_sizes) instance using the current [point size](https://fonts.google.com/knowledge/glossary/point_size).

hanspagel:
I think this is ready to be merged, needs your final approvement @marclave. 🙏 Let me know if I can further help with that.

marclave (approved):
tested! let's go!

marclave (changes requested):
wait just thinking if we can add a flag to quickly add google fonts 🤔

tmastrom:
@marclave @cameronrohani @hanspagel I added documentation and an example for using custom fonts 👍. Thanks to @hwkr for the help!

PR #1522, opened 2024-04-25
chore: replace swagger.json examples with openapi.json, remove unused swagger.json

hanspagel:
We’re using the term `swagger.json` (old, sounds proprietary) in a few places, let’s use `openapi.json` (new) instead.

I’ve also found a `swagger.json` file, that isn’t used.


marclave (approved):
openapi is the future! nice clean up @hanspagel :-)

PR #1523, opened 2024-04-25
feat: use Scalar Galaxy example everywhere

hanspagel:
A while ago we’ve created the Scalar Galaxy example specification. This PR adds it as an example, instead of the Swagger Petstore one.

It’s probably not perfect, but let’s just take it from here I‘d say. :)

There’s a minor styling issue though, the [`details` tags don’t have an arrow](https://github.com/scalar/scalar/issues/1524) (or any other visual indication) anymore.

Screenshot 2024-04-25 at 14 52 30

Screenshot 2024-04-25 at 14 47 08


marclave (approved):
to infinity and beyond

PR #1525, opened 2024-04-25
build: update scalar components package icons

antlio:
this pr adds new icons to the scalar components package.


marclave (approved):
nice @antlio ✨

PR #1537, opened 2024-04-26
fix: update ClientSelector styles fix #1536

### LINKED ISSUE #1536: [UI] Click on "more" label does not open client libraries select options
### What happens?

image

Click on "more" label does not open client libraries select options.

When I see cursor: pointer I expected that some interaction happens on click.

Open API reference.

### Swagger/OpenAPI Example

_No response_

soGit:
**Problem**
Currently, when user focus on "more >" text, cursor pointer appears, but select does not open on click/touch.

**Explanation**
That happens because select's parent element has padding

**Solution**
Set select size same as parent

fix #1536


marclave (approved):
awesome contribution @soGit ! ✨

PR #1544, opened 2024-04-26
feat: add image support to tag, operation and webhook descriptions, fix #1543

### LINKED ISSUE #1543: Markdown images are not supported in endpoint description
### What happens?

Hi,

According to this line, `withImages` props is not set (default value is false) : https://github.com/scalar/scalar/blob/2700c5c9ad4ab555ff5404ead0c6874ec61136ae/packages/api-reference/src/components/Content/Operation/EndpointDetails.vue#L18

- OpenAPI V3.1

Markdown images support in endpoints descripions
 
Is it possible to enable image support? or make it configurable ?

With MarkdownRender unit test ?

### Swagger/OpenAPI Example

_No response_

hanspagel:
We support Markdown in a lot of places, but allow images only in the introduction.

This PR adds image support to the tag description, operation description and webhook description.

See #1543 for more context.

### DIFF
--- a/.changeset/large-cameras-remember.md
+++ b/.changeset/large-cameras-remember.md
@@ -1,0 +1,5 @@
+---
+"@scalar/api-reference": patch
+---
+
+feat: add image support to tag, operation and webhook descriptions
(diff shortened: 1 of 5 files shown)


marclave (approved):
this is a picture perfect fix! ✨

Issue #1498, opened 2024-04-23
Double Description when there are additional properties


hwkr:
### What happens?

![Google Chrome-2024-04-23-13-38-48@2x](https://uploads.linear.app/06a32804-030f-45a1-98aa-79d02a076af5/4ff7bd83-5168-46f2-b854-216e2b5a75f1/c15d1bf8-a153-4d5a-85f9-7586b74344d0?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzA2YTMyODA0LTAzMGYtNDVhMS05OGFhLTc5ZDAyYTA3NmFmNS80ZmY3YmQ4My01MTY4LTQ2ZjItYjg1NC0yMTZlMmI1YTc1ZjEvYzE1ZDFiZjgtYTE1My00ZDVhLTg1ZjktNzU4NmI3NDM0NGQwIiwiaWF0IjoxNzE0MTQ3NDUyLCJleHAiOjMzMjg0NzA3NDUyfQ.jYwyAASkx5uh5IEz8hyu8ybAUnKNwI3ODjh1_aTRTBA "=543x")

### What did you expect to happen?

It should probably show the description once?

### How can we reproduce the issue?

Spec:

```json
{
  "openapi": "3.1.0",
  "info": {
    "title": "Hello World",
    "version": "1.0.0"
  },
  "paths": {
    "/": {
      "get": {
        "requestBody": {
          "description": "I am a description",
          "content": {
            "application/json": {
              "schema": {
                "type": "object",
                "properties": {
                  "metadata": {
                    "type": "object",
                    "additionalProperties": {},
                    "description": "I am another description"
                  }
                }
              }
            }
          }
        }
      }
    }
  }
}
```

### Swagger/OpenAPI Example

*No response*

NOTE: It appears the description from the content object is rendered twice

[...]

PR #1545, opened 2024-04-26
fix: additional properties duplicate description

antlio:
**Problem**
as describe in #1498, description are duplicated when there is additional properties + request body description are not displayed.

**Solution**
this pr fixes the duplicate description + introduces the request body description:
image


marclave (approved):
✨ nice fix @antlio ! we are so back

PR #1547, opened 2024-04-26
fix: don't minify theme css

hwkr:
Because we want to use the theme css in our theme editor we need to have the line breaks in there.

## Before

Code-2024-04-26-15-14-37@2x

## After

Code-2024-04-26-15-14-48@2x

### DIFF
--- a/.changeset/wild-peas-fly.md
+++ b/.changeset/wild-peas-fly.md
@@ -1,0 +1,5 @@
+---
+"@scalar/themes": patch
+---
+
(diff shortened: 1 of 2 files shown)


amritk (approved):
Would be nice if we could only unminify the CSS, but not the end of the world

hwkr:
> Would be nice if we could only unminify the CSS, but not the end of the world

This does only unminify CSS, the js around it is still minified

marclave (approved):

PR #1553, opened 2024-04-29
chore: add scalar use toasts back in

antlio:
this pr re-introduces @scalar-usetoasts component + adds vue-sonner to it.


packages/use-toasts/README.md
```diff
@@ -0,0 +1,37 @@
...
+```vue
+// ChildComponent.vue
+<script setup>
+import { useToasts } from '@scalar/use-toasts'
+
+const { toast } = useToasts()
+
+const sendMessage = () => {
+  toast('This is a message from the toaster!', 'success', { timeout: 2000 })
+}
```
geoffgscott, line 35: Can you confirm this is the actual interface? I think it might be different with vue-sonner

antlio: yes that's [how it's being used currently](https://github.com/scalar/scalar/blob/5c5bc45857082ddb2149cf949cc837ab3ad70ec1/packages/api-reference/src/hooks/useClipboard.ts)

hwkr:
This looks good! I wouldn't mind adding some styles to make it look more like the other Scalar stuff. The old toast library had quite a bit of stuff like a loading ring and stuff.

hwkr (approved):
Lgtm!

packages/use-toasts/src/ScalarToast.test.ts
```diff
@@ -0,0 +1,7 @@
+import { describe, expect, it } from 'vitest'
+
+describe('example', () => {
```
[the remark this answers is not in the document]
antlio: added!

PR #1556, opened 2024-04-30
chore: deploy fastify example

tmastrom:
This PR creates a Dockerfile for `@scalar-examples/fastify-api-reference` 
It uses `turbo` to create an optimized docker image. 
It was necessary to move the base tsconfig.json file to it's own package so `turbo` can prune dependencies for a single package/example.  

Todo: 
- [ ] cloudbuild steps 
- [ ] update `@scalar/api-reference` to fetch example spec from the new spec package


geoffgscott:
The TSConfig is an interesting one. I'm open but I'm not clear on the use case.

I don't want to move away from the base image. This was done to reduce CI time. We only ever need to build the whole project once an then each distribution can just copy the files it needs from the base. It's a very significant reduction in overall build time (and cost) and a very low complexity approach.

PR #1557, opened 2024-04-30
feat: add context / dropdown menu to component library

hwkr:
Todo:

- [x] Make tests actually test something
- [x] Extract resize logic into a hook?
- [x] Create checkbox thing in component library? Or maybe it should be in the other repo?

Google Chrome-2024-04-29-19-30-12@2x

### DIFF
--- a/packages/components/src/components/ScalarFloating/useResizeWithTarget.ts
+++ b/packages/components/src/components/ScalarFloating/useResizeWithTarget.ts
@@ -1,11 +1,11 @@
-import { type ComputedRef, type Ref, computed, ref, watch } from 'vue'
+import { type MaybeRefOrGetter, computed, ref, toValue, watch } from 'vue'
 
 type ResizeOptions = {
-  enabled?: Ref<boolean> | ComputedRef<boolean>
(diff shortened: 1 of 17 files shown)


hanspagel:
I just make it a draft until we worked on the remaining tasks if that’s OK. ✌️

packages/components/src/components/ScalarFloating/useResizeWithTarget.ts
```diff
@@ -0,0 +1,35 @@
+import { type ComputedRef, type Ref, computed, ref, watch } from 'vue'
+
+type ResizeOptions = {
+  enabled?: Ref<boolean> | ComputedRef<boolean>
+}
+
+export function useResizeWithTarget(
+  target: Ref<Element | undefined>,
```
geoffgscott, line 8: Best to use `MayebRefOrGetter` in hooks for for flexible composition
This then requires `toValue(target)` when using the value.

geoffgscott (commented):
Looks great once we at the `MaybeRefOrGetter` stuff

geoffgscott (approved):

PR #1559, opened 2024-04-30
fix: swagger 2.0 definitions don’t show up in list of models, fix #1558

### LINKED ISSUE #1558: Models don't show up in swagger version 2
### What happens?

I'm using the following url `https://petstore.swagger.io/v2/swagger.json` and models don't show up.

Models to be rendered 😎

Go to the provided sandbox URL and see 👀

### Swagger/OpenAPI Example

https://sandbox.scalar.com/e/oiEXa

hanspagel:
Currently, Swagger 2.0 definitions don't show up in the list of Models.

We’re looking into `specification.components.schemas` only, but it used to be in `specification.definitions` in Swagger 2.0.

This PR adds a new getter `getModels` to get all models from OpenAPI 3.x and Swagger 2.0 specifications. All direct references to `specification.components.schemas` are replaced with the getter.

See #1558


marclave (approved):
unreal work @hanspagel ! _we are so back_

PR #1560, opened 2024-04-30
fix: theme not selectable in the editor

antlio:
**Problem**
some themes are not selectable from the api reference editor.

**Solution**
this pr introduces the theme label value in the `changeTheme` event in order to base the theme selection on it in the app.


marclave (approved):
nice fix @antlio

PR #1571, opened 2024-05-01
fix: avoid prepending path when running standalone from an html file without a server

amritk:
this was due to loading from a file:// path


marclave (approved):
we are so back

on the path

great find @amritk

PR #1575, opened 2024-05-01
feat: containerize fastify example and deploy with cloudbuild

tmastrom:
Setup fastify api reference, containerize, and add cloudbuild script to deploy example app.


marclave (approved):
@tmastrom can fix up conflicts and we can get this in : )

PR #1576, opened 2024-05-01
refactor: switch to inter variable

hwkr:
Add support for variable width inter font. Fixes DOC-1755

@tmastrom these `woff2` files need to be uploaded to the CDN: [inter-variable.zip](https://github.com/scalar/scalar/files/15180645/inter-variable.zip)

Behold, variableness 

https://github.com/scalar/scalar/assets/6374090/edf2e6fc-f9a0-4206-b054-45c2502ff1b5


tmastrom (approved):
Nice catch, I'll update the files on the CDN

PR #1579, opened 2024-05-01
feat: containerize react example app

tmastrom:
Containerize react example app. Adds Dockerfile and cloudbuild script. Uses `serve` package to serve the static js from the dist/ folder.


geoffgscott (approved):
Looks good.

PR #1583, opened 2024-05-02
docs: use @scalar/galaxy instead of petstore example

hanspagel:
Currently, we’re still using the petstore HTTP url in a lot of places.

This PR switches to the `@scalar/galaxy` jsdelivr URL instead. :)

I’ve also simplified the quickstart example and made a few wording changes to the root README (focusing more on OpenAPI, less on Swagger).


marclave (approved):
literally out of this world 😮‍💨

PR #1584, opened 2024-05-02
chore: use same example page title everywhere

hanspagel:
Currently, we’re using `API Reference` as the page title everywhere.

This PR proposes to use `Scalar API Reference` as the default (can be overwritten through configuration).

People refer to our API reference as ”Scalar”, I think it’s good to use that in the default page title.


marclave (approved):
page titles are so back @hanspagel crushing it :)

PR #1587, opened 2024-05-02
docs: offer to make a custom theme

hanspagel:
I thought it would be fun to offer our help with custom themes.

We do this anyway, why not promote it? :)

Also, I love to show the Windows 98 example to people. This PR adds a link to it in the styling section.


marclave (approved):
love this @hanspagel we should drop my calendar link too 

https://cal.com/scalar

i'll add a theme option :)

PR #1591, opened 2024-05-02
fix: sidebar headings don’t reset

antlio:
When using and then removing a specification from the editor heading entries are still displayed.

This PR properly resets the headings.

BTW The same issue exists for webhooks, but for a different reason. We’ll tackle this in a separate PR.

image


marclave (approved):
great fix :)

Issue #906, opened 2024-01-31
Add multiple levels of headings to the sidebar


hanspagel:
Currently adding markdown headers in spec description results in adding additional items in the sidebar (example spec: https://github.com/outline/openapi/blob/main/spec3.json). However, only one level (the highest which appears in the markdown) of heading is added. It looks like this is made by design: [source](https://github.com/scalar/scalar/blob/584849b5de65fffe123bed2b33d3a53c8ca7dc2b/packages/api-reference/src/components/Content/Introduction/Description.vue#L23C1-L25C66). I think it would be great if we can add at least one more level of heading so it would match the structure of endpoints documentation, where for each tag we can have collapsible section:
<img width="413" alt="image" src="https://github.com/scalar/scalar/assets/17458785/11c5c46f-a53b-4f88-97a0-1f0b5e520f86">

_Originally posted by @SebastianBienert in https://github.com/scalar/scalar/discussions/905_

hanspagel:
Added in #1592

PR #1592, opened 2024-05-02
feat: add multiple levels of headings to the sidebar

hanspagel:
Currently, we’re looking for Markdown headings in the `info.description` field and only add the lowest level (eg. h1) to the sidebar.

This PR adds the 2nd lowest level (eg. h2) to the sidebar, too.

* If there’s only one level of headings, they are all just added to the sidebar.
* If there’s two levels of headings, they will be added to the sidebar. The 2nd lowest level (eg. h2) will be nested under the lowest level (eg. h1).
* If there’s more than two levels (eg. h1, h2, h3), the highest level (eg. h3) will be ignored (for the sidebar).

See #906 for more context.

### DIFF
--- a/.changeset/small-lies-exist.md
+++ b/.changeset/small-lies-exist.md
@@ -1,0 +1,5 @@
+---
+"@scalar/api-reference": patch
+---
+
+feat: show two levels of headings in the sidebar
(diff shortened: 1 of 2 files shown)


marclave (approved):
UNREAL @hanspagel <3

gazugafan:
Incredible! 

One problem I think I just noticed, though... I see the sub-items in the sidebar are linking appropriately, but the IDs they're linking to aren't being rendered in the description. Only the top-level headings are getting an ID via the markdown renderer, so only the top-level links in the sidebar are working. Here's a simple reproduction...
https://jsfiddle.net/gazugafan/xhw8u61a/40/

If you click on the Heading in the sidebar, that works fine. If you click on the Sub-heading, though, the page doesn't scroll anywhere. If you inspect the HTML, you can see that the sidebar Sub-Heading is linking to #description/sub-heading, which makes sense, but there's no matching ID near the sub-heading in the content.

PR #1601, opened 2024-05-03
feat: add events and tests to standalone references

### LINKED ISSUE #1340: Internal style appending continuously in head tag when using CDN link in single page application on navigation.
### What happens?

I integrated scalar API documentation via [CDN link](https://github.com/scalar/scalar?tab=readme-ov-file#from-a-cdn) in angular SPA project. but there is a issue like Internal style appending continuously in head tag in single page application DOM on navigation change. Due to this it will impact on web performance and it's a bad structure of DOM.

Here i am adding the code. Help me to fix this bug.

Here is the component file:

```typescript
import { AfterViewInit, Component, OnDestroy, OnInit, Renderer2 } from '@angular/core';

@Component({
  selector: 'app-api-list',
  standalone: true,
  imports: [],
  templateUrl: './api-list.component.html',
  styleUrls: ['./api-list.component.scss'],
})
export class ApiListComponent implements OnInit, AfterViewInit, OnDestroy {
  constructor() { }

  ngOnInit(): void {
    this.loadScript('https://cdn.jsdelivr.net/npm/@scalar/api-reference').then((res: any) => {
      console.log('res:', res);
    }).catch(error => {
      console.error('Error loading script:', error);
    });
  }

  ngOnDestroy(): void {
  }

  ngAfterViewInit(): void {
  }

  private loadScript(url: string): Promise<void> {
    return new Promise<void>((resolve, reject) => {
      const scriptElement = document.createElement('script');
      scriptElement.src = url;
      scriptElement.onload = () => resolve();
      scriptElement.onerror = (error) => reject(error);
      document.body.appendChild(scriptElement);
    });
  }
}
``` 

Here is the html file:

```html
<div
id="api-reference"
data-url="https://petstore3.swagger.io/api/v3/openapi.json"
data-proxy-url="https://api.scalar.com/request-proxy"></div>
``` 

You can check visually here and suggest me proper solution if i am miss any configuration then also please help me.

https://github.com/scalar/scalar/assets/56481473/527d8cf8-129b-4e2a-ae57-753342da689c

The internal style append should only happen once.

1. Setup CDN in Angular project as i already mentioned code above
2. Setup API documentation page in particular url slug
3. Navigate to another page
4. Go to the existing page on which we loaded the API documentation via CDN link
5. See head tag style in DOM via inspect element in browser

### Swagger/OpenAPI Example

_No response_

amritk:
Added two user-requested events to the standalone app. 

Also added some tests to util, had to use sleep as we are outside of vue and flush promises didn't work. If the sleep becomes a problem I will remove and add a better solution.

Instructions can be found [here](https://github.com/scalar/scalar/blob/321485c913239e75a2466a5eaa3c29ea379b6d52/README.md#events-beta)

closes #1340

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,6 +111,39 @@ If you’d like to add a request proxy for the API client (to avoid CORS issues)
 </script>
 ```
 
+#### Events [beta]
+
+We have recently added two events to the standalone CDN build only.
+
+##### scalar:reload-references
+
+Reload the references, this will re-mount the app in case you have switched pages or the dom
+elements have been removed.
+
+```ts
+document.dispatchEvent(new Event('scalar:reload-references'))
+```
+
+##### scalar:reload-references
+
+If you have updated the config or spec, you can trigger this event with the new payload to update
+the app. It should update reactively so you do not need to trigger the reload event above after.
+
+```ts
+import { type ReferenceProps } from './types'
+
+const ev = new CustomEvent('scalar:update-references-config', {
+  detail: {
+    configuration: {
+      spec: {
+        url: 'https://cdn.scalar.com/spec/openapi_petstore.json',
+      },
+    },
+  } satisfies ReferenProps,
+})
+document.dispatchEvent(ev)
+```
+
 ### With Nuxt
 
 You can easily run Scalar API References in Nuxt via the module:
--- a/README.md
+++ b/README.md
@@ -124,7 +124,7 @@ elements have been removed.
 document.dispatchEvent(new Event('scalar:reload-references'))
 ```
 
-##### scalar:reload-references
+##### scalar:update-references-config
 
 If you have updated the config or spec, you can trigger this event with the new payload to update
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -111,6 +111,39 @@ If you’d like to add a request proxy for the API client (to avoid CORS issues)
...
+##### scalar:reload-references
+
+Reload the references, this will re-mount the app in case you have switched pages or the dom
+elements have been removed.
+
+```ts
+document.dispatchEvent(new Event('scalar:reload-references'))
+```
+
+##### scalar:reload-references
```
hanspagel, line 127: I think the heading should say `scalar:update-references-config`. ✌️

hanspagel (changes requested):
OMG, great PR

Just found a wrong heading, that’s all

README.md
amritk: nice catch!

marclave (approved):
awesome work @amritk !!! this is SO cool

PR #1605, opened 2024-05-03
fix: disable scroll observer on sidebar click

amritk:
Because handleclick wraps the whole sidebar section, `isIntersectionEnabled` was getting enabled early and was bugging out the scroll, causing the wrong sidebar elements to get selected. This PR fixes that.


antlio (approved):
good catch 🙏

marclave (approved):
✨ nice fix @amritk

Issue #920, opened 2024-02-01
Scalar not rendering correctly when using components $ref


rlindskog:
### What happens?

- docs aren't rendering the response bodies when using $ref
- libraries aren't showing the request bodies or query params

### What did you expect to happen?

- I expect the docs to show response bodies when using $ref
- I expect the libraries to show request bodies or query params
- 
### How can we reproduce the issue?

[this sandbox](https://sandbox.scalar.com/e/BNs9P)

Example schema
```json
{
  "components": {
    "schemas": {
      "Message": {
        "properties": {
          "message": {
            "type": "string"
          }
        },
        "required": [
          "message"
        ],
        "type": "object"
      },
      "Todo": {
        "properties": {
          "complete": {
            "type": "boolean"
          },
          "title": {
            "type": "string"
          }
        },
        "required": [
          "complete",
          "title"
        ],
        "type": "object"
      },
      "UpdateTodo": {
        "description": "Todo",
        "properties": {
          "description": {
            "type": "string"
          },
          "done": {
            "type": "boolean"
          }
        },
        "required": [
          "description",
          "done"
        ],
        "type": "object"
      }
    }
  },
  "info": {
    "title": "Test",
    "version": "1.0"
  },
  "openapi": "3.0.0",
  "paths": {
    "/hello": {
      "get": {
        "parameters": [
          {
            "deprecated": false,
            "explode": true,
            "in": "query",
            "name": "name",
            "required": false,
            "schema": {
              "type": "string"
            }
          }
        ],
        "responses": {
          "200": {
            "content": {
              "application/json; charset=utf-8": {
                "schema": {
                  "$ref": "#/components/schemas/Message"
                }
              }
            },
            "description": ""
          }
        },
        "summary": "some docs"
      }
    },
    "/todos": {
      "post": {
        "requestBody": {
          "content": {
            "application/json; charset=utf-8": {
              "schema": {
                "$ref": "#/components/schemas/UpdateTodo"
              }
            }
          },
          "required": true
        },
        "responses": {
          "200": {
            "content": {
              "application/json; charset=utf-8": {
                "schema": {
                  "$ref": "#/components/schemas/Todo"
                }
              }
            },
            "description": "Successul response"
          },
          "404": {
            "description": "Error response"
          }
        },
        "summary": "This is todos"
      }
    }
  },
  "servers": [
    {
      "url": "http://127.0.0.1:3001/api"
    }
  ],
  "tags": []
}
```

[...]

PR #1606, opened 2024-05-03
fix: mimetypes with charsets and variants are ignored, see #920

hanspagel:
Currently, we expect request bodies and example responses to be in `application/json`, variants like `application/json; charset=utf-8` or `application/vnd.api+json` (which are totally valid) are ignored.

That’s why we’re just looking for that string. I’ve added a normalize function, that transforms all those variants to simple mime types like `application/json`.

See #920


marclave (approved):
@hanspagel awesome work, look at all theses tests! <3

PR #1612, opened 2024-05-04
fix: docusaurus README example

x-delfino:
There were a couple of erraneous brackets in the docusaurus package README


marclave (approved):
@x-delfino great find! those brackets are useful if you want to have multiple integrations but you're right we only need one :)

PR #1621, opened 2024-05-06
ci: generate provenance statements

hanspagel:
Provenance statements give a proof that our published packages are what’s in this repository. This is a typical security concern users have. That’s why I suggest to enable it for this repository. :)

Note: It might not work on the first attempt, it’s hard to test without actually releasing. If you like, I can test this on a smaller repository first.

Read more:

> You can generate provenance statements for the packages you publish. This allows you to publicly establish where a package was built and who published a package, which can increase supply-chain security for your packages.

https://docs.npmjs.com/generating-provenance-statements


marclave (approved):
official 😏

PR #1622, opened 2024-05-06
feat: go proxy cloud build deploy

tmastrom:
Dockerfile and cloudbuild files for building and deploying the proxy


hanspagel:
Which domain is that using then? Can we go with `proxy.scalar.com`? I love beautiful names. 😅

PR #1628, opened 2024-05-07
fix: code block broken

hanspagel:
Just adding a ` to fix this broken code block:

Screenshot 2024-05-07 at 13 15 11

And while I’m on it, let’s remove the petstore stuff from the example. :)


marclave (approved):
nice catch ✨

PR #1630, opened 2024-05-07
chore: fetch example specification from jsdelivr

hanspagel:
What if we don’t bundle the @scalar/galaxy example, but fetch it from jsdelivr? This should decrease the bundle size. 🤔


marclave (approved):
galaxy brain idea, nice @hanspagel

hanspagel:
Not even 1% though. 😅

PR #1636, opened 2024-05-07
chore: use prism instead of highlightjs

hanspagel:
Look what I found! We’re using highlight.js for the syntax highlighting in Markdown and it’s huge!

| Package          | Size      |
|----------------- | --------- |
| rehype-highlight | 3.54KiB   |
| highlight.js	   | 354.87KiB |

But we’re also using prismjs for code blocks. What if … we replace highlight.js with prismjs … 🤔 

EDIT: Results are in: 💥 1.77MiB (-6.4%) 💥

### DIFF
--- a/.changeset/eight-penguins-cough.md
+++ b/.changeset/eight-penguins-cough.md
@@ -1,0 +1,5 @@
+---
+"@scalar/api-reference": patch
+---
+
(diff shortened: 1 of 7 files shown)


marclave:
omg i love this

marclave (approved):
OMG the savings 🤯 🤯

PR #1643, opened 2024-05-07
fix: hide default caret on <summary> on Safari

hwkr:
Before:

Safari-2024-05-07-15-54-51@2x

After:

Safari-2024-05-07-15-54-53@2x


marclave (approved):
:)

PR #1646, opened 2024-05-08
chore: lazy load API client

hanspagel:
`@scalar/api-client` is pretty heavy. It’s the only reason we still have CodeMirror (for the request body and the address bar).

@amritk had the wonderful idea to lazy load the API client. This PR does exactly this, and the ESM build improves:

```diff
-dist/index.js 1,650.72 kB
+dist/index.js 1.60 kB
+dist/index-Bw-GQvDT.js 1,314.59 kB
```

This means:
1. The first file that needs to be loaded in ESM envs is 1kB instead of 1,600kB.
2. The biggest chunk is 1,300 kB instead of 1,600 kB.
3. Users that never open the API client, will never load the ~1,500 kB (only applies to ESM envs, like our app).

CDN users don’t benefit from the change yet, because it’s still bundling everything in one file. But at some point (once the ESM CDN build works), they'll experience the same benefits. :)

### DIFF
--- a/.changeset/healthy-pears-build.md
+++ b/.changeset/healthy-pears-build.md
@@ -1,0 +1,5 @@
+---
+"@scalar/api-reference": patch
+---
+
+chore: lazy load api client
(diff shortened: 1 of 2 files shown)


amritk (approved):

marclave (approved):
nothing lazy about these performance gains ✨

marclave (approved):
✨

PR #1647, opened 2024-05-08
feat: include spec title in downloaded spec filename

antlio:
**Problem**
`spec` + extension is currently used as the file name when a spec is downloaded from the reference doc, which can make it difficult to find if using multiple APIs.

**Solution**
this pr is a proposal to include the spec title in the downloaded filename.

### DIFF
--- a/packages/api-reference/src/components/Content/Introduction/Introduction.vue
+++ b/packages/api-reference/src/components/Content/Introduction/Introduction.vue
@@ -1,5 +1,6 @@
 <script setup lang="ts">
 import type { OpenAPIV2, OpenAPIV3, OpenAPIV3_1 } from '@scalar/openapi-parser'
+import GithubSlugger from 'github-slugger'
 import { computed } from 'vue'
 
 import type { Spec } from '../../../types'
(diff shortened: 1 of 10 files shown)


packages/api-reference/src/components/Content/Introduction/Introduction.vue
```diff
@@ -25,6 +25,10 @@ const props = defineProps<{
 const specVersion = computed(() => {
   return props.parsedSpec.openapi ?? props.parsedSpec.swagger ?? ''
 })
+
+const formattedSpecTitle = computed(() => {
+  return props.info.title?.toLowerCase().replace(/\s+/g, '-')
```
hanspagel, line 30: We could also use the github-slugger, it’s in the dependencies already: https://github.com/scalar/scalar/blob/7fb8d1285b8b11503effa7d067ceca5fd0e86176/packages/api-reference/src/helpers/getHeadingsFromMarkdown.ts#L3

Wdyt?

hanspagel (approved):
I love those UX additions so much!!

packages/api-reference/src/components/Content/Introduction/Introduction.vue
antlio: good call!

marclave (approved):
ah! this is SUCH a great idea :)

PR #1648, opened 2024-05-08
chore: add deprecation warnings (still disabled)

hanspagel:
Once we switch to the new proxy, we need to overwrite the old proxy URL (https://api.scalar.com/request-proxy) with the new URL (https://proxy.scalar.com) and warn users about it (affects probably everyone).

And we want to warn if a custom proxy URL is used (shouldn’t affect many people).

This PR adds the code already, but disables the messages until the proxy is ready.

Screenshot 2024-05-08 at 11 37 16
Screenshot 2024-05-08 at 11 34 31


marclave (approved):
🥳 nice work @hanspagel

PR #1649, opened 2024-05-08
fix: start page links broken

hanspagel:
Two example links are broken for the `examples/web` we’re using, this PR fixes them. ✨


marclave (approved):
so much cleanup he can't be stopped @hanspagel

PR #1650, opened 2024-05-08
fix: wrong types export for mock server 

hanspagel:
Oops, the mock server build process was messed up and the types weren't properly exported.

This PR updates the build process and should fix the type export. 👍


marclave (approved):
:) we are so back

PR #1651, opened 2024-05-08
fix(mock-server): openapi.json returns yaml string

hanspagel:
The mock server has an `/openapi.json` endpoint, which isn’t working very well.

[If you pass YAML, /openapi.json returns the YAML string …](https://galaxy.scalar.com/openapi.json)

Anyway, this PR makes it all awesome and even adds an `openapi.yaml` endpoint. :)


marclave (approved):
ooooo this is awesome ✨

PR #1652, opened 2024-05-08
fix: sidebar is empty locally

hanspagel:
Since we removed the editor from @scalar/api-reference the local demo was broken. We’ve re-added Monaco a while ago, but the sidebar was empty. I’m using this a lot to play around with specifications, so I’m eager to finally fix this again. :)

This PR makes the sidebar reactive again, which fixes the local demo (the one with the editor):

http://localhost:5050/api-reference


marclave (approved):
nice @hanspagel ! i hit this yesterday ✨

PR #1654, opened 2024-05-08
docs: Fix Next.js API Spec Link

KiARC:
There was a typo causing it to link to the Nest.js page instead.

**Problem**
Currently, the link labeled "Next.js API Reference" actually links to the Ne**s**t.js API Reference

**Explanation**
This happens because there is an s where there should be an x

**Solution**
With this PR I fixed the typo.


hanspagel:
Great find!!

PR #1657, opened 2024-05-08
fix: base docker builder hits memory limit or something

hanspagel:
CI is broken, the docker build complains about a list of strings and say it can’t analyse it.

image

When I try it locally it bails at another point. I’m pretty confident we’re hitting a memory limit or something here.

With this PR we’re using turbo to build the packages in the Dockerfile and make sure it’s not running more than 2 processes in parallel.

Let’s try if that helps! It does locally.


amritk (approved):
So `build:packages` already calls turbo, but lets try it 

```json
    "build:packages": "turbo --filter './packages/**' build",
```

marclave (approved):

PR #1662, opened 2024-05-09
fix: api client request body display

antlio:
**Problem**
request body is displayed for each http method in the api client, even though it is not useful for each of them.

**Solution**
this pr is a proposal to introduce a condition on the request body display on selected http methods.


marclave (approved):
keeps the ui clean :) nice work @antlio

PR #1668, opened 2024-05-09
feat: add icons from Scalar Docs

hwkr:
Adds a bunch of new icons pulled from docs.scalar.com, not sure if we'll use all of them but it's a starting point.


amritk (approved):
Icons _should_ be lazy loaded so we can add as many as we want! 

Looks like it might be affecting the standalone build size... Anyway thats a fix we can do on the standalone side, feel free to add all the icons!

geoffgscott (approved):

PR #1675, opened 2024-05-09
fix: uppercase required property

cameronrohani:
Why was 'required' so AGGRESSIVE, let's fix that

Before:
image

After:
image


marclave (approved):
nice :)

PR #1677, opened 2024-05-10
fix: drag and drop

amritk:
Went down a rabbit hole of trying to get anchor links to drag properly. Turns out they got their own drag business going on. Just going to switch to divs with click listeners for now


marclave (approved):
🐰

PR #1684, opened 2024-05-10
feat: add api client response empty state

antlio:
this pr introduces the response empty state ascii + style in the api client as seen below:

image


hanspagel:
Alternative wording proposal:

“Fill the void and send your first request.”

probably needs a line break

antlio:
> Alternative wording proposal:
> 
> “Fill the void and send your first request.”
> 
> probably needs a line break

love it! since that screen appears when we change endpoints, what about "Fill the void and send your request" as it might not be the first one?

cameronrohani:
Awesome work! 😍

marclave (approved):
nothing empty about this state 😏 great work @antlio

PR #1688, opened 2024-05-10
docs: replace all mentions of docs.scalar.com

hanspagel:
> ⚠️ The new domain doesn’t work yet. Don’t merge the PR before it does.

Currently, we’re pointing to `docs.scalar.com` in a few places. Let’s change this to `app.scalar.com` (once it works). :)


hanspagel:
Currently not relevant

PR #1692, opened 2024-05-10
fix: vertical alignment of client response empty state

cameronrohani:
v e r t i c a l y aligned
image


marclave (approved):
look at those calcs, @cameronrohani has his phd fr

PR #1694, opened 2024-05-10
feat: api client address bar style

antlio:
this pr updates the api client address bar style as seen below:

image


cameronrohani (approved):
amazing!

marclave (approved):
awesome work @antlio !

PR #1695, opened 2024-05-10
feat: api client toggle sidebar

antlio:
this pr is a proposal to use the `meta + b` shortcut to toggle the api client sidebar visibility.


hanspagel:
Keys.b is just b, isn’t it? It’s keys.meta_b or sth

marclave (approved):
ooo fun, we will have to consolidate all our hotkeys to one central location for 🦇 but this is _slick_ nice addition @antlio

PR #1701, opened 2024-05-11
fix: api client address bar url

### LINKED ISSUE #1687: API client layout is breaking when the URL is too long
API client layout is breaking when the URL is too long

image

_Originally posted by @Amorim33 in https://github.com/scalar/scalar/discussions/1686_

antlio:
**Problem**
as highlighted by @Amorim33 and in #1687, breaking layout when url is too long.

**Solution**
this pr fixes #1687 by allowing scrolling and adding fade to long url.

image

### DIFF
--- a/.changeset/clean-glasses-sip.md
+++ b/.changeset/clean-glasses-sip.md
@@ -1,0 +1,5 @@
+---
+"@scalar/api-client": patch
+---
+
+fix: api client address bar url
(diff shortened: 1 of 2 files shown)


marclave (approved):
woah @antlio nice fix 😎 i saw this happen on the sentry docs friday and forgot to make an issue but seems like @Amorim33 was on it too 💯

PR #1714, opened 2024-05-13
chore: bring back local urls for development

hanspagel:
In #659 we’ve added the URLs of the deployed examples, but I’m missing the local URLs. I love how convenient it is to open a locally running example. 

So why don’t we add both? 😊


marclave (approved):
🕺

PR #1715, opened 2024-05-13
feat: api client modal size

antlio:
**Problem**
the recent update of the api client has reduced the size of the modal in an api reference which may lack covered space, especially in the case where the sidebar is open :

image
 

**Solution**
this pr is a proposal to increase the size of the api client modal, as seen below:

image


antlio:
closing this in favor of #1720

PR #1718, opened 2024-05-13
fix: color single security scheme

adjsky:
**Problem**
Currently, if `securitySchemes` is defined with a single scheme, uncolored security scheme will be displayed in the API client.

Before:

After:

**Explanation**
The issue is... the needed styles weren't applied to text if a single scheme is provided.

**Solution**
Put `.security-scheme-selector span` into a separate class, and use it for scheme labels.

### DIFF
--- a/.changeset/proud-boxes-punch.md
+++ b/.changeset/proud-boxes-punch.md
@@ -1,0 +1,5 @@
+---
+"@scalar/api-client": patch
+---
+
+fix: color single security scheme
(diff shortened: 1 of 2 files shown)


adjsky:
Oh... after opening PR I noticed that `CollapsibleSection` adds extra styles to any span inside options. Plans changed, so I abandoned the previous solution, and decided to put `.security-scheme-selector span` into a separate class, and use it for labels.

marclave (approved):
nice work @adjsky !!

PR #1721, opened 2024-05-13
feat: add more icons and utility

amritk:
- moved httpStatusCodes to `oas-utils`
- added a new size to scalar button
- change height and width to size for square items
- adjusted sizes in icon button + allow overwriting classes


marclave (approved):
good call :)

PR #1729, opened 2024-05-14
feat: hide operations with `x-internal: true`

hanspagel:
As requested in https://github.com/scalar/scalar/discussions/1728, this PR adds the ability to hide some operations with `x-internal: true`.

I’ve added an example to the `@scalar/api-reference` README. While I was on it, I’ve added examples for `x-displayName` and `x-tagGroup` aswell.

I was also able to remove handful of `@ts-expect error` comments.

### DIFF
--- a/.changeset/fuzzy-baboons-yell.md
+++ b/.changeset/fuzzy-baboons-yell.md
@@ -1,0 +1,5 @@
+---
+"@scalar/api-reference": patch
+---
+
+feat: hide operations with x-internal: true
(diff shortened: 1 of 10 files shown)


marclave (approved):
✨ tested and works GREAT, this one is a fantastic extension to add AND we have docs 🤤 

@hanspagel fantastic work as always ✨

mburumaxwell:
Is there a way to make this optional?

Issue #1716, opened 2024-05-13
Nuxt: Broken since 0.0.35


max06:
### What happens?

`The requested module '/_nuxt/@fs/workspaces/nimbus/node_modules/boolbase/index.js?v=c13543d7' does not provide an export named 'default'`

This did not break in 0.0.34.

### What did you expect to happen?

It should not break

### How can we reproduce the issue?

- Have a nuxt app (tested in 3.11.2).
- Install module as documented (and pin using `bun add @scalar/nuxt@0.0.35`)
- Start devserver and render random openapi schema
- Observe app failure

amritk:
Hello @max06 thanks for reporting the issue! We are on the case 🔍 . In the meanwhile, I think version `0.0.33` _should_ work.

max06:
Hey @amritk - thanks for the headsup!

0.0.34 is also working fine, just in case. Based on the commit history 0.0.35 looked like it could cause issues so I started testing at 0.0.34 - lucky hit 🍀

amritk:
hmm strange, because I found another issue in 0.0.34. Could be that the issue is only in dev

max06:
Good to know, I'll keep my eyes open! 🙇🏼

hanspagel:
On it! :)

0.34 gives me the following error:

```
500 The requested module '/_nuxt/node_modules/.pnpm/debug@4.3.4/node_modules/debug/src/browser.js?v=bd279dc8' does not provide an export named 'default'
```

PR #1730, opened 2024-05-14
chore: switch from rehype-prism to @shikijs/rehype

hanspagel:
> ⚠️ Don’t merge yet. WIP

We run into problems with the two prismjs versions we have (see #1716).

Actually, I think [shiki is the most modern syntax highlighting package](https://github.com/shikijs/shiki) (ESM yay) and it would be great to use it. And there’s a shiki editor, which could replace CodeMirror for us: https://shikicode.vercel.app/

Last time I tried it, it added 8 megabytes to the CDN build. But I think we can do better, so let’s try it again.

EDIT: Results are in:

```
dist/browser/standalone.js       10,252.92 kB
```

Let’s see what we can do. :)

### DIFF
--- a/packages/api-reference/package.json
+++ b/packages/api-reference/package.json
@@ -71,6 +71,7 @@
     "@scalar/use-modal": "workspace:*",
     "@scalar/use-toasts": "workspace:*",
     "@scalar/use-tooltip": "workspace:*",
+    "@shikijs/rehype": "^1.5.2",
     "@unhead/schema": "^1.9.5",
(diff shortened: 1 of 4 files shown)


amritk:
heyyy I saw this too when I was debugging this 😄 and was thinking we could check it out!

Val-istar-Guo:
shiki is better  😁

hanspagel:
Shiki is awesome, modern and powerful, but the core is 800kb already. I get it down to just that, but this is still too much to ship. No shiki for us. :(

PR #1735, opened 2024-05-14
feat: add Dockerfile and README

captainsafia:
- Add Dockerfile for running .NET sample locally
- Add .gitignore to sample for .NET-related files
- Add README with build and run instructions


examples/DotNet/Program.cs
```diff
@@ -1,9 +1,3 @@
-using Microsoft.AspNetCore.Authentication;
```
captainsafia, line 1: Some of these using were not require for the specific sample (Microsoft.AspNetCore.Authentication was not needed since no authentication was confirmed).

For others, like `Microsoft.AspNetCore.Builder` we're taking advantage of ASP.NET Core's implicit usings feature to trim down the number of imports we're using.

marclave (approved):
@captainsafia WOW, thanks so much for adding the docker image + adding docs ✨ ✨ ✨ ✨

PR #1737, opened 2024-05-15
feat: add official Scalar .NET integration

marclave:
we woke up today to a wonderful guide written on how to add Scalar to .NET with the anticipation of the amazing new 9.0 release with OpenAPI support out of the box!

@captainsafia wrote the wonderful docs and the integration for Scalar, now its bundled and available as a package so we can add more functionality with config, we also added it to the root README.

some things left to do:
- [ ] custom CDN
- [ ] custom page title
- [ ] maybe a generic to handle the scalar config object?


packages/scalar.aspnetcore/Scalar.AspNetCore.csproj
```diff
@@ -0,0 +1,14 @@
...
+
+  <PropertyGroup>
+    <TargetFramework>net9.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+    <Version>1.0.1</Version>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.AspNetCore.OpenApi" Version="9.0.0-preview.4.24260.3" />
```
captainsafia, line 11: You shouldn't need a package reference to the `Microsoft.AspNetCore.OpenApi` package in `Scalar.AspNetCore` since you don't rely on any of the APIs in it. Instead, we can assume that the use has installed the package in their web project and exposed the OpenAPI endpoint the Scalar API will point to from there.

packages/scalar.aspnetcore/Scalar.AspNetCore.cs
```diff
@@ -0,0 +1,36 @@
...
+            return endpoints.MapGet("/scalar/{documentName}", (string documentName) => Results.Content($$"""
+            <!doctype html>
+            <html>
+            <head>
+                <title>Scalar API Reference -- {{documentName}}</title>
+                <meta charset="utf-8" />
+                <meta name="viewport" content="width=device-width, initial-scale=1" />
+            </head>
+            <body>
+                <script id="api-reference" data-url="/openapi/{{documentName}}.json"></script>
```
captainsafia, line 21: We provide an option for users to customize the route their OpenAPI document will be served at.

```
app.MapOpenApi(); // use default route of `/openapi/{{documentName}}.json`
app.MapOpenApi("/{{documentName}}/openapi.json");
```

We'll want to consume the data URL in the config object. I assume that will come later.

captainsafia (approved):
Exciting! Left some comments inline but looks good to me overall!

marclave:
awesome! i added some github issues so we can fast follow and get this in! ✨ 

https://github.com/scalar/scalar/issues/1738
https://github.com/scalar/scalar/issues/1739
https://github.com/scalar/scalar/issues/1740

renerlemes:
@marclave is possible to use with dotnet 8?

marclave:
> @marclave is possible to use with dotnet 8?

yes you can use it with version 8 in the latest nuget release ✨

renerlemes:
I was able to use it on dotnet 8, but I also needed to use Swagger to generate the open API json file. Is it possible to use without Swagger?

How do I generate the json file?

captainsafia:
@marclave I think to support this fully in .NET 8, the Scalar.AspNetCore package would need to support a way to set the data-url here:

https://github.com/scalar/scalar/blob/e2905ac1aba39f3a9542349d5137933b8178b3bd/packages/scalar.aspnetcore/Scalar.AspNetCore.cs#L45

So that we can point it at the OpenAPI file created in packages other than Microsoft.AspNetCore.OpenApi (which is only supported in .NET 9).

marclave:
@captainsafia @renerlemes we added the options to the package but we need to add some documentation 

https://github.com/scalar/scalar/blob/main/packages/scalar.aspnetcore/ScalarOptions.cs

I will add some today :) but you can pass in the data url ✨

renerlemes:
@marclave and @captainsafia thanks for support. I´ll try to use again.

With this change they made, will I still need the Swagger package to generate the Open API File and pass in the data url?

When you do the documentation, if you can inform me here and I will do new tests.

steingran:
@renerlemes Did you ever figure out how to use this in .NET 8? I am trying to do that as well, and am hitting a brick wall so far

captainsafia:
> With this change they made, will I still need the Swagger package to generate the Open API File and pass in the data url?

Yes, you'll still need to use a separate package to generate the OpenAPI document that Scalar is consuming. If you're targeting .NET 8 or below, you can use Swashbuckle or NSwag to generate the packages. If you're using .NET 9, you can use the support in the built-in Microsoft.AspNetCore.OpenApi package (see [here](https://learn.microsoft.com/en-us/aspnet/core/fundamentals/minimal-apis/aspnetcore-openapi?view=aspnetcore-9.0&tabs=visual-studio)).

> @renerlemes Did you ever figure out how to use this in .NET 8? I am trying to do that as well, and am hitting a brick wall so far

If you're running into issues that are more specific to .NET, feel free to file an issue on the dotnet/aspnetcore repo and I'd be happy to help.

steingran:
Thanks @captainsafia! The issue I run into is with Scalar. I cannot set the data-url, which is "/swagger/v1/swagger.json". The pattern for this in Scalar seems to be (as you pointed out in an earlier post) as specified here: https://github.com/scalar/scalar/blob/e2905ac1aba39f3a9542349d5137933b8178b3bd/packages/scalar.aspnetcore/Scalar.AspNetCore.cs#L45
And that data-url does not match my pattern at all. Can we set the whole data-url using the ScalarOptions.cs class?

captainsafia:
@steingran Ah, you're right.

@marclave Are you open to making the data-url customizable as well? If so, I'd be happy to open a PR.

marclave:
> @steingran Ah, you're right.
> 
> @marclave Are you open to making the data-url customizable as well? If so, I'd be happy to open a PR.

absolutely open! ✨

HalfToothed:
@marclave we should include how to use Scaler with .NET 8 in the documentation, i myself had very hard time figuring out how to use Scaler with .NET 8.
very few people use .NET 9 preview so i would be happy to work on this issue

marclave:
> @marclave we should include how to use Scaler with .NET 8 in the documentation, i myself had very hard time figuring out how to use Scaler with .NET 8. very few people use .NET 9 preview so i would be happy to work on this issue

hi @HalfToothed ! that would be amazing! Were the main friction points around a lack of documentation?

HalfToothed:
Yes @marclave  in documentation this is the only example which shows how to integrate scaler with .NET. 
Screenshot 2024-08-12 083259

we can provide another example using .NET 8
```
var builder = WebApplication.CreateBuilder(args);

builder.Services.AddControllers();

builder.Services.AddEndpointsApiExplorer();
builder.Services.AddSwaggerGen();

var app = builder.Build();

if (app.Environment.IsDevelopment())
{
    app.UseSwagger();
    app.UseScalar();
}

app.MapControllers();

app.Run();
```

captainsafia:
@marclave I can help you add additional documentation on this if you'd like.

This is mostly about showing people how to use the new data-url options I just added in https://github.com/scalar/scalar/pull/2722 since some OpenAPI packages in .NET host their JSON file at a different endpoint in the app.

HalfToothed:
@captainsafia ya sure, i should create a new issue for updating the docs of .NET only ?

PR #1743, opened 2024-05-15
chore: add bundle analyze scripts

hanspagel:
I use `vite-bundle-analyzer` way too often to understand what’s bundled. This PR adds `pnpm analyze:standalone` as a script for convenience. :)

The script makes a build, analyzes the bundle and opens this in your browser:

Screenshot 2024-05-15 at 12 00 19


marclave (approved):
ah i missed these analyzers, thanks for adding this @hanspagel ✨

PR #1745, opened 2024-05-15
refactor: use lowlight for syntax highlighting in markdown, see #1716

### LINKED ISSUE #1716: Nuxt: Broken since 0.0.35

hanspagel:
This PR rolls back #1636 and re-introduces `rehype-highlight` (which uses lowlight for syntax highlighting).

This should fix #1716, but also increase the bundle size again.

I’ve got some ideas to level up our syntax highlighting game, but I think we’ve got more important things on our plate right now. So it has to wait a little bit. :)

### DIFF
--- a/.changeset/red-grapes-type.md
+++ b/.changeset/red-grapes-type.md
@@ -1,0 +1,5 @@
+---
+"@scalar/api-reference": patch
+---
+
+refactor: use lowlight instead of prism in the markdown component (rollback 64024a5)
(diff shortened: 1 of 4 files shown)


amritk:
huh closed?

PR #1746, opened 2024-05-15
refactor: use lowlight for syntax highlighting in markdown, see #1716

### LINKED ISSUE #1716: Nuxt: Broken since 0.0.35

hanspagel:
This PR rolls back #1636 and re-introduces `rehype-highlight` (which uses lowlight for syntax highlighting).

This should fix #1716, but also increase the bundle size again.

I’ve got some ideas to level up our syntax highlighting game, but I think we’ve got more important things on our plate right now. So it has to wait a little bit. :)

EDIT: I don’t get the relative CI check to pass. I even disabled the check, but it’s still there. Might be a bug on their end. Let’s just ignore it, we know the bundle size increases.


packages/api-reference/src/components/MarkdownRenderer/MarkdownRenderer.vue
```diff
@@ -322,7 +324,7 @@ onServerPrefetch(async () => await sleep(1))
 
 <style lang="postcss">
 .markdown {
-  pre code[class*='language-'] {
+  pre code.hljs {
```
amritk, line 327:
```suggestion
  pre code[class*='language-'], pre code.hljs {
```
should have just left both classes tbh in case we go back

amritk: but I guess we can always just come back to this PR and steal the code

amritk (approved):
Lets gooOOooo

tmastrom (approved):
Nice work! I added a basic playwright ui test for the nuxt integration

hanspagel:
This so great! Thanks for chiming in here!

marclave (approved):
tested and works great! nice work @hanspagel

PR #1749, opened 2024-05-15
chore: update vue and storybook

tmastrom:
updates `vue` and `vue-router` and updates storybook components with [updated control types](https://storybook.js.org/docs/api/arg-types#control) ('string' is no longer a valid option)

Made this a separate PR from nuxt testing as there is still some work to do and I will try to integrate @hanspagel latest work.


amritk:
This PR is all testing so I will await the 🦅 👁️

hanspagel:
I wonder if we should just update all packages if we need to test everything thoroughly anyway? :D

Either way, love it! :)

tmastrom:
Let's update everything at once to save on manual QA

PR #1753, opened 2024-05-15
fix: client popup blur

cameronrohani:
remove transform to fix windows users blur


edorgeville:
You're not going to be happy... it's still blurry!
But I got something more now:

https://github.com/scalar/scalar/assets/2685884/f41cd1b0-406b-4cb8-82b6-2730fc5f49c4

I think the blur is caused by the use of percents in `translate3d`.

marclave (approved):

PR #1756, opened 2024-05-16
feat(mock-server): add wildcard CORS headers

hanspagel:
Currently, the `@scalar/mock-server` doesn’t allow cross origin requests.

This PR adds loose CORS headers to make development (that’s what a mock server is for) easier.


marclave (approved):
nice @hanspagel ! we should think about how we can configure mocking for scalar eventually as a hosted offering!

PR #1757, opened 2024-05-16
feat(mock-server): return responses other than `200 OK`

hanspagel:
I’m testing the new proxy with the Scalar Galaxy mock server and just found out it’ only returning responses for the status code `200 OK`, but we defined some others, e.g. `201 Resource Created`.

This PR looks for the default keyword and a few other status codes. 🏄


marclave (approved):
👌 @hanspagel

marclave:
@hanspagel just needs to fix up the merge conflicts!

PR #1759, opened 2024-05-16
feat(api-reference): show text/plain example responses, fix #1758

### LINKED ISSUE #1758: text/plain of response not shown correctly
### What happens?

If a response returns a text/plain with an example, UI only shows "no Body"

Body should show the example text

```yml
post:
      summary: " (store)"
      description: "\n\n _app/controllers/test_controller.ts_ - **store**"
      parameters: []
      tags:
        - "BEST"
      responses:
        401:
          description: "Returns **401** (Unauthorized)"
        403:
          description: "Returns **403** (Forbidden)"
        404:
          content:
            text/plain:
              example: "Product not found"
          description: "Returns **404** (Not Found) as **text/plain**"
```
image

### Swagger/OpenAPI Example

_No response_

hanspagel:
We’re only showing json example responses, this PR adds support for text/plain responses, too.

See #1758

### DIFF
--- a/.changeset/smart-eels-work.md
+++ b/.changeset/smart-eels-work.md
@@ -1,0 +1,5 @@
+---
+"@scalar/api-reference": patch
+---
+
+feat: show text/plain example responses, too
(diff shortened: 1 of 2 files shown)


marclave (approved):
nice fix! I wonder if we should also handle * + other content types? @hanspagel

PR #1760, opened 2024-05-16
fix(api-reference): `x-displayName` is not used in the search modal and the tag component

hanspagel:
I struck me in the middle of the night: I forgot to add `x-displayName` to the search modal and the tag component.

This PR fixes both issues. :)

Screenshot 2024-05-16 at 13 15 38

### DIFF
--- a/.changeset/hot-needles-end.md
+++ b/.changeset/hot-needles-end.md
@@ -1,0 +1,5 @@
+---
+"@scalar/api-reference": patch
+---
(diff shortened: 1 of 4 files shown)


marclave (approved):
awesome work @hanspagel ! I wonder if for some of these extensions we should handle it in parse.ts so in case we need to showcase tag information anywhere its handled after parse?

PR #1763, opened 2024-05-16
feat: client ui updates

cameronrohani:
Before:
image

After:
image

Note: there's still a few things I'm not happy with about this but need it in for now 🙈

### DIFF
--- a/.changeset/many-ways-smell.md
+++ b/.changeset/many-ways-smell.md
@@ -1,0 +1,6 @@
+---
+"@scalar/use-codemirror": patch
+"@scalar/api-client": patch
+---
+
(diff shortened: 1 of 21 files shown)


antlio:
what do you think about giving that checked box a stronger color? also row addition call to action could be more aligned
image

cameronrohani:
> what do you think about giving that checked box a stronger color? also row addition call to action text could keep its alignment with previous one? image

feel free to change it up, otherwise I'll get to it asap :)

marclave (approved):
nice work @cameronrohani  & @antlio, much cleaner 🧼

PR #1766, opened 2024-05-16
refactor: add .NET 8 compatibility

kyllian:
**Problem**
The `Scalar.AspNetCore` package does not support LTS.

**Explanation**
.NET 8 is LTS; odd number releases are not. Currently, `Scalar.AspNetCore` only supports the .NET 9 preview.

**Solution**
Add multi-framework targeting for both .NET 8 and .NET 9. The project currently depends on the .NET 9 preview version of the `Microsoft.AspNetCore.OpenApi` package, and in order to support .NET 8, this dependency had to be downgraded to .NET 8.

I've included a .NET `.gitignore` file as well as an `.sln` file. Please let me know if those need to be excluded from this merge.


marclave (approved):
awesome work @kyllian !!!!

PR #1769, opened 2024-05-17
fix: client selector style

antlio:
**problem**
client selector icon depending on display size do not keep their width.

**solution**
this pr add padding and min-width on icons.

image


marclave (approved):
👁️ 4 detail!

PR #1775, opened 2024-05-17
feat(api-reference): cdn custom header example

hwkr:
Another swing at making it easier to add a header to the CDN references. The main change here is that we add `display: flex;` and `min-height: 0;` to the injected references container which should make it easier to add elements around it and _shouldn't_ break anyones layouts.

Google Chrome-2024-05-17-15-47-15@2x

### DIFF
--- a/README.md
+++ b/README.md
@@ -751,6 +751,39 @@ see [`legacyTheme.css`](https://github.com/scalar/scalar/tree/main/packages/them
 and [`updatedTheme.css`](https://github.com/scalar/scalar/tree/main/packages/themes/src/fixtures/updatedTheme.css)
 in the [`@scalar/themes`](https://github.com/scalar/scalar/tree/main/packages/themes/) package.
 
+### Custom Header with CDN
+
+If you'd like to add custom elements around the references app you can do so by wrapping it in a height restricting layout such as a grid or flexbox.
+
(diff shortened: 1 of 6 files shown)


examples/cdn-api-reference/src/public/api-reference-cdn-localhost.html
```diff
@@ -25,6 +25,6 @@
       var apiReference = document.getElementById('api-reference')
       apiReference.dataset.configuration = JSON.stringify(configuration)
     </script>
-    <script src="http://localhost:5053/@scalar/fastify-api-reference/js/browser.js"></script>
+    <script src="http://localhost:5053/reference/@scalar/fastify-api-reference/js/browser.js"></script>
```
hwkr, line 28: This seems to have been broken at some point.

amritk (approved):
Shouldn't we ALWAYS be flexin?

hanspagel (changes requested):
This is great! When testing it like this:

```bash
pnpm --filter @scalar-examples/cdn-api-reference dev
```

<http://127.0.0.1:3173/api-reference-local-header.html>

I see double scrollbars, though. That’s not how it’s supposed to look, is it?

Screenshot 2024-06-19 at 17 07 35

marclave:
closed in favour of https://github.com/scalar/scalar/pull/2198

PR #1780, opened 2024-05-19
feat: Adding Options to Scalar.AspNetCore

MarioGK:
**Problem**
Currently, there is no way to configure any of the configuration options using Scalar.AspNetCore nuget package

**Solution**
I added a new `MapScalarApiReference` extension which accepts an `Action<ScalarOptions>` which you can use to configure the options, i might have not added all the configuration but i added all of those i could find in the documentation.

There is also the `MapScalarApiReference` extension without any arguments with sensible default options

Please tell me if the code requires more features or "massaging" to get this PR merged as i would really like to use these features with the official nuget package and i think some folks as well.

### DIFF
--- a/packages/scalar.aspnetcore/Scalar.AspNetCore.cs
+++ b/packages/scalar.aspnetcore/Scalar.AspNetCore.cs
@@ -1,4 +1,6 @@
-﻿using Microsoft.AspNetCore.Builder;
+﻿using System.Text.Json;
+using System.Text.Json.Serialization;
+using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Http;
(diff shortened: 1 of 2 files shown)


marclave:
hey @MarioGK ! this is AWESOME thank you so much for this PR!! I agree making it more flexible will be better, and this is exactly that! I think we just need to add some more documentation to the README now that we have `ScalarOptions` for the users.

Feel free to add some examples to the README in the dot net package, I will try to get that done later tonight but feel free to add it earlier ✨

marclave:
@MarioGK can follow the deploy here :)

https://www.nuget.org/packages/Scalar.AspNetCore/1.1.1

should be released soon ✨

PR #1788, opened 2024-05-20
fix: client request response UI

cameronrohani:
A lot of minor touch ups to req / response

before:
image

after:
image


marclave (approved):
✨

PR #1791, opened 2024-05-21
fix: client selector responsive

antlio:
**Problem**
responsive issue spotted in sandbox: 

image

**Solution**
this pr favor the ellipsis icon only according to the selector width:

image

### DIFF
--- a/.changeset/breezy-roses-tap.md
+++ b/.changeset/breezy-roses-tap.md
@@ -1,0 +1,5 @@
+---
+"@scalar/api-reference": patch
+---
+
+fix: api client selector responsive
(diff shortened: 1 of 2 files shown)


hwkr (commented):
Looks like the content still clips at some screen sizes? I'm not sure how much better it can actually be though.

image

antlio:
> Looks like the content still clips at some screen sizes? I'm not sure how much better it can actually be though.
> 

might want to set a dropdown at some point?

hwkr:
@cameronrohani and I were chatting and wondering about just removing the words for mobile

antlio:
> @cameronrohani and I were chatting and wondering about just removing the words for mobile

nice! that's what i thought too and did

marclave (approved):
much cleaner and looks like the comments were addressed! thanks for taking the initiative here @antlio

PR #1792, opened 2024-05-21
docs(docusaurus): multiple api definitions example

hanspagel:
Someone asked [whether it’s possible to render multiple API definitions](https://github.com/scalar/scalar/discussions/1787) with the docusaurus plugin, and yes it is!

This PR adds an example to the README.


marclave (approved):
awesome work on the docs ✨

Issue #1786, opened 2024-05-20
Npm scalar install is conflicting with git scalar internal tool


jr-araujo:
### What happens?

When I'm trying to install Scalar using npm, the scalar bin conflicts with the one GitHub uses to monitor and control large files.

Below is the image depicting the scenario where I got the error when trying to install the Scalar tool:
![image](https://github.com/scalar/scalar/assets/11546614/1d849fc6-b05b-4d79-9e27-fa858c5bd57f)

Here is the tool Github has to monitor large files:
![image](https://github.com/scalar/scalar/assets/11546614/ebddc12e-4da8-437a-8b96-7681f938bf69)conflicts with the one GitHub uses

### What did you expect to happen?

The expectation is too simple: When installing the Scalar tool, it must be installed as expected without conflicting with any other around.

### How can we reproduce the issue?

With git installed, just run the command below:

1. npm -g install @scalar/cli

End then, you can see the issue in the image above.

hanspagel:
Good point. If you don’t use the Microsoft Scalar CLI, you can just replace it:

```bash
npm -g install @scalar/cli --force
```

Let me think about an alternative solutions for users who’d like to keep both.

hanspagel:
I tried to add an alias to the package.json in #1794, but that creates new problems:

```
@scalar/galaxy:build:  ERR_PNPM_DLX_MULTIPLE_BINS  Could not determine executable to run. @scalar/cli has multiple binaries: scalar, scalar-cli
```

I can’t really find a way to solve this. 🤔 Actually, I think it’s good as it is now. You can install it and overwrite Microsoft’s `scalar`, or just use `npx` or `pnpm dlx` and it’s all documented.

PR #1794, opened 2024-05-21
docs(cli): naming conflicts

hanspagel:
Users might [run into issues when installing the CLI](https://github.com/scalar/scalar/issues/1786), this PR adds some information to the README to help them to find a workaround.


marclave (approved):
nice doc fix :) just chatted w @geoffgscott on this but I think this makes the most sense for now!

PR #1796, opened 2024-05-21
feat(mock-server): dynamic values for generated examples

hanspagel:
Currently, the mock server just returns static JSON or a generated example (based on a given schema).

Requests to `/planets/{planetId}` are handled, but return a static value.

With this PR, `x-variable` is introduced as an extension to the specification, which tells the mock server to use a given variable for a property.

**Example Schema**

```yaml
    Planet:
      type: object
      properties:
        id:
          type: integer
          x-variable: planetId
```

**Example Response for `/planets/123`**

```json
{
  "id": 123,
}
```

I consider this an experimental feature for now, so documentation yet.

### DIFF
--- a/.changeset/selfish-points-stare.md
+++ b/.changeset/selfish-points-stare.md
@@ -1,0 +1,7 @@
+---
+"@scalar/mock-server": patch
+"@scalar/oas-utils": patch
+"@scalar/galaxy": patch
+---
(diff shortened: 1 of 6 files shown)


marclave (approved):
:O this is cool :) i gotta get my mock on

PR #1797, opened 2024-05-21
fix(api-client): doesn’t render preview for mimetype variations like `application/foobar+json`

hanspagel:
Currently, we’re only showing a response body preview for `application/json` (and very few other types), but not for variants like `application/foobar+json`.

With this PR we’re using the recently introduced `normalizeMimeType` helper (and remove the package we used before) to decide whether to render a preview.

Related: https://github.com/scalar/scalar/discussions/1742#discussioncomment-9462321


marclave (approved):
ahhh this is awesome ✨, nice work @hanspagel

PR #1798, opened 2024-05-21
chore(cli): `scalar serve` instead of `scalar reference`

hanspagel:
The CLI has the wonderful `scalar reference openapi.json` command, but `reference` never felt like a good fit.

It’s not that we’re _referencing_ something, we’re _serving_ an API reference.

This PR renames it to `scalar serve openapi.json`, and keeps `reference` as an alias.


marclave (approved):
i dig the serve ✨ 🚀

PR #1799, opened 2024-05-21
fix(api-reference): server url with variable is prepended with current URL, fix #1755

### LINKED ISSUE #1755: When using parameterized server URL, parameter editor not displayed on first load
### What happens?

When there are variables in the server URL, the Base URL section does not show an editor for those variables on first load.

The variable editor should be displayed.

Go to <https://docs.scalar.com/swagger-editor>, import URL <https://raw.githubusercontent.com/gravitee-io/gravitee-api-management/master/gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/resources/openapi/openapi-apis.yaml>,  after loading, there is no parameter editor:

image

(The screenshot above also shows that issue #1035 still exists)

But if we click the format code button on the top right corner of the spec editor, the parameter editor will appear:

image

### Swagger/OpenAPI Example

https://raw.githubusercontent.com/gravitee-io/gravitee-api-management/master/gravitee-apim-rest-api/gravitee-apim-rest-api-management-v2/gravitee-apim-rest-api-management-v2-rest/src/main/resources/openapi/openapi-apis.yaml

hanspagel:
When a server URL starts with a variable e.g. `{protocol}`, the URL is prepended with the current URL.

We’ve got a check for http/https and even file in place, but not for variables.

This PR extends the check to work with variables.

See #1755

### DIFF
--- a/.changeset/itchy-ants-talk.md
+++ b/.changeset/itchy-ants-talk.md
@@ -1,0 +1,5 @@
+---
+"@scalar/api-reference": patch
+---
+
+fix: server url with variable is overwritten with current url
--- a/packages/api-reference/src/components/Content/BaseUrl/BaseUrl.vue
+++ b/packages/api-reference/src/components/Content/BaseUrl/BaseUrl.vue
@@ -51,6 +51,7 @@ watch(
     })
   },
   {
+    immediate: true,
     deep: true,
   },
 )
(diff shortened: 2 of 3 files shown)


marclave (approved):
ah! nice @hanspagel 😎

PR #1800, opened 2024-05-21
fix: rehype-highlight transpilation was missed

### LINKED ISSUE #1783: Nuxt missing deps
### What happens?

Just tried setting up nuxt scalar since you introduced multiple api specs but it keeps crashing with missing deps

working demo ;)

https://codesandbox.io/p/devbox/nuxt-playground-forked-32ms33

once the hydration kicks in You will see 500 error

image

first I thought it might be related to pnpm modules resolution so tried shameful hoisting, then tried npm, yarn but still the same

### Swagger/OpenAPI Example

_No response_

amritk:
This was missed when we went back to highlight

closes #1783

### DIFF
--- a/.changeset/many-points-deliver.md
+++ b/.changeset/many-points-deliver.md
@@ -1,0 +1,5 @@
+---
+"@scalar/nuxt": patch
+---
+
+fix: nuxt transpilation commonjs errors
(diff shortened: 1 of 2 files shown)


marclave (approved):
we are so back @amritk

PR #1801, opened 2024-05-21
fix: add scalar component scope to missing components

hwkr:
Because we scope our tailwind CSS coming out of the components library all scalar components need to have the `scalar-component` class on their root element (which I forgot to do for the dropdown).

@amritk we might want to rethink at some point how were handling scoping here or if we even need to handle it anymore.


amritk:
Agreed with the re-thinking as well, maybe ticket it up

marclave (approved):
scope city! ✨

PR #1803, opened 2024-05-21
chore: provide working example for scalar_fastapi

rbange:
**Problem**
Currently, …

**Explanation**
This happens because …

**Solution**
With this PR …


marclave (approved):
✨ nice fix @rbange !!

Issue #1811, opened 2024-05-22
Docusaurus: OpenAPI 3.1 webhooks do not render


federicobond:
### What happens?

They show up in the sidebar with the id of the operation (instead of the summary) but there is no content when the item is selected.

<img width="868" alt="Captura de pantalla 2024-05-22 a la(s) 11 22 55 a  m" src="https://github.com/scalar/scalar/assets/138426/f6f2d35c-be08-439a-9ccd-09825a00a61e">

### What did you expect to happen?

I expected them to show up the same way the show up in [docs.scalar.com](https://docs.scalar.com/swagger-editor)

### How can we reproduce the issue?

1. Create a new Docusaurus site
2. Configure the Scalar plugin with an OpenAPI 3.1 spec containing webhooks

federicobond:
I can confirm this is fixed in the latest release.

PR #1812, opened 2024-05-22
fix: webhook operation parsing

federicobond:
**Problem**
Currently, OpenAPI 3.1 webhooks do not render correctly. See issue https://github.com/scalar/scalar/issues/1811

**Explanation**
This happens because when transforming webhooks in the `parse` function from `@scalar/api-reference`, the originalWebhook variable is assigned to the `PathItem` object instead of the `Operation`, as implied by [this cast](https://github.com/scalar/scalar/blob/22bb58f97660b28a797d0d52f21c33e68fd079ab/packages/api-reference/src/helpers/parse.ts#L91-L94).

**Solution**
With this PR, the originalWebhook variable is correctly assigned to the Operation.

**Note:** the PathItem also has summary and description fields. It might be useful to fall back to them if they are missing in the operation.

### DIFF
--- a/.changeset/poor-tips-develop.md
+++ b/.changeset/poor-tips-develop.md
@@ -1,0 +1,5 @@
+---
+"@scalar/api-reference": patch
+---
+
+fix: webhook operation parsing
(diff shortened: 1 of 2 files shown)


marclave (commented):
nice work @federicobond ! 1 line makes all the difference :)want to add a changeset and we can get this in? just a patch would be perfect

hanspagel:
Yesss, great contribution!

hanspagel (approved):

marclave:
> @marclave like this?

perfect!!

marclave (approved):

PR #1816, opened 2024-05-22
fix: add support for union types in getExampleFromSchema

federicobond:
**Problem**
Currently, Scalar shows a warning like this when the spec declares union types:
```
[getExampleFromSchema] Unknown property type "string,null".
```

**Explanation**
This happens because there is no explicit support for union types in `getExampleFromSchema`. See section [7.6.1](https://json-schema.org/draft/2020-12/json-schema-core#section-7.6.1) of the JSON Schema spec for reference.

**Solution**
With this PR, `getExampleFromSchema` will understand union types and return null for nullable unions or an example for the first types otherwise. Let me know if this makes sense for you.

### DIFF
--- a/.changeset/smooth-jeans-protect.md
+++ b/.changeset/smooth-jeans-protect.md
@@ -2,4 +2,4 @@
 "@scalar/oas-utils": patch
 ---
 
-fix: add support for union types in getExampleFromSchema
+feat: union types in getExampleFromSchema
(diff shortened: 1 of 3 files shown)


hanspagel (approved):
Wow, what a great PR!

marclave (approved):
new support? tests? what a perfect PR @federicobond !

PR #1819, opened 2024-05-22
fix: response body does not show a preview when the content-type is undefined, fix #1818

### LINKED ISSUE #1818: Response body not showing when using api-reference v1.22.45
### What happens?

Hi! Great openapi UI, but i get an error when using the latest version.

When using hono-api-reference and v1.22.45 of `api-reference` the body is not shown. The error shown in the UI is: "Can’t render a preview. The Content-Type header is missing or unknown."

Content-Type is set by Hono.

Error in console:

```
TypeError: Cannot read properties of undefined (reading 'replace')
    at sp (api-reference@1.22.45:40:159200)
    at api-reference@1.22.45:40:776184
    at ae.fn (api-reference@1.22.45:28:8905)
    at ae.run (api-reference@1.22.45:28:1475)
    at get value (api-reference@1.22.45:28:9152)
    at api-reference@1.22.45:40:776252
    at ae.fn (api-reference@1.22.45:28:8905)
    at ae.run (api-reference@1.22.45:28:1475)
    at get value (api-reference@1.22.45:28:9152)
    at api-reference@1.22.45:40:776592
``` 

Downgrading to v1.22.44 by setting this in config does work:

```
app.get(
  '/',
  apiReference({
    cdn: 'https://cdn.jsdelivr.net/npm/@scalar/api-reference@1.22.44',
    spec: {
      url: '/openapi',
    },
  }),
)
```

Response is shown in the UI

Test any endpoint in the UI, the response is not shown.

### Swagger/OpenAPI Example

_No response_

hanspagel:
Currently, the response body preview doesn’t show up when the content-type header isn’t set. That’s bad.

This PR makes sure it’ll still work even if the content-type header isn’t set.

See #1818

### DIFF
--- a/.changeset/gold-peas-report.md
+++ b/.changeset/gold-peas-report.md
@@ -1,0 +1,5 @@
+---
+"@scalar/oas-utils": patch
+---
+
+fix: response body does not show a preview when the content-type is undefined
--- a/packages/oas-utils/src/normalizeMimeType.ts
+++ b/packages/oas-utils/src/normalizeMimeType.ts
@@ -1,6 +1,16 @@
 import type { ContentType } from './types'
 
-export function normalizeMimeType(contentType: string) {
+/**
+ * Normalizes a MIME type to a standard format.
+ *
+ * Input: application/problem+json; charset=utf-8
+ * Output: application/json
+ */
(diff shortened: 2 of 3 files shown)


marclave (approved):
tested and works!

PR #1821, opened 2024-05-22
feat: scalar button component type prop

antlio:
this pr adds a type prop to the scalar button component in order to be able to use it as a `submit` action within a form for instance, along the addition of the `reset` type

### DIFF
--- a/packages/components/src/components/ScalarButton/ScalarButton.vue
+++ b/packages/components/src/components/ScalarButton/ScalarButton.vue
@@ -15,11 +15,13 @@ withDefaults(
     loading?: LoadingState
     size?: Variants['size']
     variant?: Variants['variant']
+    type?: 'button' | 'submit' | 'reset'
   }>(),
   {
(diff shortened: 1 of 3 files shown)


packages/components/src/components/ScalarButton/ScalarButton.vue
```diff
@@ -15,11 +15,13 @@ withDefaults(
     loading?: LoadingState
     size?: Variants['size']
     variant?: Variants['variant']
+    type?: 'button' | 'submit'
```
hwkr, line 18: We should probably have the "reset" option as well since it's a valid option

hwkr (approved):

packages/components/src/components/ScalarButton/ScalarButton.vue
antlio: good call! added

marclave (approved):

PR #1823, opened 2024-05-22
docs(api-reference): improve css examples

hanspagel:
We have some CSS examples in the READMEs, but it’s not very obvious that it’s CSS.

This PR wraps the CSS examples in a `style` tag, which could help to make it more obvious. :)


tmastrom (approved):
Thank you from a Vue noob 😄

PR #1824, opened 2024-05-22
feat: scalar cli default display

antlio:
this pr is a proposal to :
- favor available commands `--help` when no argument has been provided over the init command
- add help in case of error (wrong argument given)


marclave (approved):
we have gained help : )

PR #1826, opened 2024-05-22
feat: scalar cli prompts

antlio:
this pr is a proposal to use the beautiful @natemoo-re / clack prompts package over the current prompts one, within the cli init command + adds next step command logs


marclave (approved):
this is fun! ✨

PR #1833, opened 2024-05-22
fix: don't try to access contentType value which doesn't exist

hwkr:
I feel like we shouldn't allow `any` as a variable type.

### DIFF
--- a/packages/api-client/src/components/ApiClient/Response/ResponseBody.vue
+++ b/packages/api-client/src/components/ApiClient/Response/ResponseBody.vue
@@ -16,14 +16,14 @@ const mediaType = computed(() => {
   const headers = normalizeHeaders(props.response?.headers)
 
   // Get the content-type header
-  const contentTypeHeader = headers['content-type']
+  const contentTypeHeader = `${headers['content-type']}`
 
   if (!contentTypeHeader) {
     return null
   }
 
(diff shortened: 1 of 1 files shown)


packages/api-client/src/components/ApiClient/Response/ResponseBody.vue
```diff
@@ -16,14 +16,14 @@ const mediaType = computed(() => {
   const headers = normalizeHeaders(props.response?.headers)
 
   // Get the content-type header
-  const contentTypeHeader = headers['content-type']
+  const contentTypeHeader = `${headers['content-type']}`
```
hwkr, line 19: I'm coercing this to a string now so it gets typed as `string` instead of `any`.

packages/api-client/src/components/ApiClient/Response/ResponseBody.vue
```diff
@@ -16,14 +16,14 @@ const mediaType = computed(() => {
...
 
   if (!contentTypeHeader) {
     return null
   }
 
   // application/foobar+json; charset=utf-8 -> application/json
-  return normalizeMimeType(contentTypeHeader?.value)
+  return normalizeMimeType(contentTypeHeader)
```
hwkr, line 26: This is the problem.

marclave (approved):
tested and we got syntax highlighting back!

PR #1839, opened 2024-05-23
feat: Angular integration

geoffgscott:
Adds a basic package for angular integration

Theme CSS is not getting loaded correctly right now and reactive props have not be configured.


hanspagel:
Let’s keep the branch, but close the PR for now. Can’t wait to get back to it, though.

PR #1841, opened 2024-05-23
feat: create esm javascript build docs

geoffgscott:
Adds an ESM integration for references. 

Improvements and Future Work:
- The mounted references are always 100% viewport height. Ideally it should be flex and just fill its parent container
- Other integrations should likely use the exported function as their main entrypoint. Standalone should only be used for the script tag integrations (and not in bundled code).

### DIFF
--- a/examples/esm-api-reference/index.html
+++ b/examples/esm-api-reference/index.html
@@ -1,0 +1,33 @@
+<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <link
+      rel="icon"
+      type="image/svg+xml"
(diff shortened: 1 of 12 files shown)


marclave (commented):
woah this is neat! I tested it and it worked but im so unfamiliar with esm lets get @amritk and @hanspagel to join in on the review : )

amritk (approved):
Beeeeeauty

geoffgscott:
@marclave ESM is really being misused here and in this case basically just means providing a JS function that can mount the app. 

The standalone build has a bunch of logic at the top level of the module to mount to an element on page load. We just need a little better control for integrations.

PR #1842, opened 2024-05-24
feat: create scalar tailwind preset

hwkr:
This creates a Scalar [Tailwind preset](https://tailwindcss.com/docs/presets) that defines all our core tailwind configuration. Different projects can extend the preset with more context specific configuration.

Originally the ideas was to export the core tailwind preset from `@scalar/themes` but I ran into issues with Storybook because importing anything from `@scalar/themes` tries to import a CSS file which makes Storybook Vite angry.

We probably shouldn't auto load `base.css` into the page just because someone imports a tailwind preset out of `@scalar/themes` but I wasn't really sure how to fix that. I'm hoping @amritk might know of a good way to handle it. In the mean time I just split up the configurations / preset so it can be moved over easier as a next step.


geoffgscott (approved):
LGTM

marclave (approved):
nice work @hwkr lets get this in so we can iterate and have an amazing foundation

PR #1843, opened 2024-05-24
feat(#doc-1845): fix auto importing css

amritk:
This PR is the best of both worlds, we generate the `style.css` files. If the user does not import them, we import them in JS and throw a warning.

I disabled the path aliasing in the examples or it just pulls the styles from the components directly.

Good way to test is to comment out 
```ts
// import '@scalar/api-reference-react/style.css'
```
from `examples/react/src/App.tsx`
then refresh and you should see the warning 
image

### DIFF
--- a/.changeset/chatty-crews-allow.md
+++ b/.changeset/chatty-crews-allow.md
@@ -1,0 +1,14 @@
+---
+"@scalar/api-reference-react": minor
+"@scalar/api-client-proxy": minor
+"@scalar/api-client-react": minor
+"@scalar/api-reference": minor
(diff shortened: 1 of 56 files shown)


hanspagel:
Why does the bundle size increase by ~200kb? 🤔

amritk:
That is a good question. Standalone shouldn't even be affected 🤔

hwkr (approved):
Lgtm!

marclave (approved):

marclave:
tested! lets get this in 😤

PR #1844, opened 2024-05-24
feat: add support for x-additionalPropertiesName

mpminardi:
**Problem**
Currently, scalar does not have support for the [x-additionalPropertiesName extension](https://redocly.com/docs/api-reference-docs/specification-extensions/x-additional-properties-name/) which redocly uses to support displaying a more descriptive property name for `additionalProperties`.

**Explanation**
This happens because there is currently no logic for this and scalar instead defaults to `additional properties` for the property heading.

**Solution**
With this PR `x-additionalPropertiesName` is supported and overrides the default `additional properties` heading if provided.

***Example Using x-additionalPropertiesName***
image


mpminardi:
Please let me know if there are any tests or other places I should be updating / anything missing from this PR! Can look at adding an example of usage to galaxy as well if that would be helpful / desired.

PR #1854, opened 2024-05-24
fix: update password icon

cameronrohani:
Update password show / hide icon to feel more consistent with the font weight

After:
image
image

Before:
image


marclave (approved):
before:

image


after:

image

Issue #1859, opened 2024-05-26
Authentication section does not handle oauth2 securitySchemes properly


swimand:
### What happens?

I have configured the openapi scheme to include a oauth2 securityScheme. The Scalar UI however does not present the option to authenticate properly. The image below shows the output when i have oauth2 with a "password" flow.
![image](https://github.com/scalar/scalar/assets/91246428/37899291-312f-4c0c-9fe5-b12882c504cd)

### What did you expect to happen?

I expected the possibility of entering the same values as shown in Swagger UI. Here are images of the values that swagger presents with the same openapi scheme, within the same application: 
![image](https://github.com/scalar/scalar/assets/91246428/09d0e00a-8ddf-4b40-9aad-b53c14bf7df0)

### How can we reproduce the issue?

Use the following schema to replicate in sandbox: 
```
{
  "openapi": "3.1.0",
  "info": {
    "title": "Hello World",
    "version": "1.0.0"
  },
  "paths": {},
  "components": {
    "securitySchemes": {
      "auth": {
        "type": "oauth2",
        "flows": {
          "password": {
            "tokenUrl": " /v1/token",
            "scopes": {}
          }
        }
      }
    }
  }
}
```

or create a new python fastapi project with scalar and configure a schema with the following code: 

```
        openapi_schema["components"]["securitySchemes"] = {
            "auth": {
                "type": "oauth2",
                "flows": {
                    "password": {
                        "tokenUrl": api_config.path + "/v1/oauth2/token",
                        "scopes": {}
                    }
                }
            }
        }
```

hanspagel:
You’re so right. Let’s add this feature! On it.

hanspagel:
Can you try again and let me know if it works?

hanspagel:
Closing it for now, but let me know if you face any issues and I’ll reopen it in a heartbeat. 👍

swimand:
Looks great! Fast work! 🤩
However there are a few parameters missing for full support, such as client secret and Client Credentials Location. 

There is also a weird clipping occurring for the view password button and Authorize button: 

![image](https://github.com/scalar/scalar/assets/91246428/01c626e3-3c55-447c-b8d0-e8e7021e7fbb)

Is there a way to install this version on my application to also test out the actual login logic?

PR #1862, opened 2024-05-27
feat: OAuth 2.0 password flow

hanspagel:
This PR adds support for the OpenAuth 2.0 password flow.

To achieve this, the API client needs to know everything about the server url (and variables, if set), so we have to move the whole server state from `@scalar/api-reference` to `@scalar/api-client`, which was long overdue anyway.

Now, the API client is able to request password grants (from the API, with the provided server URL and endpoint) for you.

See #1859

**Additional improvements**
* `toast()` has support for a `description` now, this helps to split longer texts

**Before**
Screenshot 2024-05-27 at 14 50 31

**After**
Screenshot 2024-05-27 at 14 53 04

**Errors**
Screenshot 2024-05-28 at 11 47 28


marclave (approved):
awesome work @hanspagel !

we will need to move this into the new app for functionality too : )

PR #1864, opened 2024-05-27
feat: copy server url with replaced variables

hanspagel:
If you click on the server URL, it’s copied to the clipboard. Even if it has variables like `https://example.com/{version}`.

With this example, we try to replace the variables first, but keep the placeholders if the variables aren’t defined.

 *  `https://example.com/{foo}` (without foo) copies `https://example.com/{foo}`
 *  `https://example.com/{foo}` (with foo: 123) copies `https://example.com/123`
 
Screenshot 2024-05-27 at 15 23 56

### DIFF
--- a/.changeset/neat-singers-cry.md
+++ b/.changeset/neat-singers-cry.md
@@ -1,0 +1,5 @@
+---
+"@scalar/api-reference": patch
+---
+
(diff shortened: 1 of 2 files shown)


marclave (approved):
🔥 🔥 🔥 🔥 🔥

PR #1866, opened 2024-05-27
feat(oas-utils): generate random values based on the `format`

hanspagel:
To generate example responses, we’ll just use empty values.

As suggested in https://github.com/scalar/scalar/discussions/1825 this PR generates example values (in some cases, e.g. response examples) based on the format.


marclave (approved):
this is so fun!

PR #1868, opened 2024-05-27
feat(api-reference): support x-enumDescriptions

mpminardi:
**Problem**
Currently, scalar does not have support for the [x-enumDescriptions extension](https://redocly.com/docs/api-reference-docs/specification-extensions/x-enum-descriptions/) which redocly uses to support displaying a table with individual descriptions for enum values.

**Explanation**
This happens because there is currently no support in scalar for this extension.

**Solution**
With this PR `x-enumDescriptions` is supported and displays a table with enum value -> description mappings when populated for an enum.

***Sample***
image

***Example OpenAPI***
```yaml
openapi: 3.1.0
info:
  title: Scalar Galaxy
  version: 1.0.0
components:
  schemas:
    User:
      type: object
      properties:
        status:
          type: string
          enum:
            - ACTIVE
            - INACTIVE
        role:
          type: string
          enum:
            - ADMIN
            - MEMBER
          x-enumDescriptions:
            ADMIN: Can do **all** of the things!
            MEMBER: Can do ~all~ _some_ things
```

### DIFF
--- a/packages/api-reference/src/components/Content/Schema/SchemaProperty.vue
+++ b/packages/api-reference/src/components/Content/Schema/SchemaProperty.vue
@@ -1,5 +1,11 @@
 <script lang="ts" setup>
 import { MarkdownRenderer } from '../../MarkdownRenderer'
+import {
+  SimpleCell,
+  SimpleHeader,
+  SimpleRow,
+  SimpleTable,
+} from '../../SimpleTable'
 import Schema from './Schema.vue'
 import SchemaPropertyHeading from './SchemaPropertyHeading.vue'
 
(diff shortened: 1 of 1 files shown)


mpminardi:
Not sure how y'all were wanting to support giving enum values individual descriptions but figured I'd throw something quick up that uses the [x-enumDescriptions extension from redocly](https://redocly.com/docs/api-reference-docs/specification-extensions/x-enum-descriptions/).

Let me know if anything else is required here / needs to change / if you are planning on going another direction here! Will also conflict with https://github.com/scalar/scalar/pull/1865/files so will need to rebase / adjust slightly if this change is desired but don't think that should require major change / effort.

federicobond:
This is very useful! I'll add that the "native" way to do this in OpenAPI is to use `anyOf` with const values, as described [here](https://github.com/OAI/OpenAPI-Specification/issues/348#issuecomment-336194030). Perhaps we can normalize all enums into a standard format for display.

PR #1870, opened 2024-05-27
fix: api reference auth form style

antlio:
**Problem**
authentication block style is broken as seen below:
image

**Explanation**
lack of specificity using `where` is not maintaining style one reference is published. this pr is a proposal to remove `where` usage in order to ensure higher style specificity.


marclave (approved):
wow, awesome fix @antlio

PR #1872, opened 2024-05-27
fix: set theme variables to Docusaurus theme at root

hwkr:
Set the Docusaurus theme variables on `<body>` instead of `.scalar-api-reference` so they apply to toasts (and anything else that's outside of that element, headless ui?) 

Also moves the references CSS variables to a `scalar-config` CSS layer so they can be overridden without using `!important`.

## Before:

Google Chrome-2024-05-27-16-25-53@2x

## After: 

Google Chrome-2024-05-27-16-22-37@2x

### DIFF
--- a/.changeset/big-cherries-grin.md
+++ b/.changeset/big-cherries-grin.md
@@ -1,0 +1,5 @@
+---
+"@scalar/docusaurus": patch
+---
+
+fix: apply docusaurus theme to body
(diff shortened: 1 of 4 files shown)


amritk (approved):
Layers, gotta love em

cameronrohani (approved):

marclave (approved):

PR #1873, opened 2024-05-27
feat(mock-server): http basic auth

hanspagel:
I missed a way to test authentication with our mock server today.

So, I’ve had to do a little midnight hack session and … add HTTP basic authentication to the mock server. :)


marclave (approved):
tested and great : )

PR #1875, opened 2024-05-28
fix: small images destroy the layout

hanspagel:
Small images are displayed as inline content. Maybe we could set them to `display: block`?

**Before**
Screenshot 2024-05-28 at 11 23 02

**After**
Screenshot 2024-05-28 at 11 23 10


marclave (commented):
nice fix @hanspagel :)

marclave (approved):

PR #1881, opened 2024-05-28
fix: remove hardcoded dark-mode class from ExampleRequest

mpminardi:
**Problem**
Currently, the example request section renders as dark mode regardless of whether dark mode is enabled.

**Explanation**
This happens because a `dark-mode` class has been hardcoded on the `Card` component.

**Solution**
With this PR the dark mode setting is respected and the card renders properly in both light and dark mode.

***Example Before***
image

***Example After Light***
image

***Example After Dark***
image

### DIFF
--- a/packages/api-reference/src/components/Content/Operation/ExampleRequest.vue
+++ b/packages/api-reference/src/components/Content/Operation/ExampleRequest.vue
@@ -159,7 +159,7 @@ computed(() => {
 })
 </script>
 <template>
-  <Card class="dark-mode">
+  <Card>
     <CardHeader muted>
       <div class="request-header">
         <HttpMethod


cameronrohani:
Hey @mpminardi, thank you for the PR.

This was actually an intentional design decision.

The idea is that having more contrast between two cards makes it easier to quickly parse and distinguish between them. We don't follow this through in the night mode as it becomes jarring to have a large white box in night mode.

If you don't like this design you can technically override it by writing custom theme CSS by appending the dark-mode class after the light mode class i.e: `.light-mode .dark-mode`


Here's a full theme file for reference:
https://gist.github.com/cameronrohani/8457944afe237fa796545e0bb52ae575

Hope this helps

mpminardi:
> Hey @mpminardi, thank you for the PR.
> 
> This was actually an intentional design decision.
> 
> The idea is that having more contrast between two cards makes it easier to quickly parse and distinguish between them. We don't follow this through in the night mode as it becomes jarring to have a large white box in night mode.
> 
> If you don't like this design you can technically override it by writing custom theme CSS by appending the dark-mode class after the light mode class i.e: `.light-mode .dark-mode`
> 
> 
> Here's a full theme file for reference:
> https://gist.github.com/cameronrohani/8457944afe237fa796545e0bb52ae575
> 
> Hope this helps

Ahh fair enough, will look at overriding on my end then. Thanks for the context and the link to the theme!

cameronrohani:
> Ahh fair enough, will look at overriding on my end then. Thanks for the context and the link to the theme!

Of course, I appreciate you thinking about it and making the PR regardless 😄 

If you need any help with doing custom themes don't hesitate to reach out

mpminardi:
> > Ahh fair enough, will look at overriding on my end then. Thanks for the context and the link to the theme!
> 
> Of course, I appreciate you thinking about it and making the PR regardless 😄
> 
> If you need any help with doing custom themes don't hesitate to reach out

Oh yeah for sure, and will do. Thank you for the quick responses!

PR #1883, opened 2024-05-29
feat: add listbox component

hwkr:
https://github.com/scalar/scalar/assets/6374090/cbd883fb-9bc2-4573-9a89-c84dc837f379


packages/components/src/components/ScalarListbox/ScalarListbox.vue
```diff
@@ -0,0 +1,86 @@
...
+              :class="
+                cx(variants({ active, selected, disabled: option.disabled }))
+              ">
+              <span class="inline-block min-w-0 flex-1 truncate">{{
+                option.label
+              }}</span>
+              <ScalarIcon
+                v-if="selected"
+                icon="Checkmark"
+                size="sm" />
```
antlio, line 77: what do you think about prepending it to the text instead for better scan-ability?

antlio (approved):
love that!

marclave (approved):

packages/components/src/components/ScalarListbox/ScalarListbox.vue
hwkr: We could! This is just where it was in FlowListbox

PR #1884, opened 2024-05-29
test: scalar api reference

tmastrom:
Test the standalone api reference in CI to catch issues in feature branches before they are pushed to the CDN. This PR creates a shared testing util so changes to the e2e test propagate to the local and live CDN versions. This PR only adds basic UI testing but snapshot testing can be added easily if we decide that we need it.

### DIFF
--- a/examples/cdn-api-reference/src/public/api-reference-cdn-local.html
+++ b/examples/cdn-api-reference/src/public/api-reference-cdn-local.html
@@ -21,10 +21,9 @@
       var configuration = {
         theme: 'purple',
       }
-
       var apiReference = document.getElementById('api-reference')
       apiReference.dataset.configuration = JSON.stringify(configuration)
     </script>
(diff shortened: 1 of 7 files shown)


examples/cdn-api-reference/src/public/api-reference-cdn-local.html
```diff
@@ -21,10 +21,9 @@
       var configuration = {
         theme: 'purple',
       }
-
       var apiReference = document.getElementById('api-reference')
       apiReference.dataset.configuration = JSON.stringify(configuration)
     </script>
-    <script src="http://localhost:5053/@scalar/fastify-api-reference/js/browser.js"></script>
+    <script src="http://localhost:5053/reference/@scalar/fastify-api-reference/js/browser.js"></script>
```
hwkr, line 27: Oh I also fixed this in one of my PRs 😂

hwkr (approved):
lgtm!

marclave (approved):
🕺 nice @tmastrom !

PR #1885, opened 2024-05-29
refactor: remove :where prefix in favor of cascade layers

hwkr:
So this this will unscope our components tailwind classes in favor of just putting them in a layer that's super easy to override. This means we might technically have leaky component css **but** it should never override a consuming apps classes.


marclave (approved):
learned a lot from this! nice work @hwkr

PR #1886, opened 2024-05-29
refactor: remove component app install

hwkr:
Removes the component app install, technically this is a breaking change but I don't think anyone was using this feature.

cc @amritk


marclave (approved):
nice :)

PR #1887, opened 2024-05-29
chore(cli): provide alternative name for the CLI

hanspagel:
Currently, users might run into conflicts with the existing `scalar` CLI that Microsoft built.

When installing, they can overwrite the other CLI and use ours as `scalar`. There’s also a note in the README on the matter. I’ve did a ton of reading, trying to find a good solution, here’s a short summary:

* We could just change the name to be something else, but I couldn’t come up with one that is as cool as `scalar`.
* We could ship executables with bun, which we’ll probably do at some point in the future, but it would require us to build/support/publish binaries for the various platforms in a convenient way, which adds a lot of complexity in multiple areas that I’d like to avoid at this point.
* We can’t hook into the install process to react to the error.

This PR suggests a pragmatic solution: We’ll just register the CLI under two names. If people run into a conflict and want to keep Microsoft’s `scalar`, they can just ignore the error and use `scalar-cli`.

This PR doesn’t change the docs, because I need to test it first, before we promote the solution. I couldn’t find a way to reproduce the issue when installing or linking the source files, so we need to publish to npm to test the solution.

### DIFF
--- a/.changeset/poor-ghosts-matter.md
+++ b/.changeset/poor-ghosts-matter.md
@@ -1,0 +1,5 @@
+---
+"@scalar/cli": patch
+---
+
+chore: provide alternative name for the CLI (`scalar-cli`)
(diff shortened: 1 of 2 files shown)


marclave (approved):
ah nice a reasonable solution ✨

PR #1888, opened 2024-05-29
fix: client library default value ignores `hiddenClients` configuration, fix #1861

### LINKED ISSUE #1861: hiddenClients still defaults curl even when it is supposed be disabled
### What happens?

as mentioned here: https://github.com/scalar/scalar/issues/1250
I have setted:
`hiddenClients: { php: true, python: true, c: true, node: true, javascript: true, go: true, java: true, ruby: true, shell: true, clojure: true, csharp: true, kotlin: true, objc: true, swift: true, r: true, powershell: true, ocaml: true, curl: true, },
`
It still makes curl default:
image

and upon clicking there is only http
image

thank you guys!

show correct clients that is set in hiddenClients

1. Set hiddenClients curl
2. Curl is still present

### OpenAPI Document

_No response_

hanspagel:
Currently, the `hiddenClients` configuration is ignored when determining the default client library. This PR fixes it.

To achieve this I merged `useHttpClients` and `useHttpClientStores`. The separation didn't really made sense after looking at it and it was the root cause of the mentioned regression.

**Additional improvements**

* Hides the client libraries the example requests entirely when all clients are hidden

See #1861

**Show just one or a few clients**
Screenshot 2024-05-29 at 13 16 08

**Hide all clients and the example requests (still shows `x-customExamples`)**
Screenshot 2024-05-29 at 12 16 15

Screenshot 2024-05-29 at 12 17 21


tylermann:
One thing I noticed when trying out this option is that it also hides any mention of the API URL. For example:

image

without the clients hidden then it includes the `POST /v1/account/billing/payment-session` in this example:
image

I'm thinking the ideal behavior would be to still include the `POST /v1/account/billing/payment-session` with the copy to clipboard icon, but no client selector or example request code.

PR #1891, opened 2024-05-29
feat: Adds new unified package for code highlight and markdown rendering

geoffgscott:
This package will be used as the basis for all remark/rehype pipelines where we are rendering code and markdown. 

Provides a common place for transforms and themes. 

Testing:
```
pnpm dev
```


marclave (approved):
looks great code wise! just getting @cameronrohani to do a quick overview on styles.

we can also fast follow in another PR ✨

PR #1902, opened 2024-05-30
feat(api-reference): normalize schemas with unnecessary nesting or single-value enums

federicobond:
**Problem**
Currently, Scalar docs show an ALLOF badge even in cases where there is a single item, causing unnecessary nesting.

**Explanation**
Some automated tools for generating schemas may use `allOf` keywords with a single value. These can be safely normalized to avoid the nesting. Redoc already does something similar.

**Solution**
With this PR a seam is introduced to allow for normalizing schemas. This is also useful to simplify frontend code such as this the one introduced in https://github.com/scalar/scalar/pull/1865 for handling const values.

No changes have been made to frontend code yet, but an extension of this PR could take advantage of this seam so that it's easy to handle enum descriptions in any form:

```yaml
allOf:
  - const: A
     description: A desc
  - const: B
     description: B desc
```
and 

```yaml
type: string
enum: [A, B]
x-enumDescriptions:
  A: A desc
  B: B desc
```


packages/api-reference/src/components/Content/Operation/RequestBody.vue
```diff
@@ -20,6 +24,12 @@ if (prop.requestBody?.content) {
     selectedContentType.value = contentTypes.value[0] as ContentType
   }
 }
+
+const schema = computed(() => {
+  return normalizeSchema(
+    prop?.requestBody?.content?.[selectedContentType.value]?.schema,
+  )
```
federicobond, line 31: Looking at the build failures, I probably fumbled this part, because I have no experience with Vue.

hanspagel (changes requested):
Great idea! I think this could produce max call stack exceeded errors for circular references, though.

We could prevent this if we do the normalisation before we’re dereferencing. 🤔

But [that part of the code is a beautiful mess](https://github.com/scalar/scalar/blob/main/packages/api-reference/src/helpers/parse.ts#L2-L3) and I would feel the need to use the opportunity and refactor it. 👀 

Maybe we’ll wait a week or two until we merge The Secret Project™ into `scalar/scalar`. Wdyt @marclave?

marclave:
> Great idea! I think this could produce max call stack exceeded errors for circular references, though.
> 
> We could prevent this if we do the normalisation before we’re dereferencing. 🤔
> 
> But [that part of the code is a beautiful mess](https://github.com/scalar/scalar/blob/main/packages/api-reference/src/helpers/parse.ts#L2-L3), which is the root cause of dozens of @ts-ignores and I would feel the need to use the opportunity and refactor it. 👀
> 
> Maybe we’ll wait a week or two until we merge The Secret Project™ into `scalar/scalar`. Wdyt @marclave?



> Great idea! I think this could produce max call stack exceeded errors for circular references, though.
> 
> We could prevent this if we do the normalisation before we’re dereferencing. 🤔
> 
> But [that part of the code is a beautiful mess](https://github.com/scalar/scalar/blob/main/packages/api-reference/src/helpers/parse.ts#L2-L3), which is the root cause of dozens of @ts-ignores and I would feel the need to use the opportunity and refactor it. 👀
> 
> Maybe we’ll wait a week or two until we merge The Secret Project™ into `scalar/scalar`. Wdyt @marclave?

yeah i think waiting is probably the best play here :)

hanspagel:
We’re doing some heavy lifting again. 😅 I should really use the gap between those huge PRs to get this PR here in, sorry. I’ll try to do better.

federicobond:
Hi @hanspagel @marclave, just wondering about this and the status of the refactoring you mentioned.

federicobond:
Awesome, let me know if I can be of help!

hanspagel:
Man, I’m just not getting to it. I have something in a branch, but can’t find the time to finish it up. Anyway, it’s on my list and I won’t forget about it and will put together an updated PR soon. Thanks for the idea!

PR #1903, opened 2024-05-30
chore: lint icon svgs

hwkr:
Cleans up our icon library and adds a linter to enforce certain icon svg properties that we want (and prevents us from accidentally setting bad properties).

Rules that are now enforced:
- The root `<svg>` element must have:
  -  `xmlns` set to `http://www.w3.org/2000/svg`
  - `viewbox` set, starting from `0, 0` and 1:1 (square)
  - `fill` set to `currentColor` or `none`, this keeps svgs from defaulting to `#000` for fill if unset
  - no width and height attributes (so it's responsive)
- all fills must be either `currentColor` or `none`
- if `stroke` is set then `stroke-linecap` and 'stroke-linejoin' must be `round` for consistency
- 'stroke-width' must not be set on any element (it should default to `1` and we can override it with CSS)
- up to one `<title>` element is allow
- only `<path>`, `<circle>` and `<rect>` shapes are allowed, we should be able to create any icon we want to with these


packages/components/src/components/ScalarIcon/.svglintrc.js
```diff
@@ -0,0 +1,101 @@
...
+      '*': false,
+    },
+    attr: [
+      {
+        // Ensure that the SVG element has the appropriate attributes
+        'rule::selector': 'svg',
+        // Don't allow extraneous attributes
+        'rule::whitelist': true,
+        // ViewBox must be square and namespace must be set
+        'viewBox': /^0 0 (\d+) \1$/,
```
antlio, line 34: should we set a specific viewbox size at some point?

antlio (approved):
great move!

packages/components/src/components/ScalarIcon/.svglintrc.js
hwkr: Yeah I think so, I just didn't want to reformat all the icons as part of this PR. Currently I have 12px set up as the standard in the readme since line icons with a stroke width of 1 look about right at that size.

marclave (approved):
:O huge lets gooo ✨ nice work @hwkr

PR #1908, opened 2024-05-30
chore: fix all linter errors

hanspagel:
I just found out we removed ESLint from CI, I’m not even sure if this was intentional. Anyway, I think it’s good this way. It would have added some extra work in some PRs, without really preventing any bad errors.

That said, I feel better if we go through the errors and warnings every now and then, though. This PR fixes all ESLint errors and some of the warnings (mostly unused code or imports).

Note: It seems the play button doesn’t support the `data-proxy-url` API, because the relevant code isn’t used. I think it’s fine to remove it. It’s not mentioned in the README (for the play button), no one complained and you can always pass the proxy with the configuration object anyway.

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -28,6 +28,7 @@ module.exports = {
     '**/tailwind.config.ts',
     '**/vite.standalone.config.ts',
     '**/cdn/**',
+    '**/hydrateClient.d.ts',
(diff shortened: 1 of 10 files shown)


marclave (approved):
awesome chore @hanspagel and thanks for the detailed PR!

Issue #1765, opened 2024-05-16
Fix auto importing CSS


amritk:
### What happens?

Currently we use `vite-plugin-lib-inject-css` and it auto adds an import to the css files we don't have to. Its convenient but there are two cases where it doesn't work:
- when the script gets reloaded, as seen on angular
- SSR

So far we have been hacking around it by transpiling references and components on SSR. 

### What should we do?

We should either configure the `vite-plugin-lib-inject-css` or roll our own version which:
- does not auto import css on SSR
- only imports styles once

PR #1909, opened 2024-05-30
fix: HMR doesn’t work in examples/web

hanspagel:
With #1765 the resolve alias was removed from `examples/web`, this broke HMR. I think it was removed because the aliases broke for @scalar/api-reference/style.css and @scalar/api-client/style.css (we only had an exemption for @scalar/components/style.css)

This PR re-introduces the aliases and changes them to *not* match `@scalar/${anything}/style.css`.

Let’s test the deploy preview to see if the CSS is still fine and all.

### DIFF
--- a/examples/web/vite.config.ts
+++ b/examples/web/vite.config.ts
@@ -1,4 +1,5 @@
 import vue from '@vitejs/plugin-vue'
+import path from 'node:path'
 import { defineConfig } from 'vite'
 
 // https://vitejs.dev/config/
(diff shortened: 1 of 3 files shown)


amritk:
Actually it was removed on purpose! Just so we can test the packages properly since they behave differently with the alias, it was hiding some bugs. I think we should just have two vite configs one for dev and one for actual testing. What do you think @hanspagel?

hanspagel:
@amritk sounds good. can we merge this though? And follow with the testing config?

amritk:
Sure but just remember that this aliasing hides bugs sometimes

amritk (changes requested):
This PR is no longer relevant with #1940, we'll come up with another strategy

PR #1912, opened 2024-05-30
feat: do not include properties with unknown values in generated examples, fix #933

### LINKED ISSUE #933: Do not include properties with unknown values in generated code
### What happens?

When request body is generated, it includes properties that are not required and have no examples values. So it include these as `""` (empty string) if the type is string.

It should not include these properties, since it is very likely that an empty string is invalid. It also feels unnecessary to include these, since they are after all not required.

Create a schema with three string type properties, 

(A) with an example value and being required .
(B) without an example and not being required.
(C) with an example and not being required.

A and C should be included in generated request, B should not.

### Swagger/OpenAPI Example

_No response_

hanspagel:
Currently, we’re generating empty request bodies from the schema, which always includes all properties. Sometimes, we can prefill the fields with values, but mostly we can’t. This can be annoying when there’s a ton of optional properties (with invalid values).

With this PR only required properties and properties with an example value are added to the request body, others are omitted, as suggested in #933.

**Other changes**
* The request body didn’t show properties that have `required: true` as required (only if they were in `required: […]` one level above).
* With `additionalProperties` you can define arbitrary properties. We used to pick `someKey` as an example. I changed it to `{{key}}` because I think it’s more clear. It’s the variable syntax we use in other places already.

**Example**
```yaml
openapi: 3.1.0
info:
  title: Example
  version: 1.0.0
paths:
  '/planets':
    post:
      tags:
        - Planets
      summary: Create a planet
      requestBody:
        description: Planet
        content:
          application/json:
            schema:
              '$ref': '#/components/schemas/Planet'
      responses:
        '201':
          description: Created
          content:
            application/json:
              schema:
                '$ref': '#/components/schemas/Planet'
components:
  schemas:
    Planet:
      type: object
      required:
        - id
      properties:
        # example + required
        id:
          type: integer
          examples:
            - 1
        # example + required
        name:
          type: string
          examples:
            - Mars
          required: true
        # example + not required
        description:
          type:
            - string
            - 'null'
          examples:
            - The red planet
        # empty + not required
        size:
          type:
            - string

```

**Preview**

Note: There’s no `size` in the example request.

Screenshot 2024-05-30 at 18 03 30

### DIFF
--- a/.changeset/small-zoos-sell.md
+++ b/.changeset/small-zoos-sell.md
@@ -1,0 +1,6 @@
+---
+"@scalar/api-reference": patch
+"@scalar/oas-utils": patch
+---
+
+feat: omit empty and not required properties from the generated request body
--- a/packages/api-reference/src/components/Content/Models/ModelsAccordion.vue
+++ b/packages/api-reference/src/components/Content/Models/ModelsAccordion.vue
@@ -63,9 +63,8 @@ const { getModelId } = useNavState()
           :key="property"
           :name="property"
           :required="
-            schema.required &&
-            !!schema.required.length &&
-            schema.required.includes(property)
+            schema.required?.includes(property) ||
+            schema.properties?.[property]?.required === true
           "
           :value="value as OpenAPIV3_1.SchemaObject['properties']" />
       </div>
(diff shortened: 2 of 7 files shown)


marclave (approved):
DANG nice fix @hanspagel and tests 🤯 unreal

PR #1914, opened 2024-05-30
fix: sidebar webhooks

antlio:
**Problem**
webhooks sidebar heading doesn't get toggled when navigating over a webhook section, nor when we navigate toward webhook section url.

https://github.com/scalar/scalar/assets/14966155/9e08e635-be30-4a40-9896-14021ec1fd2f

**Solution**
this pr is a proposal in order to open webhooks sidebar heading group on a webhook section scroll.


marclave (approved):
nice fix @antlio ! ✨

PR #1915, opened 2024-05-30
fix: remove console log in use nav state

antlio:
missed a console log within the #1914 pr


marclave (approved):
haha i missed it also! : )

PR #1920, opened 2024-05-31
fix: markdown navigation

antlio:
**Problem**
right now clicking through description items in the sidebar doesn't navigate us to the right item.

**Solution**
this pr is a proposal to add ids to description headings to enable navigation.

### DIFF
--- a/.changeset/odd-pugs-rescue.md
+++ b/.changeset/odd-pugs-rescue.md
@@ -1,0 +1,5 @@
+---
+"@scalar/api-reference": patch
+---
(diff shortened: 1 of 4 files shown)


antlio:
as discussed with @hanspagel, will take another path to fix that as we have some already made logic here that could work!

marclave (approved):
wow awesome fix @antlio !!!

antlio:
finally we'll take another path here in order to setup section per heading which will make a two in one fix

PR #1921, opened 2024-05-31
chore(galaxy): remove headings from details tag

hanspagel:
We’ve had a special case in our example: Headings inside an `<details>` tag. Which isn’t really compatible with our way to split the Markdown content in sections. Fixing it (for example adding ids to the headings, not splitting the content in sections) would make it harder to show an accurate active state in the sidebar.

This PR just removes the headings from inside the `<details>` tag. Looks the same, but isn’t anymore in the sidebar, which is actually better IMHO.

### DIFF
--- a/.changeset/perfect-bags-swim.md
+++ b/.changeset/perfect-bags-swim.md
@@ -1,0 +1,5 @@
+---
+"@scalar/galaxy": patch
+---
+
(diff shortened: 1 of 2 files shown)


marclave (approved):
nice :)

PR #1922, opened 2024-05-31
chore(galaxy): add a preview script

hanspagel:
This PR adds a script to preview the `@scalar/galaxy` example and watch for changes. Great to see the results right-away while editing the example specification.


marclave (approved):
🔥

PR #1924, opened 2024-05-31
fix: markdown navigation (rollback)

hanspagel:
Reverts scalar/scalar#1920

It adds duplicate IDs, and we have a better solution already.


marclave (approved):
image

PR #1930, opened 2024-05-31
feat: move tailwind preset from `@scalar/components` to `@scalar/themes`

hwkr:
Took me a minute but I figured out how to do it

### DIFF
--- a/packages/themes/package.json
+++ b/packages/themes/package.json
@@ -37,16 +37,20 @@
     "./presets/default.css": "./dist/presets/default.css",
     "./presets/moon.css": "./dist/presets/moon.css",
     "./presets/purple.css": "./dist/presets/purple.css",
-    "./presets/solarized.css": "./dist/presets/solarized.css"
+    "./presets/solarized.css": "./dist/presets/solarized.css",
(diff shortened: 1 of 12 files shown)


packages/themes/package.json
```diff
@@ -37,16 +37,20 @@
...
     "./presets/purple.css": "./dist/presets/purple.css",
-    "./presets/solarized.css": "./dist/presets/solarized.css"
+    "./presets/solarized.css": "./dist/presets/solarized.css",
+    "./tailwind": "./src/tailwind.ts"
   },
   "files": [
     "dist",
-    "CHANGELOG.md"
+    "CHANGELOG.md",
+    "src/tailwind.ts"
```
hwkr, line 46: I don't know if this line is actually needed 🤔

packages/themes/src/tailwind.ts
```diff
@@ -125,5 +123,4 @@ export default {
       },
     },
   },
-  plugins: [headlessPlugin, colorMix()],
```
hwkr, line 128: These plugins should actually just be in `@scalar/components` in hindsight

marclave (approved):
lets go!

Issue #1624, opened 2024-05-06
Switching between response codes with multiple named examples causes an error


gsmcdonald:
### What happens?

An error occurs when switching between response codes with different names for examples

![image](https://github.com/scalar/scalar/assets/36003378/715ee1a7-227b-4f1f-89fb-1cd76bba68e2)

### What did you expect to happen?

The selected response code's example to show.

### How can we reproduce the issue?

1. Go to a request with multiple response codes where two response codes have different named examples (or number of examples).
2. Switching between response codes
3. An error will occur

The error will be in the console in the sandbox, but I'm not seeing the effects there (but I am seeing them when using a project at docs.scalar.com)

### Swagger/OpenAPI Example

https://sandbox.scalar.com/e/s4ysq

gsmcdonald:
I think the issue is that the `selectedExampleKey` value isn't reset when switching tabs, which causes an error when attempting to find that value in the examples object [here](https://github.com/scalar/scalar/blob/7254f6eae5c7fbacac52a18e10a71979369d1d11/packages/api-reference/src/components/Content/Operation/PathResponses/PathResponses.vue#L77).

I'm not sure if the expected behavior would be to clear the value so that the first response is selected when switching tabs or potentially just checking if the key exists in the `examples` object (in case you wanted to maintain state between response codes with the same responses).

hanspagel:
Great bug report! Let’s fix this!

marclave:
> @hanspagel What do you think about something like this? I haven't tested it because of some build issues, but I think the logic adds up.
> 
> ```
> /**
>  * Gets the first example response if there are multiple example responses
>  * or the only example if there is only one example response.
>  */
> const getFirstExampleResponse = () => {
>   if (!hasMultipleExamples.value)
>     return currentJsonResponse.value.example;
>   else if (Array.isArray(currentJsonResponse.value.examples))
>     return currentJsonResponse.value.examples[0]
>   else {
>     const firstProperty = Object.keys(currentJsonResponse.value.examples)[0];
>     return currentJsonResponse.value.examples[firstProperty];
>   }
> }
> 
> const currentResponseWithExample = computed(() => ({
>   ...currentJsonResponse.value,
>   example:
>     hasMultipleExamples.value && selectedExampleKey.value
>       ? currentJsonResponse.value.examples[selectedExampleKey.value].value ??
>         currentJsonResponse.value.examples[selectedExampleKey.value]
>       : getFirstExampleResponse(),
> }))
> 
[...]

PR #1934, opened 2024-06-03
fix: different response examples for different status codes

gsmcdonald:
Fixing issue where status codes having different example responses would cause an error.

**Problem**
Currently, an error occurs when switching between response codes with different names for examples. This is issue #1624 

**Explanation**
This happens because the `selectedExampleKey` value isn't reset when switching tabs, which causes an error when attempting to find that value in the examples object

**Solution**
With this PR, the `selectedExampleKey` value is reset when changing tabs and the logic to determine the first example has been updated to account for a single example, a list of examples, and a object of examples (named examples).

### DIFF
--- a/.changeset/light-snails-bathe.md
+++ b/.changeset/light-snails-bathe.md
@@ -1,0 +1,5 @@
+---
+"@scalar/api-reference": minor
+---
+
+fix: different response examples for different status codes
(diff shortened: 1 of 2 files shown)


marclave:
@gsmcdonald seems like we gotta run 

```
pnpm format
```

cause the build is failing : )

gsmcdonald:
> @gsmcdonald seems like we gotta run
> 
> ```
> pnpm format
> ```
> 
> cause the build is failing : )

Sorry about that @marclave. I think that should be fixed now. Thank you!

marclave (approved):
nice work @gsmcdonald !!!

PR #1935, opened 2024-06-03
fix: change listbox styles

cameronrohani:
before:
image

after:

https://github.com/scalar/scalar/assets/6201407/84cd2a75-129a-4e85-a06b-9d21cc42b288


hwkr (approved):
Did a little cleanup! Should be good to go!

marclave (approved):
we are so back!

PR #1937, opened 2024-06-03
fix: update dropdown styles

cameronrohani:
Create better variance between dropdown items + dropdown toggles
before:
image

after:
image


marclave (approved):
nice polish :)

PR #1940, opened 2024-06-04
feat: welcome the new client app 👀 

geoffgscott:
Merges the experimental client.

Side effects:
- Remove aliasing in api-client and api-references
- Slight reorganization of oas-utils


examples/web/src/components/MonacoEditor.vue
```diff
@@ -18,7 +18,7 @@ const props = withDefaults(
 // const isDark = useDarkMode(
 const model = defineModel<string>()
 
-const monacoEditorRef = ref<HTMLElement | null>(null)
+const monacoEditorRef: Ref<HTMLElement | null> = ref(null)
```
hanspagel, line 21: I’ve seen that kind of change a few times, is that different/better for some reason? Just wondering.

packages/api-reference/src/types.ts
```diff
@@ -7,13 +7,29 @@ import type {
...
+  title: string
+  extname: `.${string}` | null
+  default: string
+}
+
+export type AvailableTarget = TargetInfo & {
+  clients: ClientInfo[]
+}
+// ---------------------------------------------------------------------------
+
```
hanspagel, line 32: The types are exported through `httpsnippet-lite/dist/types/targets/targets`, aren’t they? 👀

hanspagel (changes requested):
I’m so excited to welcome the new API client in scalar/scalar!! 🤩  Just two things:

1. Do we want to publish under `@scalar/client-app`? From what I understand, we’d like to make it `@scalar/api-client@2.0`. Maybe we just add a `private: true` to the package.json until we settled on a name.

2. Can we keep the aliases until we have a better solution in place? Not having them at all, would really slow me down in development. 😬 If we need to make sure they are not used during build, we can add a conditional that only enables them in development (but probably in a seperate PR?).

examples/web/src/components/MonacoEditor.vue
geoffgscott: There is a Vue TS bug that throws if you do it using the generic. Very annoying.

packages/api-reference/src/types.ts
geoffgscott: Was not playing nice with the TS update (or the tsconfig change not sure exactly). But I also don't feel comfortable grabbing unexposed types deep down. Its annoying to have to manually declare them but if the original package changes then we will get compiler errors so it's at least safe.

packages/api-reference/vite.config.ts
```diff
@@ -78,18 +77,10 @@ export default defineConfig({
...
-        // @scalar/* -> packages/*/
-        // (not @scalar/*/style.css)
-        find: /^@scalar\/(?!(openapi-parser|snippetz|galaxy|themes\/style.css|components\/style\.css|components\b))(.+)/,
-        replacement: path.resolve(__dirname, '../$2/src/index.ts'),
-      },
-    ],
+    alias: {
+      '@': fileURLToPath(new URL('./src', import.meta.url)),
+    },
+    dedupe: ['vue'],
```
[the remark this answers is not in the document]
geoffgscott: I don't think it matters much either way. We can experiment with imports in another PR

PR #1943, opened 2024-06-05
chore: upgrade pnpm and set packagManager field in package.json

tmastrom:
upgrade pnpm and set packagManager field in package.json to avoid compatibility issues


hanspagel (changes requested):
Awesome! :) Keep in mind that the pnpm version is configured in the GitHub Action workflows, too. Here is one example:

https://github.com/scalar/scalar/blob/main/.github/workflows/ci.yml#L17-L19

amritk:
Wth I was just about to upgrade pnpm... Stop reading my mind @tmastrom

PR #1944, opened 2024-06-05
fix(#doc 1973): nuxt side menu basepath if set to ignores host

### LINKED ISSUE #1942: Nuxt: side menu basePath if set to '/' ignores host
### What happens?

when pathRouting basePath is set to '/' main navigation redirects to 

https://tag/menu-element

instead of 

https://host/tag/menu-element

correct redirect

https://codesandbox.io/p/devbox/nuxt-playground-forked-32ms33

click on any menu item 
image

### OpenAPI Document

_No response_

edit: I actually see that codesandbox redirects almost correctly just // after host

image

nav link is still incorrect
image

image

so that double / causes issue on localhost dev env unless we boot nuxt with custom host

edit2: culprit is here so with base path / it will be //id

https://github.com/scalar/scalar/blob/f82a76776be134c2f5263727fc912a287f8d330a/packages/api-reference/src/components/Sidebar/SidebarElement.vue#L48

amritk:
Fixes #1942


marclave (approved):
🕺

PR #1946, opened 2024-06-05
feat: scalar dropdown offset

antlio:
this pr introduces the offset prop in the scalar dropdown usage for better position capabilities.

### DIFF
--- a/.changeset/chilly-melons-approve.md
+++ b/.changeset/chilly-melons-approve.md
@@ -1,0 +1,5 @@
+---
+"@scalar/components": patch
+---
+
+feat: scalar dropdown and listbox offset prop addition


hwkr (approved):
Looks good! Should probably also expose it on ScalarPopover and ScalarListbox.

Also we could add a prop for shift while we're at it if you want.

antlio:
> Looks good! Should probably also expose it on ScalarPopover and ScalarListbox.
> 
> Also we could add a prop for shift while we're at it if you want.

good call! i think it'll be already possible once in for `ScalarPopover` as it uses `v-bind="$props"`

marclave (approved):
nice nice nice :)

PR #1948, opened 2024-06-05
feat: improve serverurl interaction in omnibar server page

antlio:
this pr introduces a first server view version in the client-app package as seen below:

image


marclave (approved):
I think the exact data structure on the servers might change a bit but happy to have a place for it in the mean time ✨

PR #1949, opened 2024-06-05
feat: support OAuth2 client credentials flow

mpminardi:
**Problem**
Currently, the client credentials flow for oauth is not supported.

**Explanation**
This happens because support has not yet been built out.

**Solution**
With this PR the client credentials flow is supported.

***Sample in UI***
image

***Sample in UI after successful auth***
image

### DIFF
--- a/packages/api-client/src/components/ApiClient/Request/RequestAuthentication/SecurityScheme.vue
+++ b/packages/api-client/src/components/ApiClient/Request/RequestAuthentication/SecurityScheme.vue
@@ -100,6 +100,15 @@ const handleOAuth2PasswordInput = (event: Event) => {
   })
 }
 
+const handleOpenAuth2ClientSecretInput = (event: Event) => {
+  setAuthentication({
+    oAuth2: {
+      ...authentication.oAuth2,
+      clientSecret: (event.target as HTMLInputElement).value,
+    },
+  })
+}
+
 function getOpenAuth2AuthorizationUrl(flow: any) {
   // https://example.com/oauth/authorize?
   //   response_type=token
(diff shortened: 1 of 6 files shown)


packages/api-client/src/components/ApiClient/Request/RequestAuthentication/SecurityScheme.vue
```diff
@@ -190,6 +199,74 @@ function authorizeWithPassword(
...
+
+  // TODO: Proxy this request
+  fetch(
+    options?.proxy
+      ? redirectToProxy(options?.proxy, url.toString())
+      : url.toString(),
+    {
+      method: 'POST',
+      headers: {
+        'Authorization': `Basic ${btoa(`${authentication.oAuth2.clientId}:${authentication.oAuth2.clientSecret}`)}`,
```
mpminardi, line 241: Currently have this as both basic auth and in the post form body as both are accepted for client credentials. Not sure if having both will cause problems for some implementations of the flow, so could potentially have some form of switch to pick between the two.

packages/api-client/src/components/ApiClient/Request/RequestAuthentication/SecurityScheme.vue
```diff
@@ -303,9 +380,50 @@ const startAuthentication = (url: string) => {
             Reset
           </CardFormButton>
         </template>
+        <CardFormRows>
```
mpminardi, line 383: Copied this section into the logic for the implicit flow as the `v-else` at the end was causing issues with rendering as it was previously acting as a fall-through case for both implicit and password. Can do refactoring if this is not desired.

packages/api-client/src/components/ApiClient/Request/RequestAuthentication/SecurityScheme.vue
```diff
@@ -368,6 +486,83 @@ const startAuthentication = (url: string) => {
           </CardFormGroup>
         </CardFormRows>
       </template>
+      <!-- Client Credentials Flow -->
+      <template
+        v-else-if="
+          (value as OpenAPIV3.OAuth2SecurityScheme).flows.clientCredentials
+        ">
+        <template v-if="authentication.oAuth2.accessToken">
```
mpminardi, line 494: Essentially blanket copy / pasted this from the implicit flow to give feedback on the fact that a token was successfully fetched / allow for reseting the token.

This currently looks like the following:
image

Can split this into a separate row / remove entirely for now / do something else if this looks a little too wonky.

amritk:
hey @mpminardi do you know of an example spec/server so we can test this? How were you testing?

mpminardi:
> hey @mpminardi do you know of an example spec/server so we can test this? How were you testing?

Hey @amritk! Had been testing this with [tailscale OAuth clients](https://tailscale.com/kb/1215/oauth-clients) which use the client credentials flow.

Have a sandbox link [here](https://sandbox.scalar.com/e/MBh70) which should have a minimal reproducer for being able to get a token / hit an endpoint with it.

Are authorization code / client credentials supported now as of https://github.com/scalar/scalar/pull/2215?

amritk:
Not in this client, but in the new one! Coming soon 😁

amritk:
Alright @mpminardi its out now! Give it a try on the new client. Authcode and client credentials are available on the client only for now so you won't see them on the api reference side (yet)

LordParsley:
> Alright @mpminardi its out now! Give it a try on the new client. Authcode and client credentials are available on the client only for now so you won't see them on the api reference side (yet)

@amritk any idea when we might see client credentials on the api reference? 🙏

https://docs.scalar.com/ & `"@scalar/api-reference": "^1.24.60",` don't render those inputs as far as I can tell:

**clientCredentials** ❌

Screenshot 2024-07-30 at 10 15 24

**implicit** ✅ 

Screenshot 2024-07-30 at 10 16 20

**password** ✅

Screenshot 2024-07-30 at 10 15 48

PR #1951, opened 2024-06-05
fix: client scrolling in sidebar + dont remove request type on hover

cameronrohani:
Fix scrolling in sidebar + button position

before:
image

after:
image


marclave (approved):
🕺

PR #1958, opened 2024-06-05
fix: client address bar

cameronrohani:
fix
- min width of address bar
- add empty state to address bar
- z-index of history
- fix checkmark padding in scalar list component


marclave (approved):
🥵

PR #1962, opened 2024-06-05
fix: client response UI

cameronrohani:
remove text inputs for headers response
reduce font size in body
change some border values etc

after:
image


marclave (approved):
CTO coming in with some clutch VUE

PR #1963, opened 2024-06-05
fix: missing headers and cookies on send request

marclave:
before we couldnt add headers or cookies to requests, we also break up the body appropriately now

### DIFF
--- a/packages/client-app/src/libs/sendRequest.ts
+++ b/packages/client-app/src/libs/sendRequest.ts
@@ -1,0 +1,92 @@
+import type {
+  RequestInstance,
+  RequestInstanceParameter,
+  RequestRef,
+  ResponseInstance,
+} from '@scalar/oas-utils/entities/workspace/spec'
+import type { AxiosError, AxiosRequestConfig } from 'axios'
+import axios from 'axios'
+
+const paramsReducer = (params: RequestInstanceParameter[] = []) =>
+  params.reduce(
+    (acc, param) => {
+      if (!param.key) return acc
+      acc[param.key] = param.value
+      return acc
+    },
+    {} as Record<string, string>,
+  )
+
+/**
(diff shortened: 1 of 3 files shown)


packages/client-app/src/libs/sendRequest.ts
```diff
@@ -0,0 +1,92 @@
+import type {
+  RequestInstance,
+  RequestInstanceParameter,
+  RequestRef,
+  ResponseInstance,
+} from '@scalar/oas-utils/entities/workspace/spec'
+import type { AxiosError, AxiosRequestConfig } from 'axios'
+import axios from 'axios'
+
+const paramsReducer = (params: RequestInstanceParameter[] = []) =>
```
amritk, line 10: would just add a short jsdoc comment here

amritk (approved):

packages/client-app/src/libs/sendRequest.ts
marclave: oh yeah let me add some comments : )

marclave: done

packages/client-app/src/libs/sendRequest.ts
```diff
@@ -7,6 +7,11 @@ import type {
 import type { AxiosError, AxiosRequestConfig } from 'axios'
 import axios from 'axios'
 
+/**
+ * Convert the parameters to an object
+ * @param RequestInstanceParameter
+ * @returns Record<string, string>
```
amritk, line 13: we actually don't need these when using typescript! Only useful if you want to actually describe the parameter, just the type can be inferred from typescript

marclave: removed we are so back

PR #1965, opened 2024-06-06
feat: Migrate to standardized build configs

geoffgscott:
Moves us to unified set of build configs for consistency (and hopefully improved tree-shaking)

- Externalizes ALL dependencies so nothing is bundled by default


amritk (changes requested):
Need to re-run pnpm i with the new version (9)

packages/build-tooling/src/build-options.ts
```diff
@@ -0,0 +1,144 @@
+import typescript from '@rollup/plugin-typescript'
+import { readFileSync } from 'fs'
+import type { InputPluginOption, Plugin, RollupOptions } from 'rollup'
+import type { BuildOptions, LibraryOptions } from 'vite'
+
+type StrictPluginOptions = RollupOptions & { plugins?: InputPluginOption[] }
+
+/**
+ * Generate rollup options for a Scalar package build
+ * DO NOT USE FOR CLIENT BUNDLING
```
hanspagel, line 10: Might be just me, but I don’t know what “client bundling” is. Can you help me to understand it? Or maybe we could use something like:

> Generate Rollup options to build packages
>
> Warning: This works for packages only, DO NOT use this to bundle something that is directly served to the user.

And the comment for `createViteBuildOptions` could be the same, as it’s for the same use case, but for Vite, right?

hanspagel (approved):
Looks awesome to me

geoffgscott:
> Need to re-run pnpm i with the new version (9)

Done and done. Also added pnpm to the engines so it errors if we are using the wrong one.

PR #1966, opened 2024-06-06
feat: handle file uploads in request body

marclave:
needs some polish from @cameronrohani :)

image

### DIFF
--- a/packages/client-app/src/libs/sendRequest.ts
+++ b/packages/client-app/src/libs/sendRequest.ts
@@ -36,28 +36,33 @@ export const sendRequest = async (
     }
   })
 
+  const headers = paramsReducer(instance.parameters.headers)
+
   let data: FormData | string | File | null = null
 
   if (instance.body.activeBody === 'binary' && instance.body.binary) {
+    headers['Content-Type'] = 'multipart/form-data'
+
     const bodyFormData = new FormData()
     bodyFormData.append(instance.body.binary.name, instance.body.binary)
+    data = bodyFormData
   } else if (instance.body.activeBody === 'raw') {
     data = instance.body.raw.value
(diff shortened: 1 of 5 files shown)


amritk (approved):
🪀

packages/client-app/src/views/Request/RequestSection/RequestTable.vue
```diff
@@ -71,17 +72,27 @@ const handleFileUpload = (idx: number) => {
       <div
```
antlio, line 72: we could use `DataTableCell` in here to maintain right table markup

packages/client-app/src/libs/sendRequest.ts
```diff
@@ -36,26 +36,34 @@ export const sendRequest = async (
...
+
   let data: FormData | string | File | null = null
 
   if (instance.body.activeBody === 'binary' && instance.body.binary) {
-    data = instance.body.binary
+    headers['Content-Type'] = 'multipart/form-data'
+
+    const bodyFormData = new FormData()
+    bodyFormData.append(instance.body.binary.name, instance.body.binary)
+    data = bodyFormData
```
geoffgscott, line 48: If the body type is binary then the fetch body is directly a blob. 

Multipart form can also be used for files but the structure is different and you can include mixed data (files and string). 

From a UX perspective if binary is selected there should just be a single file upload box and nothing else.

For multipart form we should show the DataTable and each value cell can be a file OR a string. 

It matter quite a bit with regards to importing spec definitions because servers have to handle multipart form data very differently.

geoffgscott (commented):
Looks good but are we handling Multipart form separately from Binary? It wasn't super clear from the Code/UX

marclave:
closing in favour of #1977

PR #1969, opened 2024-06-06
chore: update `@scalar/openapi-parser`

hanspagel:
This PR integrates some big updates to `@scalar/openapi-parser`. Here is what we’ll get from it:

**Fetch referenced files from URLs**

Let’s say you’d like to import the `Planet` schema from our Scalar Galaxy example:

```json
{
  "$ref": "https://galaxy.scalar.com/openapi.json#/components/schemas/Planet"
}
```

You can test this here: https://openapi-parser.pages.dev/

**Smaller bundle size**

We updated the build process of `@scalar/openapi-parser` and did a handful of optimizations, so we can eventually shave off ~200kb.

**Way faster**

The parser is just way, way faster now. It has more features, but is smarter than ever.

Note: Currently, when dereferencing the document is validated first, which is a pretty heavy process (takes ages, adds 200kb to the bundle). But we didn’t do anything with the results of the validation anyway, we only did a `console.log`. With that PR this step is skipped. The new `dereference` function is just derefencing, without any validation.

### DIFF
--- a/packages/oas-utils/src/transforms/import-spec.ts
+++ b/packages/oas-utils/src/transforms/import-spec.ts
@@ -7,7 +7,7 @@ import type { RequestMethod } from '@/helpers'
 import { parseJsonOrYaml } from '@/helpers/parse'
 import { schemaModel } from '@/helpers/schema-model'
 import type { AnyObject } from '@/types'
-import { openapi } from '@scalar/openapi-parser'
+import { dereference, load } from '@scalar/openapi-parser'
 import type { OpenAPIV3_1 } from 'openapi-types'
 
 const PARAM_DICTIONARY = {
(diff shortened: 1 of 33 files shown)


packages/oas-utils/src/transforms/import-spec.ts
```diff
@@ -153,7 +156,9 @@ export const importSpecToWorkspace = async (payload: ImportSpecPayload) => {
   const collection = createCollection({
     spec: {
       openapi: parsedSpec.openapi,
+      // @ts-expect-error For some reason the types don't match
       info: schema?.info,
+      // @ts-expect-error For some reason the types don't match
```
hanspagel, line 161: Those shouldn’t be necessary once #2045 is merged.

packages/use-codemirror/src/index.ts
```diff
@@ -1,5 +1,4 @@
 export { type Extension } from '@codemirror/state'
-// @ts-expect-error y-codemirror is not exporting types correctly
```
hanspagel, line 2: I don’t know why this is gone with this PR … but I’ll take it.

packages/oas-utils/src/transforms/import-spec.ts
hanspagel: gone

marclave (approved):
so many improvements, lets get this live!

PR #1972, opened 2024-06-06
fix(api-client): scoping request component styles

gsmcdonald:
**Problem**
Currently, styles can bleed into other applications.

**Explanation**
This happens because styles in `Request.vue` are not scoped, which can cause styles to bleed when Scalar is used in another application.

**Solution**
With this PR, the CSS rules in `Request.vue` will be scoped using the top-most class in the component (`scalar-api-client__main__left`).


gsmcdonald:
@marclave It looks like scoping the style block itself would cause issues with the styling, but I couldn't see any difference from this change when using the local demo.

packages/api-client/src/components/ApiClient/Request/Request.vue
```diff
@@ -92,318 +92,321 @@ const readOnly = true
...
-  color: var(--scalar-color-3);
-}
-.check input {
-  position: absolute;
-  opacity: 0;
-  cursor: pointer;
-  height: 0;
-  width: 0;
-}
+.scalar-api-client__main__left {
```
hwkr, line 96: I think since were using nesting here we might need to set `lang="postcss"` on the `<style>` block?

gsmcdonald: Done!

gsmcdonald:
Converting back to a draft PR while I cleanup my rebase issues.

hwkr (commented):
Hi @gsmcdonald! Sorry for the delay reviewing, looks like CSS nesting is now baseline as of 2023 so I was wrong and we probably don't actually need postcss here. I also gave the branch a test and it seems to be working as intended!

That being said we're actually about to overhaul the API client modal and move it to mostly tailwind css which should also solve the problem you were having here. I'm not actually sure what our timeline is for that so I'm going to let @marclave weigh in.

gsmcdonald:
> Hi @gsmcdonald! Sorry for the delay reviewing, looks like CSS nesting is now baseline as of 2023 so I was wrong and we probably don't actually need postcss here. I also gave the branch a test and it seems to be working as intended!
> 
> That being said we're actually about to overhaul the API client modal and move it to mostly tailwind css which should also solve the problem you were having here. I'm not actually sure what our timeline is for that so I'm going to let @marclave weigh in.

@hwkr No problem. I talked to a UX designer, and he told me that but that not all browser versions support it (so if that's a concern, postcss might still be a good idea). We would need to get it updated this week to stay on track. Will any of the classes/structure change with the overhaul?

marclave:
hey @gsmcdonald ! we are currently revamping our api client, let me talk with @cameronrohani on this right now and lets see if we can get this in :)

gsmcdonald:
Good afternoon, @marclave Any update on this?

marclave:
awesome work @gsmcdonald ! lets get this as its long over due ✨

marclave (approved):
lets go!

gsmcdonald:
@marclave Awesome. Thank you!

PR #1977, opened 2024-06-06
feat: add file uploads to request body

marclave:
continuation of https://github.com/scalar/scalar/pull/1966

needs some polish from @cameronrohani :)

image
image

### DIFF
--- a/packages/client-app/src/components/DataTable/DataTableCell.vue
+++ b/packages/client-app/src/components/DataTable/DataTableCell.vue
@@ -11,7 +11,7 @@ withDefaults(
 <template>
   <component
     :is="is"
-    class="min-h-[31px] min-w-0 border-b-[.5px] border-r-[.5px] border-r-[inherit] text-sm last:border-r-0 group-last:border-b-transparent"
+    class="min-h-[31px] h-[31px] min-w-0 border-b-[.5px] border-r-[.5px] border-r-[inherit] text-sm last:border-r-0 group-last:border-b-transparent"
     role="cell">
     <slot />
   </component>
--- a/packages/client-app/src/views/Request/RequestSection/RequestBody.vue
+++ b/packages/client-app/src/views/Request/RequestSection/RequestBody.vue
@@ -194,7 +194,7 @@ const handleFileUploadFormData = async (rowIdx: number) => {
         const updatedParams = [...currentParams]
         updatedParams[rowIdx] = {
           ...updatedParams[rowIdx],
-          binary: file,
+          file,
         }
         updateRequestExample(
           activeRequest.value.uid,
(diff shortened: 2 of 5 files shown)


packages/client-app/src/components/DataTable/DataTableCell.vue
```diff
@@ -11,7 +11,7 @@ withDefaults(
 <template>
   <component
     :is="is"
-    class="min-h-[31px] min-w-0 border-b-[.5px] border-r-[.5px] border-r-[inherit] text-sm last:border-r-0 group-last:border-b-transparent"
+    class="min-h-[31px] h-[31px] min-w-0 border-b-[.5px] border-r-[.5px] border-r-[inherit] text-sm last:border-r-0 group-last:border-b-transparent"
```
amritk, line 14: I strongly suggest we don't use pixel values!!! Its an anti-pattern

amritk: just snap to the grid 😄

packages/client-app/src/views/Request/RequestSection/RequestBody.vue
```diff
@@ -228,15 +314,28 @@ const handleFileUpload = (idx: number) => {
...
-                <span>Upload File</span>
-                <ScalarIcon
-                  class="ml-1"
-                  icon="Upload"
-                  size="xs" />
-              </ScalarButton>
+              <template v-if="activeExample?.body.binary">
+                <span class="text-c-2">{{
+                  activeExample?.body.binary.name
+                }}</span>
```
amritk, line 320: nit:
```suggestion
                <span class="text-c-2">
                  {{  activeExample?.body.binary.name  }}
                </span>
```

amritk (approved):
Left comment about the px thing, rest looks good

packages/client-app/src/components/DataTable/DataTableCell.vue
marclave: @antlio added this 👀 can we add a lint rule for this?

amritk: Well there are rare occasions where it's ok. But like 99% if the time we shouldn't. Just need to get all the devs on board!

If we are doing it a lot then our config is wrong and it should be added there.

marclave:
worked with @geoffgscott on the binary file upload option, we will fast follow with removing axios here and using fetch : )

packages/client-app/src/components/DataTable/DataTableCell.vue
antlio: not disagreeing but snapping the grid wouldnt maintain height consistency here.

i'll give a global eye on this and take another path to reach both 👍

amritk: Then we adjust some other spots. If everything is lined up then you don't need use any pixel values

PR #1981, opened 2024-06-07
chore: delete @scalar/api-client-proxy

hanspagel:
I don’t think we’ll need the source for the old `@scalar/api-client-proxy` anymore. This PR removes it from the repository.


marclave:
@hanspagel we should probably keep it here for posterity, also we are still technically supporting it for another 5 months waas the plan before we removed it : )

but lets remove it EOY ✨

PR #1982, opened 2024-06-07
chore: enable cache for dev command

hanspagel:
With the switch from resolve aliases to builds, it’s necessary to restart `pnpm run dev` to see cross package changes. 

Currently, this takes close to a minute to restart, which makes me lose the focus every time I want to see a change in action.

Turns out, we disabled the cache for the `dev task` and everything is built again (even the packages that didn’t change). There might be a reason that I don’t know of, but if we’d use the cache, it would be faster. ⚡

### DIFF
--- a/turbo.json
+++ b/turbo.json
@@ -7,7 +7,6 @@
       "dependsOn": ["^build"]
     },
     "dev": {
-      "cache": false,
       "persistent": true,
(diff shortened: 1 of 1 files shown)


marclave:
closing in favour of https://github.com/scalar/scalar/pull/1991

PR #1983, opened 2024-06-07
feat: use example values for path variables

hanspagel:
For the example requests, variables are replaced from `{curlyBrackets}` to `__UNDERSCORES__`, because curly brackets would be URL encoded.

Anyway, if we an example value is provided, we could actually use the example value instead. 🤔 

**Preview**

Note the `planets/123` in the example request:

Screenshot 2024-06-07 at 12 14 52

**Example**

```yaml
openapi: 3.1.0
info:
  title: Scalar Galaxy
  version: 1.0.0
paths:
  '/planets/{planetId}':
    get:
      summary: Get a planet
      parameters:
        - name: planetId
          in: path
          schema:
            examples:
              - 123
```

### DIFF
--- a/.changeset/proud-pots-work.md
+++ b/.changeset/proud-pots-work.md
@@ -1,0 +1,6 @@
+---
+"@scalar/oas-utils": patch
+"@scalar/api-reference": patch
+---
+
+feat: use example values for path variables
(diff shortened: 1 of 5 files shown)


marclave (approved):
🕺 nice work @hanspagel

PR #1984, opened 2024-06-07
feat(api-client): better error output for failed requests

hanspagel:
If an request fails, we just show an empty request body. With this PR the Axios error message is shown.

As an alternative, we could also add some min-height or show a special component for empty response bodies. But I’m always staring at the response body and would love to see if something goes wrong right-away.

**Before**
Screenshot 2024-06-07 at 12 38 38

**After**
Screenshot 2024-06-07 at 12 36 29


marclave (approved):
nice : )

PR #1985, opened 2024-06-07
fix(api-client): URL is not encoded when sent to the proxy

hanspagel:
Currently, we just pass the request URL to the proxy (https://proxy.scalar.com?scalar_url=OUR_TARGET_URL_HERE) and this works for most cases.

It turns out, not in all cases. :) Actually we need to encode the target URL when adding it to the `scalar_url` parameter, otherwise we’ll get an error from the proxy:

**✅ Works (regular URL)**
https://proxy.scalar.com/?scalar_url=https://example.com/

**❌ Doesn’t work (has a whitespace)**
https://proxy.scalar.com/?scalar_url=https://example.com/?query=apple+banana

**✅ Works (encoded)**
https://proxy.scalar.com/?scalar_url=https%3A%2F%2Fexample.com%3Fquery%3Dapple%2Bbanana

### DIFF
--- a/.changeset/brave-islands-explode.md
+++ b/.changeset/brave-islands-explode.md
@@ -1,0 +1,5 @@
+---
+"@scalar/api-client": patch
+---
+
+fix: URL is not encoded when sent to the proxy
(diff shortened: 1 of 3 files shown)


geoffgscott (commented):
@marclave Can you confirm this won't cause any Go issues? Sometimes server frameworks aren't all that good about handling encoding the way you expect (I'm looking at you Flask).

hanspagel:
As far as I understand it, characters *can* be percentage encoded, but don’t have to be encoded. That’s why it was working in most cases. With more encoding, it just works in more/all cases. And Go seems to understand the percentage based encoding well, I’ve tested a bunch of other characters.

PR #1988, opened 2024-06-07
fix(api-reference): default example response is always the first example

hanspagel:
Currently, we’re always moving the `default` example response to the first position. Turns out, most people use `default` as a fallback and it doesn’t really make sense to have it as the first example response.

With this PR, all example responses are ordered alphabetically (new!) and default just comes after all numbers (new!).

Wdyt @marclave?

More context: https://github.com/scalar/scalar/discussions/1148


marclave (approved):
i dig it : )

PR #1991, opened 2024-06-07
chore: add dev web command

amritk:
Unfortunately its not that great, but better than what we had before! Just use this for development on the web example. Maybe we'll add watch next week!


marclave (approved):
nice :)

PR #2000, opened 2024-06-09
fix: handle extrict minimum and maximum properties

Amorim33:
Fix https://github.com/scalar/scalar/discussions/1956

## Problem
Currently, `min` and `max` equal to 0 restrictions are not rendered, also the `extrictMinimum`and `extrictMaximum`properties are not handled. 

image

## Explanation

This happens because `0` is falsy in Javascript, so the condition isn't satisfied and the min|max property is not rendered.
Also components for "extrictMinimum" and "extrictMaximum" are missing.

**Solution**
With this PR the condition to check if `value.minimum`or `value.maximum` are defined were refactored to check if the values are different from `undefined`, so "min=0" satisfies the condition.
"greater than" and "less then" schema property details were added to handle extrict minimum and maximum.

image

> ⚠️ "greater than" and "less than" were chose to avoid breaking changes, but it would be also possible to follow the Redoc's solution.
image

## OpenAPI Properties

```json
"balance": {
  "type": "integer",
  "minimum": 0,
  "exclusiveMinimum": true,
  "description": "Saldo devedor em centavos."
},
"remainingTerm": {
  "type": "integer",
  "minimum": 0,
  "description": "Prazo remanescente em meses."
},
```

## Reference
https://swagger.io/docs/specification/data-models/data-types/

### DIFF
--- a/packages/api-reference/src/components/Content/Schema/SchemaPropertyHeading.vue
+++ b/packages/api-reference/src/components/Content/Schema/SchemaPropertyHeading.vue
@@ -86,11 +86,23 @@ const rules = ['oneOf', 'anyOf', 'allOf', 'not']
       <SchemaPropertyDetail v-if="value.format">{{
         value.format
       }}</SchemaPropertyDetail>
-      <SchemaPropertyDetail v-if="value.minimum">
+      <SchemaPropertyDetail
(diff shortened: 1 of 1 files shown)


marclave (approved):
@Amorim33 UNREAL PR :) great work ✨

PR #2002, opened 2024-06-10
feat: request sidebar dropdown position

antlio:
this pr adds teleport capability to the scalar floating component in order to enhance side menu position in client app package:

image

### DIFF
--- a/packages/components/src/components/ScalarFloating/ScalarFloating.vue
+++ b/packages/components/src/components/ScalarFloating/ScalarFloating.vue
@@ -59,7 +59,7 @@ const { floatingStyles, middlewareData } = useFloating(targetRef, floatingRef, {
   placement: computed(() => props.placement),
   whileElementsMounted: autoUpdate,
   middleware: computed(() => [
-    offset(props.offset || 5),
+    offset(5),
     flip(),
     shift(),
     ...(props.middleware ?? []),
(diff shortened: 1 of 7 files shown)


packages/components/src/components/ScalarFloating/types.ts
```diff
@@ -17,9 +17,8 @@ export type FloatingOptions = {
    */
   middleware?: Middleware[]
   /**
-   * The offset distance between the floating element and its reference element.
-   * This can be used to adjust the spacing or alignment of the floating element.
-   * @default 5
+   * Whether the floating element is open or not.
+   * @default false
    */
-  offset?: number
```
antlio, line 24: removing this prop as being finally not needed

packages/components/src/components/ScalarFloating/ScalarFloating.vue
```diff
@@ -72,15 +72,19 @@ const { floatingStyles, middlewareData } = useFloating(targetRef, floatingRef, {
...
-    class="relative z-context"
-    :style="floatingStyles"
-    v-bind="$attrs">
-    <slot
-      :data="middlewareData"
-      :height="targetHeight"
-      name="floating"
-      :width="targetWidth" />
-  </div>
+  <Teleport
```
hwkr, line 75: I think we should make the teleport opt-in since it can cause some accessibility and styling issues. You can use the [`disabled` prop](https://vuejs.org/guide/built-ins/teleport#disabling-teleport) to have it turned off by default but then have a prop to turn it on.

packages/components/src/components/ScalarFloating/ScalarFloating.vue
```diff
@@ -72,15 +72,19 @@ const { floatingStyles, middlewareData } = useFloating(targetRef, floatingRef, {
...
+    v-if="isOpen"
```
hwkr, line 76: Do we need to have a v-if for this? I think headless should handle the rendering of the element

packages/components/src/components/ScalarFloating/ScalarFloating.vue
```diff
@@ -59,7 +59,7 @@ const { floatingStyles, middlewareData } = useFloating(targetRef, floatingRef, {
   placement: computed(() => props.placement),
   whileElementsMounted: autoUpdate,
   middleware: computed(() => [
-    offset(props.offset || 5),
+    offset(5),
```
hwkr, line 62: Did you mean to remove this prop?

marclave (approved):
nice work @antlio 

once @hwkr 's feedback is address we can get this merged ✨

packages/components/src/components/ScalarFloating/ScalarFloating.vue
antlio: updated accordingly, using `teleport` with enable it from now on

PR #2012, opened 2024-06-10
feat: address bar server interaction

antlio:
this pr favors scalar dropdown component usage for client app address bar server list in order to display call to action:

image


marclave (approved):
nice work @antlio 🥇

PR #2013, opened 2024-06-10
fix: add important to response header tabs

marclave:
before:
image

after:
image


cameronrohani (approved):
this seems important

hwkr:
Hmmmm this probably shouldn't need the important there... Also it's not broken for me on preview, checking dev now

Google Chrome-2024-06-10-13-24-23@2x

hwkr:
Looks like a CSS load order issue, we can just bump the specificity for the tab override.

PR without the `!important` #2014

PR #2015, opened 2024-06-10
fix: autoimport css in standalone

amritk:
This fixes the auto importing of css. The ID's were missing per app so it thought the style was already loaded.


marclave (approved):
nice fix! tested and validated ✨

geoffgscott (approved):

PR #2016, opened 2024-06-10
chore: remove @scalar/use-modal package

hwkr:
Removes the `@scalar/use-modal` package and migrate all usages to `ScalarModal` in `@scalar/components`.

@hanspagel has already [deprecated the NPM package](https://www.npmjs.com/package/@scalar/use-modal).

There's a couple issues still with scrolling the search modal and history modals but _I think_ those should be fixed in #2015.

This is at least an improvement since the modals actually render on screen again.


marclave (approved):

geoffgscott:
LGTM lets go

PR #2018, opened 2024-06-11
feat: api client modal package

amritk:
This PR stands up the API client modal package. Its not complete but its working! I figure we can merge this here then follow up with moving the addressbar + setting config, routes etc (ez stuff with the clientRouter)

image

test with `pnpm dev:modal`

Also this is the ESM version so it should work in any framework!


marclave (approved):
the future is here today, awesome work @amritk i can't wait to get this fully swapped.

packages/components/src/components/ScalarModal/ScalarModal.vue
```diff
@@ -37,7 +37,7 @@ const modal = cva({
       sm: 'max-w-screen-sm',
       md: 'max-w-screen-md',
       lg: 'max-w-screen-lg',
-      full: 'mt-0',
+      full: 'mt-0 overflow-hidden',
```
antlio, line 40: what did you encounter here?

packages/api-client-modal/playground/main.ts
```diff
@@ -1,5 +1,5 @@
 import { createScalarClient } from '@/api-client-modal'
-import content from '@scalar/galaxy/latest.json?raw'
+import content from '@scalar/galaxy/latest.json'
```
antlio, line 2: add to add this back to test locally

amritk: hmmm will ts-ignore

PR #2019, opened 2024-06-11
fix: client app fixtures

antlio:
this pr fixes a list of client app package ui issues:
- duplicated border
- wrong focus state
- missing overflow
- shrinking profile icon


marclave (approved):
🕺

PR #2020, opened 2024-06-11
fix: make custom scroll :global again

hwkr:
Looks like something with our changes in build tooling changed how we're parsing CSS modules and we can't use `:global` the way we were before.

We can probably change how we're doing this in the future but for now this PR fixes the regression.

## Before / After

https://github.com/scalar/scalar/assets/6374090/30c5ca0e-da47-46be-a75a-7744e16cf887

### DIFF
--- a/packages/themes/src/components/ScrollbarStyles.vue
+++ b/packages/themes/src/components/ScrollbarStyles.vue
@@ -13,62 +13,62 @@ useApplyClasses('#headlessui-portal-root', scrollbars)
 </template>
 <style lang="postcss" module>
 @layer scalar-utilities {
-  .scrollbars :global {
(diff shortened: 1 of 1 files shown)


cameronrohani (approved):

amritk (changes requested):
Can we move this to a css file? then if you want the component you can import the component, else you can import the css file?

hwkr:
> Can we move this to a css file? then if you want the component you can import the component, else you can import the css file?

Yep! There's actually a ticket for that, but this PR is just to fix the regression

amritk (approved):

marclave (approved):
:)

PR #2021, opened 2024-06-11
chore: expand standarized build tooling

geoffgscott:
- Improves the rollup config with support for json, yaml, css, and automatic package exports updating
- Adds support for aliasing and a standard alias export function 
- Reorganized the code-highlight package a bit


marclave (approved):
nice work @geoffgscott !

PR #2024, opened 2024-06-11
fix: release build

amritk:
Currently our release build is broken due to this reference to the removed `use-modal` package. This should fix it


marclave (approved):
huge!

PR #2025, opened 2024-06-11
wip: export css files instead of vue components

hwkr:
Alright this is very work in progress but here's kind of what I've been trying to do with the `@scalar/themes` exports.

Problems:
* I want to import `reset.module.css` as a CSS module in `ApiReferenceLayout.vue` but it doesn't seem to like that. Vite says it supports CSS modules natively(?) but I can't tell if there's something we need to install / configure.
* I was going to just create a little function to inject our stylesheets into the `<head>` but I don't think that will work in SSR so I made a `<Style>` component and I don't know where it should live (we probably need to use it in a number of places including `api-references` and `api-client`).
* We have duplicate versions of reset in order to have a (scoped) module reset and a root (target's `html, body`) reset. It would be nice to use `@import` or something but I don't know how to make that work in vite.
* Probably some other stuff I haven't thought of yet


amritk:
Closed in favour of 
https://github.com/scalar/scalar/pull/2135

PR #2031, opened 2024-06-12
feat: client integration

amritk:
Kinda working!


hanspagel (changes requested):
> dont merge until do you do the feature flag!
> 
> OP

hanspagel:
Added a feature flag:

> To enable the new API client, copy `packages/api-reference/.env.example` to `packages/api-reference/.env`.

marclave (approved):
:) 

going to temporarily increase bundle size, but that will get slimmed down once we remove the old client soon™️ !

PR #2034, opened 2024-06-12
chore: remove Makefile

hanspagel:
With the new dev setup I don’t use the Makefile anymore. This PR removes it. ✨ Less files in root = good. :)


marclave (approved):
i defs miss my makefile but makes sense ✨

PR #2036, opened 2024-06-12
fix: api client modal action modal fix

antlio:
this pr fixes the client app action modal style for the api client modal:

image


marclave (approved):
🕺

PR #2039, opened 2024-06-12
chore: upgrade vitest coverage

tmastrom:
Breaking out some work that I did while down the rabbit hole trying to fix the the `themes` snapshot testing. 

This PR re-enables most of the themes testing except for two tests that using `toMatchFileSnapshot()` which still requires some work to fix our environment. 

Also fixes the tsconfig path in the playwright package and upgrades `@vitest/coverage-v8` to `1.6.0` fixing some peer dependency warnings.


marclave (approved):
👯 nice @tmastrom

PR #2041, opened 2024-06-12
fix: zod url doesnt allow relative url for parsing

marclave:
**Problem**
openapi can have relative URLs for servers, which a good use case is for our integrations which are hosted docs on the api server

**Explanation**
zod url doesnt allow for relative url

**Solution**
we just make sure its a valid string now


amritk (approved):
yolo

PR #2045, opened 2024-06-13
feat(client-app): add defaults for all resources and deep merge

amritk:
We should pretty much just remove zod at some point since we aren't really parsing (everythings optional)


geoffgscott (approved):

geoffgscott:
> We should pretty much just remove zod at some point since we aren't really parsing (everythings optional)

We should not remove zod. Optional properties are still type checked so we can ensure they match the type at runtime. Otherwise the follow could happen on a HTTP request return or something:

```ts
type MyObj = {
  name?: string
}

// Invalid object. Will pass nullish checks but with the wrong type
myObj = {
  name: 10
}


```

marclave (approved):

PR #2046, opened 2024-06-13
fix: make sidebar responsive

cameronrohani:
R e s p o n s i v e
https://github.com/scalar/scalar/assets/6201407/6e8d4a0a-382e-4d0e-a57a-dc85f3889f2b

Responsive

### DIFF
--- a/packages/client-app/src/views/Request/Request.vue
+++ b/packages/client-app/src/views/Request/Request.vue
@@ -253,16 +253,20 @@ const getBackgroundColor = () => {
       </div>
     </div>
     <ViewLayout>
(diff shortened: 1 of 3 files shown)


marclave (approved):
🔥

packages/client-app/src/views/Request/Request.vue
```diff
@@ -339,4 +346,14 @@ const getBackgroundColor = () => {
 .request-text-color:active {
   color: var(--scalar-color-1);
 }
+@media screen and (max-width: 780px) {
```
antlio, line 349: should we use `@screen` here instead?

PR #2047, opened 2024-06-13
fix: remove click events from http verb + adding server in address bar

cameronrohani:
Removed click events and chevron from HTTP verb
image

and removed the add server
image


marclave (approved):
unreal work @cameronrohani

cameronrohani:
> unreal work @cameronrohani

image

PR #2050, opened 2024-06-13
feat(client-app): add missing descriptions to param tables

marclave:


marclave:
@cameronrohani verbally approved

PR #2051, opened 2024-06-13
fix: checkbox hover style

cameronrohani:
image

just like the good ol days


marclave (approved):
for some reason i feel like we are just getting started, nice work @cameronrohani

PR #2053, opened 2024-06-13
feat(client-app): add example payload to request body

marclave:
things left in this PR

image

- [x] need to match contentType to the workspace active body


packages/client-app/src/views/Request/RequestSection/RequestBody.vue
```diff
@@ -266,7 +289,7 @@ watch(
     </template>
     <template v-else>
       <DataTable :columns="['']">
-        <DataTableRow>
+        <!-- <DataTableRow>
```
antlio, line 292: this has been commented out until we support all content type

marclave:
awesome work getting this over the finish line @antlio !!

PR #2057, opened 2024-06-13
feat: client app localhost request

antlio:
this pr update client app send requests lib to prevent proxy usage in case of localhost usage as done previously.


marclave (approved):
nice :)

PR #2064, opened 2024-06-13
feat: hot-reload in oas-utils

amritk:
Added hot reloading to oas-utils since we make SO many changes in here when working on the client.

Converted to draft because it doesn't seem to be working with commonjs

Update: we are so close!


packages/oas-utils/package.json
```diff
@@ -115,5 +57,97 @@
   },
   "peerDependencies": {
     "axios": "^1.5"
+  },
+  "imports": {
```
amritk, line 61: all auto-generated

amritk:
the world was not ready for this

PR #2065, opened 2024-06-14
feat: port tests to sendRequest in new client app

marclave:
just missing 2! but we can add those once we get variable replacement done 
image


amritk (approved):
gogogo

PR #2073, opened 2024-06-14
feat(client-app): client app reponse meta infos

antlio:
this pr adds response meta informations in the client app:

image


marclave (approved):
absolutely cooking @antlio !!!

PR #2076, opened 2024-06-14
chore(client-app): switch server form fields

hanspagel:
It’s bugging me, that the optional field is coming first. And I don’t think it’s a description, it’s more a label or something. And the placeholder wasn’t really a good fit.

**Before**
Screenshot 2024-06-14 at 12 22 32

**After**
image


marclave (approved):
I dig it :)

PR #2077, opened 2024-06-14
fix(client-app): preview and raw are swapped

hanspagel:
“Preview” shows the raw data and “Raw” shows the formatted and highlighted code, let’s fix this:

**Before**
Screenshot 2024-06-14 at 12 30 25

**After**
Screenshot 2024-06-14 at 12 38 49
Screenshot 2024-06-14 at 12 38 45


hanspagel:
I still think that needs to be tackled, but we got important stuff to do right now. I plan to come back to that branch at some point.

PR #2078, opened 2024-06-14
fix(client-app): response body is not pretty printed

hanspagel:
The response body doesn’t have line breaks. This PR updates the `ScalarCodeBlock` to pretty print JS objects, arrays and JSON strings.

Note: This is fixed already, but I’ve also merged two pretty print functions. :)

**Before**
Screenshot 2024-06-14 at 12 38 45

**After**
Screenshot 2024-06-14 at 13 13 23


marclave (approved):
🕺

PR #2079, opened 2024-06-14
fix(proxy-server): only some headers are shown

hanspagel:
Since we switched to the new proxy, only some headers are shown. By default, only some headers can be accessed for cross-origin requests.

This PR adds the `Access-Control-Expose-Header` with a `*` wildcard value to allow JS to access all headers.

Read more: https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS#access-control-expose-headers

**Preview**
Screenshot 2024-06-14 at 13 32 49

EDIT: This must be a flaky test? Can you take a look @geoffgscott?

>  FAIL  |@scalar/object-utils| src/mutator-record/mutations.test.ts > Handles history rolling > Rolls history back to initial then forward to modified state
AssertionError: expected { storage: [ { …(3) }, …(2) ], …(2) } to not deeply equal { storage: [ { …(3) }, …(2) ], …(2) }


marclave (approved):
😮‍💨 niiiiice @hanspagel

PR #2080, opened 2024-06-14
fix: api client modal search

antlio:
this pr fixes the api reference search being display on hotkey pressed while api client modal is open in order to displays its own.

### DIFF
--- a/packages/api-reference/src/components/api-client-bus.ts
+++ b/packages/api-reference/src/components/api-client-bus.ts
@@ -5,3 +5,4 @@ const apiClientBusKey: EventBusKey<OpenClientPayload> = Symbol()
 
 /** Event bus to open the API Client */
 export const apiClientBus = useEventBus(apiClientBusKey)
+export const modalStateBus = useEventBus('modalState')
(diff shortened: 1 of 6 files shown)


marclave (approved):
very smart fix @antlio ! just needs a changeset then we can merge :)

packages/api-reference/src/components/api-client-bus.ts
```diff
@@ -5,3 +5,4 @@ const apiClientBusKey: EventBusKey<OpenClientPayload> = Symbol()
 
 /** Event bus to open the API Client */
 export const apiClientBus = useEventBus(apiClientBusKey)
+export const modalStateBus = useEventBus('modalState')
```
marclave, line 8: lets call this searchModalState

marclave (changes requested):

packages/api-reference/src/components/api-client-bus.ts
antlio: being the api client modal state, should we still use search here?

marclave: ah lets just keep it nvm : )

marclave (approved):

PR #2082, opened 2024-06-14
chore: remove `pack-packages` script

hanspagel:
The `pack-packages` script is broken for a while, because some packages don’t have a version number. It seems no one is using this anymore, so let’s remove it.


geoffgscott (approved):
@amritk we don't need this for the linking?

amritk (changes requested):
I still use this script for testing! I'll just fix it

PR #2083, opened 2024-06-14
feat(client-app): request select

antlio:
this pr favors scalar listbox component over select in request view:

image


marclave (approved):
🕺

PR #2084, opened 2024-06-14
feat(client-app): hotkey request sidebar toggle

antlio:
this pr enables request sidebar toggle on ctrl/cmd key + b.


marclave (approved):
oo i like this :)

marclave:
we should probably add some form of hover effect as well

PR #2091, opened 2024-06-14
feat: new `@scalar/api-reference-editor` package

geoffgscott:
Adds an editor package to the repo to enable spec editing with internal or external state. 

Derived from: https://github.com/scalar/scalar/pull/1490

Also includes a fix to remove the npx call of `@scalar/cli` during build tooling which is dangerous.


hanspagel:
I’ve extracted the new @scalar/galaxy build script changes to #2118, rebased the PR and added a changeset. ✌️

packages/api-reference-editor/src/components/EditorInput.vue
```diff
@@ -0,0 +1,90 @@
...
+</script>
+
+<template>
+  <div
+    ref="codeMirrorRef"
+    class="editor-input">
+    <slot></slot>
+  </div>
+</template>
+<style>
```
marclave, line 55: lets scope this :)

packages/api-reference/src/types.ts
```diff
@@ -250,7 +250,19 @@ export type ReferenceLayoutSlot =
   | 'sidebar-start'
   | 'sidebar-end'
 
+export type ReferenceLayoutSlots = {
+  [x in ReferenceLayoutSlot]: (props: ReferenceSlotProps) => any
```
marclave, line 254: do we need to specify any here? maybe a comment would help : )

packages/api-reference/src/components/Layouts/Layouts.vue
```diff
@@ -0,0 +1,39 @@
+<script setup lang="ts">
+/**
+ * This component allows dynamic selection of various layout configurations
+ */
+import type { ReferenceLayoutProps, ReferenceLayoutSlots } from '../../types'
+import ClassicLayout from './ClassicLayout.vue'
+import ModernLayout from './ModernLayout.vue'
+
+// eslint-disable-next-line vue/no-unused-properties
```
marclave, line 9: should we just pick the props we use from this?

marclave (changes requested):
i get an error for `[vite-node] [ERR_LOAD_URL] ./nuxt-theme.css` on nuxt 

also we should add some documentation to root README (even if just a link to the package readme) and remove our `isEditable` prop

packages/api-reference/src/components/Layouts/Layouts.vue
geoffgscott: No we are passing props through.

packages/api-reference/src/types.ts
geoffgscott: I'm not sure this is previous code that was moved.

packages/api-reference-editor/src/components/EditorInput.vue
geoffgscott: Done.

marclave (approved):

PR #2098, opened 2024-06-14
feat: Add vite plugin to restart dev server when workspace files change

geoffgscott:
Adds a vite plugin to watch for workspace file changes. When files in package `dist` folders change we restart the vite server automatically.


marclave (approved):
1 less context switch! let's goo

PR #2118, opened 2024-06-17
chore(galaxy): use Node script instead of bash script

hanspagel:
Currently, we use a bash script and the CLI to build/format the `@scalar/galaxy` package. If this breaks, for some reason, it’s hard to fix and blocks CI.

This PR uses a Node script instead. If that breaks and blocks CI, it’s easy to modify.

Note: This is extracted from #2091.


marclave (approved):
nice :) tested and works!

PR #2120, opened 2024-06-17
feat(docusaurus): add yaml support

hanspagel:
Currently, the docusaurus plugin fetches the OpenAPI document, but supports JSON only. If someone tried to add a YAML url, this exception was thrown:

```
> docusaurus start --port=5063 --no-open

[INFO] Starting the development server...

[ERROR] SyntaxError: Unexpected token 'o', "openapi: 3"... is not valid JSON
    at JSON.parse (<anonymous>)
```

With this PR we’re removing the code to fetch the OpenAPI document, so `@scalar/api-reference` can fetch the document and we all get YAML support for free. :)

### DIFF
--- a/.changeset/old-knives-smell.md
+++ b/.changeset/old-knives-smell.md
@@ -1,0 +1,5 @@
+---
+'@scalar/docusaurus': patch
+---
+
+feat: add Yaml support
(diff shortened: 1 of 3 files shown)


marclave:
@hanspagel dont we need to pass in the configuration still? : )

hanspagel:
Oh damn, yes, of course!

marclave:
closing in favour of #2155

PR #2122, opened 2024-06-17
docs: detailed Docusaurus integration guide

hanspagel:
This PR adds a new integration guide, just for Docusaurus. :)


documentation/docusaurus.md
```diff
@@ -0,0 +1,76 @@
...
+```bash
+pnpm add @scalar/docusaurus
+```
+
+Done! There’s just one more step required: Adding the plugin to your Docusaurus configuration.
+
+```ts
+// docusaurus.config.js (or .ts if you’re project is in TypeScript)
+import { ScalarOptions } from '@scalar/docusaurus'
+
```
antlio, line 51: should we describe both js/ts config examples? this would throw an error if not using ts so it might be misleading

hanspagel: It had a comment helping with TypeScript/JavaScript, but I just added both examples. :)

marclave (approved):
🕺

PR #2124, opened 2024-06-17
chore: enable minification again

hanspagel:
The CDN broke and as a quick fix we disabled minifcation for the CDN/standalone build in #2108.

This PR enables it again and a lot more. :)

This PR:
* Enables minification again
* Configures Terser so it doesn’t break the build
* Runs `@scalar/api-reference` jsdelivr tests hourly
* Runs `@scalar/api-reference` standalone build tests for every PR
* Comes with a CDN example for `@scalar/play-button`
* Fixes the `@scalar/play-button` build

Currently, we’re using the file from the fastify dist folder. With that PR we’re using the `@scalar/api-reference` and `@scalar/play-button` builds directly.

**Fast follow**

* Snapshot tests are disabled. I’ll enable them again.
* The tests for `@scalar/play-button` are disabled, they need to be written, but I didn’t bother as they’d fail for the CDN build that’s published right now anyway. Once that’s fixed (with this PR) we can ff with a proper test.
* We need a notification when the scheduled test fails.


packages/api-reference/vite.standalone.config.ts
```diff
@@ -38,7 +38,18 @@ export default defineConfig({
...
+    // * They’re using terser, too.
+    // * Copying their options fixes the build.
+    // * `max_line_len: 80` is the one setting that makes the difference.
+    //
+    // Source: https://github.com/highlightjs/highlight.js/blob/b9ae5fea90514b864f2c9b2889d7d3302d6156dc/tools/build_config.js#L58-L73
+    terserOptions: {
+      format: {
+        max_line_len: 80,
+      },
+    },
```
hanspagel, line 52: That’s the fix. 😅

geoffgscott (approved):
Approved but the new example folder should really be moved to @scalar/api-reference/playground. 

Ideally we should just fork the CDN deploy into its own package.

marclave (approved):
🕺

PR #2126, opened 2024-06-17
feat(client-app): section navigation

antlio:
this pr fallbacks on 'all' section when navigating to an endpoint without body if the latter is active + displays body on delete http method.

### DIFF
--- a/packages/client-app/src/views/Request/RequestSection/RequestSection.vue
+++ b/packages/client-app/src/views/Request/RequestSection/RequestSection.vue
@@ -11,7 +11,7 @@ import { computed, ref, watch } from 'vue'
 
 const { activeRequest, activeExample } = useWorkspace()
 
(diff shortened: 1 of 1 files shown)


packages/client-app/src/views/Request/RequestSection/RequestSection.vue
```diff
@@ -37,6 +37,16 @@ const sections = computed(() => {
 type ActiveSections = (typeof sections.value)[number]
 
 const activeSection = ref<ActiveSections>('All')
+
+watch(activeRequest, (newRequest) => {
+  console.log(newRequest)
```
hanspagel, line 42: That’s just debug output, isn’t it? Should we remove it?

antlio: omg yes 🐝

marclave (approved):
🌟

PR #2129, opened 2024-06-17
chore: do not run ci on node 21

tmastrom:
**Problem**
Currently, our CI is taking a long time.

**Explanation**
This happens because we are testing on node 18, 20, and 21

**Solution**
With this PR we will only run testing in CI on node 18 (maintenance) and 20 (LTS)


marclave (approved):
we are so back 🫡

PR #2135, opened 2024-06-17
feat(#doc-1996): themes css exports

amritk:
Scope css under `.scalar-app` instead of the layers so we don't get beat out by users resets.

I started from @hwkr's PR and just added a bit more scoping 

Decided to leave fonts and themes for another ticket, so things remaining in this one:
- [x] adding class to headless ui root
- [x] tailwind custom classes


packages/themes/src/style.css
```diff
@@ -0,0 +1,8 @@
+/* Import core stylesheets  */
+@import 'reset.css';
+@import 'scrollbars.css';
+@import 'fonts.css';
+
+/* Import a copy of the base variables and default theme */
+@import 'variables.css';
```
hwkr, line 7: I think we want to put these in a layer here no?

hwkr (commented):
Left a couple comments! I'll give it a test in the AM if that's cool

packages/themes/src/style.css
amritk: The specificity was too low in the layer, also I think we are scrapping all layers

packages/components/src/tailwind/tailwind.css
```diff
@@ -1,42 +1,32 @@
+@import '@scalar/themes/style.css';
+
 /**
```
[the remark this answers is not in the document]
amritk: true! Hey I got the tailwind nesting working too (not in this PR)

cameronrohani (approved):
great work

hanspagel (dismissed):
Sorry, I’m not able to do a full review, but just saw that the `DefaultFonts` component is removed which is responsible to handle the font loading. Is there something else in place already or do we break the `withDefaultFonts` setting? 👀

packages/themes/src/style.css
hwkr: Well the specificity shouldn't matter for the variables, they just need to be set

hwkr (approved):
Cool beans, we can tweak this more later if needed

hwkr:
@hanspagel the font loading is now handled in [`getThemeStyles`](https://github.com/scalar/scalar/pull/2135/files#diff-b13c2501faaf239caa960ed5c3b2f833b369e977ac692d3898415cae7b24bc52R115) but the loading functionality is preserved.

marclave (approved):
lets goooo

PR #2143, opened 2024-06-18
fix(client-app): display hide icon in reference only

antlio:
this pr conditions the hide action icon from the client app in order for it to be displayed only when used within a reference as a client modal.

**before**
image


marclave (approved):
nice :)

PR #2145, opened 2024-06-18
feat(client-app): sidebar toggle hotkey

antlio:
currently we have to key up and key down both keys in order to toggle the sidebar, however sometimes we might keep the meta key down to have a glimpse on the navigation.

this pr is a proposal to favor useEventListener over whenever, i haven't been able to reach the same result otherwise.

### DIFF
--- a/packages/client-app/src/views/Request/Request.vue
+++ b/packages/client-app/src/views/Request/Request.vue
@@ -17,7 +17,7 @@ import type { DraggingItem, HoveredItem } from '@scalar/draggable'
 import type { Collection } from '@scalar/oas-utils/entities/workspace/collection'
 import { REQUEST_METHODS, type RequestMethod } from '@scalar/oas-utils/helpers'
 import { isMacOS } from '@scalar/use-tooltip'
-import { useMagicKeys, whenever } from '@vueuse/core'
+import { useEventListener, useMagicKeys } from '@vueuse/core'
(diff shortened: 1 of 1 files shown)


packages/client-app/src/views/Request/Request.vue
```diff
@@ -211,10 +211,12 @@ const getBackgroundColor = () => {
...
-whenever(
-  isMacOS() ? keys.meta_b : keys.ctrl_b,
-  () => (showSideBar.value = !showSideBar.value),
-)
+
+useEventListener(document, 'keydown', (event) => {
+  if ((isMacOS() ? keys.meta.value : keys.ctrl.value) && event.key === 'b') {
+    showSideBar.value = !showSideBar.value
+  }
+})
```
amritk, line 219: we can probably factor this into a helper later so it works everywhere

amritk (approved):
What were you thinking on meta press? a modal which shows all of the hotkeys?

packages/client-app/src/views/Request/Request.vue
amritk: also would be easier to change if we need to

amritk (changes requested):
Now i'm not sure anymore, ping me!

amritk (approved):

marclave (approved):
lets test it out!

PR #2147, opened 2024-06-18
chore(cdn-api-reference): simplify cdn server

hanspagel:
I saw the CDN example had code to serve all static files already, so there’s no need to register all routes individually.


marclave (approved):
✨

PR #2148, opened 2024-06-18
ci: [revert] use GitHub bot to create release PRs

hanspagel:
Reverts scalar/scalar#2117

Doesn’t seem to work, GitHub Actions says: `HttpError: Resource not accessible by integration`


marclave (approved):
hansbot is so back

PR #2149, opened 2024-06-18
ci: send a slack message for all test jsdelivr workflow runs

hanspagel:
WIP


hanspagel:
Doesn’t work. Im mildly annoyed.

PR #2150, opened 2024-06-18
feat(client-app): environment name editable

antlio:
this pr enables is a first addition in order to be able to edit non default environment name in the client app:

image


marclave (approved):
🕺 nice @antlio !

PR #2152, opened 2024-06-18
feat(next): URL support

hanspagel:
I just found out that our Next.js integration doesn’t have support for URLs as promoted in the README:

```diff
import { ApiReference } from '@scalar/nextjs-api-reference'

const config = {
  spec: {
+    url: 'https://cdn.jsdelivr.net/npm/@scalar/galaxy/dist/latest.json',
  },
}

export const GET = ApiReference(config)
```

This PR adds support for URLs.

Before, the plugin always passed `content`, but ignored `url`. 

This PR also adds support for YAML content, the CDN version doesn’t supports it yet though (not if it’s passed as content of the `script` tag, only if it’s in the configuration object).

### DIFF
--- a/.changeset/gorgeous-sheep-check.md
+++ b/.changeset/gorgeous-sheep-check.md
@@ -1,0 +1,5 @@
+---
+'@scalar/nextjs-api-reference': patch
+---
+
+feat: add URL support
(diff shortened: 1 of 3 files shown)


marclave (approved):
i dig this! : ) we might want to move this to a helper function since its kind of agnostic of framework, and we can use it for our similar integrations ✨

PR #2153, opened 2024-06-18
fix: support commonjs docusaurus builds

marclave:
**Problem**
If you are using an older version of docusaurus it breaks our api reference react package

**Explanation**
because we dont handle commonjs

**Solution**
we load the standalone versions since we arent doing ssr anyways :)


marclave:
releasing to test ✨

PR #2155, opened 2024-06-18
fix: dont parse json from fetch just as text

marclave:
**Problem**
we .json() so we cant load yaml

**Explanation**
This happens because yaml cant be parsed as json

**Solution**
we just grab it as text :)


hanspagel (approved):
We could also just pass the URL to the frontend and let the client fetch the document, but this should do the trick already. Let’s get this in!

PR #2159, opened 2024-06-18
fix(standalone): HTML API does not parse YAML content

hanspagel:
WIP Tests are failing. :)

I saw that the HTML API does not support YAML content inside the `<script>` tag, this PR should fix it.

I’ve added some more test cases for the various ways to pass content to the CDN build. 

Once we released this fix here, we can enable the jsdelivr test for YAML content inside the `script` tag, too.

### DIFF
--- a/.changeset/blue-pillows-drop.md
+++ b/.changeset/blue-pillows-drop.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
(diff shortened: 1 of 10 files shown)


marclave:
closing in favour of #2155

hanspagel:
This is something completely different! This is for the HTML API, not just Docusaurus. 👍

marclave (approved):
nice work @hanspagel 👯

PR #2162, opened 2024-06-18
fix: standalone always defaults to light mode

marclave:
**Problem**
you could not force initial dark mode

**Explanation**
because lightmode was always added

**Solution**
With this PR we check config on initial set


amritk (approved):
Send it for now but eventually we'll default in the css

PR #2166, opened 2024-06-19
chore: add build watcher tooling

geoffgscott:
Adds some tooling for automated builds. There are probably some edge cases if you are quickly saving in multiple packages so don't do that. 

Basic usage:
```bash
# From workspace root start the build watcher
pnpm build:watch @scalar/working-package
```

```bash
# Run the local dev environment. Usage of the vite-workspace-restart plugin is recommended
pnpm --filter @scalar/working-package dev

```


packages/build-watch/package.json
hanspagel, line 1: Does that have to be a package? 🤔 We don’t use it as a dependency somewhere. We could consider to move it to `/tooling` and leave it as a script.

hanspagel (changes requested):
Thanks for making our dev setup even nicer!

When I run `pnpm build:watch @scalar/oas-utils` and change something in `oas-utils`I see the following in the logs:

```test
Building @scalar/play-button
-------------------------------------------------------------------
[WARN]: @scalar/play-button cancelled. [object Object]
```

What does that mean? 🤔 

I can’t really test it, though. It makes my computer unusable. `kernel_task` is going crazy. Might be the file watcher? Do we need to exclude node_modules or something?

Screenshot 2024-06-21 at 16 22 46

Ah, it’s actually Node processes (makes sense):

image

marclave:
unfortunately this is an old one, gunna close this cause surprisingly not tooo many complaints with the new flow in awhile :)

PR #2167, opened 2024-06-19
fix(api-client): reset style sunset

antlio:
following #2135 merge, reset style wrapper could be now removed.


marclave (approved):
nice! tested and validated ✨

PR #2168, opened 2024-06-19
ci: type check only once

hanspagel:
Some tasks don’t need to be run in Node 18 and 20, I think `types:check` is one of them.

With this PR it’s only executed for Node 20, not for Node 18.

As this takes ~1m it should speed up CI a little bit.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,7 +34,8 @@ jobs:
         run: pnpm format:check
       - name: Build
         run: pnpm turbo build
-      - name: Check types
+      - if: matrix.node-version == 20
+        name: Check types
         run: pnpm turbo types:check
       - name: Run tests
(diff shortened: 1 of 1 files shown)


marclave:
shouldnt we check the types on all versions? i got hosed by File type not being supported in version 18 and CI saved me :)

will defer to @geoffgscott and @amritk

amritk:
@marclave node doesn't run typescript so I don't think it makes a difference. The version that matters is whats specified in your tsconfig.

hanspagel:
For those rare cases we can run it on the release PRs in Node 18 and 20. wdyt

hanspagel:
Rebased and updated: Runs on Node 20 by default and on Node 18 and 20 in the release PR.

marclave (approved):
cool :)

PR #2170, opened 2024-06-19
fix(docusaurus): doesn’t support (relative) URLs

hanspagel:
We brought back the CommonJS build for Docusaurus yesterday, but I think we broke URLs.

I’ve refactored the server side function to do the following:
* Execute a callback for the content, if provided
* Delete the `spec.content` attribute from the configuration and add it as content to the script tag (good, because we just use raw JSON/YAML here):
  * strings are just printed as they are
  * objects are JSON.stringified
* The remaining configuration is added as an attribute

Note: With this PR URLs are fetched in the frontend, which adds support for relative URLs (see https://github.com/scalar/scalar/discussions/2171). To fetch the document on the server side, just add your fetch call to the configuration and pass the result as `spec.content`. :)

This PR also comes with examples for the various ways to pass content to Docusaurus.

Screenshot 2024-06-19 at 14 16 08

### DIFF
--- a/packages/docusaurus/src/index.ts
+++ b/packages/docusaurus/src/index.ts
@@ -18,15 +18,6 @@ const ScalarDocusaurus = (
     name: '@scalar/docusaurus',
 
     async loadContent() {
-      // Check if we need to download a spec
-      if (options.configuration?.spec?.url) {
-        const resp = await fetch(options.configuration.spec.url)
-        const content = await resp.text()
-        return {
-          configuration: { ...options.configuration, spec: { content } },
-        }
-      }
-
       return options
     },
 
--- a/.changeset/silly-beds-retire.md
+++ b/.changeset/silly-beds-retire.md
@@ -1,0 +1,5 @@
+---
+'@scalar/docusaurus': patch
+---
+
+feat: add support for relative URLs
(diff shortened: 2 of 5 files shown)


hkosova:
Suggestion: maybe the Docusaurus plugin should pass its `url` param (if provided) to [`useBrokenLinks().collectLink(...)`](https://docusaurus.io/docs/docusaurus-core#useBrokenLinks) so that invalid relative URLs will be reported during `docusaurus build` by Docusaurus' built-in link checker.

packages/docusaurus/src/index.ts
```diff
@@ -18,15 +18,6 @@ const ScalarDocusaurus = (
     name: '@scalar/docusaurus',
 
     async loadContent() {
-      // Check if we need to download a spec
```
marclave, line 21: lets make sure we dont delete this :)

marclave (changes requested):

marclave:
i also get these errors :) 

but seems like my comment on the loadContent was just a cached build

image

hanspagel:
@marclave this one is fixed in #2159 :)

hanspagel:
> Suggestion: maybe the Docusaurus plugin should pass its `url` param (if provided) to [`useBrokenLinks().collectLink(...)`](https://docusaurus.io/docs/docusaurus-core#useBrokenLinks) so that invalid relative URLs will be reported during `docusaurus build` by Docusaurus' built-in link checker.

I love that! Unfortunately, I’m not able to get that to work properly. 😵 I’ll try to submit a separate PR for it.

marclave (approved):
🫡

PR #2173, opened 2024-06-19
fix(components): scalar icon ref

antlio:
this pr favors shallow ref over ref usage in scalar icon component and ensures that only the reference to the icon component is reactive + fix log errors.


marclave:
@antlio what are the "reactive + fix log errors" :)

marclave (approved):
i see the errors lets get this in 😅

PR #2175, opened 2024-06-19
fix(client-app): global style updates

antlio:
this pr favor scalar-app parent class over layer following #2135.

⚠️ current `@scalar/themes` tailwind preset config is breaking the usage of responsive modifiers + some utilities, this pr is not to be merge before including an update on the color utility namings.


amritk:
Ready for review, I didn't fix the popups and menus and what not since we aren't using them on the modal. Can deal with those later. But everything else should be good to go. The font-size for xxs has grown `11px` -> `12px` so there is a  slight difference there

packages/client-app/src/assets/tailwind.css
```diff
@@ -1,18 +1,78 @@
...
+      --tw-translate-x: -50%;
+      top: 50%;
+      --tw-translate-y: -50%;
+      transform: translate(var(--tw-translate-x), var(--tw-translate-y))
+        rotate(var(--tw-rotate)) skewX(var(--tw-skew-x)) skewY(var(--tw-skew-y))
+        scaleX(var(--tw-scale-x)) scaleY(var(--tw-scale-y));
+    }
   }
+
+  text-rendering: optimizeLegibility;
```
amritk, line 70: @hwkr do we need these here? Are these client specific or should they come from the main reset?

packages/client-app/src/components/DataTable/DataTableCell.vue
```diff
@@ -11,7 +11,7 @@ withDefaults(
 <template>
   <component
     :is="is"
-    class="h-8 min-w-0 border-b-1/2 border-r-1/2 border-r-[inherit] text-sm last:border-r-0 group-last:border-b-transparent w-fit min-w-full has-[input[type=checkbox]]:w-8 has-[input[type=checkbox]]:min-w-8"
+    class="h-8 border-b-1/2 border-r-1/2 text-sm last:border-r-0 group-last:border-b-transparent"
```
antlio, line 14: does this maintain square on checkbox cell + hover?

packages/client-app/src/components/DataTable/DataTableInput.vue
```diff
@@ -94,3 +96,10 @@ const handleDropdownMouseUp = () => {
       @select="handleSelect" />
   </DataTableCell>
 </template>
+
+<style scoped>
+/* Tailwind placeholder is busted */
+input::placeholder {
```
antlio, line 102: do we style it already from the theme reset style?

packages/components/src/components/ScalarFloating/ScalarFloating.vue
```diff
@@ -76,15 +76,17 @@ const { floatingStyles, middlewareData } = useFloating(targetRef, floatingRef, {
...
-    <div
-      ref="floatingRef"
-      class="relative z-overlay"
-      :style="floatingStyles">
-      <slot
-        :data="middlewareData"
-        :height="targetHeight"
-        name="floating"
-        :width="targetWidth" />
+    <div class="scalar-app">
```
antlio, line 79: wondering if using scalar-components would make sense here or not

packages/client-app/src/assets/tailwind.css
hwkr: That's a great question, I'd just do whatever the [Tailwind preflight](https://github.com/tailwindlabs/tailwindcss/blob/next/packages/tailwindcss/preflight.css) does.

packages/client-app/src/components/DataTable/DataTableInput.vue
hwkr: I think so https://github.com/scalar/scalar/blob/main/packages/themes/src/reset.css#L62

packages/components/src/components/ScalarFloating/ScalarFloating.vue
hwkr: I could but we would then have to add more classes to the reset and stuff (probably not worth it).

packages/client-app/index.html
```diff
@@ -26,7 +26,7 @@
   <body>
     <div
       id="app"
-      class="app"></div>
+      class="app scalar-app scalar-client"></div>
```
hwkr, line 29: Do we want a class of `.app` at all?

packages/client-app/src/components/DataTable/DataTableCell.vue
hwkr: Seems to? 

Google Chrome-2024-06-19-16-30-26@2x

packages/client-app/src/components/DataTable/DataTableInput.vue
amritk: yea but didn't show up. You wanna work ur magic brynn? or we can just leave it, not the biggest thing 🤷🏾

packages/client-app/index.html
amritk: not really, don't think it does anything

packages/components/src/components/ScalarFloating/ScalarFloating.vue
amritk: the system currently works!

marclave (approved):
tested and works :)

packages/client-app/src/components/DataTable/DataTableInput.vue
hwkr: Oh weird, I'll look

packages/client-app/src/components/DataTable/DataTableCell.vue
antlio: it's not well squared anymore, will add it back!

PR #2179, opened 2024-06-19
fix: remove layers in docusaurus and fix precedence issue

marclave:
**Problem**
docusaurus has a precedence error with the show more icon, we can fix it properly later and we also forgot to remove the layers


cameronrohani (approved):
nice

PR #2182, opened 2024-06-20
fix: bring back request history for client app

marclave:
I think we can make this better... but here for feedback!

some ideas:

- [ ] we move the request history off of activeRequest for storing responses and link it to each example
- [ ] do we replace the example or enter a sort of draft state (replace response + request)

we can also think about this post modal integration

https://github.com/scalar/scalar/assets/6176314/6e77a32f-35db-4cca-980e-790befb9fdf9


antlio:
awesome work! wondering if we should check for active open tab so it doens't open a new one for the same history

antlio (approved):

PR #2184, opened 2024-06-20
fix(client-app): teleport fixtures

antlio:
this pr fixes broken styles in teleported components as seen below:

image

### DIFF
--- a/packages/components/src/components/ScalarFloating/ScalarFloating.vue
+++ b/packages/components/src/components/ScalarFloating/ScalarFloating.vue
@@ -76,7 +76,7 @@ const { floatingStyles, middlewareData } = useFloating(targetRef, floatingRef, {
     v-if="isOpen"
     :disabled="!teleport"
     to="body">
-    <div class="scalar-app">
+    <div class="scalar-app scalar-client">
       <div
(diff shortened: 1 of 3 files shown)


amritk (approved):
Let's go!

packages/components/src/components/ScalarFloating/ScalarFloating.vue
```diff
@@ -76,7 +76,7 @@ const { floatingStyles, middlewareData } = useFloating(targetRef, floatingRef, {
     v-if="isOpen"
     :disabled="!teleport"
     to="body">
-    <div class="scalar-app">
+    <div class="scalar-app scalar-client">
```
hwkr, line 79: This probably shouldn't be hard coded into the components library 🤔 Could we add a prop or something?

antlio: so `scalar-client` would be pass from the client app to the components using scalar floating under the hood then?

hwkr: Yeah exactly!

antlio:
closing in favor of #2209

PR #2187, opened 2024-06-20
ci(update-contributors): do not auto approve contributor update PRs

hanspagel:
We used to auto approve and merge the contributor update PRs and I wanted to fix that. But it somehow feels wrong to auto skip the queue, so let’s just remove it and manually merge the PRs. If it’s too many, we can change the schedule from daily to weekly or twice a week.


marclave (approved):
:)

PR #2188, opened 2024-06-20
feat(api-reference): reinstate new client modal

antlio:
this pr reinstate new client modal after being disable in #2136 following following #2135.


packages/api-reference/src/components/ApiReferenceLayout.vue
```diff
@@ -129,9 +129,9 @@ onBeforeMount(() => {
       ),
     )
     if (headlessRoot) {
-      ;(headlessRoot.addedNodes[0] as HTMLDivElement).classList.add(
-        'scalar-app',
-      )
+      const el = headlessRoot.addedNodes[0] as HTMLDivElement
```
amritk, line 132: I think we have this a few places now, lets factor it out to a common util

amritk (approved):

marclave (approved):
we are SO back 🕺

PR #2191, opened 2024-06-20
feat(client-app): disable button while requesting

antlio:
this pr disables the send button while a request is being sent.


amritk (approved):
Yup thats it

marclave (approved):
this is great! i would sometimes spam the button but this is a much better solution ✨

PR #2192, opened 2024-06-20
fix(client-app): add data table cell checkbox style back

antlio:
this pr adds back the checkbox style on data table cell component for the client app:

image


marclave (approved):
we are so back :)

PR #2193, opened 2024-06-20
fix(client-app): update input outline style

antlio:
this pr fixes some outline in the client-app package.

image


marclave (approved):
ah @cameronrohani and i were just talking about this ✨

PR #2194, opened 2024-06-20
feat(use-tooltip): update

antlio:
this pr update use tooltip package usage and dependencies, and fixes the following:

image


marclave (approved):
niiiice ✨

PR #2196, opened 2024-06-20
docs: update the package table

hanspagel:
Oops, there are two broken links in the README. Let’s update the package table and remove swagger-editor and swagger-parser, and add a few new ones. I’ve also gone through the description and made them all shorter.

And you know? We can switch to relative links. That’s way cooler, because it points to the files in the same branch/PR then.

### DIFF
--- a/README.md
+++ b/README.md
@@ -182,7 +182,7 @@ export default defineNuxtConfig({
 })
 ```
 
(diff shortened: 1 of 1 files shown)


marclave (approved):
image

PR #2198, opened 2024-06-20
feat: allow custom header & footers

cameronrohani:
This PR allows for users to easily add custom headers and footers by removing overflow:auto from `.api-reference` and instead relying on `<html>` for scrolling

We're already had to take this approach on Docusaurus so I think it will make sense to move forward with this approach for scalar

In order to support scalar users who have hacked adding a header I created `--scalar-y-offset` which applies the offset on how far scalar is pushed down on the page and assumes that is the desired header height. I imagine we'll remove this in a few months. 

Here's how a user will add a header (I'll add an example to the CDN as soon as this gets in 😄 )
image

### DIFF
--- a/.changeset/honest-geese-chew.md
+++ b/.changeset/honest-geese-chew.md
@@ -1,0 +1,7 @@
+---
+'@scalar/api-reference': patch
+'@scalar/client-app': patch
+'@scalar/docusaurus': patch
+---
+
+feat: allow custom headers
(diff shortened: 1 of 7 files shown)


marclave (approved):
🫡 @amritk approved verbally and @hwkr was chatting w cam on this, lets get it in!

amritk:
Just toss the comment in explaining it. I will probably touch it when I go back and fix lazy loading

PR #2199, opened 2024-06-21
feat(client-app): add implicit flow to the client app

amritk:
Gets the implicit flow for auth going. This has a bunch of code for testing in it, so we should remove that before merging but its good for testing!


.eslintrc.cjs
```diff
@@ -97,7 +97,7 @@ module.exports = {
         ],
         '@typescript-eslint/prefer-optional-chain': 'warn',
         'camelcase': 'off',
-        'max-len': ['warn', { code: 120, comments: null }],
```
amritk, line 100: I only added this for comments lol, removing as this is invalid, and we already have prettier 😄

packages/client-app/src/store/workspace.ts
```diff
@@ -534,9 +534,32 @@ async function importSpecFile(spec: string | AnyObject) {
       SecurityScheme
     >,
   ).forEach(([key, securityScheme]) =>
-    securitySchemeMutators.add(
-      createSecurityScheme({ ...securityScheme, uid: key }),
-    ),
+    // TODO remove Temp for testing
```
amritk, line 537: going to leave this testing code in, until auth is complete

marclave (approved):
🥵 🥵 🥵 🥵 🥵 🥵 🥵 

unreal @amritk, can remove testing code and get this in

PR #2200, opened 2024-06-21
fix(express-api-reference): explicitly set content-type to text/html

soulchild:
**Problem**
When used together with security middlewares like [Helmet](https://helmetjs.github.io/) which disable content sniffing, or manually setting the header `X-Content-Type-Options` to `nosniff`, the API documentation is rendered as plain text.

**Explanation**
`@scalar/express-api-reference` doesn't set any content-type.

**Solution**
This PR explicitly sets the content-type of the initial HTML response to `text/html`.

### DIFF
--- a/.changeset/clever-mice-kick.md
+++ b/.changeset/clever-mice-kick.md
@@ -2,4 +2,4 @@
 '@scalar/express-api-reference': patch
 ---
 
-Fix missing content-type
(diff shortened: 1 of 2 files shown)


marclave (approved):
awesome @soulchild ! thanks for fixing this ✨

PR #2201, opened 2024-06-21
fix(components): tailwind specificity

antlio:
this pr adds the `:where` pseudo element to tailwind component stylesheet in order to maintain low specificity on component usages + fixes api reference style on new api client modal activation.

image


marclave (approved):
👯

PR #2202, opened 2024-06-21
docs: trim down root README

hanspagel:
Hot take: The root README is too long.

This PR moves most of the content to the package READMEs or separate Markdown files in the `documentation/` folder. The separate files need love, too, but let’s focus on the root README in this PR.

Other changes:
* I’ve updated the features list in the root README
* I’ve made the `ci.yml` name shorter, so the badge becomes smaller

Some links are broken now, because they point to `main` and the new files aren’t in `main` yet. But I think we have to use absolute URLs for Markdown files that are published on npmjs.com.

Preview: https://github.com/scalar/scalar/blob/de76754db35942c206858c2143428ceeaeb90d27/README.md

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,4 +1,4 @@
-name: Check build, types, and units tests are passing
+name: Check build, types, and units tests
 
 on:
   pull_request:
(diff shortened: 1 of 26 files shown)


marclave (approved):
huge improvement :)

PR #2203, opened 2024-06-21
chore: remove ESLint `max-len` for comments

hanspagel:
We recently introduced a warning for code lines and comments longer than 120 characters. 

This highlights a lot of the existing comments, and I don't think it makes much sense to manually wrap them to fit into that length. It’s just one thing that I need to manually deal with and think about, especially when updating comments and adding content somewhere before the line break.

For some coments, it doesn’t make sense to wrap them:

Screenshot 2024-06-21 at 12 18 55

And if someone is annoyed by the long comments, there’s a setting in VS Code (and other editors) to automatically wrap lines.

Long code lines = complexity, long comments = valuable information.

I’d say, let’s keep the limit for the code, but not for the comments. :)

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -97,7 +97,7 @@ module.exports = {
         ],
         '@typescript-eslint/prefer-optional-chain': 'warn',
         'camelcase': 'off',
-        'max-len': ['warn', { code: 120, comments: 120 }],
+        'max-len': ['warn', { code: 120, comments: null }],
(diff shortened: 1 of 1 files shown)


marclave (approved):
agreed :)

amritk:
Ah I added this specifically as we are going to start enforcing js-doc comments and they look much better wrapped! But I will just manually wrap my own instead if we don't like it.

PR #2204, opened 2024-06-21
feat(cli): new `scalar void` command

hanspagel:
This adds a `void` command to the CLI to quickly boot up the new `@scalar/void-server`. :)

Convenient to quickly spin up a dummy API.


marclave (approved):
:)

PR #2205, opened 2024-06-21
feat(components): scalar tooltip

antlio:
this pr adds ScalarTooltip component, in order to be used in the client app package. n.b this could replace the usage of `useTooltip` package which can be remove after merge.


marclave (approved):
🤾‍♂️

PR #2206, opened 2024-06-21
feat(client-app): hide header in read only mode

antlio:
this pr is a proposal to hide the client modal header in read only mode as it could be misleading:

image


marclave (approved):
fantastic idea ✨

PR #2207, opened 2024-06-21
fix(client-app): http methods editable style

antlio:
this pr style http methods component when not editable in order to avoid clickable expectation.


marclave (approved):
ah much better @antlio !!

PR #2208, opened 2024-06-21
feat(void): XML and HTML response

hanspagel:
This PR adds HTML and XML responses to the `@scalar/void-server`, they come up when you request an URL that ends with `.html` or `.xml` or send an `Accept: application/xml` or `Accept: text/html` header.

Screenshot 2024-06-21 at 16 02 40


marclave (approved):
tested and works, nice : )

PR #2209, opened 2024-06-21
feat(client-app): add scalar-client class to teleported dropdowns

hwkr:
ScalarDropdown [merges the class onto the dropdown](https://github.com/scalar/scalar/blob/main/packages/components/src/components/ScalarDropdown/ScalarDropdown.vue#L24) so we can just add the class directly to the dropdown component.

## Before:

Google Chrome-2024-06-21-09-25-08@2x

## After:

Google Chrome-2024-06-21-09-24-15@2x

### DIFF
--- a/packages/client-app/index.html
+++ b/packages/client-app/index.html
@@ -25,7 +25,7 @@
 
   <body>
     <div
-      id="app"
+      id="scalar-client"
       class="scalar-app scalar-client"></div>
     <script
       type="module"
--- a/.changeset/six-planets-buy.md
+++ b/.changeset/six-planets-buy.md
@@ -1,0 +1,5 @@
+---
+'@scalar/client-app': patch
+---
+
+feat(client-app): add scalar-client class to teleported dropdowns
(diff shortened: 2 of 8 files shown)


amritk (approved):
Perfect

marclave (approved):

antlio (approved):
👍 lgtm we have a couple others, don't know if it's worth adding every time we teleport through client app

amritk:
@antlio is right actually, can't we teleport it into a div that already has the class?

hwkr:
> @antlio is right actually, can't we teleport it into a div that already has the class?

Yeah we can, the only thing is floating doesn't know anything about the app setup so the only safe place for it to teleport to is `body` but we could make that configurable (and teleport to say, `#scalar-client` or something)

packages/client-app/index.html
```diff
@@ -25,7 +25,7 @@
 
   <body>
     <div
-      id="app"
+      id="scalar-client"
```
hwkr, line 28: This probably shouldn't be `app` anyway to avoid id collisions.

PR #2214, opened 2024-06-22
fix: query params not added to request url

marclave:
**Problem**
No query params are sent to the proxy

**Explanation**
This happens because we dont add them to the url

**Solution**
With this PR we add them to the URL

### DIFF
--- a/.changeset/shaggy-llamas-unite.md
+++ b/.changeset/shaggy-llamas-unite.md
@@ -1,0 +1,5 @@
+---
+'@scalar/client-app': patch
+---
+
+fix: query params not added to request url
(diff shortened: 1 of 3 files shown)


amritk (approved):
we could use a regex as well, but if it works, it works!

hanspagel (changes requested):
This is great !Test is failing, though. Didn’t look into it yet.

hanspagel:
I think I fixed and improved it, but needs a re-review! @amritk @marclave

amritk (approved):
That is much better, nice work @hanspagel

PR #2222, opened 2024-06-24
chore: warn about missing comments for exported functions

hanspagel:
This PR adds an ESlint plugin to warn about out exported functions without JSDoc comments. The lint script isn’t executed in CI, eventually we should add it. But for now it’s enough to help us (or at least me) to add comments where they are missing.

Once we’re done with adding a comment for all existing functions, we can add it to CI or even error on missing  JSDoc comments or apply the rule to all functions … Step by step. :)

Screenshot 2024-06-24 at 12 27 52

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -124,7 +124,10 @@ module.exports = {
             format: ['PascalCase'],
           },
         ],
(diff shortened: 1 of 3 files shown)


marclave (approved):
i really dig this 🔥

PR #2224, opened 2024-06-24
ci: lint all Markdown files

hanspagel:
This PR adds a new GitHub Action Workflow to check all Markdown files for broken links:

```md
(headings)[#does-not-exist]
(files)[does-not-exist.md]
(urls)[https://github.com/DOES/NOT/EXIST]
```

The workflow is failing because there’s a `LICENSE` file missing in the repository, which will be added with that PR - so we can just ignore it here.


marclave (commented):
@hanspagel seems like the lint is failing 👀

marclave (approved):
expected fail , my bad :P

PR #2225, opened 2024-06-24
feat(cli): add more attributes to `scalar init`

hanspagel:
The `scalar init` command generates a `scalar.config.json`.  This PR extends the `scalar init` command to add the required attributes `subdomain`, an empty `guides` array and adds a `--force` flag to overwrite existing configurations without further interaction.

It’s not perfect, but one step closer to connect the CLI to docs.scalar.com.

Screenshot 2024-06-24 at 15 21 38

### DIFF
--- a/.changeset/pretty-rats-melt.md
+++ b/.changeset/pretty-rats-melt.md
@@ -1,0 +1,5 @@
+---
+'@scalar/cli': patch
+---
+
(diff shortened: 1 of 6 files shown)


marclave (approved):
🕺 ah this is so cooool! we are getting closer !!

PR #2229, opened 2024-06-24
feat: fix spotify spec and change back to galaxy

amritk:
- fixed the spec!
- rolled back to galaxy!

will need to come up with a better default function as this won't work


antlio (approved):
lgtm! 👾

marclave (approved):

PR #2232, opened 2024-06-24
feat: client credentials flow

amritk:
This PR adds the client credentials oauth 2 flow. 

Test data IS enabled so you can test both authorization code flow as well as client credentials (they use the same server method now) to make sure there's no regressions

sorry @cameronrohani I ended up doing some front end... I tried to resist. We can work out any conflicts

### DIFF
--- a/packages/oas-utils/src/entities/workspace/security/security-schemes.ts
+++ b/packages/oas-utils/src/entities/workspace/security/security-schemes.ts
@@ -176,5 +176,5 @@ export type SecuritySchemePayload = z.input<typeof securityScheme>
 
 /** Create Security Scheme with defaults */
 export const createSecurityScheme = (payload: SecuritySchemePayload) =>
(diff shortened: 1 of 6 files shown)


packages/oas-utils/src/entities/workspace/security/security-schemes.ts
```diff
@@ -176,4 +176,5 @@ export type SecuritySchemePayload = z.input<typeof securityScheme>
 
 /** Create Security Scheme with defaults */
 export const createSecurityScheme = (payload: SecuritySchemePayload) =>
-  deepMerge(securityScheme.parse({ type: payload.type }), payload)
+  // deepMerge(securityScheme.parse({ type: payload.type }), payload)
```
amritk, line 179: need to change this back before merging so we don't break on schema errors

marclave (approved):
🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥

PR #2233, opened 2024-06-25
fix: add null check for schema type on request

marclave:
the modal cant close cause we dont check the schema before the type! this fixes that : )


amritk (approved):
Safe

PR #2234, opened 2024-06-25
fix: ensure trailing empty parameter item

marclave:
we would often end up with no ability to add an item! now if theres an item we check to see if we can add an empty row on mount :)


amritk (approved):
Seems reasonable

PR #2236, opened 2024-06-25
chore: add more ESLint JSDoc rules

hanspagel:
This PR enables a few more recommended ESLint JSDoc rules, mostly focused on formatting and checking validity (for params and stuff). The formatting related rules can be automatically fixed, so nothing we need to manually correct.

Helps us to keep the comments in a consistent style. :)


marclave (approved):
i love these rules

PR #2239, opened 2024-06-25
fix(components): scalar button loading state

antlio:
this pr updates the scalar button component in order to fix the loading state:

**before** (loading markup always shown)
image

**after**
(is chilling)
image

(is loading)
image

### DIFF
--- a/.changeset/modern-frogs-reply.md
+++ b/.changeset/modern-frogs-reply.md
@@ -1,0 +1,5 @@
+---
+'@scalar/components': patch
+---
+
+fix: scalar button loading state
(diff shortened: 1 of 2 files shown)


marclave (approved):
nice glad we got something now : )

hwkr:
Looks good! I think we used to have the loading on the side and always show the label (rather than hide the label when we show the loading spinner). But personally I do like it this way, @cameronrohani thoughts?

antlio:
> Looks good! I think we used to have the loading on the side and always show the label (rather than hide the label when we show the loading spinner). But personally I do like it this way, @cameronrohani thoughts?

yes! favored keeping it the same size while making it change state vs being it stretched out but up to discuss if need :)

PR #2240, opened 2024-06-25
feat(galaxy): add void.scalar.com as an alternative server

hanspagel:
This PR adds <https://void.scalar.com> as an alternative server URL to our OpenAPI Example.

This URL just responds with the request data and is great to play around, could be a great addition to the OpenAPI example.


marclave (approved):
✨

PR #2242, opened 2024-06-25
feat(client-app): server variables

hanspagel:
This PR adds support for server variables to `@scalar/client-app`:

Screenshot 2024-06-25 at 13 16 50

```js
{
  "openapi": "3.1.0",
  "info": {
    "title": "Servers with Variables",
    "version": "1.0.0"
  },
  "servers": [
    {
      "url": "https://{username}.gigantic-server.com:{port}/{basePath}",
      "description": "The production API server",
      "variables": {
        "username": {
          "default": "demo",
          "description": "this value is assigned by the service provider, in this example `gigantic-server.com`"
        },
        "port": {
          "enum": [
            "8443",
            "443"
          ],
          "default": "8443"
        },
        "basePath": {
          "default": "v2"
        }
      }
    }
  ],
  "paths": {}
}
```

### DIFF
--- a/packages/client-app/src/components/AddressBar/AddressBarServer.vue
+++ b/packages/client-app/src/components/AddressBar/AddressBarServer.vue
@@ -36,6 +36,20 @@ const updateSelectedServer = (serverUid: string) => {
 const isSelectedServer = (serverId: string) => {
   return activeCollection.value?.selectedServerUid === serverId
 }
+
+// Replace variables in URL
+const serverUrl = computed(() => {
+  const server = servers[activeCollection.value?.selectedServerUid ?? '']
+  const url = server?.url
+
+  // TODO: use `replaceVariables` from `@scalar/oas-utils/helpers`
+  // Note: Currently, it’s in @scalar/api-client, but that’s about to change.
+  const singleCurlyBrackets = /{\s*([\w.-]+)\s*}/g
+  return url.replace(singleCurlyBrackets, (match, key) => {
+    const variable = server?.variables?.[key]
+    return variable?.value || variable?.default || variable?.enum?.[0] || match
+  })
+})
 </script>
 <template>
   <template v-if="serverOptions && !workspace.isReadOnly">
@@ -48,7 +62,7 @@ const isSelectedServer = (serverId: string) => {
         class="font-code lg:text-sm text-xs whitespace-nowrap border border-b-3 border-solid rounded px-1.5 text-c-2"
         type="button"
         @click.stop>
-        {{ servers[activeCollection?.selectedServerUid ?? '']?.url }}
+        {{ serverUrl }}
       </button>
       <template #items>
         <ScalarDropdownItem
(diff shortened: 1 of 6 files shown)


packages/client-app/src/components/AddressBar/AddressBarServer.vue
```diff
@@ -36,6 +36,20 @@ const updateSelectedServer = (serverUid: string) => {
...
+  const url = server?.url
+
+  // TODO: use `replaceVariables` from `@scalar/oas-utils/helpers`
+  // Note: Currently, it’s in @scalar/api-client, but that’s about to change.
+  const singleCurlyBrackets = /{\s*([\w.-]+)\s*}/g
+  return url.replace(singleCurlyBrackets, (match, key) => {
+    const variable = server?.variables?.[key]
+    return variable?.value || variable?.default || variable?.enum?.[0] || match
+  })
+})
```
geoffgscott, line 52: This will error at runtime when the selectedServerUid does not exist in the array.

I would suggest:
```ts
if(!servers.length) return ''
const server = servers[activeCollection.value?.selectedServerUid ?? ''] ?? servers[0]
const url = server.url
```

or  the following which will correctly assign the type to URL

```ts
const url = server?.url as string | undefined
```

packages/client-app/src/components/AddressBar/AddressBarServer.vue
```diff
@@ -36,6 +36,20 @@ const updateSelectedServer = (serverUid: string) => {
 const isSelectedServer = (serverId: string) => {
   return activeCollection.value?.selectedServerUid === serverId
 }
+
+// Replace variables in URL
```
geoffgscott, line 40: For variable definitions and functions the `/** Comment */` format is better because it shows up in intellisense. Rather than labelling the computed property with whats going on inside its better to label the variable with what it represents

packages/client-app/src/views/Servers/ServerForm.vue
```diff
@@ -33,12 +33,55 @@ const updateServer = (key: string, value: string) => {
...
+  const variables = activeServer.value?.variables ?? {}
+
+  const keys = Object.keys(variables)
+
+  return keys.map((key: string) => ({
+    key,
+    label: key,
+    placeholder: variables?.[key]?.default ?? '',
+  }))
+})
```
geoffgscott, line 52: Object.entries would tidy this up a bit.

```suggestion
const variableOptions = computed(() =>
  Object.entries(activeServer.value?.variables ?? {}).map(([key, val]) => ({
    key,
    label: key,
    placeholder: val ?? '',
  })),
)
```

packages/client-app/src/views/Servers/ServerForm.vue
```diff
@@ -33,12 +33,55 @@ const updateServer = (key: string, value: string) => {
...
+    (acc, variable) => ({
+      ...acc,
+      [variable]:
+        activeServer.value.variables?.[variable].value ??
+        activeServer.value.variables?.[variable].default ??
+        '',
+    }),
+    {},
+  )
+})
```
geoffgscott, line 68: Entries cleans it up and reduce doesn't type infer so the type param is a good idea. 
```suggestion
/** Get values from the workspace, use `default` as fallback */
const variablesData = computed(() => {
  return Object.entries(activeServer.value.variables ?? {}).reduce<
    Record<string, string>
  >((acc, [key, entry]) => {
    acc[key] = entry.value ?? entry.default
    return acc
  }, {})
})
```

PR #2245, opened 2024-06-25
refactor(api-reference): servers

hanspagel:
This refactors how we deal with server URLs in `@scalar/api-reference`:

* New `features/BaseUrl` folder
* Keeps all the relevant files close to each other
* Introduces a new root component to do *all the data wrangling* in one spot, instead of tons of different files with watchers and access to the `useServerStore`
* Tests for the normalization of the servers list

Note: This is still not using the `@scalar/client-app`, but once #2242 is done, we can do all the wiring in the new root component.

I see a lot more things to do, but this PR is already way bigger than it should be. 🙈 I’ll create separate PRs to further improve it and add more tests.


packages/api-reference/src/features/BaseUrl/ServerForm.vue
```diff
@@ -0,0 +1,147 @@
...
+              v-if="servers && servers.length > 1"
+              :value="selectedRef"
+              @input="updateSelectedIndex">
+              <option
+                v-for="(serverOption, index) in servers"
+                :key="index"
+                :value="index">
+                {{ serverOption.url }}
+              </option>
+            </select>
```
geoffgscott, line 57: Please no more native selects. We have the component library select now.

marclave (approved):
tested, works great, tests. its an amazing PR <3

PR #2249, opened 2024-06-25
fix: handle no tags spec import

marclave:
**Problem**
Currently, if a spec has no tags we fail

**Explanation**
This happens because although tags arent required, we dont have a createTags helper with a default

**Solution**
With this PR it creates a temporary default tag

image


amritk (approved):
not the prettiest, but it'll work

PR #2250, opened 2024-06-25
fix: type modalstate

amritk:
@antlio using that symbol helper we can type the bus


marclave (approved):
i do not long for the unknown.

nice work @amritk

amritk:
no changeset as its only types, will changeset my next pr

PR #2251, opened 2024-06-25
fix: client app binary file and multipart ui

cameronrohani:
simplify file upload a bit... we should go all out with fixing this up and making a great UX here over the next few weeks but just a start :) 

image

image

image


marclave (approved):
changeset then we are good to go! nice polish : )

PR #2255, opened 2024-06-25
chore: upgrade typescript

amritk:
I was doing some filtering and I realized we could use some [inferred type predicates](https://devblogs.microsoft.com/typescript/announcing-typescript-5-5/#inferred-type-predicates) from typescript 5.5

before (5.4.3):
image

after (5.5.2):
image

Also do we need typescript in every package? Or is having it in the base package.json enough?

### DIFF
--- a/.changeset/pink-badgers-raise.md
+++ b/.changeset/pink-badgers-raise.md
@@ -1,0 +1,35 @@
+---
+'@scalar/fastify-api-reference': patch
+'@scalar/api-client-proxy': patch
+'@scalar/build-tooling': patch
+'@scalar/echo-server': patch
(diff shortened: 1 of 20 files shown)


hanspagel:
> Also do we need typescript in every package? Or is having it in the base package.json enough?

Would be great to remove it from the packages! It would be cool if `types:check` still works when I’m in a single package,  though. Not sure if this requires the typescript dependency in this package then.

marclave (approved):

amritk:
Yep correct, only needed in the root. Removed! I bet it was just auto installed by the app bootstraps

PR #2261, opened 2024-06-26
fix: URL is not encoded when sent to the proxy

hanspagel:
When sending multiple query parameters, only the first one is proxied to the target URL.

This is because we used `encodeURI` instead of `encodeURIComponent` in a few places.

```js
encodeURI('https://example.com/?color=red&size=large')
// 'https://example.com/?color=red&size=large'

encodeURIComponent('https://example.com/?color=red&size=large')
// 'https%3A%2F%2Fexample.com%2F%3Fcolor%3Dred%26size%3Dlarge'
```

We’re not encoding the URL (https://… is perfectly fine for an URL), but a parameter.

**❌ encodeURI**
https://proxy.scalar.com/?scalar_url=https://void.scalar.com?foo=bar&bar=foo

**✅ encodeURIComponent**
https://proxy.scalar.com/?scalar_url=https%3A%2F%2Fvoid.scalar.com%3Ffoo%3Dbar%26bar%3Dfoo%27

### DIFF
--- a/.changeset/lovely-monkeys-deny.md
+++ b/.changeset/lovely-monkeys-deny.md
@@ -1,0 +1,7 @@
+---
+'@scalar/api-client': patch
+'@scalar/client-app': patch
+'@scalar/oas-utils': patch
+---
+
+fix: URL is not encoded when sent to the proxy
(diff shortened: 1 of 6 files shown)


marclave (approved):
ooo nice fix @hanspagel

PR #2266, opened 2024-06-26
feat(client-app): address bar request history

antlio:
this pr introduces a timestamp in request in order to sort request history item most recent first.


marclave (approved):
nice :)

PR #2267, opened 2024-06-26
fix(client-app): cookies are not sent (partial fix)

hanspagel:
Cross-origin cookies are hard, read more in the comment I’ve added in this PR.

With this PR cookies are at least sent from one domain to the other (✅ *.scalar.com -> proxy.scalar.com), it doesn’t work locally or in other setups (e.g. ❌ *.apidocumentation.com -> proxy.scalar.com).

I think we should introduce a custom Cookie header (e.g. `X-Cookie`) that is used when requests are sent to the proxy to circumvent the browser limitations.

Related PR to the proxy: #2265


marclave (approved):
🍪

PR #2269, opened 2024-06-26
feat: update server url in client-app

hanspagel:
WIP

This PR tries to update the server URL in `@scalar/client-app`.

- [ ] Delete all existing servers
- [x] Add a new server (with the string that `getUrlFromServerState` gives us)
- [ ] Switch to the new server

But I struggle with the remaining tasks. Can someone help here?

Note: This is a quick workaround to make the server stuff a non-blocking issue. We’ll fast-follow with a PR to wire it up properly.


marclave (approved):
HUGE lets go :)

PR #2270, opened 2024-06-26
fix(client-app): address bar request history items

antlio:
this pr updates the request history item style and removes italic on pathname:

**before**
image

**after**
image


marclave (approved):
:)

PR #2273, opened 2024-06-26
chore(client-app): merge modal package into client-app

amritk:
- merged the client-modal into the client app
- moved the adding of the scalar container classes to a helper in components 
- and its not best practice, but duplicated some helpers from oas-utils into components to remove components dependency on oas. now it should no longer build any time we change an entitity speeding up build times by a lot!


packages/client-app/src/views/Request/RequestSection/RequestAuth.vue
```diff
@@ -175,7 +175,7 @@ const updateScheme: UpdateScheme = (path, value) =>
               id="http-basic-username"
               class="text-c-2"
               :modelValue="activeSecurityScheme.scheme.value"
-              placeholder="Username"
+              placeholder="ScalarEnjoyer01"
```
antlio, line 178: 😍

packages/client-app/src/views/Request/RequestSection/RequestAuth.vue
```diff
@@ -184,7 +184,7 @@ const updateScheme: UpdateScheme = (path, value) =>
             <RequestAuthDataTableInput
               id="http-basic-password"
               :modelValue="activeSecurityScheme.scheme.secondValue"
-              placeholder="Password"
+              placeholder="xxxxxx"
```
antlio, line 187: maybe?
```suggestion
              placeholder="••••••"
```

amritk: Ya I tried this but is it too close to an actual value?

antlio: yes not wrong, color is dimmed down, maybe we should stick to password?

amritk: My issue with password is it's the same as the label so doesn't really help. The other thing we could do is a fake password like 123ABC

marclave (approved):
alright lets get this in.

we can FF with:
- remove duplicate helpers amrit mentioned
- fix react
- fix docusuarus

PR #2275, opened 2024-06-26
chore: remove random themeclasses object

amritk:
While there is a valid case for an object which holds classes like that, we weren't using it. It was a glorified wrapper around `flex-1`


marclave (approved):
🫡 cleanup!

PR #2276, opened 2024-06-26
style: remove random firefox padding

amritk:
Not sure why this was added, maybe @cameronrohani or @hwkr have some insight
image


marclave (approved):
can merge once the false is removed :)

PR #2280, opened 2024-06-27
feat(client-app): address bar request

antlio:
this pr updates the client app address bar component in order to enable path update in the request.


marclave (approved):
🫡

PR #2281, opened 2024-06-27
fix(client-app):  request sidebar icon

antlio:
this pr updates request sidebar icon accordingly to workspace read only state:

`isReadOnly`
image

`!isReadOnly`
image


marclave:
@antlio thoughts on removing the top level if in read only mode for now? so no Scalar Galaxy? :)

antlio:
> @antlio thoughts on removing the top level if in read only mode for now? so no Scalar Galaxy? :)

yes makes sense!

marclave (approved):

PR #2283, opened 2024-06-27
feat: cache deps and build in CI

tmastrom:
Currently, we are wasting CI time rebuilding the packages for different workflows. This PR aims to create a central build and then cache that and run validation and testing.

### DIFF
--- a/.github/actions/cached-deps.yml
+++ b/.github/actions/cached-deps.yml
@@ -1,0 +1,17 @@
+name: 'Get & Cache Dependencies'
+description: 'Get & Cache Dependencies(via pnpm) for faster builds'
+runs:
+  using: 'composite'
+  steps:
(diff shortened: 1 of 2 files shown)


hanspagel:
As discussed, it sounds great to re-use the build for all workflows, but I don’t think it would speed up CI.

In reality, all workflows are done before ci.yml. Everything that we’d add to ci.yml would make it run longer, or if it would need to wait for something, it would take longer, too.

At least in one workflow we need to - no matter how we do it - have those steps processing after each other:
install, build, test, there’s no shortcut here. And those are the steps that take long.

Also, Turborepo has a great cache which works well locally, but doesn’t work that well in CI. There might be a way to fix this, if we build once and let other workflows wait for it, but we could use the existing Turborepo cache then.

PR #2285, opened 2024-06-27
style: no dropdown for a single server

hanspagel:
We don’t need to have a dropdown for a single server. :)

**Before**

Screenshot 2024-06-27 at 17 55 53

**After**

Screenshot 2024-06-27 at 17 56 38


marclave (approved):
love this <3

antlio (approved):
great catch! we could also set it at a upper level and only have a button markup for multiple servers, but still good for now!

PR #2286, opened 2024-06-27
fix: remove hardcoded proxy url for new client

marclave:
now the dev:modal works :)

image


packages/client-app/src/libs/sendRequest.ts
hanspagel, line 1: `shouldUseProxy` should also check if the proxyUrl is set or not

marclave: local requests wont work with the proxy tho :)

amritk (approved):
LGTM!

PR #2287, opened 2024-06-27
fix: request history routing for client app

marclave:
before i wasnt using the right router, now i am :)


amritk (approved):
vue! lets ball

antlio (approved):
great catch! tested and validated

PR #2289, opened 2024-06-27
feat(oas-utils): move object-utils to rollup

geoffgscott:
Moves object utils to our standard rollup builds


hanspagel:
It’s all coming together! Let’s review and merge this on Tuesday, not before the big release today, I’d say. :v:

hanspagel (changes requested):
There should be a `*.d.ts` file for every source file, right?

Screenshot 2024-06-28 at 16 11 26

packages/object-utils/package.json
```diff
@@ -18,14 +18,14 @@
     "node": ">=18"
   },
   "scripts": {
-    "build": "vite build && pnpm types:build && tsc-alias -p tsconfig.build.json",
+    "build": "scalar-build-rollup",
```
hanspagel, line 21: We should consider to empty the dist folder in `scalar-build-rollup`. Keeping old files can be confusing when  testing locally.

geoffgscott:
> There should be a `*.d.ts` file for every source file, right?
> 
> Screenshot 2024-06-28 at 16 11 26

Yes and it is:
image


These commands are not new they have been in use for several weeks now so nothing should change between packages (unless the original tsconfig file had issues)

hanspagel:
Ah, I didn’t execute `types:build`, I thought that happens with `build`, too. :v:

hanspagel (approved):

geoffgscott:
> Ah, I didn’t execute `types:build`, I thought that happens with `build`, too. ✌️

It does execute as part of `scalar-build-rollup`. It's not auto-deleting the dist right now though so maybe you just had extra files in there.

marclave (approved):

PR #2292, opened 2024-06-27
fix: move css imports to style tags

marclave:
this would break org, so now it is good!


amritk (approved):
LGTM! Testing now but should be fine. Also fixes nuxt among other things!

PR #2296, opened 2024-06-27
refactor: move `<HttpMethod />` component to `@scalar/api-reference`

hanspagel:
As we’re about to say good bye to `@scalar/api-client` v1 it’s time to remove imports to it.

This PR moves the `<HttpMethod />`, which we use in a lot of places, to `@scalar/api-reference`.

⚠️ The `TestRequestButton` had a `HttpMethod` as a wrapper. I don't think we need this anymore, or do we have different colors for the button in other/older themes? @cameronrohani

### DIFF
--- a/.changeset/warm-lamps-wash.md
+++ b/.changeset/warm-lamps-wash.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
+refactor: add HttpMethod component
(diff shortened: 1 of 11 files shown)


amritk (approved):
Nice, later we can even move to the new HTTPMethod component, or move that one into components

marclave (approved):

PR #2301, opened 2024-06-28
feat: remove reliance on api-client helpers

amritk:
Api client helpers, fixtures, components, hooks, stores have been copied into api-reference/src/legacy.

This is the beginning of the end (for the old client)


packages/api-reference/src/legacy/components/CardForm.vue
```diff
@@ -0,0 +1,13 @@
+<template>
```
amritk, line 1: No need to review below this

amritk: Just copy and pasted. BUT we are missing hans latest changes I think

marclave (approved):

PR #2302, opened 2024-06-28
feat: hello `@scalar/api-client` v2 👋 

marclave:
This PR makes our new API Client the default for everyone. It’s a complete rewrite and it’s amazing:

* Full rewrite
* Tons of bug fixes
* Framework-agnostic 👀 
* Fully based on the OpenAPI specification
* Improved UI (had some learnings from v1)

Amazing work by @amritk, @geoffgscott, @antlio, @hwkr and the whole team! 👏 

Join our Discord to celebrate the release! https://discord.gg/scalar

**Preview**

Screenshot 2024-06-28 at 11 02 47

@hanspagel hijacked this comment


packages/api-reference-editor/tsconfig.json
```diff
@@ -1,9 +1,13 @@
 {
   "extends": "../../tsconfig.json",
+  "include": ["src"],
+  "exclude": ["dist", "vite.config.ts", "node_modules"],
```
antlio, line 4: nice 😮‍💨

hanspagel: actually, we should include the vite.config.ts, excluding it happened just by accident (copied from api-reference) :D

geoffgscott: We don't need to exclude anything that is not included.

marclave: what things are not included? also this is outdated its 

```
"exclude": ["dist", "node_modules"],
```

hanspagel: Yeah, but then it’s already there when we remove rootDir. :)

hanspagel: @marclave include say "src", so vite…, node_modules and dist isn’t included (yet).

marclave: @hanspagel ah no i meant want things do not need to add to the `exclude` array

hanspagel: include only includes src/ and src/ doesn't have vite.config.ts, node_modules and dist, so there’s no point in excluding the files that are one level higher than src.

PR #2303, opened 2024-06-28
fix(api-client): set param example value

antlio:
this pr sets back param example values, which displays the first occurence:

image

### DIFF
--- a/packages/api-client/src/store/workspace.ts
+++ b/packages/api-client/src/store/workspace.ts
@@ -178,7 +178,13 @@ const createParamInstance = (param: OpenAPIV3_1.ParameterObject) =>
   createRequestExampleParameter({
     key: param.name,
     value:
-      param.schema && 'default' in param.schema ? param.schema.default : '',
+      param.schema && 'default' in param.schema
+        ? param.schema.default
+        : param.schema &&
+            'examples' in param.schema &&
+            param.schema.examples.length > 0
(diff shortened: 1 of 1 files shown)


packages/api-client/src/store/workspace.ts
```diff
@@ -178,7 +178,11 @@ const createParamInstance = (param: OpenAPIV3_1.ParameterObject) =>
   createRequestExampleParameter({
     key: param.name,
     value:
-      param.schema && 'default' in param.schema ? param.schema.default : '',
+      param.schema && 'default' in param.schema
+        ? param.schema.default
+        : param.schema && 'examples' in param.schema
+          ? param.schema.examples[0]
```
hanspagel, line 184: In theory, examples could be an empty array. 🤡

hanspagel (changes requested):

packages/api-client/src/store/workspace.ts
antlio: it'll just display no value right now, and `__PLANETID__` in the reference example. wdyt?

marclave: we can probably down the road have a helper function to get default values from openapi spec, i feel like theres a ton of theoretical limits we hit ;)

marclave (approved):
i dreamed of this last night and woke up and its here!!

PR #2304, opened 2024-06-28
feat(api-client): required param hint

antlio:
this pr adds required hint for input according to param required state which gets hidden on input focus:

image


hanspagel (approved):
This change was required

marclave (approved):
nice work @antlio ! I think @cameronrohani and I had mocked up some required kind of pill like things for these inputs but lets get this in :)

PR #2305, opened 2024-06-28
fix(api-reference): sidebar font weight

antlio:
this pr sets the right font weight for the method in the api reference sidebar (700 -> 600):

image


hanspagel (approved):
YESS

marclave (approved):
nice :)

PR #2272, opened 2024-06-26
fix(api-reference): missing servers

zer0npc:
**Problem**
The configuration of servers has no effect

PR #2306, opened 2024-06-28
fix(api-reference): servers are ignored

hanspagel:
We somehow didn’t pass the list of servers (anymore). This passes the servers list down to the `<BaseUrl />` component as documented. :)

Based on #2272.


marclave (approved):
👯

PR #2307, opened 2024-06-28
chore: use new `redirectToProxy` helper

hanspagel:
The old `redirectToProxy` helper, that was removed in #2297, somehow reappeared.

This PR removes it and switches everything to the new one. :)


marclave (approved):
out with the old! in with the new, nice @hanspagel

PR #2309, opened 2024-06-28
ci: use Node 18 only for release PRs

hanspagel:
With this PR we’re going to use Node 20 by default, and release PRs will also be tested in Node 18.

This should speed up CI for all PRs, but the release PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,11 +6,30 @@ on:
       - main
 
 jobs:
+  # Always use Node 20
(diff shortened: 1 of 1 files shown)


marclave (approved):
im happy with this ✨

hanspagel:
Waiting for @geoffgscott’s review :v:

geoffgscott (approved):
Looks good. Probably won't reduce CI time much because they are run in parallel but will reduce CI run minutes which is always a win.

PR #2310, opened 2024-06-28
ci: publish packages on Stackblitz

hanspagel:
With this PR our packages are published on Stackblitz for all release PRs. 

This is great, because we can test the published version without publishing to npm (and all our users).

The publishing takes ~4 minutes, so I made it a separate workflow which can run in parallel to the `ci.yml`.


marclave (approved):
@hanspagel this is 🤯 :)

PR #2311, opened 2024-06-28
feat(api-client): search modal

antlio:
this pr moves the search modal component to the request vue in api client package, in order to make it displayable on hotkey even if the sidebar if hidden.


marclave (approved):
✨

PR #2312, opened 2024-06-28
chore(components): clean up env defintion file

hanspagel:
~~Hey, I’ve found some prism types here. Let’s clean this up a bit, before the weekend comes.~~

Hey, this became a collaboration of @geoffgscott and me to make the `@scalar/components` TypeScript configuration and Vite configuration more like what we use for the other packages. ✨ 

One step closer to use the same configuration for all packages.


marclave (approved):
sounds good to me : )

PR #2317, opened 2024-06-28
fix: api client top position mobile

cameronrohani:
new api client top position on mobile is too low, this pr fixes that!

before:
image

after:
image


marclave (approved):
🫡

PR #2323, opened 2024-06-28
fix: default proxy for client app and revert to scalar galaxy

marclave:


amritk (approved):
Just leave spotify comment OR I can bring it back for testing

PR #2329, opened 2024-06-29
fix: remove some toRaw

amritk:
it was bugging me, this makes more sense

### DIFF
--- a/packages/api-client/src/store/workspace.ts
+++ b/packages/api-client/src/store/workspace.ts
@@ -48,7 +48,7 @@ import {
   setNestedValue,
 } from '@scalar/object-utils/nested'
 import type { AnyObject, OpenAPIV3_1 } from '@scalar/openapi-parser'
-import { computed, reactive, readonly } from 'vue'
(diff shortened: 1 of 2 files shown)


marclave:
@amritk wait why does this work 🤯 what changed 🤯

amritk:
Removed a couple toRaws, but Hans pointed out we can do an even better job

amritk:
Alright now its much better

packages/api-client/src/store/workspace.ts
```diff
@@ -545,7 +545,8 @@ const deleteServer = (serverUid: string, collectionUid: string) => {
 // ---------------------------------------------------------------------------
 
 /** Helper function to import a OpenAPI spec file into the local workspace */
-async function importSpecFile(spec: string | AnyObject) {
+async function importSpecFile(_spec: string | AnyObject) {
```
marclave, line 548: bit of a nit but doesnt underscore usually mean unused for function params?

marclave (approved):
ah nice now i see why this works :)

PR #2330, opened 2024-06-29
fix: only send query params with a value

marclave:
**Problem**
Currently, we send query param values regardless of if a value is in there or not

**Explanation**
This happens because we send it always

**Solution**
With this PR we only send it if theres a value


MrSharpp:
tested, its working!

Issue #2334, opened 2024-06-30
bug: path-scoped parameters are not shown/included in `Test Request` functionality


lrstanley:
### What happens?

Path-level parameters are not rendered as an available input option when using the `Test Request` functionality, however, they are rendered within the documentation itself.

### What did you expect to happen?

I would expect `PathItem` level parameters to show up in `Test Request`, just like method/operation parameters.

Most of the time, parameters are defined within each individual operation. However, OpenAPI 3.0+ allows for providing "common" parameters that are shared across all methods of the same path to be stored at the path level.

Ref: https://swagger.io/docs/specification/describing-parameters/#common-for-path

### How can we reproduce the issue?

1. Use the provided spec.
2. Go to a operation which requires a path parameter, e.g. `Retrieve a pet`
3. Hit `Test Request`
4. See that there is no option to provide `petID` (despite it showing in the documentation correctly).

Shows in documentation correctly:
![](https://cdn.liam.sh/share/2024/06/firefox_vsBblgIN5X.png)

However, in the `Test Request` window, path parameters aren't an option:
![](https://cdn.liam.sh/share/2024/06/firefox_Gj9zha0LQH.png)

When defined in the operation/method, they do show up, however.

### OpenAPI Document

https://cdn.liam.sh/share/2024/06/openapi.json

marclave:
@lrstanley thanks for reporting this! I got a fix here :) you can test it in the deploy preview too, but you can see it working with the spec you shared in the screenshot ✨ 

https://github.com/scalar/scalar/pull/2335

PR #2335, opened 2024-06-30
fix: ensure path and operation params are merged on import

marclave:
**Problem**
Currently, we miss path level params on import

**Explanation**
This happens because we dont check them only operation params

**Solution**
With this PR we check both :) 

image

### DIFF
--- a/packages/oas-utils/src/transforms/import-spec.ts
+++ b/packages/oas-utils/src/transforms/import-spec.ts
@@ -72,8 +72,14 @@ export const importSpecToWorkspace = async (spec: string | AnyObject) => {
         cookies: {},
       }
 
+      // An operation can have component level parameters as well :)
+      const pathAndOperationParameters = [
+        ...path.parameters,
+        operation.parameters,
+      ].filter((p) => p)
+
       // Loop over params to set request params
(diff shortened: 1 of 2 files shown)


marclave:
https://github.com/scalar/scalar/issues/2334

packages/oas-utils/src/transforms/import-spec.ts
```diff
@@ -72,8 +72,14 @@ export const importSpecToWorkspace = async (spec: string | AnyObject) => {
         cookies: {},
       }
 
+      // An operation can have component level parameters as well :)
+      const pathAndOperationParameters = [
+        ...path.parameters,
+        operation.parameters,
```
amritk, line 78: U want the ... Here?

amritk (changes requested):

packages/oas-utils/src/transforms/import-spec.ts
marclave: @amritk yeah i actually wanted it in both! but i forgot to commit it up :despair:

all fixed now, theyre both arrays of objects. but sometimes both are null

amritk (approved):
Some types missing, but thats not from this PR

PR #2338, opened 2024-07-01
fix(api-client): required path params

antlio:
this pr fixes the display of the required infos on required item + prevent the update of required path param key that can currently be updated and removed.


marclave (approved):
nice 😎

PR #2339, opened 2024-07-01
fix(api-client): add scalar classes to headless ui root

antlio:
this pr adds scalar classes to headless ui root within api client package to maintain style within the workspace usage:

image


marclave (approved):
🫡

PR #2340, opened 2024-07-01
feat(void-server): routes for all HTTP errors

hanspagel:
I missed that feature: This PR adds routes for all HTTP errors to `@scalar/void-server`.

We’ve already head /404, but now there’s also /503 and others.

Great to test error handling.


marclave (approved):
🥵

PR #2341, opened 2024-07-01
feat(void-server): (multipart) form data

hanspagel:
I’ve missed that feature: This PR adds support for (multipart) form data to `@scalar/void-server`, including support for for fields containing a single or multiple fields and blob requests (just a blob, nothing else).

Great help to debug the file upload.


marclave (approved):
niiiiiiice :)

PR #2342, opened 2024-07-01
chore(build-tooling): remove `dist/` folder by default

hanspagel:
With this PR our Rollup script removes the `dist/` folder by default.

Keeping the `dist/` folder can be confusing, or make you think the build works, but it maybe just kept the files from before your changes.

If - for some reason - that’s not the expected behaviour, you can disable the plugin like this:

```ts
createRollupConfig({
  emptyOutDir: false,
})
```

The Rollup plugin is taken from the official Rollup repository.

### DIFF
--- a/.changeset/gold-horses-travel.md
+++ b/.changeset/gold-horses-travel.md
@@ -1,0 +1,5 @@
+---
+'@scalar/build-tooling': patch
+---
+
+feat: rollup removes dist folder by default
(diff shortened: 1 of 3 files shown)


marclave (commented):
works for me, once @geoffgscott gives review should be good to go :)

geoffgscott (approved):
Looks good but untested. Nice work Hans

PR #2343, opened 2024-07-01
fix(play-button): does not work with yaml

hanspagel:
The `@scalar/play-button` doesn’t support YAML (as shown in the example in the README). This PR fixes it and some more issues:

* We’ve had a `await resp.json()` in there, so it wasn’t working with anything else than JSON. We’re now just passing whatever content we receive to the `parse` function.
* To open a specified operation, it just looked in the first tag (`parsedSpec.tags?.[0]?.operations?.find(…)`), so it didn’t work for operations in other tags.
* We were looking for the specified operation on load, not on click. This broke for URLs that first need to be fetched.
* The example mentioned operationIds from the petstore example, but used the `@scalar/galaxy` example.
* The package ignored `script` tag content, this PR adds support for it.

This PR also adds basic tests for the local `@scalar/play-button` script.

### DIFF
--- a/.changeset/rare-kangaroos-sit.md
+++ b/.changeset/rare-kangaroos-sit.md
@@ -1,0 +1,5 @@
+---
+'@scalar/play-button': patch
+---
+
+fix: does not work with YAML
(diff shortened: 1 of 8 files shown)


marclave (approved):
so rad nice work @hanspagel !!!

you are the king of tests

PR #2345, opened 2024-07-01
feat(api-client): data table input enum select

antlio:
this pr adds an input enum select component to the api client package, in order to be able to select an enum value when available, along maintaining the capacity of adding our own value:

image

### DIFF
--- a/.changeset/dirty-parrots-sin.md
+++ b/.changeset/dirty-parrots-sin.md
@@ -1,0 +1,5 @@
+---
+'@scalar/oas-utils': patch
+---
+
+feat: oas-utils request example parameters enum
--- a/.changeset/serious-clouds-pretend.md
+++ b/.changeset/serious-clouds-pretend.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
+
+feat: api client data table input enum select component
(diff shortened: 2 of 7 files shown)


marclave (commented):
@antlio this looks great to me! I think some padding is off but adding @cameronrohani for you two to talk about it.


Something we also need to think about is this being kind of "strict" mode, where we should be able to switch from an enum to a normal input.

this is important for two things:

1) in "strict" mode a la popup, people might want to go outside the spec for _some reasons_ (maybe out of date docs)
2) when in the dedicated API client app, people might be working with an enum but also want to switch on the fly! so we should let ppl change :)

wdyt @antlio @hanspagel @amritk

antlio:
> @antlio this looks great to me! I think some padding is off but adding @cameronrohani for you two to talk about it.
> 

if about the dropdown, this comes from the components (dropdown, listbox). will fix if we decide too


> Something we also need to think about is this being kind of "strict" mode, where we should be able to switch from an enum to a normal input.
> 

this is what add value allows to, anything else in mind, or approach?

hanspagel:
I think it should be possible to pass any value in any mode. :) When playing around with an API it should always be possible to pass whatever you like, even if it’s just to test the error handling.

This PR still guides users with the information we get from the OpenAPI document, so I really dig the approach of this PR!

I think the UX is great, too.

hanspagel (approved):
Great UX! I tested the PR, but didn’t look at the code, though.

marclave (approved):

PR #2346, opened 2024-07-01
feat(api-client): tab focus style

antlio:
this pr adds tab index to sidebar child item in order to make them focusable + custom style focus visible:

image


marclave (approved):
tabindex! lets goo :)

PR #2349, opened 2024-07-02
fix(api client): add cookies

antlio:
this pr fixes the zod error when adding a cookie by adding missing uid + default value to cookie schema. another take could be to use `cookieSchema.parse(deepMerge({}, payload))` vs `deepMerge(cookieSchema.parse({}), payload)` to only validate merged result and allow empty object, but i favored here adding default value to maintain code consistency.

image

### DIFF
--- a/.changeset/mean-falcons-thank.md
+++ b/.changeset/mean-falcons-thank.md
@@ -1,0 +1,6 @@
+---
+'@scalar/api-client': patch
+'@scalar/oas-utils': patch
+---
+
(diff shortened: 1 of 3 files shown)


marclave (approved):
safe 🍪's!

amritk:
Ya so currently we actually don't use the Zod validation! I been using it for development, and for the types. But we usually just parse an empty object and merge with defaults since we don't want this to break for users in prod.

PR #2354, opened 2024-07-02
chore: update workspace dependencies

hanspagel:
This PR updates some of the workspace dependencies. So fresh, so clean, so up to date. :)

Note: The GitHub Action workflows that use Playwright restore the executables from any Playwright version, that’s failing. I removed the `restore-keys`, so it only restores executable for the exact version.


hanspagel:
Closing in favor if #2385 and #2399.

PR #2355, opened 2024-07-02
chore(echo-server): remove example

hanspagel:
This PR removes the `@scalar/echo-server` example and switches to a `playground/` example in the new `@scalar/void-server` package. Cleans it up, moves us closer to the structure we’d like to have (no dedicated `examples/` folder).


marclave (approved):
awesome clean up @hanspagel

PR #2361, opened 2024-07-02
fix: @scalar/use-codemirror missing, fix #2358

### LINKED ISSUE #2358: Missing dependencies in 1.24.28
### What happens?

```
vite: X [ERROR] Could not resolve "@scalar/use-codemirror"
vite:
vite:     node_modules/@scalar/api-client/dist/components/CodeInput/CodeInput.vue2.js:2:53:
vite:       2 │ ...eCodeMirror as w, colorPicker as P } from "@scalar/use-codemirror";
vite:         ╵                                              ~~~~~~~~~~~~~~~~~~~~~~~~
vite:
vite:   You can mark the path "@scalar/use-codemirror" as external to exclude it from the bundle, which will remove this error and leave the unresolved path in the bundle.
```

Expect that the following code in Vue works:

``` vue
<script setup lang="ts">
import { ApiReference } from "@scalar/api-reference";
import "@scalar/api-reference/style.css";
</script>
```

- `npm i @scalar/api-reference`
- `npm ls @scalar/use-codemirror`
- Observe:
```
MyProject@0.1.0
└── (empty)
```

### OpenAPI Document

_No response_

hanspagel:
This PR moves three of the devDependencies of `@scalar/api-client` to the dependencies list.

Reproduced the issue with `pnpm add https://pkg.pr.new/scalar/scalar/%40scalar%2Fapi-client@49da03d` in a fresh project, had to install those three dependencies to make the build pass.

See #2358


amritk (approved):
LGTM

marclave (approved):

PR #2363, opened 2024-07-02
fix: remove tailwind preflight from api client app

hwkr:
This should fix the leaky reset styles that were coming out of `@scalar/api-client`. It's worth doing some testing to make sure removing the Tailwind preflight didn't break anything.

### DIFF
--- a/packages/themes/src/reset.css
+++ b/packages/themes/src/reset.css
@@ -1,7 +1,7 @@
 /*
   Custom reset (loosely) based on Tailwind Preflight
 
-  @see https://github.com/tailwindlabs/tailwindcss/blob/next/packages/tailwindcss/preflight.css
+  @see https://github.com/tailwindlabs/tailwindcss/blob/3.3/src/css/preflight.css
 */
 
 /* Use :where to lower specificity */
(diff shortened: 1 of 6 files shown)


amritk (approved):
There are slight differences but IMO this one looks better. Will tag in @cameronrohani to check it out

packages/themes/src/reset.css
```diff
@@ -75,4 +80,15 @@
     outline: 1px solid var(--scalar-color-accent);
     outline-offset: 1px;
   }
+
+  /** Set the default cursor for buttons. */
+  button,
```
antlio, line 85: seems like the button in api client are tight off, should we inherit their display?

antlio:
couple of things in the api client in workspace mode, but that can be tackled in another pr! other than button, seems good to go!

packages/themes/src/reset.css
hwkr: I'm not sure what you mean by tight off?

antlio: sorry typo 😂 i meant within api client the sidebar icon button + history button are bigger than usual

hwkr: Ahhh okay, I'll have a look

hwkr: Fixed!

marclave (approved):
lets get @cameronrohani to give this a test and we can get it merged!

cameronrohani:
Sidebar items shouldn't be hitting the end - not sure if related to this PR 
image

hwkr:
> Sidebar items shouldn't be hitting the end - not sure if related to this PR image

Looks like it's the same on main, I'll do a fix though

PR #2364, opened 2024-07-03
fix(api-client): address bar history infos

antlio:
this pr fixes the address bar history item informations being changed when navigating within items in the api client package, by removing the activeRequest response value being stored on click.

to date when navigating within the history, selecting an item is updating the activeRequest value and hence setting the infos towards all the items as below:

https://github.com/scalar/scalar/assets/14966155/b0344bf3-a12e-4582-9364-f698af83cc56


marclave (approved):
ah nice! i did this and idk why i did haha ✨

PR #2366, opened 2024-07-03
ci: cache build, run jobs in parallel ⚡

hanspagel:
With this PR the one job in our main GitHub Action workflow is split up into multiple jobs that all wait for a single build. That way, jobs can be run in parallel and speed up CI.

Note: This makes the ci.yml way longer (+94 LOC) and one could argue also harder to understand, but personally, I think it’s worth it. [Reusable workflows](https://docs.github.com/en/actions/using-workflows/reusing-workflows#using-inputs-and-secrets-in-a-reusable-workflow) could help here. I’ll look into that soon.

**Additional changes**
* Every PR starts from a global Turborepo cache, but keeps its own separate cache for every PR. This should decrease the cache miss rate.

## Before: 6m 24s

The build cache never really worked for all packages, the cache miss rate was pretty high, so this was the fastest possible run before:

Screenshot 2024-07-03 at 13 13 09

## After: 2m 33s

Restoring the pnpm cache and installing cached dependencies takes 20-30s already, but build time is down to 1s (no cache miss for the build), so this is the fastest possible run with this PR:

Screenshot 2024-07-03 at 12 58 56

### DIFF
--- a/.github/workflows/deploy-preview-examples.yml
+++ b/.github/workflows/deploy-preview-examples.yml
@@ -50,9 +50,10 @@ jobs:
         uses: actions/cache@v4
         with:
           path: .turbo
-          key: ${{ runner.os }}-turbo-node-${{ matrix.node-version }}
+          key: turbo-${{ runner.os }}-node-${{ matrix.node-version }}-branch-${{ github.head_ref }}
           restore-keys: |
-            ${{ runner.os }}-turbo-node-
+            turbo-${{ runner.os }}-node-${{ matrix.node-version }}-branch-
+            turbo-${{ runner.os }}-node-
       - name: Build
         run: cd examples/web && pnpm turbo run build
         env:
--- a/.github/workflows/publish-on-stackblitz.yml
+++ b/.github/workflows/publish-on-stackblitz.yml
@@ -24,9 +24,10 @@ jobs:
         uses: actions/cache@v4
         with:
           path: .turbo
-          key: ${{ runner.os }}-turbo-node-20
+          key: turbo-${{ runner.os }}-node-20-branch-${{ github.head_ref }}
           restore-keys: |
-            ${{ runner.os }}-turbo-node-
+            turbo-${{ runner.os }}-node-20-branch-
+            turbo-${{ runner.os }}-node-
       - name: Build packages
         run: pnpm build:packages
       - name: Publish on Stackblitz
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,4 +1,4 @@
-name: Check build, types, and units tests
+name: Check build, types, and unit tests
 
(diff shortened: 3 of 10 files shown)


.github/workflows/publish-on-stackblitz.yml
```diff
@@ -24,9 +24,10 @@ jobs:
         uses: actions/cache@v4
         with:
           path: .turbo
-          key: ${{ runner.os }}-turbo-node-20
+          key: turbo-${{ runner.os }}-node-20-branch-${{ github.head_ref }}
           restore-keys: |
-            ${{ runner.os }}-turbo-node-
+            turbo-${{ runner.os }}-node-20-branch-
+            turbo-${{ runner.os }}-node-
```
tmastrom, line 30: Nice work! I can see the build cache being hit by all these other test workflows. I wasn't understanding how that part works. I assumed that github ran these workflows in parallel? Is there only redundancy on the first push to the branch?

.github/workflows/deploy-preview-examples.yml
```diff
@@ -50,9 +50,10 @@ jobs:
         uses: actions/cache@v4
         with:
           path: .turbo
-          key: ${{ runner.os }}-turbo-node-${{ matrix.node-version }}
+          key: turbo-${{ runner.os }}-node-${{ matrix.node-version }}-branch-${{ github.head_ref }}
```
tmastrom, line 53: I like the idea of caching per branch instead of per commit 👍

tmastrom (approved):
Amazing reduction in CI time! I agree in the future as the workflow grows it could be cleaner to use reusable workflows but for now I think this is still readable and maintainable with 3 steps.

.github/workflows/publish-on-stackblitz.yml
hanspagel: Yes! It tries to restore the cache from those sources:

1. the exact same branch: turbo-${{ runner.os }}-node-20-branch-${{ github.head_ref }}
2. any other branch: turbo-${{ runner.os }}-node-20-branch-*
3. or just the same Node version: turbo-${{ runner.os }}-node-*

The first commit for a PR should restore the cache from any other branch, for the following commits it’s taking the cache from the exact same branch.

> **restore-keys:** Optional A string containing alternative restore keys, with each restore key placed on a new line. If no cache hit occurs for key, these restore keys are used sequentially in the order provided to find and restore a cache. For example:

Source: https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#using-the-cache-action

tmastrom: Thanks for the clarification. That makes sense now 😄

marclave (approved):
huge gaiiiiins

PR #2369, opened 2024-07-03
fix(scalar.aspnetcore): additional brackets break the configuration, fix #2368

### LINKED ISSUE #2368: Configuration Incorrectly Generated In .Net Package
### What happens?

When using the .Net Package and adding in configuration options an additional { is generated in the outputed HTML which results in an error and the inability to use Scalar Options. 

```csharp
app.MapScalarApiReference(c =>
{
    c.Theme = "purple";
    c.EndpointPathPrefix = "docs";
    c.SearchHotkey = "?";
});
```

This then produces the following HTML 

```html
<!doctype html>
<html>
<head>
    <title>Scalar API Reference -- v1</title>
    <meta charset="utf-8" />
    <meta name="viewport" content="width=device-width, initial-scale=1" />
</head>
<body>
    <script id="api-reference" data-url="/openapi/v1.json"></script>
    <script>
    var configuration = {
        {
  "theme": "purple",
  "searchHotkey": "?"
}
    }

    document.getElementById('api-reference').dataset.configuration =
        JSON.stringify(configuration)
    </script>
</body>
</html>
```

The configuration should only have one bracket. 
This would then allow for the configuration to be loaded correctly.

1. Create a New .Net Project with Scalar Support
2. Add Configuration Options such as Theme 
3. Run Project and Direct over to Document
4. Theme will fail to load

### OpenAPI Document

_No response_

hanspagel:
This PR removes the additional brackets from the configuration output, which otherwise break the configuration.

See #2368

How do we test this integration? @marclave

### DIFF
--- a/packages/scalar.aspnetcore/Scalar.AspNetCore.cs
+++ b/packages/scalar.aspnetcore/Scalar.AspNetCore.cs
@@ -44,9 +44,7 @@ namespace Scalar.AspNetCore
                           <body>
                               <script id="api-reference" data-url="/openapi/{{documentName}}.json"></script>
                               <script>
-                              var configuration = {
-                                  {{configurationJson}}
-                              }
+                              var configuration = {{configurationJson}}
 
                               document.getElementById('api-reference').dataset.configuration =
                                   JSON.stringify(configuration)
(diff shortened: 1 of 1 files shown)


marclave:
@hanspagel I did this https://stackoverflow.com/questions/53841776/install-nupkg-locally-on-macos-using-dotnet-or-nuget :) let me know if you need help tomorrow

marclave (approved):

PR #2370, opened 2024-07-03
feat(api-client): address bar history infos

antlio:
this pr adds response duration infos to the history items + a different ordering proposal as seen below:

**before**
image

**after**
image


marclave (approved):
i like this! i think we might need to align some things, but we can do that in another PR after we play around with it ✨

PR #2373, opened 2024-07-03
fix(void-server): Node 18 does not have File

hanspagel:
We didn’t run the tests in CI (read more in #2367), but now we do. Turns out `createVoidServer` uses `File` in Node, which isn’t available in Node 18.

This PR adds File from the `undici` package (which is what the recent Node versions use under the hood). This should make CI pass for Node 18 again, at least it does locally.


marclave (approved):
:)

PR #2374, opened 2024-07-03
feat(void-server): force JSON responses

hanspagel:
I just wanted to force a JSON response, so I’ve added more file extensions to force a specific content type and refactored the routes to just use one single route for `.json`, `.html`, `.xml` and `.zip`.


marclave (approved):
:)

PR #2380, opened 2024-07-04
fix(api-reference): don't constrain markdown image height

hwkr:
Chatted with @cameronrohani and it sounds like we want to constrain markdown images by width not height 👍 

## Before

Google Chrome-2024-07-03-13-22-34@2x

## After

Google Chrome-2024-07-03-23-59-42@2x

### DIFF
--- a/.changeset/sharp-jeans-prove.md
+++ b/.changeset/sharp-jeans-prove.md
@@ -1,0 +1,6 @@
+---
+'@scalar/code-highlight': patch
+'@scalar/docusaurus': patch
+---
+
+fix: don't constrain markdown image height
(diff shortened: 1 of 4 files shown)


antlio (commented):
should we maintain a constrain or fix full width on classic layout?

image

hwkr:
> should we maintain a constrain or fix full width on classic layout?
> 
> 
> 
> image
> 
> 

We could just add some extra css for the classic specifically, not sure what @cameronrohani wants

Good catch though 🙏

hwkr:
Chatted with @cameronrohani and updated classic layout to contrain width

Google Chrome-2024-07-04-13-18-51@2x

cameronrohani (approved):

marclave (approved):

PR #2384, opened 2024-07-04
fix(api-client): cookies navigation

antlio:
this pr redirects to the page of the cookie on creation.


marclave (commented):
what will happen in the API Client modal @antlio ? : )

antlio:
> what will happen in the API Client modal @antlio ? : )

only in the workspace, but to be updated according to our plan 👀

marclave (approved):

PR #2389, opened 2024-07-04
fix: client search missing a few things!

amritk:
Looks like the search modal on the client was missing a couple things. Also added a fail safe (watch) to avoid the current race condition on getting requests.

Before:
image
image

After:
image

### DIFF
--- a/packages/components/src/components/ScalarSearchResults/ScalarSearchResultItem.vue
+++ b/packages/components/src/components/ScalarSearchResults/ScalarSearchResultItem.vue
@@ -23,7 +23,7 @@ const attrs = computed(() => {
       v-bind="attrs.rest"
       :class="
         cx(
-          'group flex cursor-pointer gap-2 rounded px-3 py-1.5 no-underline hover:bg-b-2',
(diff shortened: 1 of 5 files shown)


marclave (approved):

packages/components/src/components/ScalarSearchResults/ScalarSearchResultItem.vue
```diff
@@ -23,21 +23,23 @@ const attrs = computed(() => {
...
+          'group flex cursor-pointer gap-2.5 rounded px-3 py-1.5 no-underline hover:bg-b-2',
           {
             'bg-b-2': active,
           },
           attrs.className,
         )
       ">
       <!-- Icon -->
-      <div class="text-c-3 group-hover:text-c-1">
+      <div
```
hwkr, line 34: This is just to line up the icon with the first line?

amritk: @cameronrohani

cameronrohani: feel free to do it differently but needs to be aligned with first line of text no matter the font size

PR #2390, opened 2024-07-04
fix(themes): reset list styles and body margin

hwkr:
Turns out there was a little bit more resetting to do for the client app.

The reset on `<body>` is in a layer so it should be easy to override.

### DIFF
--- a/packages/themes/src/reset.css
+++ b/packages/themes/src/reset.css
@@ -4,6 +4,14 @@
   @see https://github.com/tailwindlabs/tailwindcss/blob/3.3/src/css/preflight.css
 */
 
+/* Reset <body> in a layer  */
+@layer scalar-base {
+  body {
+    margin: 0;
+    line-height: inherit;
+  }
+}
+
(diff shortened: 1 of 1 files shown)


cameronrohani (approved):
nice

packages/themes/src/reset.css
```diff
@@ -4,6 +4,14 @@
   @see https://github.com/tailwindlabs/tailwindcss/blob/3.3/src/css/preflight.css
 */
 
+/* Reset <body> in a layer  */
```
amritk, line 7: This will leak out no?

amritk: I jinxed myself, put it in!
image

hwkr: We could put it in the client-app but by nature targeting `<body>` will always leak out of our app.

amritk: send ittt

amritk (approved):
Lets get it in

marclave (approved):

PR #2394, opened 2024-07-05
docs: Add reply.code in fastify example-route

inyourtime:
**Problem**
Currently, in the Fastify documentation, the response schema for the `example-route` is defined with a status of 201. However, that route returns a status of 200. This discrepancy causes tests for that route to receive an empty object instead of the correct response.

**Solution**
With this PR, I aim to add `reply.code(201)` before sending the response object.

### DIFF
--- a/documentation/fastify.md
+++ b/documentation/fastify.md
@@ -155,7 +155,7 @@ fastify.put(
     },
   },
   (req, reply) => {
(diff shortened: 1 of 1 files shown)


marclave (approved):
awesome! thanks for your contribution @inyourtime !!

PR #2396, opened 2024-07-05
fix(api-client): param components style

antlio:
this pr fixes dropdown button component style issue and adds enhancements, in order to hide the `Required` hint only when the peer input is focus vs previously where both input focus (key/value) were hiding it.

### DIFF
--- a/.changeset/famous-ducks-repair.md
+++ b/.changeset/famous-ducks-repair.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
+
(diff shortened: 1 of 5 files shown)


marclave (changes requested):
image

for some reason the icons on the left nav affected

antlio:
> image
> 
> 
> 
> for some reason the icons on the left nav affected

was coming from another pr but is fixed by #2390
i'll rebase accordingly

marclave (approved):
:)

PR #2397, opened 2024-07-05
feat(api-client): show sidebar by default in workspace

antlio:
this pr displays request sidebar by default in workspace, while keeping it hidden in reference.


marclave (approved):
niiiice ✨

PR #2399, opened 2024-07-05
ci: do not restore cache for other playwright versions

hanspagel:

The Cache action has a great feature: It can restore the cache for other keys, in our cases from other Playwright binaries. But that’s not really what we want, because the dependencies are incompatible then. :(

Let’s only restore the cache for the exact same Playwright version.

Oh, and getting the Playwright version didn’t work at all. This PR fixes that, too.

Extracted from #2354.

### DIFF
--- a/.github/workflows/test-cdn-jsdelivr.yml
+++ b/.github/workflows/test-cdn-jsdelivr.yml
@@ -40,7 +40,7 @@ jobs:
         run: pnpm build:packages
       - name: Get Playwright version
         id: playwright-version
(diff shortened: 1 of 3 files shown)


tmastrom (commented):
This looks good. TIL `restore-keys` will overwrite a stale cache. Also noticing the the addition of `actions/cache/restore` and `actions/cache/save` but using the base `cache` looks best for this use case

marclave (approved):

PR #2400, opened 2024-07-05
ci: don’t scope turborepo cache to the branch

hanspagel:

I’ve reread the documentation on the cache action and it says:

> The cache is scoped to the key, [version](https://github.com/actions/cache?tab=readme-ov-file#cache-version), and branch. The default branch cache is available to other branches.

Soo … there’s no need to add the branch to the cache key. And we want a fresh build for different Node versions anyway, so we can remove this restore key, too.

With this PR the Turborepo cache has access to the cache of the main branch for the first run of a PR, but only to the same Node version and the same branch for the next runs.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,10 +53,7 @@ jobs:
         uses: actions/cache@v4
         with:
           path: .turbo
-          key: turbo-${{ runner.os }}-node-${{ matrix.node-version }}-branch-${{ github.head_ref }}
-          restore-keys: |
-            turbo-${{ runner.os }}-node-${{ matrix.node-version }}-branch-
-            turbo-${{ runner.os }}-node-
+          key: turbo-${{ runner.os }}-node-${{ matrix.node-version }}
       - name: Build
         run: pnpm turbo build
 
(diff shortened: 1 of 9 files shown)


tmastrom (commented):
In this case, would we want to utilize the `restore-keys` on the initial CI run for a PR? 

From the cache action docs for [caching strategies](https://github.com/actions/cache/blob/main/caching-strategies.md)
"If cache is not found matching the primary key, restore keys can be used to download the closest matching cache that was recently created. This ensures that the build/install step will need to additionally fetch just a handful of newer dependencies, and hence saving build time."

tmastrom (commented):
In this case, would we want to utilize the `restore-keys` on the initial CI run for a PR? 

From the cache action docs for [caching strategies](https://github.com/actions/cache/blob/main/caching-strategies.md)
"If cache is not found matching the primary key, restore keys can be used to download the closest matching cache that was recently created. This ensures that the build/install step will need to additionally fetch just a handful of newer dependencies, and hence saving build time."

hanspagel:
We do already - but without the additional restore-keys!

This is the same as as in `main`:
`turbo-${{ runner.os }}-node-${{ matrix.node-version }}`

And as quoted in the PR description, the workflow has access to the cache of the default branch. 🤩

marclave (approved):

PR #2401, opened 2024-07-05
ci: only write to the turborepo cache once

hanspagel:

I’ve seen the type:check job often doesn’t have cache for all the examples. My theory is that the workflows that only do `build:packages` overwrite the cache with just the packages and then jobs that start later only have that cache.

But the GitHub Action has a great feature, where you can only restore the cache, but not write to it. So this PR just restores the cache in all cases, and writes to it only in the ci.yml workflow and only in the build job.

Let’s see if we can increase the cache hit rate that way. :)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -97,7 +97,7 @@ jobs:
       - name: Install dependencies
         run: pnpm install
       - name: Turborepo cache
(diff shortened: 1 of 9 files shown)


tmastrom (approved):
Sound logic! I am optimistic about this fix

marclave (approved):

PR #2402, opened 2024-07-05
fix: default array value

antlio:
this pr normalizes the value displayed in reference / client in case we receive an array value as default, [as brought up in x](https://x.com/seancassiere/status/1809011620141748607/photo/1) and as seen below:

**before**
image
image

**after**
image
image


marclave (approved):
🔥

PR #2403, opened 2024-07-05
fix: can’t reach localhost

hanspagel:
For some reason this test is failing in Node 18 (but not Node 20). Whatever, let’s just use 127.0.0.1 instead of localhost, like we do in the other tests.


marclave (approved):
:) nice nice

PR #2404, opened 2024-07-05
fix(api-client): request table tooltip value

antlio:
this pr sets nowrap for request table tooltip component value over pre to better handle potential unsupported usage:

**before**
image

**after**
image


marclave (approved):
🔥

PR #2405, opened 2024-07-05
feat(api-reference): scalar icons

antlio:
this pr is a proposal to update some icons + moves the copy action capability to the scalar codeblock component in order to make it available in both request / response examples:

nb. it does not display copy action when response value is `null`

**before**
image

**after**
image
image

[Uploading scalar-copy.mp4…](https://github.com/scalar/scalar/assets/14966155/d031f9f0-4915-49c3-9021-229297a478f5)


marclave (approved):
much better improvement :) let's ship this 🫡

PR #2409, opened 2024-07-05
feat(doc-1882): save client data to localStorage

amritk:
Saves and loads data from localStorage. It persists now! Requires heavy testing before launch but looks ok

I ended up removing the workspace/routing from this PR, it can be added later. For now it just takes the first workspace but it will support multiple workspaces


marclave (approved):
LETS GOOO

PR #2415, opened 2024-07-06
feat: add empty state + drafts collection

marclave:
Currently we always load the galaxy client, but now we have a ✨ drafts ✨ collection which wont be saved to the cloud, and when you initially go to the app it will have the My First Request as the default

Handled scenarios with:
- no servers 
- keep drafts collection at the bottom

image

some things to improve:
 - make drafts have its own icon
 - ensure we dont have > 1 example for each draft request since its not an openapi compliant request
 - ensure Auth shows up always


marclave:
getting this in for now : )

PR #2416, opened 2024-07-07
feat(api-client): new command palette

marclave:
@cameronrohani and I have been thinking a lot about what this looks like, and we feel good about making it less so of a modal and more so of a command palette!

we will eventually change the second step, but for the sake of keeping the PR small you can get an idea of what it feels like

https://github.com/scalar/scalar/assets/6176314/d3024639-17b6-4668-9b25-21caf74e83fe


hanspagel:
OMG, NICE! I didn't review the code, but just looking at the screen recording: It seems strange that it’s vertically centered. I’d expect to not change the position based on the list of results.

https://github.com/scalar/scalar/assets/1577992/7f9532a3-9c7c-49e6-8480-5a064a76fdc6

cameronrohani (approved):
nice!

PR #2418, opened 2024-07-07
feat(api-client): add actions for requests in the sidebar

marclave:
now we can rename and delete examples + requests!

There was a duplicate item, but thats a bit more involved and we can add it later :) 

also there was a `content before !` in tailwind for the component library maybe @hwkr and @amritk can touch base on that ✨ 
https://github.com/scalar/scalar/pull/2418/files#diff-56f8d484750bbaf10d8d267641bb388b91f388c953715e9a5d2078d97ad1f898L150

also, not sure whats up with our modal but we need to fix that too :) seems to be almost full height of page regardless of variant specified

### DIFF
--- a/packages/api-client/src/views/Request/RequestSidebarItem.vue
+++ b/packages/api-client/src/views/Request/RequestSidebarItem.vue
@@ -150,7 +150,12 @@ const showChildren = computed(
           <div class="relative">
             <RequestSidebarItemMenu
               v-if="!isReadOnly"
-              :item="item" />
+              :item="item"
+              :parentUid="
+                parentUids.length
+                  ? parentUids[parentUids.length - 1]
+                  : undefined
+              " />
             <span class="flex">
               &hairsp;
               <HttpMethod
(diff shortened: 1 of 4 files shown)


packages/api-client/src/views/Request/RequestSidebarItem.vue
```diff
@@ -150,7 +150,12 @@ const showChildren = computed(
           <div class="relative">
             <RequestSidebarItemMenu
               v-if="!isReadOnly"
-              :item="item" />
+              :item="item"
+              :parentUid="
```
hwkr, line 154: Does typescript not like this?

```suggestion
              :parentUid="parentUids[parentUids.length - 1]"
```

hwkr (approved):
Looks like we have a background color issue with the text box but that's not an issue with this PR :)

Google Chrome-2024-07-09-18-31-39@2x

antlio:
> Looks like we have a background color issue with the text box but that's not an issue with this PR :)
> 

good catch! @marclave didn't we say that we wouldn't use `ScalarTextField` though?

marclave:
closing in favor of a new PR post sidebar dragging

PR #2420, opened 2024-07-07
feat(api-client): add pre/post request scripts  🤖

marclave:
🚨 new feature alert 🚨 

Scripting in the Scalar API Client : ) it's pure javascript, with access to a `scalar` object, with some namespaces for functionality depending on if you are in the pre or post script sections

**Present**
Pre Request:
- can access the `request` and add headers : ) (do we give access to the AxiosConfig and let users modify or slight helpers like i did here?)

Post Request:
- tests! you can now run tests and expect values : )
- you also get access to the `response` object to check whats going on : )

image

**Future**
- we should add the scalar object to the Codemirror config so we can add code completion + hints
- we should fix javascript script highlighting
- add a `x-scalar-scripts` OpenAPI extension so we can add these on the fly ✨ 
- Add way more test functionality (do we just use `vitest` browser only?
- This is just at the request level, but we can eventually do: Folder, Collection and workspace level : )
- We should also be able to send requests from the pre request script (foundationally request chaining 😏 )

i disabled this for the Modal mode, so we can get this merged and play around with it as we build : ) but it's pretty fun!!

### DIFF
--- a/packages/api-client/src/libs/scalarScripts.ts
+++ b/packages/api-client/src/libs/scalarScripts.ts
@@ -1,0 +1,74 @@
+import type { ResponseTest } from '@scalar/oas-utils/entities/workspace/spec'
+import type { AxiosRequestConfig, AxiosResponse } from 'axios'
+
+function getPreRequestScalarObject(config: AxiosRequestConfig) {
+  const _config = config
+  return {
+    environment: {
+      set: (key: string, value: string) => {},
+      get: (key: string) => '',
+    },
+    request: {
+      headers: {
+        add({ key, value }: { key: string; value: string }) {
+          _config.headers = _config.headers || {}
+          _config.headers[key] = value
+        },
+      },
+    },
+    _config,
+  }
+}
+export const executePreRequestScript = (
+  script: string,
+  config: AxiosRequestConfig,
+) => {
+  const scalar = getPreRequestScalarObject(config)
+  const executeScript = new Function('scalar', script)
+
+  const output = executeScript(scalar)
+
+  return { output, config: scalar._config }
+}
+
+function expect(actual: any) {
+  return {
+    toBe: (expected: any) => {
+      if (actual !== expected) {
+        throw new Error(`Expected ${actual} to be ${expected}`)
+      }
+    },
+  }
+}
+
+function getPostRequestScriptObject(response?: AxiosResponse) {
+  const testResults: ResponseTest[] = []
+  return {
+    sub: (a: number, b: number) => a - b,
+    test: (title: string, func: () => boolean) => {
+      try {
+        func()
+        testResults.push({ title, passed: true })
+      } catch (e) {
+        testResults.push({ title, passed: false, error: e as string })
+      }
+    },
+    expect,
+    response,
+    _testResults: testResults,
+  }
+}
+export const executePostRequestScript = (
+  script: string,
+  response?: AxiosResponse,
+) => {
+  const scalar = getPostRequestScriptObject(response)
+  const executeScript = new Function('scalar', script)
+
+  const output = executeScript(scalar)
+
+  return {
+    output,
+    testResults: scalar._testResults,
+  }
+}
--- a/packages/api-client/src/libs/sendRequest.ts
+++ b/packages/api-client/src/libs/sendRequest.ts
@@ -7,10 +7,16 @@ import type {
   RequestExample,
   RequestExampleParameter,
   ResponseInstance,
+  ResponseTest,
 } from '@scalar/oas-utils/entities/workspace/spec'
 import { redirectToProxy, shouldUseProxy } from '@scalar/oas-utils/helpers'
 import axios, { type AxiosError, type AxiosRequestConfig } from 'axios'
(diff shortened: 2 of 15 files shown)


antlio:
this seems to be missing some styling, need someone to jump? @marclave

packages/api-client/src/libs/scalarScripts.ts
```diff
@@ -0,0 +1,74 @@
+import type { ResponseTest } from '@scalar/oas-utils/entities/workspace/spec'
+import type { AxiosRequestConfig, AxiosResponse } from 'axios'
+
+function getPreRequestScalarObject(config: AxiosRequestConfig) {
```
geoffgscott, line 4: Before we launch this let's move to vanilla fetch.

packages/api-client/src/libs/scalarScripts.ts
```diff
@@ -0,0 +1,74 @@
...
+  const executeScript = new Function('scalar', script)
+
+  const output = executeScript(scalar)
+
+  return { output, config: scalar._config }
+}
+
+function expect(actual: any) {
+  return {
+    toBe: (expected: any) => {
```
geoffgscott, line 36: Would be nice to have user generated messages here. Should take an optional second argument that is  `(actual: any, expected: any): string`

geoffgscott: If its provided then we should run that function to generate the message and fallback to our message

packages/api-client/src/libs/scalarScripts.ts
```diff
@@ -0,0 +1,74 @@
...
+        throw new Error(`Expected ${actual} to be ${expected}`)
+      }
+    },
+  }
+}
+
+function getPostRequestScriptObject(response?: AxiosResponse) {
+  const testResults: ResponseTest[] = []
+  return {
+    sub: (a: number, b: number) => a - b,
```
geoffgscott, line 47: All of these return elements need proper doc strings

packages/api-client/src/libs/scalarScripts.ts
```diff
@@ -0,0 +1,74 @@
...
+      try {
+        func()
+        testResults.push({ title, passed: true })
+      } catch (e) {
+        testResults.push({ title, passed: false, error: e as string })
+      }
+    },
+    expect,
+    response,
+    _testResults: testResults,
```
geoffgscott, line 58: underscore is not a pattern we generally use in our codebase.

packages/api-client/src/libs/sendRequest.ts
```diff
@@ -42,6 +48,7 @@ export const sendRequest = async (
   sentTime?: number
   request?: RequestExample
   response?: ResponseInstance
+  testResults?: ResponseTest[]
 }> => {
   let url = rawUrl
```
geoffgscott, line 53: No `let`.

packages/api-client/src/libs/sendRequest.ts
```diff
@@ -176,15 +183,29 @@ export const sendRequest = async (
   // Append new query string to the URL
   url = `${urlWithoutQueryString}${queryString ? '?' + queryString : ''}`
 
-  const config: AxiosRequestConfig = {
+  let config: AxiosRequestConfig = {
```
geoffgscott, line 186: No let. Use object assign, objectMerge or override properties as needed.

packages/api-client/src/libs/sendRequest.ts
```diff
@@ -203,23 +224,51 @@ export const sendRequest = async (
...
+    // We can handle the post request scripts now
+    if (example.postSendScript) {
+      try {
+        const postRequest = executePostRequestScript(
+          example.postSendScript,
+          response,
+        )
+        testResults = postRequest.testResults
+      } catch (error) {
+        console.warn('Post-send script error:', error)
```
geoffgscott, line 236: Some display of these errors to the user would be sweet. Toast?

packages/api-client/src/libs/sendRequest.ts
```diff
@@ -203,23 +224,51 @@ export const sendRequest = async (
...
       sentTime: Date.now(),
       request: example,
       response: {
         ...response,
         duration: Date.now() - startTime,
       },
+      testResults,
     }
   } catch (error) {
     const axiosError = error as AxiosError
```
geoffgscott, line 250: This is a bad typecasts. Its not guaranteed to be an axios error.

geoffgscott (changes requested):
A few changes needed. Awesome start.

marclave:
moving this to draft for now :)

marclave:
well this ticket got stale quick : ) let's re-add it back post workspace entities PR

bretthowell714-source (approved):

dastanko:
Any specific reason why this feature is not deployed?

packages/api-client/src/libs/scalarScripts.ts
fiverrwixs-cmyk: hi

PR #2421, opened 2024-07-08
feat: add `void.scalar.com` and `galaxy.scalar.com` deployment

hanspagel:
WIP

**Tasks**

- [ ] Deployment of the void server fails, need to fix that or separate the PR for both projects 

This PR migrates https://github.com/scalar/void and https://github.com/scalar/galaxy, which just hold everything that’s necessary to deploy to void.scalar.com and galaxy.scalar.com via Cloudflare.

Intially, I wanted to move them to the respective package folders, but two reasons why I didn’t:

* To properly integrate with Cloudflare, we need some additional tooling (wrangler, different Vite setup …) which didn’t feel like a good fit for the package folder.
* `galaxy.scalar.com` is combining @scalar/galaxy and @scalar/mock-server and it’s not really clear in which package it should live.

That said, I still think most of the examples should be moved to the package folders. But probably not `proxy-server`, `void-server` and `galaxy-server` - the things that are deploy as standalone projects intended for production use.


geoffgscott:
@hanspagel We should move these to GCP away from cloudflare. A single devops pipeline is just easier to manage and monitor.

PR #2422, opened 2024-07-08
feat(use-codemirror): add json linter

antlio:
this pr adds the lint option in use-codemirror package and enable json linter in api client request body (unstyled):

image


marclave (approved):
nice work @antlio

PR #2424, opened 2024-07-08
feat(api-client): deploy `@scalar/api-client` to `client.scalar.com`

hanspagel:
This PR adds a `@scalar/api-client` standalone example, which is deployed to https://client.scalar.com. Not yet, but once this PR is in main. :)

Here is the preview:
https://eba72db2.client-980.pages.dev/


marclave (approved):
@hanspagel i wanna merge this so bad but idk why nuxt is failing :despair:

PR #2425, opened 2024-07-08
fix(components): scalar modal

antlio:
this pr fixes scalar modal style issues as seen below:

**before**
image

**after**
image


hwkr:
🙏

PR #2428, opened 2024-07-08
fix: no rendering of fonts

marclave:
A recent upgrade to vue caused an error to not render the fonts! Ruh roh, this is fixed now with a simpler computed prop

image


cameronrohani (approved):
I can breathe again

PR #2432, opened 2024-07-09
fix: missing fonts in scalar app

marclave:
we were not loading the fonts in the scalar app :despair:


cameronrohani (approved):
thank you marc!

PR #2433, opened 2024-07-09
fix: make client icons match text font weights

cameronrohani:
after:
image

image


marclave (approved):
we have never been more back

antlio:
nice ⚡️should we also ser their size down too? feels slightly misaligned wdyt?

cameronrohani:
> nice ⚡️should we also ser their size down too? feels slightly misaligned wdyt?

Feel free to play around with it 👍

PR #2437, opened 2024-07-09
chore: rename `createScalarApiClient` to `createApiClientModal` (RFC)

hanspagel:
This PR renames `createScalarApiClient` to `createApiClientModal`, and this is why:

* The method mounts a modal, which needs to be opened to be visible.
* It’s living in `packages/api-client/src/Modal/api-client-modal.ts`
* The comment says: `/** Initialize Scalar API Client Modal */`

I’d also like to propose to drop the `Scalar` in the function name, for the following reasons:

* It’s consistent with what we have already: https://gist.github.com/hanspagel/a696fa7371263dedf9f3d11f2122807c
* It’s very unlikely we run into any conflicts with `createApiClientModal`
* `createScalarApiClientModal` becomes so long, I don’t like to read it anymore

Looking for feedback!

## Alternative

If we’d keep the `Scalar` in the method name, I’d rename the others as following:

* `createScalarApiClientModal`
* `createScalarMockServer`
* `createScalarVoidServer`

### DIFF
--- a/.changeset/two-cougars-draw.md
+++ b/.changeset/two-cougars-draw.md
@@ -1,0 +1,8 @@
+---
+'@scalar/api-client-react': patch
+'@scalar/api-reference': patch
+'@scalar/play-button': patch
+'@scalar/api-client': patch
+---
+
+refactor!: rename `createScalarApiClient` to `createApiClientModal`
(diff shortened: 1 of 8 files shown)


marclave (approved):
🫡

PR #2438, opened 2024-07-09
fix(api-reference): long base url

antlio:
this pr fixes long base url and make it scrollable as seen below:

**before**
image

**after**
image


marclave (approved):
nice fix @antlio  : )

PR #2439, opened 2024-07-09
chore: ignore changesets for deprecated packages

hanspagel:
This PR adds two of our deprecated packages to the changeset ignore list. We shouldn’t bother with changesets for packages we don't release anyway.


marclave (approved):
galaxy brain nice @hanspagel

PR #2440, opened 2024-07-09
feat(api-client): new `createApiClientApp` method 👀 

hanspagel:
This PR adds a new `createApiClientApp` method to mount the full-blown API Client. 👀 

**Preview**

Screenshot 2024-07-09 at 11 40 12


marclave (approved):
🫡

PR #2443, opened 2024-07-09
feat(api-client): request input

antlio:
this pr is a first stone in order to set up variable injection, by adding the code input component in request table relying on code mirror.

let's make sure the different recent added use case are well supported here too (min max, required)


marclave:
couple initial things, we should pass a flag to disable `tab` for single line inputs so you can tab focus into the next element : )

antlio (commented):
awesome addition, lgtm!

PR #2444, opened 2024-07-09
chore: revert ignore changesets for deprecated packages

marclave:
revert https://github.com/scalar/scalar/pull/2439


cameronrohani (approved):
based

PR #2445, opened 2024-07-09
feat(api-client): variable replacement

marclave:
super wip but feel free to take from it. :)

https://github.com/scalar/scalar/assets/6176314/bf793ac1-a228-4784-b22d-1fffa044edd0


packages/api-client/package.json
```diff
@@ -107,6 +107,8 @@
   ],
   "module": "dist/index.js",
   "dependencies": {
+    "@codemirror/state": "^6.4.0",
+    "@codemirror/view": "^6.23.1",
```
antlio, line 111: these are already added within `use-codemirror` component, should we maintain that change?

marclave: im not entirely sure tbh

marclave: probs not but we would need to export them : )

amritk (approved):

PR #2446, opened 2024-07-10
fix(api-client): add keepalive to request for snappier loads

marclave:
now it loads super snappy! ✨


cameronrohani (approved):
Router View and keep alive, crazy

PR #2449, opened 2024-07-10
fix(components): copy button in `<ScalarCodeBlock />`

antlio:
this pr fixes the copy element position in scalar code block component while making it sticky for better reachability as seen below:

**before**
image
image

**after**
image


marclave (approved):
niceeeee @antlio !

PR #2456, opened 2024-07-11
feat(api-client): input select for enum and boolean

antlio:
this pr makes the previous data table input enum select more generic, add it to the code input component to bring back the enum support while bringing it to boolean type in the meantime:

image

### DIFF
--- a/.changeset/lovely-garlics-buy.md
+++ b/.changeset/lovely-garlics-buy.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
+
+feat: display select for boolean type
--- a/.changeset/wild-lies-cover.md
+++ b/.changeset/wild-lies-cover.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
+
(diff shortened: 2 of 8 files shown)


hanspagel (commented):
I didn‘t review the code, so sorry if you did it already: Fields may be nullable, should we add an explicit `null` option?

Not looked up, but there are two different ways to make a field nullable, something like that:

**OpenAPI 3.1**
```yaml
type:
- boolean
- null
```

**OpenAPI 2.x/3.0**
```yaml
type: boolean
nullable: true
```

hwkr (approved):
Looks good! 

Might be nice to push focus to the text input when you click "Add Value" but only if it's not a lot of work.

https://github.com/scalar/scalar/assets/6374090/c1683db2-b21d-40f0-82b7-4acb8660ff31

antlio:
> Looks good!
> 
> Might be nice to push focus to the text input when you click "Add Value" but only if it's not a lot of work.
> 

great call! added!

marclave (approved):

PR #2459, opened 2024-07-11
feat(api-client): request sidebar dragging

antlio:
this pr is a proposals to support sidebar dragging while adding some transition sugar:

[Uploading scalar-sidebar.mp4…](https://github.com/scalar/scalar/assets/14966155/feae7a15-e9b9-4836-80de-870afa235d07)

### DIFF
--- a/packages/api-client/src/components/Sidebar/Sidebar.vue
+++ b/packages/api-client/src/components/Sidebar/Sidebar.vue
@@ -13,14 +13,14 @@ const emit = defineEmits<{
 const { isReadOnly } = useWorkspace()
 const isDragging = ref(false)
 const sidebarRef = ref<HTMLElement | null>(null)
-const sidebarWidth = ref('250px')
+const sidebarWidth = ref('280px')
 
 /**
  * Resets the sidebar width to the default value to prevent
  * having it shrinked on click display after dragging.
  */
 const initialSidebarWidth = () => {
-  sidebarWidth.value = '250px'
+  sidebarWidth.value = '280px'
 }
 
 const startDrag = (event: MouseEvent) => {
@@ -29,7 +29,7 @@ const startDrag = (event: MouseEvent) => {
   const startX = event.clientX
   /** Current sidebar width when dragging starts */
   const startWidth = parseInt(
-    getComputedStyle(sidebarRef.value!).width || '250px',
+    getComputedStyle(sidebarRef.value!).width || '280px',
     10,
   )
 
@@ -38,7 +38,7 @@ const startDrag = (event: MouseEvent) => {
     document.body.classList.add('dragging')
     let newWidth = startWidth + dragEvent.clientX - startX
     if (newWidth > 360) newWidth = 360
-    if (newWidth < 160) {
+    if (newWidth < 240) {
       newWidth = 0
       emit('update:showSideBar', false)
     } else {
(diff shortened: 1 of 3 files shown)


cameronrohani:
Not sure if this makes sense to pursue _right in this moment_ but there's some things that would ideally happen if we were to add this:

- Sidebar size doesn't save on page refresh / collapse
https://github.com/user-attachments/assets/7093cfed-81b6-4208-ad94-e8dd99bbd012

- Sidebar icon flickers when hide / show state 
https://github.com/user-attachments/assets/1fce654f-574d-46b1-bbaf-0cc51b4be283

Also If we add this here should we be adding it in other places i.e request / response?

antlio:
@cameronrohani 
> Sidebar size doesn't save on page refresh / collapse

added

> Sidebar icon flickers when hide / show state

this is not related to that PR, however we should def fix it!

> Also If we add this here should we be adding it in other places i.e request / response?

we could if its worth it!

packages/api-client/src/components/Sidebar/Sidebar.vue
```diff
@@ -128,10 +128,6 @@ watch(
   height: 100%;
   z-index: 99;
 }
-.sidebar-enter-active,
-.sidebar-leave-active {
-  transition: width 0.15s ease !important;
-}
 .sidebar-enter-from,
 .sidebar-leave-to {
   width: 0 !important;
```
antlio, line 143: does this feel too harsh without any transition?

packages/api-client/src/components/Sidebar/Sidebar.vue
```diff
@@ -158,7 +154,5 @@ watch(
   width: 100%;
   height: 100%;
   z-index: 100;
-  backdrop-filter: blur(1.5px);
-  transition: backdrop-filter 0.15s ease-in-out;
```
antlio: before element is to be removed if we don't rely on backdrop filter / transition

packages/api-client/src/components/Sidebar/Sidebar.vue
```diff
@@ -1,23 +1,179 @@
...
 const { isReadOnly } = useWorkspace()
+const isDragging = ref(false)
+const sidebarRef = ref<HTMLElement | null>(null)
+const sidebarWidth = ref(localStorage.getItem('sidebarWidth') || '280px')
+
+/**
+ * Resets the sidebar width to the default value to prevent
+ * having it shrinked on click display after dragging.
+ */
+const initialSidebarWidth = () => {
```
geoffgscott, line 20: Nitpick but would prefer a function declaration here and an action word for function naming such as:

```ts
function setInitialSidebarWidth()
```

marclave:
awesome pr @antlio ! i think there are some second order effects here that we need to think about in regards to resizing everything other sections, but also let's revisit post invite only release of the app ✨

hanspagel:
What if we get this in, but make sure it’s disabled/not active yet? me doesn’t like stale PRs @marclave @antlio

antlio:
> What if we get this in, but make sure it’s disabled/not active yet? me doesn’t like stale PRs @marclave @antlio

i cleaned/rebased it in case - will let @marclave make the last move (close/merge)

marclave:
let's get it in :) its a good idea, we can think second order effects later <3 great work @antlio

cameronrohani:
Awesome excited to get this in

I think the main side effect that we'll need to fix is having the references in the client will need to persist the sidebar size ideally...

I think we should keep the transition for close (did I get rid of it? I can try to add it back)

cameronrohani:
after marc and I played around with it for a bit we decided to remove all transitions, ideally in the closed state it would keep the additional border if ur cursor is still clicked down (see vscode)

PR #2462, opened 2024-07-11
fix(themes): don't minify themes css

hwkr:
Should fix the issue with the themes editor CSS being minified.

image


hwkr:
It looks like this increases the bundle size a tiny bit, @hanspagel thoughts?

hwkr:
> can we conditionally not minify string imports? we really only need this for the custom css editor?

Yeah just for the theme editor I think, I can try! I'm not 100% sure how to do that but I'll play around, @amritk do you have any ideas>

amritk:
Output both minified and non-minified versions and just grab the one you need

hanspagel:
CSS can’t really be minified, it’s just the whitespace that is removed. And no matter how we serve the CSS, it should all be gzipped/brotlied/…, where spaces are well compressed. I’d say there is no need to bother.

hanspagel (approved):
I tested the PR and with the `?raw` imports I still saw minified CSS for the presets in the `dist/` folder.

I tried `cssMinify: false` in the Vite config, and that really made the change for me. The CSS files in the `dist/` folder aren’t minifed anymore, no matter if the presets are imported with `?inline` or `?raw`.

I’m pretty sure I saw `cssMinify: false` in an earlier version of this PR, though. Am I missing something here? 🤔

hwkr:
@hanspagel `?raw` still minimizes the the CSS files but the theme editor uses the CSS string imported into JS which are in the `dist/presets/*.css2.js` files. Setting `?raw` stops those from being minified.

I'm fine with doing it either way but the `?raw` option should be working 🤷

PR #2474, opened 2024-07-12
fix: make default tag check computed for reactivity bugs

marclave:
**Problem**
Currently in the docs we lose reactivity for the show more button

**Explanation**
This happens because onMounted post parsedSpec

**Solution**
With this PR we make it a computed property just incase : )


cameronrohani (approved):
genius, thanks for fixing this 😎

PR #2480, opened 2024-07-13
fix: add max height to search modal

marclave:
can make it better later but marc-gpt css has been implemented

before:
image

after:
image


cameronrohani (approved):
AFUFU CSS

PR #2483, opened 2024-07-13
fix: only add body data if body data is not null

marclave:
**Problem**
Currently we always send the body on http requests

**Explanation**
This happens because we dont do any null checks

**Solution**
With this PR we null check and only add it the body is not null


packages/api-client/src/libs/sendRequest.ts
```diff
@@ -180,9 +180,11 @@ export const sendRequest = async (
     url: redirectToProxy(proxyUrl, url),
     method: request.method,
     headers,
-    data,
   }
 
+  console.log(data)
```
amritk, line 185: Remove log

amritk (approved):

PR #2484, opened 2024-07-14
fix(api-client): update active body when content type changes

inyourtime:
**Problem**
Currently, when I test a request and the content type of the body is JSON, changing the content type of the body to others, such as multipart/form-data, does not update the activeBody. This results in the body remaining as JSON when the request is sent.

**Solution**
With this PR, I have added a function to update the activeBody when the content type changes.

This is my second contribution. If there's anything else, you can let me know. 😅

Screenshot 2567-07-14 at 21 27 40
Screenshot 2567-07-14 at 21 27 51

### DIFF
--- a/.changeset/late-keys-push.md
+++ b/.changeset/late-keys-push.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
+
+fix: update active body when content type changes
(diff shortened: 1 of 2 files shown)


marclave (approved):
great work @inyourtime !! love to see a second contribution ✨ 

Are you able to add a patch changeset for `api-client` :) then we can get this merged and released!

PR #2488, opened 2024-07-15
fix: add reply response in fastify-api-reference example.

inyourtime:
**Problem**
Currently, when I try using the example from the `fastify-api-reference` and send a request, I don't get anything because it doesn't return a response and the API request remains pending.

**Solution**
With this PR, I ensure the API responds correctly.

Screenshot 2567-07-15 at 14 27 35

### DIFF
--- a/examples/fastify-api-reference/src/index.ts
+++ b/examples/fastify-api-reference/src/index.ts
@@ -27,7 +27,7 @@ await fastify.register(fastifySwagger, {
 })
 
 // Add some routes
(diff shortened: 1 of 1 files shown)


hanspagel:
Good catch!

hanspagel (approved):
Great contribution, let’s get this in! I’ve just updated the example a little bit more.

marclave (approved):
awesome contribution @inyourtime

PR #2489, opened 2024-07-15
ci: update turborepo cache after build

hanspagel:
I noticed that the Turborepo cache is (still) not leveraged properly. We’re doing the build and then pick up the cache in other jobs, just to still see a high cache miss rate. Note how the type check and others are still taking forever:

Screenshot 2024-07-15 at 12 16 04

So I’ve checked the logs and noticed this one in the Post Cache step:

> Cache hit occurred on the primary key turbo-Linux-node-20, not saving cache.

So … it found the cache and decided to not update the cache. This is not what we want. :)

With this PR the cache should be updated after the full build.

For a full-rebuild it looks like this now (notice how the type check and others are way faster):

Screenshot 2024-07-15 at 12 28 18

Because build time is down to 0-1s in those jobs:

Screenshot 2024-07-15 at 12 30 02

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -56,6 +56,11 @@ jobs:
           key: turbo-${{ runner.os }}-node-${{ matrix.node-version }}
       - name: Build
         run: pnpm turbo build
+      - name: Update Turborepo cache
+        uses: actions/cache/save@v4
+        with:
+          path: .turbo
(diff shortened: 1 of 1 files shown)


hanspagel:
Sorry, I need this badly. I have a bug that only appears in CI and I wait 10mins to get results on my changes.

I don’t see how this could break anything, so I’m skipping the queue here.

PR #2494, opened 2024-07-15
feat(doc-2095): handle multiple workspace part 1 - routing

amritk:
This only does the routing portion, but it seems to be working. Just need to see if I caught every case. Just poke around to different pages and see if anything breaks.


marclave (approved):
working perfect 🫡

PR #2497, opened 2024-07-15
fix(components): polish history icon

hwkr:
## Before / After:

https://github.com/user-attachments/assets/f109af61-fae7-41ee-b754-20cdd3b2dd7b


cameronrohani (approved):
awesome work!

marclave (approved):

PR #2498, opened 2024-07-16
feat(components): add outline icons

hwkr:
Adds a number of new outline icons to be used in the Dashboard.

Google Chrome-2024-07-15-17-48-20@2x

It adds the following icons:

- House Icon
- Pencil Icon
- Google Icon

It also updates two icons:

1. The Link icon which isn't used anywhere (yet).
2. The Play icon which is now a line icon by default.

The Play icon can still have a fill added to it which I've done for all of the places we use it already. Video to make sure the Play icon is visually the same:

https://github.com/user-attachments/assets/58c443fc-8819-4691-8f2b-96d6c85feaad


geoffgscott (approved):
Looks good but are those style changes to client intentional?

hwkr:
> Looks good but are those style changes to client intentional?

Yeah that's to account for the changes to the play icon

cameronrohani (changes requested):
I think we were wanting to keep the play button filled to draw more attention to it, but other than that everything looks great!

cameronrohani (approved):
nvm saw the vid, great work

marclave (approved):

PR #2502, opened 2024-07-16
feat(api-reference): hide schemas with `x-internal: true`, fix #2468

### LINKED ISSUE #2468: internal schema object not excluded from models
> when you set  x-internal: true as a property on a schema object definition, it is excluded from the "Models" section of the generated docs.

_Originally posted by JP in [discord](https://discord.com/channels/1135330207960678410/1135330208627564597/1259991302221856898)

hanspagel:
We hide operations and webhooks with `x-internal: true`, but we don't hide schemas. With this PR schemas/models with `x-internal: true` are hidden, too.

See #2468

### DIFF
--- a/.changeset/four-emus-pull.md
+++ b/.changeset/four-emus-pull.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
+feat: hide schemas with `x-internal: true`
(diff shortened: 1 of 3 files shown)


marclave (approved):
:galaxy-brain:

nice @hanspagel

PR #2504, opened 2024-07-16
feat(api-reference): configure `defaultHttpClient`, fix #2075

### LINKED ISSUE #2075: Configure default HTTP client
Hi all,
We would love to be able to set a defaultClient. the reason is that our likely target audience is a TS/JS developer, so instead of cUrl we would like to show a Node option by default. Is it correct that this currently is not possible or did I overlook this option? Would others also benefit from setting this default? 

Thnks!

_Originally posted by @flipvh in https://github.com/scalar/scalar/discussions/2074_

hanspagel:
This PR introduces a new configuration attribute `defaultHttpClient` to configure the HTTP client, that’s selected by default.

Comes with tests and docs.

The `useHttpClientStore` grew a lot already, so I took the opportunity and improved/added the code comments.

See #2075.


marclave (approved):
@hanspagel you are just COOKING

marclave:
@hanspagel can merge after conflicts ✨

PR #2508, opened 2024-07-16
chore: rename Webhook to Webhooks

hanspagel:
In the sidebar, we have Model*s*, but for webhooks we just write Webhook. 🤔 This PR adds an S, so we end up with:

Screenshot 2024-07-16 at 16 44 51


marclave (approved):
Good call :)

PR #2514, opened 2024-07-16
feat(doc 2095): handle multiple workspace part 2

amritk:
This is the rest of the code to handle multiple workspaces. Now some of these are drastic changes as we go from one workspace -> many so it needs some solid testing.

I will continue testing it as this PR goes up.

To reset just hit `localStorage.clear()` and refresh


marclave (approved):
the future is here ✨

PR #2517, opened 2024-07-17
fix(api-reference): set background color on body, fix #2496

### LINKED ISSUE #2496: iOS Dark Mode white status bar
#### On safari on iOS, the collapsed navigation bar shows as white when in dark mode on the default theme.

I assume this is the case on all themes.

#### On the kepler theme, both the navigation bar and status bar appear white.

---

default                | kepler
:----------------:|:----------------:
92AFA735-3653-4ADA-A608-42DF6E4E870B_1_102_o | 435F3893-7B38-4D89-ADC7-FDE03E4AE85D_1_102_o

---

### Workaround
This seems to be caused by the fact that background-color is not set. By adding custom css
```css
body {
  background-color: var(--scalar-background-1);
}
```

we can achieve the following output (which fixes the issue on all themes I've tested)

---

kepler
:-------:
477A153D-3319-4317-BE83-7DF25E4CF927_1_102_o

---

### Environment

@scalar/hono-api-reference@^0.5.107

### Steps

1. Add a basic scalar reference route
```ts
app.get("/docs", apiReference({
  spec: {
    url: SPEC_URL,
  },
});
```
2. Navigate to that route using Safari on an iOS device
3. Scroll down and observe the collapsed navigation bar is white

### OpenAPI Document

https://forestpark.cecs.pdx.edu/staging/v1/openapi

hwkr:
Sets the background color on the `<body>` tag in a layer so it's set for iOS. It turns out this also colors the address bar in addition to the top chrome which is nice.

Since it's in a layer it _should_ be easy to override if anyone doesn't want it set for their use.

Fixes #2496

### DIFF
--- a/.changeset/tidy-peas-call.md
+++ b/.changeset/tidy-peas-call.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
+fix: set background color on body
--- a/packages/api-reference/src/components/ApiReference.vue
+++ b/packages/api-reference/src/components/ApiReference.vue
@@ -104,7 +104,10 @@ const { parsedSpec, rawSpec } = useReactiveSpec({
   </Layouts>
 </template>
 <style>
-body {
-  margin: 0;
+@layer scalar-base {
+  body {
+    margin: 0;
+    background-color: var(--scalar-background-1);
+  }
 }
 </style>


hanspagel (commented):
Sorry, I’m not up to date on the `@layer` thing. Do we want to introduce this again?

But looks good to me! Would be great to get @cameronrohani’s review.

cameronrohani (approved):
hopefully no one has weird margin values on body 😨

hwkr:
> hopefully no one has weird margin values on body 😨 

I mean if they do presumably they want it that way?

PR #2518, opened 2024-07-17
chore: make react a peer dependency, fix #2482

### LINKED ISSUE #2482: Cannot read properties of null (reading 'useRef')
### What happens?

image

run success

// src/App.jsx
import { ApiReferenceReact } from '@scalar/api-reference-react'
import React, { useState } from 'react';

function App() {
  return (
    <ApiReferenceReact
      configuration={{
        spec: {
          url: 'https://cdn.jsdelivr.net/npm/@scalar/galaxy/dist/latest.yaml',
        },
      }}
    />
  )
}

export default App

--------
I just use demo code to run , but throw error;

 "dependencies": {
    "@ant-design/icons": "^5.3.6",
    "@docusaurus/core": "^3.1.1",
    "@docusaurus/preset-classic": "^3.1.1",
    "@mdx-js/react": "^3.0.0",
    "@scalar/docusaurus": "0.4.44",
    "antd": "^5.16.1",
    "clsx": "^2.0.0",
    "fs-extra": "^11.2.0",
    "prism-react-renderer": "^2.3.0",
    "react": "^18.2.0",
    "react-dom": "^18.2.0"
  },

### OpenAPI Document

_No response_

hanspagel:
As correctly stated in https://github.com/scalar/scalar/issues/2482#issuecomment-2231993359, React and Nextjs should be peer dependencies.

Means, they have to be installed, but we don’t care about the exact version as long as it’s in the React 18.* or Next 14.* range.

See #2482 for more context on why it’s probably fixing the issue.

### DIFF
--- a/.changeset/pink-dots-march.md
+++ b/.changeset/pink-dots-march.md
@@ -1,0 +1,7 @@
+---
+'@scalar/nextjs-api-reference': patch
+'@scalar/api-reference-react': patch
+'@scalar/docusaurus': patch
+---
+
+chore: make react and next a peer dependency
--- a/packages/api-reference-react/package.json
+++ b/packages/api-reference-react/package.json
@@ -42,8 +42,7 @@
   ],
   "module": "./dist/index.js",
   "dependencies": {
-    "@scalar/api-reference": "workspace:*",
-    "react": "^18.3.1"
+    "@scalar/api-reference": "workspace:*"
   },
   "devDependencies": {
     "@types/react": "^18.2.60",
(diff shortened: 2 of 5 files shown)


marclave:
adding @amritk cause i think there was a reason why we did this

marclave (approved):

amritk:
Yes it was due to yarn not installing peer dependencies so we just moved them all to deps. And as stated users using a react package will have react installed so this should be fine.

PR #2521, opened 2024-07-17
fix: table checkbox for required params on import

marclave:
this pr https://github.com/scalar/scalar/pull/2469 unfortunately introduced a regression for the api client standalone app, now we just set enabled on import :)

### DIFF
--- a/packages/api-client/src/store/workspace.ts
+++ b/packages/api-client/src/store/workspace.ts
@@ -197,6 +197,7 @@ const createParamInstance = (param: OpenAPIV3_1.ParameterObject) =>
           : '',
     description: param.description,
     required: param.required,
+    enabled: !!param.required,
     enum: param.schema?.enum,
     type: param.schema?.type,
     format: param.schema?.format,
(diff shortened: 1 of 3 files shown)


packages/api-client/src/store/workspace.ts
```diff
@@ -197,6 +197,7 @@ const createParamInstance = (param: OpenAPIV3_1.ParameterObject) =>
           : '',
     description: param.description,
     required: param.required,
+    enabled: !!param.required,
```
amritk, line 200:
```suggestion
    enabled: Boolean(param.required),
```
nit pick but this one is a bit more succinct

amritk: but also shouldn't even be needed?

amritk (approved):
Nit pick comment/question

packages/api-client/src/store/workspace.ts
marclave: woah ive never seen this before 🤯 yeah i think youre right, lemme remove

marclave: well, it could be undefined 
image

PR #2523, opened 2024-07-17
feat: revamp address bar

marclave:
now we have servers as env vars, you can select and update the entire address bar!


marclave:
verbally approved w @cameronrohani @amritk 🫡

PR #2524, opened 2024-07-17
chore: hoist workspace packages

tmastrom:
Hoist the following packages to the workspace level

tsc-alias
vite-node
vue-tsc
eslint
@typescript-eslint/eslint-plugin
tsconfig-paths
nodemon
ts-node


marclave (commented):
i unfortunately am not the build person + have a bit of 🫨 from build stuff. 

so gunna let @geoffgscott @hanspagel @amritk test + validate this across all our packages +  testing in fresh installs of apps from stackblitz

geoffgscott (approved):
Looks good. @marclave This doesn't impact end packages as it's dev tooling hoisting only. It would fail if someone tried to use fork our workspace and use npm but I don't think that's a use case we care about.

marclave:
> Looks good. @marclave This doesn't impact end packages as it's dev tooling hoisting only. It would fail if someone tried to use fork our workspace and use npm but I don't think that's a use case we care about.

cool can fix up the conflicts and we can get this in!

hanspagel (approved):
Yesss, it’s awesome. :) Thanks @tmastrom!

PR #2526, opened 2024-07-18
fix(api-client): all request bodies are sent as form data, fix #2511

### LINKED ISSUE #2511: JSON Request Body Sent as FormData
### What happens?

When testing the request, selecting the JSON type for the request body results in the data being sent as FormData instead of JSON. This leads to unexpected behavior in the API and makes it difficult to test JSON endpoints properly.

Version: 
"@scalar/fastify-api-reference": "^1.24.46"

The request should be sent with the body formatted as JSON.

Steps to Reproduce:

1. Navigate to the request testing interface.
2. Select the request type as JSON.
3. Enter a JSON object in the request body.
4. Send the request.

### OpenAPI Document

`
    "/auth/login": {
      "post": {
        "operationId": "auth_login",
        "summary": "Login",
        "tags": ["auth"],
        "security": [{ "noAuth": [] }],
        "requestBody": {
          "required": true,
          "content": {
            "application/json": {
              "schema": {
                "type": "object",
                "properties": {
                  "email": { "type": "string", "format": "email", "description": "The email address of the user." },
                  "password": { "type": "string", "description": "The password of the user in plain-text." }
                },
                "required": ["email", "password"]
              }
            }
          }
        },`

marclave:
**Problem**
Currently we don't save the raw encoding value and if you change we dont properly set the content-type

**Explanation**
This happens because we dont handle it

**Solution**
With this PR we save both the raw encoding value and set the Content-Type header depending on the content type

https://github.com/user-attachments/assets/771c273a-e29c-496d-9a79-ba9e9bf2a000

### DIFF
--- a/packages/api-client/src/views/Request/RequestSection/RequestBody.vue
+++ b/packages/api-client/src/views/Request/RequestSection/RequestBody.vue
@@ -9,7 +9,7 @@ import { useWorkspace } from '@/store/workspace'
 import { ScalarButton, ScalarIcon, ScalarListbox } from '@scalar/components'
 import { createRequestExampleParameter } from '@scalar/oas-utils/entities/workspace/spec'
 import type { CodeMirrorLanguage } from '@scalar/use-codemirror'
-import { computed, nextTick, ref, watch } from 'vue'
+import { computed, nextTick, ref, unref, watch } from 'vue'
 
 import RequestTable from './RequestTable.vue'
 
@@ -181,14 +181,24 @@ const updateRequestBody = (content: string) => {
 // }
 
 const updateActiveBody = (type: keyof typeof contentTypeOptions) => {
+  let contentTypeHeader = ''
   let bodyType: 'raw' | 'formData' | 'binary' = 'raw'
 
   if (type === 'multipartForm' || type === 'formUrlEncoded') {
     bodyType = 'formData'
+    contentTypeHeader = 'multipart/form-data'
   } else if (type === 'binaryFile') {
     bodyType = 'binary'
-  } else {
+    contentTypeHeader = 'application/octet-stream'
+  } else if (type !== 'none') {
     bodyType = 'raw'
+    const mappedType = type === 'other' ? 'html' : type
+    contentTypeHeader = `application/${mappedType}`
+    requestExampleMutators.edit(
+      activeExample.value.uid,
+      'body.raw.encoding',
+      mappedType,
+    )
   }
 
   requestExampleMutators.edit(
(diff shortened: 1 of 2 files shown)


packages/api-client/src/views/Request/RequestSection/RequestBody.vue
```diff
@@ -181,21 +181,52 @@ const updateRequestBody = (content: string) => {
...
+  } else if (type !== 'none') {
     bodyType = 'raw'
+    const mappedType = type === 'other' ? 'html' : type
+    contentTypeHeader = `application/${mappedType}`
+    requestExampleMutators.edit(
+      activeExample.value.uid,
+      'body.raw.encoding',
+      mappedType,
+    )
   }
```
geoffgscott, line 202: This logic should really be made functional. Rather than a load of if/else statements and a lot of lets a function that takes the type and just returns `{ bodyType, contentTypeHeader }` would be clearer (and more testable)

packages/api-client/src/views/Request/RequestSection/RequestBody.vue
```diff
@@ -181,21 +181,52 @@ const updateRequestBody = (content: string) => {
...
   )
+
+  if (contentTypeHeader) {
+    // now lets handle the header
+    const headersWithoutContentType = unref(
+      activeExample.value,
+    ).parameters.headers.filter(
+      (header) => header.key.toLowerCase() !== 'Content-Type'.toLowerCase(),
+    )
+    console.log('headersWithoutContentType', headersWithoutContentType)
```
geoffgscott, line 217: Stray log

packages/api-client/src/views/Request/RequestSection/RequestBody.vue
```diff
@@ -276,16 +307,16 @@ const selectedContentType = computed({
...
+  if (activeExample.value.body.activeBody === 'formData') {
+    if (
+      activeExample.value.body.formData &&
+      activeExample.value.body.formData.value.length > 0
+    ) {
+      return 'multipartForm'
+    }
+  } else if (activeExample.value.body.activeBody === 'raw') {
     return activeExample.value.body.raw.encoding
+  }
```
geoffgscott, line 319: The wrapper `if` is not needed here
```suggestion
  if (
    activeExample.value.body.activeBody === 'formData' &&
    activeExample.value.body.formData &&
    activeExample.value.body.formData.value.length > 0
  )
    return 'multipartForm'

  if (activeExample.value.body.activeBody === 'raw')
    return activeExample.value.body.raw.encoding
```

geoffgscott (commented):
Some style changes recommended

marclave:
OMG the weird syncing bug was related to https://github.com/scalar/scalar/pull/2569

🕺 

it works perfectly now

PR #2527, opened 2024-07-18
fix(api-client): make topbar draggable and fix offset

cameronrohani:
f i x e d
https://github.com/user-attachments/assets/8497524a-e072-4045-afa8-3f25b84fe2ad


marclave (approved):
this literally changes everything ✨

PR #2530, opened 2024-07-18
chore: exclude peer dependencies to avoid React version conflicts

hanspagel:
With #2518 we moved `react` and `next` to the peer dependencies. The build configuration excludes all dependencies from the bundle, but not the peer dependencies. This would have led to version conflicts because we were about to ship a specific React version to users.

With this PR, we’re also excluding peer dependencies for the relevant packages. This way, users won’t run into React/Next.js version conflicts.


marclave (approved):
🤯 nice

PR #2533, opened 2024-07-18
feat: store workspace in localStorage when using `createApiClientApp`

hanspagel:
We’ve had this ApiClientApp.vue and the App.vue in the root folder `@scalar/api-client`, both were pretty similar. The ApiClientApp.vue just lacked localStorage support. To keep the same structure between the Modal and the App “layout“ this PR introduced a layouts folder where both live, and have the same structure (but different code).

The default example in the root is gone, we’re using the playground instead. Which should be the same as we’ve had in the root.


amritk (approved):
code LGTM! Nice work hans. Did not test

marclave (approved):

PR #2537, opened 2024-07-18
feat: moved command palette to the client app base

amritk:
Need some QA but I believe this should do it. Now we can trigger the command palette from anywhere. Should be trivial to add the hotkey opening as well!

- also leaned more towards type inference instead of setting types
- switched the event bus event to be an object so we can pass metadata in the future (pre-select request etc)

### DIFF
--- a/packages/api-client/src/components/CommandPalette/TheCommandPalette.vue
+++ b/packages/api-client/src/components/CommandPalette/TheCommandPalette.vue
@@ -78,17 +78,17 @@ const availableCommands = [
       {
         name: 'Add Server',
         icon: 'Brackets',
-        route: '/servers',
+        path: '/servers',
       },
       {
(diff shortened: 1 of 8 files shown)


packages/api-client/src/components/CommandPalette/TheCommandPalette.vue
```diff
@@ -118,38 +114,29 @@ const searchResultsWithPlaceholderResults = computed(() =>
...
+
+  executeCommand(command)
 })
 
 whenever(keys.ArrowDown, () => {
-  if (!props.state.open) {
-    return
-  }
-
-  if (!window) {
```
amritk, line 150: Don't think we need to check for window here as the server can't hit keys

marclave (approved):
✨ niiiiceee

PR #2540, opened 2024-07-19
fix(docusaurus): theme dark mode variables + button color + classic theme icon uniformity

cameronrohani:
image

image

image

image


marclave (approved):
we are so back 🫡

PR #2541, opened 2024-07-19
feat(doc-2064): make examples functional in api client

amritk:
Made requests functional. We still have the double enter problem but I will leave that for another PR.
- fixed the phantom galaxy workspace
- built out the request example command palette
- added add and delete for examples and requests
- added routing for examples
- switched back to routerLink instead of @click handler on sidebar

### DIFF
--- a/packages/api-client/src/components/CommandPalette/CommandPaletteExample.vue
+++ b/packages/api-client/src/components/CommandPalette/CommandPaletteExample.vue
@@ -13,18 +13,23 @@ const emits = defineEmits<{
   (event: 'close'): void
 }>()
 
-const { activeWorkspaceRequests } = useWorkspace()
+const { activeRequest, activeWorkspaceRequests } = useWorkspace()
 const exampleName = ref('')
-const selectedRequest = ref(activeWorkspaceRequests.value[0])
+const selectedRequest = ref(activeRequest.value)
 
 function handleSelect(request: any) {
+  console.log('handleSelectasinsd')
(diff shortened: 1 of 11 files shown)


packages/api-client/src/components/CommandPalette/CommandPaletteExample.vue
```diff
@@ -7,24 +7,54 @@ import {
...
+  activeWorkspaceRequests,
+  requests,
+  requestExampleMutators,
+} = useWorkspace()
+
 const exampleName = ref('')
-const selectedRequest = ref(activeWorkspaceRequests.value[0])
+const selectedRequest = ref(
+  // Ensure we pre-select the correct request
+  requests[props.metaData ?? ''] ?? activeRequest.value,
```
marclave, line 35: `requests[props.metaData ?? '']` wont this error? ohhh but then it defaults

marclave (approved):
literally woz energy, tested across all our examples + apps 🫡 

he just can't be stopped

packages/api-client/src/components/CommandPalette/CommandPaletteExample.vue
amritk: Up this bit allows us to delete all requests. We can probably add an empty state later

amritk: Yep exactly. It will default

PR #2542, opened 2024-07-19
fix(api-client): can’t navigate to request history entry

marclave:
we properly use a mutator to store in localstorage
we properly navigate to historical requests


amritk (approved):
I would add that comment back in, other than that looks legit

PR #2544, opened 2024-07-19
fix(api-client): remove click.stop page refresh action item

marclave:
**Problem**
Currently, when you click `...` twice  it reloads the page

**Explanation**
This happens because of a click.stop headless recursive event

**Solution**
With this PR we just dont click.stop


amritk (approved):
Perfect

PR #2547, opened 2024-07-19
fix(api-client): search modal isn’t centered

marclave:
**before**
was stuck to the bottom

**after :)** 
image

### DIFF
--- a/packages/components/src/components/ScalarModal/ScalarModal.vue
+++ b/packages/components/src/components/ScalarModal/ScalarModal.vue
@@ -117,7 +117,6 @@ export const useModal = () =>
   top: 0;
   bottom: 0;
   right: 0;
-  margin: auto;
 }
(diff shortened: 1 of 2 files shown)


packages/components/src/components/ScalarModal/ScalarModal.vue
```diff
@@ -117,7 +117,6 @@ export const useModal = () =>
   top: 0;
   bottom: 0;
   right: 0;
-  margin: auto;
```
hwkr, line 120: 🙌

hwkr (approved):

packages/components/src/components/ScalarModal/ScalarModal.vue
hwkr, line 120: 🙌

PR #2548, opened 2024-07-19
feat(api-client): add cookie empty state

marclave:
we now have a default cookie added to a workspace : )


amritk (approved):
It works! But unrelated to this PR: how do we edit the name of a cookie

marclave:
> It works! But unrelated to this PR: how do we edit the name of a cookie

haha yeah that whole UX needs improvement

PR #2550, opened 2024-07-19
fix(client.scalar.com): localStorage isn’t enabled

amritk:
We are back!


marclave (approved):
we test on prod!

packages/api-client-app/electron.vite.config.ts
```diff
@@ -5,6 +5,9 @@ import { resolve } from 'path'
 export default defineConfig({
   main: {
     plugins: [externalizeDepsPlugin()],
+    define: {
```
amritk, line 8: @hanspagel this doesn't work, need a way to pass the env var in. Any ideas?

hanspagel (changes requested):
Why does it have to be an env var, can’t we make this a regular setting? 🤔 The same way we pass a specification.

marclave:
@hanspagel we can make it a different config later, its fine for now being a define : )

PR #2557, opened 2024-07-19
feat(api-client): create a default workspace if no OpenAPI document is passed to the modal

hanspagel:
Currently, we log an error when no OpenAPI document is passed to the API client modal.

This PR just creates an empty workspace instead. :)

See #2555


marclave (approved):
yeah i dig this ✨

PR #2559, opened 2024-07-19
feat(api-client): make the side nav a draggable region

hanspagel:
This PR makes the side nav (but not the buttons inside) a draggable region, so it’s easier to move the app window around:

https://github.com/user-attachments/assets/f14b822d-2343-4090-9169-ed71e9ab9d05

### DIFF
--- a/.changeset/cold-trains-guess.md
+++ b/.changeset/cold-trains-guess.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
+
+feat: make side navigation a drag region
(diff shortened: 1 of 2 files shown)


geoffgscott:
I don't really think this is expected behavior for a desktop app. Do you have examples of someone else doing it?

marclave:
@hanspagel i also couldnt find examples of this ux :)

hanspagel:
Not common at all! But we don’t have much space to drag the window between the tabs and after I tried to move the window that way twice, I just added it. 

Do you think it’d hurt to have it? I mean, it’s just doing nothing otherwise.

We could also think about moving the tabs out of the title bar, which is what they recommend in the human design guidelines, but that’d be a bigger change:
 
Screenshot 2024-07-22 at 13 15 37

cameronrohani:
I say we do it. I've added some transparency to the background (https://github.com/scalar/scalar/pull/2605) which makes it feel more draggable additionally I've found a few examples since of people doing it i.e arc browser

marclave (approved):

PR #2561, opened 2024-07-19
fix: release workflow permissions

tmastrom:
Add permissions for publishing to npm. Documentation [here](https://docs.npmjs.com/generating-provenance-statements#example-github-actions-workflow)


marclave (approved):
we are so back

PR #2563, opened 2024-07-19
chore(api-references): use thickness instead of css overrides

hwkr:
Just switched to using the thickness prop

Looks the same:

https://github.com/user-attachments/assets/246a58f0-d9b9-4c32-89fc-7163ae070cdc


cameronrohani (approved):

marclave (approved):
we are so back

PR #2565, opened 2024-07-20
fix(components): provide a placeholder for icons while loading

hwkr:
This add a placeholder `<svg>` while the async component is loading to prevent layout shift.

I think by the nature of async components it needs to load them in anytime the view is mounted / unmounted, it doesn't send additional requests for the icons once they've been loaded in. Apparently the small amount of time it takes them to load from cache is enough to cause a flicker but this PR minimizes that flicker.

## Before / After

https://github.com/user-attachments/assets/52761105-03b7-498c-88e7-a37e84e41931

### DIFF
--- a/.changeset/gentle-plants-cough.md
+++ b/.changeset/gentle-plants-cough.md
@@ -1,0 +1,5 @@
+---
+'@scalar/components': patch
+---
+
+fix: provide a placeholder for icons while loading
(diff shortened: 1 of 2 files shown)


amritk:
Loading from cache should be instant, there should be no delay no matter how small. We should look into that

hwkr:
> Loading from cache should be instant, there should be no delay no matter how small. We should look into that

These also a delay for loading the UI so I think it's just Vue rendering, we can have a look together if you want though.

amritk:
I'll test the PR tom and see if my original issue was fixed

amritk (approved):
Alright my test case has been solved. I think we can call this complete until it comes up again! Nice work

marclave (approved):

PR #2567, opened 2024-07-20
docs: update import to @scalar/hono-api-reference 

inyourtime:
This PR updates the README.md in `hono-api-reference` to correct the import.


marclave (approved):
ah great docs update  @inyourtime  :) tysm

PR #2568, opened 2024-07-20
fix: latest changes broke testing base url

marclave:
Currently we use a default of localhost, but this breaks for our opensource integrations that rely on the servers origin to fallback to for the server, now we default to the origin the server is hosted on! we might want to eventually do something different for the Scalar App where we actually _dont_ have a server populated if theres no server, and have an empty varible in an error state so the user can update it : )

but that can be a future thing

### DIFF
--- a/.changeset/shiny-trees-hunt.md
+++ b/.changeset/shiny-trees-hunt.md
@@ -1,0 +1,5 @@
+---
+'@scalar/oas-utils': patch
+---
+
+fix: use window location origin if no servers
--- a/packages/oas-utils/src/transforms/import-spec.ts
+++ b/packages/oas-utils/src/transforms/import-spec.ts
@@ -147,7 +147,10 @@ export const importSpecToWorkspace = async (spec: string | AnyObject) => {
     ? parsedSpec.servers!
     : [
         {
-          url: 'http://localhost',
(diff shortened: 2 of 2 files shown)


hanspagel:
Shouldn’t the server be overwritten by the reference (curently)? 👀

marclave:
> Shouldn’t the server be overwritten by the reference (curently)? 👀

it is, but this is if theres no servers

marclave:
@hanspagel scroll up a little bit and you'll see this comment

https://github.com/scalar/scalar/pull/2568/files#diff-792e7cc2a0535a4610c0eb90ee123a169c650eb52fbab0ddbd9b0a4fe383a9f2R145

hanspagel:
But the reference has various ways to fall back if no server is provided, and should then pass the result to the client.

marclave:
> But the reference has various ways to fall back if no server is provided, and should then pass the result to the client.

even if that were the case, the api client loads on it's own as a separate app so we need this agnostic of the references

marclave:
going to merge to get a fix in cause all of our local server integrations aren't working right now

marclave:
@hanspagel we can do your way next week : )

PR #2570, opened 2024-07-20
feat(api-client): rename requests and use delete modal

marclave:
now we have a delete modal and a rename modal 🕺 

also hides duplicate for now

https://github.com/user-attachments/assets/37329e38-af28-478d-b7bb-de40365e8bca


amritk (changes requested):
I think we can be a little more efficient on this one 🤓

amritk (approved):
On second thought, its good to go

Issue #2503, opened 2024-07-16
Freeze when performing a test request


clement4saunier:
### What happens?

Whenever I send a test request, the whole UI freezes forever. I have pretty heavy routes but even with smaller ones it still happens to some degree. 

https://github.com/user-attachments/assets/207e1453-f6ec-489e-b3a6-8f9a5bc1bf08

### What did you expect to happen?

When trying with swagger, there is a non-blocking loading state and the json ends up showing in little time. Due to it being completely stuck it's hard to tell what's going on, but I was once able to make a performance test and it showed the whole blocking time was UI rendering, the response from the api is also pretty quick

### How can we reproduce the issue?

Use `elysia` with `@elysiajs/swagger` on the latest version, scalar is the default provider.
Return 700kb of data on the request, run test request on scalar UI.

clement4saunier:
I love how modern scalar feels, but this is really a deal breaker

hanspagel:
Bummer! We recently revamped our API client, but I don’t think we did perf testing already. There might be some low hanging fruits here. 🤔

clement4saunier:
> Bummer! We recently revamped our API client, but I don’t think we did perf testing already. There might be some low hanging fruits here. 🤔

Yeah I've seen that you've been iterating quite a lot ! I went through some previous versions to test the same thing and had no luck finding one that was working fine, but I definitely seen big positive ui changes. Do you know of any previous versions that could have more stable requests ?

marclave:
> > Bummer! We recently revamped our API client, but I don’t think we did perf testing already. There might be some low hanging fruits here. 🤔
> 
> Yeah I've seen that you've been iterating quite a lot ! I went through some previous versions to test the same thing and had no luck finding one that was working fine, but I definitely seen big positive ui changes. Do you know of any previous versions that could have more stable requests ?

hey @clement4saunier ! im working on this perf issue rn, you can go back to 1.20 and it should be good.

we do some stuff with codemirror, so i might remove the "preview" feature for now : ) will update this thread in time!

marclave:
@clement4saunier after some serious debugging i found a bunch of performance issues:
- headless ui is causing re-render's way too much
- our ScalarCodeBlock is not performant at all for highlighting large (>500KB) responses
- I need to remove some `v-if`'s to `v-show`s :) 
going to be a ton of small PRs to improve this, but today im going to swap out ScalarCodeBlock to use CodeMirror (which does virtualization) really well and get a PR up : ) 

[...]

PR #2573, opened 2024-07-21
fix: remove ScalarCodeBlock that freezes browser in response preview

marclave:
**Problem**
Currently HighlightJS freezes for large responses > 700KB. https://github.com/scalar/scalar/issues/2503

Codemirror does a great job w large files + virtualization, so lets remove Highlight.js for the API Client Responses

In this PR I also remove "preview" mode, since it needs to be re-thought out :)


amritk (approved):
OoOoooo all deletions!! Did not test but I see red, I approve

PR #2575, opened 2024-07-21
fix(api-client): electrical wiring for gitbook

cameronrohani:
image


marclave (approved):
didnt know our CTO was elec/acc

PR #2577, opened 2024-07-21
chore(api-client-app): temp fix icon

marclave:


packages/api-client-app/src/main/index.ts
```diff
@@ -14,7 +16,7 @@ function createWindow(): void {
     titleBarStyle: 'hidden',
     trafficLightPosition: { x: 11, y: 12 },
     autoHideMenuBar: true,
-    ...(process.platform === 'linux' ? { icon } : {}),
+    icon: iconPath,
```
hanspagel, line 19: If that’s not set, it’s using the app icon. Let’s not change it, until we need to. :v:

packages/api-client-app/electron-builder.yml
hanspagel, line 1: That file won’t be used with toDesktop, you can ignore it.

packages/api-client-app/build/icon.png
hanspagel, line 1: Ohhh, where is the Scalar icon going? 🥲 I love the ASCII graphic, but I think the Scalar logo is the better app icon.

packages/api-client-app/src/main/index.ts
```diff
@@ -2,7 +2,9 @@ import { electronApp, is, optimizer } from '@electron-toolkit/utils'
 import { BrowserWindow, app, ipcMain, shell } from 'electron'
 import { join } from 'path'
 
-import icon from '../../build/icon.png?asset'
+const icon = join(process.env.PUBLIC || app.getAppPath(), 'build', 'icon.png')
+
+const iconPath = is.dev ? join(__dirname, '../../build/icon.png') : icon
```
hanspagel, line 7: What’s that doing/changing? 🤔

packages/api-client-app/build/icon.png
cameronrohani: You could be 100% right, any way I can test multiple logos/icons for this ya'll?

marclave:
closing cause idk what i was cooking here 😅

PR #2580, opened 2024-07-22
fix(mock-server): tries to register non-http methods as routes, fix #2560

### LINKED ISSUE #2560: Mock server fails to start if OpenAPI contains parameters at the path
### What happens?

The mock command parses parameters as an operation and fails with the following

```
[INFO] Example (OpenAPI v3.0)
[INFO] 1 paths, 2 operations

Available Paths

PARAMETERS /example/{exampleId}
GET    /example/{exampleId}

file:///workspace/node_modules/@scalar/mock-server/dist/createMockServer.js:87
            app[method](route, (c) => {
                       ^

TypeError: app[method] is not a function
    at file:///workspace/node_modules/@scalar/mock-server/dist/createMockServer.js:87:24
    at Array.forEach (<anonymous>)
    at file:///workspace/node_modules/@scalar/mock-server/dist/createMockServer.js:65:57
    at Array.forEach (<anonymous>)
    at createMockServer (file:///workspace/node_modules/@scalar/mock-server/dist/createMockServer.js:63:45)
    at async bootServer (file:///workspace/node_modules/@scalar/cli/dist/commands/mock/MockCommand.js:80:17)
    at async Command.<anonymous> (file:///workspace/node_modules/@scalar/cli/dist/commands/mock/MockCommand.js:66:18)

Node.js v18.20.2
```

Based on the sample spec below I would expect to only see a single GET operation

There are also a number of other valid fields on the path that might cause similar issues https://swagger.io/specification/v3/#path-item-object
- summary
- description
- servers
- parameters

Example Spec

```
openapi: 3.0.3
info:
  title: Example
  version: 1.0.0
paths:

  /example/{exampleId}:
    parameters: 
      - name: exampleId
        in: path
        description: Example ID.
        required: true
        schema:
          type: string
    get:
      summary: Get Example
      description: Get Example
      operationId: getExample
      responses:
        '200':
          description: Example.
        '404':
          description: Example not found.
      
```

`npx scalar mock ./example.yaml`

### OpenAPI Document

https://sandbox.scalar.com/e/qWABf

hanspagel:
The `@scalar/mock-server` just tries to add every key under `paths` as a route to Hono, which is non-sense. OpenAPI documents can have other attributes on that level, e.g. parameters.

This is funny, because TypeScript complained about this and was totally right and I just added a `@ts-expect-error` everywhere. 🤡 

Anyway, this PR fixes it. We’re only adding operations (which can only found in specific keys) as routes to Hono.

See #2560


marclave (approved):

marclave:
nice work @hanspagel ! readin through the code got me feeling 🫨 w TS :)

PR #2587, opened 2024-07-22
feat: revived the api-client-react package

amritk:
We include context/provider so that the modal can be launched from anywhere inside.


marclave (approved):
let's ship it!

PR #2588, opened 2024-07-22
fix(api-client): add header to client subpages

cameronrohani:
subpages now don't shift the whole app up
https://github.com/user-attachments/assets/15ba49d8-335f-42b2-b34b-75b92673dd4b


marclave (approved):
so good

PR #2597, opened 2024-07-23
feat: toDesktop builds

hanspagel:
This PR hooks up the Electron app with `toDesktop`. :) ~You~ I (because I have access) can now go to the Electron app folder and run `pnpm todesktop:build` to build, code sign and notarize the macOS app in the cloud.

This PR also fixes some issues we’ve seen in the app, most notable the empty start screen. This was caused by the history router, which doesn’t work in the Electron context. That’s why I introduced a new router based on **hash** routes.

With this PR the package is also integrated into the changeset workflow (set to `private: true`, so not publish to npm) to manage the version.


marclave (approved):
so rad!!

PR #2599, opened 2024-07-23
feat(api-client-react): keep use-client directive after build

amritk:
Need 'use-client' at the top of the file for next.js


marclave (approved):
LGTM!

PR #2601, opened 2024-07-23
fix(components): set component font size to 13px

hwkr:
Seems like at some point this got bumped to 12px in a bunch of places. @cameronrohani can you make sure the dropdowns looks the way you want them in the client / app / references?


cameronrohani (approved):
thanks! looks better to me

marclave (approved):

PR #2602, opened 2024-07-23
feat(components): create combo box component

hwkr:
It's a combobox! We might want to add some more features in the future but I think this is pretty good for now.

https://github.com/user-attachments/assets/11617f4d-8ab4-4bab-b853-5d68eb51f426


cameronrohani (approved):

marclave (approved):
new box who this, nice work @hwkr !

PR #2604, opened 2024-07-23
fix(api-reference): allow scrolling to collapsed sections via hash

### LINKED ISSUE #2603: Internal markdown links don't work if the section is collapsed
### What happens?

https://github.com/user-attachments/assets/bb760c7f-73aa-4c49-8167-ecf884b87327

We need to expand the section so it can be scrolled to.

1. Add an internal link to an endpoint in the description markdown
2. While the section is collapsed, try to click the link.

### OpenAPI Document

_No response_

hwkr:
Make sure the section is expanded so we can scroll to it on hash change. Also adds some internal links to the Scalar Galaxy description.

Fixes #2603

## Before: 

https://github.com/user-attachments/assets/2309ab56-b83c-48db-8f02-e510782c557f

## After:

https://github.com/user-attachments/assets/574b62a1-82d4-4a24-890e-938c65a9ad74

### DIFF
--- a/packages/api-reference/src/components/ApiReferenceLayout.vue
+++ b/packages/api-reference/src/components/ApiReferenceLayout.vue
@@ -107,9 +107,15 @@ onBeforeMount(() => updateHash())
 const scrollToSection = async (id?: string) => {
   isIntersectionEnabled.value = false
   updateHash()
-
-  if (id) scrollToId(id)
-  else documentEl.value?.scrollTo(0, 0)
+  if (id) {
+    // Open the section if it's collapsed
+    const sectionId = getSectionId(id)
+    if (sectionId !== id) {
+      setCollapsedSidebarItem(getSectionId(id), true)
+      await sleep(100)
+    }
+    scrollToId(id)
+  } else documentEl.value?.scrollTo(0, 0)
 
   await sleep(100)
   isIntersectionEnabled.value = true
--- a/.changeset/smooth-chairs-kneel.md
+++ b/.changeset/smooth-chairs-kneel.md
@@ -1,0 +1,6 @@
+---
+'@scalar/api-reference': patch
+'@scalar/galaxy': patch
+---
+
+fix(api-reference): allow scrolling to collapsed sections via hash
--- a/packages/galaxy/src/specifications/3.1.yaml
+++ b/packages/galaxy/src/specifications/3.1.yaml
@@ -2,7 +2,7 @@ openapi: 3.1.0
 info:
   title: Scalar Galaxy
   description: |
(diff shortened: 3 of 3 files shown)


packages/api-reference/src/components/ApiReferenceLayout.vue
```diff
@@ -107,9 +108,15 @@ onBeforeMount(() => updateHash())
...
-
-  if (id) scrollToId(id)
-  else documentEl.value?.scrollTo(0, 0)
+  if (id) {
+    // Open the section if it's collapsed
+    const sectionId = getSectionId(id)
+    if (sectionId !== id && !isCollapsedSidebarItem(sectionId)) {
+      setCollapsedSidebarItem(getSectionId(id), true)
+      await sleep(100)
+    }
```
amritk, line 117:
```suggestion
      setCollapsedSidebarItem(getSectionId(id), true)
      await sleep(100)
```

tbh you can just set it to true, its just a hashmap

packages/api-reference/src/hooks/useSidebar.ts
```diff
@@ -74,6 +74,10 @@ function setCollapsedSidebarItem(key: string, value: boolean) {
   collapsedSidebarItems[key] = value
 }
 
+function isCollapsedSidebarItem(key: string) {
+  return collapsedSidebarItems[key] ?? false
+}
+
```
amritk, line 80:
```suggestion
```

packages/api-reference/src/hooks/useSidebar.ts
```diff
@@ -360,6 +364,7 @@ export function useSidebar(options?: ParsedSpecOption & TagsSorterOption) {
     collapsedSidebarItems,
     toggleCollapsedSidebarItem,
     setCollapsedSidebarItem,
+    isCollapsedSidebarItem,
```
amritk, line 367:
```suggestion
```

amritk (approved):
so this will probably work 99% of the time and we do it everywhere already.

However, if you want to do the correct (new) way: we use mutation observer to detect when the element has loaded then scroll to it. 

But I'll approve this one and let you decide what you want to do as its not really a blocker

packages/api-reference/src/components/ApiReferenceLayout.vue
hwkr: We need to confirm sectionId !== id otherwise it triggers sometimes when it shouldn't

hanspagel (approved):
Great work! Approving the changes to the Scalar Galaxy example. 🏄

marclave (approved):

PR #2605, opened 2024-07-24
fix(api-client): some api client app clean up (font sizes, spacing, etc)

cameronrohani:
before:
image

after:
image

### DIFF
--- a/packages/api-client/src/components/SideNav/SideNav.vue
+++ b/packages/api-client/src/components/SideNav/SideNav.vue
@@ -11,17 +11,18 @@ const { currentRoute } = useRouter()
 <template>
   <nav
     aria-label="Side Navigation"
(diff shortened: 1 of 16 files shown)


packages/api-client/src/components/SideNav/SideNav.vue
```diff
@@ -11,17 +11,18 @@ const { currentRoute } = useRouter()
...
       <li
-        v-for="({ icon, name }, i) in ROUTES"
+        v-for="({ icon, name, prettyName }, i) in ROUTES"
         :key="i">
         <SideNavLink
           :active="(currentRoute.name as string | undefined)?.startsWith(name)"
           :icon="icon"
-          :name="name">
+          :name="name"
+          :prettyName="prettyName">
```
hwkr, line 25: nit: this could probably be called `label`

PR #2607, opened 2024-07-24
feat(api-client): enable multiple api clients

amritk:
We need to be able to have multiple separate api clients running. With the current setup our workspace store was global. It is now scoped to a vue app.

Whats left:

- [x] codevariablewidget
- [x] topnav
- [x] bug in api-client-react
- [x] intial empty workspace buggin with resources
- [x] activeRouterParams
- [x] play button
- [x] play button indentation
- [x] next.js
- [x] code variable


packages/api-client-react/playground/main.tsx
```diff
@@ -7,7 +7,7 @@ import App from './App'
 ReactDOM.createRoot(document.getElementById('root')!).render(
   <React.StrictMode>
     <ApiClientModalProvider
-      configuration={{
```
amritk, line 10: sometimes we use config, sometimes we use configuration. Which is it?

marclave (approved):
very happy with this! ✨

PR #2616, opened 2024-07-25
fix: remove layout code from scalar-app

cameronrohani:
with the addition of the "scalar-app" class to headless ui we needed to remove layout code that was being written on 'scalar-app'


marclave (approved):
:)

PR #2619, opened 2024-07-25
fix(api-client): client.scalar.com and add command for testing

amritk:
we can now test with `pnpm dev:client`

@hanspagel would it be possible to just move this into the api-client package?

### DIFF
--- a/examples/api-client/src/main.ts
+++ b/examples/api-client/src/main.ts
@@ -1,9 +1,7 @@
-import { router } from '@scalar/api-client'
-import { ApiClientApp } from '@scalar/api-client/layouts/App'
+import { createApiClientApp } from '@scalar/api-client/layouts/App'
 import '@scalar/api-client/style.css'
-import { createApp } from 'vue'
 
-const app = createApp(ApiClientApp)
-
-app.use(router)
-app.mount('#scalar-client')
+// Initialize
+await createApiClientApp(document.getElementById('scalar-client'), {
+  proxyUrl: 'https://proxy.scalar.com',
+})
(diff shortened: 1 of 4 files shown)


hanspagel:
My 🌶️ take: Everything that’s exposed or promoted, should have it’s own directory.

* In theory, we could move it to the package folder, but it has a few cloudflare depencies. Would feel wrong to have those in the package.
* Even if we trim those, what about the Electron app? This has pretty heavy dependencies, custom scripts, other ways to test …  that we can’t get rid of. 

So I’d get rid of all the generic examples, but I’d keep a `projects/` folder for the stuff that we expose/market:

```
└── projects
    ├── app
    ├── client.scalar.com
    ├── galaxy.scalar.com
    ├── sandbox.scalar.com
    └── void.scalar.com
```

(I wouldn’t use the domain names with a dot, but you get the idea.)

hanspagel:
I'm seeing errors on client.scalar.com, this PR probably fixes them, right?

Screenshot 2024-07-25 at 11 28 30

marclave:
> My 🌶️ take: Everything that’s exposed or promoted, should have it’s own directory.
> 
> * In theory, we could move it to the package folder, but it has a few cloudflare depencies. It would feel wrong to have those in the package.
> * Even if we trim those, what about the Electron app? The app has heavy dependencies, custom scripts, other ways to test …  that we can’t get rid of.
> 
> So I’d get rid of all the generic examples, but I’d keep a `projects/` folder for the stuff that we expose/market (no matter if it’s just a few lines or huge):
> 
> ```
> └── projects
>     ├── app
>     ├── client.scalar.com
>     ├── galaxy.scalar.com
>     ├── sandbox.scalar.com
>     └── void.scalar.com
> ```
> 
> (I wouldn’t use the domain names with a dot, but you get the idea.)
> 
> Re: Your PR
> 
> Would you mind to add a short explanation on why we’re making it a non-Vue app? I’ve had some ideas for it in mind, and they would require it to be a Vue app, so I’m wondering what it would cost to keep it like that.

this is what we do in org! so i am down for this ✨

PR #2620, opened 2024-07-25
feat(components): update combobox styles

hwkr:
Some small updates to the combobox styles based on how it would be used for request auth.

### DIFF
--- a/.changeset/giant-balloons-add.md
+++ b/.changeset/giant-balloons-add.md
@@ -1,0 +1,5 @@
+---
+'@scalar/components': patch
+---
+
+feat(components): update combobox styles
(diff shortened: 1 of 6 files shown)


cameronrohani:
Wow great work!

Should we add a placeholder - could just be "Search"

Also it seems like the search input is taking up too much vertical space - can we tighten that?

cameronrohani:
Wow great work!

Should we add a placeholder - could just be "Search"

Also it seems like the search input is taking up too much vertical space - can we tighten that?

amritk (changes requested):
I would remove all of the integration parts for now as auth is changing completely. Once we add multiple we can integrate

marclave (approved):

amritk (approved):
Send it

PR #2628, opened 2024-07-25
feat(api-client-app): store window size and position

hanspagel:
With this PR the window size and position of the app is restored on load.

In case the window is placed outside the display (even if it’s just parts of it), for example when switching from a big to a small screen, it’s moved inside the viewport again.


marclave (approved):
i was wondering how we would do this!!! nice @hanspagel

PR #2630, opened 2024-07-25
feat(api-reference): performance logging

hanspagel:
I’m debugging the performance and thought it would be cool to have a few key metrics in the console. We already had something for the dereferencing, this PR adds some metrics about the OpenAPI document fetching.

Open the browser console to see it.

```bash
fetch: 1552 ms (https://sandbox.scalar.com/files/korOI/openapi.yaml)
size: 621 kB
dereference: 181 ms
```


marclave (approved):
i dig this : )

PR #2633, opened 2024-07-25
feat: improve search modal performance, fix #2586

### LINKED ISSUE #2586: Search is very slow to open
### What happens?

I'd expect hitting Cmd+K to open the Search box nearly instantaneously for my API. 

Instead it's taking ~2-3 seconds to open. 

Go to - https://sandbox.scalar.com/e/korOI

Enter preview mode. Hit `Cmd + K`. See how long it takes to show up. Also pressing the esc key doesn't close the search box until the 2nd time you press it. See attached screen recording: 

https://github.com/user-attachments/assets/e0cdb712-6884-4fd9-aa90-f3e2a9f981ea

Perhaps there is a way to have an empty search box come up instead of a pre-populated one? 

Thanks! Would love to use scalar for our docs but the slow search response time is a bit of a non-starter..

### OpenAPI Document

https://sandbox.scalar.com/e/korOI

hanspagel:
This PR:

- Moves all the search code to a new `Search/` feature folder.
- Moves the code to fill the fuse search index to a `useSearchIndex` hook.
- Adds a basic test for the `useSearchIndex` hook.
- Adds the keyboard shortcut code for the search modal to a new `useKeyboardNavigation` hook.

But most importantly:

This makes opening the search modal way faster by limiting the displayed search results to 25. Rendering all can be very slow, and I just assume users don’t open the search modal to scroll through all entries.

See #2586 

Note: I’ve also introduced a new `Search` component, which eventually should do all the data wrangling, so the `SearchModal` is responsible for the rendering only. But when I found the cause for the slow modal, I stopped the refactor. I’ll continue in a separate PR.


marclave (approved):
absolutely wonderful! ✨

PR #2638, opened 2024-07-26
feat(api-client): enable multiple auth

amritk:
Its all about to change...


amritk:
This PR is no longer needed. New crazier PR inc soon

PR #2641, opened 2024-07-26
ci: reusable composite actions

hanspagel:
This PR introduces a composite action for the steps that are required to do a (full) build of the repository. A build (even if cached) is required for most jobs, so we’ve repeated the code over and over. But no more! This PR puts it into a single place.

Makes the CI workflow file smaller and a litte easier to grasp. :)

Read more: https://docs.github.com/en/actions/using-workflows/avoiding-duplication


marclave (approved):
ah nice this is a lot easier to understand! great work as always @hanspagel

PR #2643, opened 2024-07-26
feat(mock-server): headers, redirects and all content types

hanspagel:
This PR adds three great features to `@scalar/mock-server`:

**Headers**

Adds headers (defined in the OpenAPI document) to the mocked response.

**Redirects**

You can now define a redirect (`Location` header and status code 30*)

**Content types**

No matter what content type is given, it tries to pick the appropriate content type (based on the `Accept` header, falls back to JSON or what’s available). Even works with XML.

### DIFF
--- a/.changeset/cold-birds-doubt.md
+++ b/.changeset/cold-birds-doubt.md
@@ -1,0 +1,5 @@
+---
+'@scalar/mock-server': patch
+---
+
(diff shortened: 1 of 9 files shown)


marclave (approved):
🤯 code looks awesome, lets get this in great work 🤤

PR #2644, opened 2024-07-26
docs(api-client-app): release process

hanspagel:
This PR adds a description of the manual release process for our app.

Yes, we’ll automate this eventually, but for now this is how it works. :)


marclave (approved):
awesome : ) stoked to be running these also ;)

PR #2645, opened 2024-07-26
ci: output all published packages

hanspagel:
With this PR a list of all published packages should be echo’d in CI, right after publishing.

I’m just testing if and how that works to eventually hook into the process to automate the release of our app.

Unfortunately, I couldn’t come up with another way to test it, then actually adding it to the workflow and waiting for the next release. Even if it has errors, it shouldn’t break the release workflow, because it comes *after* everything is published.


marclave (approved):
it was allll echo'd ✨

PR #2650, opened 2024-07-26
fix(hono-api-reference): move api-reference to devDependency

amritk:
We had included @scalar/api-reference as a dependency but the hono package actually uses the cdn standalone. This moves it to a devDependency for the types.

from #2649


marclave (approved):

hanspagel:
Are we exposing the ReferenceConfiguration type though? Then we’d need to keep the package in the dependencies.

amritk:
@hanspagel Ah you are right, but it doesn't make sense to add a giant dependency for just a type. Should we just copy paste it over?

Should we copy it over?

hanspagel:
Yes, maybe we need `@scalar/types`, which could have the few types that we really want to expose and promote.

Sounds easier than it is though. Just as an example `ReferenceConfiguration` also includes other types like `AuthenticationState` and other complex stuff. :)

I’d like to work on it, but it’s still background processing in my mind.

amritk:
Yea actually a lot of the types already exist inside of oas-utils including auth. we can just move them into scalar/types

amritk:
Closed in favour of #2820

PR #2651, opened 2024-07-26
fix(api-client): tw-bg-base gitbook bug and some app polish

cameronrohani:
quick fix for when tw-bg-base isn't defined on cm-pill and some overall scalar app polish


marclave (approved):
🖖

PR #2654, opened 2024-07-26
ci: do not create PR comment in main

hanspagel:
In #2621 I’ve added a step to create PR comments, but that fails in `main` — because there’s no PR then. :)

This PR just skips the step in `main`.


marclave (approved):
<3 ur the best

PR #2656, opened 2024-07-26
ci: enable title linter again

hanspagel:
Ooops, I renamed the title linter workflow in some of the recent CI PRs. That was a mistake.

Another workflow showed the name of this workflow and I’ve tried to disable it to get rid of that bug, but I found another solution eventually.

Anyway, let’s just rename/enable it again. :)


marclave (approved):
<3

PR #2657, opened 2024-07-26
chore: retry flaky test

hanspagel:
There is this one test that is failing every now and then, and it really bugs me to manually rerun the jobs. 😅

This PR makes sure the test is executed up to 3 times. Still better then disabling it I’d say, and it’s probably just a race condition or something anyway.


marclave (approved):
literally galaxy brain

PR #2661, opened 2024-07-27
feat(api-client): new auth structure

amritk:
Alright new auth! Need to test both modal and client. Big changes

What's missing:
- [ ] need some ux for adding scopes


packages/api-client/src/components/DataTable/DataTableInput.vue
```diff
@@ -72,12 +74,13 @@ const handleDropdownMouseUp = () => {
     :class="containerClass">
     <div
       v-if="$slots.default"
-      class="text-c-2 flex min-w-[100px] items-center border-r-1/2 pl-2 pr-0">
+      class="text-c-2 flex min-w-[120px] items-center border-r-1/2 pl-2 pr-0">
```
amritk, line 77: I found this super strange, we are just using hard coded divs inside of a table. If we were using a table this would have resized automatically

marclave (approved):
🫡

PR #2662, opened 2024-07-27
fix: actually add workspace cookies to request

marclave:
**Problem**
Currently we dont send workspaces cookies to a request with the appropriate domain

**Explanation**
This happens because we dont check the domains

**Solution**
now we do!


amritk (approved):
Looks legit

PR #2663, opened 2024-07-27
feat(api-client): add collections page routing

marclave:
big draft energy, but this adds a collection page where we will eventually add:
- headers
- auth
- cookies

that get applied to all requests in the collection, which means we can add "inherit from parent" for auth :)


marclave:
closing in favor of something else 😅

PR #2667, opened 2024-07-28
feat: add environment selector to addressbar

marclave:
Now we have an active environment selector! Still needs some styling from @cameronrohani but this is a reallllly cool feature because, its probably the first openapi extension we should support asap!

image
image

and we dont show create environment in readOnlyMode : )

@hanspagel @amritk we should do something like

```
x-scalarEnvironments:
  activeEnvironment: some_id
  global: `string of environments ORRRR key values (which makes more sense probably`
  some_id: `some id yahooo`

```

then we can parse an OpenAPI spec, and have environments loaded for our users! 🤯 

Apart from styling we should also hook this into variable highlighting to pick the proper environment : )

We also might want to think about doing collection level environments vs workspace? so maybe the extension is specific if its a collection vs workspace collection

### DIFF
--- a/packages/api-client/src/views/Request/Request.vue
+++ b/packages/api-client/src/views/Request/Request.vue
@@ -1,6 +1,7 @@
 <script setup lang="ts">
 import { Sidebar } from '@/components'
 import AddressBar from '@/components/AddressBar/AddressBar.vue'
+import EnvironmentSelector from '@/components/EnvironmentSelector/EnvironmentSelector.vue'
 import SearchButton from '@/components/Search/SearchButton.vue'
 import SearchModal from '@/components/Search/SearchModal.vue'
 import SidebarButton from '@/components/Sidebar/SidebarButton.vue'
@@ -282,6 +283,7 @@ const getBackgroundColor = () => {
         </a>
       </div>
       <AddressBar />
+      <EnvironmentSelector />
       <div
         class="flex flex-row items-center gap-1 lg:px-1 lg:mb-0 mb-0.5 lg:flex-1 justify-end w-6/12">
         <!-- TODO: There should be an `ìsModal` flag instead -->
--- a/packages/api-client/src/store/workspace.ts
+++ b/packages/api-client/src/store/workspace.ts
@@ -497,6 +497,10 @@ export const createWorkspaceStore = (router: Router, persistData = true) => {
       workspaces[Object.keys(workspaces)[0]],
   )
 
+  const activeEnvironment = computed(
+    () => environments[activeWorkspace.value?.activeEnvironmentId ?? 'default'],
+  )
+
   /** Ordered list of the active workspace's collections with drafts last */
   const activeWorkspaceCollections = computed(() =>
     activeWorkspace.value?.collectionUids
@@ -819,6 +823,7 @@ export const createWorkspaceStore = (router: Router, persistData = true) => {
     activeWorkspaceServers,
     activeParsedEnvironments,
     activeWorkspaceRequests,
+    activeEnvironment,
     modalState,
     isReadOnly,
     router,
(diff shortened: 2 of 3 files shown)


cameronrohani:
image


@hwkr sorry to tag you on another but I'm not sure what's going on here

when you click the env selector it shifts a few px's to the right, maybe it's something super obvious but I couldn't seem to figure it out

https://github.com/user-attachments/assets/5471044b-ca58-42f0-b27b-704be1a8f3c2

hwkr:
@cameronrohani fixed :)

packages/api-client/src/components/EnvironmentSelector/EnvironmentSelector.vue
```diff
@@ -0,0 +1,102 @@
...
+const updateSelected = (uid: string) => {
+  workspaceMutators.edit(activeWorkspace.value.uid, 'activeEnvironmentId', uid)
+}
+
+const createNewEnvironment = () =>
+  router.push(`/workspace/${activeWorkspace.value.uid}/environment`)
+
+const envs = computed(() => {
+  return Object.values(environments).filter((env) => env.uid !== 'default')
+})
```
amritk, line 31:
```suggestion
const envs = computed(() => Object.values(environments).filter((env) => env.uid !== 'default'))
```
don't need returns with arrow functions ;)

amritk (approved):
LGTM!

hanspagel:
Rad! If someone is about to add the OpenAPI extension, pls consult so we can make it really OpenAPI-like together. 👍

packages/api-client/src/views/Request/Request.vue
```diff
@@ -285,6 +286,7 @@ const getBackgroundColor = () => {
       <div
         class="flex flex-row items-center gap-1 lg:px-1 lg:mb-0 mb-0.5 lg:flex-1 justify-end w-6/12">
         <!-- TODO: There should be an `ìsModal` flag instead -->
+        <EnvironmentSelector v-if="!activeWorkspace.isReadOnly" />
```
hanspagel, line 289: The TODO is for the line below I think

```suggestion
        <EnvironmentSelector v-if="!activeWorkspace.isReadOnly" />
        <!-- TODO: There should be an `ìsModal` flag instead -->
```

packages/api-client/src/components/EnvironmentSelector/EnvironmentSelector.vue
```diff
@@ -0,0 +1,102 @@
...
+</script>
+<template>
+  <div>
+    <ScalarDropdown>
+      <ScalarButton
+        class="font-normal h-auto justify-start py-1.5 px-1.5 text-c-1 hover:bg-b-3 text-c-2 w-fit"
+        fullWidth
+        variant="ghost">
+        <h2 class="font-medium m-0 text-sm flex gap-1.5 items-center">
+          {{ activeEnvironment?.name ?? 'No Environment' }}
```
antlio, line 41: should we display the `Global Environment` name here to be consistant with what we display in the environment page?

packages/api-client/src/components/EnvironmentSelector/EnvironmentSelector.vue
```diff
@@ -0,0 +1,102 @@
...
+              activeWorkspace.activeEnvironmentId === ''
+                ? 'bg-blue text-b-1'
+                : 'text-transparent'
+            ">
+            <ScalarIcon
+              class="size-2.5"
+              icon="Checkmark"
+              thickness="3.5" />
+          </div>
+          No Environment
```
antlio, line 65: same feeback as [for the button wording](https://github.com/scalar/scalar/pull/2667/files#r1704213789)

packages/api-client/src/views/Request/Request.vue
antlio: good catch, fixed

hanspagel (changes requested):
Oh, this is great! Just one thing I noticed: The `Add Environment Variable` actually creates a new environment. Maybe we can make it say `Add Environment` instead?

Ah, and the environment color isn’t persisted! After reload they are all grey.

hanspagel (approved):
Let‘s get this in! :)

PR #2686, opened 2024-07-28
fix(api-client): upload file in form-data with key

inyourtime:
**Problem**
Currently, when I try to upload a file with form-data, the key is sent as the file name instead of the correct key.

Screenshot 2567-07-28 at 18 19 09

**Solution**
With this PR, I ensure that the correct key is used instead of the file name when sending the request. Additionally, if both a file and a value are present, the file will be used for the request.

### DIFF
--- a/.changeset/wet-dolphins-poke.md
+++ b/.changeset/wet-dolphins-poke.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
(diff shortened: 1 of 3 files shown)


marclave (approved):
@inyourtime AMAZING PR

and tests!? 🤯 an example of a perfect pr

PR #2688, opened 2024-07-28
fix: superfluous yaml or json parse before spec parsing

marclave:
we only used the parsedSpec for checking "servers" but we really dont need that at all, we should let our amazing openapi-parser handle all things loading of spec and we then use the result

this fixes a bug where if a yaml file has newlines, or wrapped in string characters that it fails to load 😨


amritk (approved):
Sick play

PR #2690, opened 2024-07-28
feat(api-client): new settings page

marclave:
**Tasks**

- [x] If I change the theme and reload, it’s still the same theme.
- [x] If I uncheck the Scalar Proxy and reload, it’s enabled again.

currently we have no way to toggle the proxy in the scalar app, we are gunna have a settings page with shortcuts and a whole bunch of other things but for the time being we can just have a toggle for our proxy (and i added a theme picker)

https://github.com/user-attachments/assets/814b87dd-7f66-4553-804d-0553d3914302


amritk (approved):
Code looks good, would prefer a bit more comments bit its ok

hanspagel (changes requested):
Not sure if the home icon is the best fit for the settings page. 🤔 

When I hover the checkbox, I get a strange and huge box.

Tested here: https://975d05b1.client-980.pages.dev/workspace/default/settings/General

Screenshot 2024-08-08 at 13 35 57

geoffgscott:
+1 to @hanspagel comment that the home icon is bad choice. Should be a ⚙️.

antlio:
rebased, fixed and updated with a proposal

hanspagel (changes requested):
* If I change the theme and reload, it’s still the same theme.
* If I uncheck the Scalar Proxy and reload, it’s enabled again.

Can we persist this setting, too?

Screenshot 2024-08-27 at 15 04 48

I love the new settings icon! Isn’t settings worth an entry in the sidebar, though? Would be great to have the settings icon on top of the help (question mark) icon in the bottom left corner.

antlio:
> The text in the table doesn’t wrap on small screens. We can improve this in a seperate PR, though.

fixed by #2971

cameronrohani:
ignore this comment, doing some cross browser testing on gh images

cameronrohani:
<p>
	<a href="https://client.scalar.com/" target="_blank"></a>
	<a href="https://client.scalar.com/" target="_blank"></a>
</p>

hanspagel:
Did you hijack this PR to upload assets for the README? 😅

cameronrohani:
> Did you hijack this PR to upload assets for the README? 😅

it's for cross browser testing / mobile testing 😹  + for some reason svg's dont render on mobile in private repos so i just leave comments here and hope no one sees 😬

antlio:
@amritk do you mind giving a look on saving the state of the proxy option?

amritk:
> @amritk do you mind giving a look on saving the state of the proxy option?

Yep but not until later this week

antlio:
i set the local storage proxy so should be good to go @amritk @hanspagel

hanspagel (changes requested):
Looks great! Unfortunately, it seems themes aren't correctly loaded/changed. Some colors change, but not the background color for example.

image

antlio:
> Looks great! Unfortunately, it seems themes aren't correctly loaded/changed. Some colors change, but not the background color for example.

fair catch, that's because of the way client has been brought up to life first, i was thinking of fast follow fixing this but will do in that pr!

hanspagel (approved):
FINALLY

packages/api-client/src/hooks/useDarkModeState.ts
```diff
@@ -1,7 +1,7 @@
-import { ref, watch } from 'vue'
+import { onMounted, onUnmounted, ref, watch } from 'vue'
 
-// State
-const isDark = ref<boolean>(false)
+// State - null to indicate system preference
+const isDark = ref<boolean | null>(null)
```
hwkr, line 4: Might be nice to make this a union type instead, nullable booleans can be a little confusing
```suggestion
const colorScheme = ref<'light' | 'dark' | 'system'>('system')
```

antlio: yes good call! i updated in the last commit in that sense 💪

antlio:
latest updates:
- fixes route redirection to settings
- sets sidebar element action props to prevent display in setting ones
- fixes setting view border spotted by @hanspagel 
- favors color mode over is dark as suggested by @hwkr

hanspagel:
Looks like the dark mode setting is loaded on the settings page only. When I refresh on another page it’s using light mode, instead of the system preference (dark).

https://github.com/user-attachments/assets/a0bafaeb-4fe4-45f9-85c8-1eaa0cc78c6c

The sidebar hover effect makes the text disappear, and the active request state has a strange border (theme: moon).

Screenshot 2024-10-09 at 10 57 53

antlio:
latest updates:

- fixes color mode issue on other views caught by @hanspagel 
- updates border color presets of themes caught by @hanspagel

packages/api-client/src/hooks/useDarkModeState.ts
```diff
@@ -1,73 +1,98 @@
-import { ref, watch } from 'vue'
+import { onMounted, onUnmounted, ref, watch } from 'vue'
 
-// State
-const isDark = ref<boolean>(false)
+// State - null to indicate system preference
```
hwkr, line 3: Probably don't need this comment anymore

hanspagel:
```suggestion

```

Issue #2670, opened 2024-07-28
Websocket support


marclave:

Makeuptehrani:
Is it supported for Android?

Its4Nik:
@hanspagel is there any plan on this being done in the near future? I am currently writing an API with websockets and would love this feature. 

If not, could you tell me how this should be integrated? Then I would try and see what I could do.

MarkLee425:
> [@hanspagel](https://github.com/hanspagel) is there any plan on this being done in the near future? I am currently writing an API with websockets and would love this feature.
> 
> If not, could you tell me how this should be integrated? Then I would try and see what I could do.

Same here, I am currently using websocket with swagger and swagger-socket. Thinking to change this if there is support of websocket 🤔

MrSquaare:
More generally, this could support of the [AsyncAPI specification](https://www.asyncapi.com/docs/reference/specification/v3.0.0)

riprsa:
ngl, i would love to see WS support

nathan-muon:
Would love to see this as well!

PR #2691, opened 2024-07-28
feat(api-client): websocket support, see #2670

marclave:
wanted to take a super rough hacked-prototype-swing at this for a few hours and a lot of things came up! I wanted to just make it so we can add it to the drafts collection.. but then it made me think about the #2663 PR 🤔 

so some initial thoughts:

CollectionRoute:
- have a type in the collection -> OpenAPI | AsyncAPI
- render OpenAPI Collection Viewer vs AsyncAPI Viewer
- AsyncAPI Collection Viewer has different renders for "Websocket vs Graphql" that are root level
- we still have a "collection page" for both buuuuut theyre pretty different (no tests for websocket and stuff)
- we still render the same sidebar for both, but the address bar is different.
-  this actually kind of solves our addressbar conundrum, cause we probably wanna abstract out the sidebar to always render maybe we use a slot at the root of the routerview!

but yeah on a personal note i would love to have websocket support cause omg

https://github.com/user-attachments/assets/f2306aaf-caa9-4db9-8ff1-6d4bd1319232

### DIFF
--- a/packages/api-client/src/components/CommandPalette/CommandPaletteWebsocket.vue
+++ b/packages/api-client/src/components/CommandPalette/CommandPaletteWebsocket.vue
@@ -1,0 +1,130 @@
+<script setup lang="ts">
+import HttpMethod from '@/components/HttpMethod/HttpMethod.vue'
+import { useWorkspace } from '@/store/workspace'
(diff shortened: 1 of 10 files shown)


marclave:
closing in favor of coming back to this post workspace entities refactor

PR #2693, opened 2024-07-29
feat: add scalar-go lib under go integration

bdpiprava:


marclave:
@bdpiprava could you run `pnpm format` in the root :) seems like that test failed in ci

great pr! stoked to have this alongside our golang libs ✨

bdpiprava:
@marclave - Done

marclave (approved):
:)

PR #2697, opened 2024-07-30
feat(api-client): add drag and drop functionality

amritk:
Cases that need to be handled

- [x] collection cannot go into another collection
- [x] requests/folders cannot be loose outside of a collection
- [x] cannot drag into a request/examples, can only drag into folders or collections
- [x] handle exec data for collection re-order


marclave (approved):
drop it like its hot

PR #2706, opened 2024-07-30
feat(api-client): add enter to submit in command pallette inputs

amritk:


marclave (approved):
nice! ✨ works much better

PR #2709, opened 2024-07-31
feat(api-client): add response previews

hwkr:
Adds response previews for a number of different file / MIME types including but not limited to:

- HTML / Webpages
- Images like pngs, svgs and jpegs
- Audio and video like mp3s, mp4s and webm
- PDFs

Because we store our data as strings in localstorage we can't (easily) serialize the binary data so a request with binary data will show "Preview Unavailable" until you send the request again.

If you want to test it out I created a repo with a bunch of example files ([hwkr/ExampleFiles](https://github.com/hwkr/ExampleFiles)), I recommend serving them locally with something like [http-server](https://www.npmjs.com/package/http-server) so you can get correct MIME types in your content-type header (GitHub serves a lot of files as `text/plain`).

https://github.com/user-attachments/assets/55e7af53-e9aa-4e8f-b0ed-1ddaf2d309df


marclave (approved):
lets get this IN! great work @hwkr

Issue #2701, opened 2024-07-30
Delete Collections


marclave:
### What happens?

We should be able to delete collections with the context menu like how we support request modifications

### What did you expect to happen?

a

### How can we reproduce the issue?

a

PR #2710, opened 2024-07-31
feat(#2701): allow deleting and renaming folders and collections

amritk:
We added the ability to delete and rename collections as well as folders


packages/api-client/src/components/CommandPalette/CommandPaletteCollection.vue
```diff
@@ -42,7 +42,10 @@ onMounted(() => {
         id="collectionanme"
         ref="collectionInput"
         v-model="collectionName"
+        autocomplete="off"
         class="border-transparent outline-none w-full pl-8 text-sm min-h-8 py-1.5"
+        data-form-type="other"
```
amritk, line 47: stops the password managers from getting in the way

marclave (approved):
tested and based let's gooo

PR #2714, opened 2024-07-31
fix(client-app): dropdown add variable redirection 

antlio:
this pr is an update proposal in order to fix the redirection to the environment page, when add variable is selected from the environment variable dropdown.


marclave (approved):
yeah i like this

PR #2715, opened 2024-07-31
fix(client-app): back to requests redirection link

antlio:
this pr fixes the back to requests redirection link to be effective when navigating from another workspace than default:

image


marclave (approved):
nice @antlio ✨

PR #2718, opened 2024-07-31
feat(client-app): cookie removal

antlio:
this pr adds a cookie removal method proposal for the client app, in order to allow cookies to be removed + fallback on the last existing one.


marclave (approved):
nice work @antlio ✨

PR #2719, opened 2024-07-31
feat: use pathstring when req summary isn't available

brunogrcsada:
I've done quite a lot of specs previously were the path of the request is just used as the default identifier, which makes the most sense to me if there is no summary. No need to add the type as a prefix/suffix since we already display it.

This is just a simple fallback so that the summary is set to something sensible rather than being empty or having the same default label for all requests.

image


marclave (approved):
awesome @brunogrcsada !!

could you add a changeset :) then we can get this merged and released!

PR #2720, opened 2024-07-31
fix: word break in sidebar items and ellipsis positioning

cameronrohani:
fixes word break in sidebar

fixes ellipsis display: none -> display: flex causing word to wrap in sidebar

fixes sidebar icon stroke width

adds more contrast to multi auth


marclave (approved):
nice work @cameronrohani !

PR #2721, opened 2024-07-31
feat(api-client): support body params when importing 2.0 specs

brunogrcsada:
Not sure if there's a better way we want to abstract the different spec versions, but for now I've just included a very clear separate path for using body params, since requestBody was only introduced in 3.0

Documentation here: https://swagger.io/docs/specification/2-0/describing-request-body/

Happy to hear any thoughts if you would like this done differently; but works for now :D

### DIFF
--- a/packages/api-client/src/store/workspace.ts
+++ b/packages/api-client/src/store/workspace.ts
@@ -287,23 +287,39 @@ export const createWorkspaceStore = (router: Router, persistData = true) => {
       },
     }
 
-    if (request.requestBody) {
-      const requestBody = getRequestBodyFromOperation({
+    const bodyParams = Object.values(request.parameters.body)
+
+    let requestBody
+
+    if (bodyParams.length) {
+      requestBody = getRequestBodyFromOperation({
+        httpVerb: request.method,
+        path: request.path,
+        information: {
+          parameters: bodyParams,
+        },
+      })
+    } else if (request.requestBody) {
+      requestBody = getRequestBodyFromOperation({
         httpVerb: request.method,
         path: request.path,
         information: {
           requestBody: request.requestBody,
         },
       })
-      if (requestBody?.postData?.['mimeType'] === 'application/json') {
-        parameters.headers.push({
-          key: 'Content-Type',
-          value: 'application/json',
-          enabled: true,
-        })
-        body.activeBody = 'raw'
-        body.raw.value = requestBody.postData.text
-      }
+    }
+
+    if (
+      requestBody &&
+      requestBody?.postData?.['mimeType'] === 'application/json'
+    ) {
+      parameters.headers.push({
+        key: 'Content-Type',
+        value: 'application/json',
+        enabled: true,
+      })
+      body.activeBody = 'raw'
+      body.raw.value = requestBody.postData.text
     }
 
     // Check all current examples for the title and iterate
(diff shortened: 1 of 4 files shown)


packages/api-client/src/store/workspace.ts
```diff
@@ -287,23 +287,39 @@ export const createWorkspaceStore = (router: Router, persistData = true) => {
...
-          enabled: true,
-        })
-        body.activeBody = 'raw'
-        body.raw.value = requestBody.postData.text
-      }
+    }
+
+    if (
+      requestBody &&
+      requestBody?.postData?.['mimeType'] === 'application/json'
```
amritk, line 314:
```suggestion
    if (requestBody?.postData?.['mimeType'] === 'application/json'
```
the check for requestBody isn't requred since we are checking the mimeType already

packages/api-client/src/store/workspace.ts
```diff
@@ -287,23 +287,39 @@ export const createWorkspaceStore = (router: Router, persistData = true) => {
...
+        },
+      })
+    } else if (request.requestBody) {
+      requestBody = getRequestBodyFromOperation({
         httpVerb: request.method,
         path: request.path,
         information: {
           requestBody: request.requestBody,
         },
       })
```
amritk, line 309: going to tag @geoffgscott in here for the functional way to write this

amritk (commented):
hey thanks for the sweet PR! Just one code style question for @geoffgscott but other than that should be good to go!

hanspagel (changes requested):
Great contribution!

I don't think the body parameters are in `request.parameters.body`, but `request.parameters` is an array which can contain a parameter which has `{ in: 'body' }`. 🤔 

https://github.com/OAI/OpenAPI-Specification/blob/main/versions/2.0.md#fixed-fields-7

And I think it’d be cleaner to do this inside `getRequestBodyFromOperation`.

packages/api-client/src/store/workspace.ts
```diff
@@ -287,23 +287,36 @@ export const createWorkspaceStore = (router: Router, persistData = true) => {
...
+    if (bodyParams.length) {
+      requestBody = getRequestBodyFromOperation({
+        httpVerb: request.method,
+        path: request.path,
+        information: {
+          parameters: bodyParams,
+        },
+      })
+    } else if (request.requestBody) {
+      requestBody = getRequestBodyFromOperation({
```
geoffgscott, line 303: @amritk This looks like it is redoing work from `getRequestBodyFromOperation` (or mutating it's logic?)

In that function we already check for body parameters and return it first. To get body parameters it uses the more complicated method of:
```ts
const bodyParameters = getParametersFromOperation(operation, 'body', false)
```
But otherwise does the same thing. 

If we truly want this logic (and not improving `getRequestBodyFromOperation`) then I would suggest:
```ts
const requestBody =
  bodyParams.length || request.requestBody
    ? getRequestBodyFromOperation({
        httpVerb: request.method,
        path: request.path,
        information: {
          requestBody: bodyParams.length ? bodyParams : request.requestBody,
        },
      })
    : undefined
```

However I think these changes need to be in the `getRequestBodyFromOperation`

geoffgscott: All of this logic needs significant commenting as well. The intent is a bit hard to follow here.

hanspagel:
We’re now upgrading to OpenAPI 3.1 on the fly and I think that solves the issues, too, but need to double-check.

hanspagel:
It’s working partially (see #3430), so we can close this here. Thank you so much!

PR #2722, opened 2024-08-01
feat: support customizing DataUrl via ScalarOptions

captainsafia:
Addressing user feedback from https://github.com/scalar/scalar/pull/1737#issuecomment-2171906715.

Instead of exposing the `DataUrl` as a single string property, I've expsoed customization for it as a function that takes the document name as an argument.

Someone can configure the options like so:

```csharp
{
	options.DataUrlFunc = (documentName) => $"/swagger/{documentName}.json";
});
```

### DIFF
--- a/packages/scalar.aspnetcore/Scalar.AspNetCore.csproj
+++ b/packages/scalar.aspnetcore/Scalar.AspNetCore.csproj
@@ -8,7 +8,7 @@
 	</PropertyGroup>
 
 	<ItemGroup>
-		<PackageReference Include="Microsoft.AspNetCore.OpenApi" Version="8.0.5" />
+		<FrameworkReference Include="Microsoft.AspNetCore.App" />
 	</ItemGroup>
 
 </Project>
--- a/packages/scalar.aspnetcore/ScalarOptions.cs
+++ b/packages/scalar.aspnetcore/ScalarOptions.cs
@@ -8,7 +8,7 @@ public class ScalarOptions
     [JsonIgnore]
     public string EndpointPathPrefix { get; set; } = "/scalar";
 
-    [JsonIgnore] 
+    [JsonIgnore]
     public string? Title { get; set; }
 
     public string Theme { get; set; } = "purple";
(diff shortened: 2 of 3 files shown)


packages/scalar.aspnetcore/Scalar.AspNetCore.csproj
```diff
@@ -7,8 +7,4 @@
 		<Version>1.1.0</Version>
 	</PropertyGroup>
 
-	<ItemGroup>
-		<PackageReference Include="Microsoft.AspNetCore.OpenApi" Version="8.0.5" />
-	</ItemGroup>
-
-</Project>
+</Project>
```
captainsafia, line 10: Since the package doesn't directly use any of the APIs in `Microsoft.AspNetCore.OpenApi` we can remove this package reference.

Typically, we'd expect that users install the two packages alongside each other:

```
<PackageReference Include="Microsoft.AspNetCore.OpenApi" Version="9.0.0-preview.6" />
<PackageReference Include="Scalar.AspNetCore" Version="1.1.0" />
```

Then they can manage each package version independently.

marclave (approved):
amazing tysm @captainsafia !!!!

packages/scalar.aspnetcore/ScalarOptions.cs
```diff
@@ -28,6 +28,12 @@ public class ScalarOptions
     public Dictionary<string, string>? Metadata { get; set; }
 
     public ScalarAuthenticationOptions? Authentication { get; set; }
+
+    /// <summary>
+    /// Provides a function for setting the data URL for the API reference
+    /// given the document name.
+    /// </summary>
+    public Func<string, string>? DataUrlFunc { get; set; }
```
WeihanLi, line 36: What about another name like `DataUrlProvider`/`DateUrlFactory`?

captainsafia: `Provider` and `Factory` suffixes, IMO, indicate that the type you might expect to pass is more complex (e.g. some class implementation). Something like `GenerateDataUrl` might work as well.

PR #2725, opened 2024-08-01
feat(api-client): added application/problem+json media type

zsilbi:
**Problem**
Currently, `application/problem+json` responses are displayed as binary files.

**Explanation**
This happens because it is not added to the list of supported media types.

**Solution**
With this PR I added it to the list of supported media types.


marclave (approved):
awesome contribution @zsilbi ✨

PR #2727, opened 2024-08-01
fix(oas-utils): I'm a teapot 🫖

thibaultleouay:
**Problem**
I'm a teapot was missing 😭 

**Explanation**

**Solution**
With this PR we will have the best status code ever


marclave (approved):
LOVE this :)

PR #2732, opened 2024-08-01
fix(api-client): authentication UI

antlio:
this pr updates api client authentication counter logic + enhances select style as seen below:

**before**
image

**after**
image

### DIFF
--- a/.changeset/rare-pumas-help.md
+++ b/.changeset/rare-pumas-help.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
+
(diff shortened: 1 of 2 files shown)


cameronrohani:
Awesome work! Something to think about in the future could be putting selected items in pills with x buttons? I believe I've seen this pattern a handful of times before

antlio:
> Awesome work! Something to think about in the future could be putting selected items in pills with x buttons? I believe I've seen this pattern a handful of times before

great idea, will follow up with this!

marclave (approved):

PR #2733, opened 2024-08-01
fix(use-codemirror): lint

antlio:
this pr updates scalar/use-codemirror package in order to enable linting only on given content to avoid following behavior:

image


hanspagel:
Can we also not lint if the body is empty?

marclave (approved):

PR #2738, opened 2024-08-01
fix(api-client-app): allow blob: as source in Content-Security-Policy

hwkr:
Turns out our electron app is a little more locked down than a browser so we have to whitelist displaying `blob:` sources (which is what we use to display response previews).

## Before

Electron-2024-08-01-16-59-59@2x
Electron-2024-08-01-16-58-39@2x

## After

Electron-2024-08-01-16-55-42@2x
Electron-2024-08-01-16-55-22@2x

### DIFF
--- a/.changeset/fifty-walls-cheer.md
+++ b/.changeset/fifty-walls-cheer.md
@@ -1,0 +1,5 @@
+---
+'scalar-api-client': patch
+---
+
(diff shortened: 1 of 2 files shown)


marclave (approved):
:galaxy-brain:

PR #2746, opened 2024-08-02
fix(api-client): rebuild address bar history with grid

hwkr:
Fixes some of the layout issues with the address bar history and reverses the order so the most recent entry is at the top. 

There seems to be another bug where clicking a history item doesn't always mutate the address bar properly and that's still outstanding.

## Before

Google Chrome-2024-08-01-21-31-37@2x

## After

Google Chrome-2024-08-01-21-30-40@2x

### DIFF
--- a/.changeset/beige-pumas-heal.md
+++ b/.changeset/beige-pumas-heal.md
@@ -1,0 +1,6 @@
+---
+'@scalar/api-client': patch
+'@scalar/themes': patch
+---
+
+fix(api-client): rebuild address bar history with grid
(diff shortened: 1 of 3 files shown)


cameronrohani:
Awesome work😁

hanspagel:
Didn’t look at the code, but love it! One thing: On the screenshot 400 is red and 404 is grey, which can be confusing.

Can we make all 2** codes green, all 3** codes blue, all 4** errors yellow and all 5** errors red?

marclave (approved):

PR #2762, opened 2024-08-03
feat(cli): `scalar check` command to check `scalar.config.json` files

tmastrom:
This PR adds the `lint` cli command to validate `scalar.config.json` files using a JSON schema and `ajv`

JSON schema for the ScalarConfig type generated using https://github.com/vega/ts-json-schema-generator

This PR is a WIP and would love some feedback! 

### Todo
- [ ] improve file input paths 
- [ ] fetch input files from url 
- [ ] error formatting 
- [ ] more error output 
       - would be really cool to output all potential errors instead of just throwing the first one 
       - iteratively run the validator while dynamically updating the validator type to find all possible errors?


packages/cli/src/commands/lint/LintCommand.ts
```diff
@@ -0,0 +1,63 @@
...
+
+/**
+ * Lint users scalar configs (scalar.config.json files)
+ */
+export function LintCommand() {
+  const cmd = new Command('lint')
+
+  cmd.description('Lint users scalar configs')
+  cmd.argument('[file]', 'File to lint')
+  cmd.action(async (inputArgument: string) => {
```
hanspagel, line 16: I’d call the command `scalar check`, to not confuse it with validation/linting of OpenAPI documents.

Eventually the `scalar check` command could basically make sure everything is set up properly and in place, not just “lint” a file.

packages/cli/src/commands/lint/types.ts
```diff
@@ -0,0 +1,77 @@
+type ConfigSidebarPage = {
```
hanspagel, line 1: I think we should move the types, the schema and the code to check whether a given string matches the `ScalarConfig` schema to a separate `@scalar/config` package.

The CLI could then just import and execute the code, and we could eventually use the same types, validation and so on in docs.scalar.com.

Wdyt?

hanspagel:
Love it!! Added two comments regarding the naming and structure of this approach, but I think we’re on the right track. Can’t wait to have this in. 😊

packages/cli/src/commands/lint/LintCommand.ts
tmastrom: I like this update! `check` sounds better and allows future flexibility

packages/cli/src/commands/lint/types.ts
tmastrom: Yes this is more organized 👍

PR #2763, opened 2024-08-03
feat(docusaurus): add showNavLink config option

mason-at-pieces:
**Problem**
Currently, there is no way to disable the nav links generated by the docusaurus plugin.

**Solution**
With this PR users can now choose to enable/disable the nav link for any spec file they add to their docs.

Default will stay as `true` but users can now simply set `showNavLinks` to false to remove the nav link

### DIFF
--- a/.changeset/ninety-beds-bathe.md
+++ b/.changeset/ninety-beds-bathe.md
@@ -1,5 +1,5 @@
 ---
-'@scalar/docusaurus': minor
+'@scalar/docusaurus': patch
 ---
 
 Added option to remove spec link from navigation bar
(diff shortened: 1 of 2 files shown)


marclave (approved):
amazing @mason-at-pieces !!

code looks good, im about to test but could you add a changeset 

```
pnpm changeset
```

as a patch for docusaurus : )

marclave:
tested and works great :)

mason-at-pieces:
> as a patch for docusaurus : )

oh sorry, I did a minor instead of a patch. Whats the best way to fix this?

marclave:
> > as a patch for docusaurus : )
> 
> oh sorry, I did a minor instead of a patch. Whats the best way to fix this?

i just renamed it to a patch :) lets merge it!

PR #2764, opened 2024-08-03
feat(api-client): improve client request loading and feedback

hwkr:
Redid the address bar loading logic so it actually reflects whether or not the request has loaded yet, the logic itself is a little complicated but I think it makes for a nice UX.

Basically while we're waiting on a request the loading bar progresses asymptotically towards 85% (i.e. it will never actually get to 85%). Then when the request gets back we fill the rest of the bar linearly in 400ms which is the timing we were using before to fill the entire bar. 400ms feels about long enough that you can tell whats happening while still feeling snappy.

The tl;dr; is fast requests feel the same as they did before but with slow requests you can actually tell that the request hasn't loaded yet.

Other changes:
- Made the loading bar a little bit darker
- Added toasts for all Axios errors (some of the error messages suck but better than silently failing in my opinion, we can make the error handling better in the future).

https://github.com/user-attachments/assets/8183cbe5-2481-44e5-bb4f-cbac0e50abbd

### DIFF
--- a/.changeset/witty-rings-peel.md
+++ b/.changeset/witty-rings-peel.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
+
+feat(api-client): improve client request loading and feedback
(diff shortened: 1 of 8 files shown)


hanspagel (approved):
Love it!! Way better.

I’ve added dedicated error messages for empty and invalid URLs.

Screenshot 2024-08-05 at 12 47 01
Screenshot 2024-08-05 at 12 46 46

marclave (approved):

Issue #2761, opened 2024-08-02
Configurations servers.url override is not being used for the test requests (show up in other places correctly)


tnrich:
### What happens?

<img width="1079" alt="image" src="https://github.com/user-attachments/assets/2755ddc4-9a9e-4480-8eb4-674e978ded6f">

![image](https://github.com/user-attachments/assets/63b6a0ef-15eb-4211-8a21-e9147f21cb69)

![image](https://github.com/user-attachments/assets/b6a937b3-591e-4f1a-a60c-d50021bf76ed)

![image](https://github.com/user-attachments/assets/3af5a87e-9b29-4fb3-a778-aacecb2568dc)

![image](https://github.com/user-attachments/assets/5394fc1d-1c48-4bf8-9187-6458668017fc)

![image](https://github.com/user-attachments/assets/bbf8c5b7-49d7-4f48-8e8e-99a84e62d973)

### What did you expect to happen?

I would expect the servers override specified in the config to also work for the test requests! 

### How can we reproduce the issue?

I'm using this file: 

``` 
/npm/@scalar/api-reference@1.24.67/dist/browser/standalone.js
```

Do something similar to the steps outlined above using that file and you will see the issue. I wasn't able to reproduce this in the sandbox since I don't know how to set config overrides within the sandbox itself. 

### OpenAPI Document

https://sandbox.scalar.com/e/XIMVA

tnrich:
Thanks 🙏!

PR #2766, opened 2024-08-04
fix: add overload to servers as prop for create client app

marclave:
we allow for overrides of the servers in the references but we need to initialize the api client as well : )

this fixes https://github.com/scalar/scalar/issues/2761

```
        servers: [
          {
            url: 'https://api.scalar.com',
            description: 'Scalar API',
          },
        ],
```

this now works

image

### DIFF
--- a/packages/api-reference/src/components/ApiClientModal.vue
+++ b/packages/api-reference/src/components/ApiClientModal.vue
@@ -46,8 +46,6 @@ onMounted(async () => {
       const serverUrl = getUrlFromServerState(server)
       if (serverUrl) updateServerUrl(serverUrl)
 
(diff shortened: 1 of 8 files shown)


packages/api-reference/src/components/ApiClientModal.vue
```diff
@@ -44,6 +46,8 @@ onMounted(async () => {
       const serverUrl = getUrlFromServerState(server)
       if (serverUrl) updateServerUrl(serverUrl)
 
+      console.log(props.servers, 'marc', serverUrl, server)
+
```
antlio, line 50: should we ✂️ that log?

PR #2777, opened 2024-08-05
fix(api-client): authentication select

antlio:
this pr updates the api client authentication combobox in order to only display select auth and filter out unselected ones, to avoid infinite list:

image

### DIFF
--- a/packages/api-client/src/views/Request/RequestSection/RequestAuth.vue
+++ b/packages/api-client/src/views/Request/RequestSection/RequestAuth.vue
@@ -107,8 +107,14 @@ const schemeOptions = computed<SecuritySchemeOption[] | SecuritySchemeGroup[]>(
         }
       })
 
+      /** Filter out when an auth option get unchecked */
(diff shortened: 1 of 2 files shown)


packages/api-client/src/views/Request/RequestSection/RequestAuth.vue
```diff
@@ -276,7 +282,7 @@ const unselectAuth = (id: string) => {
               <RequestAuthDataTableInput
                 :id="`http-bearer-token-${scheme.uid}`"
                 :modelValue="scheme.value"
-                placeholder="QUxMIFlPVVIgQkFTRSBBUkUgQkVMT05HIFRPIFVT"
+                placeholder="Token"
```
antlio, line 285: i felt that the previous placeholder was too close from a real value making it hard to distinguish, any thougths? @amritk

marclave: it was an easter egg, but i think we can move this easter egg to something else :)

marclave (approved):
great :)

PR #2778, opened 2024-08-05
ci: split ci and release workflow (again)

hanspagel:
WIP

Sorry for all the back and forth, but I think it’s better to to do small changes on a regular basis then doing everything at once, and testing CI changes isn’t straight forward. Anyway:

Sometimes the CI workflow isn’t executed in the release PR. This seems to happen when the PR is rebased (by GitHub Actions).

This PR tries to fix this.


hanspagel:
I don’t know. This feels brittle.

PR #2781, opened 2024-08-05
fix(api-reference): markdown headings, fix #2760, fix #2004

### LINKED ISSUE #2004: Can’t scroll to 2nd-level headings
Initially, we only added the highest heading level (e.g. h1) to the sidebar. But at some point we changed it to show the two highest levels (e.g. h1 and h2) in the sidebar. 

Unfortunately, scrolling to the second level doesn’t work, because we’re only wrapping the highest level in a section.

> [Scalar sub-section.webm](https://github.com/scalar/scalar/assets/42624869/94bd24c9-133f-4444-a135-61193df48b6d)
> 
> When adding a H3 (###) to the description section, Scalar creates a sub-section in the left sidebar. However when the item menu is clicked, the page does not scroll to the selected sub-section.

_Originally posted by @Amorim33 in https://github.com/scalar/scalar/discussions/1954_

### LINKED ISSUE #2760: code blocks broken by markdown-like text
### What happens?

When previewing a spec reference, I noticed that some of the description section was broken. After some inspection, it appears that code block rendering is broken by this code block snippet:
```
    ```python
    # Example: GET /api/v2/test/resource HTTP/1.1
    # Signature Component: GET/api/v2/test/resource
    digester.write(request_method + request_uri)

    # Update the digester for further chaining
    digester = hmac.New(sha256, digester.hash())
    ```
```
The python code comment `# Example: ....` is mistakenly interpreted as markdown, which breaks the code block and begins treating the rest of the text as markdown, resulting in these artifacts:
![image](https://github.com/user-attachments/assets/c452e244-1c15-4223-b75e-700bb63549ad)
![image](https://github.com/user-attachments/assets/6489bc02-b720-4088-bbcd-fc505f3e77c6)


The code block should halt markdown format parsing until the code block is closed, preserving output for languages like python which use `#` for comments.

It should look like this:
![image](https://github.com/user-attachments/assets/c5c58201-0364-4575-ae66-a7617aaf5f40)


1. Go to https://docs.scalar.com/swagger-editor
2. Import this spec: https://raw.githubusercontent.com/SpecterOps/BloodHound/main/packages/go/openapi/doc/openapi.json
3. Notice the incorrect formatting of code blocks and out of places links in the sidebar nav

### OpenAPI Document

_No response_

hanspagel:
This PR refactors how we deal with headings in the description.

**Before**
We were splitting the content in multiple sections using regex, to then wrap the headings **and** the content below in an `IntersectionObserver`.

This broke for description with Markdown in code blocks (see #2760) and didn’t work for 2nd level headings (see #2004).

**After**
With this PR the Markdown AST is used to find headings. The content is still split into multiple sections, but only the headings (*not* the content below) are wrapped in an `IntersectionObserver`.

_Note: The sidebar doesn’t show the active state for the sidebar entries all the time anymore, because only the headings are wrapped. Unfortunately, this isn’t possible to do right (hmu if you have questions)._


marclave (approved):
awesome :)

sircodemane:
this is fantastic, thank you for the quick resolution! ❤️

PR #2784, opened 2024-08-05
docs: documented new showNavLink option in docusaurus plugin

mason-at-pieces:
**Problem**
New `showNavLink` option for the docusaurus plugin did not have any documentation besides the type file that it existed.

**Solution**
Now it is documented


hanspagel (approved):
Good call! 👏

marclave (approved):

PR #2785, opened 2024-08-06
chore(fastify-api-reference): change default routePrefix

inyourtime:
This PR updates the default `routePrefix` in the `fastify-api-reference` configuration from an empty string to `/reference`. For helps avoid potential conflicts with root route in the application.


hanspagel (approved):
Niice! 👏

marclave (approved):

Issue #2736, opened 2024-08-01
max call stack exceeded with circular references in request body


knvs97:
### What happens?

When we try to open API Client of an endpoint with request body in Specification with circular dependency on the request body definition, the API Client goes into an infinite loop and breaks with `RangeError: Maximum call stack size exceeded`. The same error occurs on loading the specification, but does not break the page but slows down the entire page.

I have observed that this is happening in below lines of code since there is no check for cyclic dependency between payload schemas
```
// requests.js
const createRequest = (payload) => deepMerge(requestSchema.parse({}), payload);

// deepmerge.js
if (isObject(source[key])) {
  if (!target[key])
    Object.assign(target, { [key]: {} });
  deepMerge(target[key], source[key]);
}```

### What did you expect to happen?

Request payloads should be allowed to have self references and the same should not break the Test Request functionality or slow down the page

### How can we reproduce the issue?

Steps to reproduce the issue:

1. Load the below specification
```---
openapi: "3.0.1"
info:
  title: "Test OpenAPI definition"
  version: "1.0.0"
servers:
- url: "http://localhost:8080"
  description: "Generated server url"
paths:
  /api/v1/updateEmployee:
    put:
      tags:
      - "update-employee-controller"
      operationId: "updateEmployee"
      requestBody:
        content:
          application/json:
            schema:
              type: "array"
              exampleSetFlag: false
              items:
                $ref: "#/components/schemas/Employee"
                exampleSetFlag: false
            exampleSetFlag: false
        required: true
      responses:
        "400":
          description: "Bad Request"
          content:
            '*/*':
              schema:
                $ref: "#/components/schemas/Response"
                exampleSetFlag: false
              exampleSetFlag: false
        "500":
          description: "Internal Server Error"
          content:
            '*/*':
              schema:
                $ref: "#/components/schemas/Response"
                exampleSetFlag: false
              exampleSetFlag: false
        "200":
          description: "OK"
          content:
            '*/*':
              schema:
                $ref: "#/components/schemas/Response"
                exampleSetFlag: false
              exampleSetFlag: false
components:
  schemas:
    AppResponse:
      type: "object"
      properties:
        message:
          type: "string"
          exampleSetFlag: false
        response:
          type: "object"
          exampleSetFlag: false
        path:
          type: "string"
          exampleSetFlag: false
      exampleSetFlag: false
    Employee:
      type: "object"
      properties:
        id:
          type: "integer"
          format: "int64"
          exampleSetFlag: false
[...]

PR #2786, opened 2024-08-06
test(api-client): createRequest with circular reference

hanspagel:
This PR has a ~~failing~~ test for #2736. It used to throw a max call stack exceeded error. Run it like this:

```
pnpm test createRequest
```

The actual issue was fixed in #2805 and #2807.


marclave (approved):
love this :)

PR #2793, opened 2024-08-06
fix(api-client): move web hash router to a factory function

### LINKED ISSUE #2791: Unexpected `/` added to hash
### What happens?

It seems like our change to using the WebRequestRouter in #2597 added a `/` to the beginning of the hash on the navigate which is causing a number of comparisons to fail.

This is making tags not expand / collapse properly and making sidebar items not collapse correctly.

There shouldn't be a `/` in the id on navigate.

Click on any of the sidebar items or tags.

### OpenAPI Document

_No response_

hwkr:
This took me way to long to debug but fixes #2791 and also the issue with sidebar items not highlighting when active.

@hanspagel I tested the app and everything seemed to be working fine still but if you want to confirm that wouldn't hurt 🙏

### DIFF
--- a/.changeset/nice-bikes-type.md
+++ b/.changeset/nice-bikes-type.md
@@ -1,0 +1,6 @@
+---
+'scalar-api-client': patch
+'@scalar/api-client': patch
+---
+
+fix(api-client): move web hash router to a factory function
--- a/packages/api-client-app/src/renderer/src/main.ts
+++ b/packages/api-client-app/src/renderer/src/main.ts
@@ -1,4 +1,4 @@
-import { webHashRouter } from '@scalar/api-client'
+import { createWebHashRouter } from '@scalar/api-client'
 import { createApiClientApp } from '@scalar/api-client/layouts/App'
 import '@scalar/api-client/style.css'
 import type { IpcRendererEvent } from 'electron'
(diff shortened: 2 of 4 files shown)


hwkr:
@hanspagel @amritk @marclave I wasn't sure if we should also switch the [normal router](https://github.com/scalar/scalar/blob/main/packages/api-client/src/router.ts#L132) to be a factory function since it seems like doing multiple calls to the router history functions may have weird side effects 🤔

amritk:
Hey so we needed the factory function in the modal due to multiple specs on one page, just curious why it was required here?

hwkr:
> Hey so we needed the factory function in the modal due to multiple specs on one page, just curious why it was required here?

See https://github.com/scalar/scalar/issues/2791#issuecomment-2272090188

Tl;dr; is calling `createWebHashHistory` has side effects

hanspagel:
Wow, what a journey!

hanspagel (approved):
Must have been crazy to hunt that bug, I’d watch if they’d have a documentary on Netflix on it. The fix works well!

hanspagel (changes requested):
Oops, ignore this comment here. :)

hanspagel (approved):

amritk (approved):
Okay I checked it out, holy what a bug. We should report it to vue-router! Also nice fix

marclave (approved):
:)

PR #2797, opened 2024-08-07
feat(api-client): move some features over to the new hotkey structure including comand palette

amritk:
- moved over to our global keydown eventbus
- simplified some logic


marclave (approved):
can merge after stray log!

Issue #2006, opened 2024-06-10
Array enum with exactly 1 enum does not render as a enum


hanspagel:
### Discussed in https://github.com/scalar/scalar/discussions/1905

<div type='discussions-op-text'>

<sup>Originally posted by **westmark** May 30, 2024</sup>
Hi! I have schema that is an array of string enums. It only contains one enum. Currently it renders as an `array[] string` with no "available options".  If I add another enum, it renders as an `array[] string enum` with all available options included. This has to be a bug, right? Swagger renders a single enum array as expected.

(Yes an array with just 1 enum is pretty weird, but it is just the start. More will be added later)</div>

soGit:
Check Elements vs Scalar

Issue #1846, opened 2024-05-24
treat single-value enums as constants


marclave:
### Discussed in https://github.com/scalar/scalar/discussions/1837

<div type='discussions-op-text'>

<sup>Originally posted by **federicobond** May 22, 2024</sup>
Since OpenAPI 3.0 does not have a way of defining constant property values, it's common practice to use single-value enums to represent such values.

The Scalar UI currently treats these the same as other enums but it would be better for readers if it was rendered in a way that made that distinction clear.

Check out this discussion for reference: https://github.com/OAI/OpenAPI-Specification/issues/1313

OpenAPI 3.1 appears to add support for `const` directly, which would also be a nice addition to Scalar too.</div>

PR #1212, opened 2024-03-17
fix: enum under items is not being displayed

CJHwong:
**Problem**
The enum definition under 'items' was omitted in the provided YAML specification. Below is a simple spec to test out:

```yaml
openapi: 3.1.0
info:
  title: Enum Test
  version: 0.0.1
paths:
  /api:
    post:
      operationId: TestEnumUnderItems
      summary: Test Enum Under Items
      requestBody:
        content:
          application/json:
            schema:
              type: object
              properties:
                field_with_enum:
                  type: array
                  items:
                    type: string
                    enum:
                      - dog
                      - cat
                      - duck
```

**Explanation**
The OpenAPI specification does not clearly state how enums should be handled under 'items'. However, based on my experience with other popular API documentation tools like Redocly and Element, I believe this is a potential oversight in how scalar renders the schema.

**Solution**
I tried to obtain the enumeration from both the property's value and 'items' if possible.

Before:
before

After:
after

PR #2803, opened 2024-08-07
fix(api-reference): improve enum support

soGit:
**Explanation**
Scalar has multiple issues & improvements related to Enum support in query parameters:
https://github.com/scalar/scalar/pull/1212
https://github.com/scalar/scalar/issues/1846
https://github.com/scalar/scalar/issues/2006

Here is a sandbox to check the the current support state. https://sandbox.scalar.com/e/8lwnj

**Solution**
With this PR following improvements was implemented:
- display enum value with a single value enum
- do not display 'enum' if it is an array of enum values
- display up to 5 enum values in a line (display flex) + >5 - (display grid)
- add 'allowed values' for array values if it is enum

Links:
https://github.com/scalar/scalar/discussions/2798

### DIFF
--- a/packages/api-reference/src/components/Content/Schema/SchemaProperty.vue
+++ b/packages/api-reference/src/components/Content/Schema/SchemaProperty.vue
@@ -65,6 +65,14 @@ const generatePropertyDescription = function (property?: Record<string, any>) {
   return descriptions[property.type][property.format || '_default']
 }
 
+const isEnum = function (value?: Record<string, any>): boolean {
+  return !!value?.enum
+}
+
+const isEnumArray = function (value?: Record<string, any>): boolean {
+  return !!value?.items?.enum
+}
+
 const getEnumFromValue = function (value?: Record<string, any>): any[] | [] {
   return value?.enum || value?.items?.enum || []
 }
@@ -83,7 +91,7 @@ const rules = ['oneOf', 'anyOf', 'allOf', 'not']
     ]">
     <SchemaPropertyHeading
       :additional="additional"
-      :enum="getEnumFromValue(value).length > 1"
+      :enum="isEnum(value)"
       :required="required"
       :value="value">
       <template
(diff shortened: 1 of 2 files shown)


packages/api-reference/src/components/Content/Schema/SchemaProperty.vue
```diff
@@ -283,6 +300,14 @@ const rules = ['oneOf', 'anyOf', 'allOf', 'not']
   border-radius: var(--scalar-radius-lg);
 }
 
+.property-enum-values {
+  display: flex;
+  gap: 6px;
+}
+.property-enum-values-long {
+  display: grid;
+  grid-template-columns: 1fr 1fr 1fr 1fr 1fr;
```
soGit, line 309: we might use https://developer.mozilla.org/en-US/docs/Web/CSS/repeat

or define is a better approach, for example. Display enum values as chips or code blocks

hanspagel:
What a great PR! Just a quick note: The first issue you mentioned is fixed already, the second seems to be not relevant.

I’m tagging @cameronrohani here, so he can check how it’s rendered and if it’s what we’re aiming for.

soGit:
> What a great PR! Just a quick note: The first issue you mentioned is fixed already, the second seems to be not relevant.
> 
> I’m tagging @cameronrohani here, so he can check how it’s rendered and if it’s what we’re aiming for.

Sure. I've linked the first issue, because in this Pull Requests presents same improvements related to it.
Thank you for pointing out on the second link, it should be https://github.com/scalar/scalar/issues/1846 (updated in the initial description as well)

cameronrohani:
Hey @soGit Marc and I will take a look in to this one this week 😄

hanspagel:
Man, we can definitely do better. But there’s so much that changed with enums lately, that I have to close this PR. I know that @cameronrohani already looked at your PR and took some inspiration from it. Thanks for investing the time to submit this PR!

PR #2804, opened 2024-08-07
feat(api-client): add loading state and cancellation for slow requests

hwkr:
Adds a loading state for slow requests (appears when the request takes more than 1 second) and a cancel button.

Also fixed a bug where you can click the request button multiple times.

https://github.com/user-attachments/assets/efa3b2f5-2100-4b66-a721-e84ddbd5c866

### DIFF
--- a/.changeset/wise-goats-develop.md
+++ b/.changeset/wise-goats-develop.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
+
+feat(api-client): add loading state and cancellation for slow requests
--- a/packages/api-client/src/components/AddressBar/AddressBar.vue
+++ b/packages/api-client/src/components/AddressBar/AddressBar.vue
@@ -77,7 +77,7 @@ function load() {
 }
 
 function startLoading() {
-  if (isRequesting.value) return
+  if (interval.value) return
   isRequesting.value = true
   interval.value = setInterval(load, 20)
 }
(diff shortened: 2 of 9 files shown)


cameronrohani:
Great work! last we talked we mentioned putting the cancel in the omnibar just curious why we went this direction?

hwkr:
> Great work! last we talked we mentioned putting the cancel in the omnibar just curious why we went this direction?

Hey! I tried it in the omnibar and it felt kinda full / less clear. Sorry I was going to send you a video, do you think it would be better in the address bar as well / instead?

hanspagel (approved):
I think it’s great to put it where the spinner is! Love this PR

geoffgscott:
Looks awesome. Nice work @hwkr

amritk (approved):
So it looks good but there is one thing you can do. Since executeRequest and cancelRequests are both related, just make a requestBus that has two events. There should be examples in one of the other busses on how to do typed multiple events. Other than that good to go, nice work!

hwkr:
> So it looks good but there is one thing you can do. Since executeRequest and cancelRequests are both related, just make a requestBus that has two events. There should be examples in one of the other busses on how to do typed multiple events. Other than that good to go, nice work!

Yeah I did it that way originally and then separated them because it seemed more clear I think? I can have a look at it again though.

amritk:
Nah that's fine if you already tried it and it didn't work, all good

cameronrohani (approved):
great work! :)

marclave (approved):

PR #2805, opened 2024-08-07
fix: Move to ts-deepmerge to better handle circular refs

geoffgscott:
The current deep merge utility does not handle the circular refs returned by the parser.

Long term we need to re-evaluate parser $ref resolution. Short term this should fix issue with circular refs. 

There may be side effects with rendering request bodies.


marclave (approved):
let's get it out :)

PR #2807, opened 2024-08-07
feat: Move to flatted for localstorage for handling of circular refs

geoffgscott:
Migrate to flatted for localStorage


amritk (approved):
Smart move! Untested but looks solid

marclave (approved):
we are so back (flatted)

PR #2811, opened 2024-08-08
fix(api-client): topnav alignment

antlio:
this pr fixes topnav misalignment as seen below:

**before**
image

**after**
image


marclave (approved):
absolutely crushed it, nice work @antlio

PR #2814, opened 2024-08-08
fix(api-client): hotkeys modifier

antlio:
after merging #2797 hotkey doesn't require the key combinations anymore (meta + letter for instance) and event are fired when pressing only the letter. this pr is an attempt to bring back the combination requirement when specified in the bus accordingly.


amritk (approved):
Okay this works, but the only thing its missing from before is ensure modifier is false, thats why it was === instead of checking for true. But I don't think that is a very big usecase so nice work!

marclave (approved):

PR #2817, opened 2024-08-08
ci(api-client-app): automatically release new app versions

hanspagel:
With this PR, a new versions of the app should be built and released in CI. The manual release process should be a thing of the past. 👀 

I’ve added the necessary secrets to the repository already.

We could also do just the build in CI and keep the actual release manual - whatever you prefer. :)


marclave (approved):
awesome work @hanspagel : )

PR #2818, opened 2024-08-08
ci: omit the build step before the type check

hanspagel:
The type check is one of the slower jobs, which isn’t really cached (yet). 

But I’m not sure whether we need a build before we run the type check. 🤔 Let’s try without that, that would be an easy way to run it earlier and speed it up.


hanspagel:
Nah, we need the build. :)

PR #2820, opened 2024-08-08
feat(types): new `@scalar/types` package, fix #1906

### LINKED ISSUE #1906: `@scalar/fastify-api-reference` requires ReferenceConfiguration type from `@scalar/api-reference`
### What happens?

When building a typescript node app without skipping library type-checking, build crashes as it cannot find `@scalar/api-reference`
image

Ability to build without setting `skipLibCheck` to true

1. Create a fastify project with typescript
2. set `skipLibCheck: false`
3. Use `@scalar/fastify-api-reference` 
4. Try to build

### OpenAPI Document

_No response_

amritk:
At first I tried to go NO dependency, but that was a lot of extra copy and paste. Instead I went LOW dependency and we just have 3 small ones. This is a nice happy medium I think, can go full copy and paste if needed.

closes #1906 
references #2649

here are the current dependencies:
```bash
@scalar/openapi-types link:../openapi-types
@scalar/themes link:../themes
@unhead/schema 1.9.15
├── hookable 5.5.3
└── zhead 2.2.4
```


packages/api-reference-types/package.json
```diff
@@ -0,0 +1,39 @@
...
+  "name": "@scalar/api-reference-types",
+  "description": "Shared types for the references",
+  "license": "MIT",
+  "author": "Scalar (https://github.com/scalar)",
+  "homepage": "https://github.com/scalar/scalar",
+  "bugs": "https://github.com/scalar/scalar/issues/new/choose",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/scalar/scalar.git",
+    "directory": "packages/apoi"
```
marclave, line 11: is this correct? :)

amritk: Lawl

packages/types/src/legacy/reference-config.ts
```diff
@@ -1,11 +1,160 @@
...
-import type { RequestMethod } from '@/helpers'
-import type { OpenAPIV2, OpenAPIV3, OpenAPIV3_1 } from '@scalar/openapi-parser'
-import type { HarRequest } from 'httpsnippet-lite'
+import type { ThemeId } from '@scalar/themes'
+import type { UseSeoMetaInput } from '@unhead/schema'
+import type { HarRequest, TargetId } from 'httpsnippet-lite'
+import type { OpenAPIV2, OpenAPIV3, OpenAPIV3_1 } from 'openapi-types'
 
+/**
+ * @deprecated This type does not provide much type safety, we shouldn't really use it
```
hanspagel, line 7: But this is exactly what it’s for: For objects where we don’t know anything about the shape. 🤔

hanspagel:
OMG. great work! The `openapi-types` have a few issues for our use cases. It basically assumes we’re dealing with an OpenAPI document which strictly adheres to the standard, but that’s not the case. We’re dealing with any kind of user input. :)

The types we’re using in the parser seem to be better fit and they allow us to remove the `openapi-types` packages from most packages.

I think I’d use the same approach for the parser and move those types to `@scalar/openapi-types`, then we’d be able to use that as a very light dependency in `@scalar/types`. 👍

packages/oas-utils/package.json
```diff
@@ -34,10 +34,6 @@
   "type": "module",
   "main": "dist/index.js",
   "exports": {
-    ".": {
-      "import": "./dist/index.js",
-      "types": "./dist/index.d.ts"
-    },
```
hanspagel, line 40: That seems dangerous, do we really want to remove this? 🤔

packages/types/package.json
```diff
@@ -0,0 +1,39 @@
...
+  "type": "module",
+  "types": "dist/index.d.ts",
+  "files": [
+    "dist",
+    "CHANGELOG.md"
+  ],
+  "dependencies": {
+    "@scalar/themes": "workspace:*",
+    "@unhead/schema": "^1.9.5",
+    "httpsnippet-lite": "^3.0.5"
```
hanspagel, line 37: I agree, let’s copy the types from `httpsnippet-lite` and get rid of the dependency. :)

hanspagel (changes requested):

packages/oas-utils/package.json
hanspagel: Ah, this used to only export types so we don’t need it anymore.

packages/types/src/legacy/reference-config.ts
geoffgscott: Probably `Record<string, unknown>` would be a better choice

packages/types/package.json
```diff
@@ -0,0 +1,39 @@
...
+    "lint:fix": "eslint .  --fix"
+  },
+  "type": "module",
+  "types": "dist/index.d.ts",
+  "files": [
+    "dist",
+    "CHANGELOG.md"
+  ],
+  "dependencies": {
+    "@scalar/openapi-parser": "^0.7.2",
```
geoffgscott, line 34: I think this eliminates the whole point of the package? :joy:

packages/types/src/index.ts
geoffgscott, line 1: I don't really believe in hoisting all the exports to the top level of the package. They should be logically grouped by domain within the package.

We can just import from `@scalar/types/legacy`

This manages the import namespace a lot more effectively (especially with legacy types where we might want to use the same name later).

packages/types/package.json
hanspagel: Not really! The @scalar/openapi-parser dependency tree is way smaller, also I’ve already submitted a PR to separate the types into a dependency free `@scalar/openapi-types`, so we can use that very soon.

amritk: Wait a sec how did this get here lol

packages/types/src/index.ts
amritk: Haha sure sure, tomato potato 😜

packages/types/package.json
amritk: Oh looks like hans made some changes. I was like what haha

amritk: @hanspagel these are the dependencies from the openapi parser
```bash
ajv 8.17.1
├── fast-deep-equal 3.1.3
├── fast-uri 3.0.1
├── json-schema-traverse 1.0.0
└── require-from-string 2.0.2
ajv-draft-04 1.0.0
└─┬ ajv 8.17.1 peer
  ├── fast-deep-equal 3.1.3
  ├── fast-uri 3.0.1
  ├── json-schema-traverse 1.0.0
  └── require-from-string 2.0.2
ajv-formats 3.0.1
└─┬ ajv 8.17.1 peer
  ├── fast-deep-equal 3.1.3
  ├── fast-uri 3.0.1
  ├── json-schema-traverse 1.0.0
  └── require-from-string 2.0.2
jsonpointer 5.0.1
leven 4.0.0
yaml 2.4.5
```
ajv, ajv-draft-04, ajv-draft-04 and jsonpointer  all cause trouble due to them being commonjs
yaml causes problems in other ways

We need that new types package

marclave (approved):
let's get it in!!!!

packages/types/src/legacy/reference-config.ts
```diff
@@ -1,11 +1,165 @@
-import type { ExternalDocumentation } from '@/entities/workspace/collection'
-import type { RequestMethod } from '@/helpers'
-import type { OpenAPIV2, OpenAPIV3, OpenAPIV3_1 } from '@scalar/openapi-parser'
-import type { HarRequest } from 'httpsnippet-lite'
+import type { HarRequest, TargetId } from '#external'
```
geoffgscott, line 1: No external imports please

geoffgscott (changes requested):
No `#imports` please. It's just not a pattern we have much of a use case for or use anywhere else.

geoffgscott (approved):
LGTM

PR #2823, opened 2024-08-09
fix(api-client): sidebar dragging

antlio:
this pr introduces the enableResizer prop for the sidebar component in order to only enable dragging / resizing on request view, and not on others like cookies, environments.

### DIFF
--- a/.changeset/mean-balloons-guess.md
+++ b/.changeset/mean-balloons-guess.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
+
+fix: enable sidebar resizer on request page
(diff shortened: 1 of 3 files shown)


marclave (commented):
vscode seems to keep the width different panes, wdyt about keeping it?

image
image

image

cameronrohani:
@marclave I was thinking we should do this too :galaxy-brain:

antlio:
because once shrinked its not openable anymore, but we could set a min width instead then on non request page?

cameronrohani:
Yeah could work with min width I suppose! Let's do it? 😄

antlio:
i ended up setting globally the sidebar width, while removing the drag and hide capability as it didn't feel seamless for now.

marclave (approved):
dig it :)

PR #2825, opened 2024-08-09
feat(api-reference): `RequestExample` embed

hanspagel:
This PR adds a standalone `RequestExample` Vue component.

This and all future components will need to dereference the OpenAPI document. We probably don’t want to do this in every embed, so I’ve also added an `OpenApiDocument` which does the data wrangling and can then be used by one or multiple instances of the `ExampleRequest` component.

In order to bring this to life, the components need to be really encapsulated, so I’ve had to move some code. I think that’s a great way to improve our code quality. :)

Supports the relevant parts of the configuration: proxy, spec, hiddenClients, defaultHttpClient, baseServerUrl, servers, authentication.

**Usage**

_Note: While this PR adds the embeds, it doesn’t expose them yet. Let’s polish them a little bite more before making them available to everyone. :)_

```vue
<script setup>
// TODO: The package doesn’t have this entrypoint yet.
import { OpenApiDocument, ExampleRequest } from '@scalar/api-reference/embeds'
</script>

<template>
  <OpenApiDocument
    v-slot="{ configuration, parsedSpec }"
    :configuration="{
      proxy: 'https://proxy.scalar.com',
      spec: {
      }
    }">
  
    <ExampleRequest
      :configuration="configuration"
      :operation="parsedSpec.tags?.[1]?.operations?.[1]"
      :parsedSpec="parsedSpec" />
  
  </OpenApiDocument>
</template>
```

**Playground**

```bash
cd packages/api-reference
pnpm playground:embed
```

**Preview**

Screenshot 2024-08-09 at 21 23 02


geoffgscott:
@hanspagel I think the approach around OpenApiDocument needs to be discussed a bit more. Looking at this the plan looks to be using it in third party consumers to provision specs for different types of embedded components. 

I think we can do this better with a hook that exposes both the Vue reactive objects and an event based system that would be easier to interface with in other applications.

```ts
// Usage anywhere in the app. 
const { configuration, parsedSpec, addSpec  } = useOASDocument({
  key: 'some-spec-id',
  /** Event based interface for non-vue frameworks. */ 
  onSpecChange: (spec) => console.log(spec),
})

// Function used to add a spec to the app using a namespaced key
// This allows a spec to be uploaded/parsed once and referenced everywhere using the same key
addSpec({
  key: 'some-spec-id',
    ...baseConfig,
})
```

With your above point about tidying up the code by keeping code encapsulated I think this could help a lot with writing broader parts of the reference codebase as we can just provide a key to components and they can call useOASDocument() internally to get the data they need.

hanspagel:
True. It was just slightly easier/faster to move the code from a component to another component and I wanted it to be hidden, but merged before I’m OOO. Anyway, I’ll rebase and add to the PR. ✌️

marclave (approved):

hanspagel:
Cool! I’ll merge this and take over the responsibility to further improve the implementation according to @geoffgscott’s feedback. 🤝

PR #2831, opened 2024-08-10
feat(api-client): added multi auth

### LINKED ISSUE #2744: complex security requirements render incorrectly
### What happens?

Top level `security` requirements are rendered incorrectly when using complex security methods (such as HMAC signed requests). Something like a bearer token scheme is fine as it's a single `securityScheme` component, but for signed requests it could require multiple combined `securityScheme` components:
```json
  "security": [
    {
      "JWTBearerToken": []
    },
    {
      "SignedRequest": [],
      "RequestDate": [],
      "HMACSignature": []
    }
  ],
```
In several places where these security requirements are used, they are treated as individual auth schemes, rather than combined:
image
Example code snippets also generate code using only one of the pieces as well. 

These complex security requirements should be rendered as a single combined component, and reflected similarly in code snippets. The Test Request client is less affected as it allows you to select the check boxes for each of securitySchemes, but ideally it would also handle complex security requirements more appropriately.

This is something I may be willing to open a PR for, but it would take some time to familiarize myself with the code base and design an appropriate solution.

1. Go to https://docs.scalar.com/swagger-editor
2. Import this spec: https://raw.githubusercontent.com/SpecterOps/BloodHound/main/packages/go/openapi/doc/openapi.json
3. Notice the behavior of the security selector in the top right, endpoint code snippet examples, and test request client

For comparison, this is how it is handled in other similar tooling:
image
image
image

### OpenAPI Document

_No response_

amritk:
closes #2744

Not the best code, but jammed multi id into a string with delimiter so the combobox could handle it


marclave (approved):
wow amazing work @amritk !!

PR #2833, opened 2024-08-10
feat: add polish to models

marclave:
aligning the models with collapsable states for a better consumption of information density


antlio:
nice! we should set that up with the sidebar and then good to go!

PR #2834, opened 2024-08-10
feat(fastify-api-reference): add hooks option for API reference route

inyourtime:
This PR adds a `hooks` option to the API reference route, allowing you to include hooks such as basic authentication to protect your API reference route or add other custom functionality as needed.

example with basic authentication:

https://github.com/user-attachments/assets/aaaebe7e-a814-4e31-9dd7-7bd0ea6d0d2f


marclave (approved):
okay this is actually so rad, nice work @inyourtime !!

PR #2838, opened 2024-08-10
fix: make api ref use variable border width + change the default value

cameronrohani:
scalar api ref wasn't using --scalar-border-width, so I fixed that + made the default value to 0.5px to match the api client app


marclave (approved):
:)

PR #2843, opened 2024-08-12
fix(oas-utils): request default tag

antlio:
this pr fixes the `default` tag issue that is now added to every request instead of just request without tag, which was duplicating them in the api client and creating an unneeded `default` collection:

image


amritk (approved):
Seems reasonable!

marclave (approved):

PR #2848, opened 2024-08-12
fix: make tag background have more contrast when unopened

cameronrohani:
For consideration - make tag background have more contrast when unopened?

image


marclave:
O M G

i love love love this

marclave (approved):

PR #2851, opened 2024-08-12
fix: README.md

timheuer:
Updating .NET readme usage snippet

- Updating `app.MapOpenAPi()` to match how .NET 9 default templates use it (behind IsDevelopment guard).
- Updating usage text to indicate the default URL of the usage snippet would be


marclave (approved):
nice PR @timheuer ! thanks for updating the docs ✨

marclave:
@timheuer can you run `pnpm format` in root? : )

timheuer:
> @timheuer can you run `pnpm format` in root? : )

done/pr updated

PR #2853, opened 2024-08-13
feat: deploy void server playground to cloud run

tmastrom:
**Problem**
Currently, we are deploying the void server to cloudflare from a separate repo

**Solution**
With this PR we create `@scalar-examples/void-server` and deploy it to cloud run using cloud build.


geoffgscott (changes requested):
Needs to deploy from playground not a separate example package.

PR #2854, opened 2024-08-13
chore: remove cloudbuild step to deploy deleted hono example

tmastrom:
**Problem**
Currently, we have a cloud build step to deploy an example that no longer exists.


geoffgscott (changes requested):
@tmastrom The example has moved to the playground folder. Can we just update to deploy from there?

geoffgscott:
@tmastrom This is already pointing the correct Dockerfile. 

`'./packages/hono-api-reference/Dockerfile',`

PR #2857, opened 2024-08-13
feat: deploy void-server to cloud run

tmastrom:
With this PR, the `void-server` playground example is deployed to cloud run.

### DIFF
--- a/packages/void-server/Dockerfile
+++ b/packages/void-server/Dockerfile
@@ -4,30 +4,15 @@ WORKDIR /app
 
 ENV BUILD_PLAYGROUND=1
 
-# process init system
-RUN apt-get update && apt-get install -y --no-install-recommends dumb-init
-RUN npm install pnpm@9.2.0 --global
-RUN pnpm config set store-dir ~/.pnpm-store
+# Build the package with the BUILD_PLAYGROUND flag set
+RUN pnpm --filter @scalar/void-server build
 
-# Copy and build the void server package
-COPY pnpm-lock.yaml .
-COPY pnpm-workspace.yaml .
-COPY package.json .
-COPY tsconfig.json .
-COPY turbo.json .
-COPY packages/build-tooling ./packages/build-tooling
-COPY packages/void-server ./packages/void-server
-
-RUN pnpm --filter @scalar/void-server install --frozen-lockfile && \
-    pnpm --filter @scalar/void-server build
-
-
-FROM node:18-bullseye-slim AS runner
-# Copy from the previous stage to keep this image minimal
+FROM node:20-bullseye-slim AS runner
 COPY --from=builder /usr/bin/dumb-init /usr/bin/dumb-init
 
 ENV NODE_ENV=production
 ENV PORT=$PORT
+ENV HOST='0.0.0.0'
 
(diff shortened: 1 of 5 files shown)


packages/void-server/Dockerfile
```diff
@@ -0,0 +1,43 @@
...
+RUN pnpm config set store-dir ~/.pnpm-store
+
+# Copy and build the void server package
+COPY pnpm-lock.yaml .
+COPY pnpm-workspace.yaml .
+COPY package.json .
+COPY tsconfig.json .
+COPY turbo.json .
+COPY packages/build-tooling ./packages/build-tooling
+COPY packages/void-server ./packages/void-server
```
geoffgscott, line 19: base builder will include all of these file

packages/void-server/Dockerfile
```diff
@@ -0,0 +1,43 @@
+ARG BASE_IMAGE
+FROM ${BASE_IMAGE} AS builder
+WORKDIR /app
+
+ENV BUILD_PLAYGROUND=1
+
+# process init system
+RUN apt-get update && apt-get install -y --no-install-recommends dumb-init
+RUN npm install pnpm@9.2.0 --global
+RUN pnpm config set store-dir ~/.pnpm-store
```
geoffgscott, line 10: May as well add this to base builder

packages/void-server/Dockerfile
```diff
@@ -0,0 +1,43 @@
...
+COPY pnpm-lock.yaml .
+COPY pnpm-workspace.yaml .
+COPY package.json .
+COPY tsconfig.json .
+COPY turbo.json .
+COPY packages/build-tooling ./packages/build-tooling
+COPY packages/void-server ./packages/void-server
+
+RUN pnpm --filter @scalar/void-server install --frozen-lockfile && \
+    pnpm --filter @scalar/void-server build
```
geoffgscott, line 22: This is good. I think best practice is probably to let base builder build without the flag and then do a second build of the playgrounds here.

packages/void-server/Dockerfile
```diff
@@ -0,0 +1,43 @@
...
+
+
+FROM node:18-bullseye-slim AS runner
```
geoffgscott, line 25: May as well move to 20

packages/void-server/Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+
+# Build the package with the BUILD_PLAYGROUND flag set
+RUN pnpm --filter @scalar/void-server build
+
+FROM node:20-bullseye-slim AS runner
+COPY --from=builder /usr/bin/dumb-init /usr/bin/dumb-init
+
+ENV NODE_ENV=production
+ENV PORT=$PORT
+ENV HOST='0.0.0.0'
```
geoffgscott, line 15: @tmastrom What are we using these for? $PORT is already exposed and I don't see us having a need to change the host?

geoffgscott (approved):
Looks good.

packages/void-server/Dockerfile
tmastrom: These aren't necessary

PR #2861, opened 2024-08-13
fix(components): rebuild code block x-y scroll box

hwkr:
Looks like when we switched from codemirror to highlight JS we lost a bunch of our custom scroll classes that were targeting the code block. This PR should put things back to the way they were.

## Before / After

https://github.com/user-attachments/assets/779aff99-72df-4910-8f3b-80e25f9ecc92


cameronrohani (approved):
amazing!

marclave (approved):

PR #2862, opened 2024-08-13
fix(themes): always show obtrusive scrollbars

hwkr:
Not sure if this is exactly what we want to do but it's probably the best behavior in terms of accessibility and respecting user preferences.

This PR updates scrollbars to be similar to Linear's where if obtrusive scrollbars are turned on (e.g. configured in MacOS or on Windows etc.) then we show the scroll bars. If we only show the obtrusive scrollbars on hover (the previous behavior) then we get weird padding gutters that clip the content of the scroll box.

**Note:** You have to have obtrusive scrollbars turned on to see this.

See https://linear.app/api-docs/issue/DOC-2209/show-obtrusive-scroll-bars

## Before

Google Chrome-2024-08-13-17-00-25@2x

## After

Google Chrome-2024-08-13-17-00-05@2x


cameronrohani (approved):

geoffgscott (approved):

marclave (approved):
@hwkr nice work, lets get this in after conflicts are fixed ❇️

PR #2865, opened 2024-08-14
fix(api-client): close tab hotkey event prevent

antlio:
this pr updates the app native shortcut to be able to use `meta + w` to close tab as expected:

image


amritk (changes requested):
As per our slack discussion, we should prevent default where the listener is. So wherever we handle the closeTAb event, prevent default there. 

EDIT: nvm lets handle this inside electron

marclave (approved):

PR #2867, opened 2024-08-14
fix(api-client): command palette navigation

antlio:
this pr fixes the navigation issue cause by the focused search input in the command palette modal:

https://github.com/user-attachments/assets/8369c324-e4fe-420f-8b1c-eb23b639a74e

### DIFF
--- a/packages/api-client/src/views/Request/Request.vue
+++ b/packages/api-client/src/views/Request/Request.vue
@@ -143,7 +143,12 @@ const executeRequest = async () => {
 /** Handle hotkey events from the bus */
 const handleHotKey = (event: HotKeyEvents) => {
   if (event.toggleSidebar) showSideBar.value = !showSideBar.value
-  if (event.openCommandPalette) commandPaletteBus.emit()
+
+  // We prevent default on open command so we can use it on the web
+  if (event.openCommandPalette) {
+    event.openCommandPalette.preventDefault()
+    commandPaletteBus.emit()
+  }
 }
 
 onMounted(() => {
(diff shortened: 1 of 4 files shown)


amritk (commented):
Two things to check here. 

How does it work on textareas, will it just blur instead of going down a line? For this we can just ensure its an input then handle textareas differently. OR currently we have single hotkeys disabled when in inputs. Maybe we can decide which keys to allow and it should just work I think. Just message me about this one!

Can we modify the code in the commandpallet so it selects through the input as well? It should be in the `handleArrowKey` function. This can be done in another ticket/pr so its not blocking but the first point is.

amritk:
Actually I got an idea, let me do something here

amritk:
Alright I made some changes, now the input stays focussed and still allows you to both navigate as well as type

[screen-recorder-wed-aug-14-2024-11-19-32.webm](https://github.com/user-attachments/assets/6840e842-a78a-459b-9017-2e11a19a808e)

packages/api-client/src/views/Request/Request.vue
```diff
@@ -143,7 +143,12 @@ const executeRequest = async () => {
 /** Handle hotkey events from the bus */
 const handleHotKey = (event: HotKeyEvents) => {
   if (event.toggleSidebar) showSideBar.value = !showSideBar.value
-  if (event.openCommandPalette) commandPaletteBus.emit()
+
+  // We prevent default on open command so we can use it on the web
```
amritk, line 147: was using this for testing, but seems sensible its a common practice to hijack this command for search

marclave (approved):
niiiice 😈

PR #2868, opened 2024-08-14
feat(galaxy): migrate galaxy to monorepo

tmastrom:
This PR 
* updates the `@scalar/galaxy` to build with our rollup tooling. 
* adds a playground example in `@scalar/galaxy` to serve the galaxy spec with the `@scalar/mock-server`
*  creates a Dockerfile and cloudbuild script to deploy the the `@scalar/mock-server` example (the mock-server and galaxy playground examples are the same so we can deploy either one)


packages/galaxy/rollup.config.ts
```diff
@@ -0,0 +1,23 @@
...
+      dest: 'dist',
+      rename: 'latest.yaml',
+    },
+  ],
+  options: {
+    input: entries,
+  },
+})
+
+await addPackageFileExports({ entries })
```
tmastrom, line 23: I am having issues with this overwriting the the .json and .yaml file exports in the package.json

packages/mock-server/Dockerfile
```diff
@@ -0,0 +1,51 @@
...
+COPY turbo.json .
+COPY packages ./packages
+RUN pnpm install
+RUN pnpm turbo \
+    run build \
+    --filter='./packages/*' \
+    # … don’t build too many in parallel, we don’t want to hit a memory limit.
+    --concurrency=2
+
+FROM base AS builder
```
geoffgscott, line 27: We can just use base builder here.

packages/galaxy/rollup.config.ts
geoffgscott: That is precisely what it does. In your case I would suggest just manually provision the exports field.

geoffgscott (approved):
Looks good.

marclave (changes requested):
we should add a changeset to this so we know the deploys to npm are working

https://www.npmjs.com/package/@scalar/galaxy

PR #2882, opened 2024-08-16
fix(components): refactor loading to match icon sizes

hwkr:
Switches the `<ScalarLoading>` component to use the same sizes as our `<ScalarIcon>` component


marclave (approved):
:)

PR #2888, opened 2024-08-16
feat(api-client): top nav context menu

antlio:
this pr adds context menu on top nav as seen below:

https://github.com/user-attachments/assets/ad4c62fb-a2b8-4255-a7e9-124af56915f4


marclave (approved):
LOVE this ✨

PR #2894, opened 2024-08-16
fix(api-client): force code mirror to scroll on overflow

hwkr:
## Before

Google Chrome-2024-08-16-13-44-02@2x

## After

Google Chrome-2024-08-16-13-43-42@2x


marclave (approved):
we are so back ✨

PR #2899, opened 2024-08-16
chore(api-reference): wrapper to get examples from `@scalar/snippetz` and `httpsnippet-lite`

tmastrom:
Migrate `@scalar/snippetz` to the monorepo

This PR is just the migration and some exploration to create a wrapper for the `httpsnippet-lite` methods we currently use. Unifying this API specifically for usage in `api-reference` will follow.


hanspagel (changes requested):
Great work! Just at a glance: The `scalar.config.json` was deleted for some reason and the recent changesets are missing. I would have prefered to do a release in the old repository before the migration, maybe we’ll do that and update the PR here. I’ll check what’s the best way to get this in asap.

tmastrom:
> Great work! Just at a glance: The `scalar.config.json` was deleted for some reason and the recent changesets are missing. I would have prefered to do a release in the old repository before the migration, maybe we’ll do that and update the PR here. I’ll check what’s the best way to get this in asap. 🙌
> 
> Just personally, I don’t like the `Snippet` prefix in the type names, and would go with an import alias `TargetId as SnippetzTargetId` if the prefix is needed in our use case.

Added the `scalar.config.json` back! Not sure why that was deleted. 

Can you explain more about the changesets? Do we want to merge the changesets currently in this repo with the changeset in the old snippetz repo? 

Would you prefer `TargetId` and `ScalarTargetId` to differentiate the types?

hanspagel:
I think I’d 

1. release @scalar/snippetz in the old repository first (because they were major changes that aren’t released yet)
2. submit a new PR here to update @scalar/snippetz,
3. then pick up this PR here again to migrate the code and
4. submit a separate PR for the new unified httpsnippet-lite/snippetz API

I plan to do this over the next days. :v:

hanspagel:
Closed in favor of #2995 🫡

PR #2911, opened 2024-08-20
fix(api-client): request sidebar item new tab event

antlio:
this pr use export and use getModifiers function to be able to rely on the meta key no matter the OS.


amritk (approved):
LGTM, don't forget to add a changeset!

PR #2913, opened 2024-08-20
feat(api-client): command palette input required

antlio:
this pr prevents command palette data submission if input value is empty, set error state toast accordingly and comments out the server link which will be reactivated once the server view is set:

image

### DIFF
--- a/.changeset/selfish-paws-jog.md
+++ b/.changeset/selfish-paws-jog.md
@@ -1,0 +1,5 @@
+---
+'@scalar/components': patch
+---
+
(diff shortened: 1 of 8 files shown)


cameronrohani:
Let's try to use existing colors rather than applying an opacity?

Perhaps bg-3 and color-3

antlio:
> Let's try to use existing colors rather than applying an opacity?
> 
> Perhaps bg-3 and color-3

updated / fixed to rely on disabled styles

marclave (approved):

PR #2914, opened 2024-08-20
feat(nuxt): add dynamic styles import, update README

romanhrynevych:
In this PR, I made these changes:
1. Update Nuxt default config, as in Docs mentioned default route `/scalar` [commit](https://github.com/scalar/scalar/commit/71ff33b05572e78138be88fa54fb9cb845a72dc8)
image

2. Added `theme` config because I want to use the default theme, not only `nuxt` once; I think this will be better if the user could select from predefined ones or an exclusive `nuxt` theme, which is the default; also updated README.md file

https://github.com/user-attachments/assets/01c4f182-25d7-44a6-9ba8-1a010813833e

### DIFF
--- a/.changeset/long-carrots-own.md
+++ b/.changeset/long-carrots-own.md
@@ -1,5 +1,5 @@
 ---
-'@scalar/nuxt': minor
+'@scalar/nuxt': patch
 ---
 
 Added dynamic theme configuration, update README
--- a/.changeset/long-carrots-own.md
+++ b/.changeset/long-carrots-own.md
@@ -1,0 +1,5 @@
+---
+'@scalar/nuxt': minor
+---
+
+Added dynamic theme configuration, update README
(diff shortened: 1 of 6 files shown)


marclave (approved):
what an absolutely amazing PR @romanhrynevych !

ah we should keep the route as /docs :) so if you could change that + the docs we can get this merged :)

romanhrynevych:
> what an absolutely amazing PR @romanhrynevych !
> 
> ah we should keep the route as /docs :) so if you could change that + the docs we can get this merged :)

Thanks! First I think that `/scalar` as in README are right 😅. Changed in [8d84213](https://github.com/scalar/scalar/pull/2914/commits/8d84213455ca29a0b8a37ba9162b204ba8e5fd61)

marclave (approved):

amritk:
hey @romanhrynevych unfortunately I might have to roll back some of the theme changes, they don't seem to be working well with ssr at the moment.
[nuxty.webm](https://github.com/user-attachments/assets/55fb358b-fdef-4eeb-b259-39f84f89b45e)

I believe the solution would be to host the theme css files then link to them through useHead. That should work

PR #2915, opened 2024-08-21
feat(ts-to-openapi): new TypeScript to OpenAPI library

amritk:
It's still early but I have added a bunch of tests. Next step would be to add some automated documentation which would be nice. This package contains a library required to generate openapi specs from typescript.


marclave (approved):
played around with it this weekend 🤯 🤯 🤯 🤯 🤯 🤯 🤯 🤯 🤯 🤯

PR #2935, opened 2024-08-23
feat(api-client): main topnav context menu

antlio:
this pr adds context menu also on the unique topnav:

image


marclave (approved):
niiiice @antlio ✨

PR #2943, opened 2024-08-26
fix(api-client): variable removal breaking line

antlio:
this pr fixes the line addition when removing variable in the api client + some style (pill centering) as seen below:

https://github.com/user-attachments/assets/f610b57a-0267-44dc-a34d-60405149d0ac

### DIFF
--- a/.changeset/eight-beds-peel.md
+++ b/.changeset/eight-beds-peel.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
+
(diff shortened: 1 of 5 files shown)


amritk (approved):
We should also allow you to hit the down arrow after typing opening braces `{{`

antlio:
> We should also allow you to hit the down arrow after typing opening braces `{{`

in progress in another branch 😏

marclave (approved):
i dig this :)

PR #2945, opened 2024-08-26
feat(api-reference): multipart and url encoded request data, fix #2902, fix #1451

### LINKED ISSUE #1451: Example requests with `application/x-www-form-urlencoded`
### What happens?

In the OpenAPI document, there is a requestBody definition for application/x-www-form-urlencoded type. However, when creating a curl request adhering to this definition, it has been observed that the data appears to be empty when sent via the --data parameter. This issue seems to stem from the expected data format not being received properly, despite creating the curl request in the correct format.

Could you please assist in resolving this issue?

data

Reviewed the requestBody definition in the OpenAPI document.
Created the curl request correctly, but observed that the data is empty.
If possible, could you explain the reason for this issue and provide a solution?

1)Reviewed the requestBody definition in the OpenAPI document.
2)Created the curl request correctly, but observed that the data is empty.

### Swagger/OpenAPI Example

```
post:
  tags:
    - "Form"
  summary: Add new question
  description: Add new question to specified form.
  operationId: formQuestions
  requestBody:
    content:
        schema:
          $ref: '#/components/schemas/Questions'
```

### LINKED ISSUE #2902: Error parsing the request body if the content type is ‘application/x-www-form-urlencoded’
### What happens?

When I set the content type of requestBody to application/x-www-form-urlencoded, it cannot be rendered correctly. For example, the schema of requestBody is as follows:
```yaml
"/oauth/token":
    post:
      tags:
        - OAuth
      summary: Get Token
      security: []
      requestBody:
        description: Get Token
        required: true
        content:
            schema:
              type: object
              properties:
                client_id:
                  type: string
                  example: 123345xxxx
```
The rendered output is as follows:
image

The example should be displayed.

Copy the above code into a yaml file as a path. 

### OpenAPI Document

_No response_

hanspagel:
Long overdue! This PR adds support for request bodies with the `application/x-www-form-urlencode` and `multipart/form-data` content types. :)

See #2902 for more context.

### DIFF
--- a/packages/oas-utils/src/spec-getters/getRequestBodyFromOperation.ts
+++ b/packages/oas-utils/src/spec-getters/getRequestBodyFromOperation.ts
@@ -11,7 +11,7 @@ import { getParametersFromOperation } from './getParametersFromOperation'
  * that represent the key-value pairs of the object.
  */
 function getParamsFromObject(
-  obj: any,
+  obj: AnyObject,
   prefix = '',
 ): {
   name: string
--- a/packages/oas-utils/src/spec-getters/getRequestBodyFromOperation.ts
+++ b/packages/oas-utils/src/spec-getters/getRequestBodyFromOperation.ts
@@ -7,6 +7,28 @@ import { getExampleFromSchema } from './getExampleFromSchema'
 import { getParametersFromOperation } from './getParametersFromOperation'
 
 /**
+ * Transform the object into a nested array of objects
+ * that represent the key-value pairs of the object.
+ */
+function getParamsFromObject(
+  obj: any,
+  prefix = '',
+): {
+  name: string
+  value: any
+}[] {
+  return Object.entries(obj).flatMap(([key, value]) => {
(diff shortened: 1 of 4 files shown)


packages/oas-utils/src/spec-getters/getRequestBodyFromOperation.ts
```diff
@@ -6,6 +6,28 @@ import { prettyPrintJson } from '../helpers/prettyPrintJson'
 import { getExampleFromSchema } from './getExampleFromSchema'
 import { getParametersFromOperation } from './getParametersFromOperation'
 
+/**
+ * Transform the object into a nested array of objects
+ * that represent the key-value pairs of the object.
+ */
+function getParamsFromObject(
+  obj: any,
```
amritk, line 14:
```suggestion
  obj: AnyObject,
```
😉

amritk (approved):
Looks good

marclave (approved):
approved but should add @amritk 's comment ✨

PR #2946, opened 2024-08-26
fix(api-reference): incorrect example for schemas with anyOf, fix #2218

### LINKED ISSUE #2218: Schema with anyOf section that contains two objects is rendered invalid
### What happens?

I have tried this schema(https://sandbox.scalar.com/e/1T1zN) for a path response. It has an `anyOf` section which consists of two objects and a normal property.

```json
{
  "title": "Foo",
  "type": "object",
  "anyOf": [
    {
      "type": "object",
      "required": [
        "a"
      ],
      "properties": {
        "a": {
          "type": "integer",
          "format": "int32"
        }
      }
    },
    {
      "type": "object",
      "required": [
        "b"
      ],
      "properties": {
        "b": {
          "type": "string"
        }
      }
    }
  ],
  "required": [
    "c"
  ],
  "properties": {
    "c": {
      "type": "boolean"
    }
  }
}

```

The rendered model has only field `c` and don't show that there should be either field `a` or `b`. In request example it also shows extra `mode` and `omitEmptyAndOptionalProperties` fields for some reason, but does provide `a` field.
image
If `anyOf` section is removed then `mode` and `omitEmptyAndOptionalProperties` fields disappear.

I stumbled upon this when generating docs from rust structures(I don't know much about json schemas and openapi spec) using `serde` and `schemars`. This may be common when you want to deserialize into a struct that can have one of two fields.
```rust
#[derive(Debug, Deserialize, JsonSchema)]
#[serde(untagged)]
enum Bar {
    A { a: i32 },
    B { b: String },
}

#[derive(Debug, Deserialize, JsonSchema)]
struct Foo {
    #[serde(flatten)]
    v: Bar,
    c: bool,
}
```

I expected model to have a `c` field and show normal `anyOf` section to document that either `a` or `b` field must be present. I also expected request example to not have `mode` and `omitEmptyAndOptionalProperties` fields.

1. Open https://sandbox.scalar.com/e/1T1zN
2. There are extra fields in the request example and fields `a`, `b` are missing from model definition.

### OpenAPI Document

https://sandbox.scalar.com/e/1T1zN

hanspagel:
I’ve had a nasty typo: Instead of passing options to `getExampleFromSchema` I just merged the options with the result. 🙈 

This PR fixes it and adds a basic test for schemas containing `anyOf`.

See #2218

### DIFF
--- a/.changeset/selfish-ladybugs-pretend.md
+++ b/.changeset/selfish-ladybugs-pretend.md
@@ -1,0 +1,5 @@
+---
+'@scalar/oas-utils': patch
+---
+
+fix: incorrect example for schemas with anyOf
--- a/packages/oas-utils/src/spec-getters/getExampleFromSchema.test.ts
+++ b/packages/oas-utils/src/spec-getters/getExampleFromSchema.test.ts
@@ -834,4 +834,43 @@ describe('getExampleFromSchema', () => {
       },
     })
   })
+
+  it('works with anyOf', () => {
+    expect(
+      getExampleFromSchema({
+        title: 'Foo',
+        type: 'object',
+        anyOf: [
+          {
+            type: 'object',
+            required: ['a'],
+            properties: {
+              a: {
+                type: 'integer',
+                format: 'int32',
+              },
+            },
+          },
+          {
+            type: 'object',
+            required: ['b'],
+            properties: {
+              b: {
+                type: 'string',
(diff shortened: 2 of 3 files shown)


amritk (approved):
Probably right? haha

marclave (approved):

PR #2879, opened 2024-08-16
feat(openapi-parser): migrate openapi parser to monorepo

tmastrom:
This PR migrates the `openapi-parser` package to the monorepo

Real API specifications previously fetched from apis.guru and `.yaml` and `.json` testing specifications have been moved to their own Google Cloud Storage bucket `test-specifications` and are fetched on an as needed basis for testing.

PR #2954, opened 2024-08-27
chore: migrate `@scalar/openapi-parser` to `scalar/scalar`

hanspagel:
This PR migrates `@scalar/openapi-parser`, combining the recent changes and the great work in #2879.

What I’ve added on top of the work in #2879:

* Merged and migrated open PRs in `scalar/openapi-parser`
* Update imports to use the new `@scalar/openapi-types` package
* Use the root README as the new package README
* Use the real-world examples from GCP, but kept the small generic OpenAPI files (easier for local development)
* Migrated the changesets for the unreleased changes (there were a bunch)
* Removed the playground: We don’t need that anymore, we can just use @scalar/api-reference and other packages to see the parser in action.
* Not updated rollup and other dependencies

I’ve also migrated the open issues.

Note: Sorry, this PR is probably too big to properly review. But I’d be very thankful for any review, even if it’s just a quick glance. And we should make sure to test the functionality e.g. in `@scalar/api-reference` and maybe other packages (eg. use some bigger/more complex documents). I didn’t do this yet, but I can do this tomorrow.


tmastrom (approved):
Nice work consolidating the types and testing!

PR #2957, opened 2024-08-27
feat(api-client-app): check for updates

hanspagel:
This PR adds a menu entry to the app to manually check for updates.

If a new version is found, it’ll show the same popup as if the update is auto-detected.

Screenshot 2024-08-27 at 15 23 22


marclave (approved):
niiice :)

PR #2964, opened 2024-08-27
feat(nextjs-openapi): new next.js openapi integration

amritk:
This is still super early so I added the pre-alpha to the title. A lot will change and more work is needed but I wanted to get this merged before moving on to other things just so we have the basic example out there. These references are using a spec which is completely auto generated.

image


marclave (approved):
this is like, the coolest thing ive ever flippin seen.

awesome work @amritk !

PR #2965, opened 2024-08-27
fix(api-client): set active server url in example

antlio:
this pr attempts to get back the server selection from the api reference to the client by setting active server url in the active example function.


antlio:
closing this in favor of #2931

PR #2967, opened 2024-08-28
feat(api-client): remove axios

### LINKED ISSUE #2895: Test request is failing on React JS (MODULE FEDERATION)
### What happens?

I am currently trying to use the "test request" functionality but when trying to send the request in the console I get an error about axios.
I see that scalar internally adds axios to its dependencies
image

image
so, i tried to solve this on my webpack.config.js.  with the next configuration
```js
module.exports = (env) => {
    const envKeys = getEnvKeys(env);
    return {
        output: { publicPath: getOutputDomain(env) },

        entry: './src/index.tsx',

        resolve: {
            extensions: ['.tsx', '.ts', '.jsx', '.js', '.json'],
            fallback: {
                'axios': require.resolve('axios')
            }
        },

        devServer: {
            port: 8081,
            historyApiFallback: true,
            headers: {
                'Access-Control-Allow-Origin': '*',
                'Access-Control-Allow-Methods': 'GET, POST, PUT, DELETE, PATCH, OPTIONS',
                'Access-Control-Allow-Headers': 'X-Requested-With, content-type, Authorization'
            }
        },
        module: {
            unknownContextCritical: false, // suppresses a warning during compile caused by coveo's crypto polyfill which will be fixed in next major release
            rules: [
                {
                    test: /\.m?js/,
                    type: 'javascript/auto',
                    exclude: [/node_modules\/(?!@saffron\/core-components.*)/, /infrastructure/, /api/],
                    resolve: {fullySpecified: false}
                },
                {
                    test: /\.(css|s[ac]ss)$/i,
                    exclude: [/node_modules/, /infrastructure/, /api/],
                    use: ['style-loader', 'css-loader', 'postcss-loader', 'sass-loader']
                },
                {
                    test: /\.(js|jsx|ts|tsx)$/,
                    exclude: [/node_modules/, /infrastructure/, /api/],
                    use: {loader: 'babel-loader'}
                },
                {
                    test: /\.(png|jpg|gif|svg|ttf)$/i,
                    type: 'asset/resource'
                },
                {
                    test: /\.json$/,
                    type: 'json'
                }
            ]
        },

        plugins: [
            new DefinePlugin(envKeys),
            new ModuleFederationPlugin({
                name: 'generalContent',
                filename: 'remoteEntry.js',
                exposes: exposeRoutes(env),
                remotes: { shell: `shell@${getOutputDomainShell(env)}remoteEntry.js` },
                shared: {
                    ...deps,
                    axios: {
                        singleton: true,
                        eager: true,
                        requiredVersion: deps['axios']
                    },
                    '@scalar/api-reference-react': {
                        singleton: true,
                        requiredVersion: deps['@scalar/api-reference-react']
                    },
                    react: {
                        singleton: true,
                        eager: false,
                        requiredVersion: deps.react
                    },
                    'react-dom': {
                        singleton: true,
                        eager: false,
                        requiredVersion: deps['react-dom']
                    },
                    '@auth0/auth0-react': {
                        singleton: true,
                        requiredVersion: deps['@auth0/auth0-react']
                    },
                    'swr': {
                        version: '^2.2.0',
                        singleton: true,
                        eager: false
                    },
                    '@coveo/headless': {
                        singleton: false,
                        eager: false,
                        requiredVersion: deps['@coveo/headless']
                    },
                    'yaml': { // https://github.com/webpack/webpack/issues/13457
                        singleton: true,
                        eager: false,
                        requiredVersion: '^2.3.4',
                        version: '^2.3.4'
                    }
                }
            }),
            // Automatically imports React when needed
            new ProvidePlugin({React: 'react'}),
            new HtmlWebPackPlugin({template: './src/index.html'})
        ]
    };
};


```

and the implementation of the component is the next:

```js
import { useParams } from 'react-router-dom';
import { type AxiosError } from 'axios';
import useFetch from '../../hooks/useFetch/useFetch';
import { DOCUMENT_API_PATH } from '../../Constants';
import { fetchApiReference } from '../../utils/fetchWithCiam';
import useErrorNavigation from '../../hooks/useNavigationError/useNavigationError';
import { type IApiReferenceResponse } from '../../types/apiResponses';
import './ApiReferenceScalar.scss';

function ApiReferenceScalar() {
    const { refId } = useParams();
    const {data, error} = useFetch(`${DOCUMENT_API_PATH}/api-reference/${refId}`, fetchApiReference, {});
    const { navigateToError } = useErrorNavigation(error as AxiosError);
    const apiRefData = data ? JSON.parse((data as IApiReferenceResponse).openApi) : undefined;

    // TODO: remove this hack
    // this condition is circumventing a bug where these two authSchemes in an oasv3 spec throw an error
    // bc it expects to always be able to access the 'password' flow.
    // this is definitely a bug in scalar. https://github.com/scalar/scalar/issues/1960
    if(apiRefData?.components?.securitySchemes?.OAuth2?.flows?.clientCredentials || apiRefData?.components?.securitySchemes?.OAuth2?.flows?.authorizationCode){
        apiRefData.components.securitySchemes.OAuth2.flows.password = {};
    }

    if (navigateToError) return navigateToError;

    return (
        apiRefData && (
            
            <ApiReferenceReact
                configuration={{
                    spec: {content: apiRefData},
                    withDefaultFonts: false,
                    darkMode: false,
                    theme: 'none',
                    hideModels: true
                }}
            />
        )
    );
}

export default ApiReferenceScalar;


```
and the current version of the packages :

```js 
        "@scalar/api-reference-react": "^0.3.77",
        "axios": "^1.7.4"
```

I think the problem is that the component is incompatible with module federation.

I would appreciate all your help. 

I expect the test request component works normally and returns a response from the server

1. open the api reference page
2. click on test request 
image

3. send a request 

image

### OpenAPI Document

_No response_

amritk:
closes #2895 

Went for the easiest way to remove axios, serialized headers and what not. Seems to be working, needs heavy testing! Also added bit of type safety and removed what I hope was unneeded code 🦑, and I _think_ cookies should work now? need to test

### DIFF
--- a/packages/api-client/src/libs/sendRequest.ts
+++ b/packages/api-client/src/libs/sendRequest.ts
@@ -132,8 +132,13 @@ export const sendRequest = async (
   } else if (example.body.activeBody === 'raw' && example.body.raw.value) {
     data = example.body.raw.value
   } else if (example.body.activeBody === 'formData') {
-    headers['Content-Type'] = 'multipart/form-data'
-
+    /**
+     * The header has to look something like this:
+     *
+     * Content-Type: multipart/form-data; boundary=----formdata-undici-043007900459
+     *
+     * fetch() makes sure to generate this properly, we must not set it manually
+     */
     const bodyFormData = new FormData()
     if (example.body.formData.encoding === 'form-data') {
       example.body.formData.value.forEach(
--- a/packages/api-client/src/libs/sendRequest.ts
+++ b/packages/api-client/src/libs/sendRequest.ts
@@ -140,15 +140,25 @@ export const sendRequest = async (
      * fetch() makes sure to generate this properly, we must not set it manually
      */
     const bodyFormData = new FormData()
+
     if (example.body.formData.encoding === 'form-data') {
       example.body.formData.value.forEach(
         (formParam: { key: string; value: string; file?: File }) => {
-          const value = formParam.file ? formParam.file : formParam.value
-          if (formParam.key && value) {
-            bodyFormData.append(formParam.key, value)
+          // Add File to FormData
+          if (formParam.key) {
+            if (formParam.file) {
+              bodyFormData.append(
+                formParam.key,
+                formParam.file,
+                formParam.file.name,
+              )
+            } else if (formParam.value !== undefined) {
+              bodyFormData.append(formParam.key, formParam.value)
+            }
           }
         },
       )
+
       data = bodyFormData
     }
   }
@@ -269,7 +279,9 @@ export const sendRequest = async (
     signal: abortSignal,
   }
 
-  config.body = data
+  if (data) {
+    config.body = data
+  }
 
   // Start timer to get response duration
   const startTime = Date.now()
--- a/packages/api-client/src/libs/sendRequest.ts
+++ b/packages/api-client/src/libs/sendRequest.ts
@@ -269,7 +269,7 @@ export const sendRequest = async (
     signal: abortSignal,
   }
 
-  if (data) config.body = JSON.stringify(data)
+  config.body = data
 
   // Start timer to get response duration
   const startTime = Date.now()
--- a/packages/api-client/src/libs/sendRequest.ts
+++ b/packages/api-client/src/libs/sendRequest.ts
@@ -137,8 +137,10 @@ export const sendRequest = async (
      *
      * Content-Type: multipart/form-data; boundary=----formdata-undici-043007900459
      *
-     * fetch() makes sure to generate this properly, we must not set it manually
+     * fetch() makes sure to generate this properly, we must make sure to delete it
      */
+    delete headers['Content-Type']
+
     const bodyFormData = new FormData()
 
     if (example.body.formData.encoding === 'form-data') {
--- a/packages/api-client/src/libs/sendRequest.ts
+++ b/packages/api-client/src/libs/sendRequest.ts
@@ -293,8 +293,7 @@ export const sendRequest = async (
     }
   } catch (e) {
     const sentTime = Date.now()
-    const duration = Date.now() - startTime
-    const payload = { sentTime, request: example, duration }
+    const payload = { sentTime, request: example }
 
     // We have a response from fetch
     if (e instanceof Response) {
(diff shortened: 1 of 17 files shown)


marclave (approved):
some tests are failing but once they go green we can merge!

maybe @hanspagel can help

hanspagel:
OMG, I love this, so great!

I rebased the PR and fixed a few tests! A rabbit hole I got lost into:

When `fetch()` gets `FormData` with a `File` the void server only receives the beginning of the body. It seems that `fetch()` is switching to streaming for files, which might be what we want, but I’m not able test this.

Long story short: The `File` test is disabled, and we should double check files are still working properly.

amritk:
Yea I think we can do things differently with fetch but I didn't want to change too much of marcs buffer logic. We can definitely switch to using the appropriate parsing methods (json, blob etc) instead of just arrayBuffer. 

But that would mean much more changes, I'll see if file works

packages/api-client/src/libs/sendRequest.ts
```diff
@@ -93,18 +132,35 @@ export const sendRequest = async (
...
   } else if (example.body.activeBody === 'formData') {
-    headers['Content-Type'] = 'multipart/form-data'
+    /**
+     * The header has to look something like this:
+     *
+     * Content-Type: multipart/form-data; boundary=----formdata-undici-043007900459
+     *
+     * fetch() makes sure to generate this properly, we must make sure to delete it
+     */
+    delete headers['Content-Type']
```
amritk, line 142:
```suggestion
    delete headers['Content-Type']
```
alright @hanspagel I think this is what we needed, try it now! 

Not sure if we want to delete this for everything or just formData

hanspagel: I didn’t want to delete the header, because people should still be able to overwrite headers, e.g. send a wrong Content-Type header to test the behaviour of their API.

Is that fixing something specifically? Because the mentioned test, that is marked as failing, is still failing. 🤔

amritk: Yea this fixes sending as formData because it adds the boundary at the end. Otherwise we might need to manually add the end boundary

amritk:
Alright I went deep on the last test and it turns out that it is unrelated to our code. Looks like the fetch its using has an issue or we need to patch it or something. In the future we can try using node-fetch or something else which might work. Going to merge it for now!

PR #2969, opened 2024-08-28
feat(api-client): track active workspace in local storage

hwkr:
I tried this a few different ways but this seemed to be the simplest and most robust. Persists the most recent active workspace `uid` so that we can go to it when we load `/`.

Client Web:

https://github.com/user-attachments/assets/5fef215e-f953-4e25-8f1f-f9f773c597eb

Client App:

https://github.com/user-attachments/assets/92d22c16-57b9-495f-ae91-73ce581ce7a3

### DIFF
--- a/.changeset/twelve-penguins-end.md
+++ b/.changeset/twelve-penguins-end.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
+
+feat(api-client): track active workspace in local storage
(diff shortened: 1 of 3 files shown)


marclave (approved):
this looks good to me! do we need to do any SSR checks here @amritk with localstorage?

hanspagel (approved):
nice ux improvement 👏

amritk:
the client is client only, so we only need to worry about electron. Just make sure it works there

hanspagel:
But it does, but I think it would cause more conflicts in #2931, right? 😵

PR #2977, opened 2024-08-28
fix(cli): init test deletes `scalar.config.json`

hanspagel:
The `scalar init` test deleted the `scalar.config.json` in root.

This PR makes all CLI tests use files inside the CLI directory.

This is also great, because you can run `pnpm test` in the package folder again. :)


tmastrom (approved):
Looks good

PR #2982, opened 2024-08-29
chore: eslint script max lines

antlio:
this pr adds max lines rules to the eslint config in order to prevent components from containing any significant amount of business logic. this aims to encourage the creation of helper files, so as to maintain an atomic approach to the functions written and tested in them.

```
  1:1  warning  Block has too many lines (79). Maximum allowed is 50  vue/max-lines-per-block

✖ 1 problem (0 errors, 1 warning)


│  package_format (skip) no matching staged files
┃  prettier ❯ 
```

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -218,6 +218,13 @@ module.exports = {
           },
         ],
         'vue/custom-event-name-casing': ['error', 'camelCase'],
+        'vue/max-lines-per-block': [
+          'warn',
+          {
+            script: 50,
+            skipBlankLines: true,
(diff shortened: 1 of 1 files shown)


hanspagel:
LOVE it! 

I’d say let’s do 50 lines, but warn instead of error.

That would probably enough nudging for the beginning. :)

amritk (approved):
is 50 too low? guess well see. Adjust as we go. nice work!

geoffgscott (approved):

marclave (approved):

PR #2984, opened 2024-08-29
feat(api-client-app): `scalar://` links

hanspagel:
With this PR the app is registered as the default handler for `scalar://` protocol links, this allows to import an OpenAPI document through links.

Let’s say we’d love to import our Scalar Galaxy example:
`https://cdn.jsdelivr.net/npm/@scalar/galaxy/dist/latest.json`

We can encode the URI and prefix it with our protocol:
`scalar://${encodeURIComponent("https://cdn.jsdelivr.net/npm/@scalar/galaxy/dist/latest.json")}`

Which gives us this:
scalar://https%3A%2F%2Fcdn.jsdelivr.net%2Fnpm%2F%40scalar%2Fgalaxy%2Fdist%2Flatest.json

To test how this opens in the app, do the following:

1. `$ cd packages/api-client-app`
2. `$ pnpm dev`
3. `$ open scalar://https%3A%2F%2Fcdn.jsdelivr.net%2Fnpm%2F%40scalar%2Fgalaxy%2Fdist%2Flatest.json`

There’s also a simple HTML file that we’d eventually need to host somewhere, but more on this in a separate PR. :)

### DIFF
--- a/.changeset/dirty-months-jump.md
+++ b/.changeset/dirty-months-jump.md
@@ -1,0 +1,5 @@
+---
+'scalar-api-client': patch
+---
+
+feat: register as the default app for `scalar://` links
--- a/packages/api-client-app/open/index.html
+++ b/packages/api-client-app/open/index.html
@@ -1,0 +1,48 @@
+<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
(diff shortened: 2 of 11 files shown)


amritk (approved):
awwww yisssss

PR #2986, opened 2024-08-29
feat(api-client): custom icons for collections

hwkr:
Add a custom icon picker for collections ✨ 

https://github.com/user-attachments/assets/73df7d85-940b-453b-83f1-6dffee026e81


hanspagel (commented):
Oh, this is a 9% increase to the bundle size! Is that really worth the feature? Or can we find another approach? Or offer less icons? 🤔

amritk:
@hanspagel ideally we get treeshaking going so it only uses the icons that it needs. We had lazy loaded icons before but there were issues with SSR so I removed it. We can bring it back IF we can get it working with ssr but imo tree shaking would be better

hanspagel:
Most of our integrations use the CDN version which bundles everything into one file! That’s what the bundle size check tests.

I’ve built an ESM version with chunks twice already, but both versions’ entrypoint was bigger than the current CDN bundle for some reason.

Would love to pair on this at some point, but there’s just so much cool and important stuff going on. 😅

hwkr:
@amritk @hanspagel I'm not sure the best way to tackle this but I'm open to options, does this need to be bundled with the CDN version at all? Currently these icons are only used in the client so if we can make sure they're not getting bundled into the CDN version then we can avoid those issues at least for the time being

amritk:
@hwkr quick fix it is! Also on second thought I don't even think we could do treeshaking without explicit imports 
```
import { Close } from 'icons'
```
so we can probably look into lazy loading again at some point

hanspagel:
Yeah, can we do explicit imports? I think that would really help with the bundle size (didn’t test though).

marclave (approved):

packages/oas-utils/src/entities/spec/collection.ts
```diff
@@ -53,6 +53,8 @@ export const extendedCollectionSchema = z.object({
   tags: nanoidSchema.array().default([]),
   /** List of requests without tags and top level tag "folders" */
   children: nanoidSchema.array().default([]),
+  /** An icon representing the collection */
+  icon: z.string().default('interface-content-folder'),
```
hwkr, line 57: @amritk @marclave  should this be `icon` or `x-scalar-icon`? I'm not sure how this schema relates to the open api spec 🤔

amritk: You probably want to move it up to the `oasCollectionSchema` and call it `x-scalar-icon` but putting it down there isn't the end of the world. If you leave it in the bottom there as icon we will just need to convert it on the way in/out. Putting it above in the openapi spec section saves the conversion

amritk: usually we have a spec compliant schema which starts with `oas` and an extended schema which is the non spec compliant stuff

hwkr:
Alright this should be all up to date. I trimmed it down to only the icons we're showing in the selected which seems to have reduced the bundle size somewhat, we should be able to also not have the icons in the references since they aren't needed there but I'm not sure how to do that.

packages/api-client/src/components/IconSelector.vue
```diff
@@ -0,0 +1,57 @@
...
+  placement?: AlignedPlacement
+}>()
+
+const emit = defineEmits<{
+  (e: 'update:modelValue', icon: string): void
+}>()
+
+type Side = 'top' | 'bottom'
+type Alignment = 'start' | 'end'
+type AlignedPlacement = `${Side}-${Alignment}`
```
amritk, line 18: nice!

packages/scalar.aspnetcore/playground/Scalar.AspNetCore.Playground/Properties/launchSettings.json
```diff
@@ -1,15 +1,15 @@
...
-      "launchBrowser": true,
-      "launchUrl": "scalar/v1",
-      "applicationUrl": "https://localhost:7201;http://localhost:5056",
-      "environmentVariables": {
-        "ASPNETCORE_ENVIRONMENT": "Development"
-      }
-    }
-  }
-}
+﻿{
```
amritk, line 1: was this meant to be changed?

amritk (approved):
LGTM!

packages/scalar.aspnetcore/playground/Scalar.AspNetCore.Playground/Properties/launchSettings.json
hwkr: Oh no prettier did this I guess

packages/api-client/src/components/IconSelector.vue
hwkr: I stole this from inside floating ui 😂 But it is pretty slick

packages/api-client/src/views/Request/types/sidebar-item.ts
```diff
@@ -14,7 +14,8 @@ export type SidebarItem = {
   method?: RequestMethod
   link?: string
   warning?: string
-  rename: (name: string) => void
+  icon?: string
+  rename: (name: string, icon?: string) => void
```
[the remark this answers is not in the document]
hanspagel: Done :)

hanspagel (approved):
This is great! Love how that turned out.

I’d just prefer to call it `@scalar/icons` as we have `@scalar/components` (and not `@scalar/component-library`).

Agree?

packages/icon-library/package.json
```diff
@@ -0,0 +1,48 @@
+{
+  "name": "@scalar/icon-library",
```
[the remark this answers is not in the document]
hwkr: Fine with me!

PR #2989, opened 2024-08-30
feat(api-client): fix up and polish command palette

hwkr:
Started this PR to just clean up modals but ended up finding a bunch of global hotkey issues.

I moved the search and the command palette navigation hotkeys out of the global bus and onto the respective inputs. I'm not sure if there was a reason they needed to be global? I think those actions should only fire when their input is focused anyways and the global bus should be used for app wide hotkeys like CMD Enter, CMD K for the palette etc.

## Before / After

https://github.com/user-attachments/assets/78af75fc-fde0-4827-8943-5d5f8844dc41


hanspagel:
What a great PR! We have to wait until #2931 is merged, though. 😵

hanspagel (approved):
Awesome! lgtm, @amritk wdyt?

amritk:
Yup its the right play, just gunna rebase

amritk (approved):
Lets go!

PR #2991, opened 2024-08-30
chore: require some servers to be running for some test files

hanspagel:
With this PR the contributing guide is updated to list the required servers (`@scalar/void-server` + `@scalar/proxy-server`) for a complete test run.

And with this PR, in the two test files that require them, we check whether they are running and throw a helpful error if they aren’t running already.


amritk (approved):
Yes helpful!

PR #2992, opened 2024-08-30
feat(api-client-app): fetch OpenAPI documents from various sources 👀 

hanspagel:
With this PR the OpenAPI document URL is retrieved from various sources. There’s a new helper method to which you can pass any URL and it’ll do the following:

```bash
✓ returns JSON urls
✓ returns YAML urls
✓ finds URLs from sandbox URL
✓ finds URL in the CDN example
✓ works with single quote data attributes
✓ returns absolute URLs
✓ finds URLs in some wrangled configuration object
✓ finds URLs in redoc HTML
✓ finds URLs in redoc JS
```

With this PR it’s used in the app to get the OpenAPI document URLs from `scalar://` app links. You can see it in action with our Hono example here:

https://github.com/user-attachments/assets/87ebeb62-1eee-4ef3-bc51-83e404bcd14f

### DIFF
--- a/.changeset/nasty-donuts-approve.md
+++ b/.changeset/nasty-donuts-approve.md
@@ -1,0 +1,5 @@
+---
+'@scalar/oas-utils': patch
+---
+
+feat: new findOpenApiDocument helper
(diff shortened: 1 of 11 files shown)


amritk (approved):
Not sure whats going on here... but I like it 👍🏾

PR #2993, opened 2024-08-30
Update package.json path export error

fjdvchain:
Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: No "exports" main defined in /node_modules/@scalar/hono-api-reference/package.json

**Problem**
Currently, …

**Explanation**
This happens because …

**Solution**
With this PR …

### DIFF
--- a/packages/hono-api-reference/package.json
+++ b/packages/hono-api-reference/package.json
@@ -29,6 +29,7 @@
   "exports": {
     ".": {
       "import": "./dist/index.js",
+      "default": "./dist/index.js",
       "types": "./dist/index.d.ts"
     }
   },


fjdvchain:
This change in locally fixed my issue running this with `tsx`

hanspagel:
Thanks for the contribution! Do we need the `default` import? 🤔 Do you experience an issue somewhere, can you provide some more information?

fjdvchain:
> Thanks for the contribution! Do we need the `default` import? 🤔 Do you experience an issue somewhere, can you provide some more information?

I followed a fix in another repositories pull request. They came across the same issue I was having using hono scalar [here](https://github.com/nextauthjs/next-auth/pull/11224)

hanspagel:
Sorry, this PR won’t work because the exports are auto-generated. I’ve created an issue: #2994

PR #2995, opened 2024-08-30
chore(api-reference): unify `@scalar/snippetz` and `httpsnippet-lite` API

hanspagel:
WIP

This is a follow up to #2899, which adds a helper method to get code example from `@scalar/snippetz` and `httpsnippet-lite` without caring which comes from which package.

I’ve chosen to not add `httpsnippet-lite` to `@scalar/snippetz` and add a meta helper on top of both. Eventually, we’d like to get rid of `httpsnippet-lite` and I’d prefer to keep `@scalar/snippetz` light-weight. :)

Note: This is not used yet, I’m waiting for movement in #2825. Once that’s completed, I’ll replace the actual code with the new helper.

Wdyt?

**Tasks**
- [ ] Get rid of the `httpsnippet-lite` `HarRequest` type


packages/api-reference/src/helpers/createRequest.ts
```diff
@@ -0,0 +1,18 @@
+import type { Request } from '@scalar/snippetz'
+
+/**
+ * Creates a complete HAR request, even if just passed a partial request.
+ */
+export function createRequest(partialRequest?: Partial<Request>): Request {
```
tmastrom, line 6: We may need to generalize this function to accept `Partial<HarRequest>` as well

packages/api-reference/src/features/ExampleRequest/ExampleRequest.vue
```diff
@@ -113,7 +115,7 @@ async function generateSnippet() {
   }
 
   // Generate a request object
-  const request = getHarRequest(
+  const harRequest = getHarRequest(
```
tmastrom, line 118: I don't have a good understanding of the differences between `HarRequest` and `Request` types. This will be key for streamlining this api.

packages/api-reference/src/features/ExampleRequest/ExampleRequest.vue
```diff
@@ -137,25 +139,12 @@ async function generateSnippet() {
...
-    return (await snippet.convert(
-      httpClient.targetKey,
-      httpClient.clientKey,
-    )) as string
-  } catch (e) {
-    console.error('[ExampleRequest]', e)
-    return ''
-  }
+  return (
+    (await getExampleCode(harRequest as any, targetKey, clientKey as string)) ??
```
tmastrom, line 145: We need to eliminate this type casting if possible

tmastrom (changes requested):
On the right track but usage in `ExampleRequest` component is still clunky

packages/api-reference/src/features/ExampleRequest/ExampleRequest.vue
tmastrom: Not sure if it's possible yet but this `getHarRequest` may be helpful or even replace the `createRequest` helper util

packages/api-reference/src/helpers/getExampleCode.ts
```diff
@@ -0,0 +1,51 @@
...
+  type TargetId as HttpSnippetLiteTargetId,
+} from 'httpsnippet-lite'
+
+export type TargetId = HttpSnippetLiteTargetId | SnippetzTargetId
+export type ClientId = HttpSnippetLiteClientId | SnippetzClientId
+
+/**
+ * Returns a code example for given HAR request
+ */
+export async function getExampleCode(
```
tmastrom, line 19: need to tighten up this return type to just `string`

hanspagel:
Great feedback! I’ll do another round next week. ✌️

packages/api-reference/src/features/ExampleRequest/ExampleRequest.vue
hanspagel: I renamed `getHarRequest` to `mergeRequests` to make its intention more clear. :v:

I tried getting rid of HarRequest and using Request only, but I run into some issues. Need to try again tomorrow.

tmastrom (approved):
Fixed the return type in `getExampleCode` and now things are working great!

hanspagel:
Cool, let’s get this in! :) I’ve more ideas to improve this and get less reliant on httpsnippet-lite, but that’s for another PR I guess. :)

PR #2997, opened 2024-08-30
feat: Add auth select back in

geoffgscott:


amritk (approved):
Looks like it probably works

PR #2998, opened 2024-08-30
feat(cli): `scalar check` command to check `scalar.config.json` files

tmastrom:
This PR adds the `check` cli command to validate `scalar.config.json` files using [`typebox`](https://github.com/sinclairzx81/typebox) schema validation

## Example usage
```bash 
scalar check /packages/cli/src/commands/check/valid.json
```
Screenshot 2024-09-25 at 12 09 21 PM
Screenshot 2024-09-25 at 12 08 58 PM
Screenshot 2024-09-25 at 12 08 41 PM

### To Test
You can use the ` pnpm --filter cli cli:link` command to check out these latest changes locally


hanspagel (approved):
This is sooo great! I desperately need this.

I’ve added to the CLI README, update the wording a little bit, tried to improve the output for the check command and replaced the `useFileOrConfiguration` helper, because that’s for resolving to OpenAPI documents, not for configuration files.

There were a few cases where the check method return valid: true for invalid files, too. I’ve improve the utility, but the test is passing for the `valid.json` file now. Would you mind to look into this?

hanspagel (changes requested):
Sorry, I’ve overwritten your changes it seems. We need to restore them. 🫨

packages/config/src/check.ts
```diff
@@ -0,0 +1,58 @@
...
+    const result = Value.Check(ScalarConfigSchema, scalarConfigJson)
+    const errors = [...Value.Errors(ScalarConfigSchema, scalarConfigJson)]
+
+    // Early return if there is already an error with the schema at the root level
+    if (!result) {
+      return {
+        valid: errors.length === 0,
+        errors: errors,
+      }
+    }
```
tmastrom, line 23: Opinions on this early return? It is useful for type safety in the rest of the function.

packages/config/src/configTypes.ts
```diff
@@ -0,0 +1,117 @@
...
+  name: Type.String(),
+  description: Type.Optional(Type.String()),
+  sidebar: Type.Array(SidebarItemSchema),
+})
+
+// TODO: Currently only using the explicit guide schema
+export const ConfigGuideSchema = Type.Union([
+  ConfigGuideDirectorySchema,
+  ConfigGuideExplicitSchema,
+])
```
tmastrom, line 64: What is the use case for the GuideDirectory?

packages/config/src/utils.ts
```diff
@@ -0,0 +1,32 @@
...
+>(item: T, schema: any) {
+  const errors = [...Value.Errors(schema, item)]
+
+  // Add the item name for better errors
+  errors.forEach((error) => {
+    const errorMessage = `${error.message} on sidebar item: ${item.name}`
+    error.message = errorMessage
+  })
+  return errors
+}
```
tmastrom, line 21: I think this is necessary because the full path for the object being validated is lost while iterating over the nested objects. Open to improvements on this

packages/config/test/check.test.ts
```diff
@@ -0,0 +1,30 @@
...
+    console.log(result.errors)
+  })
+  it('checks the config file with valid reference object', () => {
+    const result = check(path.resolve('./test/valid-reference.json'))
+    expect(result.valid).toBe(true)
+  })
+  it.todo('throws an error for extra properties', () => {
+    const result = check(path.resolve('./test/invalid-extra.json'))
+    expect(result.valid).toBe(false)
+  })
```
tmastrom, line 29: Need to verify how critical this feature is. It is possible that core will strip unnecessary properties for us.

tmastrom (commented):
First draft working `scalar check`! 

Currently using a workaround to iterate over the sidebar items and children objects and validate those schemas individually. In an ideal world this is captured by the defined Typebox schema but I have spent too much time trying to make the recursive types work and I will try again on the next iteration.  

Left some other comments for things to add / clean up but wanted some initial feedback.

packages/config/src/check.ts
hanspagel: Great

hanspagel (approved):
This is so great! Let’s get it in, we can further improve it over time. :)

PR #3001, opened 2024-08-30
chore(api-client-app): deploy open electron app page

tmastrom:
We need an intermediate page before we open the Electron app from a link.

This PR deploys a simple vue app page to cloud run using cloud build

### TODO: 
- [ ] point it at `https://open.scalar.com/`


hanspagel (approved):
Thanks for putting this together! It’ll be an intermediate solution and we’ll eventually make it a modal on `client.scalar.com`, but this is great to have it running *now*. :)

Let’s get this in!

PR #3002, opened 2024-08-30
feat: add server selection

amritk:
Server selection working for both request server and servers, adding servers only working for collection rn

image
image

also did some work in the request command pallette, its _almost_ there


amritk:
@hanspagel this one aint going green! @geoffgscott good to merge?

PR #3014, opened 2024-09-02
feat(api-client): add sidebar request creation button

antlio:
this pr adds a create request button in empty request sidebar collection/folder to display the request creation command modal: 

https://github.com/user-attachments/assets/3b3e7e0f-5424-4bdc-844e-22a4ca33961f

### DIFF
--- a/.changeset/tasty-adults-press.md
+++ b/.changeset/tasty-adults-press.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
+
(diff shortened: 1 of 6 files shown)


hanspagel:
This is SO great! Unfortunately, we should wait for #2931 to be in to avoid further conflicts over there.

antlio:
> This is SO great! Unfortunately, we should wait for #2931 to be in to avoid further conflicts over there.

rebased

PR #3015, opened 2024-09-02
fix: autoprefixer warnings, fix #3003

### LINKED ISSUE #3003: autoprefixer warning: end value has mixed support, consider using flex-end instead
### What happens?

I'm using `api-reference-react`, included the `.css` in our project and we're running [autoprefixer](https://github.com/postcss/autoprefixer) automatically. This throws the following warnings in our project;

```
WARNING in ../../node_modules/@scalar/api-reference-react/dist/style.css (../../node_modules/@scalar/api-reference-react/dist/style.css.webpack[javascript/auto]!=!../../node_modules/css-loader/dist/cjs.js??ruleSet[1].rules[6].oneOf[4].use[1]!../../node_modules/postcss-loader/dist/cjs.js??ruleSet[1].rules[6].oneOf[4].use[2]!../../node_modules/@scalar/api-reference-react/dist/style.css)
Module Warning (from ../../node_modules/postcss-loader/dist/cjs.js):
Warning

(1:169877) autoprefixer: start value has mixed support, consider using flex-start instead

WARNING in ../../node_modules/@scalar/api-reference-react/dist/style.css (../../node_modules/@scalar/api-reference-react/dist/style.css.webpack[javascript/auto]!=!../../node_modules/css-loader/dist/cjs.js??ruleSet[1].rules[6].oneOf[4].use[1]!../../node_modules/postcss-loader/dist/cjs.js??ruleSet[1].rules[6].oneOf[4].use[2]!../../node_modules/@scalar/api-reference-react/dist/style.css)
Module Warning (from ../../node_modules/postcss-loader/dist/cjs.js):
Warning

(1:169895) autoprefixer: end value has mixed support, consider using flex-end instead
```

This is not an error but just a warning shown each time when reloading the app in dev mode; the app starts and behaves just fine. Because it shows on each reload it is however quite spammy.

Searching this codebase there are several places where `: end` and `: start` are used (e.g. `justify-content: end` and `align-items: start;`). Where possible these should be `flex-start` and `flex-end` if they're part of a flexbox layout.

I'm not 100% sure about the practical differences (if any) between `: end|start` and their `flex-` alternatives. Afaik `flex-start` is specific to `display: flex`; https://developer.mozilla.org/en-US/docs/Web/CSS/align-items#flex-start

I tried reproducing using the React project in `examples/` but got the following error running `npm install`:

```
Unsupported URL Type "workspace:": workspace:*
```

I checked a few example READMEs but I'm not sure how to resolve that error.

Most likely when adding an autoprefix postcss plugin to the  react example it'll yield the warning as well. I'm using a fairly up-to-date browserlistrc (= used by autoprefixer):

```
> 1%
not dead
```

### OpenAPI Document

_No response_

hanspagel:
This replaces some CSS values with mixed support with alternative values.

While they behave [slightly different according to MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/justify-content#flex-start), I couldn’t spot any difference in our use case.

See #3003

### DIFF
--- a/.changeset/weak-ties-rescue.md
+++ b/.changeset/weak-ties-rescue.md
@@ -1,0 +1,7 @@
+---
+'@scalar/api-reference': patch
+'@scalar/components': patch
+'@scalar/use-toasts': patch
+---
+
+fix: autprefixer warnings
--- a/packages/api-reference/src/components/Content/Operation/PathResponses/PathResponses.vue
+++ b/packages/api-reference/src/components/Content/Operation/PathResponses/PathResponses.vue
@@ -203,7 +203,7 @@ const showSchema = ref(false)
   border-top: var(--scalar-border-width) solid var(--scalar-border-color);
 }
 .response-example-selector {
-  align-self: start;
+  align-self: flex-start;
   margin: -4px;
 }
 .response-description {
--- a/packages/api-reference/src/components/Content/Schema/Schema.vue
+++ b/packages/api-reference/src/components/Content/Schema/Schema.vue
@@ -226,7 +226,7 @@ button.schema-card-title:hover {
   width: 100%;
 }
 .schema-card--compact {
-  align-self: start;
+  align-self: flex-start;
 }
 
 .schema-card--compact.schema-card--open {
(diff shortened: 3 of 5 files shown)


antlio (approved):
⚡️

PR #3016, opened 2024-09-02
feat(api-reference): allow to hide the “Test Request” button

hanspagel:
Every now and them [someone asks to disable the “Test Request” button](https://github.com/scalar/scalar/discussions/3005), which launches the embedded API client. Currently, people just hide it with CSS or even use alternatives, just because they can’t disable it. 

This PR introduces the `hideTestRequestButton` option to allow users to hide the button.

I’ve kept the footer of the card, because otherwise we’d need the logic to disable the button in the parent component, and these are 3 different components. Keeping the footer was way easier, so I went with this solution.

**Preview**

```js
{
  hideTestRequestButton: true
}
```

Screenshot 2024-09-02 at 15 23 55

### DIFF
--- a/.changeset/chilly-cycles-breathe.md
+++ b/.changeset/chilly-cycles-breathe.md
@@ -1,0 +1,6 @@
+---
+'@scalar/api-reference': patch
+'@scalar/types': patch
+---
+
+feat: allow to hide the Test Request button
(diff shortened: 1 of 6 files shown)


antlio:
nice! i'm down to hide the footer as well but can be done in another PR!

marclave (approved):

antlio:
should this be added to ASP and fast api also?

PR #3017, opened 2024-09-02
style: don't make the path variable italic

hanspagel:
I don’t think it was intentional to have the path variable italic, so this PR makes it `font-style: normal` again:

**Before**
Screenshot 2024-09-02 at 15 22 51

**After**
Screenshot 2024-09-02 at 15 22 59


hwkr (approved):
Yeah that doesn't seem right

marclave (approved):

PR #3023, opened 2024-09-03
ci: build and test on Node 18 and 20

hanspagel:
The `File` type [slips through again and again](https://github.com/scalar/scalar/actions/runs/10673537971/job/29601485700?pr=3022), and it causes the tests in the release PR to fail, because this is the only place we build and test against Node 18 (where we don’t have the `File` type).

Let’s just build and test every PR against Node 18 and 20 again.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,38 +9,11 @@ on:
       - main
 
 jobs:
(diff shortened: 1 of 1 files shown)


tmastrom (approved):
This works for me

PR #3024, opened 2024-09-03
fix: `File` is not defined (Node 18) DOC-2222

hanspagel:
The `File` type is not available in Node 18, so let’s use this object as a workaround.

### DIFF
--- a/.changeset/fluffy-eagles-do.md
+++ b/.changeset/fluffy-eagles-do.md
@@ -1,0 +1,5 @@
+---
+'@scalar/oas-utils': patch
+---
+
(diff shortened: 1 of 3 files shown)


tmastrom (commented):
This was on [my todo list](https://linear.app/api-docs/issue/DOC-2222/oas-utils-file-type) as well! Added the full File type definition (and[ Blob type ](https://developer.mozilla.org/en-US/docs/Web/API/File)which it extends) 

These schemas could be centralized and used other places in the codebase as well.

hanspagel:
OMG NICE

PR #3026, opened 2024-09-03
feat(api-reference): allow to hide search

antlio:
This PR is a proposal to introduce the `hideSearch` option to allow users to hide the search from api reference sidebar as requested in #2869.

**Preview**

```js
{
  hideSearch: true
}
```

image

### DIFF
--- a/.changeset/ten-pets-return.md
+++ b/.changeset/ten-pets-return.md
@@ -1,0 +1,6 @@
+---
+'@scalar/api-reference': patch
+'@scalar/types': patch
+---
+
+feat: allow to hide search sidebar
--- a/documentation/configuration.md
+++ b/documentation/configuration.md
@@ -102,6 +102,18 @@ Whether to show the “Test Request” button
 }
 ```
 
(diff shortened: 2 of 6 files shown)


hanspagel (commented):
Should it hide the search in the client, too? 🤔 And do we need to disable the keyboard shortcut to?

antlio:
> Should it hide the search in the client, too? 🤔 And do we need to disable the keyboard shortcut to?

no strong opinion, but i guess we could do it there too 

oops initially disabled the modal but i guess i removed it, will add again!

marclave:
> Should it hide the search in the client, too? 🤔 And do we need to disable the keyboard shortcut to?

we can keep search in the api client, cause that search is more of a filter : )

antlio:
> > Should it hide the search in the client, too? 🤔 And do we need to disable the keyboard shortcut to?
> 
> we can keep search in the api client, cause that search is more of a filter : )

should be good to go then

marclave (approved):

PR #3027, opened 2024-09-03
fix(api-reference): operation accordion test icon

antlio:
this pr hides the test icon in classic layout api reference when `hideTestRequestButton` is enabled introduced in #3016:

**before**
image

**after**
image


hanspagel:
Oh, damn! I’ve had the file open but it looks like I forgot to write the code. 😂

hanspagel (approved):

PR #3030, opened 2024-09-03
fix(api-client): request header

antlio:
this pr favors span markup over input in the api client request header on read only mode to avoid style issue:

**before**
image

**after**
image


hanspagel:
Has to wait, unfortunately :)

PR #3031, opened 2024-09-03
fix(api-reference): anchor offset

antlio:
this pr moves id to section markup level for model and webhook components in order to fix anchor offset issue as seen below:

**before**

https://github.com/user-attachments/assets/7a82f0a7-2ba1-40cd-86c1-84408164c80a

**after**

https://github.com/user-attachments/assets/46bd3508-1e62-479f-993a-0d470575cdb3

### DIFF
--- a/.changeset/green-adults-crash.md
+++ b/.changeset/green-adults-crash.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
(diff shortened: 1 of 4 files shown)


cameronrohani:
Wicked, was about to report this bug 😎

hanspagel (approved):

PR #3037, opened 2024-09-04
fix(api-reference): endpoints description

### LINKED ISSUE #2970: Wrong scroll behaviour when the description is long
### What happens?

When I open a page for the first time with a URL constructed to point to a given endpoint (eg. http://localhost:4200#tag/planets), the scroll gets me to the wrong position. It happens only when the description of the endpoints is long. I created a modified Scalar Galaxy spec which reproduces this issue. Note that it works fine after the initial load of the API reference.

It appears for both `defaultOpenAllTags` ReferenceConfiguration option set to `false` and `true` but it is easier to spot if is set to `true`.

The scroll behaviour should be the same for both long descriptions and short ones.

1. Download spec which I prepared:
[bug-example.json](https://github.com/user-attachments/files/16778644/bug-example.json)
2. Go to `#tag/planets` anchor (eg. http://localhost:4200#tag/planets)
3. Api reference points to "Get a token" section instead of "Planets" section:
image

### OpenAPI Document

_No response_

antlio:
this pr adds collapse state to endpoints component in order to clamp description when collapsed and fix #2970:

https://github.com/user-attachments/assets/49e167f0-5dde-4c6c-b02f-b03800664a82


packages/components/src/components/ScalarMarkdown/ScalarMarkdown.vue
```diff
@@ -31,9 +32,18 @@ onServerPrefetch(async () => await sleep(1))
 <template>
   <div
     class="markdown"
+    :class="{ collapsed: !isCollapsed }"
     v-html="html" />
 </template>
 
 <style>
 @import '@scalar/code-highlight/css/markdown.css';
+
```
hanspagel, line 41: Can we have two style tags here? One for the Markdown CSS and one `scoped` for the `.collapsed` CSS?

hanspagel:
LOVE IT, keep those PRs coming

packages/components/src/components/ScalarMarkdown/ScalarMarkdown.vue
antlio: oh great catch i didnt see that!

hanspagel (approved):

PR #3043, opened 2024-09-04
fix(docusaurus): remove headless ui on page change

hwkr:
## Before

Google Chrome-2024-09-04-14-51-34@2x

## After

Google Chrome-2024-09-04-14-50-26@2x

### DIFF
--- a/.changeset/little-pets-approve.md
+++ b/.changeset/little-pets-approve.md
@@ -1,0 +1,5 @@
+---
+'@scalar/docusaurus': patch
+---
+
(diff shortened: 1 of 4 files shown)


amritk (approved):
I have a feeling we are still leaking memory somewhere, but if this fixes the issue!

hwkr:
> I have a feeling we are still leaking memory somewhere, but if this fixes the issue!

Yeah this made the dev tools more responsive but we might still be leaking somewhere in general 🤔

PR #3044, opened 2024-09-04
fix: request headers and body

amritk:
Fixed request headers and body, it no longer errors. 

*I did yank out the content type stuff because IMO if the user wants it, they can add it explicitly.*

We can add it back in later if needed! Still not 100% happy with this component but we'll get there!


packages/api-client/src/views/Request/RequestSection/RequestBody.vue
```diff
@@ -30,25 +42,27 @@ const contentTypeOptions = {
...
-  yaml: 'yaml',
-  edn: 'edn',
-  other: 'html',
-} as const
-
 const codeInputLanguage = computed(() => {
-  const type = contentType.value as keyof typeof contentTypeToLanguageMap
-  return (contentTypeToLanguageMap[type] ?? 'plaintext') as CodeMirrorLanguage
+  const type = selectedContentType.value
+    .id as keyof typeof contentTypeToLanguageMap
```
amritk, line 64: weird flex eslint/prettier

amritk:
I need these changes for the next thing so im merging, can change anything later

PR #3046, opened 2024-09-05
fix(docusaurus): improve docusaurus memory issues

hwkr:
While not solving every issue this PR does improve the memory performance and DOM performance of the Scalar Docusaurus integration. Things it does do:

- Unmounts and removes the app when switching away from Scalar Docs in Docusaurus
- Checks to see if the Scalar CDN script has been loaded already rather than fetch a new copy every time
- Deletes the old configuration when refreshing the config (which hopefully will allow the browser to free up that memory)

Because our Docusaurus integration uses the CDN version of the app we will need to make sure we publish the new `standalone.js` before releasing the Docusaurus update (or at the same time?). I think we should be safe to publish the changes to `standalone.js` on their own, the only functionality change should be that the `scalar:reload-references` event will now try to mount to `script#api-reference` rather than always remounting to body.

I'm not sure if this is how we want to tackle all this @hanspagel @amritk feel free to weigh in. 

## Before

https://github.com/user-attachments/assets/8831f3fb-6c34-46b1-8294-2beb3594c45f

## After

https://github.com/user-attachments/assets/9da942f9-32ca-44ad-90d9-283c69cae833

### DIFF
--- a/packages/api-reference/src/standalone.ts
+++ b/packages/api-reference/src/standalone.ts
@@ -2,13 +2,14 @@
  * This file is the entry point for the CDN version of the API Reference.
  * It’s responsible for finding the spec and configuration in the HTML, and mounting the Vue.js app.
  */
+import type { ReferenceProps } from '@/types'
 import type { ReferenceConfiguration } from '@scalar/types/legacy'
 import { createHead } from '@unhead/vue'
 import { createApp, h, reactive } from 'vue'
 
 import { default as ApiReference } from './components/ApiReference.vue'
 
-const specScriptTag = document.getElementById('api-reference')
+const getSpecScriptTag = () => document.getElementById('api-reference')
 const specElement = document.querySelector('[data-spec]')
 const specUrlElement = document.querySelector('[data-spec-url]')
 const configurationScriptElement = document.querySelector(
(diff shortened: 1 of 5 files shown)


packages/api-reference/src/standalone.ts
```diff
@@ -164,16 +165,28 @@ if (!specUrlElement && !specElement && !specScriptTag) {
   document.addEventListener(
     'scalar:reload-references',
     () => {
-      // Check if elemenet has been removed from dom, and re-add
-      if (container && !document.body.contains(container)) {
-        document.body.appendChild(container)
+      // Check if element has been removed from dom, and re-add
+      if (!document.body.contains(container)) {
+        console.log('Re-adding container')
```
hanspagel, line 170: Can we omit the console.log here? 👀

packages/api-reference/src/standalone.ts
```diff
@@ -164,16 +165,28 @@ if (!specUrlElement && !specElement && !specScriptTag) {
...
+  // Allow user to destroy the vue app
+  document.addEventListener(
+    'scalar:destroy-references',
+    () => {
+      delete props['configuration']
+      app.unmount()
+    },
+    false,
+  )
+
```
hanspagel, line 189: This is great! Would you mind to add this to the documentation?

hanspagel (approved):
Great explanation! 

I don’t see any problems with the release. It should just work better once a new version of the CDN build is released, I think.

marclave (approved):

PR #3047, opened 2024-09-05
feat: new `ExampleResponses` embed

hanspagel:
This PR adds a new `ExampleResponses` embed (still not exposed) as a side effect of actually refactoring and cleaning up a beautiful mess:

* more comments, 
* replace code with existing helpers, 
* replace strange custom types with proper OpenAPI types

It’s also fixing a bug with `anyOf`, `allOf`, `anyOf` schemas showing a strange schema visualisation and the example response body at the same time, but in a broken layout.

When generating the example responses we’re taking those dedicated rules into account, and we have the `Show Schema` mode, so we don’t need an additional visualisiation which breaks the layout anyway.

**Preview**

Screenshot 2024-09-05 at 11 48 02


marclave (approved):
very happy with this

hwkr:
Opened a PR to display the headers in the responses section on the left #3052

PR #3049, opened 2024-09-05
feat(api-client): security schemes modal

antlio:
this pr adds security scheme button creation to auth combobox along auth modal in order to add security scheme to a collection:

https://github.com/user-attachments/assets/d43f9c1a-7e7c-4e23-b3dd-58d42e9a0d8a

some more refactor and polish to be added as i started by duplicating const type and flows but we might want to re-use `new-auth-options` ones.


hanspagel:
Let’s get this all in the super huge PR. 😊

hanspagel (approved):

PR #3051, opened 2024-09-05
chore(openapi-parser): use standard build tooling

tmastrom:
Leverage `@scalar/build-tooling` in the openapi parser package. 

This should go in after these types are updated https://github.com/scalar/scalar/pull/3028


hanspagel:
That’s so great! I think we should split this in multiple PRs, though. And before we merge any of them I need to test the tree shaking. I tried to use the build tooling and It made the tree shaking be less efficient.

Let’s start with a PR to use the same TS Config first maybe? I think that requires some changes to the code already.

tmastrom:
> That’s so great! I think we should split this in multiple PRs, though. And before we merge any of them I need to test the tree shaking. I tried to use the build tooling and It made the tree shaking be less efficient.
> 
> Let’s start with a PR to use the same TS Config first maybe? I think that requires some changes to the code already.

Good idea! Happy to break this work up for easy testing at each step. https://github.com/scalar/scalar/pull/3075

tmastrom:
Closing in favour of https://github.com/scalar/scalar/pull/3075

Issue #3054, opened 2024-09-06
scalar does not support arrays as query params


alewin:
### What happens?

🐛 Bug

When passing the following input: `http://localhost:3000?param=1&param=2`, Scalar only considers the last param value, resulting in the request being converted to `http://localhost:3000?param=2`

There is an ongoing discussion about this issue here: https://github.com/scalar/scalar/discussions/2835. As a first step, even without a UI to manage arrays, it would be helpful to have the ability to manipulate query parameters to handle arrays

### What did you expect to happen?

When passing the following input: `http://localhost:3000?param=1&param=2`, Scalar send the request without omitting the first `param` value, so `http://localhost:3000?param=1&param=2`

in this way a server application can handle the request as is, converting the `request.query.param` into `[1,2]` sends

### OpenAPI Document

https://sandbox.scalar.com/p/ZNf68

PR #3055, opened 2024-09-06
fix(api-client): arrays as query params

alewin:
**Problem**

Fix https://github.com/scalar/scalar/issues/3054

Before the fix proposal
Screenshot 2024-09-06 at 11 34 56

**Solution**
After the fix proposal
Screenshot 2024-09-06 at 11 34 16

**Comments**
this is the first step in handling arrays. in the future, one might think of supporting other ways of defining arrays, e.g. https://swagger.io/docs/specification/serialization/

### DIFF
--- a/.changeset/breezy-moons-grab.md
+++ b/.changeset/breezy-moons-grab.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
+
+avoid overwriting parameters in queries to support arrays as query params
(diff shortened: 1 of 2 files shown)


hanspagel:
Such a great PR! Unfortunately, we have this huge refactor in #2931, which we have to wait for, before we get this beauty in.

hanspagel:
Should pass CI once #3133 is merged and this PR here is rebased with it. :)

hanspagel (approved):
We rewrote the `sendRequest` method meanwhile, and it seemed to fix the issue, so this PR just adds a test to make sure it works. :) Just waiting for CI to finish before I merge it.

PR #3056, opened 2024-09-06
feat(api-reference): pass selected server to API client

bdpiprava:
- This PR set the selected server in api-reference to api-client

**Problem**
Currently, api-reference does not preserve the selected server in the main page to the `Test Request` modal. This is to set the `active server` url in the api-client

**Solution**
When retreiving the `activeExample` from workspace, set the activeServer url in the example url

https://github.com/user-attachments/assets/ace85b01-d6fa-4e0e-bb4e-2871d295ca44

### DIFF
--- a/packages/api-client/src/store/workspace.ts
+++ b/packages/api-client/src/store/workspace.ts
@@ -344,12 +344,27 @@ export const createWorkspaceStore = (router: Router, persistData = true) => {
     requestExampleMutators.delete(requestExample.uid)
   }
 
+  const withActiveServer = (requestExample: RequestExample): RequestExample => {
+    if (!activeServer.value) {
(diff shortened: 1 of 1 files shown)


bdpiprava:
@marclave  - Let me know if this makes sense.

hanspagel:
Thanks for looking into this! It might be that we fixed this already in one of the `on-hold` labelled PRs. I’ll check once we merged all those.

hanspagel:
Sorry, I have to close this PR. We have this super huge PR #2931, which includes this fix already. Thanks so much for taking the time to submit this PR! We’ll do better in the future.

bdpiprava:
Sure 👍

PR #3058, opened 2024-09-06
fix: logo light/dark mode

hanspagel:
I think we need to swap the logos.

Currently, it looks like this:

Screenshot 2024-09-06 at 15 21 20

Screenshot 2024-09-06 at 15 21 16


antlio (approved):
niceee

PR #3062, opened 2024-09-06
feat(themes): make base dark mode colors opaque

hwkr:
This fixes some issues with transparent theme colors sometimes causing inconsistent layering when placed over each other. 

For example icons that use a fill and an outline (main / this PR): 

Google Chrome-2024-09-06-14-08-43@2x
Google Chrome-2024-09-06-14-08-50@2x

This PR updates all of the base colors in the default theme (colors 1 through 3 and the border color) to be opaque. I can also update the rest of the themes and colors to all be opaque if that's something we want as well.

In order to match the existing look of the app I adjusted the border color slightly to be slightly darker in light mode and slightly lightly in dark mode. I think the change is pretty subtle?

## Before / Afters

These screenshots can be used to compare the before after if you don't feel like checking out the PR: [ComparisonScreenshots.zip](https://github.com/user-attachments/files/16912988/ComparisonScreenshots.zip)

This shows the references introduction card which is a good example of the border colors overlayed on several different background colors.

Google Chrome-2024-09-06-13-56-34

Google Chrome-2024-09-06-13-58-20

Here you can see a bunch of the pages one after the other:

https://github.com/user-attachments/assets/152cde2b-2020-45ce-85ce-22b39e11f313

### DIFF
--- a/.changeset/kind-grapes-lick.md
+++ b/.changeset/kind-grapes-lick.md
@@ -1,0 +1,5 @@
+---
+'@scalar/themes': patch
+---
+
+feat(themes): make base dark mode colors opaque
--- a/packages/themes/src/presets/default.css
+++ b/packages/themes/src/presets/default.css
@@ -10,21 +10,21 @@
   --scalar-color-3: #8e8e8e;
 
   --scalar-color-accent: #0099ff;
-  --scalar-border-color: rgba(0, 0, 0, 0.1);
+  --scalar-border-color: #dfdfdf;
 }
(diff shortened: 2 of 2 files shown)


antlio (approved):
great call, looks good to me!

hanspagel:
Love it! We should wait for approval from @cameronrohani. ✌️

cameronrohani (approved):

hanspagel (approved):
Ship it

PR #3063, opened 2024-09-07
fix(api-reference): server description color style

inyourtime:
Before:

https://github.com/user-attachments/assets/ff104f94-03ac-4c03-8929-77e53990d7b5

After:

https://github.com/user-attachments/assets/75c8fddf-67ef-4529-8586-0986cd87f13f


marclave (approved):
AWESOME fix @inyourtime

hanspagel (approved):
Great find and fix! 🤝

PR #3066, opened 2024-09-09
docs: fix broken link to laravel integration

hanspagel:
We’ve had two files for the Laravel integration, which I fixed, but I didn’t update the link in the root README it seems.


hanspagel:
The Markdown check fails in main, let’s skip the queue to get back to green. :)

PR #3072, opened 2024-09-09
chore(api-client): remove axios again

hanspagel:
WIP

**Tasks**

- [x] Make the tests run
- [x] Make the UI work
- [x] Fix related TS errors
- [x] allow non-proxy calls
- [x] Check whether we can abort a request
- [x] Setting server path works ex `https://galaxy.scalar.com/api`
- [x] fix variable replacement
- [x] add :colon variables
- [x] add error handling
- [x] make cookies work - not quite sure how to test cookies, but made it stop erroring when URL creation is failing

Then review, approve and merge :)

### DIFF
--- a/packages/api-client/src/libs/send-request.ts
+++ b/packages/api-client/src/libs/send-request.ts
@@ -311,6 +311,9 @@ export function createRequestOperation({
         cookieHeaderKeys: response.headers.getSetCookie(),
         data: responseData,
         duration: Date.now() - startTime,
+        method: request.method,
+        status: response.status,
+        path: pathname,
       },
     }
   }
--- a/packages/api-client/src/libs/send-request.ts
+++ b/packages/api-client/src/libs/send-request.ts
@@ -214,8 +214,7 @@ export function createRequestOperation({
   const pathname = replaceTemplateVariables(request.path, pathVariables)
   const urlParams = createFetchQueryParams(example, env)
   const headers = createFetchHeaders(example, env)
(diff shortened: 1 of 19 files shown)


packages/api-client/src/libs/send-request.ts
```diff
@@ -278,36 +290,25 @@ export function createRequestOperation({
...
-
-      headersToRemove
-        .map((header) => header.toLowerCase())
-        .forEach((header) => response.headers.delete(header))
-    }
-
+    const responseHeaders = normalizeHeaders(
+      response.headers,
+      shouldUseProxy(proxy, url.origin),
+    )
```
hanspagel, line 296: note to myself: make sure this is still in the code

packages/api-client/src/views/Request/Request.vue
```diff
@@ -66,39 +67,24 @@ const executeRequest = async () => {
...
+
+  // Send error toast
+  if (sendRequestError) toast(sendRequestError.message, 'error')
+  else requestHistory.push(result)
 }
 
 /** Cancel a live request */
-const cancelRequest = async () => requestAbortController.value?.abort()
+const cancelRequest = async () =>
+  requestAbortController.value?.abort('User has cancelled the request')
```
antlio, line 86: what do you think about that as the action is triggered by the user?

```suggestion
  requestAbortController.value?.abort('The request has been cancelled')
```

amritk: Sounds good to me!

PR #3080, opened 2024-09-11
feat: added error handling for empty + invalid urls

amritk:
Also centralized error messages and added a prettyError function to convert javascript error messages to human readable ones 👍🏾 

image


marclave:
@amritk i think a toast is fine for now, however @cameronrohani you mocked up the error state in the address bar we should add?

marclave (approved):

PR #3082, opened 2024-09-12
fix: non oauth security schemes

amritk:
This fixes security schemes and moves them onto the collection, then selection to the request, and restores the old UI

What this PR does not contain is Oauth fixes, thats a whole nother can of worms for tomorrow


.eslintrc.cjs
```diff
@@ -221,7 +221,7 @@ module.exports = {
         'vue/max-lines-per-block': [
           'warn',
           {
-            script: 100,
+            script: 150,
```
amritk, line 224: We can reduce it later 🫨

packages/api-client/src/libs/create-client.ts
```diff
@@ -128,8 +128,6 @@ export const createApiClient = ({
     })
   }
```
amritk, line 130: Removed since we have dataDump now

packages/api-client/src/libs/send-request.ts
```diff
@@ -264,18 +267,21 @@ export const createRequestOperation = <ResponseDataType = unknown>({
...
       if (scheme.type === 'apiKey' && exampleAuth.type === 'apiKey') {
         const value = replaceTemplateVariables(exampleAuth.value, env)
-        if (scheme.in === 'header') headers[scheme.nameKey] = value
-        if (scheme.in === 'query') urlParams.append(scheme.nameKey, value)
-        if (scheme.in === 'cookie')
-          Cookies.set(scheme.nameKey, value, cookieParams)
+        if (scheme.in === 'header') headers[exampleAuth.name] = value
+        if (scheme.in === 'query') urlParams.append(exampleAuth.name, value)
+        if (scheme.in === 'cookie') {
+          Cookies.set(exampleAuth.name, value)
```
amritk, line 281: Yea not sure what the plan is with cookies, but at least if you remove the params you can see them in dev tools (otherwise they don't show).

We should log em on the void server to test if they are working

packages/api-client/src/store/import-spec.ts
```diff
@@ -18,12 +17,6 @@ export function importSpecFileFactory({
   const importSpecFile = async (
     _spec: string | Record<string, any>,
     workspaceUid = 'default',
-    /**
-     * TODO: What do these look like?
-     * Ideally we reference some existing UIDs in the store and
-     * attach those as needed to entities below
-     */
-    overloadServers?: Spec['servers'],
   ) => {
```
amritk, line 20: We wont need this overload servers anymore, we will just update the ones on the collection

packages/api-client/src/views/Request/libs/auth.ts
```diff
@@ -0,0 +1,23 @@
+import {
+  type SecurityScheme,
+  securitySchemeExampleValueSchema,
+} from '@scalar/oas-utils/entities/spec'
+
```
amritk, line 5: Moved out some functions into another file and it still wasn't enough to hit the line limit 😓

.eslintrc.cjs
hanspagel: Funny how this increases with every PR. 😁 I think we should just remove this rule, while the intention is great, it’s not helpful in practice.

packages/api-client/src/libs/send-request.ts
hanspagel: Void Server should respond with cookies already: https://github.com/scalar/scalar/blob/main/packages/void-server/src/utils/getRequestData.ts#L44

It’s just that they are blocked for security reasons in most scenarios. 😢

packages/api-client/src/store/import-spec.ts
hanspagel: I think this is needed to overwrite what’s in the OpenAPI document through configuration. Same schema, but different place to add them.

hanspagel (changes requested):
Let’s make sure overwriting servers still works.

https://github.com/scalar/scalar/blob/main/documentation/configuration.md#servers-server

.eslintrc.cjs
amritk: haha its a good idea! I think just not right now. I would rather keep pushing it up than removing it though. It keeps us honest

packages/api-client/src/store/import-spec.ts
amritk: Going to rollback these changes and tackle it in a follow up PR

PR #3085, opened 2024-09-12
feat(components): add / update icons for dashboard

hwkr:
Adds and updates some icons for the Dashboard (and to better align them with the other line icons).

Also deleted some icons that weren't being used.

## Before

Google Chrome-2024-09-12-11-32-08@2x

## After

Google Chrome-2024-09-12-11-28-00@2x


amritk (approved):
Seems reasonable

hanspagel:
Let’s wait for @antlio, he cares a lot about our icons ✌️

antlio:
would that make sense to add the new ones [in figma first](https://www.figma.com/design/v6mJt1mK6UEh9nWJdzj2Fa/Client-WIP?node-id=3449-2&t=6YvCU0KoNe3hHthE-1)? or could be the time to extract them all in a global file too. (haven't got reviews on current proposal so haven't pursue a global update).

so we can make sure we use the right gap and increase visual balance between all icons. we can also merge this one and do this after.

hanspagel:
I’d say merge this one, but the work you proposed afterwards. :)

hanspagel (approved):

hwkr:
> nice ones!
> 
> would that make sense to add the new ones [in figma first](https://www.figma.com/design/v6mJt1mK6UEh9nWJdzj2Fa/Client-WIP?node-id=3449-2&t=6YvCU0KoNe3hHthE-1)? or could be the time to extract them all in a global file too. (haven't got reviews on current proposal so haven't pursue a global update).
> 
> so we can make sure we use the right gap and increase visual balance between all icons. we can also merge this one and do this after.

We can add those too! These ones are specifically for the dashboard though and I don't think that figma has anything similar to any of the icons I added here. We should sync on it next week :)

hanspagel:
I love every effort that is put into icons ✨

Issue #3065, opened 2024-09-08
`@scalar/fastify-api-reference` should expose an endpoint for the OpenAPI document


danwithabox:
### What happens?

I am using `@fastify/swagger` and am investigating swapping from `@fastify/swagger-ui` to `@scalar/fastify-api-reference`.

Scalar is much nicer, thank you for making it, but I've hit a roadblock:
Some of my tooling fetches the OpenAPI spec JSON file from a URL provided by `@fastify/swagger-ui`.

With `@scalar/fastify-api-reference`, it seems no such URL exists, despite https://github.com/scalar/scalar/pull/2947

Is there no endpoint I can fetch that would result in the spec JSON file?

### What did you expect to happen?

Be able to fetch the spec JSON for use in external tooling, but no endpoint exists to grab it from

hanspagel:
Ah, `@scalar/fastify-api-reference` does the rendering only. You can use `@fastify/fastify-swagger` to generate your OpenAPI document, and expose and an endpoint for it and pass that endpoint to `@scalar/fastify-api-reference`.

Does that help? I should make this more clear in the README.

danwithabox:
Cheers for the fast response, I will look into the suggestion.

I've mostly followed this documentation: https://github.com/scalar/scalar/tree/main/packages/fastify-api-reference
It does say 

> Actually, we’re picking it up automatically, so this would be enough:
> ```ts
> await fastify.register(require('@scalar/fastify-api-reference'), {
>   routePrefix: '/reference',
> })
>```

That made me assume it would be a drop-in replacement of `@fastify/fastify-swagger`, and it was, except for the download link issue.
I scrolled through https://www.npmjs.com/package/@scalar/fastify-api-reference for extra hints, to no avail.

Now though, I do see the NPM doc linking to this: https://github.com/scalar/scalar/blob/main/documentation/integrations/fastify.md#set-up-fastify-swagger-optional
Which in turn has this in its depths:
```ts
// Serve an OpenAPI file
fastify.get('/openapi.json', async (request, reply) => {
  return fastify.swagger()
})
```

Coming solely from `@fastify/swagger-ui`, I assure you this is the first time I saw `@fastify/fastify-swagger` used like that 😄

I would assume most people who come from `@fastify/swagger-ui` are in the same boat as me, and would save us digging through a few scattered docs and make the switch to Scalar trivial, if `@scalar/fastify-api-reference` would expose an endpoint by default.

[...]

PR #3098, opened 2024-09-13
feat: exposing JSON and YAML spec endpoints in `@scalar/fastify-api-reference`, mimicking `@fastify/swagger-ui`

danwithabox:
**Problem**
https://github.com/scalar/scalar/issues/3065

**Solution**
- OpenAPI spec exposed on endpoints, just like how `@fastify/swagger-ui` does it, e.g. by default:
  - `reference/json `
  - `reference/yaml`
- these endpoints are configurable
- added relevant test cases
- the download button in `@scalar/fastify-api-reference` now points to these URLs except when an URL is explicitly provided

**Notes**
Exposing the endpoints was trivial:
 - used `@scalar/openapi-parser` and its `@scalar/openapi-parser/plugins/fetch-urls` plugin to always have a spec no matter how it was provided (`spec.content`, `spec.url`, or `fastify.swagger()`) in both JSON and YAML formats
 - the endpoints are tested to make sure they are reliably there

However, making the download button work as I would find it intuitive - always pointing to the exposed endpoints, except when `spec.url` is passed in the config - brought up some stuff:

1. ❌ **ditched idea:** first, in `@scalar/fastify-api-reference`, I simplified `DownloadLink.vue` and made the relevant changes to not have client side generation of the download at all
    - `@scalar/fastify-api-reference` would always pass down `spec.url`, pointing to the configured JSON endpoint
    - 🎉 was super clean 
    - but broke one test - `"has the spec"` in `fastifyApiReference.test.ts`
      - it didn't like that when `spec.content` was passed, I rewrote it to `spec.url`
      - unsure if this would be acceptable conceptually
    - plus, I realized this is beyond the scope of what I wanted and won't touch it without input from the maintainers
      - because other packages would have to change and expose the endpoints too
      - I am guessing these ones, but maybe more:
        - `@scalar/api-reference-react`
        - `@scalar/express-api-reference`
2. ❌ **ditched compromise:** download button in `DownloadLink.vue` unaffected, just expose endpoints
    - I simply think this is annoying, I expect to be able to find the download URL from the button, when available, without scouring the documentation, just like I did when using `@fastify/swagger-ui`
    - and the button is already capable of this, so it felt like a waste to ignore it!
3. ✅ **chosen compromise:** pass `spec.url` too to `@scalar/fastify-api-reference`, alongside `spec.content`
    - the test `"has the spec"` mentioned in idea 1/ does not fail
    - uses the existing logic of `DownloadLink.vue`: if it sees an URL, it switches to the `<a href>` link mode instead of the client side generation mode
      - 🎉 the endpoint is now easily discoverable when using `@scalar/fastify-api-reference`!
      - 🎉 using `@scalar/api-reference` in other wrappers with other backends, like React with Express, etc, is unaffected
    - ⚠️ however, I had to add one line to `standalone.ts` in `@scalar/api-reference` to use `getSpecUrl()` even when `spec.content` is detected
      - this feels a bit unclean, having both content and URL may not be semantically desirable? 🤷‍♂️
      - but could not think of a way to avoid changing `standalone.ts`
      - and everything seems to work, tests pass, I did not find code relying on an assumption of only one of them existing

I didn't like 2/ at all, but 1/ felt out of scope for now, so I went with 3/ as a sensible compromise.

Please do chip in about this choice, I will change this if needed, but I would think eventually going for 1/ is the most desirable.
Always having an exposed endpoint by design, and pointing the download button to it in all cases would just be very clean, imo.

Feel free to reword the `openApiSpecEndpoint` config and the JSDoc of it in `fastifyApiReference.ts`. I thought about not making it configurable for simplicity, but I think it's better to grant that flexibility, right?

Also, this is my first PR for OSS, if I missed something do tell!
Enjoyed contributing, the repo was set up rather ergonomically :)

Cheers!

### DIFF
--- a/packages/oas-utils/package.json
+++ b/packages/oas-utils/package.json
@@ -27,8 +27,8 @@
     "lint:fix": "eslint .  --fix",
     "preview": "vite preview",
     "test": "vitest",
-    "typegen:migration": "vite-node ./src/migrations/generate-types.ts",
     "test:unit": "vitest .",
+    "typegen:migration": "vite-node ./src/migrations/generate-types.ts",
     "types:build": "scalar-types-build",
     "types:check": "scalar-types-check"
   },
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -88,7 +88,7 @@ importers:
         version: 12.3.2(typescript@5.5.2)
       ts-node:
         specifier: ^10.9.2
-        version: 10.9.2(@swc/core@1.5.29(@swc/helpers@0.5.5))(@types/node@20.14.10)(typescript@5.5.2)
+        version: 10.9.2(@swc/core@1.5.29)(@types/node@20.14.10)(typescript@5.5.2)
       tsc-alias:
         specifier: ^1.8.10
         version: 1.8.10
@@ -1491,15 +1491,24 @@ importers:
       '@fastify/basic-auth':
         specifier: ^5.1.1
         version: 5.1.1
+      '@fastify/swagger':
+        specifier: ^8.10.1
+        version: 8.14.0
       '@scalar/api-reference':
         specifier: workspace:*
         version: link:../api-reference
+      '@scalar/openapi-parser':
+        specifier: workspace:*
+        version: link:../openapi-parser
       '@vitest/coverage-v8':
         specifier: ^1.6.0
         version: 1.6.0(vitest@1.6.0(@types/node@20.14.10)(jsdom@24.1.0)(terser@5.31.1))
       fastify:
         specifier: ^4.26.2
         version: 4.28.0
+      github-slugger:
+        specifier: ^2.0.0
+        version: 2.0.0
       magic-string:
         specifier: ^0.30.8
         version: 0.30.10
@@ -1521,6 +1530,9 @@ importers:
       vue:
         specifier: ^3.4.29
         version: 3.4.31(typescript@5.5.2)
+      yaml:
+        specifier: ^2.4.5
+        version: 2.4.5
 
   packages/galaxy:
     devDependencies:
@@ -30366,7 +30378,7 @@ snapshots:
       strip-json-comments: 3.1.1
     optionalDependencies:
       '@types/node': 20.14.10
-      ts-node: 10.9.2(@swc/core@1.5.29(@swc/helpers@0.5.5))(@types/node@20.14.10)(typescript@5.5.2)
+      ts-node: 10.9.2(@swc/core@1.5.29)(@types/node@20.14.10)(typescript@5.5.2)
     transitivePeerDependencies:
       - babel-plugin-macros
       - supports-color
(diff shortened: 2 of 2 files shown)


hanspagel (changes requested):
I love approach 3! Thanks for the awesome explanation!

I think we can *not* pass the content and just use the URL in the fastify integration. Actually, I think having support for directly passing the content was a mistake. 👀 

And I'd prefer to have the endpoints to look like files:

* reference/openapi.json 
* reference/openapi.yaml

Wdyt?

packages/fastify-api-reference/src/fastifyApiReference.ts
```diff
@@ -15,15 +18,47 @@ export type FastifyApiReferenceOptions = {
...
+   * The specification is sourced from, in order of precedence:
+   * - `configuration.spec.content`
+   * - `configuration.spec.url` – fetched via `@scalar/openapi-parser/plugins/fetch-urls`
+   * - `@fastify/swagger` – if `configuration.spec` is not provided
+   *
+   * These endpoints can be used to fetch the OpenAPI specification for your own programmatic use.
+   *
+   * @default{ json: '/json', yaml: '/yaml' }
+   */
+  openApiSpecEndpoint?: {
```
hanspagel, line 44: and can we name this `openApiDocumentEndpoints` (my fav), `openApiDocuments` or `openApiFiles`?

packages/oas-utils/package.json
```diff
@@ -101,7 +101,7 @@
     "@scalar/themes": "workspace:*",
     "@scalar/types": "workspace:*",
     "nanoid": "^5.0.7",
-    "yaml": "^2.4.5",
+    "yaml": "^2.5.1",
```
hanspagel, line 104: Can we stick to the other version? I think it’d increase the bundle size otherwise because one dependency is pinned to the other version.

danwithabox: I didn't look around enough to have an input on that, and it's not like this _needs_ to be 2.5.1, so 2.4.5 it is 👍

packages/fastify-api-reference/src/fastifyApiReference.ts
danwithabox: renamed to `openApiDocumentEndpoints` 👍

danwithabox:
YAML version change and renames done.

Endpoints changed to:
- reference/openapi.json
- reference/openapi.yaml

Now passing URL to `@scalar/fastify-api-reference`, always:
 - change `standalone.ts` in `@scalar/api-reference` is reverted, no longer processing both `.url` and `.content`, just like before
 - relevant tests reworked, added a test to check whether the explicitly passed URL is respected and not getting replaced by the exposed endpoint URL

Misc:
- fixed some eslint warnings

I think that's all the requested changes! 🎉

hanspagel:
Best PR ever!

For some reason I can’t check out the PR, I think there’s something wrong on my machine. Once you/me/someone is able to fix the merge conflict, we’ll get this in ASAP.

hanspagel (approved):

danwithabox:
Somehow the merge conflicted a completely safe change to `pnpm-lock.yaml` on origin, never seen this happen 😵‍💫 
Resolved it, synced my PR to origin/main, good to go in theory 👍

hanspagel:
No worries. Actually, I don’t know what’s going on exactly, but I just want this PR in and we can further improve it in a separate PR. 😁

hanspagel:
It’s in! :) Should be released later today, let’s see.

Thanks for this great PR, huge help to improve the fastify integration! 🙏

danwithabox:
Cheers! If it ever becomes a goal to abolish the client-generated downloads en bloc, or any issue opens around Vue / TS / Fastify, I'm open for hire 😁

Can't wait for the release, can finally fully ditch Swagger UI 🥳

PR #3099, opened 2024-09-13
docs: Tidy of NuGet package metadata

timheuer:
- Added icon (using repo org icon)
- Added existing readme for aspnet to package
- Added description for package (using tag line from project)
- Modified author to be 'Scalar' (not sure about this one versus keeping marclave the author? I leaned on using the org here as the author metadata)

**Problem**
Currently when searching packages for Scalar, the 'official' package isn't differentiated in author/brand/description compared to others:
image

**Solution**
With this PR it just modifies the NuGet package metadata to provide better annotation which also results in better branding where packages are displayed (in products, on NuGet package pages)
image


marclave (approved):
this is AWESOME @timheuer ✨ 

i agree on all the above

PR #3106, opened 2024-09-15
feat: getExampleFromSchema can generate examples for anyOf schemas, fix #2417

### LINKED ISSUE #2417: `requestBody` with `anyOf` defined: no example is generated for the body
### What happens?

Screenshot 2024-07-07 at 15 18 32

An example should be generated from the first schema in the `anyOf` array

1. Go to [https://sandbox.scalar.com/e/Bj8Ak](https://sandbox.scalar.com/e/Bj8Ak)
2. Create a request where the schema body has the `anyOf` defined, even if it has just one item in the array
3. The body example is no more shown

### OpenAPI Document

https://sandbox.scalar.com/e/Bj8Ak

kevinand11:
**Problem**
Examples are not generated for schemas with `anyOf`

Currently, I used some auto gen tools to generate json schemas, and it creates unions with `anyOf` instead of `oneOf`
Examples are generated fine for `oneOf` schemas, but the `anyOf` case is completely ignored

**Solution**
With this PR, I create an implementation for the `anyOf` that is similar to the `oneOf`'s implementation


hanspagel (changes requested):
For some reason the Git history is screwed up. 🤔 Did you clone and push or fork?

So there’s a three things that’s stopping me from merging it right-away:

- [ ] The Git history seems to be off (that might be our fault), so you probably need to create a new fork
- [ ] There’s not test for the new behaviour
- [ ] There’s no changeset (run `pnpm changeset` in root)

Would you be able to work on this? Otherwise I’ll bring it over the finish line, but that will take some time. ✌️

kevinand11:
> Would you be able to work on this? Otherwise I’ll bring it over the finish line, but that will take some time. ✌️

I will attend to it later in the day

PR #3107, opened 2024-09-16
feat(api-client): response codeblock copy

antlio:
this pr adds copy capability to response code in api client as seen below:

image


hanspagel (approved):
Desperately needed

amritk (approved):
Nice!

PR #3112, opened 2024-09-16
chore: remove max-lines-per-block rule

hanspagel:
I’d love to have less lines in Vue components, but the rule doesn’t work well in practice. It just adds squiggly lines to the whole file and instead of nudging me to refactor the component it just makes me close it faster. Also, red squiggly lines are really hard to spot then.

So let’s make our components have less lines, but let’s also remove this rule for now!


amritk (approved):
Yea we can take it out as long as we all keep it in mind, and would like to bring it back one day!

PR #3113, opened 2024-09-16
fix: request example code works for some clients only, fix #3094

### LINKED ISSUE #3094: JavaScript fetch example code can't be generated
### What happens?

The "JavaScript fetch" version of the code can't be generated. The "Shell cURL" works as expected.

image

image

### Additional info

All of the other JavaScript flavours produce the same output. E.g. XHR, jQuery have the same output

image

Correctly generate the example code for the JavaScript flavours.

### OpenAPI Document

_No response_

hanspagel:
We broke the example code generation for most clients in #2995.

This PR fixes multiple issues causing the example code generation for most clients, and even deletes a dead code block.

We just didn’t pass the client id in some cases and we added an `js` alias to the `@scalar/snippetz` call, but not to the `httpsnippet-lite` call where it’s actually needed.

### DIFF
--- a/.changeset/dirty-crabs-give.md
+++ b/.changeset/dirty-crabs-give.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
+fix: request example code works for some examples only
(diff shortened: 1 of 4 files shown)


amritk (approved):
lgtm!

PR #3114, opened 2024-09-16
feat: render deeply nested schemas, fix #2337

### LINKED ISSUE #2337: Deep schema is not fully shown in example responses
### What happens?

I have a somewhat unusual, very deeply nested schema and noticed that everything below level five in the hierarchy is displayed as `null` in the example response. A limit of five suggests that this was a deliberate decision (not a bug, but a feature). Is this something which can be configured somewhere?

Example response goes all the way down 😉 

See [sandbox](https://sandbox.scalar.com/e/lZPDb). `level6` underneath `theAbyss` is rendered as `null`, but should be another object containing a property `foo` of type `number`. The actual schema is shown correctly when switching to it.

### OpenAPI Document

https://sandbox.scalar.com/e/lZPDb

hanspagel:
Currently, we just don’t render schemas more than 5 levels deep.

This PR renders deeply nested schemas, no matter how deeply they are nested.

But circular references are detected and only the first five levels are rendered in that case.

See #2337 for more context.

### DIFF
--- a/.changeset/smooth-ties-dress.md
+++ b/.changeset/smooth-ties-dress.md
@@ -1,0 +1,6 @@
+---
+'@scalar/api-reference': patch
+'@scalar/oas-utils': patch
+---
+
+feat: render deeply nested schemas, but not circular references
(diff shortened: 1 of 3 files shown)


amritk (approved):
noice

PR #3115, opened 2024-09-16
fix: whitespace in galaxy/package.json

hanspagel:
Some scripts seem to modify the line ending of this one package.json and the change comes up in every branch.

Hope fully this PR resolves the issue and doesn’t come up again.


hanspagel:
That’s not it.

PR #3116, opened 2024-09-16
feat: getExampleFromSchema can generate examples for anyOf schemas, fix #2417

### LINKED ISSUE #2417: `requestBody` with `anyOf` defined: no example is generated for the body

kevinand11:
**Problem**
Examples are not generated for schemas with `anyOf` when `type` is not defined on the object

Currently, I used some auto gen tools to generate json schemas, and it creates unions with `anyOf` instead of `oneOf`
Examples are generated fine for `oneOf` schemas, but the `anyOf` case is completely ignored

**Solution**
With this PR, I create an implementation for the `anyOf` that is similar to the `oneOf`'s implementation

### DIFF
--- a/.changeset/thick-monkeys-shout.md
+++ b/.changeset/thick-monkeys-shout.md
@@ -1,0 +1,5 @@
+---
+'@scalar/oas-utils': patch
+---
+
+Generate examples for schemas with `anyOf` defined when `type` is not defined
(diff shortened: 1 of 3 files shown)


kevinand11:
@hanspagel Reopened from https://github.com/scalar/scalar/pull/3106

PR #3095, opened 2024-09-13
fix: prepare api client for dashboard

cameronrohani:
we won't be having any sidenav's across the app with the new dashboard, so I've had to make quite a lot of changes to remedy this

image

image

Also made a small handful of visual tweaks

PR #3117, opened 2024-09-16
feat(api-client): context bar

antlio:
this pr cherry picks context bar position and style updated from #3095 as seen below:

image


amritk (approved):
👍🏾

PR #3119, opened 2024-09-16
fix: eslint does not run

hanspagel:
WIP strange TS error

ESLint was failing in multiple packages, this PR makes it run again. It doesn’t fix any linter warnings, just a few hard errors, but mostly it just adds files to the TSConfig, that were already in the ESLint scope (which triggers an error otherwise).


hanspagel:
Lol too much work

PR #3120, opened 2024-09-16
chore: fix some linter warnings

hanspagel:
This PR fixes just some linter warnings:

* ESLint fixes some things automatically
* replaced @ts-ignore with @ts-expect-error
* deleted some dead imports/variables


hanspagel:
Should be safe to go (famous last words). 😬

PR #3121, opened 2024-09-16
feat(api-client): code input foldable

antlio:
this pr adds fold capability to api client code input as seen below:

image

### DIFF
--- a/packages/use-codemirror/src/hooks/useCodeMirror.ts
+++ b/packages/use-codemirror/src/hooks/useCodeMirror.ts
@@ -15,6 +15,7 @@ import {
   type StreamLanguage,
   bracketMatching,
   defaultHighlightStyle,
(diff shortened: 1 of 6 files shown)


packages/use-codemirror/src/hooks/useCodeMirror.ts
```diff
@@ -397,9 +428,6 @@ function getCodeMirrorExtensions({
     extensions.push(jsonLinter)
   }
 
-  // Line numbers
-  if (lineNumbers) extensions.push(lineNumbersExtension())
```
hanspagel, line 401: accident?

antlio: changed the order to have the marker prepending the line numbers, otherwise it appends it

Issue #2918, opened 2024-08-21
Test Request with "Form URL Encoded" is sent with Content-Type: "multipart/form-data"


yvzn:
### What happens?

When setting "Form URL Encoded" for the type of body in the "Test Request" window, the request is sent to the server with Content-Type: "multipart/form-data; boundary=...." instead of "application/x-www-form-urlencoded"

Despite the fact that the "Headers" section displays "Content-Type": "application/x-www-form-urlencoded"

### What did you expect to happen?

The request to the server is sent with "Content-Type": "application/x-www-form-urlencoded"

### How can we reproduce the issue?

1. Go to https://sandbox.scalar.com/e/_wHkn
2. Click on "Test Request" of the first route
3. Select "Form URL Encoded" in the Body section
4. Click on "Send"
5. Inspect the request in the DevTools, it is sent with the wrong content-type

### OpenAPI Document

https://sandbox.scalar.com/e/_wHkn

yshrsmz:
@hanspagel Hi, I still see the same issue.
See https://sandbox.scalar.com/e/M67H7 for the repro. reproduction step is the same as the OP.

PR #3125, opened 2024-09-17
fix(api-reference): placeholders in the license header not replaced

yshrsmz:
**Problem**
While playing around with #2918, I found that the placeholders in [the license header](https://github.com/scalar/scalar/blob/main/packages/api-reference/license-banner-template.txt) were not correctly replaced.

**Solution**
This PR fixes the regular expressions to match the placeholder's format.


hanspagel (approved):
Dang! I saw this a dozen times and never looked into it. Thanks for finally fixing this! 🙏

PR #3126, opened 2024-09-17
feat(api-client): address bar http methods

antlio:
this pr updates http methods style and ordering in the address bar:

**before**
image

**after**
image

### DIFF
--- a/packages/oas-utils/src/helpers/httpMethods.ts
+++ b/packages/oas-utils/src/helpers/httpMethods.ts
@@ -1,5 +1,9 @@
 import type { RequestMethod } from '@/entities/spec/requests'
 
+/**
+ * HTTP methods in a specific order
+ * Do not change the order
+ */
 export const REQUEST_METHODS: {
   [x in RequestMethod]: {
     short: string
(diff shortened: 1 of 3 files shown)


hanspagel (approved):
Love it! @cameronrohani should take a look, too

packages/oas-utils/src/helpers/httpMethods.ts
```diff
@@ -1,58 +1,60 @@
 import type { RequestMethod } from '@/entities/spec/requests'
 
-export const REQUEST_METHODS: {
-  [x in RequestMethod]: {
-    short: string
-    color: string
+/** HTTP methods in a specific order */
+const ORDERED_METHODS: RequestMethod[] = [
```
amritk, line 4: I would just order these in entitites/spec/requests and put on a comment that they should not be re-ordered. Then we only have one source of truth

packages/oas-utils/src/helpers/httpMethods.ts
```diff
@@ -1,42 +1,21 @@
 import type { RequestMethod } from '@/entities/spec/requests'
 
+/**
+ * HTTP methods in a specific order
+ * Do not change the order
+ */
```
antlio, line 6: changing order in requests.ts won't impact the order to i changed it here + comment and kept previous code along @amritk

antlio: yes we kinda do as we add properties per method, but could be refactored

cameronrohani (approved):

amritk (approved):

PR #3145, opened 2024-09-19
fix(api-client): path regex helper

antlio:
this pr fixes the current path variable regex that display also the request path table for environment variables by setting up a path regex helper.

### DIFF
--- a/.changeset/early-geese-impress.md
+++ b/.changeset/early-geese-impress.md
@@ -1,0 +1,6 @@
+---
+'@scalar/api-client': patch
+'@scalar/oas-utils': patch
+---
+
+fix: adds path regex helper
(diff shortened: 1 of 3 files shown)


amritk:
hey @antlio can we use the one from send-request? or does that one not work here?

https://github.com/scalar/scalar/blob/main/packages/api-client/src/libs/string-template.ts#L16

antlio:
> hey @antlio can we use the one from send-request? or does that one not work here?
> 
> 
> 
> https://github.com/scalar/scalar/blob/main/packages/api-client/src/libs/string-template.ts#L16

just want to get the one curly braces one to be displayed in the path table. i'll look but mot sure

marclave (approved):
😎

PR #3154, opened 2024-09-20
fix: respect the proxy url config

marclave:
we werent respecting the proxy config, lets fix it! :) 

the scalar api client still has the proxy by default


amritk (approved):
Yolo

Issue #2741, opened 2024-08-01
search docs not jumping to the correct tag the 1st time


tnrich:
### What happens?

When I search for some tag in the search box, then click or hit enter, the first time the docs do NOT jump to the desired place: 
https://github.com/user-attachments/assets/364990b8-0b5c-47aa-9ae9-b4f28547517b

However, this does open the tag in the left panel: 
![image](https://github.com/user-attachments/assets/339ab676-732f-4369-af33-f9d6d310d1c9)

Repeating the same steps again, searching for the same tag, does work the 2nd time, once the tag is already open in the left panel: 

https://github.com/user-attachments/assets/741d0afb-5e1a-453c-8a2e-92a626fafdd5

### What did you expect to happen?

I would have expected the docs page to jump to the correct tag the first time. 

### How can we reproduce the issue?

Go to https://sandbox.scalar.com/p/pchA0

and repeat the steps above as described. Thanks!

### OpenAPI Document

https://sandbox.scalar.com/p/pchA0

PR #3160, opened 2024-09-20
fix(api-reference): search results scroll to wrong position

Mason-Little:
**Problem**
when using search function inside of the references there is a race condition that causes the search to sometimes not jump to the correct endpoint the first time it is searched, highlighed in this [issue](https://github.com/scalar/scalar/issues/2741)

**Solution**
This PR changes the function `onSearchResultClick` inside of `SearchModal.vue` Now it uses a `MutationObserver` to wait until the target section is loaded in the DOM before navigating to it. This ensures that the docs scroll to the correct section right away when a user clicks a search result.

### DIFF
--- a/.changeset/stupid-bananas-begin.md
+++ b/.changeset/stupid-bananas-begin.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
+fix: search results scroll to wrong position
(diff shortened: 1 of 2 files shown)


amritk (changes requested):
Alright so this will work and is how we used to do it BUT turns out its more performant to use our `lazyBus` which already exists. Now this will only work if the item we are waiting for is wrapped in lazy, but if it is we can use this

https://github.com/scalar/scalar/pull/2858/files#diff-177327574212cd0b7483ff6ea4da60e444da454855f260a9a5a2768b7a83a297R333

If the items don't use lazy already then we can use this

amritk (approved):
Nice work!!

PR #3161, opened 2024-09-20
feat(postman-to-openapi): new postman-to-openapi package

Mason-Little:
Problem
Currently, there is no Postman spec to OpenAPI converter for our references

Solution
With this PR, I created a single-file TypeScript converter package in API references that allows us to convert any Postman spec JSON to OpenAPI JSON


hanspagel:
This is awesome! I just moved the files around a little bit and added a README. I’ve also added one test, which fails for the Postman Collection example hosted on GitHub: https://github.com/postmanlabs/postman-collection/blob/develop/examples/collection-v2.json

Would you mind to look into this?

packages/postman-to-openapi/rollup.config.ts
```diff
@@ -0,0 +1,12 @@
+import { createRollupConfig } from '@scalar/build-tooling'
+import type { RollupOptions } from 'rollup'
+
+const options: RollupOptions = {
+  input: ['./src/index.ts'],
+  ...createRollupConfig({
+    // Needed to enable explicit TypeScript (which Vite does not need)
+    typescript: true,
+  }),
+}
```
geoffgscott, line 10: No destructuring needed
```suggestion
const config = createRollupConfig({
  options: {
    input: ['src/index.ts'],
  },
  typescript: true,
})
```

packages/postman-to-openapi/src/fixtures/collection-v2.json
geoffgscott, line 1: Test fixtures should go in a `test` folder adjacent to the `src` folder. This allows easy TSConfig exclusion of all test files.

packages/postman-to-openapi/env.d.ts
geoffgscott, line 1: We should not be using raw imports unless we using Vite in which case the vite shims will manage this.

packages/postman-to-openapi/src/convert.ts
```diff
@@ -0,0 +1,771 @@
...
+
+function replacePostmanVariables(collectionFile: any): any {
+  return collectionFile
+}
+/**
+ * Converts a Postman collection to an OpenAPI specification.
+ *
+ * @param {string | Record<string, any>} collectionContent - The Postman collection content, either as a JSON string or an object.
+ * @param {Options} [options={}] - Configuration options for the conversion process.
+ * @returns {Promise<string>} A promise that resolves to the OpenAPI specification as a JSON string.
```
geoffgscott, line 24: We don't need JSDocs for anything as we can directly annotate parameters using typescript.

packages/postman-to-openapi/src/convert.ts
```diff
@@ -0,0 +1,771 @@
...
+  if (typeof collectionContent === 'string') {
+    collectionFile = JSON.parse(collectionContent)
+  } else {
+    collectionFile = collectionContent
+  }
+  if (replaceVars) {
+    collectionFile = replacePostmanVariables(collectionFile)
+  }
+  const postmanJson = collectionFile.collection || collectionFile
+  const { item: items = [], variable = [] } = postmanJson
```
geoffgscott, line 54: Please do a generator function pattern here and not `let` reassignment with added destructuring

packages/postman-to-openapi/src/convert.ts
```diff
@@ -0,0 +1,771 @@
...
+  const paths: Record<string, any> = {}
+  const domains = new Set<string>()
+  const tags: Record<string, any> = {}
+  const securitySchemes: Record<string, any> = {}
+
+  for (let i = 0; i < items.length; i++) {
+    let element = items[i]
```
geoffgscott, line 61: Prefer `items.forEach`

packages/postman-to-openapi/src/convert.ts
```diff
@@ -0,0 +1,771 @@
...
+    while (element?.item != null) {
+      const { item = [], description: tagDesc = '' } = element
+      const tag = calculateFolderTag(element, folders)
+      const tagged = item.map((e: any) => ({ ...e, tag }))
+      tags[tag] = tagDesc
+      items.splice(i, 1, ...tagged)
+      element = tagged.length > 0 ? tagged.shift() : items[i]
+    }
```
geoffgscott, line 69: While loops can be difficult to follow and can have undefined exit conditions. 
We are also mutating the array we are iterating over here (items) which will cause unintended side effects (and is a bad code smell).

packages/postman-to-openapi/src/convert.ts
```diff
@@ -0,0 +1,771 @@
...
+        response,
+      } = element
+      const {
+        url,
+        method,
+        body,
+        description: rawDesc,
+        header = [],
+        auth,
+      } = request
```
geoffgscott, line 85: This feels like excessive destructuring. nothing wrong with using `element.name` which in a large function block like this will be more clear anyways.

packages/postman-to-openapi/src/convert.ts
```diff
@@ -0,0 +1,771 @@
...
+    ...parseExternalDocs(variable, externalDocs),
+    ...parseServers(domains, servers),
+    ...parseAuth(postmanJson, optsAuth, securitySchemes),
+    ...parseTags(tags),
+    paths,
+  }
+
+  return JSON.stringify(openApi, null, 4)
+}
+/* Helper Functions */
```
geoffgscott, line 129: Incorrect comment syntax for a function. If you want a module brake using the VSCode snippet `//`

geoffgscott: This module file is also getting quite long. Would suggest creating some separate module files to group some of these logically. All of the below functions need unit tests.

packages/postman-to-openapi/src/convert.ts
```diff
@@ -0,0 +1,771 @@
...
+    ? { externalDocs: { url, ...(description ? { description } : {}) } }
+    : {}
+}
+
+function parseBody(body: any = {}, method: string): any {
+  if (['GET', 'DELETE'].includes(method)) return {}
+  const { mode, raw, options = { raw } } = body
+  let content: any = {}
+  switch (mode) {
+    case 'raw': {
```
geoffgscott, line 223: No `switch` and no `let`. In the scalar codestyles you can see examples that use the early return pattern which is much easier to follow. 

```ts
if(conditionOne) return dataFormOne
if(conditionTwo) return dataFormTwo
```

packages/postman-to-openapi/src/convert.ts
```diff
@@ -0,0 +1,771 @@
...
+function parseFormData(data: any[]): any {
+  const objectSchema: any = {
+    schema: {
+      type: 'object',
+    },
+  }
+  return data.reduce((obj: any, { key, type, description, value }: any) => {
+    const { schema } = obj
+    if (isRequired(description)) {
+      ;(schema.required = schema.required || []).push(key)
```
geoffgscott, line 293: This is hard to parse syntax. This can be broken into two lines or done differently.

packages/postman-to-openapi/src/convert.ts
```diff
@@ -0,0 +1,771 @@
...
+    }
+    ;(schema.properties = schema.properties || {})[key] = {
```
geoffgscott, line 295: Same as above comment

packages/postman-to-openapi/src/convert.ts
```diff
@@ -0,0 +1,771 @@
...
+): any {
+  const parameters = [
+    ...Array.from(
+      Array.isArray(header) ? 
+        header.reduce(
+          mapParameters('header', includeHeader, paramInserter),
+          new Map(),
+        ) : [] // If not an array, default to an empty array
+    ),
+  ]
```
geoffgscott, line 327: This needs some clean up. To hard to follow.

packages/postman-to-openapi/src/convert.ts
```diff
@@ -0,0 +1,771 @@
...
+  header: any[] = [],
+  paths: string,
+  paramsMeta: any = {},
+  pathVars: any = {},
+  { includeQuery = false, includeHeader = false }: any = {},
+  paramInserter: (
+    parameterMap: Map<string, any>,
+    param: any,
+  ) => Map<string, any> = defaultParamInserter,
+): any {
```
geoffgscott, line 318: No ts `any`

packages/postman-to-openapi/src/convert.ts
```diff
@@ -0,0 +1,771 @@
...
+}
+
+function mapParameters(
+  type: string,
+  includeDisabled: boolean,
+  paramInserter: (
+    parameterMap: Map<string, any>,
+    param: any,
+  ) => Map<string, any>,
+): (parameterMap: Map<string, any>, param: any) => Map<string, any> {
```
geoffgscott, line 345: No ts `any` unless there is a very strong reason and it's well commented.

Mason-Little:
Marc Amrit and I decided it would be better to restart after syncing on this PR

PR #3162, opened 2024-09-20
feat(api-client-app): sets load analytics spa to auto

antlio:
this pr is a first attempt to get real time analytics within the api client app.


hanspagel (approved):
Let’s try this!

PR #3169, opened 2024-09-20
fix(Scalar.AspNetCore): ForceThemeMode is null by default

### LINKED ISSUE #3156: C# package missing some methods
### What happens?
1. options method missing from example [program.cs](https://github.com/scalar/scalar/blob/main/packages/scalar.aspnetcore/playground/Scalar.AspNetCore.Playground/Program.cs)

image

2. the MapScalarApiReference option darkMode not set dark mode as initial theme mode, wether i set it true / false.

also not working using the builder pattern
image

C# .NET 9 using scalar package

should be dark mode at initial page opening
image

### OpenAPI Document

_No response_

xC0dex:
**Problem**
The `ForceThemeMode` property was set to `ThemeMode.Light` by default, which caused properties like `DarkMode` to have no effect.

**Solution**
The default value is now `null` as intended.

Closes #3156


marclave (approved):
pause-shrek

PR #3170, opened 2024-09-20
chore: Add Nuget to dependabot configuration

xC0dex:
### Description

This PR adds the NuGet package manager to the dependabot configuration. The `Scalar.AspNetCore` project itself has only one dependency, but the demo and test projects have some more. I think weekly updates are fine as the dependencies are not crucial.


marclave (approved):
nice @xC0dex ✨

PR #3175, opened 2024-09-20
feat(api-client): performance upgrades for sidebar

amritk:
This improves the load time of the api-client by about 25x when looking at big specs. This was caused by the context menu + menu + modals in every instance of the sidebar items. We should aim to keep looping items as lean as we can!

Before:
sidebar load time: 2629.917236328125 ms

After:
sidebar load time: 110.170166015625 ms

@antlio I had to remove the context menu because I couldn't figure out how to pull it out of the ReqeuestSidebarItem and into the parent. Maybe you can help with that.

TODO:

- [x] accessibility/focus with arrow keys
- [x] close menu when clicking anywhere
- [x] close menu when hitting escape
- [x] Still need to figure out why the menu jumps (due to the button disapearing)
- [x] switch the bus to a ref and use the menu button wrapper

### DIFF
--- a/.changeset/brown-peaches-bathe.md
+++ b/.changeset/brown-peaches-bathe.md
@@ -1,0 +1,6 @@
+---
+'@scalar/api-client': patch
+'@scalar/components': patch
+---
+
+chore: performance upgrades for api client sidebar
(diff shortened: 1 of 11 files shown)


marclave (approved):
wagmi

antlio:
awesome!

PR #3176, opened 2024-09-20
fix: remove trailing slash in examples and in client for server

marclave:
**Problem**
Currently we have `//` if a server has a trailing slash

**Solution**
(we remove it from rendering :) )

### DIFF
--- a/packages/api-client/src/components/AddressBar/AddressBarServer.vue
+++ b/packages/api-client/src/components/AddressBar/AddressBarServer.vue
@@ -57,16 +57,23 @@ const handleAddServer = () =>
   commandPaletteBus.emit({
     commandName: 'Add Server',
   })
+
+const serverUrlWithoutTrailingSlash = computed(() => {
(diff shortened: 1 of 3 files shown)


packages/api-client/src/components/AddressBar/AddressBarServer.vue
```diff
@@ -57,16 +57,26 @@ const handleAddServer = () =>
   commandPaletteBus.emit({
     commandName: 'Add Server',
   })
+
+const serverUrlWithoutTrailingSlash = computed(() => {
+  if (activeServer.value?.url?.endsWith('/')) {
+    return activeServer.value.url.slice(0, -1)
+  }
+  return activeServer.value?.url || ''
+})
```
amritk, line 66: nit: slightly cleaner if u like

```suggestion
const serverUrlWithoutTrailingSlash = computed(() => activeServer.value?.url?.replace(/\/^/, '') || '')
```

amritk (approved):
Send it!

Issue #3182, opened 2024-09-21
UI is broken in `@scalar/hono-api-reference` on `v0.5.147`


omarkhatibco:
### What happens?

Hi, 
This has been reported on other packages as well and it seems it was fixed, but unfortunaly not in hono package I assume.

I tried to revert back to `v0.5.146` and `v0.5.146` but there were broken as well.

see the sidebar or tri to click test request, you should see a blank screen.

### What did you expect to happen?

the test request screen should work

omarkhatibco:
a quick fix for now is to pin the scalar version to `1.25.11` 

```ts
app.get(
	'/swagger',
	apiReference({
		cdn: 'https://cdn.jsdelivr.net/npm/@scalar/api-reference@1.25.11',
		spec: {
			url: '/openapi.json',
		},
	}),
)

```

NotReallyPrince:
![image](https://github.com/user-attachments/assets/f91685e3-fff6-4882-bcc9-d77e01fd8ddb)

marclave:
hey @omarkhatibco @NotReallyPrince wouod you be able to share a link to your docs or your openapi spec? :) 

can also dm it to me marc@scalar.com

omarkhatibco:
Hi @marclave,

Just sent you an email with the links, thanks for your help

NotReallyPrince:
Hi @marclave,

Just sent you an email with the links, thanks for your help

Issue #3165, opened 2024-09-20
@scalar/api-reference v1.25.12 and v1.25.13 cause the UI to crash


subzero10:
### What happens?

I am using Scalar.AspNetCore. Since the release of 1.25.12 (yesterday), the UI is crashing with the following error:
```
Failed to fetch spec from URL: ZodError: [
  {
    "code": "invalid_type",
    "expected": "string",
    "received": "array",
    "path": [
      "type"
    ],
    "message": "Expected string, received array"
  }
]
    at get error (standalone.js:18441:10)
    at w8.parse (standalone.js:18477:64)
    at standalone.js:19266:13
    at standalone.js:19264:54
    at Array.forEach (<anonymous>)
    at C6 (standalone.js:19264:29)
    at standalone.js:19373:36
    at Array.forEach (<anonymous>)
    at Y6 (standalone.js:19373:15)
    at async l (standalone.js:19537:61)
```
I traced it to the fact that we have GET requests with query parameters that are not plain strings - for example I have a query parameter that is mapped to a list of strings (List<string>).

Have you considered allowing the app to continue without crashing, even if Zod validation fails (at least for the production environment)? 

### What did you expect to happen?

The UI should not crash. I know that complex types as query parameters may not be supported out of the box, but even in that case it shouldn't break the whole tool.

amritk:
Hey @subzero10 do you think you could pass me the spec if its public? Or just  mock one up with those parameters

marclave:
@subzero10 

i tried to create a path parameter with an array type and it worked, if you can share a small snippet of your spec we can fix :) 

```
        "parameters": [
          {
            "name": "petId",
            "in": "path",
            "description": "ID of pet to update",
            "required": true,
            "schema": { "type": "array" }
          }
        ],
```

<img width="687" alt="image" src="https://github.com/user-attachments/assets/3ba4d608-ffe5-4b88-973a-6f0300b84c23">

omarkhatibco:
Hi @amritk ,

I could share with you my api ui if thats helps, it's also broken on the latest version.

url: https://api.mando.cx/swagger#tag/auth

I'm using `@scalar/hono-api-reference v 0.5.147`

marclave:
fixed in 1.25.17 :)

subzero10:
Hey @marclave, 1.25.17 seems to have improved things - the UI does not crash anymore.
I get a bunch of warnings on Dev tools, but I guess that's expected.
[...]

PR #3184, opened 2024-09-21
fix: null example bug

marclave:
fixes https://github.com/scalar/scalar/issues/3182 and https://github.com/scalar/scalar/issues/3165


amritk (approved):
yolo

PR #3188, opened 2024-09-22
chore(nestjs-api-reference, express-api-reference): move `@scalar/api-reference` and `express` from `dependencies` to `devDependencies`

### LINKED ISSUE #3159: In `@scalar/nestjs-api-reference`, `express` should be under `devDependencies` instead of `dependencies`
### What happens?

Hi, first of all thank you for this amazing project.

I was checking the npm of `@scalar/nestjs-api-reference` and I noticed that `express` is under the dependencies list. Then I went through the source code in the repo and check its usage, which turns out it's only used for type references similar to `fastify`.

As `fastify` is already in `devDependencies`, why not `express` as well?

`express` should be listed under `devDependencies` instead of `dependencies`

### OpenAPI Document

_No response_

### LINKED ISSUE #3187: In `@scalar/nestjs-api-reference` and `@scalar/express-api-reference`, the `@scalar/api-reference` should be removed from the dependencies.
### What happens?

For these packages, there is no need for `@scalar/api-reference` as a dependency, and they should use `@scalar/types` like `@scalar/fastify-api-reference`

when I install the `@scalar/express-api-reference` package.
Screenshot 2567-09-22 at 12 18 20

like `@scalar/fastify-api-reference` when i installed.
Screenshot 2567-09-22 at 12 18 38

### OpenAPI Document

_No response_

darekaze:
closes #3159
closes #3187

In `@scalar/nestjs-api-reference` and `@scalar/express-api-reference`, `@scalar/api-reference` and `express` should be listed under `devDependencies` to avoid including them in built package. Check above issues for details.

### DIFF
--- a/.changeset/nasty-penguins-argue.md
+++ b/.changeset/nasty-penguins-argue.md
@@ -1,0 +1,6 @@
+---
+'@scalar/express-api-reference': patch
+'@scalar/nestjs-api-reference': patch
+---
+
+chore: move `express` and `@scalar/api-reference` to `devDependencies`
--- a/packages/express-api-reference/package.json
+++ b/packages/express-api-reference/package.json
@@ -40,6 +40,7 @@
     "dist"
   ],
   "module": "dist/index.js",
+  "dependencies": {},
   "devDependencies": {
     "@scalar/api-reference": "workspace:*",
     "@types/express": "^4.17.21",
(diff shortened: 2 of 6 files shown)


inyourtime:
@darekaze I think we should add `@scalar/types` to the dependencies.

darekaze:
updated to use @scalar/types

marclave (approved):
thanks for the wonderful pr @darekaze :) ✨✨

and thanks @inyourtime for the perfect PR review <3 i was just telling @cameronrohani i was gunna leave that exact same review!

PR #3196, opened 2024-09-23
fix: retry flaky mutation test

tmastrom:
**Problem**
Currently, the [mutations test is flaky 
](https://github.com/scalar/scalar/actions/runs/10987367520/job/30502196429?pr=3192)
**Solution**
With this PR, the vitest syntax is updated to retry the test. Vitest docs https://vitest.dev/api/#test-api-reference


hanspagel:
Oh, great catch!

hanspagel (approved):

PR #3197, opened 2024-09-23
fix(api-client): sidebar folder toggle

antlio:
this pr enables back the sidebar item being disabled from a reference client, unless it was meant to be.


hanspagel (approved):
Ooops, yes! Let’s get this fixed.

PR #3201, opened 2024-09-23
feat(api-client): import collection modal

hanspagel:
This PR adds an import modal that comes up if you pass an URL and a title (optional) as query parameters.

I’d prefer to disable this feature and merge, to then ff in a separate PR with an improved Electron integration.

Note: The feature will be disabled until we’ve gone through the UI.

**Preview**
Screenshot 2024-10-01 at 16 08 14


antlio:
great work! found a small overflow issue on errors:
image

packages/api-client/src/components/ImportCollection/utils/isUrl.ts
```diff
@@ -0,0 +1,4 @@
+/** Checks whether the given string is an URL */
+export function isUrl(input: string | null) {
+  return input && (input.startsWith('http://') || input.startsWith('https://'))
+}
```
antlio, line 4: would it make send to use a url regex as a checker in the regexHelpers?

packages/api-client/src/components/ImportCollection/ImportNowButton.vue
```diff
@@ -0,0 +1,101 @@
+<script lang="ts" setup>
+import { isDocument } from '@/components/ImportCollection/utils/isDocument'
```
antlio, line 2: could this be removed? doesn't seem used here

packages/api-client/src/components/ImportCollection/utils/isUrl.ts
hanspagel: Hm … I think it’s fine. Way easier to read than any regex and not performance relevent.

marclave (approved):
@hanspagel let's merge after it's disabled? :) 

awesome work 🆒

packages/api-client/src/layouts/App/ApiClientApp.vue
```diff
@@ -60,6 +62,8 @@ const fontsStyleTag = computed(
 )
 </script>
 <template>
+  <!-- Listen for paste and drop events, and look for `url` query parameters to import collections -->
+  <!-- <ImportCollectionListener> -->
```
hanspagel, line 66: disabled already

PR #3194, opened 2024-09-23
fix(api-client): address typo style

antlio:
this pr fixes address bar empty url typo + adds style enhancements and loading overflow:

**before**
image

**after**
image

PR #3204, opened 2024-09-23
fix(api-client): address bar http methods

antlio:
this pr fixes the visibility of the address bar http methods dropdown introduced in #3194.


amritk (approved):
Why not

PR #3207, opened 2024-09-23
fix: modal background in dark mode

cameronrohani:
image

image


packages/themes/src/tailwind.ts
```diff
@@ -79,6 +79,7 @@ export default {
 
       // Utility Colors
       backdrop: 'rgba(0, 0, 0, 0.22)', // Overlay Backdrops
+      backdropdark: 'rgba(0, 0, 0, 0.45)', // Overlay Backdrops
```
antlio, line 82: should we camel case it?

PR #3214, opened 2024-09-23
fix(api-client): adding default extra row on route change

amritk:
This PR ensures we add the extra empty row to forms in the client on route change. Previously we were doing it when the content type changed, but if it didn't change and it was a different route there would be no row added.


marclave (approved):
:)

PR #3223, opened 2024-09-24
feat(api-reference): create new openapi pipeline

hanspagel:
WIP

We have this [`parse()`](https://github.com/scalar/scalar/blob/main/packages/api-reference/src/helpers/parse.ts) function, which loads the OpenAPI documents, dereferences them and does all kinds of strange transformations to the OpenAPI document.

I’d prefer to keep it as a regular OpenAPI document instead and build a new OpenAPI pipeline for the reference with the following goals:

- [x] Less transformations, keep the original OpenAPI document
- [ ] Maybe upgrade all documents to OpenAPI 3.1 on the fly?
- [x] Split the function into loading, dereferencing, upgrading? and normalization
- [x] Add tests
- [x] The normalization should make sure that the most important attributes are there
- [ ] Improved error handling and showing errors in the reference, not just some generic errors in the console
- [ ] Caching for the results so it’s way faster on the second run
- [ ] Reuse the result in the API client, so we only fetch documents once
- [x] Replace `useReactiveSpec`, `parse` and `type Spec` with the new pipeline and the generic `OpenAPI` type.

Feedback on the goals is highly appreciated, the code is nothing ATM.

Related #1902


hanspagel:
This is still great, but I’d like to take another approach (with a smaller impact/less changed files). Closing the PR, but keeping the branch.

PR #3224, opened 2024-09-24
fix(api-client): example sidebar modal

antlio:
this pr fixes:
- closes delete sidebar modal on action
- adds delete example warning message
- sets example parent request


amritk (approved):
I found the correct request was already being pre-selected, but the rest is good!

PR #3234, opened 2024-09-25
fix: fathom script

marclave:
**Problem**
Fathom isnt working properly

**Solution**
With this PR it might fix it


amritk (approved):
gogog

PR #3237, opened 2024-09-25
chore(api-client): create workspace wording

hanspagel:
Doing the important work over here. 😁

* If I create something, it’s obviously new.
* We use title case for other add/create actions.

**Before**
Screenshot 2024-09-25 at 10 03 11

**After**
Screenshot 2024-09-25 at 10 03 54


antlio (approved):
great catch

Issue #3241, opened 2024-09-25
Remove `@scalar/themes` from the dependencies of `@scalar/types`


inyourtime:
### What happens?

The `@scalar/types` package currently imports some types from the `@scalar/themes` package. To reduce dependencies, we need to remove the reliance on `@scalar/themes` by refactoring those types or just copy that type to `@scalar/types`.

### What did you expect to happen?

we will eliminate the `@scalar/themes` dependency from `@scalar/types`, making it more self-contained and reducing external package dependencies.

hanspagel:
That’s a great suggestion! Would you submit a PR?

inyourtime:
Sure, I will submit a PR for this.

PR #3243, opened 2024-09-25
feat: remove `@scalar/themes` from the dependencies of `@scalar/types`

inyourtime:
#3241

### DIFF
--- a/packages/themes/tsconfig.json
+++ b/packages/themes/tsconfig.json
@@ -1,20 +1,8 @@
 {
+  "extends": "../../tsconfig.json",
   "compilerOptions": {
-    "baseUrl": ".",
-    "target": "ES2022" /* Set the JavaScript language version for emitted JavaScript and include compatible library declarations. */,
-    "module": "ESNext" /* Specify what module code is generated. */,
-    "moduleResolution": "Bundler",
-    "resolveJsonModule": true,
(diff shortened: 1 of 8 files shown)


hanspagel (changes requested):
Let’s try to add types as a dependency of themes and import the ThemeId there instead of copying the type ✌️

inyourtime:
i will do it.

packages/themes/tsconfig.json
```diff
@@ -3,7 +3,7 @@
     "baseUrl": ".",
     "target": "ES2022" /* Set the JavaScript language version for emitted JavaScript and include compatible library declarations. */,
     "module": "ESNext" /* Specify what module code is generated. */,
-    "moduleResolution": "Node",
+    "moduleResolution": "Bundler",
```
amritk, line 6: hey great work! do you think you can do one better and just extend the base tsconfig?

```json
{
  "extends": "../../tsconfig.json"
}
```

amritk (changes requested):
Just requested one change to the tsconfig, but excellent work!

amritk (approved):
LGTM!

hanspagel (approved):
Thanks so much! :) Love when the dependency tree becomes smaller/simpler.

PR #3244, opened 2024-09-25
chore: update to fastify v5

VialFlorian:
Update [fastify to v5](https://github.com/fastify/fastify/releases/tag/v5.0.0), and all the fastify plugins to their new compatible v5 version

### DIFF
--- a/examples/cdn-api-reference/package.json
+++ b/examples/cdn-api-reference/package.json
@@ -24,10 +24,10 @@
   },
   "type": "module",
   "dependencies": {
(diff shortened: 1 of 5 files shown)


tmastrom:
Nice! We'll have to do this upgrade alongside a larger upgrade to node 20 and drop support for node 18. [Fastify v5 will only support node 20](https://fastify.dev/docs/latest/Guides/Migration-Guide-V5/#long-term-support-cycle).

PR #3254, opened 2024-09-25
fix(api-client-app): invalid pageview url

hanspagel:
Turns out we’ve used an invalid URL for the pageviews, which worked locally, but not in the build - for some reason.


amritk (approved):
⛵

marclave (approved):

PR #3258, opened 2024-09-26
fix(api-reference): show examples in models section

hwkr:
Shows all the schema example values in the "Models" section again when the `examples` keyword is used (e.g. an array of examples). It sounds like OAS 3.1 has introduced the ability to add multiple examples to a schema as part of its alignment with JSON schema which only [allows an array of examples](https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-validation-00#section-9.5). 

I also added some additional example data to Scalar Galaxy with multiple examples.

Google Chrome-2024-09-25-18-08-06@2x

Google Chrome-2024-09-25-18-16-57@2x

### DIFF
--- a/.changeset/clean-buckets-perform.md
+++ b/.changeset/clean-buckets-perform.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
+feat: support for multiple examples in schemas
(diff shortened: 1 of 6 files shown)


hanspagel (approved):
Finally! :) Thanks for adding to the galaxy example, too.

cameronrohani (approved):
awesome 😍

PR #3270, opened 2024-09-26
chore(oas-utils): `getExampleFromSchema` performance improvements

hanspagel:
I’m playing around with a huge OpenAPI document with huge schemas. Find the screenshot of a perf profile for opening a tag. The big blue box is `getExampleFromSchema` taking more than 1s:

image

This PR tries to improve the performance and the results are amazing: The new implementation is 1.000× faster than what we had before.

Screenshot 2024-10-02 at 12 54 46

### DIFF
--- a/.changeset/sharp-walls-bathe.md
+++ b/.changeset/sharp-walls-bathe.md
@@ -1,0 +1,5 @@
+---
+'@scalar/oas-utils': patch
+---
+
+chore: improve performance of getExampleFromSchema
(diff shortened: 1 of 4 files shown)


marclave (approved):
this is HUGE ✨

amritk (approved):
Would be nice if you included a little blurb about what exactly caused the perf increase (was it as simple as removing the pretty print?). BUT I LOVE IT. 

What is that bench? And how to use? I just use console.time/end everywhere

PR #3275, opened 2024-09-26
fix(api-client): don't teleport auth combobox to body

hwkr:
Not 100% sure why but teleporting a div onto the body causes a page scroll.

https://github.com/user-attachments/assets/a05c58dd-d931-48da-bd7d-3e478d58002a


amritk (approved):
🔥

PR #3276, opened 2024-09-26
fix(api-reference): pressing enter scroll bug

### LINKED ISSUE #3245: Hitting 'Enter' in any input field resets the url path
### What happens?

Pressing Enter inside any input field, trims the url from 
`http://localhost:5000/scalar/api#tag/accesskey/DELETE/api/accesskey/{id}`
to `http://localhost:5000/scalar/api#tag/accesskey`

Enter should send cursor to the beginning of the next line (new entry for key: values, new line inside body etc)

### OpenAPI Document

_No response_

amritk:
Turns out our global listeners were listening the whole time, we should pretty much never use `whenever` and always put the events on the element. And for any actually global events we use the event bus

Also a bunch of these search components werent doing anything

closes #3245

To test you can use the arrow keys on the search pop up 
image

and hit enter anywhere else on the page, neither should change the url


packages/api-reference/src/features/Search/Search.vue
hwkr, line 1: Was this being used for @hanspagel component library / playground?

amritk: Didn't look like it was being used anywhere. This build should break if it was

hwkr (approved):
Looks good!

antlio:
great catch

antlio (approved):
great job and catch!

hanspagel:
GREAT clean up! I think it was me who introduced whenever at some point, I copied an example from `vueuese` I think. 🙈 

I’m not sure if this is related to this PR, but the keyboard shortcuts behave really strange in the modal. It looks like it’s jumping to random positions.

https://github.com/user-attachments/assets/54106cfc-f626-49ee-9c2c-d169b3bbe12a

hanspagel (changes requested):
Let’s make sure this wasn’t introduced with that PR, or - ideally - improve it with this PR here.

antlio:
> Let’s make sure this wasn’t introduced with that PR, or - ideally - improve it with this PR here.

great catch! should be fixed with the latest commit

amritk:
Yea I wasn't sure if we wanted center or top

amritk:
K random jumpyness fixed!!

packages/api-reference/src/features/Search/SearchModal.vue
```diff
@@ -95,6 +86,26 @@ function onSearchResultClick(entry: FuseResult<FuseData>) {
   }
 }
 
+// Scroll to the currently selected result
+watch(selectedSearchResult, (index) => {
+  const newResult = searchResultsWithPlaceholderResults.value[index]
+
+  document.getElementById(newResult.item.href)?.scrollIntoView({
+    behavior: 'smooth',
+    block: 'start',
```
hwkr, line 95: Do we actually want this to be nearest?

packages/api-reference/src/features/Search/SearchModal.vue
```diff
@@ -108,7 +119,14 @@ function getFullUrlFromHash(href: string) {
 <template>
   <ScalarModal
     :state="modalState"
-    variant="search">
+    variant="search"
+    @keydown.down.stop="navigateSearchResults('down')"
```
hwkr, line 123: We probably want `.prevent` on these too so we don't move the cursor when going through results

packages/api-reference/src/features/Search/SearchModal.vue
amritk: this is currently up for debate haha anthony is on team center, idc really

hwkr: Well if we do any of the positioned ones then it moves every time you press an arrow key even if it doesn't need to.

**`block: 'start'`**:
 
https://github.com/user-attachments/assets/c659dfd9-a0ea-4688-9166-276e10f93401

**`block: 'nearest'`**:

https://github.com/user-attachments/assets/b2b83e79-cb82-4ca7-9ffa-acf1d10fce97

amritk: I'm in for nearest across the board. Will wait for the others to confirm then merge. Good call!

packages/api-reference/src/features/Search/SearchModal.vue
amritk: done

amritk:
Going to merge because it feels optimized, any gripes can come as a follow up!

PR #3278, opened 2024-09-27
feat(api-client): sets initial security scheme in the client 

### LINKED ISSUE #3264: preferredSecurityScheme Is Not Working
### What happens?

When attempting to configure Scalar to use any `preferredSecurityScheme`, the security schema is not chosen by default in the "Test Request" web client.

This results in the user needing to select a security schema from the available list before sending each-request.

Here is my code:
```html
<!doctype html>
<html>
  <head>
    <title>Scalar API Reference</title>
    <meta charset="utf-8" />
    <meta
      name="viewport"
  </head>
  <body>
    <script
      id="api-reference"
      data-url="https://cdn.jsdelivr.net/npm/@scalar/galaxy/dist/latest.yaml"></script>

    <script>
      const configuration = {
        isEditable: false,
        authentication: {
          preferredSecurityScheme: "oauth2",
          oAuth2: {
            clientId: "test"
          }
        }
      };
      const apiReference = document.getElementById("api-reference");
      apiReference.dataset.configuration = JSON.stringify(configuration);
    </script>

  </body>
</html>
```

Setting `preferredSecurityScheme` with a value matching one of the `components.securitySchemes` keys (defined in the Open API YAML), should result in that security schema being selected by default in the "Test Request" web client.

### OpenAPI Document

https://cdn.jsdelivr.net/npm/@scalar/galaxy/dist/latest.yaml

amritk:
Since security schemes are selected per request, it isn't so easy to change the currently selected scheme from the references. HOWEVER we can set the scheme on initialization!

closes #3264

### DIFF
--- a/packages/api-client/src/libs/create-client.ts
+++ b/packages/api-client/src/libs/create-client.ts
@@ -5,57 +5,24 @@ import { workspaceSchema } from '@scalar/oas-utils/entities/workspace'
 import { LS_KEYS, objectMerge } from '@scalar/oas-utils/helpers'
 import { DATA_VERSION, DATA_VERSION_LS_LEY } from '@scalar/oas-utils/migrations'
 import type { Path, PathValue } from '@scalar/object-utils/nested'
-import type { ThemeId } from '@scalar/themes'
-import type { Spec, SpecConfiguration } from '@scalar/types/legacy'
+import type {
+  AuthenticationState,
+  ReferenceConfiguration,
+  SpecConfiguration,
+} from '@scalar/types/legacy'
 import type { LiteralUnion } from 'type-fest'
 import { type Component, createApp } from 'vue'
 import type { Router } from 'vue-router'
 
 /** Configuration options for the Scalar API client */
 export type ClientConfiguration = {
-  /** The Swagger/OpenAPI spec to render */
-  spec?: SpecConfiguration
-  /** Pass in a proxy to the API client */
-  proxyUrl?: string
-  /** Pass in a theme API client */
-  themeId?: ThemeId
-  /** Whether to show the sidebar */
-  showSidebar?: boolean
-  /** The initially selected security scheme */
-  initialScheme?: string | null
-  /** override the initial servers */
-  servers?: Spec['servers']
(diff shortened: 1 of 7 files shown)


packages/api-reference/src/components/ApiClientModal.vue
```diff
@@ -25,6 +26,7 @@ onMounted(async () => {
   const { app, open, updateAuth, modalState, updateSpec, updateServer } =
     await createApiClientModal(el.value, {
       spec: props.spec ?? {},
+      initialScheme: props.preferredSecurityScheme,
```
hanspagel, line 29: Should this have the same name `preferredSecurityScheme` as in the API Reference configuration?

amritk: Yea I had it as preferred before but I found it a bit of a misnomer. I named it initial in the client because it won't get updated on change, it only gets set on import

packages/api-client/src/libs/create-client.ts
```diff
@@ -5,55 +5,24 @@ import { workspaceSchema } from '@scalar/oas-utils/entities/workspace'
...
-    | 's'
-    | 't'
-    | 'u'
-    | 'v'
-    | 'w'
-    | 'x'
-    | 'y'
-    | 'z'
-}
+  proxyUrl?: ReferenceConfiguration['proxy']
```
amritk, line 19: if we re-use the types, we only need to set the comment once. We can do really in depth comments in the types package

image

packages/api-reference/src/features/Search/SearchModal.vue
```diff
@@ -129,9 +129,11 @@ function getFullUrlFromHash(href: string) {
         @click="onSearchResultClick(entry)"
         @focus="selectedSearchResult = index">
         <span
-          :class="{ deprecated: entry.item.operation?.information?.deprecated }"
-          >{{ entry.item.title }}</span
-        >
+          :class="{
+            deprecated: entry.item.operation?.information?.deprecated,
```
amritk, line 133: classic vue prettier shenanigans

marclave (approved):
we can change the name later if need be :)

PR #3279, opened 2024-09-27
style(api-client): design tweaks to address recent feedback

cameronrohani:
pt 1 of 2 :)

make sidebar same color as everything else:
image

empty state:
image

hover state:
image


cameronrohani:
https://github.com/user-attachments/assets/1bf90ea1-6aeb-46bd-a0c4-9f3455ab18be

packages/api-client/src/components/CodeInput/CodeInput.vue
```diff
@@ -279,6 +279,7 @@ export default {
   color: var(--scalar-color-3);
   font-size: var(--scalar-small);
   line-height: 1.44;
+  border-radius: 0 0 0 3px;
```
antlio, line 282: good catch!

antlio:
nice to see some animations comin 👏 
would it make sense to find a way to display at all time the current selected filter state for the context bar?

packages/api-client/src/views/Request/RequestSidebar.vue
```diff
@@ -183,14 +186,35 @@ onBeforeUnmount(() => {
...
-        v-if="!isReadonly"
+      <div
+        class="relative z-10 pt-0 md:px-2.5 md:pb-2.5 -translate-y-full w-[inherit]"
         :class="{
           'empty-sidebar-item': activeWorkspaceRequests.length === 1,
-        }"
-        :click="commandPaletteBus.emit">
-        <template #title>Add Item</template>
-      </SidebarButton>
+          'border-t-1/2': activeWorkspaceRequests.length === 1,
```
antlio, line 193: should we set it to be less or equal to one here? currrently if we remove the all the request then the state is not displayed anymore

cameronrohani: sure! great idea

cameronrohani:
> nice to see some animations comin 👏 would it make sense to find a way to display at all time the current selected filter state for the context bar?

great idea, feel free to add to this pr or I can fast follow tomorrow :)

PR #3281, opened 2024-09-27
feat(api-client): request and response section

antlio:
this pr updates the request and response section order and naming as seen below:
- `all` displayed first
- `variables` for path variables section instead of request

**before**
image

**after**
image


hanspagel (approved):
Yess

amritk (approved):
Yup up

PR #3287, opened 2024-09-28
fix(api-references): selected server doesn’t update the API client, fix #2830

### LINKED ISSUE #2830: changes to the base url in the reference don’t update the client
When I change the base URL in the reference, it doesn’t change in the api client. That used to work. 🤔

Be aware that #2825 moves the relevant files. Maybe wait until this is merged. :)

**Sandbox**

https://sandbox.scalar.com/e/Tpnn8

**Screenshots**

Screenshot 2024-08-09 at 21 29 18
Screenshot 2024-08-09 at 21 29 23

hwkr:
Synchronizes the selected server in the references with the selected server in the api client.

I'm not sure if this is the preferred callback pattern but I did my best to keep it simple and clean up the server selector a little bit. I wasn't sure how to set up variables to test that part @hanspagel?

## Before / After

https://github.com/user-attachments/assets/f2e6cbbc-e4a0-4eca-bfb7-a9e329528ad6

### DIFF
--- a/.changeset/eleven-tigers-mix.md
+++ b/.changeset/eleven-tigers-mix.md
@@ -1,0 +1,6 @@
+---
+'@scalar/api-reference': patch
+'@scalar/api-client': patch
+---
+
+fix(api-references): sync selected server state with api client modal
--- a/packages/api-client/src/libs/create-client.ts
+++ b/packages/api-client/src/libs/create-client.ts
@@ -11,7 +11,7 @@ import type {
   SpecConfiguration,
 } from '@scalar/types/legacy'
 import type { LiteralUnion } from 'type-fest'
-import { type Component, createApp } from 'vue'
+import { type Component, createApp, watch } from 'vue'
 import type { Router } from 'vue-router'
 
 /** Configuration options for the Scalar API client */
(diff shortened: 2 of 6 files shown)


hanspagel (approved):
Finally and forever

amritk (approved):
So this is great and keeps both stores in sync, but what I would like to do is start using the same source of truth for the client + references. IF this is even possible @hanspagel but I mean it should be since its all openapi 😄

hwkr:
> So this is great and keeps both stores in sync, but what I would like to do is start using the same source of truth for the client + references. IF this is even possible @hanspagel but I mean it should be since its all openapi 😄 

Yeah I figured some sort of coordinated store would be the ideal but I figure that could be a next step?

amritk:
> Yeah I figured some sort of coordinated store would be the ideal but I figure that could be a next step?

Yep in retrospect I totally should have known what you were doing when u asked about syncing, but my mind was elsewhere. I would have said at the time we should explore this. 

But all good! Anyway me and hans should probably nail down what its going to look if we start sharing things before we get started

marclave (approved):
awesome work @hwkr !!!

PR #3288, opened 2024-09-28
fix(api-client): disable object preview due to blob XSS vulnerability

hwkr:
Turns out `blob` urls in `object` (or `iframe`) tags are executed in the same origin context as the host page which means they're vulnerable to XSS injection. This disables the object preview (which is used for html and pdf previews) for the time being until we can preview them safely. The `svg` preview should still be safe though since it's rendered with an ``.

I also updated our `Content-Security-Policy` for the app and for `examples/api-client` (which seems to be used for `client.scalar.com`) to not allow scripts from blobs and to only allow blobs to be used for `` and media tags (`<video>` and `<audio>`). This prevents scripts from running in svgs when they're opened in a new tab via right click → "Open Image in New Tab"

## `Content-Security-Policy` Before / After

https://github.com/user-attachments/assets/0590325e-6a5e-4976-8dc4-1dd15c0a4a30

### DIFF
--- a/.changeset/old-olives-itch.md
+++ b/.changeset/old-olives-itch.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
+
+fix(api-client): disable object preview due to blob XSS vulnerability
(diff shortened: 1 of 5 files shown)


hwkr:
Closing in favor of #3289

PR #3289, opened 2024-09-28
fix(api-client): add Content-Security-Policy to client.scalar.com

### LINKED ISSUE #3286: Cross Site Scripting Issue in Scalar Client
### What happens?

From the Scalar Client (https://client.scalar.com/), making a request to a page that returns javascript causes the script to be executed on `client.scalar.com`. This is a security issue, because it could allow an attacker to exfiltrate data from the user's session.

image

The Scalar Client should not permit executing scripts.

### OpenAPI Document

_No response_

hwkr:
Turns out blob urls in object (or iframe) tags are executed in the same origin context as the host page which means they're vulnerable to XSS injection. 

I updated our `Content-Security-Policy` for the app and set a `Content-Security-Policy` for examples/api-client (which seems to be used for client.scalar.com) to not allow scripts from blobs. This prevents scripts from running in html previews and in svgs when they're opened in a new tab via right click → "Open Image in New Tab"

Fixes #3286 

## `Content-Security-Policy` Before / After

https://github.com/user-attachments/assets/08340b65-70d8-4f69-b6ce-d5fd8aa40128

https://github.com/user-attachments/assets/0590325e-6a5e-4976-8dc4-1dd15c0a4a30

### DIFF
--- a/.changeset/old-olives-itch.md
+++ b/.changeset/old-olives-itch.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
+
+fix(api-client): add Content-Security-Policy to client.scalar.com
--- a/examples/api-client/index.html
+++ b/examples/api-client/index.html
@@ -8,6 +8,9 @@
     <meta
       name="viewport"
       content="width=device-width, initial-scale=1.0" />
+    <meta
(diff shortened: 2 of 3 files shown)


amritk (approved):
Security!

PR #3290, opened 2024-09-28
fix(api-reference): OAuth authorization code flow rendering

### LINKED ISSUE #1960: OAuth Authorization Code bug
### What happens?

When using `@scalar/api-reference-react@0.1.96` (`@scalar/api-reference@1.22.54`)  or later - `g.value.flows.password is undefined` is thrown when using Authorization Code security scheme.

It looks like https://github.com/scalar/scalar/pull/1862 caused this issue

I'm assuming support for this flow will come at some point however the unsupported schemes should be ignored until supported. I can try and look at raising a PR, however I can see that work on the OAuth functionality is ongoing (e.g. https://github.com/scalar/scalar/pull/1949)

1. Create schema with Authorization Code scheme:
```yaml
components:
  securitySchemes:
    OAuth2:
      type: oauth2
      flows:
        authorizationCode:
          authorizationUrl: https://example.com
          tokenUrl: https://example.com
          scopes:
            scope: scope
```
2. Use schema with `@scalar/api-reference-react@0.1.95`/`@scalar/api-reference@1.22.53` - everything works fine.
3. Use schema with `@scalar/api-reference-react@0.1.96`/`@scalar/api-reference@1.22.54`, `g.value.flows.password is undefined` is thrown

### OpenAPI Document

https://sandbox.scalar.com/e/6mZtB

carstenlebek:
## Problem
The reference currently doesn't support the OAuth Authorization Code flow due to an issue in the `SecurityScheme.vue` component. The password flow scopes have a non-null assertion, which prevents the Authorization Code flow from being rendered properly. This issue was reported in #1960.

## Solution
This PR modifies the `SecurityScheme.vue` component to correctly support the Authorization Code flow. Additionally, it adds an Authorization Code flow example to the galaxy spec for testing and demonstration purposes.

## Changes
- Updated `SecurityScheme.vue` to handle Authorization Code flow rendering
- Added Authorization Code flow example to the galaxy spec
- Removed unnecessary non-null assertion for password flow scopes

### Before:
Before changes

### After:
After changes

## Related Issues
- Fixes #1960: OAuth Authorization Code bug

## Testing
- Verified that the Authorization Code flow now renders correctly
- Ensured that existing functionality for other OAuth flows remains intact
- Tested with the updated galaxy spec example

## Additional Notes
This change resolves a long-standing issue that has affected users trying to implement OAuth Authorization Code flow. It should significantly improve the usability of the client for OAuth 2.0 implementations.

### DIFF
--- a/.changeset/empty-donkeys-tan.md
+++ b/.changeset/empty-donkeys-tan.md
@@ -1,0 +1,5 @@
+---
+'@scalar/galaxy': patch
+---
+
+feat: add oauth code authorization
--- a/.changeset/lemon-shoes-judge.md
+++ b/.changeset/lemon-shoes-judge.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
+fix: show oauth code authorization
--- a/packages/api-reference/src/legacy/components/SecurityScheme.vue
+++ b/packages/api-reference/src/legacy/components/SecurityScheme.vue
@@ -463,13 +463,15 @@ const startAuthentication = (url: string) => {
                 value !== undefined &&
                 Object.entries(
                   (value as any).flows.implicit?.scopes ??
-                    (value as any).flows.password!.scopes,
+                    (value as any).flows.password?.scopes ??
+                    (value as any).flows.authorizationCode?.scopes,
                 ).length > 0
               "
               v-model:selected="oauth2SelectedScopes"
               :scopes="
                 (value as any).flows.implicit?.scopes ??
-                (value as any).flows.password!.scopes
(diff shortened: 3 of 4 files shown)


hanspagel (approved):
AWESOME! Thanks for adding it to the OpenAPI example, too.

Just waiting for @amritk to do a review here.

amritk (approved):
Yep this is great and thanks for doing it. I want to echo what I said on the other PR about sharing more resources between client + references so we are not duplicating work. @hanspagel we should have a sit down about it!

PR #3291, opened 2024-09-29
fix(api-client): skip empty value of query parameters

inyourtime:
**Problem**
Currently, query parameters are populated even when their values are empty, leading to incorrect or unnecessary parameters being included in the request. This can cause issues when the backend expects non-empty values for certain query parameters.

**Solution**
With this PR, the logic has been updated to skip query parameters with empty values. Now, only enabled query parameters with non-empty values will be appended to the request, ensuring that the resulting query string is correct and avoids sending unnecessary parameters.

### DIFF
--- a/.changeset/fast-pandas-shake.md
+++ b/.changeset/fast-pandas-shake.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
+
+fix: skip empty value of query parameters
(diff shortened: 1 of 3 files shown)


amritk (approved):

marclave (approved):
awesome work @inyourtime !!!

PR #3297, opened 2024-09-29
fix(api-client): allow to add servers in read only mode

marclave:
**Problem**
Currently, you couldn't see the add servers if there was only 1 server in the address bar

**Solution**
With this PR we bypass this logic and show all options when in the api client

### DIFF
--- a/.changeset/empty-snails-vanish.md
+++ b/.changeset/empty-snails-vanish.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
+
+fix: force show servers when not in read only mode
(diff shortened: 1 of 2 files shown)


hanspagel (approved):
Actually, this makes my life so much easier. I missed that button all the time.

antlio:
can you add a screenshot of the expecting result? that's what i get from the example:
image

hanspagel:
For Swagger v2 (Import Collection > https://petstore.swagger.io/v2/swagger.json):

Screenshot 2024-09-30 at 15 38 26

Issue #3292, opened 2024-09-29
"Hello World" Teleport


arkhvoid:
### What happens?

Previous commit accidentally pushed a test ui "Hello World". [RequestAuth.vue#L215-L219](https://github.com/scalar/scalar/blob/1b4656474109721937fde882af5d2d05987cafc6/packages/api-client/src/views/Request/RequestSection/RequestAuth/RequestAuth.vue#L215-L219).

### What did you expect to happen?

should not display the test ui.

marclave:
😅 haha nice catch, let's fix this

PR #3298, opened 2024-09-29
fix: remove stray hello world html

marclave:
fixes https://github.com/scalar/scalar/issues/3292


amritk (approved):
Lol where was this from

PR #3299, opened 2024-09-29
fix(api-client): add correct version number in ascii preview

marclave:
before:
image

after:
image

### DIFF
--- a/.changeset/good-rabbits-invite.md
+++ b/.changeset/good-rabbits-invite.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
+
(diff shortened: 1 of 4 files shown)


hanspagel (changes requested):
This would be so cool! Unfortunately, this doesn’t work for two reasons:

1. the build fails

pnpm build
> [json] [plugin json] package.json: Could not parse JSON file

2. the build happens before we bump the version number

So we’ll always show the previous version number. Or we build a 2nd time after we bumped the version number. 🤔

marclave:
@hanspagel youre the GOAT

PR #3300, opened 2024-09-29
fix(client.scalar.com): add missing favicon to client.scalar.com

marclave:
now we have a favicon for https://client.scalar.com


antlio (approved):
nice call!

hanspagel (changes requested):
Doesn't work because of the recently introduced CSP

hanspagel (approved):
Fixed! :)

PR #3302, opened 2024-09-29
fix(api-client): add missing securityDefinitions check on import

marclave:
**Problem**
Currently when you import `"swagger": "2.0",` and using `securityDefinitions`, it breaks our client 😨 

**Solution**
With this PR we check for that field on import 😎 

https://petstore.swagger.io/v2/swagger.json

test that spec on https://client.scalar.com vs our preview 😎 

**edit** Amrit brought up that we need to handle this in the upgrade actually 🤯


hanspagel (approved):
Nice

PR #3307, opened 2024-09-30
fix(api-client): sidebar empty state

antlio:
this pr updates the empty sidebar item content introduced in #3279 to be re-introduced once their import is supported.


hanspagel (approved):
accurate

marclave (approved):

PR #3309, opened 2024-09-30
fix(api-client): too many z-index layers

hanspagel:
I faced a z-index issue in another branch, so I tried to debug it and found a lot of z-index layers throughout the API client. Some in Tailwind, some in CSS, some even conflicting with eachother.

I tried removing all and it didn’t look too bad, so @hwkr added a Tailwind config and reintroduced a z-index where actually needed.

We should test this thoroughly, though. I might have missed places where we need to add it again. Can you all help me with that? 👀 Just leave a screenshot here and we’ll fix it.

### DIFF
--- a/packages/api-client/src/layouts/Modal/ApiClientModal.vue
+++ b/packages/api-client/src/layouts/Modal/ApiClientModal.vue
@@ -44,12 +44,12 @@ onBeforeUnmount(() => {
   <div
     v-show="modalState.open"
     class="scalar">
-    <div className="scalar-container">
+    <div className="scalar-container z-overlay">
       <div className="scalar-app scalar-client scalar-app-layout">
         <RouterView key="$route.fullPath" />
       </div>
       <div
-        class="scalar-app-exit"
+        class="scalar-app-exit -z-1"
         @click="modalState.hide()"></div>
     </div>
   </div>
--- a/packages/api-client/src/layouts/Modal/ApiClientModal.vue
+++ b/packages/api-client/src/layouts/Modal/ApiClientModal.vue
@@ -74,7 +74,6 @@ onBeforeUnmount(() => {
   margin: auto;
   opacity: 0;
   animation: scalarapiclientfadein 0.35s forwards;
-  z-index: 1002;
   position: relative;
   overflow: hidden;
   border-radius: 8px;
@@ -95,7 +94,6 @@ onBeforeUnmount(() => {
   height: 100vh;
   background: #00000038;
   transition: all 0.3s ease-in-out;
-  z-index: 1000;
   cursor: pointer;
   animation: scalardrawerexitfadein 0.35s forwards;
 }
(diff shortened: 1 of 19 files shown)


antlio:
great call!

seems like we cannot click the back button anymore
image

antlio (approved):
beside my comment seems alright! great!

hwkr (changes requested):
Oh yeah we can use tailwind for this, components has a good example of setting [custom z-index config](https://github.com/scalar/scalar/blob/main/packages/components/tailwind.config.ts#L30-L34).

tl;dr;

```js
export default {
  // ...
  theme: {
    extend: {
     // ...,
      zIndex: {
        // Contextual overlays like dropdowns, popovers, tooltips
        context: '1000',
        // Full screen overlays / modals
        overlay: '10000',
      },
    },
  },
}
```

Then you use the class as `z-overlay`.

hwkr (approved):

hwkr:
Alright I updated this and cleaned up everything that seemed broken, I couldn't figure out how to test the address bar history though. For some reason the button isn't appearing for me even on main 🤔

hanspagel:
In the reference, when I open the API client modal, the background is in front of the modal.

https://976f4431-6bfa-4a93-bb53-eaea404ac630--scalar-deploy-preview.netlify.app/standalone-api-reference

Screenshot 2024-10-08 at 14 15 56

packages/api-client/src/layouts/Modal/ApiClientModal.vue
```diff
@@ -44,12 +44,12 @@ onBeforeUnmount(() => {
...
     v-show="modalState.open"
     class="scalar">
-    <div className="scalar-container">
+    <div className="scalar-container z-overlay">
       <div className="scalar-app scalar-client scalar-app-layout">
         <RouterView key="$route.fullPath" />
       </div>
       <div
-        class="scalar-app-exit"
+        class="scalar-app-exit -z-1"
```
antlio, line 52: any reason to have the z-overlay here (out of curiosity)? it seems like the `-z-1` is making it

hanspagel:
Awesome work! I think we can merge this, maybe just look into @antlio’s question above. 👌

packages/api-client/src/layouts/Modal/ApiClientModal.vue
hanspagel: Should we use another Tailwind preset instead of a number here?

antlio: can also just add the `z-overlay` as a `scalar-client` class sibling so we don't have to?

hwkr: We want to redefined the stacking context for `scalar-app-exit` so it gets moved behind `scalar-client` (which inherently has a `z-index` of 0 in the stacking context) but not behind other elements on the page. 

If we don't make it an overlay there then theres a risk of the app exit getting put behind other high z-index elements on the page the modal is embedded in.

hwkr:
Oops may have found another z-index issue

Google Chrome-2024-10-09-20-29-30@2x

PR #3310, opened 2024-09-30
ci: don’t run jobs with secrets in forks

hanspagel:
For security reasons secrets aren’t available in PRs coming from forks.

Let’s skip steps that require secrets, so they don’t fail for external contributions.

Screenshot 2024-09-30 at 15 06 54


tmastrom (approved):
Nice, we can save some ci time this way

PR #3313, opened 2024-09-30
chore(api-reference): add Tailwind

hanspagel:
I just noticed Tailwind isn’t set up in `@scalar/api-reference`, but as this is what we’d like to use in the future I thought it might be time to integrate it.

It’s scoped under a different class name, so the CSS will be duplicated. But CSS can be gzipped pretty well, and maybe we can run the API client and the API reference under the same class to avoid CSS duplication.


antlio:
seems like its breaking some:
image

hwkr:
Updated the style load order so that the client tailwind styles can beat the references tailwind styles and that seems to have fixed the issues :)

amritk:
@hwkr if I ever get around to writing that plugin we won't need to worry about it anymore 😅

hwkr:
> @hwkr if I ever get around to writing that plugin we won't need to worry about it anymore 😅 

One day 😂

hwkr:
> @hwkr if I ever get around to writing that plugin we won't need to worry about it anymore 😅 

One day 😂

marclave:
how can we test this with docusaurus, and make sure we arent going to be breaking other peoples themes and our tailwind styles dont overwrite others.

we've had a few instances of our tailwind breaking other peoples sites and would be great to not have that happen! :)

hanspagel:
Good point! We’re using the same way to scope the styles as we do in `@scalar/api-client` already, so it’s beyond my imagination to find something that makes it different to what we do there already.

hwkr:
> Good point! We’re using the same way to scope the styles as we do in `@scalar/api-client` already, so it’s beyond my imagination to find something that makes it different to what we do there already. Wdyt @hwkr?

Should be fine but doesn't hurt to double check!

hanspagel:
Checked those scenarios and it all looks good (and properly scoped), so let’s try it! :)

hanspagel:
Can’t approve my own PR, though it’s mostly @hwkr’s work, so I’ll merge it once CI is green. 👀

PR #3315, opened 2024-09-30
feat(api-reference): new fun way to test popular OpenAPI documents

hanspagel:
This PR adds a new @scalar/api-reference playground to test popular OpenAPI examples. It has a list of OpenAPI documents to pick from and a random button to load up … a random OpenAPI document.

I think we should deploy this and add it as the default preview link to PRs. :)

BTW, clicking through it with an open browser console, you can already spot a few minor issues that are not reported yet.

**Usage**
```bash
cd packages/api-reference
pnpm playground:examples
```

**Preview**
Screenshot 2024-09-30 at 17 33 14


marclave (changes requested):
as discussed with @hanspagel , lets put these somewhere else :)

marclave (approved):
😎

PR #3321, opened 2024-10-01
feat(openapi-parser): self reference the document

x-delfino:
**Problem**
Currently, openapi-parser does not support self references in `$ref` properties  e.g. openapi.json cannot reference an object with `./openapi.json#/parameters/Parameter` (I think it gets stuck in a loop of circular references). Whilst this is a bit of a weird way of doing a reference, I don't believe it's against the spec (https://json-schema.org/draft/2020-12/json-schema-core#ref) and I've come across a few instances of this in the [Azure API spec](https://github.com/Azure/azure-rest-api-specs/blob/ad60d7f8eba124edc6999677c55aba2184e303b0/specification/apimanagement/resource-manager/Microsoft.ApiManagement/stable/2016-10-10/apimanagement.json#L49).

**Solution**
This PR adds a simple check during reference resolution to see if the reference prefix matches the current filename. This doesn't account for all scenarios atm. e.g.  `../parentfolder/openapi.json#/parameters/Parameter`

### DIFF
--- a/.changeset/khaki-cheetahs-doubt.md
+++ b/.changeset/khaki-cheetahs-doubt.md
@@ -1,0 +1,5 @@
+---
+'@scalar/openapi-parser': patch
+---
+
+feat: self reference the document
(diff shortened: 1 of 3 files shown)


hanspagel:
Great find! Would you be able to add a test?

x-delfino:
I've added a test now. another scenario not supported is that this wouldn't match if the filename in the filesystem is `./openapi.json` and the $ref is to `openapi.json` (I think both are valid). we should probably resolve the full filepath and compare on that

hanspagel:
Yeah, but this is already better than what we had, so let’s get this in!

hanspagel (approved):
Awesome contribution, as always

Issue #3317, opened 2024-09-30
Variables in server urls are not usable in client


guillaumefont:
### What happens?

I have generated a doc with a server url that uses variables.

![image](https://github.com/user-attachments/assets/a32ea828-d534-4d3d-beb0-defb0afa3c2a)

The generated examples are correct, but when I want to test a request using the client, the variables cannot be set and the requests are wrong.

![image](https://github.com/user-attachments/assets/afb13ff8-6968-4bd0-b5e7-ad15491a26d2)
![image](https://github.com/user-attachments/assets/23954196-2874-43e6-a851-7a63791eac11)

### What did you expect to happen?

I would expect one of two things, either :
- the client would use the variables set in the documentation
- or the variables of the server url would appear in the path parameters to be set in the client

marclave:
in progress here https://github.com/scalar/scalar/pull/3553

guillaumefont:
Thanks @marclave !!

PR #3323, opened 2024-10-01
feat(galaxy): add variables to the second server, see #3317

hanspagel:
This PR adds variables to the second server of our OpenAPI example, great do debug #3317.


amritk (approved):
Nice work! Is it working in the client as well?

hanspagel:
Nope, this doesn’t fix the issue(s)! I thought we should do this in a separate PR.

marclave (approved):
nice 😎

PR #3324, opened 2024-10-01
feat(api-client): command palette navigation

antlio:
this pr adds the following enhancements:
- focus on first result when conducting command search
- return on global menu on delete key from a command
- enhances styles and fix border opacity

https://github.com/user-attachments/assets/3435ca91-ddd7-4fc8-83b9-061d4bb9e1c2

### DIFF
--- a/packages/api-client/src/components/CommandPalette/TheCommandPalette.vue
+++ b/packages/api-client/src/components/CommandPalette/TheCommandPalette.vue
@@ -34,7 +34,7 @@ export type CommandNames = keyof typeof PaletteComponents
 
 <script setup lang="ts">
 import { ScalarIcon, useModal } from '@scalar/components'
-import { computed, nextTick, onBeforeUnmount, onMounted, ref } from 'vue'
+import { computed, nextTick, onBeforeUnmount, onMounted, ref, watch } from 'vue'
 import { useRouter } from 'vue-router'
 
 import {
@@ -125,6 +125,14 @@ const closeHandler = () => {
   selectedSearchResult.value = -1
 }
 
+/** Reset state on back */
+const backHandler = () => {
+  // Prevent delete event from removing query command character
+  event.preventDefault()
+  activeCommand.value = null
+  nextTick(() => commandInputRef.value?.focus())
+}
+
 /** Handle execution of the command, some have routes while others show another palette */
 const executeCommand = (
   command: (typeof availableCommands)[number]['commands'][number],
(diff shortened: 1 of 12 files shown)


hanspagel:
Great improvement! In Raycast, when I go back (backspace) it still has the search query and the previous results listed, I like that. :)

Screenshot 2024-10-01 at 14 38 21

hanspagel (approved):
Great improvement

packages/api-client/src/components/CommandPalette/TheCommandPalette.vue
```diff
@@ -153,6 +161,21 @@ const openCommandPalette = ({
   nextTick(() => commandInputRef.value?.focus())
 }
 
+/** Focus on first result when conducting a search */
+watch(commandQuery, (newQuery) => {
+  if (newQuery && searchResultsWithPlaceholderResults.value.length > 0) {
+    selectedSearchResult.value = 0
+    nextTick(() => {
```
marclave, line 168: nit but we could do async await on next time to keep the indentation down :) but non blocking for this PR

```
watch(commandQuery, async (newQuery) => {
  if (newQuery && searchResultsWithPlaceholderResults.value.length > 0) {
    selectedSearchResult.value = 0
    await nextTick()

```

amritk (changes requested):
Cannot delete any characters in the command pallets. We need to add a check for `string length === 0` before going back

marclave:
feel free to merge after @amritk 's feedback is addressed @antlio <3

packages/api-client/src/components/CommandPalette/TheCommandPalette.vue
```diff
@@ -153,6 +161,21 @@ const openCommandPalette = ({
...
+      commandRefs.value[selectedSearchResult.value]?.scrollIntoView({
+        behavior: 'smooth',
+        block: 'center',
```
hwkr, line 171: Should this be `nearest` like the search modal?

antlio:
thanks for the great feedbacks, i added the following enhancements:
- maintain search query on back
- keep focus item on back
- only fire onDelete event on empty input

https://github.com/user-attachments/assets/348282d3-3c1e-4b54-a496-fffb1f98d09c

packages/api-client/src/components/CommandPalette/TheCommandPalette.vue
amritk: Nearest all the things

amritk (approved):
Copilot says it's clear

PR #3327, opened 2024-10-01
chore: move `client.scalar.com` to a new projects folder

hanspagel:
A few people were confused because `client.scalar.com` is in `examples/api-client`, and I can totally understand the confusion.

This PR moves it to its own `projects/` folder and renames it to `client-scalar-com/`, this should make clear what’s inside.

The idea is that we’ll get rid of `examples/` at some point, moving those to the package folders (as “playgrounds”) and keep user-facing stuff in `projects/`.

If you all agree, and this is merged, I’d move the Electron app to `projects/app`, too.


amritk (approved):
I'm all for it! Could also be called apps but that doesn't matter

hwkr:
> A few people were confused because client.scalar.com is in examples/api-client, and I can totally understand the confusion.

Me last week

hanspagel:
Files can easily be moved back (in case someone has a different opinion), so let’s get this in. :)

PR #3305, opened 2024-09-30
fix(api-client): uppercase method on request send, fix #3303

marclave:
fixes https://github.com/scalar/scalar/issues/3303

as per https://www.rfc-editor.org/rfc/rfc7231#section-4.1

before:

```
curl 'https://sandbox.scalar.com/test' \
  -X 'patch' \
  -H 'accept: */*' \
  -H 'accept-language: en-US,en;q=0.9' \
  -H 'content-length: 0' \
  -H 'content-type: application/json' \
  -H 'origin: https://sandbox.scalar.com' \
  -H 'priority: u=1, i' \
  -H 'referer: https://sandbox.scalar.com/p/GcxDQ' \
  -H 'sec-ch-ua: "Google Chrome";v="129", "Not=A?Brand";v="8", "Chromium";v="129"' \
  -H 'sec-ch-ua-mobile: ?0' \
  -H 'sec-ch-ua-platform: "macOS"' \
  -H 'sec-fetch-dest: empty' \
  -H 'sec-fetch-mode: cors' \
  -H 'sec-fetch-site: same-origin' \
  -H 'user-agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/129.0.0.0 Safari/537.36'
```

after

```
curl 'https://sandbox.scalar.com/test' \
  -X 'PATCH' \
  -H 'accept: */*' \
  -H 'accept-language: en-US,en;q=0.9' \
  -H 'content-length: 0' \
  -H 'content-type: application/json' \
  -H 'origin: https://sandbox.scalar.com' \
  -H 'priority: u=1, i' \
  -H 'referer: https://sandbox.scalar.com/p/GcxDQ' \
  -H 'sec-ch-ua: "Google Chrome";v="129", "Not=A?Brand";v="8", "Chromium";v="129"' \
  -H 'sec-ch-ua-mobile: ?0' \
  -H 'sec-ch-ua-platform: "macOS"' \
  -H 'sec-fetch-dest: empty' \
  -H 'sec-fetch-mode: cors' \
  -H 'sec-fetch-site: same-origin' \
  -H 'user-agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/129.0.0.0 Safari/537.36'
```

PR #3328, opened 2024-10-01
chore(api-client): add more tests

hanspagel:
This PR adds a few more tests around `createRequestOperation` and the used helpers.

Just one code change: We’re trimming whitespace from the header name now.

Note: I’ve also added one to test for the uppercase request method added in #3305, but I didn’t enable it, because it requires more code changes. Will ff with a separate PR.


marclave (approved):
✨ 

we love tests

PR #3331, opened 2024-10-01
ci: skip update-contributors workflow on forks

xC0dex:
**Problem**
This workflow runs at 3pm also on forks, which can be annoying as the job will fail.
image

**Solution**
This PR prevents the workflow from running on forks.

### DIFF
--- a/.github/workflows/update-contributors.yml
+++ b/.github/workflows/update-contributors.yml
@@ -7,6 +7,7 @@ on:
 jobs:
   update-readme:
     name: Update the README
+    if: github.ref == 'refs/heads/main'
     runs-on: ubuntu-latest
 
     steps:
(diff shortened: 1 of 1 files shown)


hanspagel:
This is great! But we should use the same check as here. https://github.com/scalar/scalar/pull/3310

xC0dex:
@hanspagel, I didn't know that you did the same thing for different workflows yesterday 🤣. In the related issue, the `github.event.pull_request` variable is used. I don't know how such an event looks like for a scheduled workflow 👀 Maybe something like this: `if: github.event.repository.fork == false`?

hanspagel:
Oh, I don’t know either. `github.repository_owner` looks good!

hanspagel (approved):

PR #3333, opened 2024-10-02
feat(Scalar.AspNetCore): add missing favicon configuration

diegolopes:
**Problem**
We got a configuration for favicons, as we can read [here](https://github.com/scalar/scalar/blob/main/documentation/configuration.md).
But for the package [Scalar.AspNetCore](https://www.nuget.org/packages/Scalar.AspNetCore), we don't have that option, leaving the documentation with no favicon at all when configuring Scalar on a new project.

image

**Solution**
With this PR, now we can specify an URL/Path to the favicon, and you can use just like any other option that is on [ScalarOptions](https://github.com/scalar/scalar/blob/main/packages/scalar.aspnetcore/src/Scalar.AspNetCore/Options/ScalarOptions.cs):

```c#
{
    options
        .WithTitle("My title")
        .WithFavicon("/favicon.png")
       // ... other options
});
``` 
I also added an example on the [playground project](https://github.com/scalar/scalar/tree/main/packages/scalar.aspnetcore/playground/Scalar.AspNetCore.Playground).


xC0dex:
Great work on this PR! Everything looks good 🥳. I double-checked the config twice and missed the Favicon 😆.
One area for potential improvement: instead of providing the static file via an endpoint, consider using the `wwwroot` folder and the `app.UseStaticFiles()` middleware. It uses built-in features and keeps the code cleaner IMO, even if it's only the playground.

hanspagel:
That’s a great find! Thanks for taking the time to contribute. Can we add a test here?

xC0dex (approved):
LGTM 🚀

diegolopes:
@xC0dex Yeah, `app.UseStaticFiles()` is the correct approach, I refactored the code to use it properly.

@hanspagel I've added some missing tests as well.

marclave (approved):
✨

marclave:
awesome work @diegolopes !!

xC0dex:
It looks like this PR and https://github.com/scalar/scalar/pull/3344 updated the same test which causes the aspnetcore test job to fail 👀

PR #3334, opened 2024-10-02
fix(api-reference): render falsely default values

### LINKED ISSUE #3314: Default values aren't rendered when they are False or 0
### What happens?

We have parameters defined in our FastAPI app like this

```
    page: int = Field(
        default=0,
        examples=[0],
        description="Page number",
    )
    limit: int = Field(
        default=25,
        examples=[25],
        description="Number of results per page.Default value is 25. Can't be higher than 500",
    )
```

that in our [opanapi.json file ](https://api.theirstack.com/openapi.json) render look like this
```
          "page": {
            "type": "integer",
            "title": "Page",
            "description": "Page number",
            "default": 0,
            "examples": [
              0
            ]
          },
          "limit": {
            "type": "integer",
            "title": "Limit",
            "description": "Number of results per page.Default value is 25. Can't be higher than 500",
            "default": 25,
            "examples": [
              25
            ]
          },
```

But then, in the interactive docs on api.theirstack.com, that uses Scalar, only defaults of fields with a default value that doesn't coerce to `false` in js are properly rendered. So the default for `limit` appears, but the one for `page` doesn't:

image

The same happens for boolean attributes. Only those where the default is `true` show a default value:

image

```
job_description_pattern_is_case_insensitive: bool | None = Field(
    True,
    description="""Whether to make searches on job descriptions case-sensitive or insensitive.""",
)
remote: bool | None = Field(
    None,
    examples=[None],
    description="True: only show remote jobs. False: only show non-remote jobs. None: show all jobs.",
)
```

All fields that have a default value defined should show that value. 

Redoc does this well - check it out at https://api.theirstack.com/redoc#tag/Jobs/operation/search_jobs_v1_jobs_search_post:
image

So does Openapi - https://api.theirstack.com/openapi#/Jobs/search_jobs_v1_jobs_search_post
image

### OpenAPI Document

https://sandbox.scalar.com/e/VaXXX

hwkr:
Fixes #3314 

## Before

Google Chrome-2024-10-01-20-32-05@2x

## After

Google Chrome-2024-10-01-20-40-05@2x

### DIFF
--- a/.changeset/mean-penguins-call.md
+++ b/.changeset/mean-penguins-call.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
+fix(api-reference): check if default value is undefined instead of falsy
--- a/packages/api-reference/src/components/Content/Schema/SchemaPropertyHeading.spec.ts
+++ b/packages/api-reference/src/components/Content/Schema/SchemaPropertyHeading.spec.ts
@@ -1,0 +1,62 @@
+import { mount } from '@vue/test-utils'
+import { describe, expect, it } from 'vitest'
+
+import SchemaPropertyHeading from './SchemaPropertyHeading.vue'
+
+describe('SchemaPropertyHeading', () => {
+  it('renders falsy default values', async () => {
+    const wrapper = mount(SchemaPropertyHeading, {
+      props: {
+        value: {
+          type: 'boolean',
+          default: false,
+        },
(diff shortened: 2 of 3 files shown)


amritk (approved):

hanspagel:
This is great! I’ve added a few tests. ✨

hanspagel (approved):

PR #3340, opened 2024-10-02
chore: remove credentials min length error

hanspagel:
There’s an error in the console when the auth credentials are shorter than 3 characters. It’s saying that we won’t mask the credentials in that case.

I can’t think of any scenario where 2 characters are confidential anyway and no one will notice when it’s in the browser console, I think we can remove it.


amritk (approved):
Yea a silly one, don't forget the changelog!

hanspagel:
A changelog is not necessary, it doesn’t change any logic, it doesn’t have to trigger a release and doesn’t have to be in the changelog. That said, we should have a changeset for 99 % of the changes, that’s maybe the 1 % here. 😁

PR #3345, opened 2024-10-02
chore: added lint rule for function definitions

amritk:
Before it never picked picked up most functions, if this is too noisey we can turn it off but I think all functions should have jsdoc comments

before:
image

after:
image


hwkr:
Ooof this is going to add a lot of warnings to existing code

amritk:
> Ooof this is going to add a lot of warnings to existing code

Yep that is the point!
image

but its just a warning so it should only annoy you and not block anything

marclave:
im in

marclave (approved):

PR #3351, opened 2024-10-03
feat(api-client): remove tabs on client.scalar.com

hwkr:
After chatting with @marclave we decided to remove the tabs on `client.scalar.com` and only have them in the electron app similar to what linear does.

It made the most sense to me to split these into two packages one for the app and one for the web version.

## Web

https://github.com/user-attachments/assets/0d838c66-c5b1-4816-bf84-c7842537b7c2

## Desktop

https://github.com/user-attachments/assets/218c2dec-4dae-4e55-815e-7ae9163f17df

### DIFF
--- a/packages/api-client/src/layouts/Web/ApiClientWeb.vue
+++ b/packages/api-client/src/layouts/Web/ApiClientWeb.vue
@@ -29,9 +29,7 @@ const { isDark } = useDarkModeState()
 const workspaceStore = useWorkspace()
 
 // Ensure we add our scalar wrapper class to the headless ui root
-onBeforeMount(async () => {
-  addScalarClassesToHeadless()
-})
+onBeforeMount(() => addScalarClassesToHeadless())
 
 /** Handles the hotkey events, we will pass in custom hotkeys here */
 const handleKeyDown = (ev: KeyboardEvent) => handleHotKeyDown(ev)
--- a/packages/api-client/src/layouts/Web/ApiClientWeb.vue
+++ b/packages/api-client/src/layouts/Web/ApiClientWeb.vue
@@ -3,9 +3,8 @@ import { TheCommandPalette } from '@/components/CommandPalette'
 // TODO: Disabled until we polished the UI.
 // import { ImportCollectionListener } from '@/components/ImportCollection'
 import SideNav from '@/components/SideNav/SideNav.vue'
-import TopNav from '@/components/TopNav/TopNav.vue'
 import { useDarkModeState } from '@/hooks'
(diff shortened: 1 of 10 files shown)


packages/api-client/src/layouts/Web/ApiClientWeb.vue
```diff
@@ -0,0 +1,96 @@
...
+  })
+})
+
+const { isDark } = useDarkModeState()
+const workspaceStore = useWorkspace()
+
+// Ensure we add our scalar wrapper class to the headless ui root
+onBeforeMount(async () => {
+  addScalarClassesToHeadless()
+})
```
amritk, line 34:
```suggestion
onBeforeMount(() => addScalarClassesToHeadless())
```
😄

amritk:
Excellent idea! Will test more thoroughly later, code looks good though

packages/api-client/src/layouts/Web/ApiClientWeb.vue
hwkr: Yep that makes sense, tbh I copied this from the App layout

amritk: barely makes a difference, just slightly cleaner looking.

amritk: I do notice we use async in a few places where we don't really wait for anything 😄

hwkr: I feel like we should have a linter for that 🤔

amritk (approved):

hanspagel (commented):
The idea itself is amazing, and the code looks good, too.

I’m a bit hesitant with the new entry point. Isn’t it better to stick with the same entrypoint but just add a few conditions around the hotkeys and the components? Or do I underestimate the number of things we need to take into account?

Not a strong opinion, just a feeling. We somehow need to look at each entry point with every release, and this seems like an additional risk to miss bugs. 

Wdyt?

hwkr:
> I’m a bit hesitant with the new entry point. Isn’t it better to stick with the same entrypoint but just add a few conditions around the hotkeys and the components? Or do I underestimate the number of things we need to take into account?

Personally I always lean towards composition over configuration where possible, it is an extra thing to test but so would be a config setting and I feel like the config setting is easier to miss when testing (but yeah we should add some tests for the different entrypoints regardless)

hanspagel:
Yeah, let’s just merge this one here. Probably better!

PR #3353, opened 2024-10-03
fix: minimize some ui clutter, clean up some type styles

cameronrohani:
before:
image

after:
image
image

prototypey but ideally ya'll play with it and let me know what you think in slack

### DIFF
--- a/packages/api-client-app/src/main/index.ts
+++ b/packages/api-client-app/src/main/index.ts
@@ -71,7 +71,7 @@ function createWindow(): void {
     height: mainWindowState.height,
     show: false,
     title: 'Scalar',
-    trafficLightPosition: { x: 9.5, y: 12 },
+    trafficLightPosition: { x: 17, y: 18 },
     // Borderless Window, for macOS only
     ...(process.platform === 'darwin' ? { titleBarStyle: 'hidden' } : {}),
     autoHideMenuBar: true,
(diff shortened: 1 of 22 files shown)


antlio:
great work, its cleaner and clearer I especially like the environment dropdown position and address bar one!

on tabs: i'm wondering if we shouldn't maintain current hierarchy, as we have a different address bar url input per tabs, it might be better to keep having it above to maintain readability flow?

on sidebar toggle: i like it there however this might implies that it works for every sidebar, which as of today is not the case (only the request sidebar can be) should we allow it for each sidebar views? and also it a bit far from it so we might wonder what it would toggle at first.

cameronrohani:
i'm fine to leave the url bar up top but the reason i prefer not doing that is so that we only take up the extra space if you want to use tabs

if we put tabs up top  we'll always have extra space up top on the app whether if people are using tabs or not

here's a prototype of them up top

image

in regards to making the sidebar toggle-able for each page most likely should.... that being said I think we should probably only show sidebar toggle at a certain screen size ++ if the sidebar is ever closed - i'm not sure if there's much reaason to close the sidebar if you're on a larger screen and I've seen this ux on other apps before

marclave:
closing in favour of some new stuff @cameronrohani and i are cooking 😈

PR #3372, opened 2024-10-04
chore(Scalar.AspNetCore): Switch editorconfig EOL config to lf

xC0dex:
### Description

This PR switches from `crlf` to `lf` EOL-config in the `packages/scalar.aspnetcore/.editorconfig` file to satisfy the `pnpm prettier` command. This should prevent any further format issues in the `Scalar.AspNetCore` solution.


marclave (approved):
nice 😎

PR #3374, opened 2024-10-04
feat(api-client): OpenAPI preview and error handling for the import

hwkr:
Started by trying to surface the errors and then ended up polishing the whole UX.

Changes:

- Shows errors when the spec fails to import or fetch
- Shows a preview (again) when you paste a spec
- Doesn't try to fetch bad URLs
- Shows a loading state while importing
- Uses a text area for the form so you can write multiple lines and lines wrap in the input area (still submits the form on enter)

## Before

https://github.com/user-attachments/assets/ea44c846-05bd-4ca3-b1a6-624b460e5bc2

## After

https://github.com/user-attachments/assets/99e8bd5d-f2cd-4ef5-bac0-dfe89a593073


hanspagel (commented):
OMG, nice! Just one thing: The CodeBlock font size looks a little bit too big compared to the other text in the modal. Should we make this a little bit smaller?

hwkr:
> OMG, nice! Just one thing: The CodeBlock font size looks a little bit too big compared to the other text in the modal. Should we make this a little bit smaller?

Oh yeah, I can make it a bit smaller 😄

antlio (approved):
great job! tested several use cases works like a charm.

we could also handle bad pasted specification with an error toast, but can be added later

hanspagel:
It’s really slow with big documents (Stripe for example: https://raw.githubusercontent.com/stripe/openapi/refs/heads/master/openapi/spec3.yaml).

Should we disable the preview for big files, does that help?

hwkr:
> It’s really slow with big documents. Stripe for example: https://raw.githubusercontent.com/stripe/openapi/refs/heads/master/openapi/spec3.yaml
> 
> Should we disable the preview for big files, does that help?

Yeah we could I'm not sure how tot detect that though on paste 🤔

hwkr:
@hanspagel Alright I investigated the pasting big files. Basically our call to `JSON.parse` is slow and there isn't anyway to make it async without using a web worker. I tried turning off the parsing on paste both in general and for large files and the textarea is also laggy when you have that much data in it and lags when you delete characters.

I think long term we should set up a web worker for our parser so we can show a loading state when processing large files. Short term locking the UI and showing the preview is actually not the worst since it avoids user errors and doesn't give them a laggy textbox (imo).

hanspagel (approved):
Thanks for looking it! Let’s get this in then! :)

PR #3376, opened 2024-10-05
fix(api-client): Basic Auth header format for empty passwords

### LINKED ISSUE #3124: API Client: Basic Auth credentials not properly encoded when password is empty
### What happens?

When using the built-in API client with Basic Authentication and providing only a username with an empty password, an Authorization header is incorrectly generated. 
The current behaviour produces a Bearer token instead of the expected Basic Auth format.

I am using package `@scalar/hono-api-reference ^0.5.146`

The API client's Authorization header should be prefixed with "Basic"
Credentials (username:password) should be Base64 encoded, even when the password is empty

### OpenAPI Document

_No response_

inyourtime:
fixes #3124 

This PR addresses an issue with Basic Authentication header formatting when the password is empty. According to [RFC 7617](https://datatracker.ietf.org/doc/html/rfc7617), the Basic Auth header must include the colon `(:)` separator between the username and password, even if the password is empty.

- Ensured that the username:password format is maintained by appending a colon after the username, even when the password is empty.

This change ensures compliance with the Basic Auth standard and prevents potential authorization issues when no password is supplied.

### DIFF
--- a/.changeset/plenty-dancers-breathe.md
+++ b/.changeset/plenty-dancers-breathe.md
@@ -1,0 +1,6 @@
+---
+'@scalar/api-client': patch
+---
+
+fix: Basic Auth should still include a colon after the username  
+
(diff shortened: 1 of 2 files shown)


amritk (approved):
Excellent find/fix

marclave (approved):
unreal fix @inyourtime

PR #3379, opened 2024-10-05
feature(Scalar.AspNetCore): support local resource files 

xC0dex:
### Todos
- Should this feature be opt-in or opt-out? Opt-out would be breaking (behavior) change as a custom `CdnUrl` would no longer be used by default.
- How to provide a `standalone.js` file into the `StaticAssets` folder?

### Description

This PR adds support for a local resource files by introducing the `LocalResources` property. This feature was requested in #3357 and I think it's expected for .NET devs to provide support for this.
When setting this property to `true` it will no longer use the `CdnUrl`. Instead, a new endpoint will be added that provides all static assets (Currently only `standalone.js`). Furthermore, no fonts from CDNs will be used.
The `standalone.js` would be located in the new `StaticAssets` folder. It's provided via the route `/scalar-assets/{file}`. This can be changed by the user using the `LocalResourcesRoutePattern` property.


hanspagel:
Yeah, we’d need to add `@scalar/api-reference` to the package.json of the .NET package to trigger updates. We’re basically releasing a new version on most days, so that would trigger a lot of .NET releases … We tend to use the CDN only.

Wdyt @marclave?

xC0dex:
> so that would trigger a lot of .NET releases

I would also say that it's relatively uncommon for a NuGet package to have many patch releases in a short period of time. However, as I mentioned in the PR description, I believe .NET users generally expect everything to be bundled inside the package. The reason for this expectation is that currently, both `NSwag` and `Swashbuckle` include the Swagger UI in their packages. In fact, I think **every** .NET API project used one of these packages, as they were the standard way to display the generated OpenAPI documentation.

What do you think about the idea of bundling the latest `standalone.js` in the NuGet package when we release a new version of `Scalar.AspNetCore`? We could make this feature opt-in, so by default, the CDN with the latest version would be used. If developers prefer to use the bundled version, they could enable it, though they might not always get the latest version.

marclave:
> > so that would trigger a lot of .NET releases
> 
> I would also say that it's relatively uncommon for a NuGet package to have many patch releases in a short period of time. However, as I mentioned in the PR description, I believe .NET users generally expect everything to be bundled inside the package. The reason for this expectation is that currently, both `NSwag` and `Swashbuckle` include the Swagger UI in their packages. In fact, I think **every** .NET API project used one of these packages, as they were the standard way to display the generated OpenAPI documentation.
> 
> What do you think about the idea of bundling the latest `standalone.js` in the NuGet package when we release a new version of `Scalar.AspNetCore`? We could make this feature opt-in, so by default, the CDN with the latest version would be used. If developers prefer to use the bundled version, they could enable it, though they might not always use the latest version.

so we would release static on minor releases and not for every patch? :) 

i think this all sounds good to me ✨

xC0dex:
> so we would release static on minor releases and not for every patch?

Now I'm confused 🤣. Could you explain this, please? My idea was that we **always** put in the latest `standalone.js` into the `Scalar.AspNetCore` package when create a new release of `Scalar.AspNetCore`. And we only release a new version of `Scalar.AspNetCore` when we have a change in it **or** when the last release is so old that we should update the `standalone.js`  in it.

marclave:
> > so we would release static on minor releases and not for every patch?
> 
> Now I'm confused 🤣. Could you explain this, please? My idea was that we **always** put in the latest `standalone.js` into the `Scalar.AspNetCore` package when create a new release of `Scalar.AspNetCore`. And we only release a new version of `Scalar.AspNetCore` when we have a change in it **or** when the last release is so old that we should update the `standalone.js` in it.

omg haha my bad, im kind of sick right now and my brain was definitely not working well.

ohhh okay, how do we decide when the last release is old enough? :)

xC0dex:
> omg haha my bad, im kind of sick right now and my brain was definitely not working well.

Stop working when you're sick! 🤧 
 
> ohhh okay, how do we decide when the last release is old enough? :)

That’s a great question. I haven’t really given it much thought yet. I’m aiming to find a good balance between 'releasing frequently' and 'only releasing when there are actual changes to the package.' We _could_ also think about a scheduled release? So, e.g., update the `Scalar.AspNetCore` package every 2 weeks if the dependency (`standalone.js`) changed.

How do we handle this for current integrations?

hanspagel:
Currently, most integrations use the CDN! I think it’s only the Fastify integration that bundles the JS.

I feel we’re shipping pretty fast, and people benefit from “over the air” updates, but that might change over the next year and it’s ok to get into a slower release cycle (for most things) eventually.

Anyway, no right answer here!

xC0dex:
Alright, I have one last idea 👀. Let’s ship `Scalar.AspNetCore` with every release of the API reference, but as a patch release. We can update the documentation to recommend installing the` Scalar.AspNetCore` package with the following command:

```bash
dotnet add package Scalar.AspNetCore -v 1.2.*
```

This command will add the dependency with a wildcard in the patch version. So it will always install the latest available version, without any noise.

hanspagel:
Perfect! This is actually what the changeset workflow does automatically (as long as `@scalar/api-reference` is a dependency in `package.json`).

xC0dex:
Great, looks like we have an approach! My question now is: How can we get the` standalone.js` file into the `StaticAssets` folder? In other projects, I downloaded artifacts from a previous job.

hanspagel:
For the Fastify integraton we’re copying it into the source folder:

https://github.com/scalar/scalar/blob/main/packages/fastify-api-reference/vite.config.ts#L11-L18

For the `Scalar.AspNetCore` we could just write a bash script. Maybe we use a package.json `"build"` script to run the bash script and then run `dotnet build`.

hanspagel (approved):
Awesome PR as always!

TIL `prepare` is a reserved name for npm scripts

Anyway, good to merge once CI is green. :)

PR #3283, opened 2024-09-27
style(themes): new selection colors

antlio:
big friday energy inspired by @JohnPhamous amazing [tweet](https://x.com/JohnPhamous/status/1770861726604419369). gave a first shot at our custom theme selection but feel free to play with it, as seen below:

<details>
<summary>alternate</summary>
image
image
</details>

<details>
<summary>default</summary>
image
image
</details>

<details>
<summary>moon</summary>
image
image
</details>

<details>
<summary>purple</summary>
image
image
</details>

<details>
<summary>solarized</summary>
image
image
</details>

<details>
<summary>blue planet</summary>
image
image
</details>

<details>
<summary>deep space</summary>
image
image
</details>

<details>
<summary>saturn</summary>
image
image
</details>

<details>
<summary>kepler 11e</summary>
image
image
</details>

<details>
<summary>mars</summary>
image
image
</details>

PR #3385, opened 2024-10-07
fix(api-client): cm focused selection colors

antlio:
this pr fixes the code mirror selection color mismatch caught by @hwkr by setting the introduced variables in #3283:

**before**
image

**after**
image


marclave (approved):
nice! the proper code mirror selection 🫡

PR #3387, opened 2024-10-07
feat(nextjs-api-reference): pin cdn version

tmastrom:
This PR updates the configuration to allow the user to specify the `@scalar/api-reference` CDN version in the `next-js` integration. It also adds documentation for pinning the version in other integrations.

Maybe we add the cdn url as an option in the universal configuration object?

It was proposed to add an option in vite config to pin the versions but I'm not seeing any specific advantages over specifying the version in the script tag


hanspagel (approved):
Long overdue! Thank you!

marclave (approved):

PR #3389, opened 2024-10-07
fix(client-scalar-com): favicon does not show

hanspagel:
The favicon doesn’t show on <https://client.scalar.com>.

Looks like we switched from `scalar.com` to `www.scalar.com` and back to `scalar.com`. This PR updates the Content Security Policy to allow both, and points to the non-www URL (which is the current default).


amritk (approved):
Nice catch

PR #3394, opened 2024-10-07
chore: nuget security updates only

hanspagel:
I love dependabot, but let’s limit it to nuget security updates for now. :)


xC0dex:
I think it's fine. We only have one dependency in the package itself. All the other dependencies are either in the playground or in the test project. 

> I love dependabot

If you love dependabot you would like to marry [Renovate](https://docs.renovatebot.com/) 🤣.

xC0dex (approved):

PR #3395, opened 2024-10-07
fix(themes): selection color variables to presets

### LINKED ISSUE #3388: Theming stopped working in the React wrapper
### What happens?

After switching to the latest version 0.3.104 of @scalar/api-reference-react, themes stopped working. Switching between light and dark mode still works, but the custom background from themes like "kepler" don't apply anymore.

The last version that worked as expected was 0.3.19.

I expected to get the same theme styling as per examples and documentation when applying it on the configuration.

```javascript
'use client';
import React from 'react';

type Props = {
    spec: Record<string, any>;
};

const Docs = ({ spec }: Props): React.ReactElement => {
    return (
        <ApiReferenceReact
            configuration={{
                spec: {
                    content: spec
                },
                layout: 'modern',
                theme: 'kepler',
                hideDownloadButton: true,
                hideModels: true,
                darkMode: true,
            }}
        />
    );
};

export default Docs;
```

### OpenAPI Document

_No response_

antlio:
this pr moves back the selection color variables to presets in order to avoid adding background variable at the variables level that breaks theming, fixes #3388

### DIFF
--- a/.changeset/thick-experts-behave.md
+++ b/.changeset/thick-experts-behave.md
@@ -1,0 +1,5 @@
+---
+'@scalar/themes': patch
+---
+
+fix: moves selection color variables to presets
--- a/packages/themes/src/presets/alternate.css
+++ b/packages/themes/src/presets/alternate.css
@@ -18,6 +18,13 @@
   --scalar-background-accent: var(--scalar-background-3);
 
   --scalar-border-color: rgba(0, 0, 0, 0.1);
+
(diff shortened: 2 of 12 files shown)


amritk (approved):
Will defer to @hwkr

hwkr:
Alright I think this is actually an issue with us loading the variables css wrong, going to open another PR.

hwkr:
Okay opened a fix for the themes layer issues in #3398

PR #3398, opened 2024-10-07
fix(themes): remove variables from theme layer and move selection to default theme

### LINKED ISSUE #3388: Theming stopped working in the React wrapper

hwkr:
Looks like we were injecting a copy of `variables.css` from `@scalar/themes` into the `scalar-theme` layer, we already import `variables.css` into the `scalar-base` layer as part of `@scalar/themes/style.css`.

It's important it's only in the `scalar-base` layer otherwise there's a risk of the default variables overriding theme variables (which is the issue in #3388). This updates the precedence to be:

1. Any user styles that they add to the page outside of a css layer
2. The theme injected into the `scalar-theme` css layer via [`getThemeStyles()`](https://github.com/scalar/scalar/blob/6279c5977c5067f21c61d4d91d7963f4a4b9a384/packages/themes/src/index.ts#L114) (includes any selection color overrides)
3. The variables & default (base) theme injected into the `scalar-base` css layer by importing [`@scalar/themes/style.css`](https://github.com/scalar/scalar/blob/6279c5977c5067f21c61d4d91d7963f4a4b9a384/packages/themes/src/style.css) (includes the default selection color configuration)

I also moved the base / default css variables for the selections into the default theme because they're more colors than root variables (they should work in either file now though). 

Fixes #3388

## Before

Google Chrome-2024-10-07-13-55-04@2x

## After

Google Chrome-2024-10-07-13-29-32@2x

### DIFF
--- a/packages/api-client/src/layouts/App/ApiClientApp.vue
+++ b/packages/api-client/src/layouts/App/ApiClientApp.vue
@@ -54,19 +54,16 @@ const handleKeyDown = (ev: KeyboardEvent) => handleHotKeyDown(ev, { hotKeys })
 onMounted(() => window.addEventListener('keydown', handleKeyDown))
 onBeforeUnmount(() => window.removeEventListener('keydown', handleKeyDown))
 
-const fontsStyleTag = computed(
+const themeStyleTag = computed(
   () =>
     workspaceStore.activeWorkspace.value &&
-    `<style>
-  ${getThemeStyles(workspaceStore.activeWorkspace.value.themeId, {
-    fonts: true,
-  })}</style>`,
+    `<style>${getThemeStyles(workspaceStore.activeWorkspace.value.themeId)}</style>`,
 )
 </script>
 <template>
   <!-- Listen for paste and drop events, and look for `url` query parameters to import collections -->
   <!-- <ImportCollectionListener> -->
-  <div v-html="fontsStyleTag"></div>
+  <div v-html="themeStyleTag"></div>
   <TopNav :openNewTab="newTab" />
 
   <!-- Ensure we have the workspace loaded from localStorage above -->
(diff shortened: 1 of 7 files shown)


packages/api-client/src/layouts/App/ApiClientApp.vue
```diff
@@ -54,19 +54,16 @@ const handleKeyDown = (ev: KeyboardEvent) => handleHotKeyDown(ev, { hotKeys })
 onMounted(() => window.addEventListener('keydown', handleKeyDown))
 onBeforeUnmount(() => window.removeEventListener('keydown', handleKeyDown))
 
-const fontsStyleTag = computed(
+const themeStyleTag = computed(
```
hwkr, line 57: Also renamed these to reflect what they're doing

antlio (approved):
great job!!

hanspagel (approved):
Great fix, awesome explanation!

cameronrohani (approved):

PR #3399, opened 2024-10-08
fix(api-reference): show path and test request when hiddenClients is true

### LINKED ISSUE #3380: When using hiddenClients: true, the endpoint uri is also hidden
### What happens?

Currently the http method + uri is part of the code snippet section, so when using the `hiddenClients: true` configuration option, it's not possible to see what uri should be used.

The endpoint e.g. `POST /newresource` is still visible when using the `hiddenClients: true` configuration option. Perhaps it can be pulled out of the right section and added under the endpoint title/description

### OpenAPI Document

_No response_

hwkr:
Cleans up the UI when there aren't any clients / example so we still show a path and the test request button.

Fixes #3380 

## Before / After

Google Chrome-2024-10-07-22-32-41@2x
Google Chrome-2024-10-07-21-50-11@2x

Google Chrome-2024-10-07-22-32-47@2x
Google Chrome-2024-10-07-22-23-43@2x

Google Chrome-2024-10-07-22-32-55@2x
Google Chrome-2024-10-07-22-29-07@2x

### DIFF
--- a/.changeset/friendly-rocks-wait.md
+++ b/.changeset/friendly-rocks-wait.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
+fix(api-reference): show path and test request when hiddenClients is true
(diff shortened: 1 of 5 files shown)


hanspagel (approved):
Cool!

cameronrohani (approved):

PR #3400, opened 2024-10-08
fix(api-client): address bar placeholder

antlio:
this pr updates placeholder display according to active server presence:

**before**
image

**after**
image


hanspagel (approved):
So fresh, so clean ✨

marclave (approved):
awesome :)

PR #3401, opened 2024-10-08
fix(api-client): command palette label

antlio:
this pr removes the command palette commands label as suggested by @marclave + increases search icon size to maintain icon alignment consistency, as seen below:

**before**
image

**after**
image

### DIFF
--- a/.changeset/clever-baboons-confess.md
+++ b/.changeset/clever-baboons-confess.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
+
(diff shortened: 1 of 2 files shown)


hanspagel (approved):
Slick!

marclave:
looks good to me removing the text, but adding @cameronrohani about increasing icon sizes

cameronrohani:
nice work! 

I think we should make icons text-c-2 to put slightly more focus on the words

amritk (approved):
Lets go

PR #3404, opened 2024-10-08
fix: add export default to hono

### LINKED ISSUE #2994: No "exports" main defined
When using the `@scalar/hono-api-reference` with tsx it seems to throw the following error:

Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: No "exports" main defined in /node_modules/@scalar/hono-api-reference/package.json

More context: #2993

amritk:
closes #2994


amritk:
@geoffgscott going to adapt the build tooling instead

PR #3407, opened 2024-10-09
fix(object-utils): setNested function doesn’t work with nested arrays

amritk:
The `setNestedValue` function didn't work for the test case I added (required for live sync). I got claude to fix it 

Also added a path parsing function which will be used to type the diffs in live sync

Also checked a benchmark and perf has slightly improved + we can remove some extra code with this one when working with formData
```ts
setNestedValue (shallow) x 37,805,234 ops/sec ±6.51% (60 runs sampled)
setNestedValueOld (shallow) x 25,364,159 ops/sec ±4.34% (72 runs sampled)
setNestedValue (deep) x 16,101,160 ops/sec ±3.30% (75 runs sampled)
setNestedValueOld (deep) x 12,860,047 ops/sec ±2.91% (75 runs sampled)
Fastest is setNestedValue (shallow)
```

### DIFF
--- a/.changeset/large-cows-hammer.md
+++ b/.changeset/large-cows-hammer.md
@@ -1,0 +1,5 @@
+---
+'@scalar/object-utils': patch
+---
+
+fix: setNestedValue function and also add new pathParsing funciton
--- a/packages/object-utils/src/nested/nested.test.ts
+++ b/packages/object-utils/src/nested/nested.test.ts
@@ -1,7 +1,7 @@
 import { clone } from '@/clone'
 import { describe, expect, test } from 'vitest'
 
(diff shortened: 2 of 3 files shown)


hanspagel (commented):
I don’t know enough about the context that method is used in, but from reading the tests it seems to be fine. :)

Maybe vitest type tests would be a great addition, example here: https://github.com/scalar/scalar/blob/main/packages/openapi-types/src/openapi-types.test-d.ts

hanspagel (approved):

amritk:
closed in favour of 
https://github.com/scalar/scalar/pull/3417

PR #3409, opened 2024-10-09
fix(api-reference): render tags in tag group order if set

hwkr:
Renders the tags based on the tag group configuration if one is set. I looked at adding the tag group titles but it seemed like maybe too much. We can add them though if we want.

Test file [tagGroups.yaml.zip](https://github.com/user-attachments/files/17302686/tagGroups.yaml.zip)

### DIFF
--- a/.changeset/happy-wolves-glow.md
+++ b/.changeset/happy-wolves-glow.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
+fix(api-reference): render tags in tag group order if set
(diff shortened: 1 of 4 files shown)


hanspagel (changes requested):
Maybe I’m doing it wrong, but this doesn’t render in the correct order:

```yaml
openapi: 3.1.0
info:
  title: Example
  version: '1.0'
tags:
  - name: planets
x-tagGroups:
  - name: galaxy
    tags:
      - foobar
      - planets
paths:
  '/planets':
    get:
      summary: Get all planets
      tags:
        - planets
  '/foobar':
    get:
      summary: Get all foobars
      tags:
        - foobar
```

sidebar has foobar, planets
content has planets, foobar

hwkr:
@hanspagel yeah it was the way I was filtering it, fixed :)

hanspagel (approved):
AWESOME, flawless

PR #3411, opened 2024-10-09
chore(openapi-parser): make `examples` in schemas an array on upgrade, fix #3373

### LINKED ISSUE #3373: Expand getting examples from requestbody
### What happens?

~~Currently we only look at top level examples when generating a request body. HOWEVER schema properties can also have examples.~~

### ~~What did you expect to happen?~~

~~A spec like~~
```json
 "requestBody": {
          "content": {
            "application/json": {
              "schema": {
                "type": "object",
                "properties": {
                  "name": {
                    "type": "string",
                    "example": "Marc"
                  }
                },
                "required": [
                  "name"
                ]
              }
            }
          },
          "required": true
        },
```
~~should correctly add the name 'Marc' in the requestBody~~

Update: Turns out we do display it correctly if it comes in the correct format, unfortunately there is an issue with the upgrader. Maybe we should only upgrade if it is an older version. But also we should fix the upgrader.

It incorrectly changes 
```json
  "requestBody": {
          "content": {
            "application/json": {
              "schema": {
                "type": "object",
                "properties": {
                  "name": {
                    "type": "string",
                    "example": "Marc"
                  }
                },
                "required": [
                  "name"
                ]
              }
            }
          },
          "required": true
        },
```
to the deprecated:
```ts
    "requestBody": {
        "content": {
            "application/json": {
                "schema": {
                    "type": "object",
                    "properties": {
                        "name": {
                            "type": "string",
                            "examples": {
                                "default": "Marc"
                            }
                        }
                    },
                    "required": [
                        "name"
                    ]
                }
            }
        },
        "required": true
    },
```
### OpenAPI Document

_No response_

hanspagel:
Currently, when upgrading an OpenAPI document to 3.1, we’re transforming all `example` attributes to an `examples` object. For schemas, it’s more common to use arrays.

With this PR we’re transforming them to arrays, and not to objects anymore (for schemas only).

Note: Probably easier to review the commits, than just the changed files.

See #3373

### DIFF
--- a/packages/openapi-parser/src/utils/traverse.ts
+++ b/packages/openapi-parser/src/utils/traverse.ts
@@ -5,7 +5,7 @@ import type { AnyObject } from '../types'
  */
 export function traverse(
   specification: AnyObject,
-  transform: (specification: AnyObject, path: string[]) => AnyObject,
+  transform: (specification: AnyObject, path?: string[]) => AnyObject,
   path: string[] = [],
 ) {
   const result: AnyObject = {}
--- a/packages/openapi-parser/src/utils/traverse.ts
+++ b/packages/openapi-parser/src/utils/traverse.ts
@@ -5,25 +5,26 @@ import type { AnyObject } from '../types'
  */
 export function traverse(
   specification: AnyObject,
-  transform: (specification: AnyObject) => AnyObject,
+  transform: (specification: AnyObject, path: string[]) => AnyObject,
+  path: string[] = [],
 ) {
   const result: AnyObject = {}
 
   for (const [key, value] of Object.entries(specification)) {
+    const currentPath = [...path, key]
     if (Array.isArray(value)) {
-      result[key] = value.map((item) => {
+      result[key] = value.map((item, index) => {
         if (typeof item === 'object' && item !== null) {
-          return traverse(item, transform)
+          return traverse(item, transform, [...currentPath, index.toString()])
         }
-
         return item
       })
     } else if (typeof value === 'object' && value !== null) {
-      result[key] = traverse(value, transform)
+      result[key] = traverse(value, transform, currentPath)
     } else {
       result[key] = value
     }
   }
 
-  return transform(result)
+  return transform(result, path)
 }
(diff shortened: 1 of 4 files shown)


amritk (approved):
Lets get it in! Needed for live sync

packages/openapi-parser/src/utils/traverse.ts
```diff
@@ -5,25 +5,26 @@ import type { AnyObject } from '../types'
  */
 export function traverse(
   specification: AnyObject,
-  transform: (specification: AnyObject) => AnyObject,
+  transform: (specification: AnyObject, path: string[]) => AnyObject,
+  path: string[] = [],
```
tmastrom, line 9: The traverse function is also used by 
- `filter`
- `getListOfReferences`
- `upgradeFromTwoToThree`

We should update those or make the path argument optional

packages/openapi-parser/src/utils/upgradeFromThreeToThreeOne.ts
```diff
@@ -6,7 +6,9 @@ import { traverse } from './traverse'
  *
  * https://www.openapis.org/blog/2021/02/16/migrating-from-openapi-3-0-to-3-1-0
  */
-export function upgradeFromThreeToThreeOne(specification: AnyObject) {
+export function upgradeFromThreeToThreeOne(originalSpecification: AnyObject) {
+  let specification = originalSpecification
```
tmastrom, line 10: was the structured clone taking too much memory?

tmastrom (changes requested):
Looks good! I think we need a couple more changes to accommodate the changes to `traverse`

packages/openapi-parser/src/utils/upgradeFromThreeToThreeOne.ts
hanspagel: It was causing errors in the pipeline syntax, looks like it’s relying on a variable reference 🫨, but I don’t want to tackle this in the same PR.

packages/openapi-parser/src/utils/traverse.ts
hanspagel: True!

PR #3412, opened 2024-10-09
style(api-client): enhances responsiveness

antlio:
this pr updates style in api client in order to fixes and enhances responsiveness glitches such as
- view header button width
- alignments
- double borders
- border radiuses

as seen below:

**before**
image

image

image

**after**
image

image

image


amritk (approved):
LGTM, maybe @cameronrohani take a look

antlio:
merging this as it is breaking ui but will fast follow up if need according to feedbacks!

PR #3413, opened 2024-10-09
feat(api-client): handle file download with filename in response

inyourtime:
This update adds the capability to download a file response with the correct filename extracted from the response headers. The feature ensures that when a file is downloaded via the API client, the filename is properly set according to the content disposition, providing a more user-friendly experience.

before

https://github.com/user-attachments/assets/368e48c7-cc04-403e-809b-19e3fa890c23

after

https://github.com/user-attachments/assets/938de78d-e0b8-498b-a94f-a16485653b00

### DIFF
--- a/.changeset/late-chairs-scream.md
+++ b/.changeset/late-chairs-scream.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
+
+feat: download response with filename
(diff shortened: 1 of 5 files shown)


antlio:
tysm, that's such a great addition!

amritk (approved):
Lets gooooooo

marclave (approved):
UNREAL work @inyourtime !!!!!! love this improvement

and tests!? 😄

PR #3414, opened 2024-10-09
feat(object-utils): new debounce function without dependencies

tmastrom:
Currently, the `object-utils` package depends on `@vueuse/core` for its `useDebounceFn` 

This PR removes the dependency, `@vueuse/core`, from `@scalar/object-utils` by creating a new dependency-less debounce function for use with mutations.

### DIFF
--- a/.changeset/weak-terms-reply.md
+++ b/.changeset/weak-terms-reply.md
@@ -1,0 +1,5 @@
+---
+"@scalar/object-utils": patch
+---
+
+feat(object-utils): new debounce function without dependencies
--- a/packages/object-utils/package.json
+++ b/packages/object-utils/package.json
@@ -69,7 +69,6 @@
   ],
   "module": "dist/index.js",
   "dependencies": {
-    "@vueuse/core": "^10.10.0",
     "flatted": "^3.3.1",
     "just-clone": "^6.2.0",
     "ts-deepmerge": "^7.0.1"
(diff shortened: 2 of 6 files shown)


hanspagel (commented):
Is that really a win, though? 

We’re using @vueuse/core in a lot of places, even `useDebounceFn` in another place, it’s well-tested, documented, tree-shakable … And having this one function in the repository doesn’t actually remove the dependency from our popular packages, but adds to the bundle size.

Or in other words: What’s the benefit of removing this dependency in just this package?

geoffgscott:
> Is that really a win, though?
> 
> We’re using @vueuse/core in a lot of places, even `useDebounceFn` in another place, it’s well-tested, documented, tree-shakable … And having this one function in the repository doesn’t actually remove the dependency from our popular packages, but adds to the bundle size.
> 
> Or in other words: What’s the benefit of removing this dependency in just this package?

This is a very low level package in our usage tree. @vueuse has many dependencies and bloats a bunch of non vue packages that consume object-utils. For an incredible small increase in package size to vue packages (where size is less of an issue) we reduce the package size in many other places.

hanspagel:
Yeah, that’s what I’m asking for: In which packages? Because it’s still heavily used in @scalar/api-reference and @scalar/api-client and those seem to be the only packages using @scalar/object-utils.

tmastrom:
> Yeah, that’s what I’m asking: Reduces the package size In which places? Because it’s still heavily used in @scalar/api-reference and @scalar/api-client and those seem to be the only packages using @scalar/object-utils.

We also use @scalar/object-utils in @scalar/oas-utils. Removing vue dependencies here will give us some bigger savings in backend packages

hanspagel (approved):
Ah, that makes sense!

PR #3417, opened 2024-10-09
fix(object-utils): setNested doesn’t work with arrays

geoffgscott:
There was a critical logic error in determining the end state for nested setting. This corrects the logic.


amritk (approved):
Need this!

PR #3423, opened 2024-10-09
feat(api-client): watch mode 👀

amritk:
Alright, this PR introduces a new feature: Watch your OpenAPI URLs for changes. Once you add to your OpenAPI document, it’ll update the API client for you. ⚡ 

Screenshot 2024-10-15 at 14 17 02

**Test**

1. Open two terminals and run `pnpm dev:galaxy` and `pnpm dev:app`.
2. Add `http://localhost:5052/_fresh/openapi.yaml` to import a collection.
3. Open a request in the newly imported collection and you should see some output in the console already.
4. You can then make changes in the `3.1.yaml` and see the API client updated shortly after. 👀 

**Future Tasks**
- example diffing - so we don't replace the whole example
- the content-type header disappearing will be fixed when I do my content-type PR
- variable speed polling


packages/api-client/src/components/CommandPalette/CommandPaletteImport.vue
```diff
@@ -65,12 +67,10 @@ async function importCollection() {
   loader.startLoading()
   try {
     if (isInputUrl.value)
-      await importSpecFromUrl(
-        inputContent.value,
-        undefined,
-        undefined,
-        activeWorkspace.value.uid,
-      )
+      await importSpecFromUrl(inputContent.value, activeWorkspace.value.uid, {
```
amritk, line 70: cleaned up the config

packages/api-client/src/components/CommandPalette/CommandPaletteImport.vue
```diff
@@ -123,16 +123,28 @@ async function importCollection() {
...
-        class="p-2 max-h-8 gap-1.5 text-xs hover:bg-b-2 relative"
-        variant="outlined"
-        @click="openSpecFileDialog">
-        JSON, or YAML File
-        <ScalarIcon
-          class="text-c-3"
-          icon="UploadSimple"
-          size="md" />
-      </ScalarButton>
+      <div class="flex flex-row items-center justify-start gap-3">
```
amritk, line 126: This needs to be removed before merge

packages/api-client/src/store/security-schemes.ts
```diff
@@ -62,47 +62,47 @@ export function extendedSecurityDataFactory({
   }
 
   /** Delete a security scheme and remove the key from its corresponding parent */
-  const deleteSecurityScheme = (scheme: SecurityScheme) => {
+  const deleteSecurityScheme = (schemeUid: string) => {
```
amritk, line 65: we weren't using the whole scheme here so just downgraded to just the uid

packages/api-client/playground/app/main.ts
```diff
@@ -1,5 +1,5 @@
 import { createApiClientApp } from '@/layouts/App'
 
 createApiClientApp(document.getElementById('scalar-client'), {
-  proxyUrl: 'https://proxy.scalar.com',
+  // proxyUrl: 'https://proxy.scalar.com',
```
amritk, line 4: need to remove this before merge

packages/api-client/src/components/CommandPalette/CommandPaletteImport.vue
hanspagel: Why do we remove the UI? I'd just enable it by default and keep it. It communicates what happens behind the scenes (super important) and gives the user a little bit more control.

hanspagel:
THIS IS SO COOL!

I’ve gone through the wording, wdyt?

Something that I experienced: When I added multiple collections with the same URL, it seems it’s watching only one of them.

And the `useOpenApiWatcher` (used to be `useLiveSync`) is deeply nested, can we flatten this with helper methods, early returns or something?

Screenshot 2024-10-10 at 10 56 43

amritk:
Love it! Nice feedback, will implement when I get back to this one!

packages/api-client/src/components/CommandPalette/CommandPaletteImport.vue
amritk: oh so anythony/cam can implement something much nicer ✨

packages/fastify-api-reference/src/fastifyApiReference.ts
```diff
@@ -251,6 +251,8 @@ const fastifyApiReference = fp<
         return reply
           .header('Content-Type', `application/json`)
           .header('Content-Disposition', `filename=${filename}.json`)
+          .header('Access-Control-Allow-Origin', '*')
```
amritk, line 254: do we want to keep this?

marclave (approved):
this is honestly so flipping cool.

@cameronrohani wanna come in and do a LITTLE bit of polish, add an active state to the sidebar or smth somewhere and we get this in :) 

if you end up needing more time @cameronrohani we can get this in and maybe comment it out from import?

packages/fastify-api-reference/src/fastifyApiReference.ts
hanspagel: I’ll make a separate PR! 👍

hanspagel:
@amritk asked for reviews, so let’s set it to Ready for Review

amritk:
Unfortunately typescript wasn't smart enough for what I was planning, maybe in future versions 😞

packages/api-client/src/views/Request/hooks/useOpenApiWatcher.ts
```diff
@@ -0,0 +1,181 @@
...
+    }
+    // Components.securitySchemes
+    else if (d.path[0] === 'components' && d.path[1] === 'securitySchemes') {
+      const payload = diffToSecuritySchemePayload(
+        d,
+        activeCollection.value,
+        securitySchemes,
+      )
+
+      if (payload?.method === 'add') securitySchemeMutators.add(...payload.args)
```
amritk, line 62: typescript did not do what I wanted it to here unfortunately so it gets pretty ugly. We can fix it up later I think

tmastrom (commented):
Very cool new feature!! Working for me locally 😄

hanspagel (approved):
PR of the month already 🏆

hanspagel:
We should test what happens when the URL disappears/throws an error. I’d expect the collection to stay in place, just not update. And ideally we’d visualize the error state in the UI.

antlio:
as discussed here a potential take on the Watch Mode display from the collection context menu:
<div>


</div>

antlio:
added in the previous commits the following ui updates (tooltip wording could be enhanced):

https://github.com/user-attachments/assets/41bf3f3d-65cb-43f8-ab82-c6e5acd58701

antlio:
added in the last commit, the watch for changes state for a collection in the request sidebar:

image

hanspagel:
Love the new toggle, but I think it should be enabled by default (at least when an URL is in the input). Wdyt?

antlio:
> Love the new toggle, but I think it should be enabled by default (at least when an URL is in the input). Wdyt?

It was and  I changed default to false intentionally to keep it disable if its not a URL, but i agree let bring it back for URL content addition!

amritk:
Done!

marclave:
@amritk @hanspagel we should add some information about ensuring they enable cors, and we should handle imports

if i run `pnpm dev` in root, then feeed in the fastify generated openapi spec
it says the import is succesful but it errors in console 
image

cameronrohani:
- reduced size of toggle
- introduced the watch mode icon 
- pushed to right because it makes more sense (to me) to be beside the import button > the file upload button

image
image

packages/api-client/src/components/CommandPalette/CommandPaletteImport.vue
```diff
@@ -123,16 +138,59 @@ async function importCollection() {
...
+            <div
+              class="grid gap-1.5 pointer-events-none max-w-[320px] w-content shadow-lg rounded bg-b-1 z-100 p-2 text-xxs leading-5 z-10 text-c-1">
+              <div class="flex items-center text-c-2">
+                <span v-if="!!inputContent && !isInputUrl"
+                  >Watch Mode is only supported with URL</span
+                >
+                <span
+                  v-else
+                  class="text-pretty"
+                  >Watch your OpenAPI URL for changes. When enable it will
```
hwkr, line 186: Typo I think

```suggestion
                  >Watch your OpenAPI URL for changes. When enabled it will
```

packages/components/src/components/ScalarToggle/ScalarToggle.vue
```diff
@@ -0,0 +1,55 @@
...
+  base: 'relative h-5 w-8 cursor-pointer rounded-full bg-b-3 transition-colors duration-300',
+  variants: {
+    checked: { true: 'bg-c-accent' },
+    disabled: { true: 'cursor-not-allowed opacity-40' },
+  },
+})
+</script>
+<template>
+  <div
+    :aria-checked="modelValue"
```
hwkr, line 31: I think these aria attributes are supposed to be on the input, also it might be good to be able to set `aria-label` or `aria-labbelledBy` on the toggle

packages/api-client/src/views/Request/RequestSidebarItem.vue
```diff
@@ -373,6 +376,34 @@ function openCommandPaletteRequest() {
...
+                  icon="Watch"
+                  size="sm"
+                  thickness="2.5" />
+              </template>
+              <template #content>
+                <div
+                  class="grid gap-1.5 pointer-events-none max-w-[320px] w-content shadow-lg rounded bg-b-1 z-100 p-2 text-xxs leading-5 z-10 text-c-1">
+                  <div class="flex items-center text-c-2">
+                    <span class="text-pretty">
+                      Your Open API document is being monitored for changes to
```
hwkr, line 400: This could maybe be a little shorter? Could also be nice to include the URL here since I don't think we show it anywhere else

Maybe:

> Watching: https://galaxy.scalar.com/openapi.yaml

hwkr (commented):
Looks good! Left a couple of comments mostly wording / typos. Had a look at the live sync code but not super in depth 🙏

packages/api-client/src/components/CommandPalette/CommandPaletteImport.vue
```diff
@@ -123,16 +138,59 @@ async function importCollection() {
...
+                !!inputContent && !isInputUrl
+                  ? 'cursor-default'
+                  : 'cursor-pointer'
+              "
+              for="watch-toggle">
+              <ScalarToggle
+                id="watch-toggle"
+                v-model="watchForChanges"
+                :disabled="!!inputContent && !isInputUrl" />
+              Watch Mode
```
hwkr, line 173: I think this could just say "Watch for changes"

antlio:
> * reduced size of toggle
> * introduced the watch mode icon
> * pushed to right because it makes more sense (to me) to be beside the import button > the file upload button

nice icon addition! i feel like we might want to handle the file upload differently at some point as it might make the visual hierarchy inconsistent, any thougts?

packages/api-client/src/views/Request/RequestSidebarItem.vue
antlio: great idea about displaying the url here 👍

antlio:
according to latest feedbacks, latest commits adds:

style and wording updates in the import modal:
image

tooltip content update in the sidebar:
image

packages/api-client/src/components/CommandPalette/CommandPaletteImport.vue
```diff
@@ -59,18 +63,28 @@ const { open: openSpecFileDialog } = useFileDialog({
...
+  if (!newVal) {
+    watchForChanges.value = false
+  }
+})
+
+watch(inputContent, (newVal) => {
+  if (!isUrl(newVal)) {
+    watchForChanges.value = false
+  }
+})
```
geoffgscott, line 76: Need comments on these. Also strongly prefer one liners for these type of statements:
 ```ts
if (!isUrl(newVal)) watchForChanges.value = false
```

File: packages/api-client/src/store/security-schemes.ts [hunk 12]
```diff
@@ -62,47 +62,47 @@ export function extendedSecurityDataFactory({
...
 
     // Remove the scheme from any collections that use it
     Object.values(collections).forEach((c) => {
-      if (scheme.uid in c.auth) {
-        const { [scheme.uid]: toDelete, ...rest } = c.auth
+      if (schemeUid in c.auth) {
+        const { [schemeUid]: toDelete, ...rest } = c.auth
         collectionMutators.edit(c.uid, 'auth', rest)
       }
     })
```
geoffgscott, line 83: Can this not be done in 1 loop?

packages/api-client/src/views/Request/Request.vue
```diff
@@ -106,6 +107,8 @@ onMounted(() => {
   events.cancelRequest.on(cancelRequest)
 })
 
+useOpenApiWatcher()
```
geoffgscott, line 110: This feels like a strange place to enable the watcher. Should be done at the App level as its a background state watcher and not specific to a single page.

packages/api-client/src/views/Request/hooks/useOpenApiWatcher.ts
```diff
@@ -0,0 +1,208 @@
...
+
+  /** Little toast helper */
+  const toastError = (type: string) =>
+    toast(
+      `[useOpenApiWatcher] changes to the ${type} were not applied`,
+      'error',
+    )
+
+  // Transforms and applies the diff to our mutators
+  const applyDiff = (d: Difference) => {
```
geoffgscott, line 55: This function is very large and very heavy. It needs to have at least some level of test coverage. 

I would suggest we pull the apply diff function out and provide the `StoreContext` as second parameter to the function. We use this pattern elsewhere in the store. This would allow separate unit testing of the apply diff function. 

The function can return a success boolean and error messages that can be toasted. 

I would also prefer an early return pattern over the massive list of else-ifs but thats more of a nit.

packages/api-client/src/views/Request/types/sidebar-item.ts
```diff
@@ -17,6 +17,8 @@ export type SidebarItem = {
   icon?: string
   edit: (name: string, icon?: string) => void
   delete: () => void
+  documentUrl?: string
+  watchForChanges?: boolean
```
geoffgscott, line 21: Out of scope but the sidebar being nested deep inside  the request view looks wrong. The same sidebar would be shared by tags, collections, and model pages.

packages/oas-utils/src/entities/spec/server.ts
```diff
@@ -37,7 +37,7 @@ export const oasServerSchema = z.object({
    * the host location is relative to the location where the OpenAPI document is being served. Variable substitutions
    * will be made when a variable is named in {brackets}.
    */
-  url: z.string().optional(),
+  url: z.string().optional().default(''),
```
geoffgscott, line 40: These types should be purely spec compliant. They can be defaulted when created.

geoffgscott (commented):
Overall looks good. A few comments.

packages/api-client/src/views/Request/Request.vue
amritk: Its only used on the request page though

packages/oas-utils/src/entities/spec/server.ts
amritk: this fixes the type thats generated

packages/api-client/src/components/CommandPalette/CommandPaletteImport.vue
amritk: agreed

amritk: but also that should be a lint rule

packages/api-client/src/store/security-schemes.ts
amritk: yup, think it was written at two different times

packages/api-client/src/views/Request/hooks/useOpenApiWatcher.ts
amritk: yea so it started as one massive function, this is actually the third time it was broken out, was just avoiding having to mock mutators but I can do that

packages/api-client/src/components/CommandPalette/CommandPaletteImport.vue
hwkr: +1 for one liner lint rule

PR #3424, opened 2024-10-10
fix(api-reference): lazy loading content jump

amritk:
This one has been a little annoying and I could have sworn I fixed it before. Basically on initial load lazy loading works fine, BUT on subsequent loads there was a strange content jump happening. Turns out this was the browser remembering the scrollPosition, which we don't want as we manually handle it with lazy loading.

enter `history.scrollRestoration = 'manual'`

Before:

https://github.com/user-attachments/assets/b9d9c8d0-4d21-4f0d-b5f2-2d5edc9c7548

After:

https://github.com/user-attachments/assets/cacad2cd-801c-4a52-8888-0d3c11d12e30

### DIFF
--- a/.changeset/hip-avocados-count.md
+++ b/.changeset/hip-avocados-count.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
+fix: lazyLoading content jump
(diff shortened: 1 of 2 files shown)


hwkr (approved):
Code looks good! :)

marclave (approved):

PR #3425, opened 2024-10-10
refactor(api-client): localize event bus in workspace store

hwkr:
Refactors the event bus to be localized to the app and available via the workspace store. I didn't want to change the event bus api too much so I pulled from [`@vueuse/core/useEventBus`]( https://github.com/vueuse/vueuse/tree/v10.10.0/packages/core/useEventBus) and modified it to create local event bus on an object rather than registering them globally via a `Symbol`. 

This fixes an issue with having two client instances on the same page where triggering an event in one client triggers it in all the clients. There's still some issues with the references using `useEventBus` but you know, one step at a time.

The most interesting / important files are:

* [`@libs/event-bus.ts`](https://github.com/scalar/scalar/pull/3425/files#diff-4616fc6972e1e5c291dfa6e88594c4951888a837983e7a8fe030c3eda58d5ba3) and [tests](https://github.com/scalar/scalar/pull/3425/files#diff-8bc485d87ebe69997aa500cf4be1c4efb07e65f7ba872bcd6784a738dc8aff67) which is a modified version of the vue use event bus
* [`@store/events`](https://github.com/scalar/scalar/pull/3425/files#diff-e01de7202ae80f42f702c013e846a9ead75f933d69611a70d3398d12c2532bb7) which is all the event busses registered to the workspace that used to live in `@libs/event-busses`

## Before

https://github.com/user-attachments/assets/96222cac-0dc3-4c5c-8961-b5391f8e1b40

## After

https://github.com/user-attachments/assets/e7c8df69-7e14-41ae-9502-163e31feff59


packages/api-client/src/store/events.ts
hwkr, line 1: This file could maybe also be called `event-busses.ts` if we want 🤔

tmastrom (commented):
This is really cool 👏

hanspagel:
Supercool! I don’t think I’m competent enough to review the code changes, though. 😁

amritk (commented):
Alright this is awesome!!! Just one question, should we have different even busses? Or just one bus now with fully typed different events?

hwkr:
> Alright this is awesome!!! Just one question, should we have different even busses? Or just one bus now with fully typed different events?

It kind of doesn't matter, we could consider the event bus to be one bus with different event types but that would involve a bit more refactoring and this most closely matches the language of the vue use package.

amritk (approved):
Tested it and works great! nice work, send it!

antlio (approved):
great refactor, seems to work nicely!

PR #3426, opened 2024-10-10
chore: remove changeset codeowners

hwkr:
I think right now we're always adding all the default codeowners on every PR that includes a changeset (which most of them do) which I don't think is what we want?


amritk (approved):
Smart haha

PR #3428, opened 2024-10-10
chore: move `packages/api-client-app` to `projects/scalar-app`

hanspagel:
This moves the Electron app to the `projects/` and renames it from `api-client-app` (folder)/`scalar-api-client` (package) to `scalar-app`.

```bash
projects
├── client-scalar-com
└── scalar-app
```

The build failed for some reason. Locally, it went away after I deleted the `node_modules` folder, so maybe we’re hitting some kind of cache here? 👀 I’ve spotted a warning about some packages requiring Node 20, so I’ve add Node 18 for the few packages that didn’t already.


packages/themes/src/presets/alternate.css
```diff
@@ -83,7 +83,7 @@
   --scalar-button-1-color: black;
 }
 
-.scalar-api-client__item,
+.scalar-app__item,
```
hwkr, line 86: Did this class actually get changed in the source?

hanspagel: GREAT CAATCH, yes!

geoffgscott:
cc @amritk @marclave @hanspagel 
I am not sure when the projects folder was instantiated but I would like to see us move to a different organization pattern in the OSS repo. 

Ideally all of the related packages of api-reference and api-client can be grouped into respective folders. I think its reasonable to just have sub-folders within the packages folder but we could also fork out the end use packages into a projected folder. 

Needs more discussion before we move things.

hanspagel:
Closing this until we agree on a strategy then.

PR #3429, opened 2024-10-10
chore: remove user without write permission from CODEOWNERS

hanspagel:
Looks like you can’t have users without full write permission in the `CODEOWNERS` file, sorry @xC0dex. :(

I need to find another way to automatically assign you as a reviewer.


xC0dex:
😢 For me, it's also ok if you manually request my review once there is a PR.

PR #3432, opened 2024-10-10
chore(Scalar.AspNetCore): remove example folder and update image path

xC0dex:
### Description

This PR removes the `examples/DotNet` folder from the repo. It's not maintained and I think it's not referenced somewhere. The `packages/scalar.aspnetcore` folder contains a `Readme.md` with a guide on how to run the playground project. This is a good example and enough IMO.
I also updated the path of the `dotnet.jpg` in the README. This README is also used for the package, and the image is not displayed correctly on in the [package description](https://www.nuget.org/packages/Scalar.AspNetCore) on nuget.


hanspagel (approved):
OMG great PR

PR #3435, opened 2024-10-10
fix(api-reference): accessibility issues

### LINKED ISSUE #3422: Accessibility in Introduction section of API Reference
### What happens?

Screen Shot 2024-10-09 at 4 19 55 PM

1. There is no visual indication of focus when the keyboard navigates over Base Url and authentication fields in the introduction card of an API Reference. The typing cursor appears in Authentication fields, but there should be a high contrast focus box that shows when we navigate to all of these active elements.
2. In the Introduction card, the labels of the Base Url, authentication type, and Client Libraries sections should be announced by a screen reader when a user interacts with the fields.
3. In authentication section in the introduction card, screen readers do not announce the current state of the ‘show password’ button (show or hide).
4. In the Client Libraries selection tab menu, the focus goes only to the More combobox, but the focus is not visually indicated in a high-contrast way. I can navigate the library drop down combobox with the keyboard, but focus does not go through the other tabs displayed horizontally. 
    1. When I move the keyboard focus to the More button in Client Libraries, the screen reader announces the name of the library that is currently selected. This is not a meaningful announcement. The screen reader needs to give more context about the purpose of this menu.
    2. The icons besides languages in the horizontal tabs of the Client Libraries menu are missing the following attributes: focusable="false”, role="presentation”, aria-hidden="true"
    3. The tab elements are missing the following attributes: role="tablist" for tab container, role="tab" for tab, role="tabpanel" for page under the tab, aria-control - for each tab referring to associated tabpanel, aria-select - true for the current page, and false for others
5. 'Download OpenAPI Document' button should have role="link" as the main function of it is to download a document. Should have attribute role="link", and screenreader should announce the 'Download OpenAPI Document' as a link.
6. There is no ‘main’ landmark on the page. 

All active elements in the introduction card should be keyboard accessible.
When the keyboard focus is on an element, there should be a visual indicator of the focus (usually a box highlighting the element). 
The screen reader should be able to make an announcement of each element that includes context for the visual label.  For example,  the Client Libraries menu options should be able to say 'Client Libraries Curl'. 
Buttons for showing or hiding elements should have screen reader announcements for the status. 
The Download OpenAPI Document button should have role=link. 
I think the 'main' landmark should wrap the introduction section and all the sections for tags/endpoints, but you may decide to have it only around the tag sections. There needs to be one somewhere on the page.

### OpenAPI Document

_No response_

antlio:
this pr fixes #3422 in order to enhance accessibility support and fix lack of as described in the issue.

### DIFF
--- a/packages/api-reference/src/components/ApiReferenceLayout.vue
+++ b/packages/api-reference/src/components/ApiReferenceLayout.vue
@@ -261,6 +261,7 @@ const themeStyleTag = computed(
   <div v-html="themeStyleTag"></div>
   <main
     ref="documentEl"
+    :aria-label="`Open API Documentation for ${parsedSpec.info?.title}`"
     class="scalar-app scalar-api-reference references-layout"
     :class="[
       {
--- a/packages/api-reference/src/components/ApiReferenceLayout.vue
+++ b/packages/api-reference/src/components/ApiReferenceLayout.vue
@@ -259,9 +259,8 @@ const themeStyleTag = computed(
 </script>
 <template>
   <div v-html="themeStyleTag"></div>
-  <main
+  <div
     ref="documentEl"
-    :aria-label="`Open API Documentation for ${parsedSpec.info?.title}`"
     class="scalar-app scalar-api-reference references-layout"
     :class="[
       {
@@ -284,8 +283,9 @@ const themeStyleTag = computed(
         name="header" />
     </div>
     <!-- Navigation (sidebar) wrapper -->
-    <aside
+    <nav
       v-if="configuration.showSidebar"
+      :aria-label="`Contents of ${parsedSpec.info?.title}`"
       class="references-navigation t-doc__sidebar">
       <!-- Navigation tree / Table of Contents -->
       <div class="references-navigation-list">
@@ -305,7 +305,7 @@ const themeStyleTag = computed(
           </template>
         </Sidebar>
       </div>
-    </aside>
+    </nav>
     <!-- Swagger file editing -->
     <div
       v-show="configuration.isEditable"
@@ -318,7 +318,9 @@ const themeStyleTag = computed(
     </div>
     <!-- Rendered reference -->
     <template v-if="showRenderedContent">
-      <div class="references-rendered">
+      <section
+        :aria-label="`Open API Documentation for ${parsedSpec.info?.title}`"
+        class="references-rendered">
         <Content
           :baseServerURL="configuration.baseServerURL"
           :layout="configuration.layout === 'classic' ? 'accordion' : 'default'"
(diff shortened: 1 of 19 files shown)


packages/api-reference/src/features/BaseUrl/ServerForm.vue
```diff
@@ -39,11 +39,15 @@ function updateVariable(name: string, value: string) {
...
+    <span
+      aria-label="Base URL"
+      class="server-form-title"
+      >Base URL</span
+    >
     <div class="server-form">
       <div class="server-form-container">
         <!-- Dropdown -->
-        <div class="server-item">
+        <div>
```
kayla-beckham, line 50: I think the selector for base url needs the aria label, or maybe aria-labelled-by. I was trying my screen reader on one of the netlify previews for this PR and I did not hear the label announced when I navigated over BaseURL and the Client Libraries menus.

antlio: hey @kayla-beckham, great feedbacks you are right! i kept it as a draft as I haven't got the time to go through all the screen reader support yet, but it will be def done in that pr :)

hwkr: Updated, I'll take a quick swing at the other blocks too

packages/api-reference/src/components/ApiReferenceLayout.vue
```diff
@@ -317,7 +318,9 @@ const themeStyleTag = computed(
     </div>
     <!-- Rendered reference -->
     <template v-if="showRenderedContent">
-      <div class="references-rendered">
+      <section
```
hwkr, line 321: @kayla-beckham since the references are often embedded in a page with other content this has to be an aria `region` landmark rather than a `main` landmark otherwise we could end up with nested `main` landmarks which is a WCAG 🙅‍♀️ 

I added a `<main>` tag to the standalone CDN version of the references which is meant to be alone on the page but if have the references embedded you'll need to add your own `main` landmark somewhere.

packages/api-reference/src/components/Content/ClientLibraries/ClientSelector.vue
```diff
@@ -78,29 +78,29 @@ const checkIfClientIsFeatured = (client: HttpClientState) =>
     <button
       v-for="client in featuredClients"
       :key="client.clientKey"
+      aria-hidden="true"
       class="client-libraries rendered-code-sdks"
       :class="{
         'client-libraries__active': isSelectedClient(client),
       }"
+      tabindex="-1"
```
antlio, line 86: this prevent the button from being outlined on tab and selected, is that what we want?

hwkr (approved):

hanspagel (approved):
What an awesome PR! Feels like a massive a11y improvement

PR #3436, opened 2024-10-10
fix(api-reference): duplicate download event listener

hanspagel:
In some cases, the download was triggered twice. We register an event listener for the download, but we don’t unregister ist when the component is unmounted. This PR fixes it.


marclave (approved):
😏 nice

PR #3437, opened 2024-10-10
chore(oas-utils): cache `getExampleFromSchema` results

hanspagel:
This PR caches the results of `getExampleFromSchema` in a map, which leads to huge performance gains:

* It’s ~10% slower for simple schemas (which are processed fast anyway).
* It’s 10× faster for simple schemas with circular references.
* It’s 100× faster for more complex schemas with circular references.


MrSharpp:
thanks!

marclave (approved):
🤯 love this

PR #3448, opened 2024-10-11
fix(api-client): command palette server collection

antlio:
this pr updates the command palette server collection dropdown displays and logic, in order to display a create collection button if no collection exists + updates the submit button state:

**before**
image

**after**
image

image


hanspagel (approved):
Great UX improvement

amritk (approved):

PR #3452, opened 2024-10-11
chore(api-client-app): delete open page

hanspagel:
This PR deletes the “open” page that I’ve built as an MVP for the import from references, we don’t need this anymore. :)


amritk (approved):
Not even sure what this is, but hans knows what hes doing

PR #3453, opened 2024-10-11
fix(fastify-api-reference): does not work with `ignoreTrailingSlash: true`

### LINKED ISSUE #3408: Method 'GET' already declared for route '/reference/'
### What happens?

I'm getting the following Fastify error with `@scalar/fastify-api-reference` versions `1.25.20` and above, if I downgrade to `1.25.19` the error goes away.

```
{
  code: "FST_ERR_DUPLICATED_ROUTE",
  name: "FastifyError",
  statusCode: 500,
  message: "Method 'GET' already declared for route '/reference/'"
}
```

I'm not doing anything particularly custom in my setup, I basically just followed the example here https://github.com/scalar/scalar/blob/main/examples/fastify-api-reference/src/index.ts

```ts
import swagger from '@fastify/swagger';
import apiReference from '@scalar/fastify-api-reference';

...

fastify.register(swagger, {
  hideUntagged: true,
  openapi: {
    openapi: '3.0.0',
    components: {
      securitySchemes: {
        bearerToken: {
          type: 'http',
          scheme: 'bearer',
          bearerFormat: 'JWT'
        }
      }
    },
    security: [
      {
        bearerToken: []
      }
    ]
  }
});
fastify.register(apiReference, {
  routePrefix: '/reference'
});
```

Expected to not get an error

### OpenAPI Document

_No response_

hanspagel:
Currently, our Fastify integration throws an error if `ignoreTrailingSlash` is set to `true`.

If this is set, Fastify registers both versions of a route (with and without a trailing slash), which we do too. So it correctly throws an error, because the route is already registered.

This PR should fix it. :)

Fixes #3408

### DIFF
--- a/.changeset/wise-doors-hide.md
+++ b/.changeset/wise-doors-hide.md
@@ -1,0 +1,5 @@
+---
+'@scalar/fastify-api-reference': patch
+---
+
+fix: doesn’t work with ignoreTrailingSlash: true
--- a/examples/fastify-api-reference/src/index.ts
+++ b/examples/fastify-api-reference/src/index.ts
@@ -5,6 +5,7 @@ import Fastify from 'fastify'
 // Init Fastify
 const fastify = Fastify({
   logger: false,
+  // ignoreTrailingSlash: true,
 })
 
 // Register Swagger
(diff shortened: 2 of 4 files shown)


amritk (approved):
LGTM! nice catch

PR #3455, opened 2024-10-11
feat(fetch): `@scalar/fetch` package

tmastrom:
Lightweight, fetch-based wrapper over an HTTP client that returns standardized and typesafe response formats.


packages/fetch/README.md
```diff
@@ -0,0 +1,118 @@
...
+
+And errors will look like
+
+```ts
+export type APIError = {
+  status: number
+  message: string
+  error: true
+  originalError: any
+}
```
hanspagel, line 109: We’d use `ApiResponse` and `ApiError` (instead of API…) in most places. I’m not super happy with the/my decision, but it’s what we use basically everywhere.

hanspagel (commented):
Great work! Are we planning to use this in other packages? If so, can you share a little bit about the plans?

hanspagel:
I think we’ll need @geoffgscott’s review here. :v:

packages/fetch/src/index.ts
```diff
@@ -0,0 +1,8 @@
+export { formatApiError, formatApiResponse } from './format'
+export { schemaModel } from './schemaModel'
+export { request } from './request'
+export type {
+  ApiError as APIError,
+  ApiResponse as APIResponse,
```
geoffgscott, line 6: I don't think we need to alias here. We can do a find/replace in scalar-org

tmastrom: I agree. This is an interesting functionality of symbol renaming. It seems to automatically alias exports if the name changes. I removed the alias'

hanspagel:
`msw` is causing the type errors 🤔

tmastrom:
> `msw` seems to force pnpm to update rollup or something and it becomes incompatible to the outdated `vite-svg-loader`
> 
> couldn’t find an update or a fork of `vite-svg-loader`, added a `@ts-expect-error` for now, let’s see if we can get back to green. :)

IT'S GREEEEN

PR #3457, opened 2024-10-11
feat(api-client): add links to HTTP header docs

hanspagel:
This PR adds links to docs for the various HTTP headers.

We’ve had this a while ago, I don’t know if it was removed intentionally, but I personally miss it. I wouldn’t mind if we’d remove the underline to make it look slicker. Wdyt @cameronrohani?

Screenshot 2024-10-11 at 15 46 21


amritk:
Strange how this is such a big bundle size increase

hanspagel:
It can be gzipped pretty well, though (repeating content):

@scalar/api-reference:build: dist/browser/standalone.js       2,517.83 kB │ gzip: 728.84 kB
@scalar/api-reference:build: dist/browser/standalone.js       2,532.16 kB │ gzip: 731.09 kB

So it’s +3kb gzipped, which I’d say is not great, but okay. I don’t know, I leave it up to our design head @cameronrohani.

cameronrohani (approved):
awesome, def need this!

PR #3458, opened 2024-10-11
chore(openapi-parser): switch to `@scalar/build-tooling`

hanspagel:
With that PR we’re using the `@scalar/build-tooling` and the default tsconfig (with a few additions) for the parser now.

I’ll tackle the few additions to the tsconfig in a separate PR, but this is already way closer to the standard than what we had before.


amritk (approved):
Looks correct

hanspagel (commented):
Ah, I can remove some of the rollup dependencies.

packages/openapi-parser/tsconfig.build.json
```diff
@@ -1,13 +1,13 @@
...
-  "exclude": ["node_modules", "dist", "test", "**/*.test.ts"],
+  "exclude": ["**/*.test.ts"],
   "compilerOptions": {
     "skipLibCheck": true,
     "declaration": true,
     "declarationMap": true,
     "emitDeclarationOnly": true,
     "outDir": "dist/",
-    "rootDir": "src/"
+    "allowImportingTsExtensions": true
```
hanspagel, line 11: Can remove this line again

PR #3460, opened 2024-10-11
fix(cli): doesn’t validate files with external references

hanspagel:
This PR tries to make the CLI work with external references (again).

**Issues**
- [ ] files prefixed with `./` can’t be found
- [ ] probably applies to `../` then, too
- [ ] the `openapi()` pipeline syntax doesn’t seem to work well with filesystems (multiple files) anymore

We need a few more tests around those issues before we fix the actual bugs.


hanspagel:
Nice, those failing tests were fixed by the recent changes to the parser it seems.

PR #3461, opened 2024-10-11
chore: upgrade Vue

amritk:
Second attempt! So far everything working that I can see except nuxt, which has an unrelated issue (ajv related)


hanspagel (approved):
Feels great! I couldn’t find any issues.

PR #3462, opened 2024-10-11
chore(openapi-types): use build tooling scripts for types

hanspagel:
I’ve had problems to build the `@scalar/openapi-types` locally, switching to the build tooling scripts for types:build and types:check helped.

Ah, and I’ve had to fix the pnpm lock file for whatever reason.


tmastrom (approved):
Nice. Makes sense to standardize

PR #3464, opened 2024-10-12
fix(api-reference): search modal a11y improvements

hwkr:
Adds screen reader support for the search modal including accessible names and aria descriptions.

### DIFF
--- a/.changeset/nasty-elephants-fry.md
+++ b/.changeset/nasty-elephants-fry.md
@@ -1,0 +1,7 @@
+---
+'@scalar/api-reference': patch
+'@scalar/components': patch
+'@scalar/themes': patch
+---
+
+fix(api-reference): search modal a11y improvements
--- a/packages/api-reference/src/features/Search/SearchButton.vue
+++ b/packages/api-reference/src/features/Search/SearchButton.vue
@@ -59,11 +59,13 @@ whenever(
       size="sm"
       thickness="2.5" />
     <div class="sidebar-search-input">
+      <span class="sr-only">Open </span>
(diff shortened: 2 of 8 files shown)


packages/api-client/package.json
```diff
@@ -77,11 +77,6 @@
       "types": "./dist/libs/importers/index.d.ts",
       "default": "./dist/libs/importers/index.js"
     },
-    "./libs/event-busses": {
-      "import": "./dist/libs/event-busses/index.js",
-      "types": "./dist/libs/event-busses/index.d.ts",
-      "default": "./dist/libs/event-busses/index.js"
-    },
```
antlio, line 84: should we keep this in another commit/pr?

antlio (approved):
overall it looks great! 

could be done in another PR but we could also keep the accent color while tabbing, but maintain a darker shade of grey for focus. what do you think? cc @cameronrohani

hwkr:
> overall it looks great! 
> 
> 
> 
> could be done in another PR but we could also keep the accent color while tabbing, but maintain a darker shade of grey for focus. what do you think? cc @cameronrohani 

That should be how it works! The search results are a little weird though because the don't actually gain focus via tab the way a normal element would, they us an aria active descendant pattern

marclave (approved):
awesome work @hwkr ! waiting on @cameronrohani to review

cameronrohani (approved):

marclave:
@hwkr feel free to merge once conflicts are fixed : )

hanspagel (approved):

PR #3471, opened 2024-10-14
fix(api-client): lower syntax highlighting threshold on responseBody

### LINKED ISSUE #3467: Random/Intermitment UI freezing happens when displaying large JSON responses
### What happens?

**Version:** 1.25.34

Noticable random and long freezing When testing any API endpoint that returns large JSON responses (~ 24 KiB download size).
It costs at least 3 seconds to display a response with 46 ms (round trip time).
Then, every UI components seems freezing, even hovering states are not displaying correctly.
After a few extra seconds, I can start scrolling the response view, but it may or may not scroll properly.
Using the small scrollbar for quickly navigating is completely broken, it has been frozen.
The UI may be unfrozen after a while or not, even on a very beefy workstation with 64 GiB RAM and a lot of processing cores.

I think we should disable syntax highlighting for large JSON responses (similar to VS Code behaviors when opening very big files).
I see this kind of issues had been addressed, but not sure why Scalar UI cannot handle ~ 24 KiB of JSON responses without freezing.

https://github.com/scalar/scalar/issues/2503
https://github.com/scalar/scalar/issues/2333

### OpenAPI Document

_No response_

amritk:
Fixes #3467

Switched the unit of measurement to bytes and dropped the threshold to 20k. Lets see how this does.

Also removed some Json parsing since it was redundant. Will fix up the tests

### DIFF
--- a/packages/api-client/src/libs/send-request.ts
+++ b/packages/api-client/src/libs/send-request.ts
@@ -32,11 +32,6 @@ export function decodeBuffer(buffer: ArrayBuffer, contentType: string) {
     const decoder = new TextDecoder(mimeType.parameters.get('charset'))
     const string = decoder.decode(buffer)
 
-    // JSON
-    if (mimeType.subtype === 'json') {
-      return JSON.parse(string)
-    }
-
     // Text
     return string
   }
@@ -218,9 +213,9 @@ export function createFetchBody(
 }
 
 /** Response from sendRequest hoisted so we can use it as the return type for createRequestOperation */
-type SendRequestResponse<ResponseDataType = unknown> = Promise<
+type SendRequestResponse = Promise<
   ErrorResponse<{
-    response: ResponseInstance<ResponseDataType>
+    response: ResponseInstance
     request: RequestExample
     timestamp: number
   }>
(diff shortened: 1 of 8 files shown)


packages/api-client/src/libs/send-request.ts
```diff
@@ -32,11 +32,6 @@ export function decodeBuffer(buffer: ArrayBuffer, contentType: string) {
     const decoder = new TextDecoder(mimeType.parameters.get('charset'))
     const string = decoder.decode(buffer)
 
-    // JSON
```
amritk, line 35: We convert it later anyway, no need here

packages/api-client/src/libs/send-request.ts
```diff
@@ -218,9 +213,9 @@ export function createFetchBody(
 }
 
 /** Response from sendRequest hoisted so we can use it as the return type for createRequestOperation */
-type SendRequestResponse<ResponseDataType = unknown> = Promise<
```
amritk, line 221: removed this extra typing as its not needed

hanspagel:
Awesome! Tried to fix the tests, let’s see. Unfortunately, the tests became uglier, because we need to JSON.parse everywhere now, but it might still be better to keep the ugly tests and remove it from the actual function.

hanspagel (approved):
Love everything that improves the performance

PR #3472, opened 2024-10-14
fix(api-client): add collection.info.version optional

### LINKED ISSUE #3463: allow info.version: null
the api client breaks if the document has `info.version: null`, let's be easy with that, it’s totally okay and actually AFAIK completely irrelevant for the client if this value is `null`. :)

image

```json
{
"openapi": "3.1.0",
"info": {
"title": "Nitro Server Routes",
"version": null
}
}
```

amritk:
fix #3463 
Turns out we already had a default, was just missing optional()

### DIFF
--- a/packages/oas-utils/src/entities/spec/spec-objects.ts
+++ b/packages/oas-utils/src/entities/spec/spec-objects.ts
@@ -57,7 +57,7 @@ export const oasInfoSchema = z.object({
    * REQUIRED. The version of the OpenAPI document (which is distinct from the OpenAPI
    * Specification version or the API implementation version).
    */
-  version: z.string().default('0.0.1'),
+  version: z.string().optional().default('0.0.1'),
 })
 
 /**
--- a/packages/oas-utils/src/entities/spec/spec-objects.ts
+++ b/packages/oas-utils/src/entities/spec/spec-objects.ts
@@ -57,7 +57,7 @@ export const oasInfoSchema = z.object({
    * REQUIRED. The version of the OpenAPI document (which is distinct from the OpenAPI
    * Specification version or the API implementation version).
    */
-  version: z.string().optional().default('0.0.1'),
+  version: z.string().optional().default('1.0'),
 })
 
 /**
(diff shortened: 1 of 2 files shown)


packages/oas-utils/src/entities/spec/spec-objects.ts
```diff
@@ -57,7 +57,7 @@ export const oasInfoSchema = z.object({
    * REQUIRED. The version of the OpenAPI document (which is distinct from the OpenAPI
    * Specification version or the API implementation version).
    */
-  version: z.string().default('0.0.1'),
+  version: z.string().optional().default('0.0.1'),
```
hanspagel, line 60: Not related to this PR, but maybe a great opportunity to clean this up:

Semantic versions start at 0.1.0, not 0.0.1. But I think the default should just be `null`. I don’t see a reason to make up a version number if the users doesn’t add one. Wdyt?

hanspagel (changes requested):

packages/oas-utils/src/entities/spec/spec-objects.ts
amritk: I disagree just because the spec says it is required, so one should always exist. The integrations should just pull it from the package.json

packages/api-client/package.json
```diff
@@ -77,11 +77,6 @@
       "types": "./dist/libs/importers/index.d.ts",
       "default": "./dist/libs/importers/index.js"
     },
-    "./libs/event-busses": {
-      "import": "./dist/libs/event-busses/index.js",
-      "types": "./dist/libs/event-busses/index.d.ts",
-      "default": "./dist/libs/event-busses/index.js"
-    },
```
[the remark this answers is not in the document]
amritk: Yea this one is from some other changes, popped up in a few of my prs

packages/oas-utils/src/entities/spec/spec-objects.ts
hanspagel: What package.json? 🤔 This is for whatever OpenAPI document the user gives us, right? There’s no package.json involved typically.

Let’s use 1.0 as the default then, though. Feels like a better fallback.

```suggestion
  version: z.string().optional().default('1.0'),
```

hanspagel (approved):
LET’S GE T   T HIS   INN

PR #3478, opened 2024-10-15
fix: style api client import modal

cameronrohani:


marclave (approved):

antlio:
Is the display of the `Import to` there at all time is on purpose? :

image

PR #3479, opened 2024-10-15
feat(api-client): add web app playground

antlio:
this pr adds and updates:
- `pnpm dev:web-app` that open the api client web version
- `pnpm dev:app` that open the api client app version

to do so a web app playground has been added in order to use the api client web component that has been introduced in #3351 to reflect the web usage locally.

**why**
currently when using `pnpm dev:app`, the api client launch the app version that does not reflect the web usage.

### DIFF
--- a/.changeset/rude-berries-hunt.md
+++ b/.changeset/rude-berries-hunt.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
+
+feat: adds api client web playground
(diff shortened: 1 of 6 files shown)


hanspagel (approved):
I think that’s great!

tmastrom (approved):
Nice!

marclave (approved):
@antlio im in! lets do it

PR #3480, opened 2024-10-15
chore: update to Electron 33

hanspagel:
This PR updates Electron from 31 to 33.

Makes sense to keep Electron (and the bundled Chromium browser) up to date, I’d say.


marclave (approved):
im in!

PR #3481, opened 2024-10-15
chore(api-client-app): enable Dependabot updates for `@todesktop/*` and `electron`

hanspagel:
This PR enables Dependabot PRs for the `@todesktop/*` dependencies (which we need to keep up to date to stay compatible with the ToDesktop Cloud) and `electron`.


marclave (approved):
;)

PR #3482, opened 2024-10-15
chore: update `@todesktop/cli`

hanspagel:
This needs to be up to date to stay compatible with the ToDesktop Cloud.

> Run pnpm --filter scalar-api-client todesktop:release:ci
>
> scalar-api-client@0.1.61 todesktop:release:ci /home/runner/work/scalar/scalar/packages/api-client-app
> todesktop release --latest --force
> 
> Your version of @todesktop/cli is out of date.
> Run npm install --save-dev @todesktop/cli to update to v1.9.6.

### DIFF
--- a/packages/api-client-app/package.json
+++ b/packages/api-client-app/package.json
@@ -45,7 +45,7 @@
   "devDependencies": {
     "@electron-toolkit/tsconfig": "^1.0.1",
     "@rushstack/eslint-patch": "^1.10.3",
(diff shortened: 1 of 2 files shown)


amritk (approved):
If we need it always up to date, should we change it to latest?

hanspagel:
It’s still pinned in the lock file then, isn’t it? So I’d go with the dependabot approach to keep it up to date.

PR #3483, opened 2024-10-15
fix(api-client): hotkey component

antlio:
this pr does the following:
- sets back the hotkey component to a ui component only
- fixes style
- moves the `meta+k` to the layout level so it can be called from any views (currently not working if we load from another view than request)
- displays hotkey element on sidebar button only on app vs web
- favors `meta+n` for handler (create request, cookie, env) while `meta+k` remains for the global command palette
- constrains event action fired to route (first naive approach) which fix the `meta+n` displaying the request command palette in every view

**before**

on cookie page (both app + web)
image
image

**after**

on cookie page app
image
image

on cookie page web
image

### DIFF
--- a/packages/api-client/src/components/ScalarHotkey.vue
+++ b/packages/api-client/src/components/ScalarHotkey.vue
@@ -1,42 +1,27 @@
 <script setup lang="ts">
-import { keyMap } from '@/hooks'
+import type { HotKeyModifiers } from '@scalar/oas-utils/entities'
 import { isMacOS } from '@scalar/use-tooltip'
-import { useMagicKeys, whenever } from '@vueuse/core'
 import { computed } from 'vue'
 
 const props = defineProps<{
-  hotkey: string
+  hotkey: {
+    modifier?: HotKeyModifiers
+    key: string
+  }
 }>()
 
-const emit = defineEmits<{ (event: 'hotkeyPressed', key: string): void }>()
+const modifier = computed(() => props.hotkey.modifier || 'meta')
 
-// Prepend hotkey based on the OS
-const modifierKey = computed(() => (isMacOS() ? '⌘' : '⌃'))
-
-// Map the key icon displayed in UI to actual keyboard key
-const resolvedHotkey = computed(() => keyMap().get(props.hotkey))
-
-const displayHotkey = computed(() => `${modifierKey.value} ${props.hotkey}`)
-
-const keys = useMagicKeys({
-  passive: false,
-  onEventFired(e) {
-    // Remove default behavior for keypress
-    if (!isMacOS() && e.ctrlKey && e.key === resolvedHotkey.value) {
-      e.preventDefault()
-      e.stopPropagation()
-    }
-  },
-})
-
-whenever(keys[`${isMacOS() ? 'meta' : 'control'}_${props.hotkey}`], () => {
-  emit('hotkeyPressed', resolvedHotkey.value || '')
+const displayHotkey = computed(() => {
+  const modifierKey =
+    modifier.value === 'meta' ? (isMacOS() ? '⌘' : '^') : modifier.value
+  return `${modifierKey} ${props.hotkey.key}`
 })
 </script>
 <template>
   <div
     v-bind="$attrs"
-    class="bg-b-1 border-b-3 inline-block overflow-hidden rounded border-1/2 text-xxs rounded-b px-1 font-medium uppercase">
+    class="border-b-3 inline-block overflow-hidden rounded border-1/2 text-xxs rounded-b px-1 font-medium uppercase">
     {{ displayHotkey }}
   </div>
 </template>
--- a/packages/api-client/src/components/ScalarHotkey.vue
+++ b/packages/api-client/src/components/ScalarHotkey.vue
@@ -4,18 +4,16 @@ import { isMacOS } from '@scalar/use-tooltip'
 import { computed } from 'vue'
 
 const props = defineProps<{
-  hotkey: {
-    modifier?: HotKeyModifiers
-    key: string
-  }
+  hotkey: string
+  modifier?: HotKeyModifiers
 }>()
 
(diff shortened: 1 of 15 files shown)


hanspagel (commented):
I dig it. We need to show the ⌘K somewhere though!

antlio:
> I dig it. We need to show the ⌘K somewhere though!

still displayed in the request view sidebar

hanspagel (approved):
I think that’s great!

amritk (dismissed):
Nice work! One question though

Do these count as global hotkeys and therefore should use the bus? Or do the qualify as local evebt listeners cc: @hwkr

packages/api-client/src/components/ScalarHotkey.vue
```diff
@@ -1,42 +1,24 @@
 <script setup lang="ts">
-import { keyMap } from '@/hooks'
 import { isMacOS } from '@scalar/use-tooltip'
-import { useMagicKeys, whenever } from '@vueuse/core'
 import { computed } from 'vue'
 
 const props = defineProps<{
-  hotkey: string
+  hotkey: {
+    modifier?: string
```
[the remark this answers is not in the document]
hwkr: Also this should probably use the modifier type from https://github.com/scalar/scalar/blob/main/packages/api-client/src/libs/hot-keys.ts#L69

packages/api-client/src/layouts/App/ApiClientApp.vue
```diff
@@ -40,9 +40,26 @@ onBeforeMount(() => addScalarClassesToHeadless())
 const handleKeyDown = (ev: KeyboardEvent) =>
   handleHotKeyDown(ev, events.hotKeys, { hotKeys })
 
+const handleHotKey = (event?: HotKeyEvent) => {
+  if (!event) return
+
+  // We prevent default on open command so we can use it on the web
+  if (event.openCommandPalette) {
+    event.openCommandPalette.preventDefault()
```
hwkr, line 48: Do we need to prevent default here?

packages/oas-utils/src/entities/hotkeys/hotkeys.ts
```diff
@@ -7,7 +7,7 @@ export const HOTKEY_EVENT_NAMES = [
   'selectSearchResult',
   'navigateSearchResultsUp',
   'openCommandPalette',
-  'openCommandPaletteRequest',
+  'triggerHandler',
```
hwkr, line 10: Could this have a more descriptive name? `createNew`?

hwkr (commented):
Looks good! Left a few comments!

I wonder if we could use the workspace store or something to avoid prop drilling the `isApp` state? We could expose the hotKey config there and just check if the create new / sidebar button hotkey config is enabled and the only show the hotkey chip if it's enabled or something

Re: @amritk I think it's fine on the bus since we do want the hotkey to globally handled we just want to handle it a little bit differrent depending on the view you're looking at

packages/api-client/src/layouts/App/ApiClientApp.vue
amritk: yea I believe it will launch the default chrome command otherwise(search google)

tmastrom: N opens a new tab for me. K and return work for the app commands though

hwkr: I think search / omnibar is CMD+L

amritk: CTRL+K 😄 
image

amritk: maybe command L for you apple nerds

packages/api-client/src/components/ScalarHotkey.vue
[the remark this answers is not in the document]
antlio: good call! ✔️

packages/api-client/src/layouts/App/ApiClientApp.vue
hwkr: Welp

amritk: is it actually command L? whats command K do then?

amritk: ah okay windows has both ctrl K and ctrl E, most sites override ctrl k but you still got e, and yes apple is confirmed L so it doesn't matter. Keep the prevent default!

packages/api-client/src/components/ScalarHotkey.vue
```diff
@@ -1,42 +1,27 @@
 <script setup lang="ts">
-import { keyMap } from '@/hooks'
+import type { HotKeyModifiers } from '@scalar/oas-utils/entities'
 import { isMacOS } from '@scalar/use-tooltip'
-import { useMagicKeys, whenever } from '@vueuse/core'
 import { computed } from 'vue'
 
 const props = defineProps<{
-  hotkey: string
+  hotkey: {
```
hwkr, line 7: I was thinking more like:
```suggestion
  hotkey: string
  modifier?: HotKeyModifiers
```

(As individual props instead of an object)

Then you don't have to change how the component is used in all those places

hwkr (approved):
Looks good! Might be nice to provide the isApp logic instead of passing it down through so many layers. Maybe something like `provide('layout', 'desktop', 'web', 'modal')`?

Doesn't need to be in this PR though

antlio:
> Looks good! Might be nice to provide the isApp logic instead of passing it down through so many layers. Maybe something like `provide('layout', 'desktop', 'web', 'modal')`?
> 
> Doesn't need to be in this PR though

i like that! let's ff with this

packages/api-client/src/components/ScalarHotkey.vue
antlio: i agree, good call!

Issue #3468, opened 2024-10-13
Too many preview apps are launching during PRs


amritk:
### What happens?

Seems like a new preview launches every time you make a change and the old ones are still up. This must be incurring some extra server costs. We should either: update/replace the initial preview OR take down the old one before launching a new one. That way there's only one currently running per PR

### What did you expect to happen?

Only one instance to be active at a time

hanspagel:
Ah, I updated the PR comment action, that broke it.

PR #3484, opened 2024-10-15
chore: downgrade to `thollander/actions-comment-pull-request@v2`

hanspagel:
v3 seems to add multiple comments instead of updating them. 👀 

More context in https://github.com/thollander/actions-comment-pull-request/issues/397

See #3468, doesn’t fix it though. There are still multiple preview links generated (at least for Netlify).


hanspagel:
Skipping the queue, because we already had v2 and it was great and the bug is just annoying.

PR #3486, opened 2024-10-15
fix(themes): reset

antlio:
this pr removes the focus visible style added in the theme reset stylesheet in #3435 in order to fix ui focus issue, while we found another better approach to it.

**before**
in reference
image

in api client
image

**after**
in reference
image

in api client
image


hanspagel (approved):
Agree

PR #3488, opened 2024-10-15
style(api-client): gap between sidebar entries

antlio:
this pr adds gap to sidenav components and updates the hover color on sidehelp:

**before**
image

**after**
image


hanspagel (approved):
Nice

PR #3490, opened 2024-10-15
chore: switch from `yaml` to `js-yaml`

hanspagel:
WIP

Looks like switching from `yaml` to `js-yaml` could save 70 kB.


tmastrom:
is `js-yaml` maintained? It looks like the last publish was [2 years ago](https://github.com/nodeca/js-yaml) 👀

hanspagel:
Oh, yes. And yaml is actually updated 3 days ago. Also I’m missing an important option there. Let’s keep the 70 kB. :)

PR #3495, opened 2024-10-16
feat(api-client): create custom workspace selector

hwkr:
Due to some how we have some of our accessibility stuff set up opening the command palette from inside another modal (e.g. the import modal) is difficult™. This creates a small dialog for creating a new workspace from inside the new import modal + some other modal related cleanup.

https://github.com/user-attachments/assets/0d6261a2-c5cc-47ba-80ef-21ddc46bf91f

Also on the todo is an overhaul of all the modals to bring them into alignment soon™


hanspagel (approved):
Looks great!

PR #3499, opened 2024-10-16
docs(api-reference): promote `url` instead of `content`

hanspagel:
`url` has better performance for larger documents, can be cached by the browser, and it’s easier to import documents to let’s say <https://client.scalar.com>.

I’ve gone through all the READMEs and made sure we’re promoting `url` instead of `content` (embedded OpenAPI documents).

~~This also marks `content` as `@deprecated`. Let me know how you feel about this, but I think we should really go all-in on the URL.~~ Just adds a `@remark`.

### DIFF
--- a/packages/api-reference/README.md
+++ b/packages/api-reference/README.md
@@ -42,8 +42,8 @@ import '@scalar/api-reference/style.css'
       content="width=device-width, initial-scale=1" />
   </head>
   <body>
-    <!-- Add your own OpenAPI/Swagger specification URL here: -->
-    <!-- Note: The example is our public proxy (to avoid CORS issues). You can remove the `data-proxy-url` attribute if you don’t need it. -->
+    <!-- Need a Custom Header? Check out this example: https://codepen.io/scalarorg/pen/VwOXqam -->
+    <!-- Note: We’re using our public proxy to avoid CORS issues. You can remove the `data-proxy-url` attribute if you don’t need it. -->
     <script
       id="api-reference"
       data-url="https://cdn.jsdelivr.net/npm/@scalar/galaxy/dist/latest.yaml"
@@ -58,6 +58,7 @@ import '@scalar/api-reference/style.css'
       document.getElementById('api-reference').dataset.configuration =
         JSON.stringify(configuration)
     </script>
+
     <script src="https://cdn.jsdelivr.net/npm/@scalar/api-reference"></script>
   </body>
 </html>
--- a/packages/express-api-reference/README.md
+++ b/packages/express-api-reference/README.md
@@ -15,27 +15,9 @@ npm install @scalar/express-api-reference
 
 ## Usage
 
-[Set up Express](https://expressjs.com/en/starter/hello-world.html) and pass an OpenAPI/Swagger spec to the `apiReference` middleware:
+[Set up Express](https://expressjs.com/en/starter/hello-world.html) and pass an URL to an OpenAPI/Swagger document to the `apiReference` middleware:
 
 > Wait, but how do we get the OpenApiSpecification? 🤔 There are multiple ways to generate an OpenAPI/Swagger file for Express. The most popular way is to use [`swagger-jsdoc`](https://github.com/Surnet/swagger-jsdoc).
-
-```ts
-import { apiReference } from '@scalar/express-api-reference'
-
-const OpenApiSpecification =
-  /* … */
-
-  app.use(
-    '/reference',
-    apiReference({
-      spec: {
-        content: OpenApiSpecification,
-      },
-    }),
-  )
-```
-
-If you’re serving an OpenAPI/Swagger file already, you can pass an URL, too:
 
 ```ts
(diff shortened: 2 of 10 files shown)


amritk (approved):
hans taking us to the future

hanspagel:
I’ll wait for @marclave as this is kind of product-related.

marclave:
> I’ll wait for @marclave as this is kind of product-related.

i wonder if we should still have some comments that you can pass in a file? i dont know the distribution but i have seen a fair share of file usage, just dont want to have the user be guessing when looking at the examples so maybe url is first then we show the string based import too

packages/api-reference/README.md
```diff
@@ -58,6 +58,7 @@ import '@scalar/api-reference/style.css'
       document.getElementById('api-reference').dataset.configuration =
         JSON.stringify(configuration)
     </script>
+
```
hwkr, line 61: Is this space intentional?

packages/express-api-reference/README.md
```diff
@@ -15,35 +15,18 @@ npm install @scalar/express-api-reference
 
 ## Usage
 
-[Set up Express](https://expressjs.com/en/starter/hello-world.html) and pass an OpenAPI/Swagger spec to the `apiReference` middleware:
+[Set up Express](https://expressjs.com/en/starter/hello-world.html) and pass an URL to an OpenAPI/Swagger document to the `apiReference` middleware:
```
hwkr, line 18: I think this should be `a`

```suggestion
[Set up Express](https://expressjs.com/en/starter/hello-world.html) and pass a URL to an OpenAPI/Swagger document to the `apiReference` middleware:
```

packages/fastify-api-reference/README.md
```diff
@@ -40,37 +40,11 @@ fastify.register(require('@scalar/fastify-api-reference'), {
 })
 ```
 
-With [@fastify/swagger](https://github.com/fastify/fastify-swagger) you can even generate your OpenAPI documents from the registered routes and directly pass it to the plugin:
+With [@fastify/swagger], we’re picking it up automatically, so this would be enough:
```
hwkr, line 43: Wording suggestion (optional)

```suggestion
With [@fastify/swagger] we automatically detect the specification, so all you need is:
```

packages/nestjs-api-reference/README.md
```diff
@@ -47,7 +47,7 @@ const OpenApiSpecification =
   )
 ```
 
-If you’re serving an OpenAPI/Swagger file already, you can pass an URL, too:
+Recommended: If you’re serving an OpenAPI/Swagger file already, you can pass an URL, too:
```
hwkr, line 50: Wording

```suggestion
Recommended: If you’re serving an OpenAPI/Swagger file already, you can also pass a URL:
```

packages/types/src/legacy/reference-config.ts
```diff
@@ -329,11 +329,17 @@ export type RequestBody = {
   content?: RequestBodyMimeTypes
 }
 
-/** For providing a OAS spec object or url to be fetched */
+/** The OpenAPI Document we’ll render */
```
hwkr, line 332:
```suggestion
/** The OpenAPI Document to be rendered */
```

packages/types/src/legacy/reference-config.ts
```diff
@@ -329,11 +329,17 @@ export type RequestBody = {
...
 export type SpecConfiguration = {
-  /** URL to a Swagger/OpenAPI file */
+  /**
+   * URL to an OpenAPI/Swagger document
```
hwkr, line 335:
```suggestion
   * A URL to an OpenAPI/Swagger document
```

hwkr (commented):
Looks good to me! Added a bunch of wording / grammar comments (sorry went a little wild) feel free to take them or leave them (or shoot me a DM if you have questions).

packages/express-api-reference/README.md
hwkr: Alright I looked it up and I'm not crazy, since URL starts with a consonant sound (**Y**ou-Are-Ell) it should be "a URL" not "an URL".

This article has a good section on acronyms https://www.grammar.com/a-vs-an-when-to-use

packages/api-reference/README.md
hanspagel: Actually, yes. Not feeling strong about it, though.

hwkr: Up to you!

hwkr (approved):

packages/express-api-reference/README.md
hwkr: I think this still needs to be an `a`

packages/nestjs-api-reference/README.md
hwkr: 'a URL'

marclave (approved):
cool nice work @hanspagel :)

PR #3500, opened 2024-10-16
docs: add  `Scalar for Laravel` to the READMEs

hanspagel:
As composer (PHP package manager) doesn’t support monorepos, our Laravel package (secretly launched) has to live in its own repository. This PR makes sure it’s mentioned in the appropriate places.


amritk (approved):
Sure why not

PR #3504, opened 2024-10-16
feat(api-reference): add integration identifiers

hanspagel:
This PR adds an `_integration` identifier to the API reference configuration. All our official integration pass a string (for example `'nextjs'`) to `@scalar/api-reference` now.

This gives us the opportunity to make framework-specific optimisations or add some helpful debug information in the future.


marclave (approved):
love this

PR #3516, opened 2024-10-17
feat(api-client): toggle watch mode

antlio:
this pr is a proposal to add the watch mode toggle in the sidebar context menu as seen below:

https://github.com/user-attachments/assets/d1a3f937-3886-4bc2-9c6a-f0b6a81f3f51

### DIFF
--- a/packages/api-client/src/views/Request/RequestSidebar.vue
+++ b/packages/api-client/src/views/Request/RequestSidebar.vue
@@ -11,7 +11,7 @@ import type { HotKeyEvent } from '@/libs'
 import { useWorkspace } from '@/store'
 import RequestSidebarItemMenu from '@/views/Request/RequestSidebarItemMenu.vue'
 import { dragHandlerFactory } from '@/views/Request/handle-drag'
-import type { SidebarMenuItem } from '@/views/Request/types'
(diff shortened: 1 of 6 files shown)


hanspagel (changes requested):
We should only show this for URLs.

I’ve uploaded a file and the collection has the toggle, too:

Screenshot 2024-10-18 at 15 32 30

antlio:
> We should only show this for URLs.
> 
> I’ve uploaded a file and the collection has the toggle, too:
> 

good catch, fixed

packages/api-client/src/views/Request/RequestSidebar.vue
```diff
@@ -103,6 +103,12 @@ onMounted(() => events.hotKeys.on(handleHotKey))
 onBeforeUnmount(() => {
   events.hotKeys.off(handleHotKey)
 })
+
+const handleToggleWatchForChanges = (item?: SidebarItem) => {
+  if (item && item.watchForChanges !== undefined) {
+    item.watchForChanges = !item.watchForChanges
+  }
+}
```
hanspagel, line 111: Is it safer to check for the presence of `item.url` and maybe more self-explanatory than `item.watchForChanges !== undefined`? 🤔

hanspagel (approved):

PR #3520, opened 2024-10-17
fix(themes): remove vendor prefixes for input placeholder

hwkr:
I'm not sure why but having the vendor prefixes here was breaking the placeholder selector. This seems to fix it in chrome and has [good compatibility](https://developer.mozilla.org/en-US/docs/Web/CSS/::placeholder#browser_compatibility).

## Before

Google Chrome-2024-10-17-14-05-19@2x

## After

Google Chrome-2024-10-17-14-06-30@2x

### DIFF
--- a/.changeset/short-cats-camp.md
+++ b/.changeset/short-cats-camp.md
@@ -1,0 +1,5 @@
+---
+'@scalar/themes': patch
+---
+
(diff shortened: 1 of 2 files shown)


cameronrohani (approved):

marclave (approved):
lets add a changeset and we can merge 😎

PR #3528, opened 2024-10-18
fix(api-client): proxy setting

antlio:
this pr updates the value of the api client's default proxy setting and adds :
- in development, proxy is disabled by default
- in production, proxy is enabled by default

### DIFF
--- a/packages/api-client/src/store/store.ts
+++ b/packages/api-client/src/store/store.ts
@@ -290,8 +290,14 @@ export const createWorkspaceStore = (
   // PROXY URL STATE
   const PROXY_URL = 'proxyUrl' as const
 
-  // Initialize proxyUrl with the value from localStorage
-  const proxyUrl = ref(localStorage.getItem(PROXY_URL) || '')
+  // Initialize proxyUrl with the value from localStorage or default to the proxy URL according to env
+  const proxyUrl = ref(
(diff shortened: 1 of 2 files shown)


packages/api-client/src/store/store.ts
```diff
@@ -290,8 +290,14 @@ export const createWorkspaceStore = (
   // PROXY URL STATE
   const PROXY_URL = 'proxyUrl' as const
 
-  // Initialize proxyUrl with the value from localStorage
-  const proxyUrl = ref(localStorage.getItem(PROXY_URL) || '')
```
hanspagel, line 294: Desperately needed, but I don’t see a `proxyUrl` to be stored in local storage. Is that really where we get it from? Because I can’t see it using the browser console. In the past it has been stored in the workspace.

marclave: lets get this in but re-assess this later in a FF

right now its hard to use https://client.scalar.com without any form of error reporting that cors is disabled.

ill make an issue

marclave (approved):

packages/api-client/src/store/store.ts
marclave: https://github.com/scalar/scalar/issues/3541

PR #3530, opened 2024-10-18
style(api-client): active element background

antlio:
this pr fixes the background inconsistencies when using themes in order to favor accent background:

**before**
image

**after**
image

### DIFF
--- a/packages/api-client/src/components/DataTable/DataTableInputSelect.vue
+++ b/packages/api-client/src/components/DataTable/DataTableInputSelect.vue
@@ -98,9 +98,11 @@ watch(addingCustomValue, (newValue) => {
             :value="option"
             @click="updateSelected(option)">
             <div
-              class="flex items-center justify-center rounded-full p-[3px] w-4 h-4 group-hover/item:shadow-border"
+              class="flex items-center justify-center rounded-full p-[3px] w-4 h-4"
               :class="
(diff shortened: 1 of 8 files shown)


packages/api-client/src/components/DataTable/DataTableInputSelect.vue
```diff
@@ -98,9 +98,11 @@ watch(addingCustomValue, (newValue) => {
             :value="option"
             @click="updateSelected(option)">
             <div
-              class="flex items-center justify-center rounded-full p-[3px] w-4 h-4 group-hover/item:shadow-border"
+              class="flex items-center justify-center rounded-full p-[3px] w-4 h-4"
               :class="
-                isSelected(option) ? 'bg-blue text-b-1' : 'text-transparent'
+                isSelected(option)
+                  ? 'bg-c-accent text-b-1'
+                  : 'group-hover:shadow-border text-transparent'
```
hwkr, line 105: Is this missing `/item`?

```suggestion
                  : 'group-hover/item:shadow-border text-transparent'
```

hwkr (approved):
Looks good! Not totally related to this PR but would should probably set a more distinct name for these groups in the component library to avoid nesting conflicts.

e.g. `group/listbox` instead of `group`

PR #3531, opened 2024-10-18
chore(object-utils): disable mutation test

hanspagel:
Sorry, but this test just fails again and again and it costs me so much time to check, retry and wait for the results again and again.

Open for other solution, but if there’s no other solution, I’d wish we’d disable the test for now.

### DIFF
--- a/packages/object-utils/src/mutator-record/mutations.test.ts
+++ b/packages/object-utils/src/mutator-record/mutations.test.ts
@@ -154,7 +154,7 @@ describe('Handles history rolling', () => {
     param2: 2000,
   }
 
-  test(
+  test.todo(
     'Rolls history back to initial then forward to modified state',
     { retry: 3 },
     () => {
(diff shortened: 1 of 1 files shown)


packages/object-utils/src/mutator-record/mutations.test.ts
```diff
@@ -154,7 +154,8 @@ describe('Handles history rolling', () => {
     param2: 2000,
   }
 
-  test(
+  // TODO: This test fails way too often, there seems to be a race condition.
+  test.todo(
```
hwkr, line 158: Nit: I think you can also do `test.skip` which is maybe more applicable? I think todo is for unimplemented tests?

marclave (approved):

amritk (approved):
Yes PLZ!

packages/object-utils/src/mutator-record/mutations.test.ts
hanspagel: Never thought about this, but you’re absolutely right.

marclave:
i tried to fix this for an hour and a half, lets merge this and get it fixed later ✨

PR #3532, opened 2024-10-18
ci: pin all actions to a commit hash

hanspagel:
Our GitHub Action workflows use actions from other repositories, and so far, we’ve referenced them by version tags. However, tags can be changed, which means someone could add unwanted code to an existing tag and compromise any repository using it.

To reduce this risk, it’s a good practice to pin actions to a specific commit hash. This ensures that the exact version of the action is used, lowering the chances of security issues or unexpected changes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,13 +24,13 @@ jobs:
 
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v4
+        uses: actions/checkout@eef61447b9ff4aafe5dcd4e0bbf5d482be7e7871
       - name: Build
         uses: ./.github/actions/build
         with:
           node-version: ${{ matrix.node-version }}
       - name: Update Turborepo cache
-        uses: actions/cache/save@v4
+        uses: actions/cache/save@3624ceb22c1c5a301c8db4169662070a689d9ea8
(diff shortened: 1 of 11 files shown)


.github/workflows/ci.yml
```diff
@@ -7,6 +7,15 @@ on:
   pull_request:
 
 jobs:
+  harden_security:
+    name: Harden Security
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@5a4ac9002d0be2fb38bd78e4b4dbde5606d7042f
```
tmastrom, line 15: why is this has different from the other `actions/checkout` ?

tmastrom (commented):
This is a good idea! wonder if we can cache these hash strings and use them everywhere to ensure they are the same everywhere?

.github/workflows/ci.yml
hanspagel: HOW DID YOU SPOT THIS

hanspagel: fixed

tmastrom: hahaha I love a good hash

marclave (approved):
im happy with this :)

PR #3537, opened 2024-10-18
feat(scalar.aspnetcore): add integration configuration property

xC0dex:
## Description

This PR adds the `_integration` property to the configuration. The .NET integration provides a new `DotNetFlag` property, which is set to `true` by default.

### DIFF
--- a/.changeset/moody-points-juggle.md
+++ b/.changeset/moody-points-juggle.md
@@ -1,0 +1,5 @@
+---
+'@scalar/aspnetcore': patch
+---
+
+feat: Add DotNetFlag configuration property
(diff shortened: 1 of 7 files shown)


hanspagel (approved):
expose framework! that’s a great name for it. Thanks for helping me out here! 👏 

ah, we said expose sounds more dangerous than it has to be. ok, let me think about it.

xC0dex:
> ah, we said expose sounds more dangerous than it has to be. ok, let me think about it.

What about `ShowFramework`? We could also consider making this not configurable 🤣.

marclave:
@xC0dex awesome! happy to merge as soon as you're good to go : )

marclave:
waiting on @hanspagel based on @xC0dex conversation

> Hans want to think about an alternative property name

marclave:
> Maybe the property could be named `DotnetFlag` and the extension `WithDotnetFlag`. A nice explanation could describe what this property is used for.

yeah lets do that!

hanspagel:
Yes, that sounds great! But wouldn’t it be DotNet (with a big N)?

xC0dex:
I'll rename it later this day.

xC0dex:
Renamed to `DotNetFlag`. Hit the green button 🚀!

marclave:
@xC0dex my all time fav button

PR #3549, opened 2024-10-18
fix: turbo build cache in CI workflow

amritk:
This ONLY fixes the cache in the CI workflow. Long term we need to re-organize our workflows so that build goes first and caches, then the rest go


hanspagel:
I don’t think this will work tbh.

amritk:
no it wont, just playing around.

amritk:
Something is wrong!

PR #3550, opened 2024-10-19
fix: turbo build cache can’t be restored while writing to it

amritk:
This one seems to have worked, hopefully its not a fluke!

image

It seems to keep caching the build action so I renamed it and it started working! Open to a different name


tmastrom (approved):
I reverted the name and it still seems to be working? Adding the restore keys an fixing needs syntax appears to be the fix? Actions are arcane...

hanspagel:
Let me debug this tomorrow ✌️

amritk:
> I reverted the name and it still seems to be working? Adding the restore keys an fixing needs syntax appears to be the fix? Actions are arcane...

I tried that and it kept showing the old keys, but I guess the cache was finally busted!

.github/actions/build/action.yml
```diff
@@ -22,11 +22,13 @@ runs:
     - name: Install dependencies
       shell: bash
       run: pnpm install
-    - name: Turborepo cache
+    - name: Cache turbo build setup
       uses: actions/cache@3624ceb22c1c5a301c8db4169662070a689d9ea8
       with:
         path: .turbo
-        key: turbo-${{ runner.os }}-node-${{ inputs.node-version }}
+        key: ${{ runner.os }}-turbo-${{ github.sha }}
```
hanspagel, line 29: This is adding the commit hash to the cache key. So new commits can’t benefit from the cache anymore, right?

amritk:
@hanspagel I think the goal is to first fix the cache within a single workflow so we don't build 3 times. We can fix caching between workflows later?

hanspagel:
I’m probably too tired, a few things seem strange to me. I’m not convinced it’s better and not just different, but that might be me. Give me until tomorrow to check again.

hanspagel:
I've tested the following scenarios:

**No cache: A random PR (with changes to just the client)**
1st run: 9m 24s
2nd run (without changes): 3m 39s

**Cache: This PR here (with changes to just the api-reference)**

All builds (except the one that changed) should have been cached already.
1st run: 7m 52s (slightly better)
2nd run: IN PROGRESS

To be honest, I don’t understand everything that is going on, but let me put together a PR with a different approach and compare the results.

hanspagel:
Closing in favor of #3584, which is only slightly different. In the other PR we make sure the cache isn’t updated in the build action, but only once in the CI workflow.

PR #3553, opened 2024-10-19
fix: support server variables in request sending

marclave:
before:

image

server variable requests would fail

now they use defaults :)
image


antlio (approved):
nice, let's ff with supporting this back!

PR #3555, opened 2024-10-19
perf(api-client): improve large response performance in Chromium

rmp135:
**Problem**
Large responses are causing heavy input lag in Chromium browsers only, most recently with #3467. This is due to a performance bug in Chromium with the general sibling CSS selector that Tailwind uses for the `divide` classes. When this class is applied to an element with many, many descendants (as is the case with large parsed JSON responses) it causes the browser to halt as it re-renders the style. Firefox and Safari do not have this issue.

**Solution**
With this PR I've manually applied the same style to the two impacted components without using this selector, attempting to keep it as close as possible to the intended look. Also, through running some tests, I've found that Chrome can now render up to 200kb JSON responses, and have increased the threshold accordingly. For interest, Firefox is able to get around 400kb without major problems.

### DIFF
--- a/packages/api-client/src/components/ViewLayout/ViewLayout.vue
+++ b/packages/api-client/src/components/ViewLayout/ViewLayout.vue
@@ -1,22 +1,3 @@
-<style scoped>
-.view-layout > :not([hidden]):not(:first-child) {
-  border-top-width: calc(calc(var(--scalar-border-width) / 2));
-  border-left-width: 0px;
-}
-.view-layout > :not([hidden]):first-child {
-  border-bottom-width: 0px;
-}
-
-@screen xl {
-  .view-layout > :not([hidden]):first-child {
-    border-right-width: 0px;
-  }
-  .view-layout > :not([hidden]):not(:first-child) {
-    border-top-width: 0px;
-    border-left-width: calc(calc(var(--scalar-border-width) / 2));
-  }
-}
-</style>
 <template>
   <div
     class="flex flex-col min-h-0 flex-1 xl:overflow-hidden md:flex-row leading-3">
--- a/packages/api-client/src/components/ViewLayout/ViewLayout.vue
+++ b/packages/api-client/src/components/ViewLayout/ViewLayout.vue
@@ -19,7 +19,7 @@
 </style>
 <template>
   <div
-    class="view-layout flex flex-col min-h-0 flex-1 xl:overflow-hidden md:flex-row leading-3">
+    class="flex flex-col min-h-0 flex-1 xl:overflow-hidden md:flex-row leading-3">
     <slot />
   </div>
 </template>
--- a/packages/api-client/src/components/ViewLayout/ViewLayout.vue
+++ b/packages/api-client/src/components/ViewLayout/ViewLayout.vue
@@ -1,6 +1,25 @@
+<style scoped>
+.view-layout > :not([hidden]):not(:first-child) {
+  border-top-width: calc(calc(var(--scalar-border-width) / 2));
+  border-left-width: 0px;
+}
+.view-layout > :not([hidden]):first-child {
+  border-bottom-width: 0px;
+}
+
+@screen xl {
+  .view-layout > :not([hidden]):first-child {
+    border-right-width: 0px;
+  }
+  .view-layout > :not([hidden]):not(:first-child) {
+    border-top-width: 0px;
+    border-left-width: calc(calc(var(--scalar-border-width) / 2));
+  }
+}
+</style>
 <template>
   <div
-    class="divide-y md:divide-y-0 divide-1 flex flex-col min-h-0 flex-1 xl:overflow-hidden md:flex-row leading-3">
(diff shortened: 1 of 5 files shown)


amritk:
hey @rmp135 thanks for the PR! I wasn't aware of that chromium bug. Unfortunately the bottleneck from what I've seen has been highlight.js tokenizing the responses. Maybe one day we can find/build a virtualized syntax highlighter. 

For now I think we can take these changes but not the bump up in threshold, as at least for me it is still very laggy.

Try this endpoint, its about 100K bytes
`https://api.github.com/meta`

amritk:
OKay something strange IS going on! I tested this branch on chrome and it was laggy, but on firefox it handles that 100k no problem. So maybe we are missing something else as well

rmp135:
I've just ran that endpoint in Chrome and although there's a small initial lag (which I'm thinking is the tokeniser / DOM insertion) it's buttery smooth after that. Previous to this fix it would _very_ laggy scrolling around. I personally think 20kb is too low.

amritk:
Okay dove deeper and you are totally right, I'm just going to remove all divides from the app!
image

amritk:
Ah you know what, I was on the wrong branch 😓 I was wondering why this wasn't working. The fix is good! Nice work lets get it in

marclave:
awesome PR @rmp135 , wanna add a patch changeset :) then we can get this released!

```
# in root
pnpm changeset
```

packages/api-client/src/components/ViewLayout/ViewLayout.vue
```diff
@@ -1,6 +1,25 @@
...
+  .view-layout > :not([hidden]):not(:first-child) {
+    border-top-width: 0px;
+    border-left-width: calc(calc(var(--scalar-border-width) / 2));
+  }
+}
+</style>
 <template>
   <div
-    class="divide-y md:divide-y-0 divide-1 flex flex-col min-h-0 flex-1 xl:overflow-hidden md:flex-row leading-3">
+    class="view-layout flex flex-col min-h-0 flex-1 xl:overflow-hidden md:flex-row leading-3">
```
hwkr, line 22: It looks like these `divide`'s weren't actually doing anything so they can probably just be removed (unless I'm missing something).

```suggestion
    class="flex flex-col min-h-0 flex-1 xl:overflow-hidden md:flex-row leading-3">
```

packages/api-client/src/components/ViewLayout/ViewLayoutContent.vue
```diff
@@ -1,6 +1,25 @@
...
+  .view-layout-content > :not([hidden]):not(:first-child) {
+    border-top-width: 0px;
+    border-left-width: calc(calc(var(--scalar-border-width) / 2));
+  }
+}
+</style>
 <template>
   <div
-    class="border-t-1/2 md:border-t-0 divide divide-y xl:divide-y-0 xl:divide-x-1/2 flex xl:flex-row flex-col custom-scroll pr-0">
+    class="view-layout-content border-t-1/2 md:border-t-0 flex xl:flex-row flex-col custom-scroll pr-0">
```
hwkr, line 22: I think we can also avoid the scoped styles here and stick to some (slightly cursed) tailwind instead:

```suggestion
    class="*:border-t-1/2 first:*:border-t-0 xl:*:border-t-0 xl:*:border-l-1/2 xl:first:*:border-l-0 flex xl:flex-row flex-col custom-scroll pr-0">
```

Also for this fix we need to remove the `md:` from `md:border-b-1/2` on [RequestSubpageHeader.vue Line 22]( https://github.com/scalar/scalar/blob/main/packages/api-client/src/views/Request/RequestSubpageHeader.vue#L22) (which should be removed anyway to be honest).

Would be good to add a comment above as well:
```html
<!-- Using *:border-* to avoid chrome performance issue with adjacent sibling selector (~) -->
```

hwkr (changes requested):
Thanks for the fix @rmp135! I added some comments so we can stick to Tailwind for fixing this if that's okay 😄

I'm curious though since as far as I can tell we only ever render the `ViewLayoutContent` with a maximum of 2 children it seems weird that would affect chrome performance for the sibling selector (since even with a large response I would expect there to only be one sibling in the tree to parse). Do you have a link to the chrome bug in question?

rmp135:
@hwkr No problem, I'm not too familiar with the voodoo magic of Tailwind. 

I don't have an exact bug report, closest I've found is [this one](https://issues.chromium.org/issues/40827378).

hwkr:
> @hwkr No problem, I'm not too familiar with the voodoo magic of Tailwind. 
> 
> 
> 
> I don't have an exact bug report, closest I've found is [this one](https://issues.chromium.org/issues/40827378).

That makes sense! That could bug definitely be it. Switching to the descendant selectors should hopefully fix the issue then

hwkr (approved):
Updated with tailwind :)

amritk (approved):
Awesome work and nice find!

PR #3564, opened 2024-10-21
fix(api-client): data cell dropdown overflow issue

### LINKED ISSUE #3543: Dropdowns are broken in the settings tab
### What happens?

You can't change the settings, clicking the dropdown menu isn't do anything:
image
The only way to do it right now is to tab into them

Be able to change the settings

### OpenAPI Document

_No response_

antlio:
this pr fixes #3543 and the dropdown overflow issue within api client data table cell - mistakenly added to fix tooltip gradient visibility, now fixed using full height:

**before**
image

**after
image

### DIFF
--- a/.changeset/green-eyes-act.md
+++ b/.changeset/green-eyes-act.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
+
+fix: data table cell overflow issue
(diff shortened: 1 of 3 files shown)


hwkr (approved):

marclave (approved):
we are so back 🫡

PR #3565, opened 2024-10-21
style(api-client): tooltip fixtures

antlio:
this pr styles tooltip position/brightness and decreases back the request body content size:

**before**
image

**after**
image

### DIFF
--- a/packages/api-client/src/views/Request/RequestSection/RequestBody.vue
+++ b/packages/api-client/src/views/Request/RequestSection/RequestBody.vue
@@ -505,3 +505,8 @@ watch(
     </DataTable>
   </ViewLayoutCollapse>
 </template>
+<style scoped>
+:deep(.cm-content) {
+  font-size: var(--scalar-mini);
+}
+</style>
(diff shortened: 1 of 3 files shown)


packages/api-client/src/views/Request/RequestSection/RequestBody.vue
```diff
@@ -505,3 +505,8 @@ watch(
     </DataTable>
   </ViewLayoutCollapse>
 </template>
+<style scoped>
+:deep(.cm-content) {
+  font-size: var(--scalar-mini);
```
hwkr, line 510: @cameronrohani I think we could probably update this in the main code block component, the default font size there is hard coded and a little big in my opinion

hwkr (approved):
Looks good!

packages/api-client/src/views/Request/RequestSection/RequestBody.vue
antlio: yes, it's actually used in the environment view, but since it's only used there we could switch the default!

PR #3567, opened 2024-10-21
fix(import): sandbox links with uppercase letters can’t be resolved

hanspagel:
Oops, sandbox links that contain uppercase letters aren’t resolved by `@scalar/import`. This PR fixes it.

And let’s allow `null` as a value, too. Let’s just try to be as flexible as possible with what we take as an input.


amritk (approved):
the `resolve.ts` file could use some early returns to flatten it out, otherwise its fine

PR #3568, opened 2024-10-21
fix(fastify-api-reference): JSON endpoint does not have CORS headers

hanspagel:
This PR adds CORS headers to the JSON endpoint again, we must have lost them in some merge conflicts.


tmastrom (approved):
Nice work adding tests to both json and yaml endpoints!

PR #3570, opened 2024-10-21
fix(api-client): request path variables url

antlio:
this pr is fast following on #3553 in order to display path variable items in tabs if available in the server.

**before**
image

**after**
image


amritk (commented):
This is more of a product one so I'll leave it for marc

marclave:
@antlio dont we want to update these server variables globally? will dm on slack

hanspagel:
+1 for defining them globally

PR #3574, opened 2024-10-21
feat(api-client): env var dropdown keyboard navigation

antlio:
this pr adds keyboard navigation within environment variable dropdown as seen below:

https://github.com/user-attachments/assets/3e958fed-4fa9-48e3-a568-924bc092e498

### DIFF
--- a/.changeset/modern-phones-jump.md
+++ b/.changeset/modern-phones-jump.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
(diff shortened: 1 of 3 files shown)


amritk (approved):
Works great. 

Unrelated but when you delete an env var and re-select it it doesn't get the nice UI
image

antlio:
> Works great.
> 

yes great catch! started to dig that up in another branch

PR #3578, opened 2024-10-21
chore: upgrade Rollup and Vite

tmastrom:
Upgrades rollup to the latest version. This will hopefully fix issues with type checking [here](https://github.com/scalar/scalar/actions/runs/11439568546/job/31823841952?pr=3455)


hanspagel:
I think we can fix the type check in the PR in a different way, but we should still update rollup. Let me try to fix it first and let’s then update rollup, if that’s ok.

hanspagel (approved):
Vite depends on Rollup, so we have to update them at the same time it seems.

PR #3580, opened 2024-10-21
style(api-reference): required string orange color

cameronrohani:
move required string rightmost + add back the 🍊 colour :) 

image


amritk (approved):
lets go! needs changeset

PR #3582, opened 2024-10-22
feat(postman-to-openapi): new Postman to OpenAPI converter 💌

Mason-Little:
# Add @scalar/postman-to-openapi Package

## Problem
I decided to build a custom Scalar Postman-to-OpenAPI converter because the existing ones are not well-maintained and often fail to convert specs from services like PayPal or Stripe due to edge cases.

## Solution
This is a custom parser designed to handle edge cases more effectively. I am still working through all the tests, so this is very much a work in progress. Any reviews would be greatly appreciated.

## Overview

This PR introduces a new package, `@scalar/postman-to-openapi`, which provides an efficient way to convert Postman collections into OpenAPI 3.0 specifications. This tool makes it easy to inject Postman specs directly into the client or references. The goal is to simplify migration from Postman to OpenAPI.

## Usage

```typescript
import { convert } from '@scalar/postman-to-openapi'

// Convert your Postman collection to OpenAPI
const result = await convert(myPostmanCollection)
```

## Features

- Converts Postman Collection v2.0/v2.1 to OpenAPI 3.1.0.
- Processes multiple request body types (JSON, form-data, URL-encoded).
- Handles common authentication methods (Basic, Bearer, API Key).
- Extracts and converts example responses.
- Maintains folder structure as tags.
- Preserves request and response headers.
- Handles path and query parameters.
- Supports server URL extraction.
- Handles large Postman collections.

## Example

Input (Postman Collection):

```json
{
  "info": {
    "name": "My API",
    "description": "API Description",
    "version": "1.0.0"
  },
  "item": [
    {
      "name": "Create User",
      "request": {
        "method": "POST",
        "url": "https://api.example.com/users",
        "body": {
          "mode": "raw",
          "raw": "{\"name\": \"John\"}"
        }
      }
    }
  ]
}
```

Output (OpenAPI 3.0):

```json
{
  "openapi": "3.0.0",
  "info": {
    "title": "My API",
    "description": "API Description",
    "version": "1.0.0"
  },
  "paths": {
    "/users": {
      "post": {
        "summary": "Create User",
        "requestBody": {
          "content": {
            "application/json": {
              "schema": {
                "type": "object",
                "properties": {
                  "name": {
                    "type": "string"
                  }
                }
              }
            }
          }
        }
      }
    }
  }
}
```

## Next Steps

- [x] Expand Testing
- [x] Refactor code to avoid mutating the generated OpenAPI spec within helper files. Instead, modifications will occur in the main `convert.ts` file.
- [x] move tests to CDN

## Credits

This package is based on the original [`postman-to-openapi`](https://github.com/joolfe/postman-to-openapi) by @joolfe. It has been adapted and enhanced for the Scalar ecosystem.


marclave:
@geoffgscott @hanspagel @Mason-Little 

for this PR, we have a lot of fixtures for our test cases, im not too familiar with github repo size limits or anything but is that something we should be worried about? 

or adding all these openapi specs do we want to move them to a separate openapi files repo or something?

geoffgscott:
> @geoffgscott @hanspagel @Mason-Little
> 
> for this PR, we have a lot of fixtures for our test cases, im not too familiar with github repo size limits or anything but is that something we should be worried about?
> 
> or adding all these openapi specs do we want to move them to a separate openapi files repo or something?

Any test specs should live on the CDN and not in the repo.

tmastrom:
> > @geoffgscott @hanspagel @Mason-Little
> > for this PR, we have a lot of fixtures for our test cases, im not too familiar with github repo size limits or anything but is that something we should be worried about?
> > or adding all these openapi specs do we want to move them to a separate openapi files repo or something?
> 
> Any test specs should live on the CDN and not in the repo. @tmastrom can assist with this.

Yes we can copy the pattern we use in `@scalar/openapi-parser` for testing 👍

packages/postman-to-openapi/src/helpers/responseHelpers.ts
```diff
@@ -0,0 +1,61 @@
...
+  return responses.reduce((openapiResponses, response) => {
+    const statusCode = response.code?.toString() || 'default'
+
+    openapiResponses[statusCode] = {
+      description: response.status || '',
+      headers: extractHeaders(response.header),
+      content: {
+        'application/json': {
+          schema: inferSchemaFromExample(response.body || ''),
+          example: tryParseJson(response.body || ''),
```
hanspagel, line 27: For OpenAPI 3.1 make this 

```js
{
  examples: {
    default: tryParseJson(response.body || ''),
  }
}
```

packages/postman-to-openapi/src/helpers/responseHelpers.ts
```diff
@@ -0,0 +1,61 @@
...
+  if (!headers || typeof headers === 'string') {
+    return undefined
+  }
+  const openapiHeaders: { [key: string]: OpenAPIV3.HeaderObject } = {}
+  if (Array.isArray(headers)) {
+    headers.forEach((header) => {
+      openapiHeaders[header.key] = {
+        schema: {
+          type: 'string',
+          example: header.value,
```
hanspagel, line 47: For OpenAPI 3.1 make this 

```js
{
  examples: [header.value],
}
```

packages/postman-to-openapi/src/convert.ts
```diff
@@ -0,0 +1,180 @@
...
+          }
+        }
+
+        // Handle the /raw endpoint specifically
+        if (normalizedPathKey === '/raw' && pathItem?.post) {
+          if (pathItem.post.requestBody?.content['text/plain']) {
+            pathItem.post.requestBody.content['application/json'] = {
+              schema: {
+                type: 'object',
+                example: '',
```
hanspagel, line 113:
```suggestion
```

I don't think we need an empty example here.

packages/postman-to-openapi/src/helpers/requestBodyHelpers.ts
```diff
@@ -0,0 +1,100 @@
...
+  body: RequestBody,
+  requestBody: OpenAPIV3.RequestBodyObject,
+) {
+  try {
+    const jsonBody = JSON.parse(body.raw || '')
+    requestBody.content = {
+      'application/json': {
+        schema: {
+          type: 'object',
+          example: jsonBody,
```
hanspagel, line 46: For OpenAPI 3.1

```suggestion
          examples: { default: jsonBody },
```

packages/postman-to-openapi/src/helpers/requestBodyHelpers.ts
```diff
@@ -0,0 +1,100 @@
...
+        },
+      },
+    }
+  } catch (error) {
+    requestBody.content = {
+      'text/plain': {
+        schema: {
+          type: 'string',
+          example: body.raw,
```
hanspagel, line 55: For OpenAPI 3.1

```suggestion
          examples: [body.raw],
```

packages/postman-to-openapi/src/helpers/requestBodyHelpers.ts
```diff
@@ -0,0 +1,100 @@
...
+    type: 'object',
+    properties: {},
+    required: [],
+  }
+  urlencoded.forEach((item: UrlEncodedParameter) => {
+    if (schema.properties) {
+      const paramObject = createParameterObject(item, 'query')
+      schema.properties[item.key] = {
+        type: 'string',
+        example: item.value,
```
hanspagel, line 87:
```suggestion
        examples: [item.value],
```

packages/postman-to-openapi/src/convert.ts
hanspagel: Alternative: If we do need an empty attribute, make this an empty array:

```suggestion
                examples: [],
```

packages/postman-to-openapi/src/convert.ts
```diff
@@ -0,0 +1,180 @@
...
+import { normalizePath } from './helpers/urlHelpers'
+import type { PostmanCollection } from './types'
+
+/**
+ * Converts a Postman Collection to an OpenAPI 3.0.0 document.
+ * This function processes the collection's information, servers, authentication,
+ * and items to create a corresponding OpenAPI structure.
+ */
+export function convert(
+  postmanCollection: PostmanCollection,
```
hanspagel, line 18: Can we allow to pass a string here, too? We’d need to `JSON.parse` it then. But I like it when things are easy with the input and strict with the output.

```suggestion
  postmanCollection: PostmanCollection | string,
```

hanspagel (changes requested):
OMG, so nice. Left a bunch of tiny comments.

packages/postman-to-openapi/package.json
```diff
@@ -0,0 +1,67 @@
...
+  "keywords": [
+    "postman",
+    "openapi",
+    "swagger",
+    "converter",
+    "transformer",
+    "export",
+    "scalar"
+  ],
+  "version": "0.0.0",
```
geoffgscott, line 22: This should probably be 0.0.1 or 1.0.0

packages/postman-to-openapi/package.json
```diff
@@ -0,0 +1,67 @@
...
+    "./css/*.css": {
+      "import": "./dist/css/*.css",
+      "require": "./dist/css/*.css",
+      "default": "./dist/css/*.css"
+    },
+    "./*.css": {
+      "import": "./dist/*.css",
+      "require": "./dist/*.css",
+      "default": "./dist/*.css"
+    }
```
geoffgscott, line 53: Should be no reason to export CSS. Might need to add a setting on the rollup config

packages/postman-to-openapi/src/convert.test.ts
```diff
@@ -0,0 +1,247 @@
+import { beforeAll, describe, expect, it } from 'vitest'
+
+import { downloadFileToMemory } from '../../openapi-parser/tests/utils/downloadFileGcp'
```
geoffgscott, line 3: No relative sibling package imports. For this particular use case you can just use a fetch function with the URL to the bucket.

packages/postman-to-openapi/env.d.ts
geoffgscott, line 1: We should not have these files around. If you need to use raw import you can only do it with a vite build which means you should have `type: ["vite/client"]` to support it.

geoffgscott (changes requested):

packages/postman-to-openapi/package.json
amritk: I thought there was an includeCss flag or something

amritk (approved):
Well written PR! Nothing really jumped out at me that hasn't been addressed

packages/postman-to-openapi/src/convert.test.ts
```diff
@@ -0,0 +1,247 @@
...
+          )
+        }),
+      ]
+
+      await Promise.all(allPromises)
+    } catch (error) {
+      console.error('Failed to download test files:', error)
+      throw error
+    }
+  }, 60000)
```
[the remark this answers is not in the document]
Mason-Little: ya that was a weird one, my network we acting up one day so I had to up that a ton, and I guess I forgot to change it back :sweat_smile:

Mason-Little:
@amritk This commit might need a bit of review since I changed files in `oas-utils` and `api-client`. I also hope I understood the review comments correctly. I added it as a commit on its own so it is easy to revert if this is not the inteaded use

amritk:
Close enough! lol works for me

packages/postman-to-openapi/src/helpers/urlHelpers.ts
```diff
@@ -0,0 +1,63 @@
...
+  let match
+
+  while ((match = templateVariableRegex.exec(path)) !== null) {
+    // match[1] is for double curly braces, match[2] is for single curly braces
+    // if neither exists, it's a colon parameter
+    const param = match[1] || match[2] || match[0].slice(1)
+    params.push(param.trim())
+  }
+
+  return [...new Set(params)] // Deduplicate parameters
```
geoffgscott, line 62: This could be simplified by just making `params` a set to begin with. Should we also be converting the string to lowerCase?

For all of these functions in urlHelpers I would like to see at least a couple simple unit tests. These could come in a follow up PR if that is preferred.

Mason-Little: as far as the version @hanspagel said in DMs that he couldn't comment on the PR but 

> "This PR includes a changeset and the release workflow will bump it to 0.1.0 on the first release.
Semantic versions start at 0.1.0.
So having 0.0.0 before the first release is correct."

hanspagel (approved):
let’s get this out

hanspagel:
Tests are failing. Just took a quick look: The recent code changes look correct, but the tests might need an update.

PR #3583, opened 2024-10-22
Fix/client always uses proxy

hanspagel:
Currently, the API client is just always using our public proxy.

We just recently merged the following code:

```ts
  const proxyUrl = ref(
    localStorage.getItem(PROXY_URL) !== null
      ? localStorage.getItem(PROXY_URL) || ''
      : import.meta.env.PROD
        ? 'https://proxy.scalar.com'
        : '',
```

This basically say, use what’s in localStorage, and if nothing is localStorage, just use proxy.scalar.com. It doesn’t take the configuration into account.

With this PR the configured proxy is passed to `createWorkspaceStore` and the decision is made like this:

1. Use what’s in localStorage
2. If nothing is set in localStorage, use the proxy from the configuration

If there’s a custom proxy configured, a third setting for the custom proxy shows up on the settings page:

Screenshot 2024-10-22 at 11 55 03

Note: I’ve modified the localStorage key. Otherwise users that already had the proxyUrl in the localStorage, wouldn’t notice any changes.

### DIFF
--- a/.changeset/tasty-donkeys-behave.md
+++ b/.changeset/tasty-donkeys-behave.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
+
+fix: API client always uses proxy.scalar.com
--- a/packages/api-client/src/libs/create-client.ts
+++ b/packages/api-client/src/libs/create-client.ts
@@ -68,6 +68,7 @@ export const createApiClient = ({
 }: CreateApiClientParams) => {
   const store = createWorkspaceStore(router, {
     useLocalStorage: persistData,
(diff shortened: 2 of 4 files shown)


antlio:
great job! we could update the description to lead to the doc about customize the proxy?

also this doesn't seem to hold the selection on reload anymore.

hanspagel:
OMG, yes, fixed this!

Mh … We’ve had to link to all configuration attributes then, right? Maybe that’s too much.

marclave (approved):

marclave:
@amritk let's just merge this than move it to workspace later

amritk (approved):
KK send it

PR #3584, opened 2024-10-22
ci: in the reusable build action, only restore from the cache

hanspagel:
Currently, the build action that we use everywhere restores *and* writes to the cache. This produces conflicts.

Screenshot 2024-10-22 at 13 09 15

With this PR we’re making sure, we’re only restoring from the cache. There’s only one place where we write to the Turborepo cache, and this is in the build job of the ci workflow. So after the build was done the first time.

**Results**

1st run: 9m 17s
2nd run: 3m 45s

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,7 +33,7 @@ jobs:
         uses: actions/cache/save@3624ceb22c1c5a301c8db4169662070a689d9ea8
         with:
           path: .turbo
-          key: turbo-${{ runner.os }}-node-${{ matrix.node-version }}
+          key: turbo-${{ runner.os }}-node-${{ matrix.node-version }}-${{ github.sha }}
 
   format:
     runs-on: ubuntu-20.04
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,7 +29,7 @@ jobs:
         uses: ./.github/actions/build
         with:
           node-version: ${{ matrix.node-version }}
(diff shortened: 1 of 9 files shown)


.github/workflows/ci.yml
```diff
@@ -29,11 +29,11 @@ jobs:
         uses: ./.github/actions/build
         with:
           node-version: ${{ matrix.node-version }}
-      - name: Update Turborepo cache
+      - name: Save Turborepo cache
         uses: actions/cache/save@3624ceb22c1c5a301c8db4169662070a689d9ea8
         with:
           path: .turbo
-          key: turbo-${{ runner.os }}-node-${{ matrix.node-version }}
+          key: turbo-${{ runner.os }}-node-${{ matrix.node-version }}-${{ github.sha }}
```
tmastrom, line 36: Does this mean the cache is per PR / push?

tmastrom (commented):
Amazing work 🙌  I can do a more detailed review tomorrow

.github/workflows/ci.yml
hanspagel: Cache is always scoped to the PR, and this scopes it to the commit hash, *BUT* the restore keys make sure it can be restored from other commits.

amritk (approved):
Looks to be working to me! Nice fix

.github/workflows/ci.yml
tmastrom: Ok thank you for the clarification

PR #3586, opened 2024-10-22
test(api-client-app): playwright smoke test

tmastrom:
**Problem**
Currently, …

**Solution**
With this PR …


hanspagel (approved):
Moarr tesstss

hanspagel:
I think I found the issue, should be resolved with #3785.

PR #3589, opened 2024-10-22
fix(api-client): address bar hotkey submission

antlio:
this pr removes the stopped propagation on key enter within code input, which is preventing the meta+enter action from submitting the request from the address bar focus.


amritk (approved):
Lets go

PR #3590, opened 2024-10-22
style(api-client): address bar loading background

antlio:
this pr increases z index on loading background in address bar component, in order to fix color inconsistencies when submitting request without focus as seen below:

**before**
image

**after**
image


amritk (approved):
nice catch

hwkr:
🔥

PR #3593, opened 2024-10-22
fix(openapi-parser): update `uri` types to be `uri-reference` in schema v3.1

rpmccarter:
According to the OpenAPI 3.1 spec ([link](https://swagger.io/specification/#:~:text=to%20RFC3986.-,Relative%20References%20in%20URLs,-Unless%20specified%20otherwise)),

> Unless specified otherwise, all properties that are URLs MAY be relative references as defined by [RFC3986](https://tools.ietf.org/html/rfc3986#section-4.2). Unless specified otherwise, relative references are resolved using the URLs defined in the [Server Object](https://swagger.io/specification/#server-object) as a Base URL. Note that these themselves MAY be relative to the referring document.

It seems the v3.0 spec already uses `uri-reference`. This PR updates v3.1 to use the same

### DIFF
--- a/.changeset/early-cobras-fail.md
+++ b/.changeset/early-cobras-fail.md
@@ -1,0 +1,5 @@
+---
+'@scalar/openapi-parser': patch
+---
+
+Allow relative URLs in v3.1 documents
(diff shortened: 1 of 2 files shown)


hanspagel:
Thanks for the PR! Does this work with variables?

rpmccarter:
@hanspagel do you mean server variables? I would imagine you aren't allowed to place server variables in the relative urls, only in the `servers` field

hanspagel:
@coderabbit review

hanspagel (approved):
Great update!

PR #3595, opened 2024-10-22
feat(api-reference): create workspace store in references

amritk:
This is the prequel to merging auth. This just imports the client workspace store into references and starts using it for the modal directly. Allows us to delete a few other files like the event bus etc since we can now manipulate the workspace store directly

Previously we didn't worry too much about SSR on the client since it was client only, now the JS will be loaded in references so it will be SSR'ed. However the client is still initialized on mounted, only the JS is loaded in SSR. So I had to create the last router factory

I think down the road we can instantiate the workspace store earlier and only create the router on mounted. That should be SSR safe and would give us the workspace store throughout references. To do this we would need to split out the parts of the store that require a router as we cannot create the router on the server. We would then be able to access `useWorkspaceStore` in references. What do you think?

## Release Notes

- **New Features**
	- Enhanced client-side authentication process.
	- Introduced a new router creation function, `createWebHistoryRouter`, improving routing capabilities.
	- Added `selectedSecuritySchemeUids` to the schema for better security management.

- **Improvements**
	- Streamlined API client initialization with the new `useApiClient` hook.
	- Updated handling of OpenAPI document loading for improved performance and reactivity.

- **Bug Fixes**
	- Adjusted fallback mechanisms for missing parameters in routing.

- **Chores**
	- Removed deprecated event bus implementation for API client interactions.
	- Removed the `useApiClientStore` and related logic, simplifying sidebar item management.


hanspagel:
@coderabbit review

hanspagel (approved):
Looks good to me! I’m hyped for having just the API client store for auth and servers in both packages.

I’m not entirely convinced we should use the workspace store for for everything in the reference, but we can chat about that. ✌️

PR #3596, opened 2024-10-23
fix(api-reference): sidebar a11y improvements

hwkr:
Various improvements to the references sidebar.

## Keyboard Focus Outlines Before & After

https://github.com/user-attachments/assets/06bcf541-fe9b-4825-8060-7f23113395d9

## Screen Reader

Keep in mind that this is voiceover in Chrome and all screen readers work a little bit differently.

### Before

https://github.com/user-attachments/assets/ff082dbf-f242-4591-b497-d510478979f1

### After

https://github.com/user-attachments/assets/d18da811-919e-4259-9ba2-7f3641ea2fcc

### DIFF
--- a/.changeset/tidy-pandas-tease.md
+++ b/.changeset/tidy-pandas-tease.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
+fix(api-reference): sidebar a11y improvements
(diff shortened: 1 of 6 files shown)


hanspagel (approved):
Nice improvements!

PR #3236, opened 2024-09-25
feat(api-reference): the button ⚡

cameronrohani:
image
image

## Release Notes

- **New Features**
  - Introduced a sidebar button for improved navigation.
  - Added a `WatchModeToggle` component to enhance user control over watch functionality.
  - Integrated an `OpenApiClientButton` for direct access to OpenAPI documentation.

- **Enhancements**
  - Improved drag-and-drop functionality with a more responsive design.
  - Updated `ImportCollectionModal` for better interactivity and visual feedback.
  - Enhanced `DarkModeToggle` with a modern slider design.
  - Added a new `error` variant to the `ScalarModal` for improved visual feedback.

- **Bug Fixes**
  - Refined error handling and response management in URL fetching functions.

- **Documentation**
  - Updated comments and documentation for clarity on future enhancements.

PR #3601, opened 2024-10-23
style(api-client): top navigation

antlio:
this pr moves the app top border to the topnav component in order to fix the glitch from the preview modal in #3236, and also sets [overscroll behavior](https://developer.mozilla.org/en-US/docs/Web/CSS/overscroll-behavior) and fix icon size consistencies between web / desktop:

**before**
import modal
image

sidenav icon
image

**after**

import modal
image

sidenav icon
image


hanspagel (approved):
Love your attention to detail

PR #3611, opened 2024-10-23
feat(api-client): standalone build

amritk:
Requesting @hanspagel help for the rest of this pipeline, or just point me in the right direction

Edit: Actually I don't think anything else is required right? Once its on NPM jsdelivr will just host it :)

Should we reduce the duplicated code?


amritk:
This one is not that big of a deal anymore since gitbook has another solution, may be valuable to anyone else that wants a few KB bundle size for the client but we can re-open if that happens

PR #3621, opened 2024-10-25
feat(api-reference): tag and operation accessibility

hwkr:
## Before / After

https://github.com/user-attachments/assets/d4bde006-5abf-480c-8d5c-cdfa321a4493

### DIFF
--- a/.changeset/sweet-lemons-whisper.md
+++ b/.changeset/sweet-lemons-whisper.md
@@ -1,0 +1,7 @@
+---
+'@scalar/api-reference': patch
+'@scalar/components': patch
+'@scalar/themes': patch
+---
+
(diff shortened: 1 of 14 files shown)


amritk (approved):
Code looks good, didn't test. Like I told anthony, maybe we should start moving away from headless/vue if its still stuck on 1.7

hwkr:
> Code looks good, didn't test. Like I told anthony, maybe we should start moving away from headless/vue if its still stuck on 1.7

Yeah I agree, I considered hand rolling the disclosure but looked into it and it would have been complicated. I'm down to either switch to another library or start building our own accessibility components.

marclave (approved):

cameronrohani (approved):

PR #3622, opened 2024-10-25
test(api-reference): make the standalone test stable

tmastrom:
**Problem**
Currently, the standalone test is flaky in CI on Node 18

**Solution**
This PR:
- adds retries 
- explicitly adds the jsdom file directive for the runtime environment
- combines the tests because they build off each other to guarantee execution order

### DIFF
--- a/packages/api-reference/src/standalone.spec.ts
+++ b/packages/api-reference/src/standalone.spec.ts
@@ -4,9 +4,9 @@ import { describe, expect, it } from 'vitest'
 import { waitFor } from './helpers/waitFor'
 import type { ReferenceProps } from './types'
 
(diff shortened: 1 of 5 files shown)


hanspagel (changes requested):
Looks like some assertions need to be updated

Screenshot 2024-10-25 at 11 37 59

packages/api-reference/src/standalone.spec.ts
```diff
@@ -1,3 +1,4 @@
+// @vitest-environment jsdom
```
hanspagel, line 1: not sure this is need! would you mind to remove it and try again?

hanspagel (approved):
Hans 🤝 Tom

Issue #2558, opened 2024-07-19
activate query params by default only if required


hanspagel:
> The API Playground client seems to set all the query parameters as "on" by default.  (See checkboxes in screenshot). 
>
> In my opinion I think that, unless the parameter is defined as required, it would be better to default to the parameters being unchecked, unless a value is explicitly entered into the value field.
>
> ![image](https://github.com/user-attachments/assets/f24a5b50-c5ee-4805-8071-f52fd01fd263)

_Originally posted by JP in https://github.com/scalar/scalar/discussions/1710_</div>

marclave:
this was driving me nuts https://github.com/scalar/scalar/pull/2569

PR #3626, opened 2024-10-25
fix(api-client): selection style

antlio:
this pr sets back selection style disabled in #3618 along adding user select property on web layout for a better app feeling + adds missing variables to use codemirror.

> [!NOTE]
> we should wait for [#2558](https://github.com/GitbookIO/gitbook/pull/2558) to be merged prior to merge this one

### DIFF
--- a/packages/themes/src/variables.css
+++ b/packages/themes/src/variables.css
@@ -49,7 +49,6 @@
 
   --scalar-text-decoration: none;
   --scalar-text-decoration-hover: underline;
-  --scalar-selection-background: rgb(177, 212, 252);
 }
 .dark-mode {
   color-scheme: dark;
--- a/packages/themes/src/variables.css
+++ b/packages/themes/src/variables.css
@@ -49,6 +49,7 @@
 
   --scalar-text-decoration: none;
   --scalar-text-decoration-hover: underline;
+  --scalar-selection-background: rgb(177, 212, 252);
 }
 .dark-mode {
   color-scheme: dark;
@@ -106,8 +107,3 @@
     --scalar-page-description: 20px;
   }
 }
-/* High contrast custom selection */
-/* *::selection {
-  background: var(--scalar-selection-background);
-  color: var(--scalar-selection-color);
-} */
--- a/.changeset/rich-llamas-live.md
+++ b/.changeset/rich-llamas-live.md
@@ -1,0 +1,7 @@
+---
+'@scalar/use-codemirror': patch
+'@scalar/api-client': patch
+'@scalar/themes': patch
+---
+
+fix: sets selection style properly"
--- a/packages/api-client/src/components/CodeInput/CodeInput.vue
+++ b/packages/api-client/src/components/CodeInput/CodeInput.vue
@@ -441,8 +441,4 @@ export default {
   border-radius: var(--scalar-radius);
   color: var(--scalar-color-1);
 }
-/* .cm-focused .cm-content ::selection {
-  background: var(--scalar-selection-background) !important;
-  color: var(--scalar-selection-color);
-} */
 </style>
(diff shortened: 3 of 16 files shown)


packages/themes/src/variables.css
```diff
@@ -107,7 +108,7 @@
   }
 }
 /* High contrast custom selection */
-/* *::selection {
+*::selection {
```
amritk, line 111: This needs to be scoped or it escapes our app into other peoples apps

amritk (changes requested):

packages/themes/src/variables.css
antlio: moved it in reset stylesheet which is scoped - might make more sense

amritk: going to tag in @hwkr to see if thats the best spot, but it does work

hwkr: I think it makes sense to be in our reset :)

amritk (approved):
yolo lets go

packages/api-client/src/layouts/Web/ApiClientWeb.vue
```diff
@@ -80,6 +80,11 @@ body {
   overscroll-behavior: none;
 }
 
+body {
+  cursor: default;
+  user-select: none;
```
hwkr, line 85: Personally I hate when web apps disable user select since it means I can't copy things. I'd probably prefer to set this in specific places rather than at the body level and risk having users not be able to copy something they otherwise should be able to copy.

Thoughts @cameronrohani?

antlio: do you have specific example of copyable thing to make sure we do not miss any? i feel like its easier to go that way vs the opposite - what do you think?

i tend to agree however in that client context i feel like it helps giving that native app feeling + copyable infos seems very few / be handled

hwkr: I mean sometimes it can be useful to copy an entire table / set of values. In this PR we can't select any of the response header labels or values or the response body let alone error messages if someone wanted to try Googling those.

I think general consensus it to only disable user select where absolutely necessary. 
Some good responses on the subject: https://ux.stackexchange.com/questions/83184/is-user-select-none-bad-ux

> The goal in employing user-select: none should always be limited to reducing user frustration from unintended selections resulting from interaction that can easily cause them when attempting to carry out the primary task purpose of a control it is used on.

> You should never use user-select: none to prevent selection of content text. The browser itself is responsible for deciding between scrolling and selection in this case and should be relied upon to do its job properly in discerning user intent between the two potential actions (selection or scrolling). While there will always be exceptions, keep in mind that these two actions are globally consistent expected interactions. Interfering with them will likely lead to your application not behaving in a manner that is expected, which is a primary source of frustration.

antlio: each headers are clickable and redirect to definition, however we should allow their values to be copyable yes! that's a good example!

antlio: no strong opinion anyway - the main point was to reduce all these unneeded selectable elements and felt easier to treat this way as we allow copy to most of the things + mostly have inputs within the app - so not against dropping it too if we feel that way

cameronrohani: seems like there could potentially be some a11y & UX implications here as brynn has pointed out so we should prioritize that and evaluate it on a case by case basis if need be? :) Cool with both of you? 😄

antlio: don't think much impact on a11y as it's only a css property so i guess it should not impact that part but will drop 👍

jmanomoon (approved):

hwkr (approved):
Looks good to me!

cameronrohani:
Awesome work @antlio! I'd like to see tweaks here before this goes in

- 'Default theme' should use browser defaults and not take over selection style
- All other themes can take over selection style but instead of doing high contrast I think we should stick to doing more opaque pastel colours and keep existing text colours

Also do we have a fallback variable in case they don't have something defined?

cameronrohani (approved):

PR #3633, opened 2024-10-25
chore: Add details on configuration.md

sanscontext:
**Problem**
Currently, there's a lil ambiguity on the `hiddenClients` config.

**Solution**
With this PR I hopefully cleared a little bit up.

(I couldn't make sense of `By default hides Unirest` on line 240 since that doesn't _seem_ to map to `fetch`, but feel free to add it back.)

### DIFF
--- a/documentation/configuration.md
+++ b/documentation/configuration.md
@@ -229,15 +229,9 @@ You can explicitly set the default HTTP client, though:
 
 #### hiddenClients?: array | true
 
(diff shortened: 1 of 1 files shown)


marclave (approved):
AWESOME docs improvement @sanscontext thank you so much :)

sanscontext:
Also lol your linter is case sensitive and I triggered it by accident. 😆

Issue #3636, opened 2024-10-26
Authentication inputs do not have an overflow 


ga1az:
### What happens?

Authentication inputs do not have an overflow when they have many characters.

![CleanShot 2024-10-26 at 02 27 18@2x](https://github.com/user-attachments/assets/b62e8cc0-a979-4a8f-b56a-365faee13ca6)

### What did you expect to happen?

![CleanShot 2024-10-26 at 02 27 46@2x](https://github.com/user-attachments/assets/ae103e88-126a-4af4-a308-fbe54a4cd821)

PR #3637, opened 2024-10-26
fix(api-client): overflow is added

ga1az:
**Problem**
This PR solves the issue #3636, authentication entries do not overflow when they have many characters.

**Solution**

image


marclave (approved):

marclave:
awesome work @ga1az !!!

PR #3642, opened 2024-10-27
style(api-client): fix border color variable

cameronrohani:
I have no clue why this bg variable changes border bottom color on live but not the local build version but need to get rid of it

### DIFF
--- a/packages/components/src/components/ScalarSearchInput/ScalarSearchInput.vue
+++ b/packages/components/src/components/ScalarSearchInput/ScalarSearchInput.vue
@@ -32,7 +32,7 @@ const attrs = computed(() => {
 })
 
 const variants = cva({
-  base: 'flex items-center rounded border bg-b-1/2 text-sm font-medium has-[:focus-visible]:bg-b-1 has-[:focus-visible]:outline',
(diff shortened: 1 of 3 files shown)


antlio (approved):
✨

packages/components/src/components/ScalarSearchInput/ScalarSearchInput.vue
```diff
@@ -32,7 +32,7 @@ const attrs = computed(() => {
 })
 
 const variants = cva({
-  base: 'flex items-center rounded border bg-b-1/2 text-sm font-medium has-[:focus-visible]:bg-b-1 has-[:focus-visible]:outline',
```
hwkr, line 35: I think tailwind might be interpreting it as `bg-b-1` with a transparency of "2" see https://tailwindcss.com/docs/background-color#changing-the-opacity

Did you try making the variable `1-2` instead of `1/2`?

cameronrohani: it is changing border colors for some reason I'm not sure of, i want it to mix bg-b-1 and bg-b-2

marclave (approved):

PR #3643, opened 2024-10-27
style(api-reference): fix some minor font sizes

cameronrohani:
fix font size / styling  of `read only` and `property detail values`


hanspagel (approved):
Love every typography tweaks

PR #3648, opened 2024-10-28
fix(api-client): environment default value

antlio:
this pr sets back the default environment variable value according to latest workspace refactoring:

image


amritk (approved):
Seems to be working 👍🏾

PR #3649, opened 2024-10-28
fix(api-client): web layout hotkey display

antlio:
this pr favors provide / inject approach to condition scalar hotkey display in the web layout vs previous `isApp` prop set.

**on desktop**
image
image

**on web**
image
image

### DIFF
--- a/packages/api-client/src/components/Sidebar/SidebarButton.vue
+++ b/packages/api-client/src/components/Sidebar/SidebarButton.vue
@@ -2,11 +2,16 @@
 import ScalarHotkey from '@/components/ScalarHotkey.vue'
 import { ScalarButton } from '@scalar/components'
 
-const props = defineProps<{
-  click: () => void
-  hotkey?: string
-  isApp?: boolean
-}>()
+const props = withDefaults(
+  defineProps<{
+    click: () => void
+    hotkey?: string
+    layout?: 'modal' | 'web' | 'desktop'
+  }>(),
+  {
+    layout: 'desktop',
+  },
+)
 
 const handleClick = () => {
   props.click()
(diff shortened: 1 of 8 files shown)


amritk (commented):
What is `isDesktop` ?

antlio:
> What is `isDesktop` ?

a way to check the layout, could use "layout" instead but then we would have to add it to every layout?

hwkr:
> > What is `isDesktop` ?
> 
> a way to check the layout, could use "layout" instead but then we would have to add it to every layout?

I feel like `layout` → `'modal' | 'web' | 'desktop'` would be good (imo)

packages/api-client/src/components/Sidebar/SidebarButton.vue
```diff
@@ -2,11 +2,16 @@
...
-const props = defineProps<{
-  click: () => void
-  hotkey?: string
-  isApp?: boolean
-}>()
+const props = withDefaults(
+  defineProps<{
+    click: () => void
+    hotkey?: string
+    isDesktop?: boolean
```
hwkr, line 9: Is there a reason to pass this still as a prop vs injecting it?

antlio: we want to have the hotkey displayed on request sidebar (while hiding it on the others). prob should be done differently

antlio:
> > > What is `isDesktop` ?
> 
> > 
> 
> > a way to check the layout, could use "layout" instead but then we would have to add it to every layout?
> 
> 
> 
> I feel like `layout` → `'modal' | 'web' | 'desktop'` would be good (imo)

would this be at the layout level?

hwkr:
> > I feel like `layout` → `'modal' | 'web' | 'desktop'` would be good (imo)
> 
> would this be at the layout level?

Yeah that's what I was thinking

marclave (approved):

PR #3651, opened 2024-10-28
fix(oas-utils): parameter example value

### LINKED ISSUE #3456: parameter example value are not retrieved
### What happens?

It seems like the api client is not retrieving the parameter example, and yet not display it as a query parameter table value:

```yaml
  parameters:
  - example: example value
    in: query
    name: parameter_name
    required: true
    type: string
```

image

### OpenAPI Document

_No response_

antlio:
this pr fixes #3456 by adding to parameter schema example and examples along setting theme aside schema examples as request example value.

this should now enable to retrieve value on the following usage:
```yaml
name: limit
in: query
description: The number of items to return
example: 123
```

### DIFF
--- a/packages/oas-utils/src/entities/spec/parameters.ts
+++ b/packages/oas-utils/src/entities/spec/parameters.ts
@@ -30,6 +30,16 @@ export const oasParameterSchema = z.object({
   content: z.unknown().optional(),
   /** Defaulted according to @url https://spec.openapis.org/oas/v3.1.0#parameter-object */
   style: parameterStyleSchema.optional(),
+  example: z.unknown().optional(),
+  examples: z
+    .record(
+      z.string(),
+      z.object({
+        value: z.unknown(),
+        summary: z.string().optional(),
+      }),
+    )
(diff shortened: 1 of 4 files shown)


packages/oas-utils/src/entities/spec/parameters.ts
```diff
@@ -30,6 +30,8 @@ export const oasParameterSchema = z.object({
   content: z.unknown().optional(),
   /** Defaulted according to @url https://spec.openapis.org/oas/v3.1.0#parameter-object */
   style: parameterStyleSchema.optional(),
+  example: z.unknown().optional(),
+  examples: z.array(z.unknown()).optional(),
```
geoffgscott, line 34: I believe this is a record keyed by example name and not an array. 

image

Can you also add a unit test for this schema? We should really start adding them for all schemas.

antlio: oh doo catch, i added this one too quickly

marclave (approved):

PR #3664, opened 2024-10-29
feat(api-client): environment variable colors

### LINKED ISSUE #2730: custom colors for environments
### What happens?

This is in fact not a bug, see https://discord.com/channels/1135330207960678410/1268127488471732245/1268235804535296030

I think giving us the option for pre-defined colors is great but we should always have the option to choose our own via a color picker. 
Maybe we can have a mix of pre-defined colors and a custom picker?

-

### OpenAPI Document

_No response_

antlio:
this pr:
- fixes environment variable pill color
- adds color modal on environment creation
- fixes #2730 by introducing custom color

https://github.com/user-attachments/assets/b239bb5e-54a1-4edf-b5a6-003e5c70fff9


amritk:
Just a thought, but should all envs have the same variables with different values? Or is that not how these things work

antlio:
> Just a thought, but should all envs have the same variables with different values? Or is that not how these things work

totally - as of right now each env has its own sets of variables - however your suggestion would make the most sense

marclave (approved):

PR #3670, opened 2024-10-29
feat(api-reference): disclosure button for enum values

antlio:
this pr adds a disclosure button over 4 enum values as proposed in #3320 in order to avoid too long list displayed at all time:

image

### DIFF
--- a/.changeset/cold-planes-compete.md
+++ b/.changeset/cold-planes-compete.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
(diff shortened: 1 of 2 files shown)


amritk (approved):
I think we should start reducing our dependency on headless, but this is fine for now

antlio:
> I think we should start reducing our dependency on headless, but this is fine for now

I agree! just followed what we have to fix but we should ff with this!

marclave (approved):

PR #3677, opened 2024-10-30
chore(galaxy): align galaxy package.json with prettier

hwkr:
No more removed new line at the end of the `galaxy/package.json`!

Also updates the fetch `package.json`.

Also runs the `ci` → `format` action _after_ the build so that our build tooling can't create prettier errors (and pass build) like was happening with the galaxy package json.

It would be nice to confirm that the `build` command doesn't leave any uncommited files, but I looked it up and it seemed complicated with GH actions (if someone wants to add that though that would be cool).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,7 +36,6 @@ jobs:
           key: turbo-${{ runner.os }}-node-${{ matrix.node-version }}-${{ github.sha }}
 
   format:
-    needs: [build]
     runs-on: ubuntu-20.04
     strategy:
       matrix:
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -44,14 +45,10 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683
-      - uses: pnpm/action-setup@fe02b34f77f8bc703788d5817da081398fad5dd2
-      - name: Use Node.js ${{ matrix.node-version }}
-        uses: actions/setup-node@39370e3970a6d050c480ffad4ff0ed4d3fdee5af
+      - name: Build
+        uses: ./.github/actions/build
```
hwkr, line 49: I copied the way `ci` → `types` works but let me know if I did this right.

amritk (approved):
🤯 wait for @hanspagel but if it gets rid of that annoyingness

.github/workflows/ci.yml
hanspagel: Unfortunately, this is is not working as you’d think. With and without your changes the build will run first and pass no matter if there’s formatting issues.

But the CI workflow will fail in both cases in the same way.

We can run prettier *before* the build and require prettier to pass before we do a build, but that would extend CI times by 40s or so.

We can chat to find a better solution! ✌️

hanspagel (approved):
OMG FINALLY

This is was so annoying

I’ve had to reset the CI workflow though, read my comment above! ✌️

.github/workflows/ci.yml
hwkr: I'm not sure I follow but sure we can chat about it

PR #3678, opened 2024-10-30
fix(api-client): api client button is always visible

hwkr:
Makes it so the button is only shown if the spec uses a URL.

We can't use `link` to check if the spec has a URL here because we construct it in a computed property and it's therefore always defined.

### DIFF
--- a/packages/api-client/src/components/OpenApiClientButton.vue
+++ b/packages/api-client/src/components/OpenApiClientButton.vue
@@ -11,18 +11,27 @@ const { integration, isDevelopment, url, buttonSource } = defineProps<{
 }>()
 
 /** Link to import an OpenAPI document */
-const href = computed(() => {
+const href = computed((): string | undefined => {
+  const absoluteUrl = makeUrlAbsolute(url)
+
+  if (!absoluteUrl?.length) {
+    return undefined
+  }
+
+  // Base URL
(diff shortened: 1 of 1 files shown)


packages/api-client/src/components/OpenApiClientButton.vue
```diff
@@ -3,37 +3,37 @@ import { ScalarIcon } from '@scalar/components'
 import { makeUrlAbsolute } from '@scalar/oas-utils/helpers'
 import { computed } from 'vue'
 
-const { integration, isDevelopment, url, buttonSource } = defineProps<{
```
hwkr, line 6: I'm pretty sure this [breaks reactivity](https://vuejs.org/guide/extras/reactivity-transform#reactive-props-destructure).

marclave:
gunna wait till @hanspagel chimes in on this since he built all this and i think he had a fix for some things so its fair to wait

hanspagel (approved):
OMG, great catch! The computed is also returning an URL in any case, I‘ve added so it returns undefined if we don’t have an URL. Just a matter of taste, but I think it’s easier to grasp when the thing in `v-if` is the same thing that is used inside.

And yes, we can add more cases here very soon, but let’s just iterate with a few smaller PRs. :)

PR #3690, opened 2024-10-30
refactor(themes): apply variables to `:root` instead of the body class

hwkr:
@antlio pointed out that we could use `:has` to apply our color variables to the `:root` instead of `body.light-mode` which I think is a good idea. It seems like `:has` has good support and this allows CSS variables to be used anywhere including on the `html`.

While I was in there I also did some tidying of themes and updated our theme snapshots to make sure these changes didn't break anything (which I'm pretty they haven't).


antlio (approved):
great job and addition of the update theme screens! tested and haven't seen any flaws so far

cameronrohani:
Does this increase the precedence of variables applied at root? 

If so does it break any custom themes?

hwkr:
> Does this increase the precedence of variables applied at root? 
> 
> If so does it break any custom themes? 
> 
> 
>  

All of this stuff still get added in a css layer so it shouldn't be able to break any themes

amritk (approved):

PR #3691, opened 2024-10-30
fix(api-reference): strange jump on lazy loading

### LINKED ISSUE #3447: Inconsistent scroll position on page load when navigating to a specific endpoint
### What happens?

First of all, massive congrats to everyone for this scalar project. It really is top notch in a lot of ways compared to other similar tools out there. It has been a joy to use and I'm excited to see how far this package goes. 

For context, I have been using the react wrapper in a next 13 app: "@scalar/api-reference-react": "^0.3.106"

          <ApiReferenceReact
            configuration={{
              spec: { content: spec },
              withDefaultFonts: false,
              hideModels: true,
              tagsSorter: 'alpha',
              searchHotKey: 'k',
              hideDarkModeToggle: true,
              hideDownloadButton: true,
              hiddenClients: specType === 'events' ? true : [],
              defaultHttpClient: {
                targetKey: 'shell',
                clientKey: 'curl',
              },
              operationsSorter: 'alpha'
            }}
          />

One issue I'm seeing pretty consistently is when I attempt to navigate to a specific api endpoint url (not just the root of the page), scalar attempts to scroll to the correct position, but it doesn't always consistently take me to the endpoint I'm looking for. Sometimes it does, sometimes it doesn't, sometimes the scroll stops on a section that needs "Show more". The problem still persists even with { defaultOpenAllTags: true }, so that leads to me believe the issue may either be with my css, or within the package itself. Potentially the calculation of page height is effecting something? Overflow-x issue? 

I do have a fixed header on top of the scalar component, where I'm setting --scalar-custom-header-height to account for the height of my custom header. Perhaps this is messing up some calculation?

root.style.setProperty('--scalar-custom-header-height', `${headerHeight}px`);

I imagine others have run into this before and I'm curious if there are any tips or recommendations for how to resolve! Please let me know if I can share more details. 

Much appreciated!

Regardless of whats tags are open or not, i expect all URLs to scroll to their appropriate position. 

### OpenAPI Document

_No response_

amritk:
closes #3447

not sure why this only happened on next.js 🤷🏾 Added ApiReferenceReact page to the next.js example as well

before:

https://github.com/user-attachments/assets/045c37a6-1bd7-4c89-b09c-ca2fe7c078c7

after:

https://github.com/user-attachments/assets/a9d744f9-049e-4665-99fa-5a53a9461eb9

### DIFF
--- a/packages/api-reference/src/components/Content/Lazy/Loading.vue
+++ b/packages/api-reference/src/components/Content/Lazy/Loading.vue
@@ -211,7 +211,7 @@ onMounted(() => {
     top: calc(var(--refs-header-height) - 1px);
   }
 }
-.references-loading-hidden-tag .section-container .section:first-child {
+.references-loading-hidden-tag .section-container > .section:first-child {
   display: none;
 }
 </style>
--- a/.changeset/shiny-tips-hang.md
+++ b/.changeset/shiny-tips-hang.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
+fix: api references jumping in next.js
--- a/examples/nextjs-api-reference/app/another-reference/page.tsx
+++ b/examples/nextjs-api-reference/app/another-reference/page.tsx
@@ -1,0 +1,27 @@
+'use client'
+
+import { ApiReferenceReact } from '@scalar/api-reference-react'
+
+export default function ApiReferencePage() {
+  return (
+    <ApiReferenceReact
+      configuration={{
+        spec: {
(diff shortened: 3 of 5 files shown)


packages/api-reference/src/components/Content/Lazy/Loading.vue
```diff
@@ -211,7 +211,7 @@ onMounted(() => {
     top: calc(var(--refs-header-height) - 1px);
   }
 }
-.references-loading-hidden-tag .section-container .section:first-child {
+.references-loading-hidden-tag .section-container > .section:first-child {
```
amritk, line 214: this is the fix

marclave (approved):
nice 😎

PR #3709, opened 2024-10-31
fix(api-client): environment content copy capability

antlio:
this pr adds the following elements to the environment view:
- removes copy capability from sidebar item
- adds copy capability to code input
- updates style (sidebar / code input)
- adds lint to code input

**lint before / after**
image
image

**style before / after**
imageimage


packages/api-client/src/components/CodeInput/CodeInput.vue
```diff
@@ -335,6 +356,48 @@ export default {
...
+.copy-button:focus-visible {
+  opacity: 100;
+}
+.scalar-code-copy {
+  align-items: flex-start;
+  display: flex;
+  inset: 0;
+  justify-content: flex-end;
+  position: sticky;
+}
```
antlio, line 370: to ff with extracting it as a component

amritk (approved):
anthony with the env vars:


v2 would be nice if we had a fix lint button, or even better on save/blur/timeout/whatever

antlio:
> v2 would be nice if we had a fix lint button, or even better on save/blur/timeout/whatever

yess could cherry pick the one from the org

PR #3710, opened 2024-10-31
fix(api-client): sidebar form danger button

antlio:
this pr fixes the style of the danger button on sidebar list form component:

**before / after**
image
image

### DIFF
--- a/packages/api-client/src/components/Sidebar/Actions/SidebarListElementForm.vue
+++ b/packages/api-client/src/components/Sidebar/Actions/SidebarListElementForm.vue
@@ -1,6 +1,5 @@
 <script setup lang="ts">
 import { ScalarButton } from '@scalar/components'
-import { cva, cx } from 'cva'
 
 defineProps<{
   danger?: boolean
@@ -11,16 +10,6 @@ const emit = defineEmits<{
   (e: 'cancel'): void
   (e: 'submit'): void
 }>()
-
-const variants = cva({
-  base: 'gap-1.5 font-medium px-2.5 h-8 shadow-none focus:outline-none',
-  variants: {
-    danger: {
-      true: 'delete-warning-button',
-      false: '',
-    },
-  },
-})
 </script>
 <template>
   <form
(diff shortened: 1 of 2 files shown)


packages/api-client/src/components/Sidebar/Actions/SidebarListElementForm.vue
```diff
@@ -36,21 +25,21 @@ const variants = cva({
         Cancel
       </ScalarButton>
       <ScalarButton
-        class="custom-scroll whitespace-nowrap"
-        :class="cx(variants({ danger }))"
-        type="submit">
+        class="gap-1.5 font-medium px-2.5 h-8 shadow-none focus:outline-none custom-scroll whitespace-nowrap"
```
amritk, line 28: so the theory behind the variants is we don't need to add other classes, we just have set variants and it would make them all uniform as well. But in practice that is never the case 😉

amritk (approved):

packages/api-client/src/components/Sidebar/Actions/SidebarListElementForm.vue
antlio: wait i have no idea what's that 😂 wrong rebase

amritk: 👍🏾 no worries, my dream is to one day have an awesome, strict design system!

amritk: yes size can usually be a prop

amritk: wait size is already a prop?

antlio: yes it is, but the sm for instance keep the height of the medium

antlio: but i'll dig that up in that pr, i think we use small button for auth or request body, so we could switch them up to `xs` and then use small for `h-8` here?

amritk: You don't need to do anything now, just keep it in mind! 🧠

hanspagel:
Dangerous to use the title in the button isn’t it? Some people have pretty long names for their operations.

antlio:
> Dangerous to use the title in the button isn’t it? Some people have pretty long names for their operations.

true, right now though it gets truncated passed a certain length, however we might want to use a generic wording. Exceptionally thought about using name here as a guard to confirm deletion of the right item

PR #3712, opened 2024-10-31
feat(galaxy): load the spec from a URL rather than as content

hwkr:
Loads the Scalar Galaxy Spec by URL rather than via `content` so you can open it in the API client.

I'm assuming the playground is what we deploy to `galaxy.scalar.com`, let me know if it's something else...

## Before (galaxy.scalar.com)

Google Chrome-2024-10-31-13-38-51@2x

## After (this PR)

Google Chrome-2024-10-31-13-38-28@2x

### DIFF
--- a/.changeset/itchy-cougars-relate.md
+++ b/.changeset/itchy-cougars-relate.md
@@ -1,0 +1,5 @@
+---
+'@scalar/galaxy': patch
+---
+
+feat(galaxy): load the spec from a URL rather than as content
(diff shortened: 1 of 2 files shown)


hanspagel (approved):
Yesss

hanspagel (changes requested):
No need to add the openapi.yaml route, it’s there already (mock server)

hwkr:
> No need to add the openapi.yaml route, it’s there already (mock server)

Ahhhh ok ok

marclave (approved):

hanspagel (approved):

PR #3713, opened 2024-10-31
fix(api-reference): pre-select first auth in the modal when no auth selected

### LINKED ISSUE #1939: Authorization header does not set when auth is optional
### What happens?

image

`Authorization` header doesn't set in CDN mode.

I think it's related with `$ref` issue.

image

image

1. Deploy with CDN
2. Send request
3. Authorization does not set

### OpenAPI Document

https://sandbox.scalar.com/e/G8yz4

amritk:
fixes #1939

This just pre-selects the first auth in modal mode if there's no preferredSecurityScheme passed in to match the reference auth.

### DIFF
--- a/.changeset/silly-planes-yawn.md
+++ b/.changeset/silly-planes-yawn.md
@@ -1,0 +1,5 @@
+---
+'@scalar/oas-utils': patch
+---
+
+fix: preselect auth in the modal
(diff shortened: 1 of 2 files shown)


marclave (approved):
awesome work @amritk

PR #3714, opened 2024-10-31
fix(api-client): default redirectURI and prefix relative redirectURIs with activeServer

### LINKED ISSUE #3699: oauth redirect URI should default to current URL
the default redirect URL should be the current URL I think, because we want to have the access code in the reference/client, not on some other domain

### LINKED ISSUE #3700: authorize button doesn‘t work with relative URLs
oauth2 authorize button doesn‘t work with relative URLs (/oauth/authorize), they should use currently selected server as an URL

amritk:
fixes #3699 
fixes #3700

### DIFF
--- a/packages/api-client/src/views/Request/libs/oauth2.ts
+++ b/packages/api-client/src/views/Request/libs/oauth2.ts
@@ -2,6 +2,7 @@ import type {
   SecuritySchemeExampleValue,
   SecuritySchemeOauth2,
   SecuritySchemeOauth2ExampleValue,
+  Server,
 } from '@scalar/oas-utils/entities/spec'
 
 /** Oauth2 security schemes which are not implicit */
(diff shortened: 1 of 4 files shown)


packages/api-client/src/views/Request/libs/oauth2.ts
```diff
@@ -52,9 +55,25 @@ export const authorizeOauth2 = (
       else if (scheme.flow.type === 'authorizationCode')
         url.searchParams.set('response_type', 'code')
 
+      // Handle relative redirect uris
+      if (scheme.flow['x-scalar-redirect-uri'].startsWith('/')) {
```
hanspagel, line 59: Let me add docs for x-scalar-redirect-uri tomorrow ✌️

hanspagel: Why do we need it?

amritk: why do we need to check for the `/` or why do we need x-scalar-redirect-uri?

amritk: if its the latter, lets discuss

hanspagel: The latter, why do we need the custom attribute and can’t just use the existing redirectUri attribute?

amritk: Can tag @geoffgscott but this is something I would like to discuss in person

PR #3715, opened 2024-10-31
feat(api-client): add tests for oauth2

### LINKED ISSUE #3702: oauth code flow should use the access code to retrieve a token
OAuth 2.0 Code Flow: The user comes back with a temporary access code, which then needs to be sent to the API to retrieve the actual token.

Currently, the access code is just used as a token.

amritk:
Added tests for what we have first, then I will add the new stuff

closes #3702

### DIFF
--- a/packages/api-client/src/views/Request/libs/oauth2.test.ts
+++ b/packages/api-client/src/views/Request/libs/oauth2.test.ts
@@ -14,7 +14,7 @@ const baseScheme: Pick<
   'uid': 'test-scheme',
   'nameKey': 'test-name-key',
   'type': 'oauth2',
-  'x-scalar-client-id': 'client123',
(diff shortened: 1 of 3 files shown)


packages/api-client/src/views/Request/libs/oauth2.test.ts
```diff
@@ -0,0 +1,565 @@
...
+        }),
+        headers: {
+          'Authorization': `Basic ${secretAuth}`,
+          'Content-Type': 'application/x-www-form-urlencoded',
+        },
+      })
+    })
+  })
+
+  // describe('PKCE Flow', () => {
```
amritk, line 351: will uncomment these in the next PR

geoffgscott (approved):

PR #3716, opened 2024-11-01
feat(galaxy): add collection icon to Scalar Galaxy ✨

hwkr:
Adds a collection icon to the Scalar Galaxy spec ~~and shows the collection icon (if present) in the import modal~~ 🥳

https://github.com/user-attachments/assets/f22a2d58-1c03-4ac5-9154-2fbdc0dbd888

### DIFF
--- a/.changeset/short-boxes-help.md
+++ b/.changeset/short-boxes-help.md
@@ -1,0 +1,5 @@
+---
+'@scalar/galaxy': patch
+---
+
+feat(galaxy): add x-scalar-icon
--- a/packages/galaxy/src/specifications/3.1.yaml
+++ b/packages/galaxy/src/specifications/3.1.yaml
@@ -66,6 +66,7 @@ security:
   - apiKeyHeader: []
   - apiKeyCookie: []
   - oAuth2: []
+x-scalar-icon: interface-favorite-stars-sparkles
 tags:
   - name: Authentication
     description:


hanspagel:
Shouldn’t live `x-scalar-icon` in `info` where the title, description and such things are? 🤔

marclave (changes requested):
we should probably use this extension? 


https://redocly.com/docs/api-reference-docs/specification-extensions/x-logo

marclave:
love this :)

hwkr:
@hanspagel @marclave Both of those things make sense to me but this is where it's [already added in the schema](https://github.com/scalar/scalar/blob/main/packages/oas-utils/src/entities/spec/collection.ts#L35) so if we want to move it, rename it or change it I think we'd need a migration (but yes I agree it makes more sense in `info`).

Also it's worth note the `x-logo` schema is pretty different from this which just supports a string that defines an icon from the icon library. I'd love to add more general support in the future but it seems like the way Redocly uses `x-logo` is also pretty different.

Google Chrome-2024-11-01-10-26-39@2x

cameronrohani:
Just chatted with Brynn we're going to remove the icon from the modal because it will _eventually_ compete with the logo 

but otherwise it's good to go from me :)

cameronrohani (approved):

cameronrohani:
closed in favour of a planned release in the next few months <3

PR #3717, opened 2024-11-01
feat(Scalar.AspNetCore): add StringSyntax attribute to urls

xC0dex:
This PR adds 2 little annotations to URL parameters to tell the IDE that a URI is expected here. IDE's like Rider or Visual Studio understand this attribute and will provide better syntax highlighting.


hanspagel (approved):
Nice DX update! ✨

PR #3719, opened 2024-11-01
fix(proxy-server): redirects with relative URLs in the `Location` header, fix #3668

### LINKED ISSUE #3668: proxy doesn't follow relative redirects
Example: https://proxy.scalar.com/?scalar_url=https%3A%2F%2Falpha.api.worknet.ai%2Fdocs%2Fv1

Response Header: `Location: /docs/v1/`

Redirected to: `https://proxy.scalar.com/docs/v1/`

hanspagel:
When the proxy receives a response containing a `Location` header with a relative URL (e.g. `Location: /foobar`), it redirects to the proxy itself.

With this PR this is fixed and the relative URL is prefixed with the original URL, that we’d like to fetch.

This PR also adds some tests (finally) to the proxy server.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -89,6 +89,12 @@ jobs:
           packages-only: 'true'
       - name: Run tests
         run: pnpm test:ci
+      - name: Setup Go
+        uses: actions/setup-go@v4
+        with:
+          go-version: '1.21'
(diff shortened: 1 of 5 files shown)


marclave (approved):

marclave:
awesome work @hanspagel 

gunna wait for 1 more review before merging

PR #3723, opened 2024-11-01
fix(api-client): request height section

antlio:
this pr fixes the height issue on request section for non readonly spotted by @hanspagel : 

**before / after**
image
image


hanspagel (approved):
🤝

amritk (approved):

PR #3727, opened 2024-11-01
test(api-reference): cdn snapshot testing (macOS)

tmastrom:
Re-enable snapshot testing on the live `api-reference`. 

The action is now running on `macos-latest`. 

The main issue I was having with maintaining the snapshots is I’m running macos but the action is running ubuntu so to update the snapshots I was running the testing in a docker container and then exporting the snapshots.

Since the main people maintaining this are on mac it could be a good idea to make the action run in a macos environment for easy updates.

We can also revisit the conversation about using a snapshot testing tool!


amritk (approved):
Seems to be ok, will I be able to update the screenshots using linux? If not maybe we take/set the screenshots in the CI of a PR

hwkr (approved):
Looks great!

PR #3728, opened 2024-11-01
refactor(themes): apply variables to :root instead of the body class

hwkr:
@antlio pointed out that we could use :has to apply our color variables to the :root instead of body.light-mode which I think is a good idea. It seems like :has has good support and this allows CSS variables to be used anywhere including on the html.

### DIFF
--- a/.changeset/quiet-feet-melt.md
+++ b/.changeset/quiet-feet-melt.md
@@ -1,0 +1,6 @@
+---
+'@scalar/api-client': patch
+'@scalar/themes': patch
+---
+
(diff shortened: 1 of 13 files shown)


amritk (approved):
Looks ok to me, how will it affect other themes not in here? Also while we are refactoring css.... should we do light-dark too? get rid of that annoying flash?

amritk:
@hwkr this ones getting stale!

hwkr:
Yeah we can close it, I think we should tackle it differently.

PR #3729, opened 2024-11-01
fix(oas-utils): relative URLs are wrong

hanspagel:
When the OpenAPI document URL is relative and we try to make it an absolute URL this is happening currently:

❌ `https://example.com/swagger` + `swagger.json` = `https://example.com/swagger/swagger.json`

But we actually want this:

✅ `https://example.com/swagger` + `swagger.json` = `https://example.com/swagger.json`

Without touching this case (trailing slash):

✅ `https://example.com/swagger/` + `swagger.json` = `https://example.com/swagger/swagger.json`

### DIFF
--- a/.changeset/lazy-moose-talk.md
+++ b/.changeset/lazy-moose-talk.md
@@ -1,0 +1,5 @@
+---
+'@scalar/oas-utils': patch
+---
+
+fix: relative URLs are created wrong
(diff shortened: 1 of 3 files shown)


amritk (approved):

amritk (changes requested):
Actually making another PR 1 sec

amritk (approved):
Nope I lied, send it

PR #3736, opened 2024-11-03
feat: added another option for rust integration

Courtcircuits:
**Problem**
Currently, I struggled a bit with the current rust integration of scalar since it was to much oriented towards axum and I couldn't make it work with actix.

**Solution**
With this PR I propose another integration of scalar in rust that focuses on integrating scalar in any rust web framework. Don't get me wrong, tamasfe's crate is really good, but its DX is too much scoped towards axum.

Here is the crate I propose : [scalar-doc.rs](https://crates.io/crates/scalar-doc)

### DIFF
--- a/documentation/integrations/rust.md
+++ b/documentation/integrations/rust.md
@@ -31,13 +31,14 @@ let router: ApiRouter = ApiRouter::new()
 ```
 
 If you want to use Scalar with Actix or any other web framework, you can do that too with the crate [scalar-doc](https://crates.io/crates/scalar-doc).
-To get started with Actix, first install the crate and activate `actix` feature : 
-  
+To get started with Actix, first install the crate and activate `actix` feature :
+
 ```bash
 cargo add scalar-doc -F actix
 ```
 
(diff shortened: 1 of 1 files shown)


marclave (approved):
hey @Courtcircuits ! thanks so much for this PR and thanks so much for making that crate!

i love the idea of having a more flexible integration, and this is a great addition to the docs :)

🦀

Courtcircuits:
> hey @Courtcircuits looks like theres a prettier issue, if you run
> 
> `pnpm format` in root it should fix it :)
> 
> ```
> Run pnpm format:check
> 
> > root@1.0.0 format:check /home/runner/work/scalar/scalar
> > pnpm prettier --check .
> 
> Checking formatting...
> [warn] documentation/integrations/rust.md
> [warn] Code style issues found in the above file. Run Prettier with --write to fix.
>  ELIFECYCLE  Command failed with exit code 1.
> ```
> 
> then we can get it merged ✨

It should be fine now :)

PR #3741, opened 2024-11-04
fix(api-client): request examples in the modal

antlio:
this pr a first attempt in order to update the api client modal example used according to api reference selected example:

**before / after**

https://github.com/user-attachments/assets/b97f088c-4388-432b-be23-1d52132b3b3a

https://github.com/user-attachments/assets/6ee66c4f-269a-4edd-9a65-ce02d0d17acc


packages/api-client/src/libs/create-client.ts
```diff
@@ -333,5 +338,26 @@ export const createApiClient = ({
...
+    updateExample: (exampleKey: string, operationId: string) => {
+      if (!exampleKey || !operationId) return
+
+      const request = Object.values(requests).find(
+        ({ operationId: reqOperationId }) => reqOperationId === operationId,
+      )
+      if (!request) return
+
+      const contentType =
+        Object.keys(request.requestBody?.content || {})[0] || 'application/json'
```
amritk, line 351: This will be fine for now, but we will probably add multiple content type support soon

packages/api-client/src/libs/create-client.ts
```diff
@@ -333,5 +338,26 @@ export const createApiClient = ({
     modalState,
     /* The workspace store */
     store,
+    /** Update the currently selected example */
+    updateExample: (exampleKey: string, operationId: string) => {
+      if (!exampleKey || !operationId) return
```
amritk, line 343: this is also good enough for now, but if requested we can also handle cases of no operationId

amritk (changes requested):
I would get ride of `useExampleStore` and just use `useApiClient` directly instead. 

```ts
const { client } = useApiClient()
client.value?.updateExample
```

no need for watchers 😄

packages/api-client/src/libs/create-client.ts
antlio: noted! it is one way to retrieve the request but can also be done differently

jpjpjp:
Super excited to see this!   I'm commenting just so I get a notification when/if this gets merged.

Excuse my ignorance about this, but I get a little confused by the use of the term "client" in the scalar world.  Is it safe to assume that changes in this PR would apply to docs generated by different libraries and not just work on the scalar client when it is fed a spec?

jpjpjp:
> hey @jpjpjp - you did well! so yes basically the api client package is consumed by the api reference ones so the work there once released and package bumped will be accessible in libraries. which one are you using?

I'm using [@scalar/express-api-reference](https://www.npmjs.com/package/@scalar/express-api-reference), so after this PR is merged, I'll keep my eye out for updates to that library.

amritk (approved):
Alright took a look and I didn't realize the store was being used elsewhere as well, thought it was just for updating the client. Approved 👍🏾

hanspagel (approved):
Niice

PR #3732, opened 2024-11-01
feat(api-client): improve oauth2 error handling + add PKCE capabilities with tests

amritk:
closes #604
closes #3705

Added a super basic front-end
image

and tested it using spotify (they support pkce)
image

PR #3751, opened 2024-11-04
fix(api-client): hide scopes when there are none

### LINKED ISSUE #3739: OAuth: Scope parameter should be optional
### What happens?

Currently, the `scope` parameter is always sent to the identity provider, even when there are no scopes defined in the OpenAPI document. This behavior is causing issues, particularly when using identity providers like Keycloak.

The `scope` parameter should only be sent to the identity provider if there are scopes selected in the API client. This aligns with the current behavior of SwaggerUI, which omits the scope parameter when no scopes are specified. Additionally, it would improve UX if the scope input field were hidden in cases where no selectable scopes are defined in the OpenAPI document. 

### OpenAPI Document

_No response_

amritk:
closes #3739

we will eventually need a way to create scopes. Also do not send them when there are none selected (don't send empty scopes=)

Will fix the tests after https://github.com/scalar/scalar/pull/3732 merges since there was some changes


xC0dex (approved):
As a C# dev, I think the TS code looks great!

hanspagel (approved):
Niice

PR #3753, opened 2024-11-05
fix ref bug

rpmccarter:
**Problem**
Currently, any properties named `$ref` get treated as references, and cause errors because their value is usually `{ type: 'string' }`. This could potentially be a problem with any dynamic-key field (such as `headers` or `scopes`, but technically not `components.schemas` since `$` is not an allowed character)

**Solution**
This PR adds a workaround which keeps track of whether the recursive function is inside a `properties` field


rpmccarter:
sorry! meant to open this against our internal fork. once I brush this up hopefully I can open a similar PR to upstream!

rpmccarter:
[here](https://github.com/scalar/scalar/pull/3753/commits/a124ee894193dcdea4f238faee0e644cb7bc7ddf#diff-953b4a1887207931765c7c8009b87713e2682e853f67af3ea28f80f0d2efd277) is the relevant fix - let me know if you want me to make a PR with it!

PR #3758, opened 2024-11-05
fix(oas-utils): path regex

antlio:
this pr fixes the path regex that is [not supported by safari](https://stackoverflow.com/questions/51568821/works-in-chrome-but-breaks-in-safari-invalid-regular-expression-invalid-group) and breaks the client rendering.

```
SyntaxError: Invalid regular expression: invalid group specifier name
```


hanspagel (approved):
Great find!

antlio:
> Great find!

great test!

PR #3759, opened 2024-11-05
chore(api-reference): remove Storybook and some unused dependencies

hanspagel:
This PR removes Storybook and some other unused dependencies from `@scalar/api-reference`.

One tiny step closer towards a clean dependency tree. :)


amritk (approved):
You would think we would get better bundle size improvements 😞

tmastrom (approved):

hanspagel:
> You would think we would get better bundle size improvements 😞

Just a sign that tree-shaking works 😅

PR #3761, opened 2024-11-05
refactor(api-client): use route names

hanspagel:
I wanted to start the week with a clean up task: Currently, we‘re using route strings everywhere in the client.

This PR suggests to move to route objects, which decouples the path of the route (e.g. `/settings`) from the actual usage (e.g. `{ name: 'settings' }`) and improves the error handling when something goes wrong (e.g. can’t find route name, missing path property …).

Ideally, we’d go a step further and use TypeScript to make the usage type-safe. But that’s for another PR.

I’ve tried to test every code that I touched to make sure it still works, would be great if someone else could click through the whole client, though.

A few related changes:
* `fallbackMissingParams` was never hit, I removed it
* routes are in a separate `routes.ts` now
* I’ve renamed a few of the active workspace helpers (saveWorkspace → saveActiveWorkspace)


antlio:
great great addition

amritk (approved):
I was just going to say, I prefer type safe routes myself but this can be a stop-gap. Lets see what @geoffgscott says.

edit: vue [recommends](https://router.vuejs.org/guide/advanced/typed-routes) [unplugin-router](https://github.com/posva/unplugin-vue-router) for typesafe routes

hanspagel:
> I was just going to say, I prefer type safe routes myself but this can be a stop-gap. Lets see what @geoffgscott says.
> 
> edit: vue [recommends](https://router.vuejs.org/guide/advanced/typed-routes) [unplugin-router](https://github.com/posva/unplugin-vue-router) for typesafe routes

Yess, we definitely need typed routes as mentioned in the PR description. 🤘 Tiny steps towards big goals.

geoffgscott:
This looks good to me. Decoupling the router parameters from the actual path is also the first step to allowing users to pass a router into the integration and allow different pathing approaches.

@amritk Agree on typesafe routes. I would suggest we just write our own `push` function that accepts an enum (object const probably) for the routes and a set of parameters as needed.

geoffgscott (approved):

PR #3762, opened 2024-11-05
feat(api-reference): add error boundaries

hanspagel:
Currently, when an error is thrown, the whole page is empty.

This PR introduces a `ScalarErrorBoundary` component, that catches errors, shows the error message, but everything else is still rendered properly.

## Before

Screenshot 2024-11-05 at 14 54 23

## After

**Base URL**

Screenshot 2024-11-05 at 14 59 56

**Sidebar**

Screenshot 2024-11-05 at 14 59 31

**Example Responses**

Screenshot 2024-11-05 at 15 02 33

**Models**

Screenshot 2024-11-05 at 15 00 47


packages/components/src/components/ScalarErrorBoundary/ScalarErrorBoundary.vue
hwkr, line 1: I wonder if we should hide the error message in a modal or a tooltip or something

hwkr:
Looks good! Might fiddle with some of the styling with Cam if that's okay. Also wouldn't mind @amritk having a look :)

packages/components/src/components/ScalarErrorBoundary/ScalarErrorBoundary.vue
amritk: I prefer showing the error message, much easier for sharing back to us

amritk (approved):
Exactly what I was thinking!

amritk:
Oh also can we move it to components and use it in the client as well!?

hwkr:
Added some tests and a storybook :) Tweaked the component itself a tiny bit too but feel free to put it back.

I also tried out some other visual treatments, not sure if I like them better or not 🤔 

### With a dotted border

Google Chrome-2024-11-06-22-38-30@2x

### Using a `<details>` and including the stack
Google Chrome-2024-11-06-22-39-14@2x
Google Chrome-2024-11-06-22-39-19@2x

hwkr (approved):

hanspagel:
Waiting for @cameronrohani ✌️

cameronrohani:
happy to spruce this up briefly, but I think @marclave had some feedback here

amritk:
Lets get it innnnn

marclave (changes requested):
lets chat about it before merging, we might wanna just do a toast or smth?

hanspagel:
With this PR we ensure that we render as much as possible (which could be 99 % of the reference), instead of just showing a blank page because of an exception in just one component. That’s so much better than what we have and so much than a toast. HMU so I can convince you. :)

amritk:
Agreed with hans, using an error boundary component is a common practice! Used it many times before

cameronrohani:
Any updates here? 😅 

Should I style it or are we wanting to reconsider the approach?

@hanspagel @marclave 👂

hanspagel:
YESS, cool. @cameronrohani do your thing!

In theory no one should see this component, and even in reality it should only be seen by very, very few people, so I’d be OK with doing styling in another PR, but whatever works for you!

marclave:
@hanspagel should we add a cta to go to our github to open an issue?

hanspagel:
If we bork the CDN, it’s the users of our users that see it, though. Maybe we don’t want to ask _them_ to report a bug, but just trust our users to find their way in that case.

marclave (approved):

hwkr:
What happened to the tests 😭

hwkr:
Oh no did I not push them 🤦

PR #3763, opened 2024-11-05
fix(api-client): select input default

### LINKED ISSUE #3734: wrong parameter values
### What happens?

Test case:
1. Chose endpoint "Get user list​" and open page "Test Request"

2. Chose endpoint "Get user info​​" and open page "Test Request".
Error! Param "locale_id" has invalid value "User"

3. Check possible values for param "locale_id"

I expected to see valid values for param "locale_id"

### OpenAPI Document

https://sandbox.scalar.com/e/KK4AY

antlio:
this pr fixes #3734 as we were relying on setting the selected value on model value, hence the wrong value was added. This pr also adds the support of the default if present.

**before / after**
image
image

### DIFF
--- a/.changeset/tricky-rings-eat.md
+++ b/.changeset/tricky-rings-eat.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
+
+fix: adds default value for params select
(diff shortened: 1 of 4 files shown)


amritk (approved):
LGTM

PR #3764, opened 2024-11-05
fix(openapi-parser): when migrating plain `example` to example objects there’s no `value` attribute

### LINKED ISSUE #3760: Parsing example values for parameters broke with v1.25.53
### What happens?

I have a query parameter defined in one of my endpoints like this:
```yaml
- name: currency-id
  in: query
  required: true
  schema:
    type: string
    format: uuid
  example: a26cdf3a-6e70-9124-385d-4d7eb9bff6b8
  x-go-name: CurrencyIDQueryParam
  x-oapi-codegen-extra-tags:
    validate: "required,uuid"
```
If I try to pass this (alongside with the whole openapi definition of course) to scalar, I'm getting an error upon loading the HTML reference page in the console:
```
standalone.js:3129 {
  "_errors": [],
  "parameters": {
    "0": {
      "_errors": [],
      "examples": {
        "_errors": [],
        "default": {
          "_errors": [
            "Expected object, received string"
          ]
        }
      }
    },
    "1": {
      "_errors": [],
      "examples": {
        "_errors": [],
        "default": {
          "_errors": [
            "Expected object, received string"
          ]
        }
      }
    },
    "2": {
      "_errors": [],
      "examples": {
        "_errors": [],
        "default": {
          "_errors": [
            "Expected object, received number"
          ]
        }
      }
    },
    "_errors": []
  }
}
```
It also logs out the problematic object:
```json
{
    ...
    "parameters": [
        ...
        {
            "name": "currency-id",
            "in": "query",
            "required": true,
            "schema": {
                "type": "string",
                "format": "uuid"
            },
            "x-go-name": "CurrencyIDQueryParam",
            "x-oapi-codegen-extra-tags": {
                "validate": "required,uuid"
            },
            "examples": {
                "default": "a26cdf3a-6e70-9124-385d-4d7eb9bff6b8"
            }
        },
        ...
    ],
    ...
}
```
As you can see, if I use the singular `example` tag to define an example for my parameter, it transforms into the `examples` one, but the content of it is wrong.

Either the transformed value should be:
```json
"examples": {
    "default":  {
        "value": "a26cdf3a-6e70-9124-385d-4d7eb9bff6b8"
    }
 }
```
or there shouldn't be any transformation at all for singular examples.

### OpenAPI Document

_No response_

hanspagel:
When upgrading an (old) example structure to the (new) example objects, for some reason I didn’t put the example in the `value` attribute. I just reread the specification and fixed this.

@amritk Does this break the API client schema? Would you mind to look into this? 

**Before**
```js
{
  examples: {
    default: 'foobar'
  }
}
```

**After**
```js
{
  examples: {
    default: {
      value: 'foobar'
    }
  }
}
```

OpenAPI 3.1 Example Object: https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#example-object

Fixes #3760.

### DIFF
--- a/.changeset/ten-cobras-run.md
+++ b/.changeset/ten-cobras-run.md
@@ -1,0 +1,5 @@
+---
+'@scalar/openapi-parser': patch
+---
+
+fix: when migrating example to example objects, the example should be inside the value attribute
--- a/packages/openapi-parser/src/utils/upgradeFromThreeToThreeOne.test.ts
+++ b/packages/openapi-parser/src/utils/upgradeFromThreeToThreeOne.test.ts
@@ -172,8 +172,8 @@ describe('upgradeFromThreeToThreeOne', () => {
     })
   })
 
-  describe('use examples not example in schemas', () => {
-    it('migrates example to examples', async () => {
+  describe('migrates example to examples', () => {
+    it('uses arrays in schemas', async () => {
       const result = upgradeFromThreeToThreeOne({
         openapi: '3.0.0',
         info: {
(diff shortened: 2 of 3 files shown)


amritk:
I think galaxy uses value so it should work

marclave (approved):

hanspagel:
Yess, cool!

PR #3771, opened 2024-11-05
chore(api-client-app): update `@todesktop/cli`

hanspagel:
This updates `@todesktop/cli`.

> Your version of @todesktop/cli is out of date.
> Run npm install --save-dev @todesktop/cli to update to v1.10.2.


amritk (approved):
Does it not run if its outta date?

hanspagel:
It doesn’t seem to be the root cause of why it’s failing in CI, but still good to stay up to date with the CLI (that interacts with the toDesktop API). 👍

PR #3778, opened 2024-11-06
chore: docs Django Ninja README.md

ivanagas:
**Problem**
Currently, the instructions don't give you all the details you need to install the Django Ninja API reference.

**Solution**
With this PR, you get both the details on installing and accessing the API reference.


marclave (approved):
nice @ivanagas 😎

PR #3783, opened 2024-11-06
feat(api-client): clear drafts

antlio:
this pr adds the clear drafts capability from request sidebar draft collection, as seen below:

https://github.com/user-attachments/assets/d9b48bc6-2cfb-477b-95c6-c9ca17f2f9eb

we might want to set a more central empty state in the view afterward.

### DIFF
--- a/.changeset/nasty-rabbits-greet.md
+++ b/.changeset/nasty-rabbits-greet.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
+
+feat: adds clear drafts action to draft collection
--- a/packages/api-client/src/store/requests.ts
+++ b/packages/api-client/src/store/requests.ts
@@ -188,3 +188,16 @@ export function findRequestParentsFactory({
 
   return findRequestParentss
 }
+
+/** First draft request" */
+export function createInitialRequest() {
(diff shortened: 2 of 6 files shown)


antlio:
> This is so great! But didn’t we have to have at least one request for some reason? 🤔

We do on the first usage of the client, no strong opinion about if we should always maintain one or not

hanspagel (approved):
I’d say it’s cool, but I’ll leave this up to @marclave and @amritk.

amritk (changes requested):
I think just create the initial my first request draft and route to it so you don't need an empty state. Either that or an empty state. Seems strange to have it blank

marclave:
> I think just create the initial my first request draft and route to it so you don't need an empty state. Either that or an empty state. Seems strange to have it blank

yes : )

antlio:
> I think just create the initial my first request draft and route to it so you don't need an empty state. Either that or an empty state. Seems strange to have it blank

I added a fallback in order to always have a first draft requests. drafts requests can be cleared in other requests are existing in the workspace, however in case of all requests being deleted, a first request is added back to the draft collection.

amritk (approved):

PR #3784, opened 2024-11-06
fix(api-reference): docker image can’t be scanned

hanspagel:
The release of the docker image failed because the security scanning failed. This PR fixes it.

* We’re now scanning specifically the Dockerfile.
* The suggestion to run caddy as a non-root user came up, so I’ve added this, too.
* All the Docker steps are only executed when something in the folder changed (to avoid blocking main with the security check at random times).


tmastrom (approved):
Nice fix. Also adding the non root user to run caddy is a good practice

PR #3785, opened 2024-11-06
fix(api-client-app): app build is missing

hanspagel:
Ready for a fun story?

* We recently fixed the turbo cache in CI.
* App builds now restored from the cache.
* But the `turbo.json` only defined `dist/**` as an output folder to restore from the cache.
* The Electron app puts the build in `out/**`, though (`electron-vite` default).
* So fixing the cache actually made the build disappear.
* This broke the todesktop build and release.

With this PR we’re telling `electron-vite` to put the build inside `dist/**` as every sane tool would do. That way, it’s restored (like all other builds) and the todesktop build/release should work again. :)


tmastrom (approved):
🙌

PR #3789, opened 2024-11-06
chore(api-reference): skip the docker build if docker isn’t available

hanspagel:
I’ve added the docker build as a `build` npm script, which is executed on `pnpm turbo build`, but fails in environments without Docker.

Let’s only build the Docker image, when Docker is actually available.


tmastrom (approved):
Great this should fix the CDN references e2e testing

amritk (approved):
Lets get it in

PR #3790, opened 2024-11-06
chore(scalar-app): rename app folder

hanspagel:
We’re talking about the _Scalar App_ most of the time, so I think it’s time to streamline the naming here:

* Rename `packages/api-client-app` to `packages/scalar-app`
* Rename the package from `scalar-api-client` to `scalar-app`.


amritk:
We can rename the app, but I'm not sure I agree with the package. It should still be client while we have references and client so they can be easily distinguished

amritk:
oh whats scalar-api-client lol

amritk (approved):

tmastrom (commented):
An argument could be made for `scalar-desktop` too but I think this works

PR #3793, opened 2024-11-06
fix(api-reference): show servers even when there's no info.title or info.description

amritk:
https://discord.com/channels/1135330207960678410/1303739932434632834

Servers are in the intro block so they don't show if there's no info.title or info.description

before:
image

So I just moved them outside as well, not the most elegant solution due to requiring all of the nesting for padding etc but good enough for now

after:
image


packages/api-reference/src/components/Content/Content.vue
```diff
@@ -66,9 +73,40 @@ const introCardsSlot = computed(() =>
...
+                  :defaultServerUrl="baseServerURL"
+                  :servers="props.servers"
+                  :specification="parsedSpec" />
+                <Authentication
+                  class="introduction-card-item"
+                  :parsedSpec="parsedSpec"
+                  :proxy="proxy" />
+                <ClientLibraries class="introduction-card-item" />
+              </div>
+            </SectionColumn>
```
hwkr, line 97: I think if you just add an empty section column after this then it will restrict it to the left hand column and match the endpoint descriptions.

amritk: but then we got an empty void on the left. Which is better?

hwkr: Empty void on the right is what I'm suggestion, keep the block left just reduce width

amritk: ooo yes brilliant will do 🫡

amritk:
Fixed! Updated screenshot
image

hwkr (approved):
Lgtm!

amritk:
ACtually going to just default the title instead which should get this working

PR #3798, opened 2024-11-07
chore(openapi-parser): update test specification location

tmastrom:
Test specifications have been moved. Updates the location to pull test specifications from


hanspagel (changes requested):
Can we call that openapi or openapi-parser instead of oas? 

I try to avoid the OAS abbr. With all the various terms (Swagger, OpenAPI, OAI, OAS) abbreviations add more confusion than they help I think.

I know, it’s just the tests, and I probably shouldn’t care, but how we talk about stuff internally heavily influences how we talk to users.

PR #3799, opened 2024-11-07
fix(api-reference): base url variables DOC-2476

### LINKED ISSUE #3697: variables in the server URL don’t use the default values
This (@scalar/galaxy) breaks the code examples:

image
image

antlio:
this pr fixes #3697 by setting default server variable and enhancing empty variable value as seen below:

**⊢ base url before / after**
image
image

**⊢ request example before / after**
image
image

### DIFF
--- a/packages/api-reference/src/features/BaseUrl/ServerVariablesSelect.vue
+++ b/packages/api-reference/src/features/BaseUrl/ServerVariablesSelect.vue
@@ -36,9 +36,13 @@ const selected = computed<ScalarListboxOption | undefined>({
       class="variable-select"
       fullWidth
       variant="ghost">
-      <span>
-        <span class="sr-only">Selected:</span>
-        {{ value }}
+      <span :class="{ 'text-c-1': value }">
+        <span
+          v-if="value"
+          class="sr-only"
+          >Selected:</span
+        >
+        {{ value || 'Select value' }}
       </span>
       <ScalarIcon
         icon="ChevronDown"
--- a/packages/api-reference/src/helpers/getExampleCode.ts
+++ b/packages/api-reference/src/helpers/getExampleCode.ts
@@ -32,6 +32,11 @@ export async function getExampleCode(
     )
   }
 
+  // Prevent snippet generation if starting by a variable
+  if (request.url.startsWith('__')) {
+    return request.url
+  }
+
   // httpsnippet-lite
   try {
     const httpSnippetLiteTargetKey = target?.replace(
--- a/packages/api-reference/src/legacy/helpers/getUrlFromServerState.ts
+++ b/packages/api-reference/src/legacy/helpers/getUrlFromServerState.ts
@@ -18,5 +18,13 @@ export function getUrlFromServerState(state: ServerState) {
       ? replaceVariables(server?.url, state.variables)
       : server?.url
 
-  return url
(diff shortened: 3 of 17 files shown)


packages/api-reference/src/legacy/helpers/getUrlFromServerState.ts
```diff
@@ -18,5 +18,18 @@
       ? replaceVariables(server?.url, state.variables)
       : server?.url
 
-  return url
+  // {id} -> __ID__
+  const urlVariables = url?.match(/{(.*?)}/g)
```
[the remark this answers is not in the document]
antlio: should we also update the one from getRequestFromOperation.ts? could refactor and use the findVariables helper too

packages/api-reference/src/features/BaseUrl/ServerVariablesSelect.vue
```diff
@@ -36,9 +36,9 @@ const selected = computed<ScalarListboxOption | undefined>({
       class="variable-select"
       fullWidth
       variant="ghost">
-      <span>
+      <span :class="{ 'text-c-1': value }">
         <span class="sr-only">Selected:</span>
```
hwkr, line 40: We can probably hide this screen reader text if `value` isn't set, otherwise it will read "Selected: Select Value" (and it can probably just say "Select Value")

hwkr (commented):
Looks good! Just left one a11y comment :)

packages/api-reference/src/features/BaseUrl/ServerVariablesSelect.vue
antlio: oh great catch!

packages/api-reference/src/helpers/getExampleCode.ts
```diff
@@ -32,6 +32,11 @@ export async function getExampleCode(
     )
   }
 
+  // Prevent snippet generation if starting by a variable
```
amritk, line 35: oh is this because the variables are not replaced?

amritk (approved):
Ty for doing this one!

packages/api-reference/src/legacy/helpers/getUrlFromServerState.ts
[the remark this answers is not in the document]
hanspagel: Yes!

packages/api-reference/src/helpers/getExampleCode.ts
hanspagel: yeah, it throws an error when an url doesn’t have a valid protocol

hanspagel (changes requested):
didn’t review, but spotted this on the screenshot. the select should have the default value selected (https).

Screenshot 2024-11-07 at 20 42 37

hanspagel (changes requested):
Damn, I don’t know why but for some reason it doesn’t change the base URL in the modal anymore.

Screenshot 2024-11-11 at 15 06 57
Screenshot 2024-11-11 at 15 06 59

antlio:
> Damn, I don’t know why but for some reason it doesn’t change the base URL in the modal anymore.
> 
 great catch, it was a close call 😄 now went with returning both origin and modified url so we can display the right one in client on change, while keep displaying the updated one in the reference

PR #3801, opened 2024-11-07
refactor(api-reference): move ESM example to a package playground

hanspagel:
This PR moves the `@scalar/api-reference` ESM example to the package folder (as another playground).

I’ve also found out that the package is working with the esm.sh CDN (not with esm.run yet, though). That wasn’t the case a while ago. To optimize this use case, I’ve added a proprietary esm.sh config to the package.json, which makes them bundle all dependencies.

Otherwise, without this setting, it’s (for some reason) loading tons of various versions of packages (~8 mb, screenshot below). The ESM CDN is still slightly larger for some reason, but it’s very close now:

**Example**

```html
<!doctype html>
<html lang="en">
  <head>
    <meta charset="UTF-8" />
    <meta
      name="viewport"
      content="width=device-width, initial-scale=1.0" />
    <title>Scalar API Reference (ESM)</title>
  </head>
  <body>
    <div id="app"></div>

    <script type="module">
      import { createScalarReferences } from 'https://esm.sh/@scalar/api-reference';

      createScalarReferences(document.getElementById('app'), {
        spec: {
        },
      })
    </script>
  </body>
</html>
```

**With `?bundle-deps` or `bundle: true` (https://esm.sh/#docs)**

| CDN | Size |
|  -- | -- |
| https://cdn.jsdelivr.net/npm/@scalar/api-reference | 665 kB |
| https://esm.sh/v135/@scalar/api-reference@1.25.58/es2022/api-reference.bundle.mjs | 831 kB |

**Without `?bundle-deps` or `bundle: true` (https://esm.sh/#docs)**

Screenshot 2024-11-07 at 14 32 02


amritk (approved):
🔥

marclave:
nice 😎

PR #3805, opened 2024-11-07
feat(api-client): change sidenav to header only for web api client view

cameronrohani:
Making navigation consistent on web, but keep app sidenav

web view:
image

app view:
image


packages/api-client/src/libs/create-client.ts
```diff
@@ -13,6 +13,11 @@ import type { LiteralUnion } from 'type-fest'
 import { type Component, createApp, watch } from 'vue'
 import type { Router } from 'vue-router'
 
+/**
+ * The layout of the client
+ */
+export type ClientLayout = 'modal' | 'web' | 'desktop'
```
hwkr, line 19: Not sure if the type should live here or with the hook but it's typed now!

hwkr:
This PR should be good to go! Wouldn't hurt to test the web and desktop versions to make sure I didn't break anything. 

Also now we have a `useLayout` hook which tell us what layout we're using anywhere in the client which we might want to put in some of the places we check for readonly mode?

## Video ✨ 

https://github.com/user-attachments/assets/9a328cf7-e5a6-4f98-8dd1-fa92803b9ce4

packages/api-client/src/libs/create-client.ts
hwkr: @amritk @hanspagel

packages/api-client/src/components/SideNav/SideNavLink.vue
```diff
@@ -1,36 +1,34 @@
...
-    <div
-      class="min-w-[37px] max-w-[37px] group-hover:bg-b-2 active:text-c-1 flex items-center justify-center rounded-lg p-[8px] scalar-app-nav-padding text-c-3"
-      :class="{
-        'bg-b-2 transition-none group-hover:cursor-auto !text-c-1': active,
-      }">
+  <component
+    :is="is ?? 'a'"
+    class="hover:bg-b-3 no-underline min-w-[37px] max-w-[37px] flex items-center justify-center rounded-lg p-2"
+    :class="{
+      'bg-b-3 transition-none hover:cursor-auto text-c-1': active,
```
antlio, line 19: should we keep it `bg-b-2` on darkmode?

packages/api-client/src/components/SideNav/DownloadAppButton.vue
```diff
@@ -0,0 +1,44 @@
+<script setup lang="ts">
+import SideNavLink from '@/components/SideNav/SideNavLink.vue'
+import { useLayout } from '@/hooks'
+import { ScalarIcon } from '@scalar/components'
+
+const { layout } = useLayout()
+</script>
+<template>
+  <SideNavLink
+    class="download-app-button sm:px-3"
```
antlio, line 10: i think we should either decrease the padding in sections or increase it here to maintain alignment

packages/api-client/src/components/SideNav/SideNavLink.vue
amritk: never a big fan of these random px values, _usually_ means something else is wrong! If we stick to the grid then we never need px!

packages/api-client/src/libs/create-client.ts
amritk: Love the hook but should we change desktop? Its not really clear as web can be desktop too. How about app or even electron? Type should live with the hook imo, always keep it close to the meat

amritk (approved):
Love it!

packages/api-client/src/components/SideNav/SideNavLink.vue
cameronrohani: @antlio most likely needs some tweaking on dark mode, me or you can fast follow on monday or I'll update this weekend :)

packages/api-client/src/components/SideNav/SideNav.vue
```diff
@@ -1,82 +1,53 @@
...
-/** Make the sidebar draggable */
-.drag-region {
-  -webkit-app-region: drag;
-}
-
-/** Clickable items must not be draggable, though */
-.no-drag-region {
-  -webkit-app-region: no-drag;
-}
-
```
cameronrohani, line 68: we need this back, then it should be good to go :)

packages/api-client/src/libs/create-client.ts
hwkr: In my mind desktop !== web whereas app can be a web, desktop or mobile app. We use toDesktop to which seems to call electron apps desktop apps right? 

Google Chrome-2024-11-08-18-40-22@2x

hwkr: I can move the types to the hook though 🫡

packages/api-client/src/components/SideNav/DownloadAppButton.vue
hwkr: In sections?

marclave (approved):

PR #3817, opened 2024-11-08
ci: enable stackblitz pre-release again

hanspagel:
We disabled the Stackblitz pre-release publish job a while ago, but I think it’s time to add it again.

The pre-release allows us to test packages like they’d be published on npm, before actually publishing to npm.


tmastrom (approved):
Good idea!

amritk (approved):
Sure thing

PR #3818, opened 2024-11-08
feat(proxy-server): follow redirects

hanspagel:
Currently, when the proxy gets a `Location` header (redirect), it just proxies it:

`https://proxy.scalar.com/?scalar_url=https://example.com/foo` responds with `Location: /bar`

And the browser will follow the redirect, requesting an URL that doesn’t exist:

`https://proxy.scalar.com/bar`

With this PR, the proxy just follows the redirect:

`https://proxy.scalar.com/?scalar_url=https://example.com/foo` responds with the _content_ of `https://example.com/bar`

But in `@scalar/import` we try to transform relative URLs to absolute URLs, so we need to know the resulting URL. To achieve this, the proxy now adds a [`X-Forwarded-Host` header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Host) and `@scalar/import` uses this to prefix relative URLs (if it’s available).

This PR also adds a test for parent paths (`https://example.com/foo/bar/` + `../openapi.json` = `https://example.com/foo/openapi.json`) and merges the `makeRelativeUrlsAbsolute` and the `makeUrlAbsolute()` helper.

### DIFF
--- a/.changeset/green-nails-sparkle.md
+++ b/.changeset/green-nails-sparkle.md
@@ -1,0 +1,5 @@
+---
+'@scalar/import': minor
+---
+
+feat: use X-Forwarded-Host header (if available)
--- a/.changeset/lemon-tomatoes-wash.md
+++ b/.changeset/lemon-tomatoes-wash.md
@@ -1,0 +1,5 @@
+---
+'@scalar/oas-utils': patch
+---
+
+feat: allow to pass a base URL to makeUrlAbsolute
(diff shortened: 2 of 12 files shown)


amritk (approved):
Sweet, nice work! I'm no gopher though so maybe @marclave can review as well

amritk (approved):
Eye ball test passed.

PR #3819, opened 2024-11-08
feat(oas-utils): add object-id value example

danjohnson95:
**Problem**
We use `format: object-id` to specify that the value should be a MongoDB Object ID. Currently, an example value isn't automatically generated for these fields.

**Solution**
With this PR, any field with `format: object-id` will now have a default example value, just like how it works for `format: uuid`.


hanspagel (approved):
Thanks, that’s great!

PR #3826, opened 2024-11-08
feat(api-client): display which auth is required on a request

### LINKED ISSUE #3613: Display which auth is required on which request
### What happens?

We currently allow users to select any auth on any request. However according to the spec, each request can specify its own security requirements. Instead of restricting the user to just those auth methods, we should show something in the UI that states which auth is required on which request without limiting the user

amritk:
closes #3613

image

@cameronrohani will replace the icon later

### DIFF
--- a/.changeset/odd-rivers-approve.md
+++ b/.changeset/odd-rivers-approve.md
@@ -1,0 +1,6 @@
+---
+'@scalar/api-client': patch
+'@scalar/components': patch
+---
+
+feat: added auth indicator for required auth on requests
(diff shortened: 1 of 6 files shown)


amritk:
sorry @hwkr I just augmented ComboBox again instead of pulling it out 😳

antlio:
great one! how would we feel about something like the following? always felt that it could be enhanced since i don't see a use case with twice the same auth would there be?
image

cc @cameronrohani

cameronrohani:
> great one! how would we feel about something like the following? always felt that it could be enhanced since i don't see a use case with twice the same auth would there be? image
> 
> cc @cameronrohani

I like it @amritk, can we do this?

amritk:
oh btw there will be 3 categories now: required auth, available auth, add new auth

amritk:
Closing with a new one coming soon!

PR #3828, opened 2024-11-08
feat(api-client): display which auth is required on a request

### LINKED ISSUE #3613: Display which auth is required on which request

amritk:
closes #3613

Based on feedback from [prev PR](https://github.com/scalar/scalar/pull/3826), added all options like this instead. Copy is up for debate as usual

Preview:
image

### DIFF
--- a/.changeset/metal-ants-travel.md
+++ b/.changeset/metal-ants-travel.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
+
+feat: indicate which auth is required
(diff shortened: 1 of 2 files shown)


marclave (approved):
this is good to go!

@cameronrohani  @antlio @hanspagel we probs wanna think about empty state + also showing _something_ in the references per endpoint also : )

PR #3829, opened 2024-11-08
fix: add missing proxy on import modal

marclave:
import modal didnt work if we needed cors : )


amritk (approved):
🧠

cameronrohani (approved):
no way, he actually did it, he actually fixed it sub 15 mins

PR #3833, opened 2024-11-09
fix(Scalar.AspNetCore): use relative path for assets

### LINKED ISSUE #3810: scalar.js not found asp net core
### What happens?

Scalar returns 404 "not found" error if the API is located in a virtual application in IIS and not on the main site. Version 1.2.31. Does anyone have any suggestions on how to fix this? Thanks

The same bevahior like api is installed on main site

### OpenAPI Document

_No response_

xC0dex:
I made a modification to the `Scalar.AspNetCore` package, so the asset is now included like this:

```html
<script src="scalar.js"></script>
```

instead of:
```html
<script src="/scalar/scalar.js"></script>
```

This adjustment allows the script to load correctly regardless of the app base path, making it more flexible in different hosting environments.

Or is this too risky? I'm uncertain.

Closes #3810


hanspagel:
Makes sense!

marclave (approved):
this makes sense to me :) 

nice work @xC0dex !!

PR #3848, opened 2024-11-11
feat(proxy-server): follow redirects DOC-2535

hanspagel:
A previous version of this PR has been merged and rolled back. This PR has more changes than the initial PR:

* Even more tests
* Refactor of the proxy, hopefully easier to follow the code
* It doesn't merge CORS headers anymore, but overwrites them (the idea of the proxy is to be easy with cross-origin requests)
* Manually tested with a bunch of URLs, too (directly, in the reference and in the API client)

Here’s the original description:

Currently, when the proxy gets a `Location` header (redirect), it just proxies it:

`https://proxy.scalar.com/?scalar_url=https://example.com/foo` responds with `Location: /bar`

And the browser will follow the redirect, requesting an URL that doesn’t exist:

`https://proxy.scalar.com/bar`

With this PR, the proxy just follows the redirect:

`https://proxy.scalar.com/?scalar_url=https://example.com/foo` responds with the _content_ of `https://example.com/bar`

But in `@scalar/import` we try to transform relative URLs to absolute URLs, so we need to know the resulting URL. To achieve this, the proxy now adds a [`X-Forwarded-Host` header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Host) and `@scalar/import` uses this to prefix relative URLs (if it’s available).

This PR also adds a test for parent paths (`https://example.com/foo/bar/` + `../openapi.json` = `https://example.com/foo/openapi.json`) and merges the `makeRelativeUrlsAbsolute` and the `makeUrlAbsolute()` helper.

The proxy renders an API reference on the root path now. This is great to quickly test this PR for example:

1. `cd examples/proxy-server`
2. `pnpm run dev`
3. `open http://localhost:5051`

Screenshot 2024-11-20 at 11 35 49

### DIFF
--- a/examples/proxy-server/main.go
+++ b/examples/proxy-server/main.go
@@ -37,45 +37,24 @@ func corsMiddleware(next http.Handler) http.Handler {
 	})
 }
 
-// Handler that forwards requests to the target and writes the response back to the original client
-func handleRequest(w http.ResponseWriter, r *http.Request) {
-	// Parse query parameters
-	queryValues := r.URL.Query()
-	target := queryValues.Get("scalar_url")
+// ProxyServer encapsulates the proxy server configuration and handlers
+type ProxyServer struct {
+	transport *http.Transport
+}
 
-	remote, _ := url.Parse(target)
-
-	// If the requested path is /ping, return a simple response.
-	if r.URL.Path == "/ping" {
-		log.Println("/ping")
-
-		w.Write([]byte("pong"))
-		return
+// NewProxyServer creates a new proxy server instance
+func NewProxyServer() *ProxyServer {
+	return &ProxyServer{
+		transport: &http.Transport{
+			TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
+		},
 	}
+}
 
-	// The URL parameter is required. Return a helpful error message if it’s not provided.
-	if target == "" {
-		http.Error(
-			w,
-			"The `scalar_url` query parameter is required. Try to add `?scalar_url=https%3A%2F%2Fgalaxy.scalar.com%2Fplanets` to the URL.", http.StatusBadRequest,
-		)
-
-		return
-	}
-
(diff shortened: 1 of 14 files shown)


packages/import/src/resolve.ts
```diff
@@ -61,6 +62,9 @@ export async function resolve(
 
       if (result.ok) {
         const content = await result.text()
+        const forwardedHost = result.headers.get('X-Forwarded-Host')
+
+        console.log('ASAS', forwardedHost)
```
amritk, line 67: we wanna keep these comments?

amritk (commented):
I will leave the go reviews for gopher @marclave

packages/import/src/resolve.ts
hanspagel: Yes, those are really important. 🤡

amritk: i meant logs lol

packages/import/src/resolve.ts
```diff
@@ -110,7 +112,8 @@ export async function resolve(
 
         // Relative or absolute URL
         if (urlOrPathOrDocument) {
-          return makeRelativeUrlsAbsolute(value, urlOrPathOrDocument)
+          console.log('make absolute', forwardedHost, urlOrPathOrDocument)
```
hanspagel, line 115:
```suggestion
```

marclave (approved):
finally finished the review and testing!

examples/proxy-server/main.go
geoffgscott, line 1: This file needs WAY more comments. Every header change should have a descriptive comment as to why.

hanspagel: I’ll add comments and an explanation in the README.

hanspagel:
Merging now. I’m in front of my computer for the next hours to fix any upcoming issues.

PR #3854, opened 2024-11-11
fix(api-client): import modal shows up for any URL

hanspagel:
Currently, the import collection modal is shown when any URL is pasted or dropped into the UI.

With this PR we’re checking whether the pasted/dropped URL is pointing to an OpenAPI document first, and only then show the modal.

It still shows the import modal when an URL is provided through a query parameter (when following a link from the references).

### DIFF
--- a/packages/api-client/src/components/ImportCollection/ImportCollectionModal.vue
+++ b/packages/api-client/src/components/ImportCollection/ImportCollectionModal.vue
@@ -18,6 +18,7 @@ import { computed, onMounted, onUnmounted, ref, watch } from 'vue'
 const props = defineProps<{
   source: string | null
   integration: string | null
+  eventType: 'drop' | 'paste' | 'query' | null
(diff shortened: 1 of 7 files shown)


packages/api-client/src/components/ImportCollection/ImportCollectionModal.vue
```diff
@@ -57,15 +58,40 @@ const version = computed(() =>
...
+        return
+      }
+
+      // Query parameters:
       prefetchUrl(value, activeWorkspace.value.proxyUrl)
+
+      modalState.show()
+
+      return
     }
```
hanspagel, line 91: This is the relevant change

antlio: works like a charm!

marclave (approved):

PR #3855, opened 2024-11-11
fix(void-server): does not respond with XML header

hanspagel:
Currently, `@scalar/void-server` replies with XML, but doesn’t add the XML content-type header.

This PR adds it.


marclave (approved):
all my homies LOVE xml

nice work :)

PR #3856, opened 2024-11-11
fix(mock-server): OpenAPI YAML endpoints don’t have a YAML content-type header

hanspagel:
Currently, `@scalar/mock-server` doesn’t add YAML http headers when serving the OpenAPI documents as YAML.

This PR adds them.


marclave (approved):
:)

PR #3862, opened 2024-11-12
test(oas-utils): more tests for `importSpecToWorkspace`

hanspagel:
This PR improve a few code comments and adds a bunch of tests for `importSpecToWorkspace`.

It’s also adding a console.log for the time it takes to create the workspace, and removes the console.log for the time it takes to create the search index. The search index was never really blocking anything, and it was rather quick, even for big files. The workspace becomes more relevant for the time to interaction, so I’d like to keep an eye on it for real-world examples. 

The diff is hard to follow, but no actual code is touched (except the mentioned logging).

### DIFF
--- a/packages/api-client/src/store/import-spec.ts
+++ b/packages/api-client/src/store/import-spec.ts
@@ -6,7 +6,7 @@ import {
   importSpecToWorkspace,
 } from '@scalar/oas-utils/transforms'
 import type { OpenAPIV3, OpenAPIV3_1 } from '@scalar/openapi-types'
-import type { ReferenceConfiguration, Spec } from '@scalar/types/legacy'
+import type { ReferenceConfiguration } from '@scalar/types/legacy'
 import { toRaw } from 'vue'
 
 /** Maps the specs by URL */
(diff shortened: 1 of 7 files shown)


packages/api-client/src/store/import-spec.ts
```diff
@@ -6,7 +6,7 @@ import {
   importSpecToWorkspace,
 } from '@scalar/oas-utils/transforms'
 import type { OpenAPIV3, OpenAPIV3_1 } from '@scalar/openapi-types'
-import type { ReferenceConfiguration, Spec } from '@scalar/types/legacy'
```
hanspagel, line 9: unused import

packages/oas-utils/src/transforms/import-spec.test.ts
```diff
@@ -121,9 +185,111 @@ describe('Import OAS Specs', () => {
       const authScheme = res.securitySchemes[5] as SecuritySchemeOauth2
       expect(authScheme['x-scalar-client-id']).toEqual(testId)
     })
+
+    it('handles empty security requirements', async () => {
+      const specWithEmptySecurity = {
+        ...galaxy,
+        security: [{}],
```
amritk, line 192: this is not an empty requirement btw, it is an optional one 😉 `[]` would be empty

amritk: not that it matters

amritk (approved):
More tests are always welcome!

marclave (approved):

PR #3863, opened 2024-11-12
chore(examples): generate cloudbuild file from script

tmastrom:
Currently, the `cloudbuild.yaml` file is getting really long as we add more examples and integrations. 

This PR creates a script to generate the `cloudbuild.json` file from a smaller environment file. I am switching to `json` format because this is much easier to read and write with nodejs. The formats are interchangeable for cloud build but we will need to update the file path to use this. 

Screenshot 2024-11-12 at 4 29 24 PM

This PR also adds the script to the pre-commit hook to ensure the cloudbuild file is up to date


package.json
```diff
@@ -78,6 +78,7 @@
     "lint:fix": "pnpm -r lint:fix",
     "pack:packages": "./scripts/pack.sh",
     "playwright:install": "npx playwright install --with-deps",
+    "scripts:generate-cloudbuild": "bun ./tooling/cloudbuild/generateCloudbuild.ts",
```
tmastrom, line 81: @geoffgscott I know you don't like these scripts but I think this warrants a new one

geoffgscott: If we want to use bun prefer the workspace level version `pnpm bun run ...`
Or we can use the workspace vite-node

geoffgscott (commented):
We should leave this as a yaml file. Avoids GCP changes and is generally a bit more readable. 

You can install `yaml` as workspace dev dependency.

package.json
hanspagel: Yeah, we use vite-node everywhere else, can we use that here, too?

hanspagel:
I am a simple man. I see tasks, I convert to draft.

tmastrom:
> We should leave this as a yaml file. Avoids GCP changes and is generally a bit more readable.
> 
> You can install `yaml` as workspace dev dependency.

I looked at keeping the yaml file format but the [GCP cloudbuild file format](https://cloud.google.com/build/docs/build-config-file-schema#yaml) is not strictly yaml and expect certain fields to be arrays of strings which the `yaml` parser turns into a list with - points.

geoffgscott (approved):

geoffgscott:
Just need to remove the pnpm-lock.yaml file from this PR as there was no package changes.

PR #3864, opened 2024-11-12
feat(api-client): environment collection variables

antlio:
this pr re-organize the environment sidebar to include collections along collection environment.

image

⊢ course of action:
- [x] Adds collection environment editor
- [x] Setup collection environment routes
- [x] Redirection on environment removal
- [x] Sets global environment from workspace
- [x] Substitutes environment variable value

### DIFF
--- a/packages/api-client/src/components/EnvironmentSelector/EnvironmentSelector.vue
+++ b/packages/api-client/src/components/EnvironmentSelector/EnvironmentSelector.vue
@@ -13,18 +13,20 @@ import type { Collection } from '@scalar/oas-utils/entities/spec'
 import { computed, onMounted, watch } from 'vue'
 import { useRouter } from 'vue-router'
 
-const {
-  activeCollection,
-  activeWorkspace,
-  activeEnvironment,
-  setActiveEnvironment,
-} = useActiveEntities()
-const { isReadOnly } = useWorkspace()
+const { activeCollection, activeWorkspace, activeEnvironment } =
+  useActiveEntities()
+const { isReadOnly, collectionMutators } = useWorkspace()
 
 const router = useRouter()
 
 const updateSelected = (uid: string) => {
-  setActiveEnvironment(uid)
+  if (activeCollection.value) {
+    collectionMutators.edit(
+      activeCollection.value.uid,
+      'x-scalar-active-environment',
+      uid,
+    )
+  }
 }
 const createNewEnvironment = () =>
   router.push({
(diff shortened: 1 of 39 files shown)


hwkr:
It looks like this branch has a bunch of old commits from main for some reason?

It probably needs to be rebased and have some commits dropped

packages/api-client/src/components/Sidebar/SidebarListElement.vue
```diff
@@ -27,11 +28,18 @@ const emit = defineEmits<{
...
+  event: MouseEvent,
+  uid: string,
+  collectionId?: string,
+) => {
+  const path = collectionId
+    ? `/workspace/default/environment/${collectionId}/${uid}`
+    : `/workspace/default/environment/${uid}`
   if (event.metaKey) {
-    window.open(uid, '_blank')
+    window.open(path, '_blank')
```
hwkr, line 40: For the future but should be make this open a tab on the desktop client?

Maybe we should handle window.open for the desktop client?

hwkr (approved):
Looks good to me!

packages/api-client/src/views/Environment/Environment.vue
```diff
@@ -92,77 +227,95 @@ function synchronizeKeyRemoval(removedKey: string) {
...
-
-/** set active environment based on the route */
-const setActiveEnvironment = () => {
-  const routeEnvironmentId = router.currentRoute.value.params
-    .environment as string
-  if (routeEnvironmentId) {
-    activeEnvironmentID.value = routeEnvironmentId
-  } else if (routeEnvironmentId === 'default') {
-    activeEnvironmentID.value = environments.default.uid
+function setActiveEnvironment() {
```
geoffgscott, line 309: This sort of setting logic should live in the store. We may want to switch the environment from multiple places.

geoffgscott (commented):
UI looks good. State setting feels like it could use some work.

packages/api-client/src/components/EnvironmentSelector/EnvironmentSelector.vue
```diff
@@ -28,11 +33,55 @@ const createNewEnvironment = () =>
...
+    Object.keys(collection['x-scalar-environments']).length === 0
+  ) {
+    setActiveEnvironment('default')
+  } else if (collection['x-scalar-environment']) {
+    setActiveEnvironment(collection['x-scalar-environment'])
+  } else {
+    const firstEnvironment = Object.keys(collection['x-scalar-environments'])[0]
+    setActiveEnvironment(firstEnvironment)
+  }
+}
```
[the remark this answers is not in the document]
antlio: ✔️

packages/api-client/src/components/Sidebar/SidebarListElement.vue
amritk: +10

amritk (approved):
Looks good! did not test... hopefully I don't get too many conflicts

tmastrom (approved):
Should we warn users that this will delete their current environment variables?

marclave (changes requested):
overall looks good, i wonder if there will eventually be a pull to have multiple "workspace" envs outside of just global! i think postman supports this so we might want to as well.

thoughts on doing that now @geoffgscott @antlio ?

also i cant seem to add more than 1 environment per colleciton, which we need to be able to add more :) 
image


also when i make a new environment, it sets it for the collection regardless of if the environment is part of that collection or not 

image

antlio:
> also i cant seem to add more than 1 environment per colleciton, which we need to be able to add more :)

when adding a new environment, you should be able to select a collection and add other environments, isn't that the case for you?

> also when i make a new environment, it sets it for the collection regardless of if the environment is part of that collection or not

good catch, will fix! it seems to happen on env creation and moving to the request page straight after

hanspagel (changes requested):
love it, but +1 for multiple global environments, if that’s possible

this strange object seems to be the default, at least when creating two envs

image

marclave (approved):
!!!! LETS GOOO !!!!

PR #3867, opened 2024-11-12
feat(components): simplify index exports

hwkr:
Not the final solution but simplifies the index file and makes it easier to export new components.

I don't think this should change the consumed API anywhere but someone who knows about packages can weigh in?


amritk (approved):
Technically violates this [biome rule](https://biomejs.dev/linter/rules/no-re-export-all/) but its not in our style guide at the moment so lets let it go. Correct it will not change anything for anyone consuming, just slight typescript performance

PR #3870, opened 2024-11-13
fix: remove useTemplateRef

### LINKED ISSUE #3859: [Vue warn] Set operation on key "value" failed: target is readonly.
I get the following warning when I run `pnpm dev` in packages/api-reference:

```
create-client.js:45 [Vue warn] Set operation on key "value" failed: target is readonly.
```

pointing to

```ts
// packages/api-client/src/libs/create-client.ts
  app.use(router)
  // Provide the workspace store for the useWorkspace hook
  app.provide('workspace', store)
  // Provide the layout for the useLayout hook
  app.provide('layout', layout)
```

What AI says: The error occurs because Vue 3's provide API creates a readonly reference by default to prevent child components from accidentally mutating provided values. In this case, the store object likely contains reactive properties that need to be mutable.

amritk:
This seems to cause issues when using the react integrations. Maybe hold off for a few versions

Closes #3859

### DIFF
--- a/.changeset/twenty-carrots-hug.md
+++ b/.changeset/twenty-carrots-hug.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
+
+fix: remove useTemplateRef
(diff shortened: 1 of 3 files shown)


hwkr:
image

hwkr (approved):
image

ja-plus:
usefull

amritk:
@hwkr I think its time to bring it back!

PR #3873, opened 2024-11-13
feat(proxy-server): add an OpenAPI document

hanspagel:
This PR adds an OpenAPI document for the Scalar Proxy.

Once #3818 is merged, I’d like to render an API reference on the proxy server, too.

Would be great to interactively test the proxy with our tooling (through the embedded API client or imported into the Scalar App).

**Preview**

https://sandbox.scalar.com/e/XREwq


amritk (approved):
Not quite sure whats going on here but it seems safe 👍🏾

PR #3877, opened 2024-11-13
ci: upload playwright test report

tmastrom:
This PR is a step towards streamlining the snapshot testing pain point. 
If the snapshot testing fails, the snapshot, as well as the snapshot diff are uploaded as github artifacts and surfaced to developers in a PR comment. 

In this iteration, the developer can download the files, assess the differences, and update the the existing snapshots with the files. 

Playwright test results folder has the following structure 
- `snapshot-expected.png` (what it looked like before)
- `snapshot-actual.png` (what it looks like right now)
- `snapshot-diff.png` (highlighting the differences)
** the diff will only exist if the snapshots are the same size. If the view size has changed, only the `-actual` will be present

Next step: Approve and update the diffs in GitHub

Inspired by https://medium.com/@haleywardo/streamlining-playwright-visual-regression-testing-with-github-actions-e077fd33c27c


marclave (approved):
nice work @tmastrom :)

hanspagel (approved):
Love this!!

PR #3879, opened 2024-11-13
style(api-client): more guided empty state in app

cameronrohani:
Add 'import collection' to empty state
image

+ big command k ascii art if <= 1 request

image


amritk (approved):
Seems legit, tested and working 👍🏾

packages/api-client/src/views/Request/ResponseSection/ResponseEmpty.vue
```diff
@@ -57,7 +94,7 @@ onBeforeUnmount(() => events.hotKeys.off(handleHotKey))
         <ScalarHotkey hotkey="↵" />
       </button>
       <button
-        v-if="!isReadOnly && layout === 'desktop'"
+        v-if="layout !== 'modal'"
```
antlio, line 97: should we keep the desktop condition here? so we do not display the meta+n on web?

cameronrohani: feel free to adjust if you think it makes sense to do so

antlio: ok will do! cause we can't use meta+n in web

PR #3888, opened 2024-11-14
fix(components): cleanup ScalarMenu and ScalarListbox

hwkr:
Cleans up ScalarMenu and ScalarListbox, mostly small style changes but I also removed the `fullWidth` prop from `ScalarListbox` because it didn't really do anything.

## ScalarMenu

Improved ScalarMenu responsive / mobile layout and also made it so you can replace the button / trigger if you want.

## ScalarListbox

Removed the fullWidth prop which wasn't really doing anything except adding an extra div around the Listbox button that makes it hard to style.

Also tweaked this dropdown while I was confirming that removing this prop wouldn't break stuff.

https://github.com/user-attachments/assets/72ce9f8e-d707-472e-ae2c-d5b308825c98


packages/components/src/components/ScalarListbox/ScalarListbox.vue
```diff
@@ -58,8 +57,6 @@ const variants = cva({
 <template>
   <Listbox
     v-slot="{ open }"
-    as="div"
```
hwkr, line 61: It will render as [`<template>`](https://headlessui.com/v1/vue/listbox#listbox) now

marclave (approved):

PR #3894, opened 2024-11-15
feat(use-hooks): add useBreakpoint hook

hwkr:
Adds a `useBreakpoint` hook so we can use our Tailwind breakpoints in a consistent way that matches how we use the in CSS. 

Fixes DOC-1338 😱 🧟 

Also fixed a bug with the API Client sidebar while I was in there (it used to be possible to do this if you opened the sidebar with a viewport between 780px and 800px)

Google Chrome-2024-11-14-20-03-21@2x

### DIFF
--- a/packages/use-hooks/src/useBreakpoints/useBreakpoints.ts
+++ b/packages/use-hooks/src/useBreakpoints/useBreakpoints.ts
@@ -1,0 +1,39 @@
+import preset from '@scalar/themes/tailwind'
+import { useMediaQuery } from '@vueuse/core'
+import { type Ref, computed, unref } from 'vue'
+
+type Screen = keyof typeof preset.theme.screens
+
+/**
+ * Exposes Tailwind CSS breakpoints as reactive media queries
+ */
+export function useBreakpoints() {
+  const screens = preset.theme.screens
+
+  const mediaQueries = Object.fromEntries(
+    Object.entries(screens).map(([breakpoint, value]) => [
+      breakpoint,
+      useMediaQuery(`(min-width: ${value})`),
+    ]),
+  ) as Record<Screen, Ref<boolean>>
+
+  // We make the breakpoints a computed object so that we can use them in templates as `breakpoints.x` instead of `breakpoints.x.value`
+  const breakpoints = computed(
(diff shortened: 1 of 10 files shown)


hwkr:
Fixes DOC-1338 😱

amritk (commented):
Not a big fan of this one! Responsiveness in JS is a bad habit. What if we just removed it completely?

hwkr:
> Not a big fan of this one! Responsiveness in JS is a bad habit. What if we just removed it completely?

Yeah I mean I don't love it either but we use it in a bunch of places already and this at least keeps it aligned with our tailwind breakpoints.

There are certain things we can only do with media queries in JS like mounting and unmounting components (otherwise sometimes we might have big hidden DOM trees still being managed by Vue).

It's also worth note that these JS media queries use MatchMedia under the hood which should have similar performance to CSS media queries (or rather it's not like it's polling the viewport size or something)

What do you think if we add a blurb in the readme and on the JSDoc reminding folks to use CSS if at all possible?

amritk (approved):
Approving with the documentation changes!

hanspagel (approved):
Love this! (Didn’t test the code.)

packages/use-hooks/src/useBreakpoints/useBreakpoints.ts
```diff
@@ -0,0 +1,41 @@
+import preset from '@scalar/themes/tailwind'
+import { useMediaQuery } from '@vueuse/core'
+import { type Ref, computed, unref } from 'vue'
+
+type Screen = keyof typeof preset.theme.screens
```
hanspagel, line 5: Can we make the import not load the whole Tailwind preset? That might help with keeping the bundle size increase smaller.

```suggestion
import { theme } from '@scalar/themes/tailwind'
import { useMediaQuery } from '@vueuse/core'
import { type Ref, computed, unref } from 'vue'

type Screen = keyof typeof theme.screens
```

hwkr: It looks like we can't do this because theme isn't actually exported separately 😞

PR #3899, opened 2024-11-15
feat(api-reference): request body property examples

### LINKED ISSUE #3545: example request isn’t used for the request body
### What happens?

When attempting to test a request, the JSON request body is not pre-populated with anything other than an empty body.
image

I would expect the same outcome as the old Swagger UI that pre-populates the request body like so (below is the same endpoint as above):
image

What do I need to configure using Scalar to achieve the same outcome?

### OpenAPI Document

_No response_

antlio:
this pr is a first proposal to fix #3545 in order to display all properties through the api client modal by:
- setting to default false the empty and optional property omission
- omitting empty and optional properties in example request reference

this enables users to get all the properties from the client as a starter.

**⊢ reference before / after**
image
image


packages/oas-utils/src/spec-getters/getRequestBodyFromOperation.ts
```diff
@@ -45,6 +45,7 @@ const mimeTypes = [
 export function getRequestBodyFromOperation(
   operation: Omit<TransformedOperation, 'httpVerb'>,
   selectedExampleKey?: string | number,
+  omitEmptyAndOptionalProperties?: boolean,
```
amritk, line 48: ~when making changes like this, its good to default the value to what it was before to prevent any unforeseen bugs.~
```suggestion
  omitEmptyAndOptionalProperties = true,
```

amritk: but maybe what we had before was bad, I'm not sure of the context exactly, in that case nevermind!

amritk (approved):

PR #3905, opened 2024-11-15
feat: Add basic tests for spec export

geoffgscott:


amritk (approved):
Once my other PR is merged, we should make sure we are ignoring all clientSecrets, passwords etc

marclave (approved):

geoffgscott (commented):
UI looks good. The state setting feels like it needs some work. Could possibly happen in a follow up PR.

PR #3906, opened 2024-11-15
fix(use-hooks): set default color mode to system instead of dark

hwkr:
Initial mode now defaults to `system` instead of `dark` 🤦 

Added some extra tests too.

### DIFF
--- a/packages/use-hooks/src/useColorMode/useColorMode.ts
+++ b/packages/use-hooks/src/useColorMode/useColorMode.ts
@@ -14,12 +14,14 @@ const colorModeSchema = z
  * A composable hook that provides color mode (dark/light) functionality.
  */
 export function useColorMode(opts: UseColorModeOptions = {}) {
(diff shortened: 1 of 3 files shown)


packages/use-hooks/src/useColorMode/useColorMode.ts
```diff
@@ -14,12 +14,14 @@ const colorModeSchema = z
  * A composable hook that provides color mode (dark/light) functionality.
  */
 export function useColorMode(opts: UseColorModeOptions = {}) {
-  const { initialColorMode = 'dark', overrideColorMode } = opts
+  const { initialColorMode = 'system', overrideColorMode } = opts
 
   /** Toggles the color mode between light and dark. */
   function toggleColorMode() {
+    const darkLightMode =
```
hwkr, line 21: Updated this part so when you click toggle it switches to the opposite of your system colorMode if colorMode is 'system'

PR #3913, opened 2024-11-16
feat(api-client): code folding to the response preview

marclave:
Replaces `ScalarCodeBlock` with the `useCodeMirror` hook.

**Problem**
Currently we dont get:
- search
- folding
- any sort of advanced interactions with the text for the future of making variables from responses

**Solution**
Lets use our codemirror package since we handle large responses with virtualization

before:
image

after:
image
*edit*
i fixed the orange 
image

### DIFF
--- a/.changeset/six-seals-accept.md
+++ b/.changeset/six-seals-accept.md
@@ -1,0 +1,6 @@
+---
+'@scalar/use-codemirror': patch
+'@scalar/api-client': patch
+---
+
+fix: replace scalarcodeblock with codemirror for response preview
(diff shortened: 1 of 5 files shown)


hanspagel:
Great idea! @amritk Can you check if that still works well with large responses?

amritk (approved):
Virtual scroll still works but its stuck in this little window, it used to take up the rest of the screen with flex. However I think that was broken somewhere else!

image

hwkr:
> Virtual scroll still works but its stuck in this little window, it used to take up the rest of the screen with flex. However I think that was broken somewhere else!
> 

I can have a look at the flex 🫡

hwkr (approved):
Looks good to me! There are some a11y issues with codemirror but I can fix those separately 🫡

PR #3917, opened 2024-11-17
fix: handle no prefix

marclave:
**Problem**
Currently, we would not handle requests without a protocol (google.com) now we do :) 
image
image

also, we werent properly joining origin with the url 🥶 

now we are 😎


amritk (approved):
Looks about right. Didn't test but it has tests  👍

hanspagel (approved):
Great UX improvement, tested it. 👍

PR #3921, opened 2024-11-18
feat(use-hooks): add computed dark light mode

hwkr:
Small fix to add a computed dark / light mode so that the toggle can show the correct value initially.

## Before

Google Chrome-2024-11-17-21-48-43@2x

## After

Google Chrome-2024-11-17-21-49-15@2x


hanspagel (approved):
Oh, yes, great!

PR #3922, opened 2024-11-18
fix(api-client): falsy value display in RequestTableTooltip

inyourtime:
This PR enhanced template logic to conditionally display `minimum`, `maximum`, and `default` properties even if their values are falsy (0, false) but defined.

before:
Screenshot 2567-11-18 at 15 40 05
after:
Screenshot 2567-11-18 at 15 40 21


hanspagel (approved):
Great PR! 👏

PR #3923, opened 2024-11-18
style(api-reference): font size fixtures

antlio:
this pr fixes font / icon size glitch in api reference as seen below:

**⊢ base url before / after**
<div>
image
image
</div>

**⊢ markdown code before / after**
<div>
image
image
</div>

**⊢ request body select before / after**
<div>
image
image
</div>


cameronrohani (approved):

amritk (approved):
Send it

PR #3925, opened 2024-11-18
feat(icons): add list of icons to the README

hanspagel:
This PR adds a list of all available icons to the README.

Great for a quick preview of what’s in the package.

Screenshot 2024-11-18 at 12 02 38

### DIFF
--- a/.changeset/cold-crabs-unite.md
+++ b/.changeset/cold-crabs-unite.md
@@ -1,0 +1,5 @@
+---
+'@scalar/icons': patch
+---
+
+feat: add list of icons to the README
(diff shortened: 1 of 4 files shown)


amritk (approved):
Ah yes, I was thinking about having a permanently hosted storybook or something but this works!

hwkr (approved):
Love the script!

These icons are supposed to be used as a set to provide users with a group of selectable icons vs `ScalarIcon` which is our UI icons, I wonder if this would make more sense in `scalar/components` either in addition to or instead of here?

Or maybe a bigger conversation is should we unify these with `ScalarIcon` or something 🤔

PR #3928, opened 2024-11-18
refactor(api-reference): make `<Operation />` a feature

hanspagel:
This PR makes the `<Operation />` component a feature. The idea is to make it less reliant on other parts of the code base, to eventually expose this as an embed.

Hard to review, but this PR mostly moves code around and doesn’t change much of the code. This is for another PR.

**Changes**

* everything that’s in the operation section lives in `@/features/Operation`
* all components are short and simple
* it imports way less from other places
* instead of `Operation` and `OperationAccordion` there’s one `Operation` component with two layouts
* `TestRequestButton` got its own feature folder, because it’s used somewhere else, too
* exposed components are in the root of the feature folder, everything else is inside a components folder
* a few simple tests for the `Operation` component, we can extend this while we wire it up to the store
* the `@/` import alias is added to the Vite configuration, the imports of the touched files are changed to use it

**Unrelated changes**

* `type Parameters` actually referred to a single parameter, I renamed it to `Parameter` and added the old name as a deprecated alias

**Next up**

Actually bring in the store and use it in one component?


packages/api-reference/src/components/ApiReferenceLayout.vue
```diff
@@ -333,7 +333,7 @@ const themeStyleTag = computed(
         class="references-rendered">
         <Content
           :baseServerURL="configuration.baseServerURL"
-          :layout="configuration.layout === 'classic' ? 'accordion' : 'default'"
+          :layout="configuration.layout"
```
hanspagel, line 336: we’re just using 'modern' and 'classic' in the Operation component now

packages/api-reference/vite.standalone.config.ts
```diff
@@ -19,6 +18,12 @@ export default defineConfig({
   define: {
     'process.env.NODE_ENV': '"production"',
   },
+  resolve: {
+    alias: {
+      '@': fileURLToPath(new URL('./src', import.meta.url)),
```
amritk, line 23: we get aliases in references now?!?!

amritk (approved):

packages/api-reference/vite.standalone.config.ts
geoffgscott: Haha we have had them for ages.

PR #3931, opened 2024-11-18
feat(api-client): split workspace store into active-entities store

amritk:
We need to remove anything router related from the workspace store so we can initialize it in the references.

### DIFF
--- a/packages/api-client/src/libs/create-client.ts
+++ b/packages/api-client/src/libs/create-client.ts
@@ -1,7 +1,14 @@
 import type { ClientLayout } from '@/hooks'
 import { loadAllResources } from '@/libs/local-storage'
-import { PathId } from '@/routes'
-import { type WorkspaceStore, createWorkspaceStore } from '@/store'
+import {
+  ACTIVE_ENTITIES_SYMBOL,
+  createActiveEntitiesStore,
+} from '@/store/active-entities'
+import {
+  WORKSPACE_SYMBOL,
+  type WorkspaceStore,
+  createWorkspaceStore,
+} from '@/store/store'
 import type { Collection, RequestMethod } from '@scalar/oas-utils/entities/spec'
 import { workspaceSchema } from '@scalar/oas-utils/entities/workspace'
 import { LS_KEYS, objectMerge } from '@scalar/oas-utils/helpers'
@@ -109,10 +116,13 @@ export const createApiClient = ({
     _store ||
     createWorkspaceStore({
       isReadOnly,
-      useLocalStorage: persistData,
-      themeId: configuration.themeId,
       proxyUrl: configuration.proxyUrl,
+      themeId: configuration.themeId,
+      useLocalStorage: persistData,
     })
+
+  // Create the router based active entities store
+  const activeEntities = createActiveEntitiesStore({ ...store, router })
 
   // Load from localStorage if available
   // Check if we have localStorage data
@@ -141,7 +151,6 @@ export const createApiClient = ({
     store.workspaceMutators.add({
       uid: 'default',
       name: 'Workspace',
-      isReadOnly,
       proxyUrl: configuration.proxyUrl,
     })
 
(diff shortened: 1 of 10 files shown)


packages/api-client/src/store/active-entities.ts
```diff
@@ -0,0 +1,185 @@
...
+  requests,
+  router,
+  servers,
+  workspaces,
+}: CreateActiveEntitiesStoreParams) => {
+  /** Gives the required UID usually per route */
+  const activeRouterParams = computed(getRouterParams(router))
+
+  /** The currently selected workspace OR the first one */
+  const activeWorkspace = computed(() => {
```
amritk, line 44: This stuff is all cut out from the other store

packages/api-client/src/libs/create-client.ts
```diff
@@ -158,13 +170,13 @@ export const createApiClient = ({
   const app = createApp(appComponent)
   app.use(router)
   // Provide the workspace store for the useWorkspace hook
-  app.provide('workspace', store)
+  app.provide(WORKSPACE_SYMBOL, store)
   // Provide the layout for the useLayout hook
   app.provide('layout', layout)
```
hwkr, line 175: I guess we could make this a symbol too 🤔

packages/api-client/src/views/Request/Request.vue
```diff
@@ -67,7 +67,7 @@ watch(isNarrow, (narrow) => (showSideBar.value = !narrow))
  */
 const selectedSecuritySchemeUids = computed(
   () =>
-    (isReadOnly.value
+    (isReadOnly
```
hwkr, line 70: Is this not reactive anymore I guess?

amritk: Correct. I couldn't think of any need for it to be reactive. Also it can also just be isModal

hwkr: Well we also have `layout = 'modal'` from [`useLayout`](https://github.com/scalar/scalar/blob/main/packages/api-client/src/hooks/useLayout.ts) we could just use that (also not reactive though).

packages/api-client/src/store/active-entities.ts
```diff
@@ -0,0 +1,189 @@
...
+  )
+
+  /** Simplified list of requests in the workspace for displaying */
+  const activeWorkspaceRequests = computed(() =>
+    activeWorkspaceCollections.value?.flatMap(
+      (collection) => collection.requests,
+    ),
+  )
+
+  /** TODO: need to merge collection environments into the global space */
```
geoffgscott, line 77: No merging into the global space. We should probably be consuming the workspace collection and merging it here with the collection env and then exporting the `activeEnvVars` object

geoffgscott (approved):
Overall looks good. Most of my comments are probably future improvements.

packages/api-client/src/views/Request/Request.vue
amritk: yea we probably should but I'll leave it for a follow up

amritk: https://github.com/scalar/scalar/issues/3942

packages/api-client/src/store/active-entities.ts
```diff
@@ -0,0 +1,189 @@
...
+  Symbol() as InjectionKey<ActiveEntitiesStore>
+
+/**
+ * The active entities store
+ *
+ * This store returns anything related to the currently active entities
+ * The only reason this is a store and not a simple hook is due to storing the current router here
+ */
+export const useActiveEntities = (): ActiveEntitiesStore => {
+  const store = inject(ACTIVE_ENTITIES_SYMBOL)
```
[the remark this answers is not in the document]
amritk: I actually did that at first but the issue was the router. I think since we have multiple routers going we ran into issues previously with the useRouter hook from vue/router. So then I just copied the store

packages/api-client/src/store/active-entities.ts
amritk: ya these comments are just copy pasted from the other store. I can just delete it

packages/api-client/src/store/store.ts
```diff
@@ -72,16 +69,12 @@ type CreateWorkspaceStoreOptions = {
...
-    useLocalStorage = true,
-    defaultProxyUrl = undefined,
-  }: CreateWorkspaceStoreOptions,
-) => {
-  /** Gives the required UID usually per route */
-  const activeRouterParams = computed(getRouterParams(router))
-
+export const createWorkspaceStore = ({
+  useLocalStorage = true,
+  isReadOnly = false,
```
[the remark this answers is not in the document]
amritk: We will be removing this in another ticket so I wouldn't worry too much

https://github.com/scalar/scalar/issues/3942

amritk:
I applied the changes to the other PR. TY!

PR #3935, opened 2024-11-19
feat(api-client): polish non-macos layouts

hwkr:
This PR cleans up some of the spacing around the tabs for the client app and add a tailwind plugin to allow us to style the app based on what platform it's on.

I was hoping we could just use the [`@todesktop/tailwind-variants`](https://github.com/todesktop/tailwind-variants) plugin but it turns out it's written for Tailwind 2 and doesn't work in our electron dev environment (because it's added by toDesktop). I wrote a custom plugin that does basically the same thing but we have to apply the classes inside of `.scalar-client` because all our tailwind variants get scoped into `.scalar-client`. If anyone has a better way to get the platform from electron into the `ApiClientApp.vue` let me know 🙏 

## Before

Electron-2024-11-19-00-31-13@2x

## After

@cameronrohani I adjusted the spacing a bit here, let me know if you want something different (it should be easier to adjust per platform now though).

Electron-2024-11-19-00-28-13@2x

I would expect Windows and Linux to looks something like this (but with a menu bar up above and no traffic lights), but I don't have a system to test on right now. Maybe @amritk you could give it a test?

Electron-2024-11-19-00-28-04@2x


antlio:
regarding the right spacing it's been added initially to be aligned with the other icons below it (selector, table) but maybe we don't want that anymore

are we sure we want this?
image

hwkr:
> regarding the right spacing it's been added initially to be aligned with the other icons below it (selector, table) but maybe we don't want that anymore

I'm not sure what selector table means, do you mean the workspace dropdown? I'm not too bothered by how we want to line it up, as long as it's consistent.

> are we sure we want this?

Do you mean the tabs being to the left or the traffic lights (because the traffic lights wouldn't actually be there on Windows or Linux where the tabs would be flush to the left).

antlio:
> I'm not sure what selector table means, do you mean the workspace dropdown? I'm not too bothered by how we want to line it up, as long as it's consistent.

i meant with these two
image

> Do you mean the tabs being to the left or the traffic lights (because the traffic lights wouldn't actually be there on Windows or Linux where the tabs would be flush to the left).

my bad i got it after and forgot to remove that part 😄 lgtm!

cameronrohani:
This is a good start, I think there's things we should consider in the future but let's get this in ASAP for now!

cameronrohani (approved):

amritk (approved):
Not sure what were looking at, but heres a screenshot

image

hwkr:
> Not sure what were looking at, but heres a screenshot
> 

@amritk can you take a screenshot with some tabs open? Also do you have window controls somewhere (close / minimize etc)

amritk:
image

no window controls unfortunately, just hotkeys

amritk:
@geoffgscott would have window controls

hanspagel:
Ok, let’s get this in then. 🏄

PR #3939, opened 2024-11-19
ci: show red cross for dead markdown links

hanspagel:
With the last PR it shows as ✅. With this PR here it should show as ❌.

Let’s test this here.

EDIT: GitHub Actions makes this an interesting challenge.


hanspagel:
not easy

PR #3954, opened 2024-11-20
feat(api-client): catch OpenAPI parsing errors

hanspagel:
Currently, `createClient` and `createApiClientModal` throw an error for invalid OpenAPI documents.

With this PR those errors are catched and a helpful message is printed to the console.

Screenshot 2024-11-20 at 11 22 55

### DIFF
--- a/packages/openapi-parser/src/utils/makeFilesystem.test.ts
+++ b/packages/openapi-parser/src/utils/makeFilesystem.test.ts
@@ -1,6 +1,5 @@
 import { describe, expect, it } from 'vitest'
 
-import { getEntrypoint } from './getEntrypoint'
 import { isFilesystem } from './isFilesystem'
 import { makeFilesystem } from './makeFilesystem'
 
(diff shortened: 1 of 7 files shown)


packages/openapi-parser/src/utils/makeFilesystem.test.ts
```diff
@@ -1,6 +1,5 @@
 import { describe, expect, it } from 'vitest'
 
-import { getEntrypoint } from './getEntrypoint'
```
hanspagel, line 3: wasn’t used

tmastrom (approved):
Nice work!

geoffgscott (changes requested):
This would be much more streamlined if we just caught the errors as needed in `importSpecFile`

For this particular case i think we should only catch the parsing errors and then log a suitable message to the console using `console.warn`

geoffgscott:
Closing in favour of https://github.com/scalar/scalar/pull/3969/files

hanspagel:
Okay, we’ve got something better!

PR #3957, opened 2024-11-20
fix(api-client): response preview pretty print

### LINKED ISSUE #3955: response preview doesn’t pretty print JSON anymore
With the recent switch from ScalarCodeBlock to CodeMirror we might have lost the pretty print for JSON responses.

antlio:
this pr fixes #3955 and #3962 along:
- response body content display pretty printed
- response body content scrollability back
- copy button style to only display it on content hover
- response headers style

**⊢ header before after**
<div>
image
image
</div>

**⊢ response before after**
<div>
image
image
</div>

### DIFF
--- a/.changeset/funny-pets-play.md
+++ b/.changeset/funny-pets-play.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
+
+style: updates response headers and copy button ui"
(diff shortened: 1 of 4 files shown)


NJDave71:
When will this be available in nuGet?   I am using 1.2.41
Thanks

marclave (approved):

PR #3966, opened 2024-11-21
feat(api-reference): web component

hwkr:
Adds a package export at `@scalar/api-reference/component.js` which registers the references as a web component 🚀 

Why would we want to do this? The only real way fully encapsulate our app (embedded in a webpage) from external styles is to either put it in an iFrame (which is likely to break scroll and all our modals) or to ship it as a web component. In theory web components and the shadow DOM promise to do exactly that. From MDN:

> An important aspect of custom elements is encapsulation, because a custom element, by definition, is a piece of reusable functionality: it might be dropped into any web page and be expected to work. So it's important that code running in the page should not be able to accidentally break a custom element by modifying its internal implementation. Shadow DOM enables you to attach a DOM tree to an element, and have the internals of this tree hidden from JavaScript and CSS running in the page.

I've added two demo pages with some CSS that selects `button` and overrides some of our styles. First start the server with `pnpm dev:reference:cdn`,  then

- http://127.0.0.1:3173/api-reference-local-leaky-no-component.html mounts the references using `standalone.js` (what we serve on the CDN). It should have lots of red buttons because of the global styles.
- http://127.0.0.1:3173/api-reference-leaky-component.html mounts the references using a web component (custom element) called `scalar-api-references` and should not have red buttons because the shadow root encapsulates our Vue app.

Everything should work in the web component version but you may find some little bugs (if so let me know and I can try to fix them). I fixed a few things already like an issue with scrollToId.

Cool benefits of this:

- It prevents all our code from leaking out of the references, we probably wouldn't need `scalar-app` wrapping our tailwind anymore if we don't want to
- It seems to load faster? Maybe it's just my computer but I don't see the UI flash the same when loading the web component
- It makes setting the configuration pretty simple

Problems / Questions:

- Headless UI attaches it's teleport to the `body` but that's outside our Shadow DOM so all our styles break.
- [Browser support is okay](https://caniuse.com/custom-elementsv1), we might need [a polyfill](https://github.com/ungap/custom-elements)
  - Update: I tested in lastest Chrome, Safari and Firefox and everything seems to be 🆗 
- Shipping an app that has to work as a custom element (e.g. in a shadow DOM) and as a normal element adds some complexity (but also we already ship the client in 3 different configurations so 🤷)

### DIFF
--- a/examples/cdn-api-reference/src/index.ts
+++ b/examples/cdn-api-reference/src/index.ts
@@ -1,5 +1,6 @@
 import fastifyStatic from '@fastify/static'
 import apiReferenceBundle from '@scalar/api-reference/browser/standalone.js?raw'
+import apiReferenceComponentBundle from '@scalar/api-reference/component.js?raw'
 import playButtonBundle from '@scalar/play-button?raw'
 import fastify from 'fastify'
 import { readdirSync } from 'node:fs'
@@ -25,6 +26,13 @@ app.get('/api-reference/standalone.js', (_request, reply) => {
     .code(200)
     .header('Content-Type', 'text/javascript; charset=utf-8')
     .send(apiReferenceBundle)
+})
+
+app.get('/api-reference/component.js', (_request, reply) => {
+  reply
+    .code(200)
+    .header('Content-Type', 'text/javascript; charset=utf-8')
+    .send(apiReferenceComponentBundle)
 })
 
 // @scalar/play-button bundle
--- a/examples/cdn-api-reference/src/public/api-reference-leaky-component.html
+++ b/examples/cdn-api-reference/src/public/api-reference-leaky-component.html
@@ -1,0 +1,58 @@
+<!doctype html>
+<html>
+  <head>
+    <title>Scalar API Reference</title>
+    <meta charset="utf-8" />
+    <meta
+      name="viewport"
+      content="width=device-width, initial-scale=1" />
+    <style>
+      :root {
+        --scalar-custom-header-height: 50px;
+        color-scheme: light dark;
+        font-family: sans-serif;
+      }
+      html {
+        max-height: 100dvh;
+      }
+      body {
+        margin: 0;
+      }
+      .custom-header {
+        display: flex;
+        align-items: center;
+        gap: 18px;
+        height: var(--scalar-custom-header-height);
+        padding: 0 18px;
+        position: sticky;
+        justify-content: space-between;
+        top: 0;
+        z-index: 100;
+
+        color: light-dark(#111, #eee);
+        background-color: light-dark(#eee, #111);
+        border-bottom: 0.5px solid light-dark(#111, #eee);
+      }
+      button {
+        background: red !important;
+        border-radius: 6px;
+        padding: 6px 12px;
+      }
+    </style>
+  </head>
+  <body>
+    <header class="custom-header">
+      <b>Leaky Styles with Custom Element</b>
+      <button><code>button { background: red !important; }</code></button>
+    </header>
+    <scalar-api-reference></scalar-api-reference>
+    <script src="/api-reference/component.js"></script>
+    <script>
+      document.querySelector('scalar-api-reference').configuration = {
+        spec: {
+          url: 'https://cdn.jsdelivr.net/npm/@scalar/galaxy/dist/latest.yaml',
+        },
+      }
+    </script>
+  </body>
+</html>
--- a/packages/api-reference/src/components/ApiReference.vue
+++ b/packages/api-reference/src/components/ApiReference.vue
@@ -63,23 +63,27 @@ const favicon = computed(() => configuration.value.favicon)
 useFavicon(favicon)
 </script>
 <template>
-  <!-- Inject any custom CSS directly into a style tag -->
-  <component
-    :is="'style'"
-    v-if="configuration?.customCss">
-    {{ configuration.customCss }}
-  </component>
-  <Layouts
-    :configuration="configuration"
-    :isDark="darkLightMode === 'dark'"
-    :parsedSpec="parsedSpec"
-    :rawSpec="rawSpec"
-    @toggleDarkMode="() => toggleColorMode()"
-    @updateContent="$emit('updateContent', $event)">
-    <template #footer><slot name="footer" /></template>
-    <!-- Expose the content end slot as a slot for the footer -->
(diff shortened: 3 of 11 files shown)


examples/cdn-api-reference/src/public/api-reference-leaky-component.html
```diff
@@ -0,0 +1,58 @@
...
+        padding: 6px 12px;
+      }
+    </style>
+  </head>
+  <body>
+    <header class="custom-header">
+      <b>Leaky Styles with Custom Element</b>
+      <button><code>button { background: red !important; }</code></button>
+    </header>
+    <scalar-api-reference></scalar-api-reference>
```
hwkr, line 48: This is the custom element! ✨

examples/cdn-api-reference/src/public/api-reference-leaky-component.html
```diff
@@ -0,0 +1,58 @@
...
+    <script src="/api-reference/component.js"></script>
+    <script>
+      document.querySelector('scalar-api-reference').configuration = {
```
hwkr, line 51: We have to set the config like this because we can't put an object directly as a DOM attribute, theres probably a clever way to make this work more like the standalone config though if we want.

packages/api-reference/src/components/ApiReference.vue
```diff
@@ -91,23 +91,27 @@ const favicon = computed(() => configuration.value.favicon)
...
-    :rawSpec="rawSpec"
-    @toggleDarkMode="() => toggleColorMode()"
-    @updateContent="$emit('updateContent', $event)">
-    <template #footer><slot name="footer" /></template>
-    <!-- Expose the content end slot as a slot for the footer -->
-    <template #content-end><slot name="footer" /></template>
-  </Layouts>
+  <div
+    class="contents"
+    :class="darkLightMode === 'dark' ? 'dark-mode' : 'light-mode'">
```
hwkr, line 96: We have to apply the color mode within the shadow DOM for it to work. We wouldn't need to do this if switch to using actual CSS color mode to set our variables.

packages/api-reference/src/features/ExampleRequest/ExampleRequest.vue
```diff
@@ -348,7 +348,7 @@ function updateHttpClient(value: string) {
   padding-left: 12px;
   padding-right: 9px;
 }
-.request-card-footer {
+.request-card-footer.request-card-footer {
```
hwkr, line 351: Due to some reordering of how the styles are injected this needed more specificity.

packages/api-reference/src/helpers/scrollToId.ts
```diff
@@ -2,5 +2,7 @@
  * Tiny wrapper around the scrollIntoView API
  */
 export const scrollToId = async (id: string) => {
-  document.getElementById(id)?.scrollIntoView()
+  const root =
```
hwkr, line 5: We need to check if we're in a shadow DOM in order for `getElementById` to work (there's probably a better solution for this)

packages/api-reference/vite.component.config.ts
```diff
@@ -0,0 +1,72 @@
...
+  }, template)
+}
+
+export default defineConfig({
+  define: {
+    'process.env.NODE_ENV': '"production"',
+  },
+  plugins: [
+    // Important: Tells vite to treat all vue sub-components as custom elements
+    vue({ customElement: true }),
```
hwkr, line 22: This part is the important, this is why we need a separate vite config, everything else is just copied from `vite.standalone.config.js`.

See https://github.com/vitejs/vite-plugin-vue/tree/main/packages/plugin-vue#using-vue-sfcs-as-custom-elements

packages/themes/src/variables.css
hwkr, line 1: We need to target both, maybe we could make this configurable instead though?

xC0dex:
This looks fantastic!
Could this potentially become an alternative to the current HTML integration? If so, I was wondering if it might make sense to transition our `Scalar.AspNetCore` package to use this web component-based integration. This approach seems like it would allow us to provide a JavaScript object as the configuration instead of a JSON string, right? 👀

hanspagel:
@xC0dex That might be the plan, but we need to do some more testing and it’ll be a slow transition to it. ✌️

examples/cdn-api-reference/src/index.ts
```diff
@@ -1,5 +1,6 @@
 import fastifyStatic from '@fastify/static'
 import apiReferenceBundle from '@scalar/api-reference/browser/standalone.js?raw'
+import apiReferenceComponentBundle from '@scalar/api-reference/component.js?raw'
```
hanspagel, line 3: should we call this web-component or something to indicate it’s not a Vue component?

examples/cdn-api-reference/src/public/api-reference-leaky-component.html
hanspagel: LOVE IT

packages/api-reference/src/features/ExampleRequest/ExampleRequest.vue
hanspagel: Should we leave a comment? 🤔

packages/api-reference/src/helpers/scrollToId.ts
hanspagel: LGTM

packages/api-reference/vite.component.config.ts
hanspagel: Should we import the other config and just overwrite the plugins attribute (and whatever is different)? Or is that confusing?

hanspagel (changes requested):
I love this PR, left just a few comments

examples/cdn-api-reference/src/index.ts
hwkr: Yeah or custom element maybe?

hanspagel: Would work too!

packages/api-reference/vite.component.config.ts
hwkr: Yeah that might be good if I can get that working

hanspagel:
I love the idea, but I hate stale PRs. Let’s close this, the branch will be there once we’re able to pick it up again. ✌️

PR #3967, opened 2024-11-21
fix: handle errors from importing invalid spec

tmastrom:
Handle `YAMLParseError` as a warning in the console. 
These errors come from trying to import an invalid spec to the workspace. 
This is a short term solution while we work towards better error handling.


geoffgscott:
Close for https://github.com/scalar/scalar/pull/3969/files

PR #3969, opened 2024-11-21
fix(openapi-parser): better handle invalid API definitions

geoffgscott:
Invalid string errors were not being caught properly and causing failures on spec changes. This captures them and assigns an appropriate error message.

### DIFF
--- a/packages/openapi-parser/src/configuration/index.ts
+++ b/packages/openapi-parser/src/configuration/index.ts
@@ -29,6 +29,7 @@ export const ERRORS = {
   EXTERNAL_REFERENCE_NOT_FOUND: 'Can’t resolve external reference: %s',
   FILE_DOES_NOT_EXIST: 'File does not exist: %s',
   NO_CONTENT: 'No content found',
+  INVALID_CONTENT: 'Unable to parse valid JSON or YAML.',
 } as const
 
 export type ValidationError = keyof typeof ERRORS
--- a/.changeset/shiny-geckos-fix.md
+++ b/.changeset/shiny-geckos-fix.md
@@ -1,0 +1,8 @@
+---
+'@scalar/openapi-parser': patch
+'@scalar/oas-utils': patch
+'@scalar/api-client': patch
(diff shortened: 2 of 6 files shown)


packages/openapi-parser/src/configuration/index.ts
```diff
@@ -29,6 +29,7 @@ export const ERRORS = {
   EXTERNAL_REFERENCE_NOT_FOUND: 'Can’t resolve external reference: %s',
   FILE_DOES_NOT_EXIST: 'File does not exist: %s',
   NO_CONTENT: 'No content found',
+  INVALID_CONTENT: 'Unable to parse valid JSON or YAML.',
```
antlio, line 32: nit, wdyt about using:

```suggestion
  INVALID_CONTENT: 'Failed to parse content as valid JSON or YAML.',
```

antlio: should we have one error per type at some point?

antlio (approved):
@hanspagel will have a better eye but looks like great enhancements to me!

hanspagel (approved):
While I think I could also integrate a few of the improvements in the parser, for the specific issue I still prefer #3954.

I think it’s a good thing that `importSpecFile` and `importSpecFromUrl` throw an error that I can then react depending on the use case, e.g. I’d like render the error differently when using the methods in the CLI.

But, if we want to go with this PR here, it needs some work. I’d like to streamline the changes in the parser with the existing code base, and currently we lost the error output for the references. Hmu to discuss this!

Screenshot 2024-11-21 at 10 51 36

hanspagel:
Okay, this is the one we want to get it. But we’ll merge a quick fix first, and I’ll go through this PR here and add to it. ✌️

hanspagel (changes requested):
Will go through it 👍

amritk (approved):
Code looks good, did not test

hanspagel:
This is great, but we need an empty PR queue. I need to pick this up again, after we applied the new formatting.

PR #3972, opened 2024-11-21
fix(oas-utils): catch errors for invalid OpenAPI documents

geoffgscott:
Short term fix to catch invalid parsing


hanspagel:
Great fix (for now)!

Added error output to the API reference again.

Added a few more (unrelated) tests while we add some for `parseSchema` anyway.

hanspagel (approved):

PR #3975, opened 2024-11-21
refactor(api-reference): deprecated proxy, make it use proxyUrl everywhere

hanspagel:
WIP

We’re using `proxy` in @scalar/api-reference and `proxyUrl` in @scalar/api-client`.

The docs wrongly mention that @scalar/api-reference is using proxyUrl.

Let’s just deprecate ~~`proxy`~~ and use `proxyUrl` everywhere to end the confusion.


amritk (approved):
Yes plz

marclave (approved):
:)

PR #3976, opened 2024-11-21
fix(api-reference): section header long content + classic layout responsive

### LINKED ISSUE #3956: Horizontal overflow in ui.
### What happens?

Horizontal overflow in `@scalar/api-reference` version `1.25.68`.

No overflow.

### OpenAPI Document

_No response_

antlio:
this pr fixes #3956 by:
- enhancing section header display with long content
- fixing classic layout responsiveness

**⊢ header before / after**
<div>
image
image 
</div>

**⊢ classic alignment + border before / after**
<div>
image
image
</div>

**⊢ classic mobile before / after**
<div>
image
image
</div>

### DIFF
--- a/packages/api-reference/src/components/Anchor/Anchor.vue
+++ b/packages/api-reference/src/components/Anchor/Anchor.vue
@@ -37,7 +37,6 @@ const getUrlWithId = (id: string) => {
 .label {
   position: relative;
   display: inline-block;
-  text-align: left;
   word-break: break-all;
 }
 .anchor {
--- a/packages/api-reference/src/components/Anchor/Anchor.vue
+++ b/packages/api-reference/src/components/Anchor/Anchor.vue
@@ -37,6 +37,8 @@ const getUrlWithId = (id: string) => {
 .label {
   position: relative;
   display: inline-block;
+  text-align: left;
+  word-break: break-all;
 }
 .anchor {
   position: relative;
--- a/packages/api-reference/src/components/Content/Content.vue
+++ b/packages/api-reference/src/components/Content/Content.vue
@@ -135,10 +135,14 @@ const introCardsSlot = computed(() =>
 }
 .introduction-card-item {
   padding: 9px;
-  border-bottom: var(--scalar-border-width) solid var(--scalar-border-color);
   display: flex;
   flex-direction: column;
-  justify-content: center;
+  justify-content: start;
+}
+@container narrow-references-container (max-width: 900px) {
+  .introduction-card-item {
+    border-bottom: var(--scalar-border-width) solid var(--scalar-border-color);
+  }
 }
 .introduction-card-item:has(.description) :deep(.server-form-container) {
   border-bottom-left-radius: 0;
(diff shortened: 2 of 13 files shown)


packages/api-reference/src/components/Anchor/Anchor.vue
```diff
@@ -37,6 +37,8 @@ const getUrlWithId = (id: string) => {
 .label {
   position: relative;
   display: inline-block;
+  text-align: left;
```
hwkr, line 40: It's weird this wasn't left already 🤔

packages/api-reference/src/components/Content/Content.vue
```diff
@@ -127,10 +127,14 @@ const introCardsSlot = computed(() =>
 }
 .introduction-card-item {
   padding: 9px;
-  border-bottom: var(--scalar-border-width) solid var(--scalar-border-color);
   display: flex;
   flex-direction: column;
-  justify-content: center;
+  justify-content: start;
+}
+@media (max-width: 900px) {
```
hwkr, line 134: I think this uses a container query elsewhere, @cameronrohani?

See [here](https://github.com/scalar/scalar/blob/main/packages/api-reference/src/components/ClassicHeader.vue#L24) or [here](https://github.com/scalar/scalar/blob/main/packages/api-reference/src/components/Content/Content.vue#L164) and the [container](https://github.com/scalar/scalar/blob/main/packages/api-reference/src/components/Content/Content.vue#L108-L111).

packages/api-reference/src/components/Section/SectionHeader.vue
```diff
@@ -12,17 +12,30 @@ withDefaults(
...
-  <LoadingSkeleton v-if="loading" />
-  <component
-    :is="`h${level}`"
-    v-else
-    class="section-header"
-    :class="{ tight }">
-    <slot />
-  </component>
+  <div class="section-header-wrapper">
+    <LoadingSkeleton v-if="loading" />
```
hwkr, line 16: To be honest I don't know if this skeleton actually works...

hwkr (changes requested):
Looks good! I think we probably just want to swap that media query for a container query, also would be good to get @cameronrohani's 👀 on this before merging because he wrote a bunch of that layout logic.

packages/api-reference/src/components/Section/SectionHeader.vue
hanspagel: I doubt it, too. Haven’t seen it in a while. 😅

antlio: I just kept (moved) what we have nothing else, but we could ff and remove it if need

hanspagel (approved):
LGTM, but I leave this to @hwkr and @cameronrohani

packages/api-reference/src/components/Section/SectionHeader.vue
hwkr: Yeah sorry that wasn't an action item for you just me noticing it.

hwkr (approved):
Looks good! Thanks for swapping those queries :)

PR #3977, opened 2024-11-21
fix(api-client): request table row addition

antlio:
this pr fixes the row addition within api client request table that was prevented when using `@scalar/api-reference`.


marclave (approved):
nice :)

amritk (approved):
Works!

PR #3982, opened 2024-11-21
fix(components): scalar menu flex not on just small screen

cameronrohani:
This menu has flex-col on larger, screens

image

this pr fixes that


amritk (approved):

hwkr (approved):

hwkr:
Needs a changes easy I think

PR #3984, opened 2024-11-22
feat(components): cleanup and extend floating components

hwkr:
Updates and cleans up a bunch of the floating components (ScalarFloating, ScalarDropdown etc.). Theres a few small breaking changes in here but since we're the only ones that use `@scalar/components` (and it's semver < 1.0.0) I don't think we should have any problems. That being said a quick test of the client and references never hurt.

### In `@scalar/components`

* All components
  * Cleaned up the stories and docs
  * Updated them to provide all their `FloatingOptions` to their `ScalarFloating` component (e.g. middleware)
  * Updated their slots to include the menu open state where available (so you can make a chevron flip over in the trigger button or whatever).
  * Defined all slots

* `ScalarFloating`
  * Added an example and a story of how to use the middleware prop to adjust the offset.
  * Updated the type export to be `ScalarFloatingOptions` to be consistent with the rest of the library.
  * Refactored the `targetRef` prop to be `target` and to take either an `HTMLElement` or a `string` for an ID and added tests to that effect. 
  * Remove the `isOpen` prop which wasn't being used except for in the `EnvironmentVariableDropdown.vue` (see below). As a primitive `ScalarFloating` shouldn't really control it's open state and rather should just handle the position of the elements. If we want to hide the floating element we can just put a `v-if` on the `#floating` slot (but all the elements using it right now handle their own open and closed state).
* `ScalarListbox`
  * Extracted the listbox checkbox to be it's own component since we use it in a bunch of places
  * Updated the the listbox to use square checkboxes if it's in multiselect mode
  * Added a story showing how multiselect works
  * Improved the types for multiselect so you can't give it the wrong `v-model` type
  * Extracted the listbox item component so it can be reused
* `ScalarCombobox`
  * Added `#before` and `after#` slots so we can add buttons before and after the combobox list (note: this only works a11y because the combobox is actually a popover and you can tab through it unlike the listbox or dropdown).
* `ScalarDropdown`
  * Remove the `isOpen` and `static` props which weren't being used except for in the `EnvironmentVariableDropdown.vue, see below.
* `ScalarPopover`
  * Just cleaned it up and typed the slots
* `ScalarMenu`
  * Cleaned the floating props
  * Removed a negative margin from the button

### Outside of `@scalar/components`

* Updated `EnvironmentSelector.vue` and `WorkspaceDropdown.vue` to use the standardized `<ScalarListboxCheckbox>`.
* Refactored `EnvironmentVariableDropdown.vue` to not use `ScalarDropdown` anymore. It wasn't using very much from `<ScalarDropdown>` and was basically just rendering an absolutely positioned box in the dropdown (while running floating UI with no reference element). At some point might be worth refactoring the EnvironmentVariableDropdown to actually use FloatingUI to target the code mirror editor column or something.
* Updated the prop name from `targetRef` to `target` in `RequestSidebarItemMenu.vue`

### DIFF
--- a/packages/api-client/src/views/Environment/EnvironmentVariableDropdown.vue
+++ b/packages/api-client/src/views/Environment/EnvironmentVariableDropdown.vue
@@ -165,6 +165,7 @@ onClickOutside(
           icon="Add" />
         Add Variable
       </ScalarButton>
+      <!-- Backdrop for the dropdown -->
       <div
         class="absolute inset-0 -z-1 rounded bg-b-1 shadow-lg brightness-lifted" />
     </div>
--- a/packages/api-client/src/views/Environment/EnvironmentVariableDropdown.vue
+++ b/packages/api-client/src/views/Environment/EnvironmentVariableDropdown.vue
@@ -1,10 +1,10 @@
 <script setup lang="ts">
 import { parseEnvVariables } from '@/libs'
 import type { ActiveEntitiesStore } from '@/store/active-entities'
-import { ScalarButton, ScalarDropdown, ScalarIcon } from '@scalar/components'
+import { ScalarButton, ScalarIcon } from '@scalar/components'
 import { onClickOutside } from '@vueuse/core'
 import Fuse from 'fuse.js'
-import { computed, onMounted, ref } from 'vue'
+import { type CSSProperties, computed, onMounted, ref } from 'vue'
 import type { Router } from 'vue-router'
 
 const props = defineProps<{
@@ -54,7 +54,7 @@ const filteredVariables = computed(() => {
   }
 
   /** filter environment variables by name */
-  const result = fuse.search(searchQuery)
+  const result = fuse.search(searchQuery, { limit: 10 })
   if (result.length > 0) {
     return result
       .map((res) => res.item)
@@ -108,6 +108,14 @@ onMounted(() => {
   selectedVariableIndex.value = 0
 })
 
+const dropdownStyle = computed<CSSProperties>(() => {
+  return {
+    left: (props.dropdownPosition?.left ?? 0) + 'px',
+    // Add a 5px offset from the editor
+    top: (props.dropdownPosition?.top ?? 0) + 5 + 'px',
+  }
+})
+
 onClickOutside(
   dropdownRef,
   () => {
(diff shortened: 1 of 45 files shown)


antlio:
great great work! 

just spotted the following so far:

<details>
<summary>environment selector out of window</summary>
image
</details>


<details>
<summary>command request selector extra gap</summary>
image
</details>

<details>
<summary>error on sidebar request dropdown click</summary>
```json
Error: Passing props on "template"!
The current component <MenuButton /> is rendering a "template".
However we need to passthrough the following props:
  - aria-controls
  - aria-expanded
  - aria-haspopup
  ```
</details>

<details>
<summary>auth dropdown doesn't open</summary>
```json
index-BeJdfT_j.js:14 TypeError: Cannot read properties of null (reading 'emitsOptions')
  ```
</details>

**Comment by amritk** [comment]
For some reason I cannot import galaxy into the client (CORS error). Going through the code now

File: packages/api-client/src/views/Environment/EnvironmentVariableDropdown.vue [hunk 1]
```diff
@@ -161,6 +165,8 @@ onClickOutside(
           icon="Add" />
         Add Variable
       </ScalarButton>
-    </template>
-  </ScalarDropdown>
+      <div
+        class="absolute inset-0 -z-1 rounded bg-b-1 shadow-lg brightness-lifted" />
```
amritk, line 169: whats this div aboot

amritk (changes requested):
Code looks good, tests 👍🏾, docs 👍🏾 

In testing I only found this issue: when you click on the menu button in the sidebar, all of the ellipsis are showing. I think we fixed this before but was re-introduced here

image

packages/api-client/src/views/Environment/EnvironmentVariableDropdown.vue
amritk: yolo

amritk (approved):

packages/api-client/src/views/Environment/EnvironmentVariableDropdown.vue
hwkr: Added a comment

antlio:
last nit is that is we open a request sidebar menu, and want to open another one, it gets now closed vs swapping as before:

image

antlio (approved):
great great work - onlly one last non blocking comment

hwkr:
> last nit is that is we open a request sidebar menu, and want to open another one, it gets now closed vs swapping as before:
> 
> image

Mmmm okay, I think maybe I'll merge then fix this after

PR #3985, opened 2024-11-22
fix(api-client): css conflict with grid

cameronrohani:
sometimes people apply classes globally to `.grid`, this stretches the content so that the layout doesn't break when people do that


antlio (approved):
legit

PR #3987, opened 2024-11-22
refactor(mock-server): use path aliases

hanspagel:
This PR introduces the path aliases we already have in other packages to the `@scalar/mock-server`.

Easier to move files around without modifying the imports.

Stole the configuration from `@scalar/oas-utils`.

Also removes the barrel file from `./utils`.


amritk (approved):
Down with barrels!

PR #3988, opened 2024-11-22
chore(api-client): remove duplicated tests

hanspagel:
I found out that I duplicated all the tests for the `send-request` lib. Might be a sign that this lib just has grown a lot, so this PR does two things:

* removing the duplicated tests, ensuring the ones we added recently are still there
* giving `send-request` its own folder to be able to split things up a bit

No code was harmed.


packages/api-client/src/libs/index.ts
```diff
@@ -11,4 +11,3 @@ export * from './environment-parser'
 export * from './isUrl'
 export * from './getOpenApiDocumentDetails'
 export * from './parse-curl'
-export { createRequestOperation } from './send-request'
```
hanspagel, line 14: duplicated export?

packages/api-client/src/libs/send-request/create-request-operation.test.ts
```diff
@@ -575,4 +575,23 @@ describe('create-request-operation', () => {
...
+    if (error) throw error
+    const [requestError, result] = await requestOperation.sendRequest()
+
+    expect(requestError).toBe(null)
+    expect(JSON.parse(result?.response.data as string)).toMatchObject({
+      method: 'GET',
+      path: '/me',
+      body: '',
+    })
+  })
```
hanspagel, line 596: this was added recently to the other test file

marclave (approved):
:)

PR #3991, opened 2024-11-22
feat(snippetz): dynamically extend types

hanspagel:
Currently, we’re just using two enums to maintain a list of supported targets and clients for `@scalar/snippetz` and those two don’t even know about each other (it would have been fine to combine `shell` with `fetch`).

With this PR, we’re revamping the types so they can be extended in the plugin files and they have a relation. TS will complain when you combine `shell` with `fetch` now. Comes with type tests.

I didn't manage to maintain an object like structure in TS like shown below, because at some point it ended in a self-reference (to extend the list of clients) one way or the other:

```js
// ❌
{
  node: ['undici', 'fetch'],
}
```

So what I did was to use a list of strings, which seems to work pretty well. Imagine it to look like that:

```js
// ✅
[
  'shell/curl',
  'node/undici',
  'node/fetch',
  // …
]
```

Note: Declaring them globally isn’t perfect, but I was so sidetracked already, that I just went with it for now. Ideally they’d only be in the list when the source code is actually loaded.

**Auto-complete**

Screenshot 2024-11-22 at 15 32 36


packages/snippetz/src/core/types-proposed.ts
```diff
@@ -0,0 +1,29 @@
...
+        : never,
+>(
+  target: TTarget,
+  client: TClient,
+): { snippet: string } => {
+  // do snippet things
+  return {
+    snippet: 'my-snippet',
+  }
+}
```
tmastrom, line 24: Here is an alternative idea for getting type inference in the snippetz. Maybe there are some drawbacks to this approach but I'm preferring this over global. Let's have more discussion!

marclave (approved):
this looks good to me interface wise! : )

packages/snippetz/src/core/types.ts
```diff
@@ -1,18 +1,34 @@
 export type { Request } from 'har-format'
 
+declare global {
+  // eslint-disable-next-line @typescript-eslint/consistent-type-definitions
```
amritk, line 4: what was this for exactly?

amritk: or is it just part of snippetz

amritk (approved):
Oh I get it, if the plugin is added then it accepts the type. Cool 👍🏾

PR #3993, opened 2024-11-22
fix(proxy-server): remove original `Origin` header

hanspagel:
For cross-origin requests, the browser adds an `Origin` header with the original URL the request is coming from. That might confuse some target servers, because the request is then coming from the proxy and not from the original URL anymore.

As the whole point of the proxy is to make it work like same-site requests, we might be able to just delete the `Origin` header. This PR does that.

ℹ️ The same fix was applied to #3848.


hanspagel:
Closing in favor of #3848

PR #3994, opened 2024-11-22
feat(api-reference): displays path server in request example

### LINKED ISSUE #3964: Support server url definition per individual endpoint
### What happens?

I believe the current OAS standard supports defining a server url per endpoint as well as globally. Currently, if you define a global server url on the entire oas file, and then you define a server url on a specific endpoint, the server url on the specific endpoint is not respected.  For example: in the below spec, `https://1231` will not be applied to the endpoint as `https://test` overrides everything. (using the api-reference-react package)

{
  "openapi": "3.0.0",
  "info": {
    "title": "test",
    "version": "test"
  },
  "servers": [
          {
            "url": "https://test"
          }
        ],
  "paths": {
    "/users/{userId}/data_c": {
      "post": {
        "servers": [
          {
            "url": "https://1231"
          }
        ],
        "tags": [
          "Labels"
        ],
        "summary": "My Testing Open APIs",
        "description": "dasfdsa",
        "operationId": "MyTestingOpenAPIs",
        "parameters": [
          {
            "name": "userId",
            "in": "path",
            "required": true,
            "schema": {
              "type": "string"
            }
          }
        ],
        "responses": {
          "200": {
            "description": "123"
          },
          "429": {
            "description": "description"
          }
        }
      }
    }
  }
}
 

If an individual endpoint defines it's own server url, the UI should display that.

### OpenAPI Document

_No response_

antlio:
this pr is a proposal to fix #3964 in order to favor the path server in the request example if it exists:

**⊢ before / after**
image
image

should be ff with setting the right server in the client modal on test request.

### DIFF
--- a/.changeset/sour-hotels-exist.md
+++ b/.changeset/sour-hotels-exist.md
@@ -1,0 +1,5 @@
+---
+'@scalar/types': patch
+---
+
+feat: adds servers to information reference config
--- a/.changeset/twelve-elephants-smell.md
+++ b/.changeset/twelve-elephants-smell.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
+feat(api-reference): displays path server in request example
--- a/packages/api-reference/src/features/ExampleRequest/ExampleRequest.vue
+++ b/packages/api-reference/src/features/ExampleRequest/ExampleRequest.vue
@@ -130,10 +130,16 @@ async function generateSnippet() {
     )
   }
 
-  // Generate a request object
+  // Generate a request from operation server or fallback to global server URL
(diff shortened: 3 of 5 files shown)


marclave (changes requested):
@antlio is it supposed to not have a full proper code example?

sorry maybe im misunderstanding the change since the request parameters are missing

antlio:
> @antlio is it supposed to not have a full proper code example?
> 
> sorry maybe im misunderstanding the change since the request parameters are missing

my bad maybe not the best example used here! that's the current behavior when there is no host value set as we cannot create the url. it was to show here the support of path server variables, otherwise it would just display as regular:

image

marclave (approved):

PR #4018, opened 2024-11-26
feat(components): allow logo in ScalarMenu to adjust width

hwkr:
Allows us to use a logo of flexible width in the ScalarMenu. We can still keep it square if we prefer but at least this way it's more flexible.

Also added an example to storybook (which use the storybook logo).

Google Chrome-2024-11-26-00-42-30@2x


marclave (approved):
rad :) lets get it in!

PR #3981, opened 2024-11-21
feat(api-reference): use API client auth

amritk:
closes #3686
closes #3884
closes #3959

Most of the file changes are deletions (40 deleted files)! I got some ideas on how we can move activeEntities to a regular hook instead of a store. but will leave that for another PR 

Also embed doesn't work, just fixed the type errors but @hanspagel mentioned we aren't using it yet. Need to come up with a better state strategy which we will do soon 🇫🇷 

New UI updates, everything is good except for oauth2 which should get redesigned soon anyway.
image
image
image

PR #4023, opened 2024-11-26
feat(api-reference): skin client auth in references

amritk:
Can probably do something cleaner codewise but we need to get this in so 🤷🏾. Gotta ensure the api client side is unchanged

image
image
image

image


packages/api-client/src/components/DataTable/DataTableCell.vue
```diff
@@ -11,7 +12,12 @@ withDefaults(
 <template>
   <component
     :is="is"
-    class="min-h-8 min-w-8 border-l-0 border-t-0 border-b-1/2 border-r-1/2 flex text-sm last:border-r-0 group-last:border-b-transparent p-0 m-0 relative"
+    :class="[
+      tw(
+        'min-h-8 min-w-8 border-l-0 border-t-0 border-b-1/2 border-r-1/2 flex text-sm last:border-r-0 group-last:border-b-transparent p-0 m-0 relative',
+        $attrs.class as string,
```
amritk, line 18: use tw-merge to override classes

packages/api-client/src/components/DataTable/DataTableInput.vue
```diff
@@ -11,7 +11,10 @@ const props = withDefaults(
     id?: string
     type?: string
     /** Class for the wrapping cell because attrs is bound to the input */
-    containerClass?: string
+    containerClass?:
```
amritk, line 14: couldn't find the class for this in vue 🕵🏾

hanspagel:
Looks rad!

packages/api-client/src/components/DataTable/DataTableInput.vue
hwkr: Looks like Vue uses `any`? Maybe we should just make our own utility type.

Cursor-2024-11-26-14-16-24@2x

amritk: lets do it

packages/api-client/src/components/DataTable/DataTableCell.vue
hwkr: Do we actually want to `as string`? tw takes a few different options (also Vue will anything and smush it onto the class so we might need `inheritAttrs: false`)?

Cursor-2024-11-26-14-23-45@2x

packages/types/src/external/vue.ts
```diff
@@ -0,0 +1,12 @@
+/**
+ * Vue class prop type
+ *
+ * Vue seems to just use any behind the scenes
+ *
+ * @see https://vuejs.org/guide/essentials/class-and-style.html#class-and-style-bindings
+ */
+export type VueClassProp =
```
hwkr, line 8: It can also be a `MaybeComputedOrRef`

hwkr (approved):
Looks good! I'll stop leaving comments now 🫨

packages/types/src/external/vue.ts
amritk: This would add vue as a dependency so either we leave it out, or just move this type into the client. Maybe I'll move it into the client as thats becoming our "core" now

packages/api-client/src/components/DataTable/DataTableCell.vue
amritk: aha no as string was just go get it to work. It comes in as any so it doesn't make a difference what we put. We can do it for posterity's sake

cameronrohani (changes requested):
Awesome work! Needs some changes on the design end, maybe @hwkr can help take this over the finish line:

1. Height /spacing of auth button needs to match height of client libraries / server, it's slightly off
image
 
2. there's no border bottom between auth + client libraries
image

3. individual auths should likely be in their own cards i.e before:
image

after:
image


When we put it all together should look like this:
image

hanspagel (changes requested):
Nit pick police here. Can we make “URL” uppercase? Token Url > Token URL, Redirect Url > Redirect URL …

amritk:
Since the code has been reviewed, I will merge this PR into the other one to create a super PR. Then I will make those design fixes from @cameronrohani. This is so I don't need to spend half my time rebasing a rebase.

image

amritk:
nooooooooooooooooooooooooooooo @hanspagel did you rebase this PR? we can't rebase PR's that are based on other branches. Its dead now 😢

amritk:
This PR is now closed. work will continue in the other one

https://github.com/scalar/scalar/pull/3981

antlio:
> This PR is now closed. work will continue in the other one
> 
> #3981

let's cherry pick the commits?

amritk:
yup already picked into the other one. I should have added a do not rebase tag! 😓

PR #4026, opened 2024-11-27
fix(api-reference): move client config to ApiReferenceLayout

amritk:
In some integrations as well as our docs, we use `ApiReferenceLayout` directly. In these cases setting hidden clients was not working. This PR moves that config mapping into the layout so it works everywhere!

### DIFF
--- a/packages/api-reference/src/components/ApiReferenceLayout.vue
+++ b/packages/api-reference/src/components/ApiReferenceLayout.vue
@@ -1,4 +1,5 @@
 <script setup lang="ts">
+import { useHttpClientStore } from '@/stores/useHttpClientStore'
 import { provideUseId } from '@headlessui/vue'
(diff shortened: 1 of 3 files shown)


packages/api-reference/src/components/ApiReferenceLayout.vue
```diff
@@ -257,6 +258,28 @@ provide(INTEGRATION_SYMBOL, () =>
     : 'vue',
 )
 
+// ---------------------------------------------------------------------------/
+// HANDLE MAPPING CONFIGURATION TO INTERNAL REFERENCE STATE
+
+/** Helper utility to map configuration props to the ApiReference internal state */
+function mapConfigToState<K extends keyof ReferenceConfiguration>(
```
amritk, line 265: copy pasted for now but it will get deleted in the other component in my other branch

marclave (approved):

PR #4028, opened 2024-11-27
fix(postman-to-openapi): null server when parsing endpoint folder

Mason-Little:
**Problem**

When importing a spec that includes an example for a specific endpoint, the parser incorrectly treats the endpoint as a folder. This causes it to skip parsing the server information.

Solution

In this PR, I updated `serverHelpers.ts ` to correctly handle nested folder structures by:

- Adding a `processItem` function that recursively goes through the collection tree.
- Making sure it processes both requests and items inside nested folders.
- Keeping the existing logic for extracting the domain but applying it to each level of the collection.

### DIFF
--- a/packages/postman-to-openapi/src/helpers/serverHelpers.ts
+++ b/packages/postman-to-openapi/src/helpers/serverHelpers.ts
@@ -11,27 +11,34 @@ import { getDomainFromUrl } from './urlHelpers'
 export function parseServers(
   postmanCollection: PostmanCollection,
 ): OpenAPIV3_1.ServerObject[] {
-  // Set to store unique domains
   const domains = new Set<string>()
 
-  if (postmanCollection.item && Array.isArray(postmanCollection.item)) {
-    postmanCollection.item.forEach((item) => {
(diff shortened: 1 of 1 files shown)


hanspagel:
Great catch! Would you mind adding a test? This sounds like a perfect opportunity to make sure we won’t introduce this bug again.

Mason-Little:
> Great catch! Would you mind adding a test? This sounds like a perfect opportunity to make sure we won’t introduce this bug again.

Looking into this bug more, the nested endpoints are causing way more issues than I first thought (e.g., response bodies, response codes, etc.). This has turned into a pretty high priority since a lot of the top Postman workspaces are using this new example format. Because this was built off tests from the old parser, I’m going to need to refactor a decent chunk of the code to fix it.

That said, the other parsers just completely error out when they hit nested endpoints, so we’re still ahead of the curve. I’ll open up a new PR later today with the fix, add more tests to make sure this doesn’t happen again, and throw in support for environment imports while I’m at it. Unfortunately, this will push the Insomnia import work back a little

PR #4037, opened 2024-11-28
fix(use-codemirror): dispatch value frame

antlio:
this pr fixes the error fired when changing request along using a variable in code mirror input:

```json
Error: Calls to EditorView.update are not allowed while an update is in progress
```


hanspagel (approved):
Cool, thankful you found it

marclave (approved):

PR #4044, opened 2024-11-28
feat(snippetz): add clients from `httpsnippet-lite`

hanspagel:
This PR moves the httpsnippet-lite code example generators to the `@scalar/snippetz` package.

That way, we can only import the converters that we still need and replace them one by one with our an own implentation.

And for all the packages that need those code examples, we have just one API and we don’t need to change anything while replace the converts.

This PR adds the following clients from `httpsnippet-lite` (even including tests for some):

* c/libcurl
* clojure/clj_http
* csharp/httpclient
* csharp/restsharp
* go/native
* http/http1.1
* java/asynchttp
* java/nethttp
* java/okhttp
* java/unirest
* js/axios
* js/jquery
* js/xhr
* kotlin/okhttp
* node/axios
* objc/nsurlsession
* ocaml/cohttp
* php/curl
* php/guzzle
* powershell/restmethod
* powershell/webrequest
* python/python3
* python/requests
* r/httr
* ruby/native
* shell/httpie
* shell/wget
* swift/nsurlsession

Also adds the aliases to `@scalar/snippetz`.

We can simplify the code example generation in a separate PR.


packages/mock-server/vite.config.ts
```diff
@@ -5,6 +5,5 @@ export default defineConfig({
   plugins: [],
   resolve: {
     alias: alias(import.meta.url),
-    dedupe: ['vue'],
```
hanspagel, line 8: just cleaning up here

packages/snippetz/playground/src/components/CodeExample.vue
```diff
@@ -1,16 +1,12 @@
...
-  type ClientId,
-  type TargetId,
-  objectToString,
-} from '@scalar/snippetz/core'
+import { type TargetId, objectToString } from '@scalar/snippetz/core'
 import { getHighlighter } from 'shikiji'
 import { onMounted, ref, watch } from 'vue'
 
 const props = defineProps<{
-  target: TargetId
```
marclave, line 12: is there a reason why we opted for string instead of a ClientID type @hanspagel ?

packages/snippetz/playground/src/components/CodeExample.vue
```diff
@@ -19,7 +15,13 @@ const highlightedResult = ref('')
 const highlightedExample = ref('')
 async function renderExample() {
   // Code
-  code.value = snippetz().print(props.target, props.client, props.request) ?? ''
+  code.value =
+    snippetz().print(
+      props.target as TargetId,
+      // @ts-expect-error I don’t know how to type this.
```
marclave, line 21: @amritk maybe can help?

marclave (commented):
AWESOME! just left a couple comments :) then it looks good to me

amritk (approved):
Did not test but seems legit 🦖

packages/snippetz/playground/src/components/CodeExample.vue
hanspagel: Thanks for bringing this up, I’ll have to fix the types in a separate PR, though. ✌️ There’s just too much that we need to change.

PR #4045, opened 2024-11-28
fix(oas-utils): updates find variables import

antlio:
this pr updates import made in `findVariables` to fix circular dependencies caught by @marclave.


hanspagel (approved):
yess

PR #4052, opened 2024-11-29
fix(fastify-api-reference): export map needs explicit types condition

yshrsmz:
**Problem**
While I'm working on my Fastify project, I noticed that `FastifyApiReferenceOptions` can not be resolved.

```ts
import type { FastifyApiReferenceOptions } from '@scalar/fastify-api-reference'
```

Above import statement shows a bellow error.

```
Module '"@scalar/fastify-api-reference"' has no exported member 'FastifyApiReferenceOptions'. Did you mean to use 'import FastifyApiReferenceOptions from "@scalar/fastify-api-reference"' instead?ts(2614)
```

**Solution**

I'm not 100% sure, but some `moduleResolution` mode does not read `types` field in `package.json` when export map exists, and export map needs `types` condition to properly resolve TypeScript types.

https://www.typescriptlang.org/docs/handbook/modules/reference.html#the-moduleresolution-compiler-option

According to the spec, IIUC, it should be able to resolve type via file extension substitution, but it can't be resolved in my environment and needs `types` condition.

### DIFF
--- a/.changeset/quick-turkeys-sniff.md
+++ b/.changeset/quick-turkeys-sniff.md
@@ -1,0 +1,5 @@
+---
+'@scalar/fastify-api-reference': patch
+---
+
+fix: explicit export map is missing the type export
--- a/examples/fastify-api-reference/src/index.ts
+++ b/examples/fastify-api-reference/src/index.ts
@@ -1,5 +1,7 @@
 import fastifySwagger from '@fastify/swagger'
-import fastifyApiReference from '@scalar/fastify-api-reference'
+import fastifyApiReference, {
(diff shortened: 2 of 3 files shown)


hanspagel:
Thanks for the PR, great catch! Those exports are auto-generated, so they might get overwritten again. I’ll debug the issue and add to this PR here.

hanspagel:
I was wrong, the build system for fastify doesn’t generate the exports. I’ve just added some minor tweaks and a changeset. Let’s get this in! :)

hanspagel (approved):

yshrsmz:
Thanks!

PR #4070, opened 2024-12-02
refactor(api-reference): remove `httpsnippet-lite`

hanspagel:
WIP

⚠️ Merge #4044 first.

We’ve integrated `httpsnippet-lite` into `@scalar/snippetz` (and will eventually remove it), but we can already remove it from the API reference and make the code simpler.


hanspagel:
Need to start fresh

PR #4073, opened 2024-12-03
fix(oas-utils): throws an error for relative file paths

hanspagel:
When we added support for requests to URLs without the http/https protocol prefix, we’ve added a check that was looking for dots in URLs to detect domains.

Turns out, relative file paths with a dot in them look like valid URLs (without a protocol). That led to an exception when e.g. `openapi.yaml` was passed as an OpenAPI document URL.

This PR tries to create a new `URL` object in the `isLocalUrl` check, which will fail for a bunch of cases and then return false - which eventually makes our packages to skip the proxy. Which is what we want for relative URLs (but also for invalid URLs).

**Example**

```html
<script
  id="api-reference"
  data-url="openapi.yaml"></script>
```

**Exception**

Screenshot 2024-12-03 at 09 05 23

### DIFF
--- a/.changeset/slimy-bananas-prove.md
+++ b/.changeset/slimy-bananas-prove.md
@@ -1,0 +1,5 @@
+---
+'@scalar/oas-utils': patch
+---
+
+fix: allow to pass relative file names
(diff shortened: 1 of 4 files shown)


antlio (approved):
goat

marclave (approved):

PR #4074, opened 2024-12-03
feat(api-reference): import from embedded OpenAPI documents

hanspagel:
This PR makes it possible to import embedded API definitions to the API client:

* always shows _the button_, even for embedded content
* fixes the import for embedded content
* falls back to passing the API reference URL for the import (if we don’t have an OpenAPI document URL)


marclave (approved):
wicked

PR #4077, opened 2024-12-03
fix(api-client): sidebar request examples

antlio:
this pr is a proposal to filter out the first request example in order to avoid duplicating the main request data, this way we only access main request example from the request, and show up manually created examples:

⊢ before / after
<div>
image
image
</div>

### DIFF
--- a/packages/api-client/src/views/Request/RequestSidebarItem.vue
+++ b/packages/api-client/src/views/Request/RequestSidebarItem.vue
@@ -124,7 +124,7 @@ const item = computed<SidebarItem>(() => {
       entity: request,
       resourceTitle: 'Request',
       warning: 'This cannot be undone. You’re about to delete the request.',
-      children: request.examples,
+      children: request.examples.slice(1),
       edit: (name: string) =>
         requestMutators.edit(request.uid, 'summary', name),
       delete: () => requestMutators.delete(request, props.parentUids[0]),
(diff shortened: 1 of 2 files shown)


amritk:
this one is for product

hanspagel (approved):
I think it’s great! The current way to show it is confusing to me, this seems more intuitive. @marclave wdyt?

marclave:
> I think it’s great! The current way to show it is confusing to me, this seems more intuitive. @marclave wdyt?

we are going to have a request page in the future, kind of like a collection page but for examples. so if we can eventually remove the filter in the future im down to have it here now :)

marclave (approved):

packages/api-client/src/views/Request/RequestSidebarItem.vue
```diff
@@ -124,7 +124,7 @@ const item = computed<SidebarItem>(() => {
       entity: request,
       resourceTitle: 'Request',
       warning: 'This cannot be undone. You’re about to delete the request.',
-      children: request.examples,
+      children: request.examples.slice(1),
```
marclave, line 127: wait doesnt slice return a shallow copy?

antlio: you right, should we avoid it?

marclave: most likely but i'd ping @geoffgscott or @amritk !

amritk (approved):
I think it should be fine, just test the reactivity

hanspagel:
It’s so reactive

PR #4081, opened 2024-12-04
feat(components): create ScalarTeleport component

hwkr:
Creates `ScalarTeleport` and `ScalarTeleportRoot` components which can be used to provide a scoped teleport root for floating elements and the like.

https://github.com/user-attachments/assets/03339b15-c4bd-470d-9d4a-1fd6af273e0b


packages/api-client/src/views/Request/components/WorkspaceDropdown.vue
```diff
@@ -117,9 +117,7 @@ const deleteWorkspace = async () => {
             <span class="text-ellipsis overflow-hidden">{{
               workspace.name
             }}</span>
-            <ScalarDropdown
-              placement="right-start"
-              teleport=".scalar-client">
+            <ScalarDropdown placement="right-start">
```
hwkr, line 120: This didn't need to be teleported as it turns out.

hanspagel (approved):
LGTM

antlio:
great work!

the method options style seems off, not sure if it's coming from the pr?
image

hwkr:
Going to say don't merge this for now while I investigate a few things 🙏

packages/api-client/src/views/Request/components/WorkspaceDropdown.vue
hwkr: Oops yep 🤦

amritk (approved):
Sweet seems to be working, lets get it in so we can finish that auth PR once and for all

packages/api-client/src/components/AddressBar/AddressBar.vue
```diff
@@ -157,7 +157,7 @@ onBeforeUnmount(() => events.hotKeys.off(handleHotKey))
               :isEditable="!isReadOnly"
               isSquare
               :method="activeRequest.method"
-              teleport=".scalar-client"
+              teleport
```
antlio, line 160: should we maintain a `scalar-client` class here to keep the style of http methods?

hanspagel (dismissed):
Just checked out the PR, I think we need to fix the request methods before we’ll get this wonderful PR in. :v:

Screenshot 2024-12-05 at 11 11 50

packages/api-client/src/components/AddressBar/AddressBar.vue
hwkr: Yep! Fixed 🫡

antlio:
i think it needs to be used as a parent otherwise `uppercase` won't be set. or should we set it at the scalar teleport components?

antlio (approved):
lets goo

hwkr:
> i think it needs to be used as a parent otherwise `uppercase` won't be set. or should we set it at the scalar teleport component level?

I don't think that will work because the parent's classes will just cascade onto the child. I'll just adjust the method options to be uppercase.

PR #4083, opened 2024-12-04
feat(api-client): use just the request path as a fallback for the sidebar

hanspagel:
We’re using `${method} - ${path}` as a fallback for the title of the request in the sidebar. But already show the request method, so using just the `request.path` should be good.

**Before**

Screenshot 2024-12-04 at 09 42 13

**After**

Screenshot 2024-12-04 at 09 42 48

### DIFF
--- a/.changeset/gentle-adults-grab.md
+++ b/.changeset/gentle-adults-grab.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
+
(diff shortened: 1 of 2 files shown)


antlio (approved):
great call!

marclave (approved):
i dig it ✨

amritk (approved):
yup

PR #4084, opened 2024-12-04
fix(proxy-server): error reading from static assets

hanspagel:
I’ve tried this already, but let’s try again: The proxy server got an API reference, but it’s not served in production:

```
Error reading index.html
```

This PR should finally add the static assets to the image.


marclave (approved):

amritk (approved):
sure why not, needs a changeset!

hanspagel:
proxy server doesn’t need a changeset, it’s just deployed on push to main and there’s no changelog.

We should maybe add a changelog, but that’s for another day.

PR #4087, opened 2024-12-04
ci: upgrade to ubuntu-22.04

hanspagel:
I just noticed we’re using `ubuntu-latest` (which will soon be bumped to `ubuntu-24.04`) and `ubuntu-20.04` in our workflows.

Let’s use `ubuntu-22.04` everywhere.

I tried `ubuntu-24.04`, but the Electron tests didn’t pass and I’m not ready to fix this.

I’d prefer the pinned version, so we can decide when to upgrade it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,7 +9,7 @@ on:
 jobs:
   harden_security:
     name: Check used GitHub Actions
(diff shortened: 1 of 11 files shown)


tmastrom (approved):
I love pinned versions 👍

PR #4092, opened 2024-12-04
fix(oas-utils): order of imported operations

### LINKED ISSUE #4088: Why i can't get proper sequence when i import JSON in Scalar desktop app
### Discussed in https://github.com/scalar/scalar/discussions/4076

<sup>Originally posted by **aditya05062002** December  3, 2024</sup>
Can some one help me why i get different sequence from original in scalar desktop app. In Scalar Document i get proper sequence like post, get, put, delete, search but in scalar desktop app i get sequence  like delete, get, post, put, get why that happening if i import same JSON into both of them 
Screenshot (81)
Screenshot (83)
</div>

antlio:
this pr fixes #4088 in order to sort method based on path VS prior method iteration as it was altering the operation order:

⊢ before / after for [galaxy](https://galaxy.scalar.com/)
<div>
image
image
</div>

### DIFF
--- a/packages/oas-utils/src/helpers/httpMethods.ts
+++ b/packages/oas-utils/src/helpers/httpMethods.ts
@@ -1,4 +1,4 @@
-import type { RequestMethod } from '@/entities/spec/requests'
+import { type RequestMethod, requestMethods } from '@/entities/spec/requests'
 
 /**
  * HTTP methods in a specific order
@@ -80,3 +80,7 @@ export const getHttpMethodInfo = (methodName: string) => {
     }
   )
 }
+
+/** Type guard which takes in a string and returns true if it is in fact an HTTPMethod */
+export const isHttpMethod = (method: string): method is RequestMethod =>
+  requestMethods.includes(method as RequestMethod)
--- a/packages/oas-utils/src/transforms/import-spec.ts
+++ b/packages/oas-utils/src/transforms/import-spec.ts
@@ -254,77 +254,79 @@ export async function importSpecToWorkspace(
     const pathServers = serverSchema.array().parse(path.servers ?? [])
     servers.push(...pathServers)
 
-    requestMethods.forEach((method) => {
-      const operation: OpenAPIV3_1.OperationObject<{
-        tags?: string[]
-        security?: OpenAPIV3_1.SecurityRequirementObject[]
-      }> = path[method as keyof typeof path]
-      if (operation && typeof operation === 'object') {
-        const operationServers = serverSchema
-          .array()
-          .parse(operation.servers ?? [])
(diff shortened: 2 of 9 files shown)


marclave (approved):

packages/oas-utils/src/transforms/import-spec.ts
```diff
@@ -254,77 +254,79 @@ export async function importSpecToWorkspace(
...
-          const keys = Object.keys(s)
-          if (keys.length) return keys[0]
-          else return []
-        })
-
-        let selectedSecuritySchemeUids: string[] = []
+    // Creates a sorted array of methods based on the path object.
+    const methods = Object.keys(path).filter((method) =>
+      requestMethods.includes(method as any),
+    )
```
amritk, line 260:
```suggestion
    const methods = Object.keys(path).filter(isHTTPMethod)
```
we have a type guard for this! you can just move it from the client to oas-utils

packages/oas-utils/src/transforms/import-spec.ts
```diff
@@ -254,77 +254,79 @@ export async function importSpecToWorkspace(
...
-              authentication.preferredSecurityScheme ?? '',
-            )
-              ? authentication.preferredSecurityScheme
-              : securityRequirements[0]
-          const uid = securitySchemeMap[name]
-          selectedSecuritySchemeUids = [uid]
-        }
+    methods.forEach((method) => {
+      const operation = path[method as keyof typeof path]
+      const operationServers = serverSchema
```
amritk, line 264:
```suggestion
      const operation = path[method]
      const operationServers = serverSchema
```
now methods should be typed :)

amritk (changes requested):
Just remove the any please!

packages/oas-utils/src/transforms/import-spec.ts
antlio: oooh right right! tysm

packages/oas-utils/src/helpers/httpMethods.ts
```diff
@@ -80,3 +80,7 @@ export const getHttpMethodInfo = (methodName: string) => {
     }
   )
 }
+
+/** Type guard which takes in a string and returns true if it is in fact an HTTPMethod */
+export const isHTTPMethod = (method: string): method is RequestMethod =>
```
hanspagel, line 85: code style police: pls use `isHttpMethod` (not `isHTTPMethod`)

hanspagel (changes requested):

packages/oas-utils/src/helpers/httpMethods.ts
antlio: great call

PR #4093, opened 2024-12-04
fix(api-client): search reactivity

antlio:
this pr fixes search results display on first character in reference and client: 

⊢ before / after
image
image


marclave (approved):
nice 😎

PR #4094, opened 2024-12-04
fix(api-reference): meta + k holding search modal

### LINKED ISSUE #2740: Holding cmd down after pressing cmd+k does not re-trigger the search box
### What happens?

Pressing cmd+k opens the docs, and if you let go of cmd, then pressing cmd+k again will close it. However if you press cmd+k again without letting go of cmd then the search box is not re-toggled. 

@hanspagel :) 

Spamming cmd+k (with or without lifting your finger from cmd) should repeatedly open and close the search box. I think it is confusing for users if you need to lift your finger from cmd/ctrl to re-trigger the hotkey since this is not usually the case with hotkeys. 

Here is an example from vscode where I don't lift my finger from the cmd key to toggle showing/hiding the integrated terminal:

https://github.com/user-attachments/assets/0d653d1a-00c3-498a-8c10-a4d9d9fcff6f

Go to the sandbox, and do the above steps. 

### OpenAPI Document

_No response_

antlio:
this pr fixes #2740 and updates `meta + k` command in api reference in order to trigger the command even while holding the meta key and re-pushing `k`.

### DIFF
--- a/.changeset/funny-hornets-arrive.md
+++ b/.changeset/funny-hornets-arrive.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
+fix: favors use event listener for meta k command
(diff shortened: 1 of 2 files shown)


hanspagel (approved):
Cracked

PR #4095, opened 2024-12-04
fix(api-client): address bar background

antlio:
this pr fixes the address bar background color that was not showing up anymore by safelisting tailwing background utilities:

⊢ before / after
image
image

### DIFF
--- a/packages/api-client/tailwind.config.ts
+++ b/packages/api-client/tailwind.config.ts
@@ -7,7 +7,12 @@ import { desktopVariants } from './src/tailwind/desktop-variants'
 
 export default {
   presets: [scalarPreset],
-  content: ['./index.html', './src/**/*.{js,jsx,ts,tsx,vue}'],
+  content: [
+    './index.html',
+    './src/**/*.{js,jsx,ts,tsx,vue}',
+    // Include oas-utils for http method colors
+    '../oas-utils/src/**/*.{js,ts,vue}',
+  ],
   corePlugins: { preflight: false },
   darkMode: ['selector', '.dark-mode'],
   theme: {
--- a/.changeset/honest-cooks-push.md
+++ b/.changeset/honest-cooks-push.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
+
+fix: updates tailwind config content


antlio:
this doesn't do

hwkr:
Could we just add them to this file? https://github.com/scalar/scalar/blob/main/packages/api-client/src/components/AddressBar/httpStatusCodeColors.ts

antlio:
> Could we just add them to this file? https://github.com/scalar/scalar/blob/main/packages/api-client/src/components/AddressBar/httpStatusCodeColors.ts

we could! but would the issue remain as we add utilities programatically?

hwkr:
> > Could we just add them to this file? https://github.com/scalar/scalar/blob/main/packages/api-client/src/components/AddressBar/httpStatusCodeColors.ts
> 
> we could! but would the issue remain as we add utilities programatically? however it's not necessarily the same topic here is it?

Tailwind decides what classes to include in the CSS by parsing all the Vue and TS files and doing a simple string search for anything that looks like it might be a Tailwind class. I think because the classes are being imported from OAS Utils that's why they aren't working.

We could also add it to the content list for the API Client but I think it's reasonable to just move it into the repo.

https://github.com/scalar/scalar/blob/main/packages/api-client/tailwind.config.ts#L10

Cursor-2024-12-05-06-55-07@2x

antlio:
@hwkr i updated the tw config content

hanspagel (approved):
Sick

packages/api-client/tailwind.config.ts
```diff
@@ -56,4 +56,9 @@ export default {
     },
   },
   plugins: [headlessPlugin, desktopVariants, colorMix()],
+  safelist: [
+    {
+      pattern: /bg-(red|green|blue|orange|yellow|purple|c-2)/,
```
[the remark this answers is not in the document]
antlio: this has been removed, favored instead the importation of the files to the config

packages/api-client/tailwind.config.ts
```diff
@@ -7,7 +7,11 @@ import { desktopVariants } from './src/tailwind/desktop-variants'
 
 export default {
   presets: [scalarPreset],
-  content: ['./index.html', './src/**/*.{js,jsx,ts,tsx,vue}'],
+  content: [
+    './index.html',
+    './src/**/*.{js,jsx,ts,tsx,vue}',
+    '../oas-utils/src/**/*.{js,ts,vue}',
```
hwkr, line 13: Could add a comment here like Hans mentioned :)

hwkr (approved):
Looks good!!

packages/api-client/tailwind.config.ts
antlio: ✅

PR #4102, opened 2024-12-05
fix(components): rename ScalaConfigurationSerializerContext to Scalar…

Jefwillems:
…ConfigurationSerializerContext

**Problem**
Currently, the dotnet package uses ScalaConfigurationSerializerContext

**Solution**

I believe this should be Scalar


xC0dex (approved):
Great spot! Thank you for the fix.

marclave (approved):

PR #4104, opened 2024-12-05
fix(api-client): variant danger style

antlio:
this pr fixes variant danger style in use that display filled background by removing unused utilities:

⊢ before / after
<div>
image
image
</div>


hwkr (approved):
Looks good!

We could probably apply this to the component itself at some point if we want. I'm guessing Cam always wants danger buttons to look like this more or less.

marclave (approved):

PR #4108, opened 2024-12-06
feat(components): create ScalarHeader component

hwkr:
Creates a standardized ScalarHeader component which can be used across Scalar apps. It also extracts some variables like header height (48px) into the core Scalar tailwind config.

This PR just updates the header in the API client to use the tailwind config but doesn't use the ScalarHeader component in the client yet because the clients navbar / header is still fairly bespoke and complicated.

The design is 90% the client header design (bottom border, background 2) which I'm thinking we should adopt for `docs.scalar.com` and `dashboard.scalar.com`. If we want them to be different though it's easy to override.

## Layout

The slot layout is designed to center the middle content relative to the overall header rather than preserving equal space between the sections (e.g. justify between). This is what we do for `docs.scalar.com`.

Google Chrome-2024-12-06-05-05-39@2x
Google Chrome-2024-12-06-05-05-44@2x

## Responsiveness

The slots are designed to preserve their content's size for as long as possible and then overflow the viewport. The idea being that a mobile menu should be swapped into the slots at small screen sizes.

https://github.com/user-attachments/assets/fc47dfc9-e2c6-41dd-aed6-a06fe5d817a3

## Buttons

The header also comes with a ScalarHeaderButton component which standardizes the button design. Once again it's 90% based on the API Client header button. This PR also adjusts the ScalarMenuButton to match the other header buttons at 32px high.

https://github.com/user-attachments/assets/520811a6-0d75-43ee-82bb-68caaeb10cde


hanspagel:
Sick work!

hwkr:
> Sick work! Can we use that for @scalar/api-reference, too? 👀

We could but we don't actually include a header there by default since it's custom themed based on the config and this is Scalar branded (in theory).

geoffgscott (approved):

PR #4115, opened 2024-12-06
style(api-client): request sidebar actions on hover

antlio:
this pr updates the style of the request sidebar item in api client in order to fix the button readability on long title:

disclaimer: moved the hover class higher to make sure the mask is on at the same time as the button getting displayed.

⊢ before / after
<div>
image
image
</dov>

### DIFF
--- a/.changeset/brave-birds-tap.md
+++ b/.changeset/brave-birds-tap.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
+
+style: updates request sidebar item mask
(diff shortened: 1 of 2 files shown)


marclave (dismissed):
chatted w @antlio and @cameronrohani we should do what notion does ya?

image
image

cameronrohani:
Yeah let's apply a mask on hover and not wrap the text please 🙏

antlio:
> Yeah let's apply a mask on hover and not wrap the text please 🙏

so keep it as it is?

hanspagel:
pls not wrap

make it one line

image

hanspagel (approved):
thanks for the update! let’s get this in.

PR #4130, opened 2024-12-09
chore: remove unused `examples/api-client` folder

hanspagel:
Found this unused folder that contains a Vite config only. Let’s remove this. ✨


tmastrom (approved):
sweep sweep 🧹

PR #4132, opened 2024-12-09
fix(fastify-api-reference): make the base route the reference

tmastrom:
Make the base route the reference. This is easier and makes more sense than having the base route return a 404 and needing to navigate to `/reference`


hanspagel (approved):
True!

PR #4137, opened 2024-12-09
feat(Scalar.AspNetCore): add layout option

### LINKED ISSUE #4126: Layout is not configurable in the dotnet package
### What happens?

The `layout` configuration option was not added in the `ScalarOptions` class, and there's no way to add custom configuration options (like an `AdditionalOptions` dictionary or something similar) [see ScalarOptions.cs](https://github.com/scalar/scalar/blob/main/packages/scalar.aspnetcore/src/Scalar.AspNetCore/Options/ScalarOptions.cs)

All supported configuration options should be available in the dotnet configuration

### OpenAPI Document

_No response_

xC0dex:
This PR adds the possibility to configure the Scalar layout in the `Scalar.AspNetCore` package.

Closes #4126

### DIFF
--- a/.changeset/popular-kiwis-cheat.md
+++ b/.changeset/popular-kiwis-cheat.md
@@ -1,0 +1,5 @@
+---
+'@scalar/aspnetcore': patch
+---
+
+feat: Add Layout option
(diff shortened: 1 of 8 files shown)


hanspagel:
Awesome contribution as always! Any chance you can add this to the README too? It might help more people to switch from what they had before.

marclave (approved):
:O nice work @xC0dex !!!

PR #4143, opened 2024-12-10
fix(api-client): tooltips have wrong z-index

antlio:
this pr updates the tooltip component in order to fix the tooltip visibility within client modal as seen below:

⊢ before / after
<div>
image
image
</div>

and also fix index issue in api client:

⊢ before / after
<div>
image
image
</div>


hanspagel (approved):
ncie

PR #4147, opened 2024-12-10
style(api-client): empty response action order

antlio:
this pr re-orders the empty response actions as seen below, nit/ocd but feels better:

⊢ before / after
<div>
image
image
</div>


hanspagel (approved):
Love it!

PR #4148, opened 2024-12-10
fix(api-client): watch mode toast

antlio:
this pr updates the watch mode toast watcher to prevent them from being displayed in editor/sandbox context:

image


hanspagel (changes requested):
cool!

hanspagel (approved):

Issue #3809, opened 2024-11-08
if requestBody is set to examples, content-type is missing in the code example.


uyrabi:
### What happens?

As the title says, it occurs in both curl and non-curl.

```sh:
## with examples

curl --request POST \
  --url https://sandbox.scalar.com/withExample \
  --data '{
  "email": "hoge@example.com",
  "password": "hogehoge"
}'
```

```sh:
## without examples

curl --request POST \
  --url https://sandbox.scalar.com/withOutExample \
  --header 'Content-Type: application/json' \
  --data '{
  "email": "marc@scalar.com",
  "password": "i-love-scalar"
}'
```

### What did you expect to happen?

It is correct that Content-type: application/json; is set in the request header

### OpenAPI Document

https://sandbox.scalar.com/e/ZM3A8

antlio:
hey @uyrabi, does that issue is still remaining? 

<img width="491" alt="image" src="https://github.com/user-attachments/assets/f4c811fd-e8bd-44f3-b24f-5466937c6283" />

uyrabi:
Hi @antlio 
I think it has been resolved. Thank you for your response.

PR #4150, opened 2024-12-10
fix(api-client): request body examples

antlio:
this pr adds back the request body border updated in #3981 + fixes broken selected example as seen in #3809

⊢ before / after
<div>
image
image
</div>


marclave (approved):
awesome :)

PR #4152, opened 2024-12-10
fix(api-reference): adds server variables in request example

### LINKED ISSUE #4151: Base url server variables
### What happens?

Server variables don't seem to be added to the example request in the api reference anymore. here an example with a `{path}` variable. introduced by #3981 

image

when empty:
image

with a value:
image

### OpenAPI Document

_No response_

antlio:
this pr fixes #4151 by updating request example hook to get server with variables being displayed.

⊢ before
image

⊢ after
when empty:
image

with a value:
image

### DIFF
--- a/.changeset/tasty-planets-brush.md
+++ b/.changeset/tasty-planets-brush.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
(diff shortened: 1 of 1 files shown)


marclave (approved):
🔥

PR #4154, opened 2024-12-10
fix(api-reference): circular dependency

hanspagel:
This PR fixes a circular dependency in the `<CompactSection />` component.

> Export "default" of module "src/components/Section/CompactSection.vue" was reexported through module "src/components/Section/index.ts" while both modules are dependencies of each other and will end up in different chunks by current Rollup settings. This scenario is not well supported at the moment as it will produce a circular dependency between chunks and will likely lead to broken execution order.


marclave (approved):
awesome catch! : )

Issue #3419, opened 2024-10-09
PATCH endpoint sending invalid HTTP Method in Web Client


dominicmallo:
### What happens?

When calling an endpoint with the `PATCH` method in the Web Client, `patch` (all lowercase) is sent in the network request. Our application uses NGINX, which treats `patch` calls as a `400 Bad Request` - expecting only `PATCH`.

This issue is not present with other HTTP Methods, like GET/POST/DELETE.

The issue was observed with `api-reference@1.25.24`

![image](https://github.com/user-attachments/assets/1307393d-80b8-489b-be58-a1309238b012)

### What did you expect to happen?

Sending requests with the `PATCH` method should result in network requests that have `PATCH` (all caps) for the request method.

dominicmallo:
This was resolved by upgrading to 1.25.32

PR #4159, opened 2024-12-11
fix(api-client): uppercase request method

Huliiiiii:
Fix regression caused by #3981.

**Problem**
Same issue as #3419.

**Solution**
Same solution as #3305.


marclave (approved):

marclave:
awesome thank you so much :))

PR #4162, opened 2024-12-11
feat(api-reference): adds example property in schema heading

### LINKED ISSUE #1971: Parameter details are missing example
### What happens?

My OpenAPI spec contains a parameter like the below

```
actor:
      name: actor
      in: query
      description: list of filters on actors, either exact actor IDs or a wildcard search on login name or display name indicated as `~search`
      required: false
      schema:
        type: array
        items:
          type: string
        example: [uc4p8fRHvJ11KTM59, ~bob]
```

The query parameters entry for that parameter does not show the example

image

If I render the same spec with Redocly, it does show the example.

image

1. Write an OpenAPI spec that includes a parameter with an example value
2. Render the Spec
3. See that the example is not shown

### OpenAPI Document

_No response_

antlio:
this pr fixes #1971 by adding example to the schema heading:

⊢ before / after
<div>
image
image
</div>

### DIFF
--- a/packages/api-reference/src/components/Content/Schema/SchemaProperty.vue
+++ b/packages/api-reference/src/components/Content/Schema/SchemaProperty.vue
@@ -107,6 +107,12 @@ const remainingEnumValues = computed(() =>
         #name>
         {{ name }}
       </template>
+      <template
+        v-if="value?.example"
+        #example>
+        Example:
+        {{ value.example }}
+      </template>
     </SchemaPropertyHeading>
     <!-- Description -->
     <div
@@ -362,7 +368,22 @@ const remainingEnumValues = computed(() =>
 }
 
 .property--compact .property-example {
-  display: none;
+  align-items: center;
+  background: transparent;
+  border: none;
+  display: flex;
+  flex-direction: row;
+  gap: 8px;
+}
+.property--compact .property-example-label,
+.property--compact .property-example-value {
+  padding: 0;
+}
+.property--compact .property-example-value {
+  background: var(--scalar-background-2);
+  border-top: 0;
+  border-radius: var(--scalar-radius);
+  padding: 2px 6px;
 }
 .property-list {
   border: var(--scalar-border-width) solid var(--scalar-border-color);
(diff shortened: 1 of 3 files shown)


hanspagel (approved):
Love the feature, but when I look at it, I think it could use some specific styling. @cameronrohani wdyt?

packages/api-reference/src/components/Content/Schema/SchemaProperty.vue
```diff
@@ -107,6 +107,12 @@ const remainingEnumValues = computed(() =>
         #name>
         {{ name }}
       </template>
+      <template
+        v-if="value?.example"
+        #example>
+        Example:
+        {{ value.example }}
```
hwkr, line 114: We should probably use the mono space font for the example, I think that's what we do in the models.

https://github.com/scalar/scalar/blob/main/packages/api-reference/src/components/Content/Schema/SchemaProperty.vue#L123-L149

Also it's worth double checking that multiline examples don't break the layout (I think multiline examples are allowed?).

antlio: I think instead we should use the following in compact too?
image

antlio: how do we feel instead about using the following for compact too:
image

antlio: cc @cameronrohani

amritk: @cameronrohani is this good to go?

antlio: i haven't the change to that yet, but can do - very quick thing to do

cameronrohani (approved):

packages/api-reference/src/components/Content/Schema/SchemaProperty.vue
antlio: @amritk i'll update accordingly and hyu!

amritk:
should we merge this @antlio ?

marclave (approved):
lets goooo @antlio

PR #4164, opened 2024-12-11
feat(oas-utils): request example parse enum array

### LINKED ISSUE #4142: Enum values don't flow through to API Client
### What happens?

Enum values are missing in when you open the api client.

In the API Docs...
Screenshot 2024-12-10 162634

In the API Client...
Screenshot 2024-12-10 162813

You can see the boolean correctly renders a select element to choose true/false, however the enums are not selectable and must be typed in each time.

The enum values should be selectable from a dropdown list. In this case with multiselect since it is an array value. 

For context, in SwaggerUI it looks something like this...

Screenshot 2024-12-10 162859

### OpenAPI Document

https://sandbox.scalar.com/e/xvuZh

antlio:
this pr fixes #4142 by update parse enum in request example entity to include array:

⊢ before / after
<div>
image
image
</div>

### DIFF
--- a/.changeset/silent-papayas-call.md
+++ b/.changeset/silent-papayas-call.md
@@ -1,0 +1,5 @@
+---
+'@scalar/oas-utils': patch
+---
+
+feat: updates parse enum to include array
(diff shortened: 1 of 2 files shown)


marclave (approved):

mahdiyari:
The resulted select box is not a multiselect. Shouldn't it allow you to select multiple items since it's an array?

antlio:
> The resulted select box is not a multiselect. Shouldn't it allow you to select multiple items since it's an array?

absolutely, opening an issue right now

PR #4172, opened 2024-12-12
feat(api-reference): extend font weights for links and sidebar items

cameronrohani:
add font weights sidebar items (active, inactive) in a css variable
add link hover colours in a variable
add link weights in a variable
add link text decoration colours in a variable

### DIFF
--- a/packages/api-reference/src/components/Sidebar/Sidebar.vue
+++ b/packages/api-reference/src/components/Sidebar/Sidebar.vue
@@ -183,6 +183,8 @@ onMounted(() => {
 <style scoped>
 .sidebar {
   --scalar-sidebar-indent-base: 12px;
+  --scalar-sidebar-font-weight-active: var(--scalar-semibold);
+  --scalar-sidebar-font-weight: var(--scalar-semibold);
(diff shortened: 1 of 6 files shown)


packages/api-reference/src/components/Sidebar/Sidebar.vue
```diff
@@ -183,6 +183,8 @@ onMounted(() => {
 <style scoped>
 .sidebar {
   --scalar-sidebar-indent-base: 12px;
+  --scalar-sidebar-font-weight-active: var(--scalar-semibold);
```
hwkr, line 186: We don't want to define these in `variables.css`? We already have a few other sidebar variables in there 

https://github.com/scalar/scalar/blob/main/packages/themes/src/variables.css#L87-L89

hwkr (approved):
Looks good!

hanspagel (approved):
LGTM, but I agree with @hwkr’s comment

hanspagel:
Cool, I think @hwkr’s comment was adressed. Let’s get this in!

PR #4173, opened 2024-12-12
fix(api-reference): correct anchor id encoding to ensure correct page scrolling

lc-soft:
**Problem**

The id attribute value of the `<section>` element is being encoded using encodeURIComponent(), for example:

image

When a document link with a hash is opened in a new tab, the browser attempts to find the element corresponding to the hash and scroll to it. This behavior is similar to the following code:

image

However, because the element cannot be found due to the encoding, the browser does not scroll to the intended section.

**Solution**

Remove `encodeURIComponent()` from [packages/api-reference/src/hooks/useNavState.ts](https://github.com/scalar/scalar/compare/main...lc-soft:scalar:lcsoft/fix-nav-id?expand=1#diff-ccdd228d6187612f62c06d9aaafbc4c0c9fa973ca37de2d80744df2b7dccdb43)

The id attribute value of the `<section>` element will then be set as follows:

image

With this change, the browser can correctly find and scroll to the intended element.

image

### DIFF
--- a/.changeset/spicy-islands-enjoy.md
+++ b/.changeset/spicy-islands-enjoy.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
+fix(api-reference): correct anchor id encoding to ensure correct page scrolling #4173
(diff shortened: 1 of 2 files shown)


hanspagel (approved):
OMG, great fix! Thanks for the explanation.

PR #4176, opened 2024-12-12
feat(api-reference): enum value visible item number increase

antlio:
this pr increases the number of enum item visible in the api reference:

image

_initially requested [in discord by @Arthur](https://discord.com/channels/1135330207960678410/1135330208627564597/1311623771710296074)_

### DIFF
--- a/.changeset/khaki-hairs-help.md
+++ b/.changeset/khaki-hairs-help.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
+feat: increases visible enum value item number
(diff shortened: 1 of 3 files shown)


hanspagel (changes requested):
No matter which limit we set, it might feel strange to not just show this one additional item.

Example: You’ve got 9 items and it shows as a list, but then you add one and we hide only one item, the 10th.

You know what we could do? We could say we show up to 9, but if it’s more than those 
we show the first 5 and hide all the remaining items. 

This might be better than just hiding this one additional item (in said example case), you’ll always have at least 5 hidden if some are hidden.

antlio:
> We could say we show up to 9, but if it’s more than those
> we show the first 5 and hide all the remaining items.

greta call let's go

hanspagel (approved):
Thanks for the update, this is great! I just can’t stand to have slice whatever in the template, so I’ve added a little bit to the PR.

arthurvanl:
thanks for implementing!

PR #4178, opened 2024-12-12
chore: make dependabot ignore some dependencies

hanspagel:
Some dependencies just require manual updates, I was not able to merge one of the dependabot PRs for them without fixing some type mismatches.


tmastrom (approved):
good idea. We don't want to just blindly update vite

PR #4182, opened 2024-12-12
fix(openapi-parser): rewrite parser

fuma-nama:
**Problem**
Currently, it uses `JSON.stringify` and requires multiple iteration.

**Solution**
With this PR, only one is needed:

The `resolve` function iterates over objects and resolve references, the resolved value will also be dereferenced to ensure nested references work. A `resolved` Set will act as a storage to detect if the object has been dereferenced, which replaces the need of `JSON.stringify` as self-referencing only happens with `$ref` in JSON schema.


packages/openapi-parser/src/utils/resolveReferences.ts
```diff
@@ -105,73 +95,79 @@ export function resolveReferences(
...
+      for (const key of Object.keys(target)) {
+        if (schema[key] === undefined) {
+          schema[key] = target[key]
         }
       }
+    }
 
-      if (typeof value === 'object' && !isCircular(value)) {
-        result = resolve(value, resolveFilesystem, resolveFile, onResolve)
+    // Iterate over the whole objecct
```
hanspagel, line 144:
```suggestion
    // Iterate over the whole object
```

hanspagel (changes requested):
OMG, this is sick!!

I think the Git history is off, though. It seems to delete the changes of your last PR.

Do you want me to fix it?

packages/openapi-parser/src/utils/resolveReferences.ts
fuma-nama: Oops I shouldn't code at night ;)

fuma-nama:
Oh I only moved that line to 132 and referenced it from `options` instead of adding parameter, it should continue to work

hanspagel:
Really great PR, some benchmarking results:

  | Test Case        | Before                                        | After                                         |
  | ---------------- | --------------------------------------------- | --------------------------------------------- |
  | No Reference     | 3.73x faster than @apidevtools/swagger-parser | 5.25x faster than @apidevtools/swagger-parser |
  | Petstore         | 1.90x slower than @scalar/openapi-parser      | 2.23x faster than @apidevtools/swagger-parser |
  | Single Reference | 1.19x faster than @apidevtools/swagger-parser | 2.34x faster than @apidevtools/swagger-parser |

hanspagel (approved):
PR of the month 🏆

PR #4183, opened 2024-12-12
feat(types): generate documentation

hanspagel:
This PR is an attempt to generate the documentation for specific types automatically.

I considered to use existing tools, but the output was really overwhelming. With the script here we can make it really as simple as its now, but just auto-generate it from the types.

**Tasks**

- [ ] merge comments
- [ ] tweak the output
- [ ] replace existing Markdown file

**Usage**

```bash
cd packages/types
pnpm run docs:dev
```

**Preview**

Screenshot 2024-12-12 at 13 08 28


hanspagel:
I’m not happy with the solution, needs a lot more work and I’m not ready to deal with that now. Keeping the branch, but closing the PR.

PR #4184, opened 2024-12-12
feat(api-reference): Adds optional hash prefixing

geoffgscott:
We have no ability to use hashes in a consuming app due to API Reference force overwriting them.

This allows an integration to a set a hash prefix that will be ignored by the references.


hanspagel (approved):
code looks good, tested the original functionality (not the added feature) and it continues to work with that PR

I assume we’re testing the new functionality in Scalar Docs

marclave (approved):

Issue #1790, opened 2024-05-21
The wildcard notation for mediaType don t support (e.g. */* or application/*)


hahaland:
### What happens?

I've noticed that the project doesn't seem to support the wildcard notation for mediaType in responses (e.g. */* or application/*). Using this notation prevents any return value types from being displayed on the page.

However, after consulting the OpenAPI 3.0 specification , it includes this rule https://swagger.io/docs/specification/media-types/).

Is it possible to fix this issue in a future update? I would greatly appreciate it.(By the way, great tool!!)

### What did you expect to happen?

When using the wildcard notation for mediaType，the responses definition can be displayed normally

### How can we reproduce the issue?

1. Select any interface definition that has a response value
2. change media type to '*/*' or 'application/*'

### Swagger/OpenAPI Example

https://sandbox.scalar.com/e/Lu1Af

hanspagel:
Great feedback! We definitely need to fix that.

antlio:
will resume work on this once #4281 is in as it might impact it.

DemonHa:
This seams to be working now!

<img width="558" height="660" alt="Image" src="https://github.com/user-attachments/assets/66d4446a-11ba-407b-a8b3-6b6e6db43719" />

PR #4185, opened 2024-12-12
feat(api-reference): support wildcard notation for mediatype

antlio:
this pr adds first stepping stones into fixing #1790 by introducing `*/*` and `application/*` content type support to the example responses.

course of action:
- [ ] content header updates
- [ ] request


hanspagel:
we need an empty PR queue to apply the formatting. we can pick this up again in the future and re-apply the changes, it’ll be easier. ✌️

PR #4196, opened 2024-12-13
ci: read from cache for subsequent builds

hanspagel:
Not related to the current issues, but just found out you can control whether turborepo writes the cache. This feature was introduced in turbo@2.2.4, that’s why updated Turborepo, too.

With this PR subsequent builds read from the local cache (that was restored by GitHub Actions), without writing to it.

https://turbo.build/repo/docs/reference/run#--cache-options

EDIT: While this feels cleaner, I can’t see any performance benefits. The whole action takes ~50s and turborepo say the build takes ~700ms with and without this PR.

### DIFF
--- a/.github/actions/build/action.yml
+++ b/.github/actions/build/action.yml
@@ -9,6 +9,10 @@ inputs:
     description: 'Whether to build packages/ only'
     required: false
     default: 'false'
+  write-cache:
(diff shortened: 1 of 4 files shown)


amritk (approved):
I'm game to see it in action

tmastrom (approved):

PR #4197, opened 2024-12-13
fix(api-client): address bar servers dropdown display

antlio:
this pr fixes the address bar servers component being display even on active collection without servers as we were fallbacking on activeServer anyway :pedro-laugh-cry: . I also propose in that PR to get the dropdown displayed for unique server in order to get access easily to the `Add Server` button.

⊢ draft request before / after
<div>
image
image
</div>

⊢ request with unique server before / after
<div>
image
image
</div>

### DIFF
--- a/.changeset/new-falcons-enjoy.md
+++ b/.changeset/new-falcons-enjoy.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
+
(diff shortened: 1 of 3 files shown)


marclave (approved):

amritk (approved):
LGTM!

PR #4202, opened 2024-12-13
chore: use composite build action instead of pnpm script

tmastrom:
With this PR, workflows use the composite build action for building packages.


hanspagel (approved):
Yess! Thanks

PR #4207, opened 2024-12-13
fix: content `null` throws an error

hanspagel:
When passing `null` (or an empty string) as the content, errors are thrown.

Screenshot 2024-12-13 at 22 03 56

This PR fixes it in multiple places to improve the handling for various use cases.

**@scalar/openapi-parser**

* `normalize()` returns `{}` for empty OpenAPI documents
* added tests for the `normalize()` utility, can’t believe we didn’t had any
* `upgrade()` (and the specific upgrade helpers) deal with empty content
* `details()` utility handles empty content
* replaced some uses of `AnyObject` with `UnknownObject` and fixed TS issues

**@scalar/oas-utils**

* skip the parser for empty content, add a console.warning

**@scalar/types**

* explicitly allow `null` in the `SpecConfiguration`

**@scalar/api-reference**

* Slightly improve the types in the horrible `parse` helper

**@scalar/nuxt**

* import `ReferenceConfig` from `@scalar/types` instead of `@scalar/api-reference`

```diff
-  specification: any,
+  specification: UnknownObject | string | undefined,
```

### DIFF
--- a/.changeset/brave-grapes-march.md
+++ b/.changeset/brave-grapes-march.md
@@ -1,0 +1,5 @@
+---
+'@scalar/types': patch
+---
+
+chore: improve the comment for UnknownObject
--- a/.changeset/silent-dingos-hide.md
+++ b/.changeset/silent-dingos-hide.md
@@ -1,5 +1,0 @@
----
-'@scalar/openapi-parser': patch
----
-
-feat: better deal with empty OpenAPI documents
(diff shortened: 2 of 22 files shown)


hwkr (approved):
Looks great! Tests are 🔥 🙏

marclave (approved):

marclave (approved):

PR #4209, opened 2024-12-13
chore: make repository url full git url

hanspagel:
> GitHub or GitLab links should be prefixed with `git+` and postfixed with `.git`. (This is also warned by npm when publishing a package).

https://publint.dev/rules#invalid_repository_value


hwkr (approved):
🙏

PR #4215, opened 2024-12-15
refactor(Scalar.AspNetCore): improved API and endpoints

### LINKED ISSUE #3901: Cannot map multiple endpoints
### What happens?

Using the ASP.NET Core package I generate two OpenAPI documents as my API has two clients with different authentication schemes and wanted to map both to separate endpoints. 

```
            {
                options.WithEndpointPrefix("/scalar/admin/{documentName}");
                options.WithOpenApiRoutePattern("/openapi/admin.json");
                options.WithTitle("Admin");
                options.WithOAuth2Authentication(authOptions =>
                {
                    authOptions.ClientId = app.Configuration["Admin:ClientId"];
                    authOptions.Scopes = [app.Configuration["Admin:Scope"]!];
                });
            })
            .AllowAnonymous();

            {
                options.WithEndpointPrefix("/scalar/partner/{documentName}");
                options.WithOpenApiRoutePattern("/openapi/partner.json");
                options.WithTitle("Partner");
                options.WithOAuth2Authentication(options =>
                {
                    options.ClientId = app.Configuration["Partner:ClientId"];
                    options.Scopes = [app.Configuration["Partner:Scope"]!];
                });
            })
            .AllowAnonymous();
```

However, the MapScalarApiReference using GetService<IOptions<ScalarOptions>>, which will return the same reference for both calls by default, so both endpoints ends up using the same options object with the values from the second configuration of the options.

I expected be able to configure the two endpoints independently.

I worked around this by forcing the options to be null:

builder.Services.AddTransient<IOptions<ScalarOptions>>(sp => null!);

But I think perhaps it would be nice with a more obvious way to do this. 

### OpenAPI Document

_No response_

### LINKED ISSUE #4032: Inject service or action when mapGet Scalar.AspNetCore
### What happens?

I tried to add a token from the HTTP context of the request, but I couldn't because the configuration is serialized before the MapGet endpoint. As a result, I can't inject or perform any actions when a request comes to the server.

I would like to inject or configure the options at runtime

### OpenAPI Document

_No response_

### LINKED ISSUE #4284: UI is pointing to the wrong spec file when API is loaded via reverse proxy
### What happens?

I have a reverse proxy which forwards the request to my backend APIs. I can open the json spec fine, however, the UI is unable to load because it is pointing to the wrong json path.

The proxy logs clearly forwards it to the right path
image

However, when the UI loads, it seems to be ignoring the service prefix.
image

The json path works when accessed thru the proxy or directly from the API.

**Additional information**
- Scalar.AspNetCore 1.2.63
- .NET 8
- Reverse proxy - YARP

UI should be loaded for APIs sitting behind a proxy.

### OpenAPI Document

_No response_

xC0dex:
Please let me know what you think. ~We have to wait until we support multiple OpenAPI documents.~

This PR introduces a couple of new features and several enhancements.

**Preparation: Support for Multiple OpenAPI Documents:**

> [!Note]
> I prepared APIs and methods for this feature but they are `internal`, so no one can use it yet. We have to wait until the API reference supports multiple OpenAPI documents. But once this is the case, we basically only have to make the methods public.

Previous Behavior: Only one OpenAPI Document was supported. The name of the OpenAPI document was part of the route  (e.g., `v1` ⇾ `/scalar/v1`).
Now: Some new APIs in the `ScalarOptions` and a redesign of the `MapScalarApiReference` method allows adding multiple OpenAPI documents through the configuration. The default route changed from `/scalar/{documentName}` to `/scalar/`. The document name is no longer part of the route by default. However, it's still possible to provide the document name in the route. This allows backwards compatibility and also overwriting the documents provided via configuration.

This is the basic usage of the new configuration API:
```c#
{
    options.AddDocument("v1", "v2");
});
```

It's also possible to implement a `DocumentNamesProvider`. This can be used to automagically fetch document names:

```c#
app.MapScalarApiReference((options, context) =>
{
    // Automagically add the OpenAPI document names based on Swashbuckle config
    var swaggerOptions = context.RequestServices.GetRequiredService<IOptions<SwaggerGeneratorOptions>>().Value;
    var documentNames = swaggerOptions.SwaggerDocs.Keys;

    options.DocumentNamesProvider = _ => documentNames;
});
```

If no document provider or document name is provided, a default document name `v1` will be added.

Due to the redesign of `MapScalarApiReference` it's now also possible to map the endpoint to the root of the API by using the new `endpointPrefix` parameter.

```c#
app.MapScalarApiReference("/");
```

**Multiple endpoints:**
Previous Behavior: Only one `MapScalarApiReference` was working as expected.
Now: It's possible to map multiple API reference endpoints, distinguished by a different `endpointPrefix`. They have scoped options so they won't conflict with other options.

```c#
app.MapScalarApiReference("/internal");
app.MapScalarApiReference("/public");
```

Later, it's also possible to add documents through the configuration (`AddDocument` is currently internal):

```c#
app.MapScalarApiReference("/internal", options => options.AddDocument("internal"));
app.MapScalarApiReference("/public", options => options.AddDocument("public"));
```

**Sub Path deployment (Reverse Proxy):**
Previous Behavior: Deploying the API with the Scalar API reference under a sub path (e.g., `/api/` → `/api/scalar/v1`) required workarounds like manually updating the `OpenApiRoutePattern` property.
Now: Developers no longer need to perform additional configurations; Sub Path deployment works seamlessly out of the box. Some JS logic determines the actual sub path and adds it to the OpenAPI document paths. 

**Custom Head and Header content:**
Previous Behavior: It was not possible to add custom HTML elements to response.
Now: It's possible to add custom content that will be added to the end of the `<head>` section and content that will be added right after the `<body>` but before the API reference.

**Addition changes:** 

- The `HttpContext` is now injected during options configuration, enabling developers to access all necessary elements from the `HttpContext`.
- Auto redirect from `/scalar` ⇾ `scalar/` to resolve relative paths correctly.
- Cache and ETag Header for static assets.
- The `EndpointPathPrefix` parameter is now marked as obsolete. A new `endpointPrefix` parameter has been introduced, decoupling the route from the options and allowing for individual configurations for each endpoint.
- More `[StringSyntax]` annotations for a better developer experience.
- Fixed a typo in `Metadata` ⇾ `MetaData`. Now the configuration works as expected.

**Breaking changes:**
- The  `EndpointPathPrefix` property of `ScalarOptions` is now obsolete. It still works when using the `MapScalarApiReference` methods without the `HttpContext` or `endpointPrefix` overloads, but developers should switch to the new `endpointPrefix` parameter.
- The workaround for the sub path deployment no longer works.

I wrote a migration guide for the breaking changes: https://github.com/scalar/scalar/issues/4362

Closes #4032
Closes #3901
Closes #4284


hanspagel:
Love the changes! And I think it’s worth to do a new major release!

We’re about to add support for multiple API defintions in the next weeks (never trust a timeline in open-source), should we wait for this to arrive? Or do you prefer the new syntax in .NET anyway? Or would we just map to whatever the `@scalar/api-reference` config looks like then?

xC0dex:
> We’re about to add support for multiple API defintions in the next weeks

That's literally **awesome**! I would definitely wait for this because this would introduce another breaking change (I guess).

hirenvadher954:
@xC0dex  First of all, thank you for making these changes. They’re amazing, and I’m planning to use it soon. This is why we’ve been sticking with Swagger, but with these updates, we can fully switch to it. Once the changes are finalized, I’d like to contribute specifically to the documentation. I encountered some challenges implementing the OAuth flow with Azure B2C due to limited documentation, so I’d love to help improve it. Once again, thank you for bringing these updates

xC0dex:
This PR is **finally** ready for review. It would be wonderful if you leave comments if something is unclear because this is a new major release. 🚀

hanspagel:
Not reviewed yet, just want to say this is a banger PR.

hwkr:
+1 for banger PR 🙌

hanspagel (approved):
OMG, looked through it and it’s awesome. Can’t wait for this to be released.

hanspagel:
The docker setup didn’t work for me. I think I fixed it, can you try whether it still works for you?

marclave (approved):

xC0dex:
> I think I fixed it, can you try whether it still works for you?

It still works!

hanspagel:
Let’s get this banger in! 👏

PR #4218, opened 2024-12-16
chore(express-api-reference): move example to package playground

tmastrom:
This PR moves the `express-api-reference` example app to the package playground folder.


hanspagel (approved):
Awesome work! The playground didn’t show any endpoints, I’ve fixed it and did some clean up work for the package. I‘ve also refactored the tests. 😎

hanspagel (changes requested):
Oh, wait. There is `'./examples/express-api-reference/Dockerfile',` in the cloudbuild.yaml, can you update this?

packages/express-api-reference/playground/src/index.ts
```diff
@@ -0,0 +1,146 @@
...
+    openapi: '3.1.0',
+    info: {
+      title: 'Express Example',
+      description:
+        'The `@scalar/express-api-reference` middleware renders a beautiful API reference based on your OpenAPI specification.',
+      version: '1.0.0',
+    },
+  },
+  // Update the path to include the current file
+  apis: [new URL(import.meta.url).pathname],
```
tmastrom, line 122: It's important we don't hardcode paths because they won't work in Docker and when served at subpaths

hanspagel (approved):

PR #4219, opened 2024-12-16
chore(fastify-api-reference): move example to package playground 

tmastrom:
This PR moves the example app to the package playground


hanspagel:
We used to keep the CJS build for the Fastify community, and our build tooling isn’t providing CJS builds.

I think we should switch to our build setup, drop CJS, modify the tests and when someone complains try to add optional CJS support to the build tooling.

PR #4229, opened 2024-12-16
fix(api-reference): SSG build is broken

hanspagel:
The SSG build is broken, let’s fix it!

## Tasks

- [x] SSG example needs to be updated
- [x] ApiReference/Layout component tree: localStorage not defined
- [x] useColorMode: ReferenceError: window is not defined
- [x] RenderedReference export can be removed
- [x] [ERROR] ReferenceError: document is not defined???
- [x] SSG checks for @scalar/use-hooks?
- [x] Error: Could not parse CSS stylesheet

## Reproduction

```bash
cd examples/ssg
pnpm build
```
## Error: Could not parse CSS stylesheet

**Source**

https://github.com/scalar/scalar/blob/main/packages/api-reference/src/components/ApiReferenceLayout.vue#L321-L329

**Error**

```
Error: Could not parse CSS stylesheet
    at exports.createStylesheet 

…

  @layer scalar-theme {
/* basic theme */
.light-mode {
  --scalar-background-1: #fff;
  --scalar-background-2: #f6f6f6;
  --scalar-background-3: #e7e7e7;
  --scalar-background-accent: #8ab4f81f;

  --scalar-color-1: #2a2f45;
  --scalar-color-2: #757575;
  --scalar-color-3: #8e8e8e;

  --scalar-color-accent: #0099ff;
  --scalar-border-color: #dfdfdf;
}
```

### DIFF
--- a/.changeset/shaggy-moons-accept.md
+++ b/.changeset/shaggy-moons-accept.md
@@ -1,0 +1,7 @@
+---
+'@scalar/api-reference': patch
+'@scalar/api-client': patch
+'@scalar/use-hooks': patch
+---
+
+feat: make it work in SSG environments
--- a/.changeset/twelve-ducks-matter.md
+++ b/.changeset/twelve-ducks-matter.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
+chore: remove RenderedReference export
--- a/examples/ssg/src/App.vue
+++ b/examples/ssg/src/App.vue
@@ -1,5 +1,6 @@
 <script setup lang="ts">
 import { ApiReference } from '@scalar/api-reference'
+import '@scalar/api-reference/style.css'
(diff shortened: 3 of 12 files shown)


amritk:
Alright fixed the document is undefined, it was the searchButton hotkeys, lets make sure it still works. 

I just used a regular event listener BUT we should be able to start using the event bus in references now! (didn't want to do it here though)

amritk:
The css issue has to do with this package [rrweb-cssom](https://www.npmjs.com/package/rrweb-cssom)

@layer doesn't work.

I found the same issue in JSDom:
https://github.com/jsdom/jsdom/issues/3785

Until they update, we can toss in an override

hanspagel:
Great work, @amritk! Thanks for chiming in. Let’s get this in ASAP.

PR #4235, opened 2024-12-16
fix(api-client): request params clear action

antlio:
this pr updates the request params clear action in the api client in order to allow clearance on enabled items, however required items will remain + it adds a tooltip information as seen below:

image

### DIFF
--- a/packages/api-client/src/views/Request/RequestSection/RequestParams.vue
+++ b/packages/api-client/src/views/Request/RequestSection/RequestParams.vue
@@ -3,7 +3,7 @@ import ViewLayoutCollapse from '@/components/ViewLayout/ViewLayoutCollapse.vue'
 import { useWorkspace } from '@/store'
 import { useActiveEntities } from '@/store/active-entities'
 import RequestTable from '@/views/Request/RequestSection/RequestTable.vue'
-import { ScalarButton } from '@scalar/components'
+import { ScalarButton, ScalarTooltip } from '@scalar/components'
 import {
   type RequestExample,
   requestExampleParametersSchema,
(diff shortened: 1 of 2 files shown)


packages/api-client/src/views/Request/RequestSection/RequestParams.vue
```diff
@@ -163,13 +161,28 @@ watch(
...
+              @click.stop="deleteAllRows">
+              Clear
+              <span class="sr-only">All {{ title }}</span>
+            </ScalarButton>
+          </template>
+          <template #content>
+            <div
+              class="grid gap-1.5 pointer-events-none min-w-48 w-content shadow-lg rounded bg-b-1 z-context p-2 text-xxs leading-5 z-10 text-c-1">
+              <div class="flex items-center text-c-2">
+                <span>Clear all non required parameters</span>
```
hanspagel, line 181:
```suggestion
                <span>Clear optional parameters</span>
```

hanspagel (changes requested):
What if we delete the required parameters, if there are only required parameters in the list? 🤔 

Two scenarios:

1) There are only required parameters and I click on clear, I might expect to actually clear the required parameters.

2) There are required and optional parameters, it might be good to only delete the optional parameters. And if someone wants to clear the remaining parameters a second click will delete the remaining parameters.

antlio:
> What if we delete the required parameters, if there are only required parameters in the list? 🤔
> 
> Two scenarios:
> 
> 1. There are only required parameters and I click on clear, I might expect to actually clear the required parameters.
> 2. There are required and optional parameters, it might be good to only delete the optional parameters. And if someone wants to clear the remaining parameters a second click will delete the remaining parameters.

Like I said, the clear action removed all parameters except the required ones - today we only clear the disabled ones

packages/api-client/src/views/Request/RequestSection/RequestParams.vue
antlio: way better!

antlio:
> 1. There are only required parameters and I click on clear, I might expect to actually clear the required parameters.

should we omit the display in that case?

hanspagel:
Ah, yes! Got it. Yeah, we can hide the Clear button then?

antlio:
> Ah, yes! Got it. Yeah, we can hide the Clear button then?

added a function to display clear button only if non required param with key or value are present

hanspagel (approved):
Sick

PR #4236, opened 2024-12-16
fix: path params disapearing

amritk:
This line seems to be deleting path params, @antlio what was it for?

Before:
image

After:
image


marclave (approved):

antlio:
great catch thank you!

the initial intention here was to manage variables being added/removed in the path, in order to get the variables table to show up or to be hidden accordingly.

however this doesn't seem to work anymore.

antlio:
it doesn't seem to fix it, will give it another look

PR #4237, opened 2024-12-16
feat(oas-utils): add isDefined type safe filter helper

amritk:
We will need this or something similar in the tsconfig fix PR


packages/oas-utils/src/helpers/is-defined.ts
```diff
@@ -0,0 +1,14 @@
...
+ *
+ * ```ts
+ * const dataArray = [1, null, 2, undefined, 3].filter(isDefined)
+ * ```
+ *
+ * @see https://jaketrent.com/post/typescript-type-safe-filter-boolean/
+ */
+export const isDefined = <T>(
+  value: T | null | undefined,
+): value is NonNullable<T> => value !== null && value !== undefined
```
hwkr, line 14: Should it not be 
```suggestion
): value is NonNullable<T> => value !== null && typeof value !== 'undefined'
```

?

marclave (approved):

packages/oas-utils/src/helpers/is-defined.ts
amritk: I think that would be the same thing?

amritk: claude agreed lol

image

Issue #4214, opened 2024-12-15
Body Json input is not visible


tim-deighton:
### What happens?

When I try to submit a POST request I cannot enter any test data because the Body/JSON input is not visible. When I 'Inspect' I notice that the CSS is Display: None.  If I disable the selector in the inspector all works as expected. I can enter data and see the response.

### What did you expect to happen?

I expected the Body input box to be visible when I choose to test a POST request.

antlio:
thank you for reporting this @tim-deighton! is it from creating a new request? do you have any screenshot or element so I can reproduce on my end? tysm!

tim-deighton:
![Screenshot 2024-12-15 154507](https://github.com/user-attachments/assets/464c831a-10bb-40eb-a666-1c283af6de72)
![Screenshot 2024-12-15 154422](https://github.com/user-attachments/assets/440f602a-559a-4e4e-a78c-80fe6c811e68)
![Screenshot 2024-12-15 154218](https://github.com/user-attachments/assets/0bbfbc99-7c22-43a2-ad0a-3db00e6870ee)

Thanks guys!!

antlio:
@infureal do you mind trying with you document in [our sandbox](https://sandbox.scalar.com/) and tell me if the issue remains? trying to replicate on my end

antlio:
> @antlio do you replicate the issue?

can you refresh, you should be able to have:
<img width="1067" alt="image" src="https://github.com/user-attachments/assets/6d2748c9-1f02-4b0c-95b9-3cbd4be0bc5b" />

however do you have any issue regarding the body input for your post endpoints?

antlio:
oh i see, as of right now we only display the body input [for the following methods](https://github.com/scalar/scalar/blob/8adfeea59607190c2e0d5159feb41b2801775389/packages/oas-utils/src/helpers/httpMethods.ts#L62) - any reason to have it displayed also for a get request? 

we took that path initially to avoid unnecessary element being display however we might reconsider it!

antlio:
@tim-deighton do you mind sharing some example of your document in [sandbox](https://sandbox.scalar.com/)? also what package are you using for the reference? (sorry all of that is for testing purposes to make sure the issue is fixe :))

tim-deighton:
Thanks for fixing Guys AMAZING job. All working now in latest release.

PR #4248, opened 2024-12-17
fix(api-client): adds request body input safety check

antlio:
this pr is part of #4214 in order to ensure the request body being displayed.

image

### DIFF
--- a/.changeset/two-kangaroos-live.md
+++ b/.changeset/two-kangaroos-live.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
+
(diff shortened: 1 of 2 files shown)


hanspagel:
This is great! But why did the previous check fail?

antlio:
> Quote reply

i don't find it
image

hanspagel:
I mean, why was `v-show="activeRequest …`  not enough?

antlio:
> I mean, why was `v-show="activeRequest …` not enough?

i'don't know i haven't been able to replicate that issue yet, this is preventing some logging error from poppin

PR #4252, opened 2024-12-17
fix(openapi-parser): correct schema upgrade for "format: binary"

lc-soft:
**Problem**
Currently, if the schema contains `format: binary`, it will be converted to undefined, resulting in the following error and the API client window will be blank.

readOnly-undefined-error

schema-undefined-debug

The content of the schema YAML file is as follows:

openapi-schema

**Solution**

With this PR, I referred to the code in the upgradeFromThreeToThreeOne.ts file and changed the upgrade result of `format: binary`.

image

### DIFF
--- a/.changeset/slow-spies-roll.md
+++ b/.changeset/slow-spies-roll.md
@@ -1,0 +1,5 @@
+---
+'@scalar/openapi-parser': patch
+---
+
+fix(openapi-parser): correct schema upgrade for "format: binary"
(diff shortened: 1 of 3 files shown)


hanspagel:
Added a bit to it:

**Binary File Uploads**

The test checks that the schema is removed for binary file uploads in application/octet-stream.

**Base64 Encoded Images**

The test verifies that the format: base64 is replaced with contentEncoding: base64 for image uploads.

**Multipart File Uploads**

The test ensures that the format: binary is replaced with contentMediaType: application/octet-stream for multipart file uploads.

Source: https://www.openapis.org/blog/2021/02/16/migrating-from-openapi-3-0-to-3-1-0

hanspagel (approved):

PR #4255, opened 2024-12-17
ci: no concurrent publish

hanspagel:
I think the recent issues we faced with publishing were related to having the `changesets/action` running in parallel doing two different things. 🤔

Anyway, publishing shouldn’t be executed in parallel and existing jobs shouldn’t be canceled, so let’s add this to the CI workflow.

image

EDIT: #4261 should make CI green

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -207,6 +207,9 @@ jobs:
     if: github.ref == 'refs/heads/main'
     runs-on: ubuntu-22.04
     timeout-minutes: 10
+    # Avoid running this job in parallel:
+    # `changesets/action` creates/updates the release branch, which shouldn’t happen in parallel.
+    # npm publish also shouldn’t happen in parallel.
     concurrency:
       group: npm-publish
       cancel-in-progress: false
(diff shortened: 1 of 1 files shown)


tmastrom (commented):
Should these be named the same concurrency group like `npm publish` or are we trying to stop issues between multiple runs of the same workflow?

hanspagel:
Yes! We can publish to multiple registries in parallel, but we can’t publish to the npm registry in parallel. Or actually it’s more about not running the changeset/action in parallel, which creates/updates the release branch (= trouble when running in parallel).

tmastrom:
> Yes! We can publish to multiple registries in parallel, but we can’t publish to the npm registry in parallel. Or actually it’s more about not running the changeset/action in parallel, which creates/updates the release branch (= trouble when running in parallel).

This sounds good to me 👍

.github/workflows/ci.yml
```diff
@@ -207,6 +207,9 @@ jobs:
     if: github.ref == 'refs/heads/main'
     runs-on: ubuntu-22.04
     timeout-minutes: 10
+    concurrency:
```
geoffgscott, line 210: Some more comments here will be helpful for the future viewers.

amritk (approved):
Seems reasonable

PR #4258, opened 2024-12-17
chore: updates CodeMirror dependencies

### LINKED ISSUE #4222: CommonJS named export issue for @uiw/codemirror-themes packages.
### What happens?

When building my docs site via a CI/CD pipeline, I'm encountering the following error for an incompatibility between ESM / CommonJS:

```
Named export 'createTheme' not found. The requested module '@uiw/codemirror-themes' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

import pkg from '@uiw/codemirror-themes';
const { createTheme: r } = pkg;
```

Notably, I haven't bumped my scalar version so not sure why the sudden incompatibility on builds. Currently using  `"@scalar/api-reference-react": "^0.3.154",`

Able to build scalar properly.

### OpenAPI Document

_No response_

hanspagel:
This PR updates `@scalar/use-codemirror` to 

* use the latest dependencies, 
* fix #4222,
* add tests,
* deletes the unused CodeMirror component,
* switches to rollup for the build,
* add a playground.

This is also an attempt to fix CI, because it’s running into issues when building `@scalar/use-codemirror` again and again.

EDIT: I think #4259 fixed the CI issue. I think we could still upgrade the dependencies, but feel free to reject the PR.

Screenshot 2024-12-17 at 15 38 11


packages/use-codemirror/src/hooks/useCodeMirror.ts
hanspagel, line 1: I’ve had to change the order to write tests.

packages/use-codemirror/src/index.ts
```diff
@@ -11,7 +11,5 @@ export {
   WidgetType,
 } from '@codemirror/view'
 
-export { CodeMirror } from './components/CodeMirror'
-export { useDropdown } from './hooks/useDropdown'
```
hanspagel, line 15: exported through hooks/index.ts now

geoffgscott (approved):
Looks good other than the 1 comment

amritk (commented):
I tried to test if this fixed the nuxt error, unfortunately it did not. BUT it could also be a linking error. 

image

geoffgscott:
> I tried to test if this fixed the nuxt error, unfortunately it did not. BUT it could also be a linking error.

Yeah i had to pin the version to the previous patch in org to make it work

PR #4261, opened 2024-12-17
ci: do not cache playwright binaries

hanspagel:
We’ve added a check whether the playwright dependencies (browsers) were restored from cache, but that check failed for Playwright version updates (for whatever reason) - and with it all tests.

https://github.com/scalar/scalar/actions/runs/12375837584/job/34541570648?pr=4255

It turns out, we don’t need to cache Playwright binaries.

> Caching browser binaries is not recommended, since the amount of time it takes to restore the cache is comparable to the time it takes to download the binaries.

Source: https://playwright.dev/docs/ci#caching-browsers

### DIFF
--- a/.github/workflows/test-cdn-jsdelivr.yml
+++ b/.github/workflows/test-cdn-jsdelivr.yml
@@ -29,15 +29,7 @@ jobs:
         with:
           packages-only: true
           node-version: ${{ matrix.node-version }}
-      - name: Get Playwright version
-        id: playwright-version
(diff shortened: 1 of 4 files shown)


tmastrom (commented):
Where does it say that playwright handles restored caching? I found this saying they don't recommend caching the binaries https://playwright.dev/docs/ci#caching-browsers

tmastrom (approved):
Thanks for the explanation!

PR #4160, opened 2024-12-11
fix(api-client): server variables

antlio:
this pr fixes #4158 by :
- bringing back the variables table on server using path variable
- fixes send request to include variable values
- style address bar server dropdown
- sets values from reference to client modal

⊢ variables before / after
<div>
image
image
</div>

⊢ address bar dropdown before / after
<div>
image
image
</div>

PR #4263, opened 2024-12-17
revert(api-reference): sets server variables values in client

antlio:
This reverts [commit 51bfd5f](https://github.com/scalar/scalar/pull/4160/commits/51bfd5f7d5258881ae3a852df6b70be3a784b0fb) which is preventing the path parameters to be displayed in the api reference client modal.

Will get back to it with a better approach based in that case on retrieving the server variables.

image


amritk (approved):
Ah yes this is why I was trying to wait for you but we couldn't wait!

PR #4274, opened 2024-12-18
fix(api-reference): correct `hash.value` encoding to ensure scrolling

lc-soft:
**Problem**
Currently, manually refreshing the page or using a URL with a hash to open a new tab does not cause the corresponding section to scroll into the view. I have fixed similar issue in PR #4173.

I found that this issue is caused by this commit:
https://github.com/scalar/scalar/commit/56c1981d9a1e0b38837fa6881f6f76a17cc7b616#diff-ccdd228d6187612f62c06d9aaafbc4c0c9fa973ca37de2d80744df2b7dccdb43L68-R77

image

**Solution**
With this PR, the section id and `hash.value` are both raw values that have not been encoded with `encodeURIComponent()`, allowing the browser to correctly find the section and scroll it into view.

### DIFF
--- a/.changeset/poor-toys-sip.md
+++ b/.changeset/poor-toys-sip.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
+fix(api-reference): correct `hash.value` encoding to ensure scrolling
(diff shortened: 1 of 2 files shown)


hanspagel:
Oops, great that you found that! Can you just help me to reproduce the issue so I can confirm the fix?

hanspagel (approved):
What a great explanation, that really helped! Thanks so much! 🙏

PR #4276, opened 2024-12-18
ci: make important jobs a requirement for the publish process

hanspagel:
Let’s not skip CI for release PRs, I’m not really convinced this is the right solution.

But let’s make the important jobs a requirement for the publish jobs.

---

~~Release PRs are created automatically by GitHub Actions. But GitHub Actions doesn’t trigger workflows for actions performed by bots (to avoid infinite loops), that’s why they don’t run when the release PR is (automatically) updated (by the bot).~~

~~This looks like CI is hanging, so let’s just skip CI for the release PR.~~

~~**But, Hans, isn’t that dangerous?**~~

~~I don’t think so! Code has passed CI three times before being published:~~

~~1. In the PR branch (the UI warns if it’s not rebased on main)
2. When it’s in main (and the release PR is created eventually)
3. Once it’s in main again (after the release PR is merged, the publishing process gets kicked off after main passed CI)~~

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,21 +3,11 @@ name: CI
 # This workflow runs on:
 # 1. Push events to the main branch
 # 2. Pull request events
-#
-# We explicitly ignore the 'changeset-release/main' branch in both cases
-# because this branch is created by our release workflow.
-#
-# Note: GitHub Actions doesn’t run workflows triggered by bot accounts.
-# We keep the ignore rule for clarity, but the bot’s PR would be skipped anyway.
-#
-# ⚠️ That could become a problem if we filter for specific files in the trigger.
 on:
   push:
     branches:
       - main
   pull_request:
-    branches-ignore:
-      - changeset-release/main
 
 jobs:
   harden_security:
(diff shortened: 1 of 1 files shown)


tmastrom (commented):
I have a few concerns, 
- This would become unsafe if we started adding trigger filters for specific files. Untested code code potentially be released if this were to be added. Maybe add a note about this
- CI status can ignored / forgotten / hidden on the `main` branch

amritk (approved):
Works for me, but address @tmastrom's concerns

hanspagel:
True! I’ve added a comment about that, and I’ve made all important jobs (not just the build) a requirement for the publish jobs.

tmastrom (approved):

.github/workflows/ci.yml
```diff
@@ -1,10 +1,25 @@
...
+# Note: GitHub Actions doesn’t run workflows triggered by bot accounts.
+# We keep the ignore rule for clarity, but the bot’s PR would be skipped anyway.
+#
+# ⚠️ That could become a problem if we filter for specific files in the trigger.
 on:
   push:
     branches:
       - main
+    branches-ignore:
+      - changeset-release/main
```
hanspagel, line 19: actually, this trigger seems to work, so let’s keep it

```suggestion
    branches-ignore:
      - changeset-release/main
```

tmastrom (approved):

PR #4278, opened 2024-12-18
feat(api-client): import curl collection

antlio:
this pr adds a collection selector to the import curl modal and fix the server being created in drafts collection requests:

⊢ before / after
<div>
image
image
</div>


hanspagel (approved):
Love it! Tested and works great.

Issue #4240, opened 2024-12-16
Server url is stuck in Mac OS app


Yolley:
### What happens?

If I try to import `scalar-galaxy` example schema that has `servers` configured to Mac OS app the server url will be pre selected across the whole application, even when I try to create a request in a separate folder/workspace.

<img width="1263" alt="image" src="https://github.com/user-attachments/assets/8d0b7ecd-cfb8-4ceb-a44e-ae691c57ab85" />
<img width="1265" alt="image" src="https://github.com/user-attachments/assets/4590273c-6f6e-4292-bdc1-845d306b6029" />

I can't remove the base url or even select a different one.

### What did you expect to happen?

The `servers` configured should applicable only to requests within the collection folder/workspace (I guess within the collection folder is more appropriate).

### OpenAPI Document

https://sandbox.scalar.com/files/m8sCN/openapi.yaml

vladshcherbin:
Guys, how is it even possible to make such releases?

Params are broken, test request params are also broken, now it's even impossible to change the request url. Is there any automated or manual testing setup?

`Scalar` looks and feels phenomenal but the amount of bugs and blocked usage is simply unbelievable 🤯

antlio:
hey @Yolley sorry about that - thank you for reporting, we just patched this in #4197 and I'm releasing the desktop app as of right now!

antlio:
@vladshcherbin thank you for your feedback - i completely understand how frustrating it must be to encounter broken functionality.. 

while we do have automated tests and manual checks in place, this release clearly slipped through, which is on me 🙏.

to address your concerns, we just fixed the issue regarding params not being displayed - this has just been released see #4242 however if any issue remain, hit me up and i'll personally look at it asap!

vladshcherbin:
@antlio macOS app is fixed, I can confirm, thank you 🙌

From what I've seen, path params are not fixed. https://github.com/scalar/scalar/pull/4242 has [example link](https://github.com/scalar/scalar/pull/4242#issuecomment-2547337745) and standalone example [still show invalid params](https://c4ae92ab-f2c5-4253-b0bb-09de142669ce--scalar-deploy-preview.netlify.app/standalone-api-reference#tag/planets/GET/planets/{planetId})

Invalid symbols in param example:

<img width="1089" alt="image" src="https://github.com/user-attachments/assets/3c7b1ce0-1be4-4d38-a070-09f923f48d67" />

No param setting in `Test Request` (I bet I used them in some version before):

[...]

PR #4280, opened 2024-12-18
fix(api-client): removes server in drafts requests

antlio:
this PR is the final touch in fixing #4240 and includes:
- send request updates
- drafts requests migration

⊢ send request updates
+ sets active server only for non drafts requests in order to avoid issue on draft usages

⊢ drafts requests migration
+ removes server from drafts request 
+ moves previously used server to the path


amritk (approved):
Seems to be working for me, nice job on migration 👍🏾

PR #4281, opened 2024-12-18
feat(api-reference): blocks 🧩

hanspagel:
This PR brings the Scalar Blocks to life: A way to embed just the operation in any environment. I’ve taken what we discussed and the existing code and tried to make it work. There are some gaps to fill, but I think we’re actually pretty close. 👀

There are three parts to the operation: the details, the code examples and the example responses. I focused on the code examples first.

**Preview**

Screenshot 2024-12-18 at 13 14 28

**Usage**

```ts
import { getLocation, createStore, createOperationBlock } from '@scalar/api-reference/blocks'
import '@scalar/api-reference/blocks/style.css'

const { store } = createStore({
})

const operationBlock = createOperationBlock({
  // If passed, the block will mount to the element during initialization.
  // element: '#scalar-block',
  store,
  location: getLocation('GET', '/planets/{planetId}'),
})

// Mount it after initialization with just a selector string …
operationBlock.mount('#scalar-block')
// or a DOM element:
// operationBlock.mount(document.getElementById('scalar-block'))
```

**Features**

* playground for the blocks (cd packages/api-reference && pnpm playground:blocks)
* took the fantasy code we wrote in Paris and made it real
* made the code examples work
* all exposed blocks take a store and a location as input (operation, code examples, example responses)
* tests for all files!

**Tasks**

- [ ] inside the API reference, pass the store to the updated components (that expect a store now)
- [ ] make custom code examples work again, I think they don’t even exist in the store yet?
- [ ] map the example responses to the store (some work)
- [ ] map the schema components to the store (lots of work)
- [ ] variables in the operation path are italic, fix this
- [ ] testing SSG-compatibility
- [ ] actually improving the SSG output
  
**Notes**

* it’s just another entrypoint for `@scalar/api-reference` now, so we can avoid moving too many files, fighting differences between the environments and moving tons of other dependencies.
* the playground imports the source files directly, eventually we want to import from the package, but that’ll make the dev cycle slower. also it gives us a faster development cyle (see comment above).


packages/api-reference/src/blocks/components/OperationBlock/OperationBlock.vue
```diff
@@ -0,0 +1,40 @@
...
+      <CodeExamplesBlock v-bind="$props" />
+    </div>
+    <div class="border rounded p-4">
+      <ExampleResponsesBlock v-bind="$props" />
+    </div>
+  </div>
+  <div v-else>
+    <p>No operation found.</p>
+    <p>location: {{ location }}</p>
+    <p>store: {{ store }}</p>
```
antlio, line 38: any reason to keep these when no operation found?

hanspagel: just for debugging purposes :)

hanspagel:
No worries, we’ll bring this one back soon! 😊

MRocholl:
> No worries, we’ll bring this one back soon! 😊

Hey Hans! Is that somewhere on the roadmap? Im unable to find any reference to this. The starlight integration discussion brought me here. Cheers :)

EDIT: Just found your new draft. https://github.com/scalar/scalar/pull/8439 :)

Issue #4290, opened 2024-12-19
Use live-sync to update the store on spec changes


amritk:
As a quick fix, we just regenerate the store any time there's a config change. BUT for a long term performant solution we should use the live-sync diff to update only what changed.

PR #4291, opened 2024-12-19
fix(api-reference): debounce client config updates to increase perf

amritk:
This is just a quick fix to reduce typing lag. The actual fix should be to switch to the live sync diff as stated in [this ticket](https://github.com/scalar/scalar/issues/4290) but this does the job for now 🙂 

to test:
```bash
pnpx turbo dev --filter @scalar/api-reference-editor
```


hanspagel (approved):
Isn’t everything temporary? 🤔 

🤡

PR #4301, opened 2024-12-19
fix(api-reference): request server in example

antlio:
this pr brings back the previous fix for #3964 following changes that occurred in #3981 by:
- displaying request server in the request example reference
- fixing the collection server selection
- updating address bar server style

| before | after |
| -------|------|
| image | image |
| alignment of the label to the text creates a big left gap  at the top | alignment now sets to the check icon and create child parent gap |

### DIFF
--- a/packages/api-client/src/components/AddressBar/AddressBarServer.vue
+++ b/packages/api-client/src/components/AddressBar/AddressBarServer.vue
@@ -22,10 +22,15 @@ const requestServerOptions = computed(() =>
 )
 
 const collectionServerOptions = computed(() =>
-  activeCollection.value?.servers?.map((serverUid: string) => ({
-    id: serverUid,
-    label: servers[serverUid]?.url ?? 'Unknown server',
-  })),
+  // Filters out servers already present in the request
+  activeCollection.value?.servers
+    ?.filter(
+      (serverUid: string) => !activeRequest.value?.servers?.includes(serverUid),
+    )
+    .map((serverUid: string) => ({
+      id: serverUid,
+      label: servers[serverUid]?.url ?? 'Unknown server',
+    })),
 )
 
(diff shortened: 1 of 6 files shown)


packages/api-client/src/components/AddressBar/AddressBarServer.vue
```diff
@@ -84,7 +89,7 @@ const serverUrlWithoutTrailingSlash = computed(() => {
 
       <template v-if="showDropdownLabels">
         <ScalarDropdownDivider />
-        <div class="text-xxs text-c-2 ml-8">Collection Servers</div>
+        <div class="text-xxs text-c-2 px-2.5 py-1">Collection</div>
```
hwkr, line 92: I'm open to adding a `ScalarDropdownTitle` or something to components if we want

packages/api-reference/src/features/Operation/hooks/useRequestExample.ts
```diff
@@ -57,15 +57,23 @@ export const useRequestExample = ({
...
     const operationServer = server.value
-    const { modifiedUrl } = getUrlFromServerState(serverState, operationServer)
     return {
       uid: operationServer?.uid || '',
-      url: modifiedUrl || '',
+      url: serverUrl.value || '',
     }
   })
 
+  console.log('SERVER URL', serverUrl.value)
```
hwkr, line 75: Extra console?

hwkr (approved):
Looks good! Just an extra console log I think

amritk (approved):
Looks great!

packages/api-reference/src/features/Operation/hooks/useRequestExample.ts
antlio: omg tysm!

PR #4303, opened 2024-12-19
fix(api-client): showSidebar config option for the client

amritk:
**Problem**
The `showSidebar` config option doesn't do anything in the client

**Solution**
We now do not render the sidebar if that config option is false. Giving us a nice performance boost on big specs. I also renamed the old `showSidebar` var to `isSidebarShowing` to more accurately reflect what it does.

**Checklist**

### DIFF
--- a/packages/api-client/src/views/Request/RequestSubpageHeader.vue
+++ b/packages/api-client/src/views/Request/RequestSubpageHeader.vue
@@ -11,7 +11,6 @@ import { useRouter } from 'vue-router'
 
 defineProps<{
   modelValue: boolean
-  isReadonly: boolean
 }>()
 
 defineEmits<{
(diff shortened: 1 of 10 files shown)


hwkr (approved):
Looks good! Would be nice to move the config object to a provide hook one day 🔜

packages/api-client/src/views/Request/RequestSubpageHeader.vue
```diff
@@ -43,16 +40,12 @@ const { currentRoute } = useRouter()
         ]"
         :modelValue="modelValue"
         @update:modelValue="$emit('update:modelValue', $event)" />
-      <!-- todo: only show this env selectorin modal -->
-      <EnvironmentSelector
```
amritk, line 47: @cameronrohani this needed to be removed right? Currently it isn't shown at all

cameronrohani: we are planning on showing env selector when it's ready for the modal, so you can remove it for now but I believe the plan is to add it back

PR #4286, opened 2024-12-18
chore(fastify-api-reference): move example to package playground

tmastrom:
Adds the `fastify` integration example to the playground folder within the package. The playground example is built with scalar build tooling and the dockerfile and cloudbuild files reflect these changes. 

The `fastify-api-reference` keeps its vite build config due to some intricacies with copying the standalone api reference.

PR #4314, opened 2024-12-20
chore(fastify-api-reference): move example to package playground

hanspagel:
Reopening #4286, but with a fixed changeset.

Should be good to go now, but keep an eye on whether main breaks after merge.


.changeset/tender-impalas-build.md
```diff
@@ -0,0 +1,6 @@
+---
+"@scalar/fastify-api-reference": patch
+"@scalar-examples/fastify-api-reference": patch
```
hanspagel, line 3:
```suggestion
```

tmastrom (approved):
Thanks for fixing the changeset @hanspagel

PR #4318, opened 2024-12-20
feat(components): host storybook on netlify

hwkr:
I think this might get moved soon but for now it at least hosts the storybook somewhere ✨


tmastrom (commented):
Looks good. We will need @geoffgscott to create the site on netlify and add the site id github secret. Once we know we want this more permanently we can move it to cloud run like the examples

marclave (approved):
@hwkr can you work w geoff to get the netlify stuff in so we can merge this :)

marclave (approved):
LETS GO

PR #4320, opened 2024-12-21
fix(api-reference): ensure store exists before trying to create store

amritk:
**Problem**
With `immediate: true` we sometimes send an empty spec to the watcher which then tries to create a store from it. 

**Solution**
We ensure a spec exists before trying to do so.

**Checklist**

- [ ] I’ve added tests for the regression or new feature. - I will add a test in Cam's PR

before:
image
after:
image


marclave (approved):

hwkr:
🔥

PR #4326, opened 2024-12-22
fix(api-client): params not saving or sending

### LINKED ISSUE #4325: Query parameters are longer being sent
### What happens?

It seem that there was a regression in one of the recent scalar versions. The web client currently does not send any query parameters, it's reproducible with an Example Galaxy API if we try to make a `Get All Planets` request with limit and offset configured.

image

I expect the API client to send query parameters if they were configured and enabled.

### OpenAPI Document

https://sandbox.scalar.com/files/u7-Dw/openapi.yaml

amritk:
**Problem**
Params aren't saving or sending (store is not being updated)

**Solution**
We rollback and use a different strategy for the stricter tsconfig update

closes #4325

**Checklist**

- [ ] I’ve added tests for the regression or new feature. - Needs tests but the fix needs to get out asap

### DIFF
--- a/packages/api-client/src/views/Request/RequestSection/RequestParams.vue
+++ b/packages/api-client/src/views/Request/RequestSection/RequestParams.vue
@@ -50,26 +50,22 @@ const updateRow = (rowIdx: number, field: 'key' | 'value', value: string) => {
   const currentParams = params.value
   if (currentParams.length > rowIdx) {
     const updatedParams = [...currentParams]
-    updatedParams[rowIdx] = {
-      ...updatedParams[rowIdx],
-      [field]: value || '',
-      value: updatedParams[rowIdx]?.value || '',
-      key: updatedParams[rowIdx]?.key || '',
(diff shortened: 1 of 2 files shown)


packages/api-client/src/views/Request/RequestSection/RequestParams.vue
```diff
@@ -50,26 +50,22 @@ const updateRow = (rowIdx: number, field: 'key' | 'value', value: string) => {
...
   if (currentParams.length > rowIdx) {
     const updatedParams = [...currentParams]
-    updatedParams[rowIdx] = {
-      ...updatedParams[rowIdx],
-      [field]: value || '',
-      value: updatedParams[rowIdx]?.value || '',
-      key: updatedParams[rowIdx]?.key || '',
-      enabled: updatedParams[rowIdx]?.enabled ?? false,
-    }
+    if (!updatedParams[rowIdx]) return
```
amritk, line 53: I rolled back the changes and added this line

marclave (approved):

PR #4337, opened 2024-12-24
chore(nextjs-api-reference): remove react peer dependency

amritk:
**Problem**
The nextjs package requires react 19.

**Solution**
With this PR we remove that requirement

**Checklist**


amritk:
I re-explored this, but it doesn't look like you can consume a package built with react 19 in react 18. Unless I'm missing something we may need to build multiple versions of the app. For now anyone on react 18 can use the older version.

https://discord.com/channels/1135330207960678410/1316466399140380752

Issue #4322, opened 2024-12-21
openapi-parser filter not working as expected


qqilihq:
### What happens?

I'm trying to chain a filter with the openapi-parser like so:

```typescript
    const spec = {
      openapi: '3.1.0',
      paths: {
        '/test': {
          get: {
            responses: {
              '200': {
                description: 'visible',
              },
            },
          },
        },
        '/hidden': {
          get: {
            'responses': {
              '200': {
                description: 'hidden',
              },
            },
            'x-internal': true,
          },
        },
      },
    }

    const result = await openapi()
      .load(spec)
      .filter((schema) => !schema['x-internal'])
      .toJson()

    console.log(result)
```

However the result printed out is always the same as the input.

### What did you expect to happen?

The example above should return a modified spec based on the filter.

hanspagel:
Bummer! Thanks for the report. Would you be up to submit a PR for it? 👀

qqilihq:
Yes, I'll give it a try at least - it came up while looking for a solution for https://github.com/scalar/scalar/issues/3434

hanspagel:
Won’t fix that one I’m afraid, we’re not using this feature anymore.

Issue #3434, opened 2024-10-10
Ability to modify spec. on the fly


hanspagel:
We're currently using fastify-swagger-ui but I'm enjoying scalar very much - good work 👍 

In our current setup using `fastify-swagger-ui` we have implemented a technique which allows to show an extended spec for “pro users”. For that, they load the spec via `api.example.com/documentation` and modify the URL in the Swagger UI URL bar from `https://api.example.com/documentation/json` to `https://api.example.com/documentation/json?expert=true`.

This approach is implemented using fastify-swagger-ui’s [`transformSpecification`](https://github.com/fastify/fastify-swagger-ui?tab=readme-ov-file#transformspecification):

```typescript
await fastify.register(fastifySwaggerUi, {
  routePrefix: '/documentation',
  transformSpecification: (swaggerObject, request) => {
    const expertSpec = request.query.expert === 'true';
    if (expertSpec) {
      return swaggerObject;
    } else {
      return trimDownSpec(swaggerObject);
    }
  },
  // …
);
```

I'm currently wondering how to best map this to scalar:

1. Is there some possibility included to detect a query parameter when accessing the spec?
2. Is there an equivalent of transformSpecification? There's a spec factory, but it doesn't provide any further context:

```typescript
await fastify.register(scalarApiReference, {
  routePrefix: '/documentation',
  configuration: {
    spec: {
      // called upoon each request, but no request-specific parameters or context available which I could use
      content: () => trimDownSpec(fastify.swagger())
    }
  },
  // …
);
```

3. Any better ideas how to implement this?

_Originally posted by @qqilihq in https://github.com/scalar/scalar/discussions/3433_

hanspagel:
Duplicate of #5317

PR #4342, opened 2024-12-25
feat(fastify-api-reference): dynamic fastify schema modifications

qqilihq:
**Problem**
See #3434 

**Solution**
* Adds a hook `processSpec` to fastify-api-reference to modify a schema on the fly. The hook has access to the current request, so it can e.g. check for a query parameter to modify the schema.
* Any query parameters on the request URL to the HTML frontend are passed on to the spec. URL. (I thought if this should be behind an explicit toggle, but I don’t assume any breaking change)
* Fixes the queue’s filter functionality (see #4322)
* Adds a `map` operation to the queue which allows arbitrary modifications to the schema (more flexible than filtering)

Bildschirmfoto 2024-12-25 um 19 19 13

**Checklist**

### DIFF
--- a/.changeset/proud-flowers-march.md
+++ b/.changeset/proud-flowers-march.md
@@ -1,0 +1,6 @@
+---
+'@scalar/fastify-api-reference': minor
+'@scalar/openapi-parser': minor
+---
+
+Add a hook to `fastify-api-reference` to conditionally modify schema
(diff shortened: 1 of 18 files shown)


hanspagel (changes requested):
I love, love, love that feature! Maybe we should put it in the core so it’s available in all integrations, though. 👀 Has some challenges though, let me think for a minute.

hanspagel:
Sorry for the silence here. We’ll integrate this eventually, but I really would love us to integrate that for all integrations vs. just for one.

qqilihq:
@hanspagel Sure - understand. Looking forward!

PR #4363, opened 2025-01-04
Geoff/repo reorg

geoffgscott:
Moves the integrations into a separate folder. Integrations should never be a dependency of another workspace package.


packages/components/src/components/ScalarFloating/ScalarFloating.vue
```diff
@@ -97,7 +97,6 @@ const { floatingStyles, middlewareData } = useFloating(targetRef, floatingRef, {
     <slot />
   </div>
   <ScalarTeleport
-    v-if="$slots.floating"
```
hwkr, line 100: This is important iirc. I think there are teleport issues if we try to create this without the floating element 🤔

hwkr: The combobox ones I think should be removed too otherwise empty divs get included.

hwkr (commented):
🔥

packages/components/src/components/ScalarFloating/ScalarFloating.vue
geoffgscott: We need to validate this behaviour with latest Vue. I am not sure if it's vue-tsc or a change to vue internals but TS is saying `$slots.floating` is always defined.

hanspagel:
Awesome initiative! Will take this over here, split it up in a few smaller PRs and try to get this all in ASAP 🫡

hanspagel:
I’ve created a bunch of new PRs from it already. Some changes are missing, but we don’t need to keep this PR open for longer, I’ll just follow with the remaining changes once a few of the other PRs are merged. ✌️

PR #4368, opened 2025-01-06
docs: add Django integration to the root README

hanspagel:
I’ve just discovered a new Scalar Django integration by @m1guer! Let’s add it to the root README. :)


marclave (approved):
awesome :)

PR #4374, opened 2025-01-06
chore(themes): remove dependency, clean up types

hanspagel:
This PR cleans up the types, the TS config and removes the dependency from `@scalar/themes`. ✨ 

Fresh, clean and lightweight. Helps to avoid circular type references.


packages/themes/vite.config.ts
```diff
@@ -26,5 +26,17 @@ export default defineConfig({
       enabled: true,
       reporter: 'text',
     },
+    /**
```
hwkr, line 29: Do we need this again? I'm not sure why it was removed 🤔

hanspagel: Geoff removed the vitest.config.ts and put it into the vite.config.ts in #4363, I just redid the changes in a separate PR. I’m fine with both approaches (two configs or one config).

packages/api-reference/src/components/ApiReferenceLayout.vue
```diff
@@ -405,7 +405,7 @@ const themeStyleTag = computed(
             v-if="configuration?.isEditable"
             #empty-state>
             <GettingStarted
-              :theme="configuration?.theme || 'default'"
+              :theme="(configuration?.theme as ThemeId) || 'default'"
```
amritk, line 408:
```suggestion
              :theme="configuration?.theme || 'default'"
```
we shouldn't need this?

amritk: also we should rarely use `as` as it bypasses type safety

amritk (changes requested):
I think we should discuss this. What was the issue with having types as a dependency? The types package was the solution to remove the duplication.

hanspagel:
I agree. I mostly copied the code blindly, let me take a look at how we can improve this.

hanspagel:
Closing this for new, maybe starting a fresh attempt next week.

PR #4294, opened 2024-12-19
style(api-client): auth redesign 

cameronrohani:
changed auth styles in client + intro cards

before:
image
image

after:
image
image

PR #4377, opened 2025-01-06
style(api-client): auth table pill

antlio:
**Problem**
selected auth background in darkmode is not highlighted as being sometime overloaded by the `cm-pill` utility usage that includes background property.

**Solution**
this pr removes the `cm-pill` utility usage while maintaining the style in order to only rely on the rightful background regarding auth type pill.

| before | after |
| -------|------|
|  image | image |
| auth type background is missing | auth type pill background is now displayed  |

**Checklist**

### DIFF
--- a/.changeset/lucky-steaks-do.md
+++ b/.changeset/lucky-steaks-do.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
+
+style: fixes dark auth pill ui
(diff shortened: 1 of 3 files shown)


hanspagel (approved):
One less `!important`, and a fixed bug, double W 🥳

cameronrohani:
PR #4294 gets rid of the pills in auth so we should prioritize getting that in :)

antlio:
> PR #4294 gets rid of the pills in auth so we should prioritize getting that in :)

oh yea it's not like i worked on it either 😂 tysm closing according to previous comment

PR #3454, opened 2024-10-11
fix(api-client): import curl query parameters

antlio:
this pr fixes query parameters importation form curl following latest changed introduced in #3267.

**after**

https://github.com/user-attachments/assets/b14a0428-dfcd-461d-b575-3a4efa2eaeb4

PR #4384, opened 2025-01-07
fix(api-client): parse curl query parameters

antlio:
**Problem**
as pointed out by @amritk in #3454 there was an unneeded condition and type mistake in the parse curl function.

**Solution**
this pr updates the type and remove the condition accordingly.

**Checklist**

### DIFF
--- a/.changeset/calm-forks-bake.md
+++ b/.changeset/calm-forks-bake.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
+
+fix: updates parse curl query params type and condition
--- a/packages/api-client/src/libs/parse-curl.ts
+++ b/packages/api-client/src/libs/parse-curl.ts
@@ -17,7 +17,7 @@ export function parseCurlCommand(curlCommand: string) {
     method?: RequestMethod
     headers?: Record<string, string>
     body?: string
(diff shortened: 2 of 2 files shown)


hanspagel (changes requested):
pls add tests 🙏

hanspagel:
Yes, something that covers the change! :)

amritk:
yes agreed with hans, I wasn't even sure what that section was doing, some comments would be good as well! Because in my testing I couldn't hit that code.

antlio:
> yes agreed with hans, I wasn't even sure what that section was doing, some comments would be good as well! Because in my testing I couldn't hit that code.

i went through it and actually that condition is not used anymore as I introduced `shell-quote`, it was originally used to get query params for the following curl:
```curl
curl --d 'key=value' https://example.com
curl --data 'key=value' https://example.com
```

will update and cover that case

hanspagel (approved):
Nice work! I’ve aded a bunch of tests to cover more (edge) cases, some were failing and I fixed them.

@antlio Can you review this?

antlio:
> Nice work! I’ve aded a bunch of tests to cover more (edge) cases, some were failing and I fixed them.
> 
> @antlio Can you review this?

great coverage tysm! everything lgtm

PR #4341, opened 2024-12-24
fix: runtime error with radix tooltip

marclave:
**Problem**
Currently we get a nasty runtime error w radix tooltip! not sure why, but gunna comment it out and resolve in the new year

https://github.com/scalar/scalar/issues/4334

**Solution**
i commented it out

**Checklist**

I’ve gone through the following:

- [x] I’ve added an explanation _why_ this change is needed.
- [x] I’ve added a changeset (`pnpm changeset`).
- [ ] I’ve added tests for the regression or new feature.
- [ ] I’ve updated the documentation.

See the [contributing guide](https://github.com/scalar/scalar/blob/main/CONTRIBUTING.md) for more information.

PR #4387, opened 2025-01-07
fix(api-client): request params tooltip runtime error

### LINKED ISSUE #4334: Api client throws error if there is no GET method in the OpenAPI spec.
### What happens?

`createApiClientApp` throw error when there is no GET method in the paths.

It should work without error. For example it causes gitbook to crash.

https://github.com/user-attachments/assets/0bcf99c9-98bc-413a-b43e-103f276dd4b4

Ekran Resmi 2024-12-23 08 08 00

### OpenAPI Document

[openapi.json](https://petstore3.swagger.io/api/v3/openapi.json)

antlio:
**Problem**
currently empty api client usage through reference is throwing the following error:
image

**Solution**
this pr fixes #4334 as a follow of #4341 and makes sure to only have the clear action available in parameter presence.

tested using api client react playground as highlighted by @marclave.

**Checklist**

### DIFF
--- a/packages/api-client/src/views/Request/RequestSection/RequestParams.vue
+++ b/packages/api-client/src/views/Request/RequestSection/RequestParams.vue
@@ -163,7 +163,6 @@ watch(
     <template #actions>
       <div
         class="text-c-2 flex whitespace-nowrap opacity-0 group-hover/params:opacity-100 has-[:focus-visible]:opacity-100 request-meta-buttons">
-        <!-- TODO fix this DOC-2740 -->
         <ScalarTooltip
           v-if="params.length > 1"
           side="right"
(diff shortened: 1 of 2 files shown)


packages/api-client/src/views/Request/RequestSection/RequestParams.vue
```diff
@@ -164,7 +164,8 @@ watch(
       <div
         class="text-c-2 flex whitespace-nowrap opacity-0 group-hover/params:opacity-100 has-[:focus-visible]:opacity-100 request-meta-buttons">
         <!-- TODO fix this DOC-2740 -->
```
hanspagel, line 166:
```suggestion
```

amritk (approved):
Code looks good! @marclave can test the case where it was happening

marclave (approved):

PR #4389, opened 2025-01-07
feat(oas-utils): examples for schemas with `patternProperties`

### LINKED ISSUE #3911: patternProperties is not supported
### What happens?

Objects with `patternProperties` are shows as `object` without properties in the UI.

Here's an example of a schema containing such an object:

```json
{
	"schema": {
		"type": "object",
		"properties": {
			"externalId": {
				"type": "object",
				"patternProperties": {
					"^(.*)$": {
						"type": "object",
						"properties": {
							"dataId": {
								"type": "string"
							},
							"link": {
								"anyOf": [
									{
										"format": "uri",
										"type": "string"
									},
									{
										"type": "null"
									}
								]
							}
						},
						"required": ["dataId", "link"]
					}
				}
			}
		}
	}
}
```

image

I expected the object to show nicely in both the schema explanation & the example.

### OpenAPI Document

https://sandbox.scalar.com/e/w4-CN

hanspagel:
**Problem**
Currently, we don't add anything to the generated examples when the schema has `patternProperties` (just a regex pattern instead of a property name).

**Solution**
With this PR we use the regex pattern as the property name, see the attached test for an example.

Fixes #3911.

**Checklist**

### DIFF
--- a/.changeset/real-dolls-walk.md
+++ b/.changeset/real-dolls-walk.md
@@ -1,0 +1,5 @@
+---
+'@scalar/oas-utils': patch
+---
+
+feat: generate examples for objects with patternProperties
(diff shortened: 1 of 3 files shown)


amritk (approved):
Looks cool 👍🏾 has test 👍🏾

PR #4390, opened 2025-01-07
fix(api-client): request sidebar watch toast

antlio:
**Problem**
watch mode state notification get displayed in the desktop app when removing a collection.

**Solution**
this pr prevents displaying watch mode state notification on collection deletion + for drafts collection and add missing type in request sidebar component file.

**Checklist**


marclave (approved):
nice fix @antlio

Issue #4349, opened 2024-12-30
Re-ordering of environments within collections not possible


sebdesalvador:
### What happens?

After creating a few environments within a collection, when trying to re-order them with the mouse, new unnamed collections appear.

### What did you expect to happen?

Environments within my collection should have been re-ordered.

hanspagel:
Oh no! Can you share a screenshot and the browser you’re using? I wasn’t able to repoduce it yet.

sebdesalvador:
<img width="622" alt="image" src="https://github.com/user-attachments/assets/ada32496-3389-4f01-983d-1e76a19bb53a" />

I'm on mac, using the standalone app. As you can see on the picture, the unnamed selected folder was created after I tried to drag and drop `test2` before `test`.

antlio:
tysm @sebdesalvador i've been able to reproduce that too! i'll fix first that behavior and then look into implementing dragging there as it was not initially added in environments - will keep updated here!

sebdesalvador:
My pleasure @antlio :)

antlio:
@sebdesalvador the error has been found and a pr opened - i'll just need a bit more time to include the dragging for environment

PR #4391, opened 2025-01-07
fix(scalar-app): drag self importation

antlio:
**Problem**
when using the desktop app, drop event creates empty collection as seen in #4349 - as relying on `text/uri-list` that is available in the app context.

to reproduce: 
- using the desktop app
- drag a request from a collection to another

| before | after |
| -------|------|
| image | image |
| an empty collection is created on drop event | no empty collection is created on drop event |

**Solution**
prevents self importation on drop event by avoiding import action for url origin.

**Checklist**

### DIFF
--- a/packages/scalar-app/src/renderer/src/main.ts
+++ b/packages/scalar-app/src/renderer/src/main.ts
@@ -78,6 +78,12 @@ async function drop(e: DragEvent) {
   // Check if the user dropped an URL
   if (e.dataTransfer?.getData('text/uri-list')) {
     const url = e.dataTransfer.getData('text/uri-list')
+    const appOrigin = window.location.origin
+
+    // Prevents self importation that is causing collection creation
+    if (url.startsWith(appOrigin)) {
+      return
+    }
 
     if (url) {
       client.store.importSpecFromUrl(url, 'default')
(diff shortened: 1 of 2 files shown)


amritk:
I was unable to reproduce the issue

antlio:
> I was unable to reproduce the issue

apologies i updated the description accordingly to clarify how to reproduce

hwkr:
> > I was unable to reproduce the issue
> 
> apologies i updated the description accordingly to clarify how to reproduce

Gave it a try and I was able to reproduce in my dev env

Electron-2025-01-08-10-58-49@2x

hwkr:
> > I was unable to reproduce the issue
> 
> 
> apologies i updated the description accordingly to clarify how to reproduce

Gave it a try and I was able to reproduce in my dev env and also confirm the fix, great work @antlio!

Electron-2025-01-08-10-58-49@2x

hwkr (approved):

packages/scalar-app/src/renderer/src/main.ts
```diff
@@ -100,4 +106,5 @@ async function drop(e: DragEvent) {
 function dragover(e: DragEvent) {
   e.preventDefault()
   e.stopPropagation()
```
hwkr, line 108: Optional but I _think_ you can remove `stopPropagation` since you added `stopImmediatePropagation`

antlio: oh that's a testing leftover, removed!

marclave (approved):

PR #4398, opened 2025-01-08
feat(components): make ScalarMenu into a dropdown

hwkr:
This updates the design for the ScalarMenu to be more similar to our ScalarDropdown and to have a nested menu for the team switcher.

https://github.com/user-attachments/assets/4b903b83-d328-4db9-8002-b6da6ed95bf5

**Checklist**


antlio:
great work! 

nit but should we keep prepending checkbox instead of appending it for the sake of consistency? 
image

in dropdown
image

amritk:
lets get some storybook deploy previews goin? what you think @tmastrom

hwkr:
> great work!
> 
> nit but should we keep prepending checkbox instead of appending it for the sake of consistency?

We could! It was a little busy with the profile picture there and the checkbox but I'm open to it.

> Did we need to make this many components? Or could we have done less components with some different variants?

That's a good question, it was originally done this way to keep it composable because we need to use the parts in a few different ways in different projects (e.g. some of them need to add redirects to the projects etc). Additionally any changes we need to make in Scalar/Scalar take an extra iteration cycle to make the change, release, and then update the scalar-org products. Keeping them composable (hopefully) makes it easier to adapt the components without have to modify the component library all the time.

amritk:
Yea thats what I mean! Like this are strict components so any change has to be made in components. BUT if we just did the standard props of something like:

link
prepend-slot (for images)
checkbox
etc

we wouldn't need to make changes to the components and can just do everything from org. But also I haven't worked on it so maybe I don't know what I'm talking about.

hwkr:
> Yea thats what I mean! Like this are strict components so any change has to be made in components. BUT if we just did the standard props of something like:
> 
> link prepend-slot (for images) checkbox etc
> 
> we wouldn't need to make changes to the components and can just do everything from org. But also I haven't worked on it so maybe I don't know what I'm talking about.

Maybe 🤔 I'm not sure which component you're talking about merging though

marclave (approved):

cameronrohani (approved):

Issue #4397, opened 2025-01-08
Cookie Parameters are not displayed in the parameter list


Catzilla:
### What happens?

[Cookie Parameters](https://swagger.io/docs/specification/v3_0/describing-parameters/#cookie-parameters) are not displayed in the operation parameter list in API Reference

### What did you expect to happen?

Cookie Parameters displayed in list, somewhere around Query Parameters or Headers

### OpenAPI Document

https://sandbox.scalar.com/e/CfLq9

PR #4399, opened 2025-01-08
fix(api-reference): cookie parameters don’t show up

hanspagel:
**Problem**
Currently, cookie parameters don’t show up in the list of parameters. 

See #4397 for more context

**Solution**
With this PR we’re adding the cookie parameters, so they don’t feel excluded anymore.

Screenshot 2025-01-08 at 10 53 26

**Example**

```yaml
openapi: 3.1.0
info:
  title: Hello World
  version: 1.0.0
components:
  parameters:
    CSRFCookie:
      name: csrf
      in: cookie
      description: Anti-CSRF cookie
      required: true
      schema:
        type: string
paths:
  /example:
    post:
      parameters:
        - $ref: '#/components/parameters/CSRFCookie'
```

**Checklist**

I’ll add this fix and a test to #4281, where I modified the relevant components already.

EDIT: Done.

### DIFF
--- a/.changeset/heavy-pugs-remain.md
+++ b/.changeset/heavy-pugs-remain.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
+fix: cookie parameters don’t show up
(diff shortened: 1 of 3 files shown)


amritk (approved):
Cookies are always so neglected

hwkr (approved):
🔥

PR #4405, opened 2025-01-08
fix(oas-utils): handles object in examples

### LINKED ISSUE #3782: ArraySchema example becomes object in requestbody.
### What happens?

I define an example in my request:
```java
@Schema(example = "101010101")
            String stationId,
            @ArraySchema(arraySchema = @Schema(example = "[\"testsvr\",\"testsvr\"]"))
            List<String> operatorIds
```

It is displayed normally in request example. 

图片

But when i click `test request` button, it becomes object in requestbody.

图片

it should be an array

### OpenAPI Document

_No response_

antlio:
**Problem**
Currently, using the following in a document  `"example": ["Portfolio1", "Portfolio2"]` gets returned as an object in the request example.

**Solution**
this pr handles that use case and fixes #3782  in order to turn the object into an array to maintain its original type.

**Checklist**

### DIFF
--- a/packages/oas-utils/src/spec-getters/getExampleFromSchema.test.ts
+++ b/packages/oas-utils/src/spec-getters/getExampleFromSchema.test.ts
@@ -1072,12 +1072,69 @@ describe('getExampleFromSchema', () => {
     })
   })
 
-  it('handles objects in examples and returns them as an array', () => {
+  it('handles multiple examples for a string type', () => {
     expect(
       getExampleFromSchema({
-        type: 'array',
-        example: ['example 1', 'example 2'],
+        type: 'string',
+        examples: {
+          example1: 'example string 1',
+          example2: 'example string 2',
+        },
       }),
-    ).toMatchObject(['example 1', 'example 2'])
+    ).toBe('example string 1') // first example
+  })
+
+  it('handles multiple examples for a number type', () => {
+    expect(
+      getExampleFromSchema({
+        type: 'number',
+        examples: {
+          example1: 42,
+          example2: 100,
+        },
+      }),
+    ).toBe(42) // first example
+  })
+
+  it('handles single example for an object type', () => {
+    expect(
+      getExampleFromSchema({
+        type: 'object',
+        example: { key: 'value' },
+      }),
+    ).toMatchObject({ key: 'value' })
+  })
+
+  it('handles multiple examples for an object type', () => {
+    expect(
+      getExampleFromSchema({
+        type: 'object',
+        examples: {
+          example1: { key1: 'value1' },
+          example2: { key2: 'value2' },
+        },
+      }),
+    ).toMatchObject({ key1: 'value1' }) // first example
+  })
+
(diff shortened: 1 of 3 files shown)


packages/oas-utils/src/spec-getters/getExampleFromSchema.test.ts
```diff
@@ -1071,4 +1071,13 @@ describe('getExampleFromSchema', () => {
       },
     })
   })
+
+  it('handles objects in examples and returns them as an array', () => {
+    expect(
+      getExampleFromSchema({
+        type: 'array',
+        examples: [{ 0: 'example 1', 1: 'example 2' }],
```
hanspagel, line 1079: That syntax seems strange and is not what’s in the PR description, pls enlighten me :D

antlio: seems like using

```yaml
example:
  - example 1
  - example 2
```

is turned into a schema `examples` object, can sync about this if need!

antlio: oops updated

packages/oas-utils/src/spec-getters/getExampleFromSchema.ts
```diff
@@ -132,9 +132,12 @@ export const getExampleFromSchema = (
     }
   }
 
-  // Use the first example, if there’s an array
+  // Use the first example, if there’s an array, and handle objects
   if (Array.isArray(schema.examples) && schema.examples.length > 0) {
-    return cache(schema, schema.examples[0])
+    const firstExample = schema.examples[0]
+    return typeof firstExample === 'object'
+      ? cache(schema, Object.values(firstExample))
```
amritk, line 139: what if the example is supposed to be an object? Wouldn't we want the whole object?

antlio: good call i guess it should remain, this was a first quick shot but i'll dig deeper regarding the casting

hanspagel (approved):
Actually, this didn’t work for strings, numbers and objects with the `examples` (the new way to provide multiple examples in OpenAPI 3.1).

I’ve added some more tests and fixed the code. ✌️

antlio:
@hanspagel i still get an object returned vs array

> Actually, this didn’t work for strings, numbers and objects with the `examples` (the new way to provide multiple examples in OpenAPI 3.1).
> 
> I’ve added some more tests and fixed the code. ✌️

I think i'm getting the object returned again instead of the array in the request example

hanspagel (changes requested):
I broke it

hanspagel (changes requested):
I think it’s working now. But I think it’s not fixing the initial bug report.

It looks like there’s somewith wrong in createRequestOperation. Before merging, we need to reproduce the issue in main and make sure it’s fixed here.

antlio:
closing this in favor of #4444

PR #4412, opened 2025-01-09
fix(api-reference): displays array enum of one

### LINKED ISSUE #2006: Array enum with exactly 1 enum does not render as a enum

antlio:
**Problem**
Currently when using only one enum value, the value doesn't get displayed in the api reference schema property.

**Solution**
this pr fixes #2006 by showing enum of 1 value in the reference.

| before | after |
| -------|------|
| image | image |
| enum list value is not being displayed | enum list value is displaying the one value |

**Checklist**

### DIFF
--- a/.changeset/tough-pillows-kick.md
+++ b/.changeset/tough-pillows-kick.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
+fix: shows enum value when there is only one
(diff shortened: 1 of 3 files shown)


hanspagel (approved):
Nice fix!!

hwkr (approved):
🥳

Issue #4357, opened 2025-01-02
Unable to Edit Field Name When Uploading Files


unnoq:
### What happens?

When uploading a file, the field name always defaults to the local file name, even if I try to edit it. Changes to the field name do not persist, and the field name remains locked to the local file name.

![image](https://github.com/user-attachments/assets/fa8d1883-8f87-4a81-b54b-1565a709ed38)

expect: Field name for field should work

### What did you expect to happen?

The field name should be editable and retain any modifications made by the user, rather than defaulting to the local file name.

RemcoGoy:
@unnoq , I think this should be fixed in #4367

unnoq:
Thank you, @RemcoGoy! I can confirm it's working now. However, there's still a bit of an issue: when uploading a file, it overrides the current field name, which is a bit inconvenient.

RemcoGoy:
@unnoq , I've added a PR that should fix this

hanspagel:
Should be released as @scalar/api-client@2.2.29 (https://github.com/scalar/scalar/pull/4379). 🥳

PR #4413, opened 2025-01-09
fix(api-client): filename overwrites key/value in multipart form body

RemcoGoy:
**Problem**
Currently, when settings data (key and value) in the multipart form body and uploading a file after. The file will overwrite the existing data you might have already set.

**Solution**
With this PR, I made sure the body will select key or value first if it exists already, otherwise it will default back to the filename for both fields.

**Issues**
solves #4357 

**Checklist**

---

Feel free to let me know if this change is what we would want. Or if I can change/improve anything.

### DIFF
--- a/.changeset/blue-bags-dance.md
+++ b/.changeset/blue-bags-dance.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
+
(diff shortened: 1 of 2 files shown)


hanspagel (approved):
What a great contribution! Tested and works great. :)

PR #4414, opened 2025-01-09
fix(api-client): allows enabled empty parameters in request

### LINKED ISSUE #4406: Enabled empty parameters are omitted
### What happens?

We're experiencing an issue with our ASP.NET Core API (using .NET 9) and Scalar when sending requests to our GetAll endpoints.

For all our GetAll endpoints, we define a request DTO that includes all the query parameters, like this:

public record GetAllItemsRequest : IPaginationInfo
{
    [DefaultValue(PaginationDefaults.DefaultPage)]
    public int Page { get; init; } = PaginationDefaults.DefaultPage;
    [DefaultValue(PaginationDefaults.DefaultPageSize)]
    public int PageSize { get; init; } = PaginationDefaults.DefaultPageSize;
    [DefaultValue("")]
    public string Query { get; init; } = "";
    public FilterDefinitionDto<AdvertisementResponse>[] Filter { get; init; } = [];
    public SortDefinitionDto<AdvertisementResponse>[] Sort { get; init; } = [];
}
When we use Scalar to send a request, the API returns an error stating that it cannot bind to the Filter and Sort parameters.

ASP.NET Core expects the Filter and Sort parameters to be included in the query string but Scalar does not include these empty query parameters in the query string. This happens even though they are explicitly selected in the UI.

This issue does not occur when using other API clients like Postman or Swagger as they correctly include the empty parameters (when selected). This might be related to issue https://github.com/scalar/scalar/pull/3291 which seems to have changed how Scalar handles empty query parameters.

_Originally posted in discussion [#4400](https://github.com/scalar/scalar/discussions/4400) by @TheDusty01_

### OpenAPI Document

_No response_

antlio:
**Problem**
Currently empty parameters are not sent through request rather enabled or not.

**Solution**
this pr fixes #4406 in order to allows enabled empty parameters to be sent while keep omitting the non enabled ones.

nb. we should make sure that this still provide a solution for #3291

**Checklist**

### DIFF
--- a/.changeset/new-spiders-hammer.md
+++ b/.changeset/new-spiders-hammer.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
+
+fix: allows enabled empty query params and omits disabled ones
--- a/packages/api-client/src/libs/send-request/create-request-operation.test.ts
+++ b/packages/api-client/src/libs/send-request/create-request-operation.test.ts
@@ -386,7 +386,34 @@ describe('create-request-operation', () => {
     expect(JSON.parse(result?.response.data as string).query).toStrictEqual({})
   })
 
-  it('should ignore query parameters with empty values', async () => {
+  it('should ignore query parameters with empty values that are not enabled', async () => {
+    const [error, requestOperation] = createRequestOperation(
+      createRequestPayload({
+        serverPayload: {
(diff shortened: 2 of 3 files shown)


hwkr (approved):
✨

hanspagel (approved):
smart

PR #4415, opened 2025-01-09
ci: update paths of the fastify and express examples

tmastrom:
Some examples have moved from the `examples` folder to `packages`. This is updates the cloudbuild env to use the correct paths. Eventually all examples will be moved to packages. This migration is in progress 😄


hanspagel (approved):
LATM

Looks awesome to me

Thanks for the quick fix

Let’s see if we can get main back to green

PR #4416, opened 2025-01-09
fix(openapi-parser): call `onDereference` after dereference process

### LINKED ISSUE #4347: `onDereference` called for non-ready schemas
### What happens?

If we try to make a log of resolved schemas inside `onDereference` callback, we will get an object before the resolving
image
outputs:
image

https://github.com/scalar/scalar/blob/ddbd0cdf277340d7b8bc5a3443c2190f24c0c32e/packages/openapi-parser/src/utils/resolveReferences.ts#L132-L141

Schemes should be resolved before the callback is called. This may be unexpected behavior if there is any schema processing going on in the callback, assuming that it has already been resolved

### OpenAPI Document

_No response_

fuma-nama:
**Problem**
fix #4347 

**Solution**
Call the hook after dereferencing the schema, and add docs about the behaviours.

Also added `errors` array to parameter instead of return value of `dereference`, so it reduces a little memory usage

**Checklist**

### DIFF
--- a/.changeset/sour-cougars-sing.md
+++ b/.changeset/sour-cougars-sing.md
@@ -1,0 +1,5 @@
+---
+"@scalar/openapi-parser": patch
+---
+
+Call `onDereference` after dereference process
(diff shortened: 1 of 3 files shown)


hanspagel (commented):
OMG, thanks for the great contribution! Looks great at a quick glance, do you want me to add tests?

fuma-nama:
Sure! Would be nice to have more testing

hanspagel (approved):
AWESOME work my friend, tysm! ✨

PR #4417, opened 2025-01-09
feat(api-reference): hide tags entirely

### LINKED ISSUE #4404: Make `x-internal: true` work with tags
As discussed yesterday on [Twitter](https://x.com/emilwidlund/status/1874761191664517222), we would like to have a way to organize the operations in our OpenAPI schema in a more advanced way than just tags.

For us, the ideal way would be to have custom OpenAPI extensions:

1. A first one to hide internal operations, e.g. `x-scalar-ignore`. It's worth to note that we already use a similar tag with Speakeasy, [`x-speakeasy-ignore`](https://www.speakeasy.com/docs/speakeasy-reference/extensions).
2. Another one to group operations, instead of using the basic tags (we have several tags on our endpoints), e.g, `x-scalar-group`, that would control the group in which the operations goes into. Ideally, this would also let us set the "display name" of the category.

_Originally posted by @frankie567 in https://github.com/scalar/scalar/discussions/4359_

We need to make sure this works in the api-reference and the api-client.

hanspagel:
**Problem**

Currently, you can hide operations and webhooks from the reference (but not from the client). But some people might want to hide entire tags (and the attached operations). 

Fixes #4404

**Solution**

With this PR you can hide operations, webhooks and tags (including their webhooks and operations) in the reference and the API client. 🤯 

Works with `x-internal: true` (as before) and specifically with `x-scalar-ignore: true`, too.

**Checklist**

**Example**

```json
{
    "openapi": "3.1.0",
    "info": {
      "title": "Sample API",
      "version": "1.0.0",
      "description": "A simple example API"
    },
    "tags": [
      {
        "name": "users",
        "description": "Operations about users",
        "x-internal": true
      }
    ],
    "components": {
      "schemas": {
        "HIDDENUSER": {
          "type": "object",
          "x-scalar-ignore": true,
          "properties": {
            "id": { "type": "integer" },
            "name": { "type": "string" }
          }
        }
      }
    },
    "webhooks": {
      "newUser": {
        "post": {
          "tags": ["users"],
          "summary": "HIDDEN WEBHOOK",
          "description": "Triggered when a new user is created in the system",
          "requestBody": {
            "content": {
              "application/json": {
                "schema": {
                  "type": "object",
                  "properties": {
                    "userId": {
                      "type": "integer",
                      "description": "The ID of the newly created user"
                    },
                    "name": {
                      "type": "string",
                      "description": "The name of the newly created user"
                    },
                    "createdAt": {
                      "type": "string",
                      "format": "date-time",
                      "description": "Timestamp of when the user was created"
                    }
                  }
                }
              }
            }
          },
          "responses": {
            "200": {
              "description": "Webhook received successfully"
            }
          }
        }
      }
    },
    "paths": {
      "/users": {
        "get": {
          "tags": ["users"],
          "summary": "OPERATION IN HIDDEN TAG",
          "description": "Returns a list of users",
          "operationId": "getUsers",
          "responses": {
            "200": {
              "description": "Successful operation",
              "content": {
                "application/json": {
                  "schema": {
                    "type": "array",
                    "items": {
                      "type": "object",
                      "properties": {
                        "id": {
                          "type": "integer"
                        },
                        "name": {
                          "type": "string"
                        }
                      }
                    }
                  }
                }
              }
            }
          }
        },
        "post": {
          "tags": ["foobar"],
          "summary": "Get all users",
          "description": "Returns a list of users",
          "operationId": "getUserssd",
          "responses": {
            "200": {
              "description": "Successful operation",
              "content": {
                "application/json": {
                  "schema": {
                    "type": "array",
                    "items": {
                      "type": "object",
                      "properties": {
                        "id": {
                          "type": "integer"
                        },
                        "name": {
                          "type": "string"
                        }
                      }
                    }
                  }
                }
              }
            }
          }
        },
        "patch": {
          "summary": "HIDDEN OPERATION",
          "description": "Returns a list of users",
          "operationId": "getUserssdasdasdas",
          "x-internal": true,
          "responses": {
            "200": {
              "description": "Successful operation",
              "content": {
                "application/json": {
                  "schema": {
                    "type": "array",
                    "items": {
                      "type": "object",
                      "properties": {
                        "id": {
                          "type": "integer"
                        },
                        "name": {
                          "type": "string"
                        }
                      }
                    }
                  }
                }
              }
            }
          }
        }
      }
    }
}
```


packages/api-client/src/views/Request/RequestSidebarItem.vue
```diff
@@ -287,10 +288,25 @@ const hasDraftRequests = computed(() => {
     item.value.children.length > 0
   )
 })
+
+/**
+ * Check if the item should be shown.
+ * This is used to hide items that are marked as hidden/internal.
+ */
+const shouldShowItem = computed(() => {
```
amritk, line 296: 😍

amritk (approved):
Love the PR! can we add some tests?

PR #4420, opened 2025-01-09
feat(api-reference): one endpoint tag collapsibility

### LINKED ISSUE #4409: Single endpoint does not display show more button
### What happens?

we've noticed that the Show More button is not displayed when there is only one endpoint for a given tag.

image

From a UX standpoint, this could be confusing for readers, especially in very long API References. It also breaks the homogeneity of the sidebar's behavior.

_originally shared by @mfrayssens-treezor in discussion [#4408 ](https://github.com/scalar/scalar/discussions/4408)_

### OpenAPI Document

_No response_

antlio:
**Problem**
Currently if a tag has only one endpoint in a multiple tag reference - it cannot be collapsible (e.g. show more button).

**Solution**
this pr fixes #4409 by enabling collapsibility to non default tag that have one endpoint, this way if a reference has no tag usage the non collapsibility remain.

| before | after |
| -------|------|
|  image | image |
| operations are displayed and are not collapsible | operations remain collapsed by default | 

**Checklist**

### DIFF
--- a/.changeset/shaggy-peas-begin.md
+++ b/.changeset/shaggy-peas-begin.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
+fix: adds collabsibility to one endpoint non default tag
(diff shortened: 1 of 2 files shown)


hwkr (approved):
🔥

PR #4426, opened 2025-01-10
fix(api-reference): schema example duplication

antlio:
**Problem**
a duplication has been introduced in #4162 for schema property example, being now shown twice.

**Solution**
this pr removes the duplication and update the style for array value.

| before | after |
| -------|------|
| image | image |

**Checklist**

- [x] I’ve added a changeset (`pnpm changeset`).s


hanspagel (commented):
Just noticed this today, and before I was able to create the issue there’s a PR, you made my day!

PR #4431, opened 2025-01-10
fix(api-client): linux enter hotkey

amritk:
**Problem**
ctrl + enter doesnt work on linux anymore

**Solution**
ideally we would remove whenever + magic keys as they have caused issues in the past, but this is a quick fix that works

Update: did the proper fix and removed magic keys + whenever

**Checklist**

### DIFF
--- a/.changeset/tame-news-grow.md
+++ b/.changeset/tame-news-grow.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
+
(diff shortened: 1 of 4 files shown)


marclave (approved):

marclave (approved):

hwkr (approved):
Looks good! +1 for using the event bus

PR #4433, opened 2025-01-10
fix(api-client): duplicate query params

### LINKED ISSUE #4428: Address bar query parameter value duplicated
### What happens?

Using the web version of 
@scalar/hono-api-reference“:”^0.5.165”

Maybe it's a bug: if you send a query with query parameters in the url, for example, like this 
/test?value=1

This query is sent as 
/test?value=1&value=1

If the parameters are written in the interface, everything works fine

image

_originally shared [in discord](https://discord.com/channels/1135330207960678410/1135330208627564597/threads/1327270303880974366) by Leraft_

### OpenAPI Document

_No response_

amritk:
**Problem**
Query params are duplicated when they are in the path string

**Solution**
We remove any query params from the path string

**Checklist**

closes #4428

### DIFF
--- a/.changeset/mighty-penguins-invite.md
+++ b/.changeset/mighty-penguins-invite.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
+
+fix: duplicate query params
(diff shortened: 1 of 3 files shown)


antlio (changes requested):
issue seems to be remaining using in the url `https://void.scalar.com/path?query=param`:

image

antlio (approved):
added a commit / test for the empty server url use case

PR #4435, opened 2025-01-11
feat(components): create dark light toggle components

hwkr:
Extracts the dark light toggle and sidebar footer from the references into the component library and standardizes the functionality.

Adds the following components:

- `ScalarColorModeToggle`: An all-in-one toggle that uses the `useColorMode` hook and sets the color mode on toggle, can be the button or an icon. 
- `ScalarColorModeToggleButton`: Just the toggle button with a model, for more complex use cases where you need to hook it up yourself you can connect the model.
- `ScalarColorModeToggleIcon`: Just the toggle icon with a control state, for more complex use cases where you need to hook it up yourself you can bind a click event.
- `ScalarSidebar`: A barebones sidebar component, the idea is we'll update it when we align our sidebars a bit better. Could also bring the drag to resize code in here.
- `ScalarSidebarFooter`: The footer that is currently in the references with the Powered by Scalar link and the toggle. Both can be overridden with slots.

This also updates the consumption of the toggles in the references (both the modern and classic layouts) but we have to use the base components (`ScalarColorModeToggleButton` and `ScalarColorModeToggleIcon`) because of how we set up the `forceDarkModeState` config.

https://github.com/user-attachments/assets/63891978-033e-423d-a908-8777d732f0c3

**Checklist**


packages/components/src/components/ScalarColorModeToggle/ScalarColorModeToggleIcon.vue
```diff
@@ -0,0 +1,106 @@
...
+    <span class="toggle-icon-sun-ray" />
+    <span class="toggle-icon-sun-ray" />
+    <span class="toggle-icon-sun-ray" />
+    <span class="toggle-icon-sun-ray" />
+    <span class="toggle-icon-ellipse">
+      <span class="toggle-icon-moon-mask" />
+    </span>
+  </component>
+</template>
+<style scoped>
```
hwkr, line 47: I tried to rewrite these styles as both tailwind and an svg animation and eventually gave up

hanspagel (approved):
OMG, great work! Just tested the functionality in the reference and skimmed through the code, looks great!

packages/components/src/components/ScalarColorModeToggle/ScalarColorModeToggleIcon.vue
amritk: this is the way, when its complex just go to css

amritk (approved):
Code looks good, good comments, good docs



Did not test as it looks like hans did already

cameronrohani (approved):

PR #4439, opened 2025-01-13
fix(api-client): sets integration theme id fallback

antlio:
**Problem**
currently importing a reference to the client creates schema type issue (e.g. with hono).

**Solution**
this pr fallback to the default theme id if the integration theme is not supported / typed.
 
**Checklist**

### DIFF
--- a/packages/api-client/src/components/ImportCollection/ImportCollectionModal.vue
+++ b/packages/api-client/src/components/ImportCollection/ImportCollectionModal.vue
@@ -14,7 +14,11 @@ import { ScalarIcon, ScalarModal, useModal } from '@scalar/components'
 import { isLocalUrl } from '@scalar/oas-utils/helpers'
 import { normalize } from '@scalar/openapi-parser'
 import type { OpenAPI } from '@scalar/openapi-types'
-import { type IntegrationThemeId, getThemeStyles } from '@scalar/themes'
+import {
+  type IntegrationThemeId,
+  getThemeStyles,
+  themeIds,
+} from '@scalar/themes'
 import { useColorMode } from '@scalar/use-hooks/useColorMode'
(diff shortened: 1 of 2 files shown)


hanspagel (approved):
Didn’t test (yet), but looks good.

packages/api-client/src/components/ImportCollection/ImportCollectionModal.vue
```diff
@@ -187,11 +191,18 @@ const themeStyleTag = computed(
 )
 
 function handleImportFinished() {
-  if (shouldShowIntegrationIcon.value && props.integration) {
+  // If the integration is not a valid theme id, set the theme to default
+  const integrationThemeId = themeIds.includes(
+    props.integration as IntegrationThemeId,
+  )
+    ? props.integration
+    : 'default'
```
amritk, line 199:
```ts
const isIntegrationThemeId = (value: unknown): value is IntegrationThemeId => 
  themeIds.includes(value as IntegrationThemeId);
```

```suggestion
  // If the integration is not a valid theme id, set the theme to default
const isIntegrationThemeId = (value: unknown): value is IntegrationThemeId => 
  themeIds.includes(value as IntegrationThemeId);

const integrationThemeId = isIntegrationThemeId(props.integration) 
  ? props.integration 
  : 'default';
```
lets move away from using `as` and instead use a type guard here

@geoffgscott should we ban the use of `as`?

antlio: yes better approach to prevent runtime error actually

amritk (approved):
I would have used `value: string` but still works

antlio:
> I would have used `value: string` but still works

though you had an opinion here - updated >>

PR #4442, opened 2025-01-13
feat(oas-utils): improve Zod schema for the OpenAPI Info Object

hanspagel:
**Problem**

Currently, the API definition parsing is failing when a contact without a (valid) email address is passed. This is very unfortunate.

**Solution**

With this PR we just omit invalid values for emails or URLs (when the specification says it MUST be an email/URL).

I’ve gone through the whole _Info Object_ and made it handle invalid data better.

**Additional changes**
* If the `info` object isn’t provided, we’ll add an `info` object with just the title now.
* The fallback for the required title is “API” now (instead of “OpenAPI Spec”).
* We don’t need the OpenAPI 3.0 type after running the definition through the upgrader.

**Checklist**


packages/oas-utils/src/transforms/import-spec.ts
```diff
@@ -134,12 +132,12 @@ export async function importSpecToWorkspace(
...
       securitySchemes: SecurityScheme[]
     }
-  | { error: true; importWarnings: string[] }
+  | { error: true; importWarnings: string[]; collection: undefined }
 > {
   const { schema, errors } = await parseSchema(spec, { shouldLoad })
   const importWarnings: string[] = [...errors.map((e) => e.message)]
 
-  if (!schema) return { importWarnings, error: true }
+  if (!schema) return { importWarnings, error: true, collection: undefined }
```
hanspagel, line 140: there might be a better solution, but without it TS is complaining when I write tests like this:

```
expect(result.collection?.info).toEqual({
  title: 'API',
})
```

packages/oas-utils/src/entities/spec/spec-objects.ts
```diff
@@ -42,7 +42,7 @@ export const oasContactSchema = z.object({
  */
 export const oasInfoSchema = z.object({
   /** REQUIRED. The title of the API. */
-  title: z.string().optional().default('OpenAPI Spec'),
+  title: z.string().catch('API'),
```
hanspagel, line 45: It’s required, but has a fallback for any invalid/undefined value.

amritk: Oh so optional + default actually has to do with the types. Optional means you don't need to pass it in the input, default means it always exists after parse. Maybe we just slap the catch on there? Not sure how it affects the types

amritk: https://www.gybex.dev/blog/zod-default-and-catch-methods

amritk: I don't think adds any value in this case according to the blog

amritk (changes requested):
Requesting changes as I believe we should use optional + default over catch in these specific cases

packages/oas-utils/src/transforms/import-spec.ts
amritk: yes in your test you need to pass the type guard first! 

https://github.com/scalar/scalar/blob/main/packages/api-client/src/libs/send-request/create-request-operation.test.ts#L132

we should actually switch to using this error response style here as well

packages/oas-utils/src/transforms/import-spec.ts
```diff
@@ -78,9 +78,7 @@ export const parseSchema = async (
      * Temporary fix for the parser returning an empty array
      * TODO: remove this once the parser is fixed
      */
-    schema: (Array.isArray(schema) ? {} : schema) as
-      | OpenAPIV3.Document
-      | OpenAPIV3_1.Document,
+    schema: (Array.isArray(schema) ? {} : schema) as OpenAPIV3_1.Document,
```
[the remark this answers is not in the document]
amritk: I need to fix the openapi types so we don't need to do this

hanspagel:
There’s a big difference, though:

* `optional.default('API')` throws an error when giving it invalid data (e.g. title: true)
* `catch(undefined)` just omits invalid data, it makes the schema more forgiving, but still keeps the data clean

packages/oas-utils/src/entities/spec/spec-objects.ts
amritk: alright yea I confirmed it, with .catch alone we don't get the proper types on output. we should use both `.catch` and `.default`

packages/oas-utils/src/entities/spec/spec-objects.ts
```diff
@@ -108,10 +128,13 @@ export const oasTagSchema = z.object({
...
-})
+export const tagSchema = oasTagSchema
+  .extend({
+    uid: nanoidSchema,
+    children: nanoidSchema.array().default([]),
+  })
+  .transform(omitUndefinedValues)
 
 export type Tag = z.infer<typeof tagSchema>
+/** @deprecated Use `Tag` instead */
```
amritk, line 139:
```suggestion
```

amritk: we need both

amritk (approved):
Minor things remain

packages/api-client/src/views/Request/libs/watch-mode.ts
```diff
@@ -126,6 +126,7 @@ const unwrapSchema = (schema: ZodSchema): ZodSchema => {
   if (schema instanceof z.ZodOptional) return unwrapSchema(schema.unwrap())
   if (schema instanceof z.ZodDefault) return unwrapSchema(schema._def.innerType)
   if (schema instanceof z.ZodEffects) return unwrapSchema(schema._def.schema)
+  if (schema instanceof z.ZodCatch) return unwrapSchema(schema._def.innerType)
```
amritk, line 129: @hanspagel and the zod traverser gets even better, maybe we should pull this out of watch mode and into its own lib. Could be useful to others

amritk (approved):

PR #4445, opened 2025-01-13
feat(api-reference): remove examples from responses

### LINKED ISSUE #4438: TypeError: e.trim is not a function in 200 array response
### What happens?

The error:
```
Oops, something went wrong here.
TypeError: e.trim is not a function
```

1. Go to https://docs.blockfrost.io/#tag/cardano--network/GET/network/eras
2. Click on `Expand 200 Returns era summaries content.`

Error in 200 response only if it's an array
```yaml
type: array
items:
  type: object
  properties:
    time:
      type: integer
      description: Starting time of the call count interval (ends midnight UTC) in UNIX time
    calls:
      type: integer
      description: Sum of all calls for a particular day
  required:
    - time
    - calls
example:
  - time: 1612543884
    calls: 42
  - time: 1614523884
    calls: 6942

```

Show the response

### OpenAPI Document

_No response_

hanspagel:
**Problem**
As a response to #1971 we’ve added the examples to all sections in the operation in #4162.

Turns out, it doesn’t really make sense to have the examples in the response list. It’s redundant with the box on the right side of the operation, and the formatting helper wasn’t really meant to deal with that kind of data.

Fixes #4438

**Solution**
With this PR the response list on the left doesn't show the example responses anymore and  the formatting helper is moved to a separate helper, got some tests and some better handling for other data types.

**Before**

Screenshot 2025-01-13 at 16 38 13

**After**

Screenshot 2025-01-13 at 16 26 23

**Checklist**

### DIFF
--- a/packages/api-reference/src/components/Content/Schema/helpers/formatExample.ts
+++ b/packages/api-reference/src/components/Content/Schema/helpers/formatExample.ts
@@ -1,7 +1,7 @@
 /**
  * Converts an example value to a string that can be displayed in the UI.
  */
-export function formatExample(example: any) {
+export function formatExample(example: unknown) {
   if (Array.isArray(example)) {
     return `[${example
       .map((item) => {
--- a/packages/api-reference/src/components/Content/Schema/helpers/formatExample.ts
+++ b/packages/api-reference/src/components/Content/Schema/helpers/formatExample.ts
@@ -1,7 +1,7 @@
 /**
  * Converts an example value to a string that can be displayed in the UI.
  */
-export function formatExample(example: unknown) {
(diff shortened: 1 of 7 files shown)


packages/api-reference/src/components/Content/Schema/helpers/formatExample.ts
```diff
@@ -0,0 +1,46 @@
+/**
+ * Converts an example value to a string that can be displayed in the UI.
+ */
+export function formatExample(example: unknown): string | number {
```
antlio, line 4: digging this

antlio (approved):
everything seems back on track on my end! great coverage 🐐

PR #4450, opened 2025-01-14
feat(api-client): adds servers page back

antlio:
**Problem**
it is not currently possible to add, update or delete server per collection.

**Solution**
this pr brings back initial work done by @hanspagel in a global server page. server variables including enum will follow along request server management that is currently under design tinkering.

| after |
| -------|
| image |
| server page brought back along add, update, delete capability |

**Checklist**

- [x] I’ve added a changeset (`pnpm changeset`)

### DIFF
--- a/.changeset/long-beans-punch.md
+++ b/.changeset/long-beans-punch.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
+
+feat: adds server page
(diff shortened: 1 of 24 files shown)


antlio:
we might want to add an empty state in that page when no collection exists cc @cameronrohani

marclave (approved):
AWESOME! lets get this in @antlio 

@cameronrohani let's follow & get @kelseyll to take a look at empty states across our api client ✨

PR #4452, opened 2025-01-14
feat(api-reference): version prefix

hanspagel:
**Problem**
Currently, a lot of people just having something like `1` or `2` as the version in their API definition, but I think the rendering might be a bit confusing. The number is just shown without any context.

Screenshot 2025-01-14 at 12 07 53

**Solution**
With this PR it’s rendered as `v1`, which might be enough to make it look like a version number.

Screenshot 2025-01-14 at 12 07 39

Wdyt?

**Checklist**

### DIFF
--- a/packages/api-reference/src/components/Content/Introduction/Introduction.vue
+++ b/packages/api-reference/src/components/Content/Introduction/Introduction.vue
@@ -38,10 +38,8 @@ const formattedSpecTitle = computed(() => {
     <Section class="introduction-section">
       <SectionContent :loading="!info.description && !info.title">
         <div class="badges">
-          <Badge v-if="info.version">
-            {{ info.version }}
-          </Badge>
-          <Badge v-if="specVersion"> OAS {{ specVersion }}</Badge>
+          <Badge v-if="info.version">v{{ info.version }}</Badge>
+          <Badge v-if="specVersion">OAS {{ specVersion }}</Badge>
         </div>
         <SectionHeader
           :level="1"
--- a/packages/api-reference/src/components/Content/Introduction/Introduction.vue
+++ b/packages/api-reference/src/components/Content/Introduction/Introduction.vue
@@ -42,11 +42,11 @@ const filenameFromTitle = computed(() => slugger.slug(props.info?.title ?? ''))
 const version = computed(() => {
   // Prefix the version with “v” if the first character is a number, don’t prefix if it’s not.
   // Don’t output anything when version is not a string.
-  return props.info?.version
-    ? `v${props.info.version.toString().match(/^\d/)?.input}`
-    : typeof props.info?.version === 'string'
-      ? props.info.version
-      : undefined
+  return typeof props.info?.version === 'string'
+    ? props.info.version.toString().match(/^\d/)
+      ? `v${props.info.version}`
+      : props.info.version
+    : undefined
 })
 </script>
 <template>
--- a/packages/api-reference/src/components/Content/Introduction/Introduction.vue
+++ b/packages/api-reference/src/components/Content/Introduction/Introduction.vue
@@ -1,7 +1,7 @@
 <script setup lang="ts">
 import type { OpenAPIV2, OpenAPIV3, OpenAPIV3_1 } from '@scalar/openapi-types'
 import type { Spec } from '@scalar/types/legacy'
-import GithubSlugger from 'github-slugger'
+import GitHubSlugger from 'github-slugger'
 import { computed } from 'vue'
 
 import DownloadLink from '../../../features/DownloadLink/DownloadLink.vue'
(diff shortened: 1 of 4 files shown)


antlio (approved):
great addition

packages/api-reference/src/components/Content/Introduction/Introduction.vue
```diff
@@ -38,10 +38,8 @@ const formattedSpecTitle = computed(() => {
     <Section class="introduction-section">
       <SectionContent :loading="!info.description && !info.title">
         <div class="badges">
-          <Badge v-if="info.version">
-            {{ info.version }}
-          </Badge>
-          <Badge v-if="specVersion"> OAS {{ specVersion }}</Badge>
+          <Badge v-if="info.version">v{{ info.version }}</Badge>
```
amritk, line 41:
```suggestion
          <Badge v-if="info.version">v{{ info.version.replace(/^v/, '') }}</Badge>
```
should we do a quick check in case they already have a v? to avoid a `v v 1`

amritk (approved):
approved with minor change

packages/api-reference/src/components/Content/Introduction/Introduction.vue
hanspagel: Man, that’s smart!

Let us take a different approach: We should check if the first string is a number, otherwise we could end up with `vstaging` and other funny things.

Also, currently it’s possible a non-string value lands there, so we need to make sure it’s a string.

I’ll add this tomorrow!

hanspagel:
Updated the component:

* to not prefix the version when the first character isn't a number (e.g. `beta`, `staging` or whatever),
* not show a version if it’s not a string, but whatever other data type
* add tests for the whole Introduction component
* remove unused CSS
* replaced one CSS class with tailwind (gap: 48px > gap-12)
* renamed some variables
* added some components

antlio:
should we add test and support on `v1.0` usage for instance?

packages/api-reference/src/components/Content/Introduction/Introduction.vue
```diff
@@ -23,33 +23,48 @@ const props = defineProps<{
...
-const slugger = new GithubSlugger()
-
-const specVersion = computed(() => {
-  return props.parsedSpec.openapi ?? props.parsedSpec.swagger ?? ''
+/**
+ * Get the OpenAPI/Swagger specification version from the API definition.
+ */
+const oasVersion = computed(() => {
+  return props.parsedSpec?.openapi ?? props.parsedSpec?.swagger ?? ''
 })
```
amritk, line 31:
```suggestion
const oasVersion = computed(() => props.parsedSpec?.openapi ?? props.parsedSpec?.swagger ?? '')
```
the beauty of arrow functions is no need to write return!

marclave (approved):

packages/api-reference/src/components/Content/Introduction/Introduction.vue
```diff
@@ -23,33 +23,48 @@ const props = defineProps<{
...
+
+/** Format the version number to be displayed in the badge */
+const version = computed(() => {
+  // Prefix the version with “v” if the first character is a number, don’t prefix if it’s not.
+  // Don’t output anything when version is not a string.
+  return typeof props.info?.version === 'string'
+    ? props.info.version.toString().match(/^\d/)
+      ? `v${props.info.version}`
+      : props.info.version
+    : undefined
```
amritk, line 49:
```suggestion
props.info?.version?.toString().startsWith('0-9') ? `v${props.info.version}` : props.info?.version
```
this version will handle numbers instead of returning undefined. completely up to you though!

packages/api-reference/src/features/DownloadLink/DownloadLink.vue
```diff
@@ -11,8 +11,14 @@ const props = defineProps<{
   specTitle?: string
 }>()
 
-const getHideDownloadButtonSymbol = inject(HIDE_DOWNLOAD_BUTTON_SYMBOL)
-const getOpenApiDocumentUrlSymbol = inject(OPENAPI_DOCUMENT_URL_SYMBOL)
+const getHideDownloadButtonSymbol = inject(
```
amritk, line 14: I removed these in my other pr

amritk (approved):
lol im commenting on this pr then i see its already merged!

packages/api-reference/src/components/Content/Introduction/Introduction.vue
hanspagel: `startsWith(0-9)` checks for the literal `0-9`, not for numbers between 0 and 9 😬 

and without the string check it would output arrays, objects, booleans … (which no one should pass, but doesn’t hurt to add the check)

PR #4453, opened 2025-01-14
chore(api-reference): remove embeds

hanspagel:
I’m splitting #4281 in a few smaller PRs.

Let’s remove the old embeds. The work was still valuable, because the components that we use in the reference were decoupled from a lot of things, but we don’t need the actual embeds anymore. The Scalar Blocks will be way better.


amritk (approved):
Noice!

PR #4454, opened 2025-01-14
feat(openapi-types): add HTTP methods

hanspagel:
I’m splitting https://github.com/scalar/scalar/pull/4281 in a few smaller PRs.

**Problem**
Currently, we have a few places where keep a type for the HTTP methods.

**Solution**
With this PR we’re adding them to the `@scalar/openapi-types` package, so we’ll have a central place for it.

**Checklist**


amritk (approved):
Sure

PR #4456, opened 2025-01-14
chore(tooling): remove old cloudbuild yaml file

tmastrom:
The cloudbuild file is now a json file and it is auto-generated on commit. This old yaml file is no longer in use


marclave (approved):
nice :)

Issue #4427, opened 2025-01-10
Headers not showing when sending request ASP.NET


fnferrero97:
### What happens?

In Scalar, I can see the headers
![image](https://github.com/user-attachments/assets/f3ea0537-7954-4fbe-8af1-f7e175572f5f)

But in the submit popup, they´re not being put in automatically. No default value, I have to manually add them.
![image](https://github.com/user-attachments/assets/ea83bd09-f45e-43b6-98ca-eebc803e8e3d)

### What did you expect to happen?

For the custom header show in the popup

antlio:
hey @fnferrero97 thank you for reporting this - should be fixed by #4460 and will be released soon. see #4527 - feel free to reach out here again if you still encounter the issue

PR #4460, opened 2025-01-14
fix(api-client): request table tooltip

### LINKED ISSUE #4418: The header parameter does not appear in the ui
### What happens?

The header parameter is not visible when trying to send a request.

image

image

I'm guessing that the error may be caused by the `Content-Type` being added rigidly because when you click clear the parameter is visible.

You can repeat this with the online editor and paste the json

https://docs.scalar.com/editor

```json
{
  "openapi": "3.0.1",
  "info": {
    "title": "xxxx.Services.xxxx | v1",
    "version": "1.0.0"
  },
  "servers": [
    {
      "url": "http://localhost:5218"
    }
  ],
  "paths": {
    "/api/webhooks/xxxx": {
      "post": {
        "tags": [
          "xxxx.Services.xxxx"
        ],
        "parameters": [
          {
            "name": "x-xxxx-signature",
            "in": "header",
            "description": "Custom header for xxxx signature",
            "required": true,
            "schema": {
              "type": "string"
            }
          }
        ],
        "responses": {
          "200": {
            "description": "OK"
          }
        }
      }
    }
  },
  "components": { },
  "tags": [
    {
      "name": "xxxx.Services.xxxx"
    }
  ]
}
```

The header parameter is visible and changeable when attempting to send an HTTP request via the UI.

### OpenAPI Document

https://sandbox.scalar.com/e/IcP-z

antlio:
**Problem**
request table tooltip encountering null item is breaking the client as highlighted in #4418 

**Solution**
this pr fixes #4418 #4427 by updating the request table tooltip logic. still works with #4387

| before | after |
| -------|------|
| image | image |
| header parameter is not displayed  | header is displayed along its tooltip accordingly | 

**Checklist**

### DIFF
--- a/packages/api-client/src/views/Request/RequestSection/RequestTable.vue
+++ b/packages/api-client/src/views/Request/RequestSection/RequestTable.vue
@@ -73,7 +73,7 @@ const flattenValue = (item: RequestExampleParameter) => {
     :columns="columns">
     <DataTableRow
       v-for="(item, idx) in items"
-      :key="idx">
+      :key="item.key">
       <label class="contents">
         <span class="sr-only">Row Enabled</span>
         <DataTableCheckbox
--- a/packages/api-client/src/views/Request/RequestSection/RequestTable.vue
+++ b/packages/api-client/src/views/Request/RequestSection/RequestTable.vue
@@ -8,6 +8,7 @@ import { ScalarButton, ScalarIcon } from '@scalar/components'
 import type { RequestExampleParameter } from '@scalar/oas-utils/entities/spec'
 import { computed } from 'vue'
 
+import { hasItemProperties } from '../libs/request'
 import RequestTableTooltip from './RequestTableTooltip.vue'
 
 withDefaults(
@@ -43,10 +44,6 @@ const handleSelectVariable = (
 
 const handleFileUpload = (idx: number) => {
   emit('uploadFile', idx)
-}
-
-const showTooltip = (item: RequestExampleParameter) => {
-  return !!(item.description || item.type || item.default || item.format)
 }
 
 const valueOutOfRange = (item: RequestExampleParameter) => {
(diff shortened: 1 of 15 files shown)


amritk:
The issue was using index as a key which 99% of the time is a big no no, so I replaced all the ones I found. eslint rules inc

packages/api-client/src/views/Request/RequestSection/RequestTable.vue
```diff
@@ -8,6 +8,7 @@ import { ScalarButton, ScalarIcon } from '@scalar/components'
 import type { RequestExampleParameter } from '@scalar/oas-utils/entities/spec'
 import { computed } from 'vue'
 
+import { hasItemProperties } from '../libs/request'
```
hanspagel, line 11: can we use the @/ alias here?

hanspagel (approved):
rad

packages/api-client/src/views/Request/RequestSection/RequestTable.vue
amritk: only if its needed/used elsewhere. if its only used here then its fine here.

packages/api-client/src/views/Request/libs/request.ts
```diff
@@ -0,0 +1,21 @@
...
+export const hasItemProperties = (item: RequestExampleParameter): boolean => {
+  return !!(
+    item.description ||
+    item.type ||
+    item.default ||
+    item.format ||
+    item.minimum ||
+    item.maximum
+  )
+}
```
amritk, line 21:
```suggestion
export const hasItemProperties = (item: RequestExampleParameter) => 
  Boolean(
    item.description ||
    item.type ||
    item.default ||
    item.format ||
    item.minimum ||
    item.maximum
  )
```
I would write it like this personally but functionally is the same

amritk (approved):

packages/api-client/src/views/Request/libs/request.ts
antlio: I agree it's more clear too

PR #4462, opened 2025-01-14
feat(api-reference): add URL customization functions

amritk:
**Problem**
We had one way to set the URL hash `tag/${tagname}/${httpVerb}${path}`

**Solution**
We allow customizing each part of the hash using the new custom ID functions (open to better naming). There is a limitation for these, we can only use them when passing in the config as JS. When using the dom data attributes like with standalone it is a security risk since we would need to eval and execute a string (however we can add this if we want).

So the best way to test is in `api-client-react` 
```tsx
// packages/api-reference-react/playground/App.tsx

<ApiReferenceReact
  configuration={{
    customOperationId: (operation) => `customOp-${operation.operationId}`,
    customTagId: (tag) => `weLoveTags-${tag.name}`,
    customModelId: (model) => `moooooodles-${model}`,
    spec: {
      content: galaxySpec,
    },
    authentication: auth,
  }}
/>
```
`pnpx turbo playground --filter @scalar/api-reference-react`

Now in this PR I also moved those provides into a single one for the whole config object and exposed it through the `useConfig` hook. Since the `useSidebar` hook uses the old global JS style I had to do some tomfoolery there. BUT it works and when we replace this sidebar with the client one it will no longer be an issue. 

**Checklist**


packages/api-reference/src/components/ApiReferenceLayout.vue
```diff
@@ -277,20 +270,8 @@ provide(WORKSPACE_SYMBOL, workspaceStore)
...
-  () => props.configuration.hideTestRequestButton,
-)
-provide(OPENAPI_DOCUMENT_URL_SYMBOL, () => props.configuration.spec?.url)
-provide(INTEGRATION_SYMBOL, () =>
-  props.configuration._integration !== null
-    ? props.configuration._integration
-    : 'vue',
-)
+// Provide the configuration
+provide(CONFIGURATION_SYMBOL, props.configuration)
```
amritk, line 274: Provide a single time for the config now

packages/types/src/legacy/reference-config.ts
```diff
@@ -175,6 +175,62 @@ export type ReferenceConfiguration = {
    * @default undefined
    */
   pathRouting?: PathRouting
+  /**
```
amritk, line 178: these good @hanspagel ?

hanspagel: `customModelId` (and the others) sound like you could pass a specific id (as a string or something).

Maybe we use a `generate*` prefix to indicate we expect a callback. And I’m used to refer to those route segments as slugs. So here’s an alternative suggestion:

`generateModelSlug`

Wdyt? Same applies to the others of course.

packages/api-reference/src/hooks/useSidebar.ts
```diff
@@ -371,6 +374,11 @@ export const scrollToOperation = (operationId: string) => {
 export function useSidebar(options?: ParsedSpecOption & SorterOption) {
   Object.assign(optionsRef, options)
 
+  // Hack navState
```
hanspagel, line 377: can you add a little bit more explanation for future us? why is that a hack, what do we need to remove it?

packages/types/src/legacy/reference-config.ts
amritk: Love it! I figured we would have to rename em lol

amritk: maybe overrideModelSlug? Since we have a generation function already? Or does having the same name imply override

hanspagel: I’d prefer generate, but I’m okay with both!

packages/api-reference/src/components/ApiReferenceLayout.vue
hanspagel: Finally!

hanspagel (changes requested):
Tested and works great!

A few things:

- [ ] The whole tag and the whole operation is passed, should we pass the whole model/schema, too?
- [x] We pass the horrible `TransformedOperation`, which we are about to remove. What if we pass `operation.information` which should also satisfy the new `Request` type (that we use in the store)?
- [ ] We probably need to add this for the navigation entries added from the description and the webhooks, too.
- [x] Yeah, the discussed names would be great. :v:

packages/api-reference/src/hooks/useSidebar.ts
amritk: oh i did above where navState is defined

amritk: https://github.com/scalar/scalar/pull/4462/files#diff-177327574212cd0b7483ff6ea4da60e444da454855f260a9a5a2768b7a83a297R30

amritk:
I just checked and models and webhooks are only names sometimes

hanspagel:
> I just checked and models and webhooks are only names sometimes

What does that mean? 🤔 We don’t have the whole object sometimes?

amritk:
> > I just checked and models and webhooks are only names sometimes
> 
> What does that mean? 🤔 We don’t have the whole object sometimes?

correct, we only have the name and would have to a lookup to get the object

amritk:
alright @hanspagel I think now it will be easy to change when we switch to entities

hanspagel (approved):
LOVE IT

PR #4464, opened 2025-01-15
fix(api-client): favors configuration integration in modal client button

antlio:
**Problem**
currently the client button in the reference client button is relaying on collection integration only and not on the reference configuration, as used in the sidebar button. in some case the client button query is not the same in the reference and in the client.

**Solution**
this pr sets the usage of the reference configuration integration in the client button in modal to ensure value usage parity between reference and client.

**Checklist**


hanspagel (approved):
Yesss!

marclave (approved):
based

PR #4465, opened 2025-01-15
feat(Scalar.AspNetCore): get all API definitions dynamically

mohsen2hasani:
Add openApi for all available versions (get api versions dynamically)

**Problem**
It's static and just defined v1 and v2

**Solution**
with a reflection on all controllers, we get available api versions


xC0dex:
Thank you for your pull request.
The current solution is based on the [eShop reference project](https://github.com/dotnet/eShop/blob/main/src/eShop.ServiceDefaults/OpenApi.Extensions.cs#L61) and it works without using Reflection. Since this project is meant to be a playground, I believe that the static solution is sufficient for our needs.

PR #4466, opened 2025-01-15
chore: new integrations/ folder

hanspagel:
**Problem**
Currently, the packages folder is a wild mix of all kinds of things. This also means, we have to build them more often than necessary. 

**Solution**
Let’s pull the integrations out and move them to a separate folder, so we can easily build all packages - without the integrations.

Some integrations are missing (nuxt, fastify), because we’re waiting for PRs to be merged. I’ll ff with a PR for those.

> ⚠️ After merge: Keep an eye on main to check whether the Docker build is good.

**Checklist**


pnpm-workspace.yaml
```diff
@@ -1,8 +1,9 @@
 packages:
   - 'examples/*'
+  - 'integrations/*'
+  - 'integrations/express-api-reference/playground'
```
geoffgscott, line 4: Should only need 1 of these

geoffgscott (commented):
Looks good but let's cleanup the naming convention of these folders at the same time. 

IMO we should just use the package we are integrating to as the name.

`express, hono, nestjs` etc

pnpm-workspace.yaml
hanspagel: We can use /**/* to simplify this!

packages/nuxt/playground/package.json
```diff
@@ -8,6 +8,6 @@
     "generate": "nuxi generate"
   },
   "dependencies": {
-    "nuxt": "^3.11.2"
+    "nuxt": "^3.12.4"
```
hanspagel, line 11: this was ignored by the previous workspace config

geoffgscott (approved):
:partying_face:

amritk:
@hanspagel should we fix tests and get this in?

hanspagel:
I try every day, but CI says no for a new reason every day. 😂

PR #4469, opened 2025-01-16
fix(api-client): settings route

antlio:
**Problem**
the settings route is accessible anymore as it has somehow been removed in #4450

**Solution**
this par adds back the missing settings route.

**Checklist**


hanspagel (approved):
Even comes with a few tests, nice!

Issue #4040, opened 2024-11-28
Read-only properties show in request body


adnweedon:
### What happens?

If I make a property in a schema read only, it stil shows when that schema is used as a request body, even though the property isn't relevant:
![image](https://github.com/user-attachments/assets/de122ea9-8e89-45ec-83ab-79c9516299ca)

I have made an example using the Pet Shop example spec, where I have made the Pet id field read only - note that it still shows for the Pet POST endpoint.

### What did you expect to happen?

I expected the read only properties not to be shown in the Body section, because they aren't relevant.

### OpenAPI Document

https://sandbox.scalar.com/#tag/pet/POST/pet

hanspagel:
Good catch! I think that was the case until recently, definitely an issue.

hanspagel:
Okay, I finally found time to test this, sorry for the super long wait. It turns out, we’re doing all we can, but those attributes are just not tagged as `read-only`/`write-only`.

I’m adding this to our example in #4477.

PR #4473, opened 2025-01-16
feat(api-reference): render operation parameters from the store

hanspagel:
I’m splitting https://github.com/scalar/scalar/pull/4281 in a few smaller PRs.

With this PR we’re rendering the operation parameters from the store. 🏄 

We don’t need the `formData` and the `body` section anymore, this was just to support Swagger 2.0, but the store upgrades to OpenAPI 3.1 on the fly anyway.

~~I’ve also made sure read-only attributes are not shown in the list of (request) parameters anymore (#4040).~~
Will submit a separate PR.

Look at how easy it becomes to test components. 🤯

```ts
it('renders query parameters', async () => {
  const wrapper = mount(OperationParameters, {
    props: {
      operation: {
        parameters: [
          {
            in: 'query',
            name: 'search',
            schema: {
              type: 'string',
            },
            required: false,
            deprecated: false,
          },
        ],
      },
    },
  })

  expect(wrapper.text()).toContain('Query Parameters')
  expect(wrapper.text()).toContain('search')
})
```

Screenshot 2025-01-16 at 10 52 13

**Checklist**


packages/api-reference/src/features/Operation/components/RequestBody.vue
```diff
@@ -4,54 +4,54 @@ import { ScalarMarkdown } from '@scalar/components'
 import type { ContentType, RequestBody } from '@scalar/types/legacy'
 import { computed, ref } from 'vue'
 
-const prop = defineProps<{ requestBody?: RequestBody }>()
+const props = defineProps<{ requestBody?: RequestBody }>()
```
amritk, line 7:
```suggestion
const { requestBody } = defineProps<{ requestBody?: RequestBody }>()
```
remember we can destructure props now! if you like

packages/api-reference/src/features/Operation/components/RequestBody.vue
```diff
@@ -4,54 +4,54 @@ import { ScalarMarkdown } from '@scalar/components'
...
 
 const contentTypes = computed(() => {
-  if (prop.requestBody?.content) {
-    return Object.keys(prop.requestBody.content)
+  if (props.requestBody?.content) {
```
amritk, line 10: The quick way to write this is 
```ts
const contentTypes = computed(() => Object.keys(props?.requestBody?.content ?? {})
```

amritk (approved):
Looks good! Did not test

PR #4474, opened 2025-01-16
ci: deploy client to staging by default

hanspagel:
**Problem**
Currently, we deploy everything that lands in main to `client.scalar.com` instantly. So we might break production more often.

**Solution**
With this PR we only deploy to production when a new version of `@scalar/api-client` is released, but deploy every commit in `main` to https://client.staging.scalar.com.

You can’t configure Cloudflare Pages to show different content on different domains. So I’ve created another project for staging.

**Checklist**


tmastrom (approved):
LGTM

amritk (approved):
smart

PR #4475, opened 2025-01-16
chore: move scalar-app to `projects/`

hanspagel:
**Problem**
Currently, we’re building the Electron app when running `pnpm build:packages`, that’s (mostly) not necessary.

**Solution**
With this PR the Scalar App is moved to the `projects/` folder, as a sibling to https://client.scalar.com. This should save build time in some cases.

**Checklist**


amritk (approved):

hanspagel:
IT JUST DOESN’T WORK SCREW THAT I WILL FIX THIS NEVER (okay, I’ll fix this soon)

PR #4477, opened 2025-01-16
feat(galaxy): add `readOnly`/`writeOnly` attributes

hanspagel:
**Problem**
Currently, the Galaxy examples shows read-only attributes (id) in the request, and write-only attributes (password) in the response.

**Solution**
With this PR we’re adding `readOnly: true` and `writeOnly: true` to attributes where applicable. This also made it possible to merge the `NewUser` and the `User` schemas.

Actually, I don’t know why I forgot about this when writing the initial document. 🤔 

**Checklist**

### DIFF
--- a/.changeset/grumpy-beans-sell.md
+++ b/.changeset/grumpy-beans-sell.md
@@ -1,0 +1,5 @@
+---
+'@scalar/galaxy': patch
+---
+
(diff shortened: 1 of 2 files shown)


amritk (approved):

hwkr (approved):
🫡

PR #4478, opened 2025-01-16
fix(mock-server): returns write-only attributes

hanspagel:
**Problem**
Currently, the mock server returns write-only attributes in the responses.

**Solution**
With this PR we’re generating the response with `mode: 'read'` to omit write-only attributes.

**Checklist**


amritk (approved):
not sure what exactly its doing but it has a test so it gets approved!

PR #4479, opened 2025-01-16
chore(use-codemirror): remove `@uiw/codemirror-themes`

### LINKED ISSUE #4222: CommonJS named export issue for @uiw/codemirror-themes packages.

hanspagel:
**Problem**

We have some CJS/ESM problems with `@uiw/codemirror-themes` for a while.

Fixes #4222

**Solution**

We’re using just a single file from the package, so let’s drop it into our code base. :)

I’ve also fixed the styling of the playground.

**Checklist**

### DIFF
--- a/packages/use-codemirror/playground/components/CodeMirror.vue
+++ b/packages/use-codemirror/playground/components/CodeMirror.vue
@@ -116,7 +116,7 @@ const { codeMirror, setCodeMirrorContent } = useCodeMirror({
   display: flex;
   gap: 0.5rem;
   margin-bottom: 1rem;
-  font-size: 14px;
+  font-size: var(--scalar-small);
 }
 
 button {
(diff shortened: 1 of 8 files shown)


packages/use-codemirror/playground/components/CodeMirror.vue
```diff
@@ -116,6 +116,7 @@ const { codeMirror, setCodeMirrorContent } = useCodeMirror({
   display: flex;
   gap: 0.5rem;
   margin-bottom: 1rem;
+  font-size: 14px;
```
antlio, line 119: if accessible: 
```suggestion
  font-size: var(--scalar-small);
```

amritk (approved):
wow hans what a smart move

PR #4485, opened 2025-01-16
feat(api-client): selected example request introduction

antlio:
**Problem**
when importing a specification document with multiple content type request, only the first content type is retrieved with no way to select the other one.

**Solution**
[POC] this pr is a proposal to import all content type request as example along example selection capability introduced in order to be able to navigate from one or another.

as different examples might be created on import - we need to validate a way to work with the imported one and the example created by users in the client - currently this is drafted by relying on a `source` field on the request example entity.

**Checklist**


amritk:
hey @antlio am I crazy, or is this working now?

antlio:
closing this ATM but to be re-open in a month 👀 in another clean state PR

Issue #4458, opened 2025-01-14
Vue/Nuxt Hydration Error


codeofsumit:
### What happens?

With the super basic vue example used in a nuxt3 project, there is an hydration error on page load:

```vue
<script setup lang="ts">
import { ApiReference } from '@scalar/api-reference';
import apiSpec from '@/api-specification/asset-api.json';
import '@scalar/api-reference/style.css';
</script>

<template>
  <div>
    <div class="flex items-center mt-8 pb-16 text-gray-600 crumbs">
      <nuxt-link class="capitalize" to="/">Home</nuxt-link>
      <nuxt-link class="capitalize" to="/docs">Docs</nuxt-link>
      <nuxt-link class="capitalize">API</nuxt-link>
    </div>
    <main class="p-0">
      <ApiReference :configuration="{ spec: { content: apiSpec } }" />
    </main>
  </div>
</template>
```

<img width="1495" alt="image" src="https://github.com/user-attachments/assets/7a2d9128-3065-493e-b5d4-3b41483bac3d" />

I'm aware there is a nuxt module for scalar but that has different issues and currently does not work due to [this issue](https://github.com/scalar/scalar/issues/4222). The Vue version works.

### What did you expect to happen?

No hydration errors.

hanspagel:
Ha @codeofsumit, didn’t know you guys are using Scalar! Thanks for the report, and the workaround. 🫶

codeofsumit:
> Ha @codeofsumit, didn’t know you guys are using Scalar! Thanks for the report, and the workaround. 🫶

We used it internally for a while and are now starting to use Scalar publicly for our first public API's. Thanks a lot for your work! 👍🚀

codeofsumit:
@hanspagel I see the label `@scalar/nuxt` - just a clarification that we are using `@scalar/vue` in a nuxt 3 project.

amritk:
Yea we often get hydration errors popping up as its primarily a client only project. We have been moving towards supporting more SSR though. What I need to do is get nuxt fixed then get some basic e2e tests going that look for any console warnings.

For fixing them its usually something pretty simple: converting the client side code to something SSR friendly.

codeofsumit:
@amritk good point. I can go client-only for now, however, I wanted to do SSR to have the API refs content ready for SEO as well.

amritk:
We should be writing some ssr/ssg tests shortly!

hanspagel:
[...]

PR #4486, opened 2025-01-16
test(api-reference): add SSR/SSG tests

hanspagel:
**Problem**
Currently, we don't have proper SSG/SSR tests and don’t even notice when we break SSG/SSR builds.

**Solution**
This PR adds two basic tests which render a super simple OpenAPI document and a big real-world OpenAPI document (GitHub’s REST API).

See #4458

**Checklist**

### DIFF
--- a/packages/api-reference/src/components/ApiReferenceLayout.test.ts
+++ b/packages/api-reference/src/components/ApiReferenceLayout.test.ts
@@ -1,6 +1,6 @@
 import { parse } from '@/helpers'
 import { renderToString } from '@vue/server-renderer'
-import { describe, expect, it } from 'vitest'
+import { describe, expect, it, vi } from 'vitest'
 import { createSSRApp, h } from 'vue'
 
 import ApiReferenceLayout from './ApiReferenceLayout.vue'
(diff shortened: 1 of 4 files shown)


packages/api-reference/src/components/ApiReferenceLayout.test.ts
```diff
@@ -0,0 +1,65 @@
...
+        }),
+    })
+
+    const html = await renderToString(app)
+
+    expect(html).toContain('Test API')
+  })
+
+  it('can render the GitHub REST API reference', async () => {
+    // Spy for console.error to avoid errors in the console
```
amritk, line 34: I would even add warnings too

amritk (approved):
Hans strikes again!

packages/api-reference/src/components/ApiReferenceLayout.test.ts
hanspagel: It’ll fail then, though. Some workspace schema work ahead. 🫡

amritk: kk just toss in a reminder somewhere to add it later!

packages/components/src/components/ScalarFloating/ScalarFloating.vue
```diff
@@ -37,15 +37,19 @@ const floatingRef: Ref<HTMLElement | null> = ref(null)
 const wrapperRef: Ref<HTMLElement | null> = ref(null)
 
 const targetRef = computed(() => {
-  // If target is a string (id), try to find it in the document
-  if (typeof props.target === 'string') {
-    const target = document.getElementById(props.target)
-    if (target) return target
-    else
-      console.warn(`ScalarFloating: Target with id="${props.target}" not found`)
+  if (typeof window !== 'undefined') {
```
hwkr, line 40: We can add an SSR unit test for ScalarFloating too if you'd like!

PR #4489, opened 2025-01-16
fix(api-reference): remove incorrect default security requirement

### LINKED ISSUE #4300: Code generator adds Authorization header
### What happens?

Hy,
I have a small problem with the Authorization header.
I am using .net9 Scalar.AspNetCore package and when I updated from 1.2.43 to 1.2.64 all my endpoints without security scheme get the request.AddHeader Authorization line.

1.2.43
image

1.2.64
image

I expect not to add the authorization header, because this endpoint has no security scheme defined. 

### OpenAPI Document

https://sandbox.scalar.com/e/7ZMuE

amritk:
fixes #4300 

**Problem**
We incorrectly set the default security requirement to optional

**Solution**
we set the default security requirement to an empty array (`[]`)

I also fixed the esm playground for easier spec testing

**Checklist**

### DIFF
--- a/packages/api-reference/playground/esm/index.html
+++ b/packages/api-reference/playground/esm/index.html
@@ -12,6 +12,7 @@
 
     <script type="module">
       import { createScalarReferences } from '../../src'
+      import test from './test.yml?raw'
 
       // CDN:
       // import { createScalarReferences } from 'https://esm.sh/@scalar/api-reference';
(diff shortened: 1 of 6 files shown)


packages/api-reference/playground/esm/index.html
hanspagel, line 1: this is a left-over, isn’t it?

amritk: leftover? I use it for testing lol

hanspagel (changes requested):
To my understanding, those are the meanings:

`security: [{}]` = requires authentication, use any security scheme
`security: []` = doesn’t require authentication, don’t use a security scheme

and if there’s no `security` attribute defined:

* if there’s no global security scheme defined: doesn’t required authentication
* if there is at least one global security scheme defined: requires authentication

With this PR we won’t stick to what the OpenAPI specification defines I think. :(

Can you further explain why you think we’d need this change?

amritk:
This is wrong!

PR #4492, opened 2025-01-17
fix(api-client): request body

antlio:
**Problem**
when selecting binary as a request body content type, raw value are being removed + style inconsistencies.

**Solution**
this pr prevents raw value from being removed on binary type selection in order to be able to come back to it, and updates ui.

| before | after |
| -------|------|
| image | image |
| missing top border + wrong alignment of the label | border added and alignment straight | 

**Checklist**

### DIFF
--- a/.changeset/odd-parrots-lick.md
+++ b/.changeset/odd-parrots-lick.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
+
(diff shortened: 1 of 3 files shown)


amritk (approved):
Code looks good, did not test

hanspagel (approved):
better

PR #4495, opened 2025-01-17
fix(api-client): path params

antlio:
**Problem**
path param addition in the path are not enabling the path params table visibility. 

**Solution**
this pr:
1. adds path variable table on variable presence in request path change
2. proposes to add back the checkbox while keeping it non updatable for consistency purposes
3. increase color of icon used in enabled checkbox to increase state information perception

| before | after |
| -------|------|
| image | image |
| variable table doesn't display checkbox - checked icon dimmed down | displays checkbox for variable + increase checked icon color | 

**Checklist**

### DIFF
--- a/.changeset/hip-rockets-sniff.md
+++ b/.changeset/hip-rockets-sniff.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
+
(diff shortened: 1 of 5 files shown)


hanspagel (approved):
Cool

PR #4501, opened 2025-01-18
fix(api-reference): complex security requirements

### LINKED ISSUE #3359: Missing combined security schema UI
### What happens?

I have a combined security schema defined as follows:
```
security:
 - bearerAuth: [ ]
   apiKeyHeader: [ ]
```
I am using @scalar/api-reference 1.25.25 package
when I try "Test Request" on one of the endpoints with authentication required there is no option for selecting combined Bearer & Subscription Key authentication type.

image

It used to look like this:
image

Is there a way now to select combined schema as the preferred security schema? 
Examples in the api-reference are now misleading because they show only one schema used (here it is --header 'X-Api-Key: YOUR_TOKEN', but I need both api key and bearer token for authentication)
image

Having option to select combined security schema which will be reflected in both api-client and api-reference examples

### OpenAPI Document

[Sandbox](https://sandbox.scalar.com/e/LSMHy)

amritk:
Complex security, this was done a few times but kept breaking whenever we refactored or changed something. Have now added tests to ensure this does not happen again!

fixes #3359

currently needs
- [x] integration tests
- [x] switch to array instead of comma

Originally I did it the easy way with much less changes, but it was not quite correct. So I switched from comma separated to using a proper array! Also switched to prop drilling so we can add integration tests :). Now the integration tests are still a little basic as we have a lot of reliance on `useActiveEntities` but I will keep drilling 

**Problem**
Complex auth was broken

**Solution**
We add back in complex auth

**To test**
load up the client and import this spec to test
`https://gist.githubusercontent.com/amritk/f64887015e998bf81ff349d3ace64cd5/raw/4499a362e1f274ef7f2f61e7f572c34dfa67e4db/test.yml`
the key part is
```json
  "security": [
    {
      "bearerAuth": [],
      "apiKeyQuery": []
    },
  ],
```
which has two security schemes in a single requirement

image
@cameronrohani I just stacked em like this, feel like it makes sense?

image

**Checklist**

### DIFF
--- a/packages/api-client/src/views/Request/RequestSection/RequestAuth/RequestAuth.vue
+++ b/packages/api-client/src/views/Request/RequestSection/RequestAuth/RequestAuth.vue
@@ -1,14 +1,14 @@
 <script setup lang="ts">
 import ViewLayoutCollapse from '@/components/ViewLayout/ViewLayoutCollapse.vue'
 import { useLayout } from '@/hooks/useLayout'
-import { useWorkspace } from '@/store'
-import { useActiveEntities } from '@/store/active-entities'
+import { useWorkspace } from '@/store/store'
+import type { SecuritySchemeOption } from '@/views/Request/consts'
 import {
-  ADD_AUTH_OPTIONS,
-  type SecuritySchemeGroup,
-  type SecuritySchemeOption,
-} from '@/views/Request/consts'
-import { displaySchemeFormatter } from '@/views/Request/libs'
+  formatComplexScheme,
+  formatScheme,
+  getSchemeOptions,
+  getSecurityRequirements,
+} from '@/views/Request/libs'
 import {
   type Icon,
   ScalarButton,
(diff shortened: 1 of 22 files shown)


antlio:
great work! i'm wondering if we could save some duplicated infos and space by playing with the label:

| before | after |
| -------|------|
| image | image |
| current PR state | proposal to use label to differentiate auth |

amritk:
> great work! i'm wondering if we could save some duplicated infos and space by playing with the label:
> 
> before	after
> image	image
> current PR state	proposal to use label to differentiate auth

yea that works, but we can fast follow, lets get this one in first!

packages/api-client/src/views/Request/RequestSection/RequestAuth/RequestAuthTab.vue
```diff
@@ -165,9 +179,12 @@ const updateScheme = <U extends string, P extends Path<SecurityScheme>>(
         :key="key">
         <OAuth2
           v-if="activeFlow === key || (ind === 0 && !activeFlow)"
+          :collection="collection"
           :flow="flow!"
           :layout="layout"
-          :scheme="scheme" />
+          :scheme="scheme"
```
amritk, line 185: Prop drillin

antlio: 🚨

packages/api-client/src/views/Request/RequestSection/RequestAuth/RequestAuth.vue
```diff
@@ -86,57 +91,53 @@ const authIndicator = computed(() => {
...
-
+const editSelectedSchemeUids = (uids: SelectedSecuritySchemeUids) => {
   // Set as selected on the collection for the modal
-  if (layout === 'modal' || layout === 'reference') {
-    collectionMutators.edit(
-      activeCollection.value.uid,
-      'selectedSecuritySchemeUids',
-      uids,
-    )
+  if (hookLayout === 'modal' || propLayout === 'reference') {
```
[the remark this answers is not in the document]
antlio: yes! i should give a deeper thought about how to make this easier to handle

marclave (approved):
unreal work @amritk 🫵🐐

PR #4502, opened 2025-01-18
feat(components): host storybook on netlify

hwkr:
Opened a new PR thinking that would fix stuff. It did not but I figured out what the problem was and now it's working!

### DIFF
--- a/packages/components/.gitignore
+++ b/packages/components/.gitignore
@@ -30,3 +30,6 @@ coverage
 
 test-results/
 playwright-report/
+
+# Storybook
+storybook-static/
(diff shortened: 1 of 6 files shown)


marclave (approved):

packages/components/.gitignore
```diff
@@ -30,3 +30,6 @@ coverage
 
 test-results/
 playwright-report/
+
+# Storybook
+storybook-static/
```
antlio, line 35: CI seems to be expecting this to deploy

hwkr: How do you mean?

antlio: this PR and main is failing

hwkr: It's looked like the main build was failing yesterday due to and install issue, I'll have a look!

PR #4505, opened 2025-01-18
fix(api-client): change OAuth State parameter to a length of 8

petercinibulk:
Make state variable a length of 8 to support OAuth providers (Kinde) who need more entropy in the state parameter

**Problem**
Currently, I cannot use Scalar to authenticate with Kinde due to the state parameter not being at least 8 characters

**Solution**
With this PR I am changing the length of the state parameter from 5 to 8.

**Checklist**

### DIFF
--- a/packages/api-client/src/views/Request/libs/oauth2.test.ts
+++ b/packages/api-client/src/views/Request/libs/oauth2.test.ts
@@ -41,7 +41,7 @@ const windowTarget = 'openAuth2Window'
 const windowFeatures = 'left=100,top=100,width=800,height=600'
 
 /** This state corresponds to Math.random() === 0.123456 */
-const state = '82mvz'
+const state = '4fzyo82m'
 const randomVal = 0.123456
 
 describe('oauth2', () => {
--- a/packages/api-client/src/views/Request/libs/oauth2.ts
+++ b/packages/api-client/src/views/Request/libs/oauth2.ts
@@ -75,7 +75,7 @@ export const authorizeOauth2 = async (
 
     // OAuth2 flows with a login popup
     else {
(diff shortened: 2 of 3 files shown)


marclave (approved):
awesome work @petercinibulk ! let's add a changeset (patch) and get this in? :)

petercinibulk:
@marclave Added the changeset. Thanks!

marclave:
@hanspagel @amritk any idea why these tests are failing?

amritk:
Just re ran it. Was cancelled

packages/api-client/src/views/Request/libs/oauth2.test.ts
```diff
@@ -41,7 +41,7 @@ const windowTarget = 'openAuth2Window'
 const windowFeatures = 'left=100,top=100,width=800,height=600'
 
 /** This state corresponds to Math.random() === 0.123456 */
-const state = '82mvz'
+const state = '4fzyo82m'
```
antlio, line 44: good catch 👍

packages/api-client/src/views/Request/libs/oauth2.ts
```diff
@@ -75,7 +75,7 @@ export const authorizeOauth2 = async (
 
     // OAuth2 flows with a login popup
     else {
-      const state = (Math.random() + 1).toString(36).substring(7)
+      const state = (Math.random() + 1).toString(36).substring(2, 10)
```
hanspagel, line 78:
```suggestion
      // Generate a random state string with the length of 8 characters
      const state = (Math.random() + 1).toString(36).substring(2, 10)
```

hanspagel (approved):
Thanks for taking the time to fix this, great PR! 🙏

PR #4509, opened 2025-01-19
fix(code-highlight): removes regex for highlight credentials

### LINKED ISSUE #4498: Using certain characters in an API key gives an Invalid regular expression exception
### What happens?

Using the character '(' or ')' as an API key results in an exception:

Image

Image

This is the expected result:

Image

### OpenAPI Document

_No response_

antlio:
**Problem**
currently when using `token(` in credentials value, the code example is not rendered.

**Solution**
this pr fixes #4498 by removing regex instance to highlight credentials.

| before | after |
| -------|------|
| image | image |
| code example is not rendered | code example get rendered |  

**Checklist**

### DIFF
--- a/.changeset/itchy-tigers-float.md
+++ b/.changeset/itchy-tigers-float.md
@@ -1,0 +1,5 @@
+---
+'@scalar/code-highlight': patch
+---
+
+fix: removes regex for highlight credentials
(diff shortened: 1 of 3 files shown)


hanspagel (changes requested):
That was quick!

Maybe we just don’t use Regex here but replace the literal string? That would make the code easier I guess. Wdyt?

hanspagel (approved):
Yesss! Simpler & better.

PR #4515, opened 2025-01-20
refactor(api-reference): use the store in `TestRequestButton`

hanspagel:
Currently, we pass the (legacy) `TransformedOperation` to the `TestRequestButton`, pass the method and the path to the API client, and then resolve the request entity from the store.

With this PR, we’re just passing the RequestEntity to the button, and then only pass the request UID to the client.

Way easier, and also another component where we can get rid of the (legacy) `TransformedOperation`.

### DIFF
--- a/packages/api-client/src/libs/create-client.ts
+++ b/packages/api-client/src/libs/create-client.ts
@@ -9,10 +9,7 @@ import {
   type WorkspaceStore,
   createWorkspaceStore,
 } from '@/store/store'
-import type {
-  RequestMethod,
-  SecurityScheme,
-} from '@scalar/oas-utils/entities/spec'
+import type { SecurityScheme } from '@scalar/oas-utils/entities/spec'
 import { workspaceSchema } from '@scalar/oas-utils/entities/workspace'
 import {
   LS_KEYS,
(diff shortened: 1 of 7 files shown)


packages/play-button/src/index.ts
```diff
@@ -111,15 +111,21 @@ if (!specUrlElement && !specElement && !specScriptTag) {
         if (specifiedOperation) {
           open({
             path: specifiedOperation.path,
-            method: specifiedOperation.httpVerb,
+            method: specifiedOperation.httpVerb.toLowerCase() as Exclude<
```
amritk, line 114:
```suggestion
            method: specifiedOperation.httpVerb.toLowerCase() as Exclude<
```
instead of as we like to use type guards now for safety. We have a few around like
https://github.com/scalar/scalar/blob/main/packages/oas-utils/src/helpers/httpMethods.ts#L85

packages/api-client/src/libs/create-client.ts
```diff
@@ -45,9 +42,18 @@ export type ClientConfiguration = {
   | '_integration'
 >
 
-export type OpenClientPayload = {
-  path: string
-  method: LiteralUnion<RequestMethod | Lowercase<RequestMethod>, string>
+export type OpenClientPayload = (
```
amritk, line 45: nice

amritk (approved):
Code looks good for the most part, one comment as we are moving to type guards but not blocking. Did not test

Issue #4523, opened 2025-01-20
fastify-api-reference does not work with TS moduleResolution mode "NodeNext"


chrisbenincasa:
### What happens?

During compilation:
```
@tunarr/server:build: ../node_modules/.pnpm/@scalar+fastify-api-reference@1.25.106/node_modules/@scalar/fastify-api-reference/dist/index.d.ts:1:25 - error TS2834: Relative import paths need explicit file extensions in ECMAScript imports when '--moduleResolution' is 'node16' or 'nodenext'. Consider adding an extension to the import path.
@tunarr/server:build: 
@tunarr/server:build: 1 export { default } from './fastifyApiReference';
@tunarr/server:build:                           ~~~~~~~~~~~~~~~~~~~~~~~
@tunarr/server:build: 
@tunarr/server:build: ../node_modules/.pnpm/@scalar+fastify-api-reference@1.25.106/node_modules/@scalar/fastify-api-reference/dist/index.d.ts:2:15 - error TS2834: Relative import paths need explicit file extensions in ECMAScript imports when '--moduleResolution' is 'node16' or 'nodenext'. Consider adding an extension to the import path.
@tunarr/server:build: 
@tunarr/server:build: 2 export * from './types';
@tunarr/server:build:                 ~~~~~~~~~
@tunarr/server:build: 
@tunarr/server:build: 
@tunarr/server:build: Found 2 errors in the same file, starting at: ../node_modules/.pnpm/@scalar+fastify-api-reference@1.25.106/node_modules/@scalar/fastify-api-reference/dist/index.d.ts:1
```

### What did you expect to happen?

Build works as expected. Guessing this is related to https://github.com/scalar/scalar/issues/4009. Seems like it has to do with the packages using moduleResolution bundler?

chrisbenincasa:
Further problems when you follow the guidance and change the moduleResolution mode (not that I really wanna do this in my project just to make this work...)

```
@tunarr/server:build: ../node_modules/.pnpm/@scalar+types@0.0.27/node_modules/@scalar/types/dist/external/snippetz.d.ts:1:53 -
 error TS2307: Cannot find module '@scalar/snippetz/types' or its corresponding type declarations.
@tunarr/server:build:                                          
@tunarr/server:build: 1 export type { HarRequest, TargetId, ClientId } from '@scalar/snippetz/types';
@tunarr/server:build:                                                       ~~~~~~~~~~~~~~~~~~~~~~~~
@tunarr/server:build:    
```

Issue #4009, opened 2024-11-25
Generated TypeScript types does not work on ESM


joshuaavalon:
### What happens?

```ts
// Module '"@scalar/fastify-api-reference"' has no exported member 'FastifyApiReferenceOptions'. Did you mean to use 'import FastifyApiReferenceOptions from "@scalar/fastify-api-reference"' instead?
import type { FastifyApiReferenceOptions } from "@scalar/fastify-api-reference";
```

### What did you expect to happen?

Types should work.

Because scalar tsconfig.json is using `"moduleResolution": "Bundler"` which allows code like
```ts
export * from "./types";
```

However, file extensions are mandatory in ESM. It should be
```ts
export * from "./types.js";
```

The *.d.ts generated by vite does not have file extensions.

PR #4528, opened 2025-01-21
fix(fastify-api-reference): can’t use with the (recommended) TS `moduleResolution: NodeNext` setting

### LINKED ISSUE #3134: node16/nodenext support for fastify plugin
### What happens?

When I import `fastifyApiReference`, the type is `any` :
```
Unsafe argument of type any assigned to a parameter of type FastifyPluginCallback<{ routePrefix: string; }, RawServerDefault, FastifyTypeProvider, FastifyBaseLogger>
```

`tsconfig.json` with `node16` :
```json
{
  "compilerOptions": {
    "outDir": "dist",
    "declaration": false

    "module": "node16",
    "moduleResolution": "node16",
    "moduleDetection": "force",
    "target": "es2022",

    "resolveJsonModule": false,
    "declaration": true,
    "sourceMap": true,
    "stripInternal": true,
    "strict": true,
    "noImplicitReturns": true,
    "noImplicitOverride": true,
    "noUnusedLocals": true,
    "noUnusedParameters": true,
    "noFallthroughCasesInSwitch": true,
    "noUncheckedIndexedAccess": true,
    "noPropertyAccessFromIndexSignature": true,
    "noEmitOnError": true,
    "forceConsistentCasingInFileNames": true,
    "skipLibCheck": true
  },
  "include": [
    "src"
  ]
}
```

I came across this issue on the fastify repo https://github.com/fastify/fastify/issues/4349, with an explanation on how to fix it in javascript plugins, but I do not know how to fix it here in this typescript plugin 🤔 
I would need some help 🙏 

I was expecting the type of `fastifyApiReference` to be defined

### OpenAPI Document

_No response_

hanspagel:
**Problem**
Currently, we use TS `moduleResolution: Bundler` which is cool, but allows vague imports, like:

```ts
import { Foobar } from './types`
// resolved to `./types/index.js` under the hood
```

Personally, I love this, BUT, this breaks in user land projects that use the more strict `moduleResolution: 'NodeNext`.

**Solution**

This PR adds a new `tsconfig.node.json` in the root, that extends the original one, but has `module: 'NodeNext'` and `moduleResolution: 'NodeNext'`, which is recommended by the [TypeScript Cheat Sheet](https://www.totaltypescript.com/tsconfig-cheat-sheet#transpiling-with-typescript) that we all love.

This TSConfig is applied to `@scalar/fastify-api-reference` and the packages that it depends on:

* `@scalar/openapi-parser`
* `@scalar/types`

This forces us to be explicit with the import (which I don't like, but that’s just a matter of taste and what I’m used to):

```ts
import { Foobar } from './types/index.ts`
```

~~Why the `.js` and not `.ts`? Because TypeScript!~~ We use `"allowImportingTsExtensions": true`.

With this change it should be possible to actually use those three packages in setups with the (recommended) `NodeNext` setting.

Fixes #3134, #4009, #4523.

**Alternatives**

* I did hours of research whether we could just slap something into the TSConfig, Rollup config or Vite config to make the imports explicit. The answer is: All solutions felt like a hack, none of them were really TypeScript-native, Vite-native or Rollup-native. It was all just different ways of a string replace script.
* I was tempted to also switch to the `@/*` alias, but that would have made the changeset harder to review.

**Checklist**


packages/fastify-api-reference/vite.config.ts
```diff
@@ -2,7 +2,7 @@ import { viteStaticCopy } from 'vite-plugin-static-copy'
 import { defineConfig } from 'vitest/config'
 
 import pkg from './package.json'
-import { nodeExternals } from './vite-plugins'
+import { nodeExternals } from './vite-plugins/nodeExternals.js'
```
Greenheart, line 5: Regarding the `.js` extensions, you can set `"allowImportingTsExtensions": true` in your `tsconfig.node.json` to allow using explicit `.ts` extensions in the code. You might prefer that, based on what you wrote in the PR description :)

chrisbenincasa: I believe allowImportingTsExtensions requires moduleResolution=bundler and I think this PR changes the resolution to NodeNext. However, in TS 5.7 `rewriteRelativeImportExtensions` was introduced for this reason (inability to use .ts extensions for code that actually produces JS output). Assuming it might be out of scope to bump the TS version in this change, though

Greenheart: Ah, that's good to know!

amritk: @chrisbenincasa we can definitely put up another PR to bump the TS version!

chrisbenincasa: @Greenheart something else must be at work there - it doesn't seem to be invalid, but vscode warns against this explicitly: 

image

but interestingly the docs only mention the emit parts: https://www.typescriptlang.org/tsconfig/#allowImportingTsExtensions

So perhaps it won't cause outright failure.

hanspagel: Yeah, it’s all not great. I think I’ll just see if I can get used to it.

amritk: there's gotta be another way! Can't we just do it in the build, so it builds out a NodeNext version?

packages/openapi-parser/src/utils/betterAjvErrors/helpers.ts
```diff
@@ -8,15 +8,15 @@ import {
...
+} from './utils.ts'
 import {
   AdditionalPropValidationError,
   DefaultValidationError,
   EnumValidationError,
   PatternValidationError,
   RequiredValidationError,
   UnevaluatedPropValidationError,
-} from './validation-errors/index'
+} from './validation-errors/index.js'
```
amritk, line 19: missed a js?

amritk:
```suggestion
} from './validation-errors/index.ts'
```

packages/openapi-parser/src/utils/index.ts
```diff
@@ -1,26 +1,26 @@
...
-export * from './toYaml'
-export * from './transformErrors'
-export * from './traverse'
-export * from './unescapeJsonPointer'
-export * from './upgrade'
-export * from './upgradeFromThreeToThreeOne'
-export * from './upgradeFromTwoToThree'
-export * from './validate'
-export * from './transform'
+export * from './dereference.ts'
```
amritk, line 1: maybe out of scope for this PR, but we can get rid of this file too. Also the export *

amritk (approved):
Awesome work! just one missed JS but will approve. As a continuation we should get rid of any barrel files as well and just import directly from the file. The only barrel files we should have are the ones exporting from the package, and even those will have explicit exports instead of `export * from ...` 😄

packages/openapi-parser/src/utils/betterAjvErrors/helpers.ts
hanspagel: Nah, it’s actually js, because this is just code I copied into the repository and never really migrated.

PR #4530, opened 2025-01-21
feat(openapi-parser): don’t load URLs/files passed as OpenAPI document

hanspagel:
WIP

Needs more work than I thought initially.

**Problem**
Currently, the parser will try to resolves URLs and files passed directly (instead of a YAML/JSON OpenAPI document). This is confusing to some users.

**Solution**
With this PR we remove this feature.

**Checklist**


hanspagel:
We need an empty PR queue to apply the new formatting. But I’ll pick this up again soon-ish.

PR #4531, opened 2025-01-21
refactor(api-reference): use server state from the client

hanspagel:
**Problem**
currently reference uses its own server store which create a lot of duplicated logic with the client.

**Solution**
Continuing @amritk’s effort: https://github.com/scalar/scalar/compare/main...amrit/sync-server-ref-client in order to use the server state in both client and reference and refactor.

**Tasks**
- [x] add a dropdown for the servers again
- [x] check whether `config.servers` works
- [x] check whether `config.baseServerURL` works
- [x] add the UI
- [ ] write tests
- [x] check what happens if no server is provided
- [x] check what happens to relative server URLs
- [x] check that variables in URLs work

**Preview**

image

**Checklist**


amritk (approved):
LGTM!

PR #4532, opened 2025-01-21
feat(api-client): address bar server variables

antlio:
**Problem**
currently users cannot set server variables from the address bar in the api client

**Solution**
this pr updates the api client address bar and welcome server variables there as seen below:

| before | after |
| -------|------|
| image | image |
| menu is not displaying server variables | menu displays server variables | 

it also removes the server variables being displayed in the variables table of the request page, which remains to be used only for path params.

**Checklist**


marclave (approved):
AWESOME work @antlio, code looks good + i tested.

what would be nice is in the dropdown of the servers in the addressbar, being able to go to that server in the servers page. so some form of "edit" which links to the server in the servers page :) but that can be fast followed

hanspagel (approved):
sick work

packages/api-client/src/components/AddressBar/AddressBar.vue
```diff
@@ -126,79 +124,72 @@ function updateRequestPath(url: string) {
...
+            @curl="$emit('importCurl', $event)"
+            @submit="handleExecuteRequest"
+            @update:modelValue="updateRequestPath" />
+          <div class="fade-right"></div>
         </div>
-      </Listbox>
+
+        <AddressBarHistory />
+        <ScalarButton
+          class="relative h-auto shrink-0 overflow-hidden !pl-2 !pr-2.5 !py-1 font-bold"
```
amritk, line 178:
```suggestion
          class="relative h-auto shrink-0 overflow-hidden pl-2 pr-2.5 py-1 font-bold"
```
we shouldn't need the importants, I believe scalarButton should be using tw-merge/cx already, right @hwkr

amritk: I just realized this isn't part of your PR

packages/api-client/src/components/AddressBar/AddressBarServerItem.vue
```diff
@@ -38,28 +38,91 @@ const updateSelectedServer = (serverUid: string) => {
...
+}
+
+const updateServerVariable = (key: string, value: string) => {
+  if (!activeServer.value) return
+
+  const variables = activeServer.value.variables || {}
+  variables[key] = { ...variables[key], default: value }
+
+  serverMutators.edit(activeServer.value.uid, 'variables', variables)
+}
```
amritk, line 71: We are setting the value as default here, would it make more sense to add a value property? Because now we can never go back to the default. Or is that ok?

packages/api-client/src/components/Server/types.ts
```diff
@@ -0,0 +1,13 @@
...
+export type ServerVariables = {
+  [variable: string]: ServerVariable
+}
+
+export type ServerVariableValues = {
+  [variable: string]: string
+}
+
+export type ServerVariable =
+  | OpenAPIV3.ServerVariableObject
```
amritk, line 12:
```suggestion
export type ServerVariable =
```
due to upgrade

amritk (approved):
Works, nice work! One question about how we want to store the value of the variable

packages/api-client/src/components/AddressBar/AddressBar.vue
hwkr: Yep! This is probably from before I made useBindCx

packages/api-client/src/components/Server/ServerVariablesForm.vue
```diff
@@ -0,0 +1,64 @@
...
+</script>
+<template>
+  <template v-if="variables && Object.keys(variables ?? {}).length">
+    <template
+      v-for="name in Object.keys(variables)"
+      :key="name">
+      <div class="group flex w-full">
+        <label
+          class="flex items-center pl-3 py-1.5 after:content-[':'] mr-1.5 group-has-[input]:mr-0"
+          :for="`${baseId}-variable-${name}`">
```
hwkr, line 43: It looks like if the text box doesn't exist this `for` is broken? 

Rather than use an id here you can just make the label element the wrapper element (e.g. `<div class="group flex w-full">` → `<label class="group flex w-full">` and turn the label into a div or span) and it should automatically associate for you

packages/api-client/src/components/Server/ServerVariablesSelect.vue
```diff
@@ -0,0 +1,52 @@
...
+})
+</script>
+<template>
+  <ScalarListbox
+    v-model="selected"
+    :options="options">
+    <ScalarButton
+      :aria-controls="controls"
+      class="h-8 p-0 py-1.5"
+      fullWidth
```
hwkr, line 36: You can just put it in a flexbox to achieve the same thing (or add `w-full` to the classes)

hwkr (approved):
A few small suggestions! Not blocking though, I know a bunch of this is pulled from the history listbox (which is absolutely positioned) but I'd be happy to pair if you want to migrate them to use `ScalarFloating`! (Should be more robust and remove a bunch of CSS)

cameronrohani (changes requested):
awesome work, you killed it!

I'm seeing that request history background is somehow altered now in dark mode in the client preview:

image

let's just make sure this isn't happening and then it'll be good to go from my end

hwkr (changes requested):
Ooops I reviewed it then did a bit more testing 😅 

Found a couple issues when I added specs with both many servers and many variables in those servers.

## Issues

- [ ] The popover doesn't scroll when you the contents are too big, instead they get squished Google Chrome-2025-01-21-21-05-04@2x
- [ ] You can't tab between inputs in the form (probably because it's a listbox)[video](https://github.com/user-attachments/assets/1de10217-57fa-43df-908e-9de5f51fee30)


## Nice to have / Non blocking issues

- [ ] The text edit box doesn't take up the full width of form (so clicking the negative space doesn't select the text box)
- [ ] When you resize the window the content is able to flow off the window (I'm guessing this is what happens to the history popover as well but it would be nice to fix it) - Nice to have / Not blocking Google Chrome-2025-01-21-21-10-14@2x

I would recommend switching this to use `ScalarPopover` since it will handle the focus trapping for you and can also help handle the scroll behaviour. Happy to pair if you want :)

Test files:[test-files.zip](https://github.com/user-attachments/files/18500192/test-files.zip) (GitHub won't let me attach YAML?!?)

packages/api-client/src/components/Server/ServerVariablesForm.vue
antlio: absolutely right will do! just copy pasted from reference so far

packages/api-client/src/components/AddressBar/AddressBar.vue
antlio: good catch can do

packages/api-client/src/components/AddressBar/AddressBarServerItem.vue
antlio: i removed this, it was initially to test out in my first iteration variables presence.

antlio:
tysm for all the feedbacks. i updated according to it + added a better visual accessibility for light/dark:

| lightmode | darkmode |
| -------|------|
| image | image |

as we should avoid deeper contrast on focus item in general for dark mode.

antlio:
> one last thing, not blocking but let's get it fixed
> 
> the input size I'm seeing isn't spanning to 100% width

i felt it was helping focusing the sight closer to the value, but can do!

packages/components/src/components/ScalarFloating/ScalarFloating.vue
```diff
@@ -77,7 +77,7 @@ const { floatingStyles, middlewareData } = useFloating(targetRef, floatingRef, {
   placement: computed(() => props.placement),
   whileElementsMounted: autoUpdate,
   middleware: computed(() => [
-    offset(5),
+    offset(props.position ?? 5),
```
hwkr, line 80: We can't actually make this a prop because it's not reactive IIRC

packages/components/src/components/ScalarFloating/ScalarFloating.vue
```diff
@@ -107,6 +107,7 @@ const { floatingStyles, middlewareData } = useFloating(targetRef, floatingRef, {
     <div
       ref="floatingRef"
       class="relative z-overlay"
+      :class="{ 'w-full': fullWidth }"
```
hwkr, line 110: I think the prop you're looking for is `resize`

hwkr (changes requested):
Just going to request changes while I do some components tweaks.

packages/api-client/src/components/AddressBar/AddressBarServer.vue
```diff
@@ -65,45 +68,47 @@ const serverUrlWithoutTrailingSlash = computed(() => {
...
-        :key="serverOption.id"
-        :serverOption="serverOption"
-        type="collection" />
-      <!-- Add Server -->
-      <template v-if="!isReadOnly">
-        <ScalarDropdownDivider />
-        <ScalarDropdownItem>
+    </ScalarButton>
+    <template #popover="{ close }">
+      <div class="menu custom-scroll flex flex-col gap-1 max-h-[300px]">
```
hwkr, line 83: `max-h-inherit` will allow you to fill but not overflow the screen height.

packages/components/src/components/ScalarPopover/ScalarPopover.vue
```diff
@@ -23,6 +23,7 @@ const { cx } = useBindCx()
       <template #floating="{ width, height }">
         <PopoverPanel
           v-slot="{ close }"
+          :middleware="middleware"
```
hwkr, line 26: I don't think this will do anything here since the middleware is handled by the floating element 🤔

packages/api-client/src/components/AddressBar/AddressBarServer.vue
```diff
@@ -65,45 +68,47 @@ const serverUrlWithoutTrailingSlash = computed(() => {
...
-    class="w-max"
-    teleport>
-    <button
-      class="font-code lg:text-sm text-xs whitespace-nowrap border border-b-3 border-solid rounded px-1.5 py-0.5 text-c-2 -outline-offset-1"
-      type="button">
+  <ScalarPopover
+    class="dark:bg-b-1 bg-b-2 border border-t-0 -top-1.25 min-w-full rounded-b-lg relative w-full"
+    placement="top-start"
+    resize
+    target="address-bar">
```
hwkr, line 75: This probably needs a uid 😢

hwkr (approved):
That was supposed to be an approval but GitHub is borked

antlio:
great work @hwkr - fixing the backdrop on darkmode and we should be good to go!

cameronrohani (approved):
amazing work!! ya'll killed it!!

PR #4533, opened 2025-01-21
feat(api-client): better server handling

hanspagel:
WIP

Need to fix tests

**Problem**

The server logic in the reference covered some edge cases that we didn't cover in the client.

**Solution**

This PR:
* moves the server logic to a separate function
* add tests
* covers some more edge cases
* adds baseServerURL to the documentation
* improves the Zod schema for the servers
  * closer to the specification
  * removes `'default'` as the default value for variables
  * if `enum` is defined: enforces `default` is a possible value
  * if a single server is invalid, it's omitted, but the others servers are still there
  * the data is validated against the schema first *and then* modified, this makes it more robust
  * updated the comments

Screenshot 2025-01-21 at 17 02 16

**Checklist**

### DIFF
--- a/packages/oas-utils/src/transforms/import-spec.ts
+++ b/packages/oas-utils/src/transforms/import-spec.ts
@@ -19,6 +19,7 @@ import {
   type SecuritySchemePayload,
   securitySchemeSchema,
 } from '@/entities/spec/security'
+import { concatenateUrlAndPath, isDefined } from '@/helpers'
 import { isHttpMethod } from '@/helpers/httpMethods'
 import { schemaModel } from '@/helpers/schema-model'
 import { keysOf } from '@scalar/object-utils/arrays'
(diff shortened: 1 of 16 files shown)


packages/oas-utils/src/transforms/import-spec.ts
```diff
@@ -478,3 +455,57 @@ export async function importSpecToWorkspace(
...
+): Server[] {
+  if (!servers || !Array.isArray(servers)) return []
+
+  return servers
+    .map((server): Server | undefined => {
+      try {
+        // Validate the server against the schema
+        const parsedSchema = serverSchema.parse(server)
+
+        console.log()
```
amritk, line 474:
```suggestion
```

amritk (approved):
Has some test and some type issues to resolve but, love the code! We should continue pulling logic out and testing it!!! One question about the OpenAPIV3_1 type is it already de-referenced? Because we have de-referenced already but the type thinks there could be a reference still, or am I mistaken.

hanspagel:
Those types are usually for the original document with unresolved references, but servers and variables don’t allow references, so there’s no difference. Not super happy with that solution, I don't know. Just wanted to get rid of another random custom type.

PR #4535, opened 2025-01-21
fix(Scalar.AspNetCore): allow anonymous for assets endpoint

### LINKED ISSUE #4529: Upgrade to `Scalar.AspetCore` version `2.0.0` breaks ui assets
### What happens?

After upgrading to `Scalar.AspetCore` version `2.0.0` the assets `/scalar/scalar.aspnetcore.js` and `/scalar/scalar.js` can't be loaded. The endpoints now return a `401 Unauthorized`.

```cs
    {
        options.WithOAuth2Authentication(oauth =>
        {
            oauth.ClientId = "foo";
            oauth.Scopes = ["bar"];
        });
    })
    // This is important because the default authentication policy of our app requires authentication
    .AllowAnonymous();
```

I've verified that this issue still persists in `2.0.1` `2.0.2`

The UI can be loaded.

Verified on version `1.2.76`

### OpenAPI Document

_No response_

xC0dex:
Closes #4529 

**Problem**
The recently introduced static assets endpoints may require authentication when a fallback authentication is configured. This results in `401 Unauthorized` responses when the browser attempts to load static assets.

**Solution**
I applied the `AllowAnonymous()` method to the static assets endpoints to ensure they bypass authentication. Additionally, I updated the test to cover this scenario.

**Checklist**

- [ ] ~I’ve updated the documentation.~ 👀

### DIFF
--- a/.changeset/strong-shirts-juggle.md
+++ b/.changeset/strong-shirts-juggle.md
@@ -1,0 +1,5 @@
+---
+'@scalar/aspnetcore': patch
+---
+
+fix: anonymous assets endpoint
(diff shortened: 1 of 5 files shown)


hanspagel (approved):
Just insane how fast you come up with those fixes

PR #4537, opened 2025-01-21
chore: update typescript

amritk:
~Bumped ts version, should be relatively safe if it builds.~
famous last words, upgrade ts led to vue-tsc. We should test this a bit heavier

Also nuxt broken right now, we can let it marinate for a bit
https://github.com/nuxt/module-builder/issues/141

Also removed concurrently as I don't think we use it


hanspagel:
Closing for now. Too many conflicts, still too much work left, not urgent enough. Removed conurrently in a separate PR #4695.

amritk:
yea @hanspagel we can try again in 5.8

PR #4543, opened 2025-01-22
fix(Scalar.AspNetCore): add missing parameter to tests

hanspagel:
**Problem**
Currently, the Scalar.AspNetCore tests fail in main. I forgot to rebase before merging the recent PR.

**Solution**
With this PR it should pass again, because we added the recently introduced parameter to the also recently introduced tests.

**Checklist**


xC0dex (approved):
Thanks for the fix 🚀

PR #4545, opened 2025-01-22
fix(api-client): broken test

hanspagel:
Omg, I broke a test with a quick refactor before merging and didn’t notice.

This should fix it.


antlio (approved):
back

antlio:
should we # reference where it broke?

PR #4551, opened 2025-01-22
feat(api-client): favors layout hook over isReadOnly

antlio:
**Problem**
we currently have two way to distinguish reference and client usages.

**Solution**
this pr favors the usage of the layout hook over the previously used isReadOnly. To be thoroughly tested to make sure no previously set behavior are now possible in client/reference and vice versa.

**Checklist**

### DIFF
--- a/.changeset/chatty-garlics-deny.md
+++ b/.changeset/chatty-garlics-deny.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
+
+feat: favors layout hook over isReadOnly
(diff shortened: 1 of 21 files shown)


amritk (changes requested):
thanks for doing this! much needed

amritk (approved):
you may need to mock useLayout to fix some tests

marclave (approved):
😎

hanspagel (approved):
FINALLY

amritk:
lets merge?

PR #4552, opened 2025-01-22
fix(api-reference):  add `.introduction-section` class to Introduction

hwkr:
Brings back the `.introduction-section` class which is used in a bunch of themes. Also adds a test so we (hopefully) don't accidentally delete the class again.


cameronrohani (approved):

cameronrohani:
in, let's get this @marclave 

image

marclave (approved):

PR #4557, opened 2025-01-23
fix(openapi-types): openapi types

amritk:
HMMM MAYBE NOT?

**Problem**
Quite a few parameters are typed as any

**Solution**
We type the missing parameters by fixing the modify type helper

**Checklist**

### DIFF
--- a/packages/oas-utils/src/entities/spec/requests.ts
+++ b/packages/oas-utils/src/entities/spec/requests.ts
@@ -102,7 +102,7 @@ export const oasRequestSchema = z.object({
   /** Hide operations */
   'x-internal': z.boolean().optional(),
   'x-scalar-ignore': z.boolean().optional(),
-}) satisfies ZodSchema<OpenAPIV3_1.OperationObject>
+})
 
 /**
  * Extended properties added to the spec definition for client usage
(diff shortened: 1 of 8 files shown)


packages/oas-utils/src/entities/spec/requests.ts
```diff
@@ -102,7 +102,7 @@ export const oasRequestSchema = z.object({
   /** Hide operations */
   'x-internal': z.boolean().optional(),
   'x-scalar-ignore': z.boolean().optional(),
-}) satisfies ZodSchema<OpenAPIV3_1.OperationObject>
```
amritk, line 105: this incorrectly tests the input instead of the inferred value, moved the check to a test

amritk:
I don't even know what the current error means

hanspagel:
Maybe a circular ref or something 🥴

amritk:
> Maybe a circular ref or something 🥴

@hanspagel 5.8 has some good stuff coming, I will try again then!

hanspagel:
We need an empty PR queue, so let’s close this and maybe pick this up later. @amritk are discussing an alternative approach already, let’s see what we end up with.

PR #4561, opened 2025-01-23
fix(components): listbox margin

antlio:
**Problem**
current margin usage in listbox is causing gap inconsistency on single list item.

**Solution**
this pr removes margin and uses gap instead.

| before | after |
| -------|------|
| image | image |
| higher top and down spaces | consistent surrounding space | 

**Checklist**

### DIFF
--- a/.changeset/metal-phones-burn.md
+++ b/.changeset/metal-phones-burn.md
@@ -1,0 +1,5 @@
+---
+'@scalar/components': patch
+---
+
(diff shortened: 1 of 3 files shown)


hwkr (approved):
This is fine with me but originally it had the extra space on purpose I believe, cc @cameronrohani

marclave (approved):

cameronrohani (approved):

PR #4564, opened 2025-01-23
feat(api-client): parameter examples support

### LINKED ISSUE #4188: Example Dropdowns in Test Request
### What happens?

In Test Request, if there is an example for a path or query parameter in the spec, the value for the example will be populated for the appropriate parameter.   Similarly, if there are example Request Bodies, a drop-down appears, allowing the user to select one of the examples.

It would be great if a similar drop-down was provided so the user could select from one of multiple examples for a query or path parameter in a similar way.  You currently support dropdowns for boolean or enumerated parameters, so it seems like the same style could also be used for examples. 

I expect there to be a dropdown that shows all the examples for a parameter if multiple examples are defined in the spec.

Photos available in this discord thread: https://discord.com/channels/1135330207960678410/1135330208627564597/1315310354015850547

It would be awesome if this issue gets picked up if someone could respond that thread when an update is available!

### OpenAPI Document

_No response_

antlio:
**Problem**
setting examples in parameters is not currently retrieved in the client and displays only the first example value.

**Solution**
this pr fixes #4188 by displaying parameters examples in the client as a dropdown.

| before | after |
| -------|------|
| image | image |

**Checklist**

### DIFF
--- a/.changeset/chilly-rivers-march.md
+++ b/.changeset/chilly-rivers-march.md
@@ -1,0 +1,6 @@
+---
+'@scalar/api-client': patch
+'@scalar/oas-utils': patch
+---
+
+feat: adds parameter examples support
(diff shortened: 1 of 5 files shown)


marclave (approved):
🤯 

awesome work : )

PR #4570, opened 2025-01-23
feat(api-client): request example array multiselect

antlio:
**Problem**
currently user cannot select multiple value in the request example dropdown selector for array.

**Solution**
this pr adds multiselect support to request example array as seen below:

https://github.com/user-attachments/assets/482ea9ef-8012-44d8-a8a1-8757aa69cd6d

**Checklist**


marclave (approved):
UNREAL @antlio , i was actually just needing this : )

PR #4573, opened 2025-01-24
fix(api-reference): inconsistent scroll position on page load when navigating to

amritk:
**Problem**
On bigger specs when we load the first operation of a tag, sometimes there is a jump and it scrolls to the wrong element.

**Solution**
We do NOT lazy load any tags until the open one, this should eliminate the jump!

**Checklist**

### DIFF
--- a/packages/api-reference/src/components/Content/Tag/TagList.vue
+++ b/packages/api-reference/src/components/Content/Tag/TagList.vue
@@ -4,13 +4,13 @@ import { Operation } from '@/features/Operation'
 import { useNavState, useSidebar } from '@/hooks'
 import { useWorkspace } from '@scalar/api-client/store'
 import { ScalarErrorBoundary } from '@scalar/components'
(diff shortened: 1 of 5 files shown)


marclave (approved):
LETS GO

packages/api-reference/src/components/Content/Tag/TagList.vue
```diff
@@ -19,19 +19,31 @@ const { getOperationId, getTagId, hash } = useNavState()
...
 
-const tagLayout = computed<typeof Tag>(() =>
+const tagLayout = computed(() =>
   props.layout === 'classic' ? TagAccordion : Tag,
 )
 
-// If the first load is models, we do not lazy load tags/operations
-const isLazy = props.layout !== 'classic' && !hash.value.startsWith('model')
+/**
+ * All tags on the list UNTIL the one afterfirst open tag should not be lazy loaded. However their operations should be.
```
amritk, line 27: Typo

PR #4576, opened 2025-01-24
fix(api-client): sidebar outline + markup

antlio:
**Problem**
currently api client sidebar are using wrong markup + outline focus is off.

**Solution**
this pr is a proposal to reset outline by offsetting it by default + update markups.

| before | after |
| -------|------|
| image | image |
| top outline focus is cropped | top outline is now fully visible | 

**Checklist**

### DIFF
--- a/packages/themes/src/reset.css
+++ b/packages/themes/src/reset.css
@@ -102,6 +102,12 @@
     border-radius: var(--scalar-radius);
   }
 
+  /** Set outline offset for buttons */
+  button:focus-visible,
+  [role='button']:focus-visible {
+    outline-offset: -1px;
+  }
+
   /** Set the default cursor for buttons. */
   button,
   [role='button'] {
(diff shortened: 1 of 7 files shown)


packages/themes/src/reset.css
```diff
@@ -32,6 +32,7 @@
     outline-width: 1px;
     outline-style: none;
     outline-color: var(--scalar-color-accent);
+    outline-offset: -1px;
```
hwkr, line 35: I don't know if we want this everywhere 🤔

antlio: yes actually it might feel too cluttered in the reference on second tests

antlio: i'll just set it up in the sidebar items then

antlio: proposing to set it only on button

hwkr: Sure

amritk (approved):
This ones more for @cameronrohani and @hwkr but lgtm!

hwkr (approved):
Looks good! Still seeing some clipping on the environment page.

Google Chrome-2025-01-25-19-52-46@2x

PR #4578, opened 2025-01-24
fix(api-client): type mismatch for the server variable

hanspagel:
Types are broken in main, let’s fix this swiftly.

### DIFF
--- a/packages/api-client/src/components/Server/types.ts
+++ b/packages/api-client/src/components/Server/types.ts
@@ -8,4 +8,6 @@ export type ServerVariableValues = {
   [variable: string]: string
 }
 
-export type ServerVariable = OpenAPIV3_1.ServerVariableObject
+export type ServerVariable =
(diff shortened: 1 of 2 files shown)


packages/api-client/src/components/Server/types.ts
```diff
@@ -8,4 +8,6 @@ export type ServerVariableValues = {
   [variable: string]: string
 }
 
-export type ServerVariable = OpenAPIV3_1.ServerVariableObject
+export type ServerVariable =
+  | OpenAPIV3.ServerVariableObject
+  | OpenAPIV3_1.ServerVariableObject
```
antlio, line 13: cc @amritk finally required to maintain 3

antlio (approved):

packages/api-client/src/components/Server/types.ts
hanspagel: just temporary though

PR #4583, opened 2025-01-24
test(api-reference): testing SSR for a bunch of real-world examples

hanspagel:
**Problem**
Currently, we only test a basic example and the GitHub REST API definition in an SSR environment. This is already good, but maybe not broad enough.

**Solution**
With this PR we’re testing a bunch of real-world examples.

**Preview**

Screenshot 2025-01-24 at 13 24 34

**Checklist**

- [ ] I’ve added a changeset (`pnpm changeset`). (just a test, doesn’t need a changeset)


marclave (approved):
OMG MORE TESTS!!

amritk (approved):
We can later check for specific things like code samples when we have them fixed up!

PR #4585, opened 2025-01-24
fix(components): floating target ref

antlio:
**Problem**
floating component is firing log error as it doesn't find id.

**Solution**
this pr sets the target ref on floating component mounted to ensure element availability.

**Checklist**

### DIFF
--- a/packages/components/src/components/ScalarFloating/ScalarFloating.vue
+++ b/packages/components/src/components/ScalarFloating/ScalarFloating.vue
@@ -44,7 +44,7 @@ const floatingRef: Ref<HTMLElement | null> = ref(null)
 const wrapperRef: Ref<HTMLElement | null> = ref(null)
 
 const targetRef = computed(() => {
-  if (typeof window !== 'undefined') {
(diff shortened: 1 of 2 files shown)


packages/components/src/components/ScalarFloating/ScalarFloating.vue
```diff
@@ -42,24 +42,27 @@ defineOptions({ inheritAttrs: false })
 
 const floatingRef: Ref<HTMLElement | null> = ref(null)
 const wrapperRef: Ref<HTMLElement | null> = ref(null)
+const targetRef = ref<HTMLElement | undefined>(undefined)
 
-const targetRef = computed(() => {
```
hwkr, line 46: We should probably keep it as computed so it can be updated but we could do a check to see if the component is mounted 🤔

antlio: good point, i updated accordingly!

hwkr (approved):
Looks good! We could add a test for this too if you want 😊

amritk (approved):

Issue #4596, opened 2025-01-27
No request imported while collection successfully created


antlio:
### What happens?

While I can import OpenAPI specifications, I receive a message stating "imported successfully," yet no requests are actually imported.

[scalar-invalid-openapi.json](https://github.com/user-attachments/files/18555311/scalar-invalid-openapi.json)

_originally shared by Germano [in discord](https://discord.com/channels/1135330207960678410/1313151751872053338)_

hanspagel:
Fixed in #4598

PR #4598, opened 2025-01-27
fix(api-client): No request imported while collection successfully created

RemcoGoy:
**Problem**
Currently, when importing a spec file, the collection is created, but no requests are shown

**Solution**
With this PR, I'm fixing the way requests are added to the children of the collection. In the example spec, `tags` was an empty list, so `if (r.tags)` evaluated to `true`, added an extra step to check list length.

solves #4596 (see this issue for example spec file)

**Checklist**

---

Which references should be added as children? Is it fine that only requests are added to the children property? Should others also be added (like examples)?

I'm also not sure if this is the correct place to add the requests to the children, or if this needs to happen anywhere else?

### DIFF
--- a/packages/oas-utils/src/transforms/import-spec.ts
+++ b/packages/oas-utils/src/transforms/import-spec.ts
@@ -382,7 +382,7 @@ export async function importSpecToWorkspace(
 
   // Add the request UIDs to the tag children (or collection root)
   requests.forEach((r) => {
-    if (r.tags) {
+    if (r.tags && r.tags.length > 0) {
       r.tags.forEach((t) => {
         tagMap[t].children.push(r.uid)
       })
--- a/packages/oas-utils/src/transforms/import-spec.ts
+++ b/packages/oas-utils/src/transforms/import-spec.ts
@@ -382,7 +382,7 @@ export async function importSpecToWorkspace(
 
   // Add the request UIDs to the tag children (or collection root)
   requests.forEach((r) => {
-    if (r.tags && r.tags.length > 0) {
+    if (r.tags?.length) {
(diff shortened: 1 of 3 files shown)


amritk (changes requested):
hey @RemcoGoy thanks for the PR! we actually do add requests to the children to the collection inside of the import function, and out here all we need to do is add to the mutators. Maybe we should document this better. 

We can come up with another fix for the issue!

RemcoGoy:
> hey @RemcoGoy thanks for the PR! we actually do add requests to the children to the collection inside of the import function, and out here all we need to do is add to the mutators. Maybe we should document this better.
> 
> We can come up with another fix for the issue!

Thanks for the reply! I'll look into it as soon as possible.

RemcoGoy:
@amritk , I think I found the actual issue this time. `r.tags` was an empty list in the example, so `if (r.tags)` evaluated to `true`. I added an extra step to check for list length.

amritk:
@RemcoGoy awesome find! Let me just suggest some minor style changes for brevity then we can merge

packages/oas-utils/src/transforms/import-spec.ts
```diff
@@ -382,7 +382,7 @@ export async function importSpecToWorkspace(
 
   // Add the request UIDs to the tag children (or collection root)
   requests.forEach((r) => {
-    if (r.tags) {
+    if (r.tags && r.tags.length > 0) {
```
amritk, line 385:
```suggestion
    if (r.tags?.length) {
```

amritk (approved):
Nice work!!

PR #4611, opened 2025-01-28
feat(ci): added bust cache to github action

amritk:
Not sure if this will work but YOLO lets test it in prod


xC0dex:
> Not sure if this will work but YOLO lets test it in prod

I like this mindset 🤣

PR #4634, opened 2025-01-29
feat(api-client): accept header

### LINKED ISSUE #4534: ACCEPT header not pre-filled
### What happens?

when I specify a mime type for requests, the accept header is not populated

Image

under headers in in the Test request modal, it should show the ACCEPT header

### OpenAPI Document

https://sandbox.scalar.com/e/8h7iI#tag/default/GET/v1/links

antlio:
**Problem**
currently we do not pre-fill the header with default ones as highlighted in #4534 

**Solution**
this pr adds `Accept` header by default and fixes #4534. we might want to adds other ones:
- Accep-Encoding
- User-Agent (in desktop app usage)
- Connection

| before | after |
| -------|------|
| image | image |
| content type header only | accept header prefilled | 

**Checklist**

- [x] Migration

### DIFF
--- a/packages/api-client/src/views/Request/RequestSection/RequestBody.vue
+++ b/packages/api-client/src/views/Request/RequestSection/RequestBody.vue
@@ -319,12 +319,23 @@ const updateActiveBody = (type: Content) => {
     }
   }
   // Add header if doesn't have one
-  else if (header)
-    headers.push({
-      key: 'Content-Type',
-      value: header,
-      enabled: true,
-    })
+  else if (header) {
+    const lastHeader = headers[headers.length - 1]
+    // Add header before last if empty to prevent empty row duplication
+    if (lastHeader && lastHeader.key === '' && lastHeader.value === '') {
+      headers.splice(headers.length - 1, 0, {
+        key: 'Content-Type',
+        value: header,
+        enabled: true,
+      })
+    } else {
+      headers.push({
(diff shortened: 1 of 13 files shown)


packages/api-client/src/views/Request/RequestSection/RequestBody.vue
```diff
@@ -319,12 +319,23 @@ const updateActiveBody = (type: Content) => {
...
   // Add header if doesn't have one
-  else if (header)
-    headers.push({
-      key: 'Content-Type',
-      value: header,
-      enabled: true,
-    })
+  else if (header) {
+    const lastHeader = headers[headers.length - 1]
+    // Add header before last if empty to prevent empty row duplication
```
amritk, line 324: very nice, but now I'm thinking should we just add this to the import? would be a little cleaner. What do you think

amritk (approved):
Awesome!

packages/api-client/src/views/Request/RequestSection/RequestBody.vue
antlio: we could but we would still need logic for when we select a body type let's say get -> post to populate it right?

amritk: ya true

amritk (approved):

PR #4639, opened 2025-01-29
fix(themes): box shadow variable

antlio:
**Problem**
recent changes to box shadow in darkmode mistakenly set it as inset which breaks the ui.

**Solution**
this pr removes the inset from the box shadow variable.

| before | after |
| -------|------|
| image | image |
| missing top border | whole border back | 

**Checklist**

### DIFF
--- a/packages/themes/src/variables.css
+++ b/packages/themes/src/variables.css
@@ -60,7 +60,7 @@
   --scalar-button-1-color: black;
 
   --scalar-shadow-1: 0 1px 3px 0 rgb(0, 0, 0, 0.1);
-  --scalar-shadow-2: inset 0 0 0 0.5px var(--scalar-border-color),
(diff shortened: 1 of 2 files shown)


marclave (approved):

packages/themes/src/variables.css
```diff
@@ -60,7 +60,7 @@
   --scalar-button-1-color: black;
 
   --scalar-shadow-1: 0 1px 3px 0 rgb(0, 0, 0, 0.1);
-  --scalar-shadow-2: inset 0 0 0 0.5px var(--scalar-border-color),
```
hwkr, line 63: @cameronrohani was this one supposed to be inset? I can't remember

hwkr (approved):
Looks good to me but @cameronrohani should have a look before we merge

cameronrohani (approved):

PR #4641, opened 2025-01-29
fix(api-client): settings UI

antlio:
**Problem**
some api client settings ui code seems duplicated giving funky borders.

**Solution**
this pr removes duplication and sets variant and color properly.

| before | after |
| -------|------|
| image | image |
| inactive checkbox border is off  | active checkbox use primary bg and border are fixed | 

**Checklist**

### DIFF
--- a/packages/api-client/src/views/Settings/SettingsGeneral.vue
+++ b/packages/api-client/src/views/Settings/SettingsGeneral.vue
@@ -2,7 +2,7 @@
 import IntegrationLogo from '@/components/ImportCollection/IntegrationLogo.vue'
 import { useWorkspace } from '@/store'
 import { useActiveEntities } from '@/store/active-entities'
-import { ScalarButton, ScalarIcon } from '@scalar/components'
+import { ScalarButton, ScalarIcon, cva, cx } from '@scalar/components'
 import {
   type IntegrationThemeId,
   type ThemeId,
@@ -59,6 +59,16 @@ const getThemeColors = (
 const changeTheme = (themeId: ThemeId) => {
   workspaceMutators.edit(activeWorkspace.value?.uid ?? '', 'themeId', themeId)
 }
+
+const buttonStyles = cva({
+  base: 'w-full shadow-none text-c-1 justify-start pl-2 gap-2 border-1/2',
+  variants: {
+    active: {
+      true: 'bg-primary text-c-1 hover:bg-inherit',
+      false: 'bg-b-1 hover:bg-b-2',
+    },
+  },
+})
 </script>
 <template>
   <div class="bg-b-1 w-full h-full overflow-auto">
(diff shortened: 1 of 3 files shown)


packages/api-client/src/views/Settings/SettingsGeneral.vue
```diff
@@ -59,6 +59,16 @@ const getThemeColors = (
 const changeTheme = (themeId: ThemeId) => {
   workspaceMutators.edit(activeWorkspace.value?.uid ?? '', 'themeId', themeId)
 }
+
+const buttonStyles = cva({
+  base: 'w-full shadow-none text-c-1 justify-start pl-2 gap-2 border-1/2',
```
amritk, line 64: So I would normally put all of my variants inside of ScalarButton as we shouldn't usually differ too much from the set button styles. BUT if this is a one off I guess its fine here. @hwkr what do you think?

+points for CVA.

packages/api-client/src/views/Settings/SettingsGeneral.vue
```diff
@@ -158,7 +176,11 @@ const changeTheme = (themeId: ThemeId) => {
                 )
               ">
               <div
-                class="flex items-center justify-center w-5 h-5 rounded-full border-[1.5px] p-1">
+                class="flex items-center justify-center w-5 h-5 rounded-full border-[1.5px] p-1"
+                :class="{
+                  'bg-c-accent border-transparent text-b-1':
+                    !activeWorkspace?.proxyUrl,
+                }">
```
amritk, line 183:
```suggestion
                :class="!activeWorkspace?.proxyUrl && 'bg-c-accent border-transparent text-b-1'">
```
I believe you can also use the short circuit for vue class. But whatever you think is better, no preference here

amritk (approved):
Just the one question for @hwkr, but the code is fine otherwise. 

tagging in @cameronrohani for visual check

packages/api-client/src/views/Settings/SettingsGeneral.vue
antlio: i usually do i guess i left as it was, let's do it!

packages/api-client/src/views/Settings/SettingsGeneral.vue
antlio: i guess i could come up with a better name too, but as we use these buttton styles in two components might make sense for a variant in the component button!

amritk: love me some variants in the button component 😄

hwkr: I wonder if this should be a different component or an extension of the button since it's kind of a toggle / selectable button. This seems fine for now though

hwkr (approved):

cameronrohani (approved):

PR #4651, opened 2025-01-30
feat(api-client): added some tests for the new server dropdown

amritk:
**Problem**
The add server button shows up on the references client modal

**Solution**
It no longer shows up there, also added tests 🧪 

**Checklist**

### DIFF
--- a/packages/api-client/src/components/Server/ServerDropdown.vue
+++ b/packages/api-client/src/components/Server/ServerDropdown.vue
@@ -1,4 +1,5 @@
 <script setup lang="ts">
+import { useLayout } from '@/hooks/useLayout'
 import { useWorkspace } from '@/store/store'
 import {
   ScalarButton,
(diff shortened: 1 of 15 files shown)


packages/api-client/src/components/Server/ServerDropdown.vue
```diff
@@ -1,5 +1,6 @@
 <script setup lang="ts">
-import { useWorkspace } from '@/store'
+import { useLayout } from '@/hooks/useLayout'
+import { useWorkspace } from '@/store/store'
```
antlio, line 3:
```suggestion
import { useLayout } from '@/hooks'
import { useWorkspace } from '@/store'
```

or should we keep import more specific in genearl?

antlio: ok reading the rest of the code i guess we might!

antlio (approved):
legit good! gave it a ui test through client / reference and seems flawless

packages/api-client/src/components/Server/ServerDropdown.vue
amritk: Yes we need to start importing from each file and not the index files. Just do it as you go. It will help Hans prepare for module: nodenext

marclave (approved):

PR #4654, opened 2025-01-30
fix(snippetz): `findPlugin` undefined

antlio:
**Problem**
playground snippetz usage is throwing an error. 

```console
Cannot read properties of undefined (reading 'findPlugin')
```

**Solution**
this pr returns object in the snippetz function + updates playground style.

| before | after |
| -------|------|
| image | image |
| example and input in one code block | display example and output distinctly | 

**Checklist**

### DIFF
--- a/.changeset/gold-buses-cheer.md
+++ b/.changeset/gold-buses-cheer.md
@@ -1,0 +1,5 @@
+---
+'@scalar/snippetz': patch
+---
+
+feat: updates playground style
--- a/.changeset/tall-brooms-change.md
+++ b/.changeset/tall-brooms-change.md
@@ -1,0 +1,5 @@
+---
+'@scalar/snippetz': patch
+---
+
+fix: updates snippetz function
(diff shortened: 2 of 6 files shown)


packages/snippetz/package.json
```diff
@@ -228,6 +228,7 @@
   "devDependencies": {
     "@scalar/build-tooling": "workspace:*",
     "@types/har-format": "^1.2.15",
-    "vite": "^5.4.10"
+    "vite": "^5.4.10",
+    "vitest": "^1.6.0"
```
hanspagel, line 232: do we need this here? should be available through the workspace already, or am I wrong? 🤔

hanspagel (changes requested):
Love it! There’s an open PR for it already, can you use the solution for the actual error from the other PR? It’s actually by @geoffgscott, I’ve just extracted it from a bigger PR and never finished it.

https://github.com/scalar/scalar/blob/ceb4164de58da3d977ff4081d1c82ea9dc54db35/packages/snippetz/src/snippetz.ts

packages/snippetz/package.json
antlio: i guess we don't if we run it from root but thought if someone would use the package they might want it - i removed it

hanspagel (approved):
Great fix!!

PR #4656, opened 2025-01-30
feat(api-client): code snippets ✂️

antlio:
This PR brings our code examples from the API reference to the API client. 👀

| preview |
| -------|
| image |
| adds request preview section | 

**Checklist**


cameronrohani:
awesome work!

2 things

we need to add a max height to the language selector otherwise it will be scrollable: 
image

other thing is, since this is connected to all of the input fields you can interact with I think we need to separate it out to imply that you can't interact with it.... i.e pushing it to the bottom like so: 
image

then opening the content above the top and applying a box shadow and reducing the scroll space of its siblings to the new height of the container.
image


Let me know if this makes sense? Let's make this awesome 🥰

antlio:
> we need to add a max height to the language selector otherwise it will be scrollable:

oh! this was supposed to be handle by [5a41f27](https://github.com/scalar/scalar/pull/4656/commits/5a41f27155d0119eadc66b4b341816226297e6cb) - will fix.

> other thing is, since this is connected to all of the input fields you can interact with I think we need to separate it out to imply that you can't interact with it.... i.e pushing it to the bottom like so:

i agree  + 1 for icon and selector position - isn't it hard to notice though? 

another take - shouldn't we take advantage of the response area where we could toggle preview/response display - this would offer optimal readability while typing <> previewing. it could then toggle back to response on request sent?

cameronrohani:
> > we need to add a max height to the language selector otherwise it will be scrollable:
> 
> oh! this was supposed to be handle by [5a41f27](https://github.com/scalar/scalar/pull/4656/commits/5a41f27155d0119eadc66b4b341816226297e6cb) - will fix.
> 
> > other thing is, since this is connected to all of the input fields you can interact with I think we need to separate it out to imply that you can't interact with it.... i.e pushing it to the bottom like so:
> 
> i agree + 1 for icon and selector position - isn't it hard to notice though?
> 
> another take - shouldn't we take advantage of the response area where we could toggle preview/response display - this would offer optimal readability while typing <> previewing. it could then toggle back to response on request sent?

[that is what I initially spec'd but the team landed on putting it on the left side IIRC](https://apidocumentation.slack.com/archives/C07RNPTKCB1/p1731959782717709)

.changeset/loud-rats-brush.md
```diff
@@ -0,0 +1,5 @@
+---
+'@scalar/snippetz': patch
+---
+
+chore: adds vitest dev dependency
```
hanspagel, line 5: this changeset can be deleted, right?

packages/snippetz/package.json
```diff
@@ -228,6 +228,7 @@
   "devDependencies": {
     "@scalar/build-tooling": "workspace:*",
     "@types/har-format": "^1.2.15",
-    "vite": "^5.4.10"
+    "vite": "^5.4.10",
+    "vitest": "^1.6.0"
```
hanspagel, line 232: do we need this?

antlio:
as discussed w/ @hanspagel in order to avoid to create unnecessary code:

course of action:
⊢ make it the same component/block that we:
⊢--- use in the reference
⊢--- use in the API client
⊢--- expose as a standalone embed

point of attention
- avoid using a global state for the selected HTTP client
- distinguished target client available in the client vs reference (that welcome custom ones)
- keep custom examples defined in the reference context only

packages/api-client/src/views/Request/RequestSection/RequestPreview.vue
```diff
@@ -58,7 +68,7 @@ const availablePlugins = computed(() => {
   <div class="sticky mt-auto bottom-0 w-full">
     <ViewLayoutCollapse
       class="group/preview -mt-0.25 w-full"
-      :defaultOpen="false"
+      :defaultOpen="true"
```
antlio, line 71: should we keep it open at all time?

hanspagel: ah, just for debugging sorry

hanspagel:
Just leaving this here for the moment: URL starts with a slash if you just put in an URL without setting up a server properly

Screenshot 2025-02-06 at 16 42 34

amritk:
Taking over, will merge over new reference code!

hwkr (approved):
Looks great! Two ideas (could be fast follows):

1. Could we use the prettier labels from language picker? And remove the stuff before the `/` since it's already in the group?

Google Chrome-2025-02-06-14-05-28@2x
Google Chrome-2025-02-06-14-06-07@2x


2. Some of the languages don't have show anything for an empty request, maybe we could add a little "Enter a URL to generate your code snippet"

Google Chrome-2025-02-06-14-07-18@2x

Let me know if you want me to hop in tackle any of these

hwkr:
Pushed some improvements to the combobox so it scrolls to the correct item when opened for long lists.

Before / After:

https://github.com/user-attachments/assets/0e527609-64fc-4415-85ed-08641989e9f1

amritk (approved):
Back in business! Not ideal as we will still get escaped vars in the URL, but it works!
image

hanspagel (approved):
Banger PR, let’s roll it out!

PR #4662, opened 2025-01-31
fix(api-reference): code example authentication

### LINKED ISSUE #4652: Scalar not respecting security
### What happens?

Scalar reports auth for endpoints that should not have auth 

Non authed endpoint represented correctly

### OpenAPI Document

https://socksproxyapi.darrenmc.dev/docs/openapi.json

antlio:
**Problem**
currently when a security scheme is selected in an api reference, the security header gets added on every endpoint regardless of the endpoint security requirement.

**Solution**
this pr adds a check for reference to the request security presence in order to only add the header accordingly and fixes #4652.

**Checklist**

### DIFF
--- a/packages/api-client/src/libs/send-request/send-request.ts
+++ b/packages/api-client/src/libs/send-request/send-request.ts
@@ -179,6 +179,7 @@ export const createRequestOperation = ({
   selectedSecuritySchemeUids = [],
   server,
   status,
+  isReadOnly,
 }: {
   environment: object | undefined
   example: RequestExample
@@ -189,6 +190,7 @@ export const createRequestOperation = ({
   selectedSecuritySchemeUids?: Operation['selectedSecuritySchemeUids']
   server?: Server
   status?: EventBus<RequestStatus>
+  isReadOnly?: boolean
 }): ErrorResponse<{
   controller: AbortController
   sendRequest: () => SendRequestResponse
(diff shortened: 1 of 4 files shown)


packages/api-client/src/libs/send-request/send-request.ts
```diff
@@ -241,52 +243,56 @@ export const createRequestOperation = ({
 
     // We flatten the array of arrays for complex auth
     const flatSelectedSecuritySchemeUids = selectedSecuritySchemeUids.flat()
+    // Handle auth schemes based on request security for reference
+    if (isReadOnly || request.security) {
```
amritk, line 247: shouldn't it be !isReadOnly?

antlio: 🧞‍♂️

hanspagel (changes requested):
I want to be strict with that PR:

* @amritk asked for a product decision and approved then, did we settle on something? Just from reading the comments I feel I miss a piece of information.
* We need tests to cover this behaviour.
* And please take into account that an OpenAPI document can have a global Security Requirement Object:

https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.1.md#openapi-object

Screenshot 2025-02-04 at 14 18 05

antlio:
> @amritk asked for a product decision and approved then, did we settle on something? Just from reading the comments I feel I miss a piece of information.

yes sorry we didn't share here but we decided to move forward with this after discussing it out w/ amrit + geoff

> need tests to cover this behaviour.

a test has been added regarding this - do you think about something more here?

> global Security Requirement Object

this should remain here, unless otherwise did you flag something?

amritk:
@hanspagel yea we do use the global object as a backup. that logic is unrelated to this but you can find the tests here:
https://github.com/scalar/scalar/blob/main/packages/api-client/src/views/Request/libs/auth.test.ts#L89

we could always use more tests though!

hanspagel:
Waiting for #4656. 👍

PR #4665, opened 2025-01-31
fix(api-reference): setting initial security when there are no requirements

amritk:
**Problem**
It was incorrectly selecting security scheme uids when there was no security requirements on the collection

**Solution**
We add a test and only select security requirements from the collection

**To Test**
You can use this spec
https://socksproxyapi.darrenmc.dev/docs/openapi.json

**Checklist**


antlio (approved):
neat

PR #4666, opened 2025-01-31
fix(Scalar.AspNetCore): rename Proxy property to ProxyUrl

xC0dex:
**Problem**
I noticed that the `Proxy` property is marked as deprecated.

**Solution**
I renamed it to `ProxyUrl`.

**Checklist**

- [ ] ~I’ve added a changeset (`pnpm changeset`).~
- [ ] ~I’ve added tests for the regression or new feature.~
- [ ] ~I’ve updated the documentation.~


marclave (approved):
nice work @xC0dex ! :)

PR #4667, opened 2025-01-31
fix(api-reference): hide auth section in references when there's no securitySchemes

amritk:
**Problem**
We show an optional auth when there's no auth defined in the schema

**Solution**
We hide the auth section
 
before: 
image
after:
image

**Checklist**

### DIFF
--- a/.changeset/nervous-rules-grin.md
+++ b/.changeset/nervous-rules-grin.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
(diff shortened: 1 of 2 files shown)


cameronrohani:
W PR!

image

antlio:
nice one! should we also hide it in the modal too? as of right now this would be displayed with empty selector: (could be ff)

image

antlio (approved):
great call! auth modal to be handle in there or fast follow

marclave (approved):

PR #4672, opened 2025-02-03
fix(api-client): request auth modal

antlio:
**Solution**
this PR is following up with #4667 in order to only display auth in security schemes presence in reference client modal also.

| before | after |
| -------|------|
| image | image |
| auth section displayed without security presence | hidden auth section in no security presence | 

**Checklist**


amritk (approved):
For some reason I thought it already existed in the client 😖

PR #4673, opened 2025-02-03
chore(tooling): fine grained proxy deployment

bgrcs:
**Problem**
Client proxy is now deployed cross-regionally, with a great improvement in response metrics (this is the pattern we will follow going forward for other services).

Now we just need some slight extra control over instance sizes and numbers.

**Solution**
Just adding two extra flags to increase the number of concurrent requests each proxy instance can handle, and the minimum number of instances - we will be updating both as needed based on metrics.

**Checklist**


hanspagel (commented):
LGTM, but needs approval by @geoffgscott.

geoffgscott (approved):
LGTM

PR #4677, opened 2025-02-03
fix(api-reference): add hash prefix to heading IDs

amritk:
**Problem**
The hash prefix was missing and not working on ORG

**Solution**
We ensure the hash prefix is added to the heading IDs

**Checklist**


hanspagel (approved):
Oops, great fix! I’ve added tests and refactored it to use import aliases and avoid going through the barrel files.

PR #4679, opened 2025-02-04
fix(api-reference): example format

### LINKED ISSUE #4678: Example `number` value being incorrectly displayed as a string
### What happens?

Using v`0.4.4`

For some reason, the example is displaying the `1` in my example here as `"1"` incorrectly in the documentation, but correct in the curl example.

Image

Image

In my OpenAPI schema, the number is of the correct type: 

Image

"Examples" in the UI should read `[1, [4,5]]`

### OpenAPI Document

_No response_

antlio:
**Problem**
currently number parameter example are being displayed as string in the reference and fixed #4678.

**Solution**
this PR updates the type formatting to maintain de type displayed.

| before | after |
| -------|------|
| image | image |
| example value as string | example value as number | 

**Checklist**

### DIFF
--- a/.changeset/fuzzy-teachers-push.md
+++ b/.changeset/fuzzy-teachers-push.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
+fix: updates format example helper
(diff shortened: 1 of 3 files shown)


hanspagel (approved):
Great fix, let’s goo

PR #4687, opened 2025-02-04
fix(components): set z-index on ScalarMenu by default

hwkr:
Because this doesn't use `ScalarFloating` we need to set a `z-index`

**Checklist**

### DIFF
--- a/packages/components/src/components/ScalarMenu/ScalarMenu.vue
+++ b/packages/components/src/components/ScalarMenu/ScalarMenu.vue
@@ -61,7 +61,7 @@ function close() {
     <DropdownMenu.Content
       align="start"
       :as="ScalarDropdownMenu"
-      class="max-h-radix-popper"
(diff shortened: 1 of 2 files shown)


amritk (approved):
Many thanks, will yolo test this in org post release

marclave (approved):

packages/components/src/components/ScalarMenu/ScalarMenu.vue
```diff
@@ -61,7 +61,7 @@ function close() {
     <DropdownMenu.Content
       align="start"
       :as="ScalarDropdownMenu"
-      class="max-h-radix-popper"
+      class="max-h-radix-popper z-context"
```
antlio, line 64: this seems to set the submenu under the menu:
image

hwkr: Ah we probably need it to use context plus

PR #4690, opened 2025-02-04
fix(api-client): content-type request 

### LINKED ISSUE #4562: Custom content type open api support
### What happens?

Got a small issue i've ran into today -- I've got a Json PATCH endpoint that requires a Content-Type header of 'application/json-patch+json' (.NET backend) to function correctly. I've got a custom document transformer that sets the expected content-type of the body correctly, and this is reflected in the documentation (see attached) but when hitting 'Test Request' it is just defaulted to application/json

Is there anyway to override this default Content-Type header for PATCH requests -- maybe on the Scalar side of things?

Image

Image

_originally posted by Ceece [in discord](https://discord.com/channels/1135330207960678410/1331254815866748979)_

antlio:
**Problem**
currently when using a custom content type it is not retrieved at the client modal level and also override the example once back to the reference.

**Solution**
this pr sets the request content-type if present and fixes #4562.

| before | after |
| -------|------|
| image | image |
| content-type is using json | content-type is using custom | 

**Checklist**

### DIFF
--- a/packages/api-client/src/views/Request/RequestSection/RequestBody.vue
+++ b/packages/api-client/src/views/Request/RequestSection/RequestBody.vue
@@ -73,7 +73,12 @@ const activeExampleContentType = computed(() => {
     return activeExample.value.body.raw.encoding
   }
 
-  return 'none'
+  // Set content type from request if present
+  const contentType = Object.keys(
+    activeRequest.value?.requestBody?.content || {},
+  )[0]
+
+  return contentType || 'none'
 })
 /** Selected ref from options above */
 const selectedContentType = computed({
--- a/.changeset/lucky-papayas-look.md
+++ b/.changeset/lucky-papayas-look.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
+
+fix: sets content type from request


packages/api-client/src/views/Request/RequestSection/RequestBody.vue
```diff
@@ -73,7 +73,12 @@ const activeExampleContentType = computed(() => {
     return activeExample.value.body.raw.encoding
   }
 
-  return 'none'
+  // Set content type from request if present
+  const contentType = Object.keys(
+    activeRequest.value?.requestBody?.content || {},
```
amritk, line 78: should we start storing content type on the example? Where did we land on example per content type again?

amritk: lets chat rq

antlio: its more of a quick fix to be refactored once the above gets in - #4485

amritk: Ok send it

amritk (approved):

PR #4692, opened 2025-02-05
feat(oas-utils): alias request with operation

amritk:
**Problem**
request was constantly bumping into the global request class causing typescript errors, also auto import didn't work. 

**Solution**
We alias request with operation which brings it more in line with our codebase + openapi spec.

**Checklist**

### DIFF
--- a/.changeset/poor-peaches-exist.md
+++ b/.changeset/poor-peaches-exist.md
@@ -1,0 +1,5 @@
+---
+'@scalar/oas-utils': patch
+---
+
+chore: alias request to operation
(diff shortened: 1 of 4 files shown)


geoffgscott (commented):
This looks great. We should probably bite the bullet and do a find/replace based on imports

amritk:
> This looks great. We should probably bite the bullet and do a find/replace based on imports
I figure we can use this for a slow change as opposed to an all at once with a bunch of changes

antlio (approved):
lgtm wiz!

PR #4698, opened 2025-02-05
fix(api-client): cookies domain

antlio:
**Problem**
currently on cookie creation the domain is required preventing the creation of a cookie without it.

**Solution**
this pr sets the domain field as optional.

| before | after |
| -------|------|
| image | image |

**Checklist**


hanspagel:
The [optional] suffix is a bit strange. Haven’t seen that anywhere, and the white colon seems off then.

I think we should just omit the text suffix.

If we want to introduce a clear visual indication for required or for optional fields, we should do that with consciously looking at all modals/forms I guess.

Screenshot 2025-02-06 at 12 24 14

hanspagel (approved):
Yesss

PR #4703, opened 2025-02-06
fix(api-client): custom parameter disappearing

WDaan:
**Problem**
Currently, when you try to provide a custom value for a parameter that has examples, the value disappears as soon as you hit enter.

https://github.com/user-attachments/assets/49c49699-b289-4d68-902a-e5fa2426cf5c

I think I tracked it down to events being emitted too much and overwriting your custom value...

**Solution**
With this PR, I removed those (unneeded?) event emits and got 'working' behavior. Meaning, the custom value is not dissapearing. The value itself however is still not being added to the list of available entries. Unsure if that's desired behavior?

https://github.com/user-attachments/assets/4c699d8d-503d-4960-a285-750763d6bbae

**Checklist**

### DIFF
--- a/packages/api-client/src/components/DataTable/DataTableInputSelect.vue
+++ b/packages/api-client/src/components/DataTable/DataTableInputSelect.vue
@@ -41,15 +41,14 @@ const updateSelected = (value: string) => {
 const addCustomValue = () => {
   if (customValue.value.trim()) {
     updateSelected(customValue.value)
-    customValue.value = ''
   }
 }
 
 const handleBlur = () => {
   if (!customValue.value.trim()) {
     emit('update:modelValue', '')
-    addingCustomValue.value = false
   }
+  addingCustomValue.value = false
 }
 
 const isSelected = (value: string) => {
--- a/.changeset/lazy-clouds-look.md
+++ b/.changeset/lazy-clouds-look.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
+
+Allow custom value in parameters


antlio (changes requested):
great contribution and fix! the only stepback is that now on blur it doesn't set the value - do you mind looking into it?

antlio:
i think this might be the one:
image

WDaan:
> i think this might be the one: image

Sorry for the delay.
You might be right, making that change and reverting the other yields the same result.

antlio (changes requested):
thank you so much for the update, last bit: i think we need to update that function as well, so we get the value when adding a custom one and sending request (without key enter):

```ts
const handleBlur = () => {
  if (!customValue.value.trim()) {
    emit('update:modelValue', '')
  }
  addingCustomValue.value = false
}

File: packages/api-client/src/components/DataTable/DataTableInputSelect.vue [hunk 1]
```diff
@@ -29,10 +29,6 @@ const addingCustomValue = ref(false)
 const customValue = ref('')
 const inputRef = ref<HTMLInputElement | null>(null)
 
-watch(customValue, (newValue) => {
-  emit('update:modelValue', newValue)
-})
-
```
antlio, line 35: we need to maintain this i'm afraid in order to get the value updated on blur

antlio (approved):
thank you so much for catching this @WDaan! 😎

hanspagel (approved):
Thanks for taking the time to fix this! Let’s get this innnn :))

PR #4709, opened 2025-02-07
feat(components): create file uploader component

hwkr:
Creates a file upload / dropzone component with two variants and several composition options. Will confirm the design with Cam soon™ 

https://github.com/user-attachments/assets/4d2df036-c7bd-40df-8c4c-b9d1ec39ff53

**Checklist**

### DIFF
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -55,6 +55,7 @@
     "@vueuse/core": "^10.10.0",
     "cva": "1.0.0-beta.2",
     "nanoid": "^5.0.7",
+    "pretty-bytes": "^6.1.1",
     "radix-vue": "^1.9.3",
     "tailwind-merge": "^2.5.5",
     "vue": "^3.5.12"
--- a/.changeset/eight-jars-dance.md
+++ b/.changeset/eight-jars-dance.md
@@ -1,0 +1,5 @@
+---
+'@scalar/components': patch
+---
+
+feat(components): create file uploader component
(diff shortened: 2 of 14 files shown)


packages/components/package.json
```diff
@@ -55,6 +55,7 @@
     "@vueuse/core": "^10.10.0",
     "cva": "1.0.0-beta.2",
     "nanoid": "^5.0.7",
+    "pretty-bytes": "^6.1.1",
```
hwkr, line 58: We already use this in `api-client` 🫡 Just brought it over here

hanspagel (approved):
incredible work

antlio:
great work! only few ux feedback:
- `upload` wording as it mades me feel i first that this would open again the file  uploader VS submitting it
- would submit button on the right make sense for consistency and natural forward pattern?
- wdyt about instead of `clear` being able to browse another file?
- won't the change of height (from empty state to file filed) cause ui issue?

image

cameronrohani (changes requested):
great work! I have some quick feedback on the design I'll dm to you asap :D

hwkr:
> great work! only few ux feedback:
> 
> * `upload` wording as it mades me feel i first that this would open again the file  uploader VS submitting it
> * would submit button on the right make sense for consistency and natural forward pattern?
> * wdyt about instead of `clear` being able to browse another file?
> * won't the change of height (from empty state to file filed) cause ui issue?
> 

Yeah these might be issues depending on the location / implementation. The file list is actually a separate component that doesn't have to be used if it doesn't fit. Most of our use cases automatically upload anyway right after file selection so I don't know if it would be used much, I mostly included it for storybook (it could definitely look nicer though 🤔)

cameronrohani (approved):

PR #4716, opened 2025-02-07
refactor(api-client): use more named routes

hanspagel:
**Problem**
Currently, when I stroll around the code base, string routes are everywhere. It’s just a matter of time, until we break a few routes.

**Solution**
With this PR we’re switching to named routes nearly entirely. There’s just two places that I couldn’t test manually and didn’t touch.

Note: We don’t have typed routes yet, but switching to route objects brings us a bit closer.

**Checklist**

### DIFF
--- a/packages/api-client/src/components/Search/useSearch.test.ts
+++ b/packages/api-client/src/components/Search/useSearch.test.ts
@@ -114,7 +114,9 @@ describe('useSearch', () => {
             path: '/path1',
             httpVerb: 'get',
             description: '',
-            link: '/workspace/workspace1/request/request1',
+            // TODO: For some reason, this is undefined when we use router.resolve() here.
+            // link: '/workspace/workspace1/request/request1',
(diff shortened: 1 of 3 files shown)


packages/api-client/src/views/Cookies/Cookies.vue
```diff
@@ -55,6 +56,7 @@ const addCookieHandler = (cookieData: {
 }
 
 const removeCookie = (uid: string) => {
+  console.log('removeCookie', uid)
```
antlio, line 59: to be eaten

amritk (approved):
hey we can fake type safe routes by just making a function that is fully typed and returns these. will give us more type safety!

packages/api-client/src/views/Cookies/Cookies.vue
hanspagel: Oops

PR #4717, opened 2025-02-07
feat(snippetz): HTTP 1.1 plugin

hanspagel:
**Problem**
Currently, we’re still using a copy of the httpsnippet-lite plugin for HTTP 1.1 examples. And this doesn’t cover many uses cases.

See https://github.com/pingdotgg/uploadthing/issues/983

**Solution**
With this PR we’re replacing it with our own plugin, to cover more edge cases (special characters) and features (file uploads, binary data, invalid URLs …).

Screenshot 2025-02-07 at 17 22 00

**Checklist**


marclave (approved):
😮‍💨 😮‍💨 😮‍💨 😮‍💨

PR #4726, opened 2025-02-10
refactor(api-reference): render example responses from store

hanspagel:
Splitting #4281 in smaller PRs:

**Problem**
Currently, we still use the deprecated transformed operation to render the example responses.

**Solution**
With this PR, we’re using the store to render example responses and add tests! Isn’t that cool?

For the example responses in the card on the right side only.

**Checklist**

### DIFF
--- a/.changeset/real-pots-shave.md
+++ b/.changeset/real-pots-shave.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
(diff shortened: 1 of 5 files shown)


amritk (approved):
Sweet! code looks good, did not test.

antlio (approved):
small round of tests on both layout - seems neat!

PR #4737, opened 2025-02-10
fix(api-client): routing broken on environment and servers page

hanspagel:
**Problem**
With #4716 I broke the routing on the servers and environment pages. We’ve used `:collectionId` before, but I refactored the code to use `PathId.Collection` (which is `collection` not `collectionId`), so we’ve had to update more places then I did in the initial PR.

**Solution**
With this PR you can click on environments and servers and see the content change again.

**Checklist**


antlio (approved):
let's go!

PR #4738, opened 2025-02-10
fix(api-client): getting started sidebar state

antlio:
**Problem**
currently the getting started state is not optimally responsive + get displayed when a request exists out of a drafts collection.

**Solution**
this pr updates responsiveness + only extracts a function to only display the state when a single request from the draft collection is added.

| before | after |
| -------|------|
| image | image |
| state is overflowing the requests | requests height according to state presence + scroll |

**Checklist**

### DIFF
--- a/packages/api-client/src/views/Request/RequestSidebar.vue
+++ b/packages/api-client/src/views/Request/RequestSidebar.vue
@@ -279,7 +279,7 @@ const toggleSearch = () => {
           @keydown.up.stop="navigateSearchResults('up')" />
       </div>
       <div
-        class="gap-1/2 flex flex-1 flex-col overflow-visible px-3 pb-3 pt-0"
+        class="gap-1/2 flex flex-1 flex-col overflow-visible overflow-y-auto px-3 pb-3 pt-0 h-[calc(100%-273.5px)]"
         :class="{
           'pb-14': layout !== 'modal',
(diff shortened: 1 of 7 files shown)


packages/api-client/src/views/Request/RequestSidebar.vue
```diff
@@ -218,6 +218,26 @@ const toggleSearch = () => {
...
+  const isDraftsRequest = draftCollection?.requests.includes(
+    activeWorkspaceRequests.value[0] ?? '',
+  )
+
+  if (!isDraftsRequest) return false
+  const isRenamed =
+    requests[draftCollection?.requests[0] ?? '']?.summary !== 'My First Request'
+
+  return hasSingleRequest && isDraftsRequest && !isRenamed
+})
```
hanspagel, line 240: We could make this a separate file and tests for the cases. Might be annoying to mock up the state, maybe we use the opportunity to add test helpers or something.

amritk: 💯

amritk: all logic should be in a helper lib, tested, then consumed in the computed

hanspagel (approved):
cool! thanks for adding the tests

PR #4740, opened 2025-02-11
feat(api-client): stricter tsconfig

amritk:
**Problem**
Our tsconfig was a bit lose, this tightens it up

**Solution**
Now there's a specific difference between an optional parameter and an undefined parameter, I just sprinkled ` | undefined` throughout so as not to break anything but we should start getting better with our types. On thing we can do is if we have an optional property, we can set it to `?: Type | null` and default to null so we explicitly allow null for an empty property instead of undefined.

Tldr: I just changed the values to make them work without changing the code and causing regressions. This is not best practice. Setting default values for optional parameters would be best practice 

**Checklist**


hwkr (approved):
Looks good to me! Played around with the types a bit and this seems to be the best compromise for sure 🙏

packages/components/src/components/ScalarListbox/ScalarListbox.vue
```diff
@@ -16,13 +16,13 @@ import ScalarListboxOption from './ScalarListboxItem.vue'
 import type { Option } from './types'
 
 type SingleSelectListboxProps = {
```
hwkr, line 18: I should maybe make this two components like I did for combobox, it would prevent this props issue

amritk: We can fix the typescript with some overloading I think

amritk:
Dug into it again and the extra verbosity is worth the type safety. The issue was vue doesn't quite handle it the same way as typescript does but it will get there.

PR #4742, opened 2025-02-11
chore: biome

hanspagel:
**Problem**

Currently, we’re using ESLint for linting and Prettier for formatting, but they are slow. Also we don’t run it in CI, for some packages it’s broken and we have way too many errors. The whole linting setup is borked.

**Solution**

With this PR we’re introducing Biome to the code base, which is faster (for the languages it supports).

* All rules are set to `'warn'`, everything is passing that way.
* The Git hooks are disabled.

That’s how we avoid merge conflicts with the open PRs. What we get, though. When add a file and use the recommended VS Code extensions, the files that we touch will already be formatted and we get all the warnings in the editor already.

I’ll ff with PRs to update the formatting of packages that don’t have any open PRs.

**Checklist**


biome.json
```diff
@@ -0,0 +1,179 @@
+{
+  "$schema": "https://biomejs.dev/schemas/1.9.4/schema.json",
```
amritk, line 2:
```suggestion
  "$schema": "./node_modules/@biomejs/biome/configuration_schema.json",
```

amritk: use the installed schema

biome.json
```diff
@@ -0,0 +1,179 @@
...
+      "venv",
+      ".venv",
+      "playwright-report",
+      ".vite-ssg-temp",
+      "pnpm-lock.yaml",
+      "fixtures"
+    ],
+    "enabled": true,
+    "lineEnding": "lf",
+    "indentWidth": 2,
```
amritk, line 38:
```suggestion
    "indentWidth": 2,
    "lineWidth": 120,
```
not sure what the default is

amritk (changes requested):
nice work! can we add a few more rules?
```json
noExcessiveCognitiveComplexity
noUndeclaredDependencies
noBarrelFile
noReExportAll
```

marclave (approved):

hanspagel:
@amritk just used the existing setup for now, we can add more rules soon

package.json
```diff
@@ -7,25 +7,24 @@
...
     "@playwright/test": "^1.49.1",
-    "@trivago/prettier-plugin-sort-imports": "^4.3.0",
     "@types/node": "^20.17.10",
-    "@typescript-eslint/eslint-plugin": "^6.21.0",
-    "@typescript-eslint/parser": "^6.21.0",
     "@vitest/coverage-v8": "^1.6.0",
-    "@vue/eslint-config-typescript": "^12.0.0",
-    "eslint": "^8.57.0",
+    "@vue/eslint-config-typescript": "^14.1.3",
+    "biome": "link:@biomejs/cli-linux-x64/biome",
```
tmastrom, line 18: https://github.com/biomejs/biome/issues/5109 Looks like this issue could be from a broken lockfile

packages/snippetz/playground/package.json
```diff
@@ -8,7 +8,7 @@
     "build": "scalar-build-vite",
     "types:check": "scalar-types-check-vue",
     "types:build": "scalar-types-build-vue",
-    "format": "scalar-format-js",
+    "format": "scalar-lint-fix",
```
tmastrom, line 11: Does it make sense to rename this to `lint:fix` ?

tmastrom (approved):
working for me 👍

PR #4743, opened 2025-02-11
feat(api-client): collection page 🎛️

hanspagel:
**Problem**

Currently, we have a few pages to edit entities that are actually attached to collections (environments, servers). The navigation between those pages is confusing.

**Solution**

With this PR, we’re introducing a collection page, where you can find related entities. Comes with the following features:

* Update title, description and icon
* Manage servers
* Toggle watch mode
* Delete collection

More? Coming soon. :)

| preview |
| -------|
| image |

**Checklist**


antlio:
added few style update - need to give a look at the highlighted issue

marclave:
reviewing now, quick look @hanspagel looks like the topnav broke :) 

image

hwkr (commented):
Gave it a test! Some UI / UX feedback:

- The folder toggle seems to have lost it's `cursor: pointer;` and isn't keyboard focusable. I'm guessing it might need to be wrapped in a `<button>`?
- When you click on a collection it doesn't seem to automatically open the collection in the sidebar, it would make sense to me that it would
- A new collection with no markdown description is automatically in preview mode and clicking the box doesn't allow you to edit it. We should probably create an empty state without the toggle so you can just click and start editing.

All of these are probably nice to have and could fast follow if that's better 👍

packages/api-client/src/layouts/App/create-api-client-app.test.ts
```diff
@@ -7,12 +7,14 @@ describe('createApiClientApp', () => {
     const element = document.createElement('div')
     expect(element).not.toBeNull()
 
-    expect(element.innerHTML).not.toContain('request')
+    console.log(element.innerHTML)
+
```
amritk, line 11:
```suggestion
```

amritk (approved):
Could change a few minor things but nothing blocking, will let @marclave merge it after final test

marclave (approved):
@antlio couple things, but can be fast followed

image
servers doesnt need `...` can just be a delete icon, we dont need 2 clicks

servers button is massive haha lets make it smaller and follow @cameronrohani 's design
image

cameronrohani:
🫡 I'll follow up with u in dm's anthony great work on the PR 🥰!

> @antlio couple things, but can be fast followed
> 
> image servers doesnt need `...` can just be a delete icon, we dont need 2 clicks
> servers button is massive haha lets make it smaller and follow @cameronrohani 's design image

antlio:
> servers doesnt need `...` can just be a delete icon, we dont need 2 clicks

it was in prevision of adding other items, but let's have the action from the element directly for now

> servers button is massive haha

added for the sake of consistency with the overview action state but can be updated

PR #4744, opened 2025-02-11
fix(api-client): empty state hotkey

antlio:
**Problem**
currently the empty state hotkeys are not using consistent sizes.

**Solution**
this pr updates the empty state hotkeys style.

| before | after |
| -------|------|
| image | image |

**Checklist**


amritk (approved):
🔥

PR #4745, opened 2025-02-11
refactor(api-reference): use store for search

hanspagel:
WIP

**Problem**
Currently, we’re still not using the store for the API reference search.

**Solution**
With this PR we’re switching to the store. :)

**Checklist**


hanspagel:
Damn, this is great, but only half done. Was more complex than I thought. And we need an empty PR queue to apply the new formatting. Need to pick this up in a new PR.

PR #4753, opened 2025-02-12
feat(components): update lock icons to align with other icons

hwkr:
As far as I can tell we don't actually use this icon anywhere, updating it for some internal work.

## Before

Google Chrome-2025-02-12-20-35-55@2x
Google Chrome-2025-02-12-20-35-49@2x

## After

Google Chrome-2025-02-12-20-34-52@2x
Google Chrome-2025-02-12-20-34-43@2x

**Checklist**


antlio (commented):
nice! just FYI there was one outlined [in figma](https://www.figma.com/design/v6mJt1mK6UEh9nWJdzj2Fa/API-Client?node-id=3897-251&t=ZFRbKVxKGJ3RaPRn-1) - proportion looks good tho!

just wondering about the open one and how it'll looks since it's not using the same off space

hwkr:
> nice! just FYI there was one outlined [in figma](https://www.figma.com/design/v6mJt1mK6UEh9nWJdzj2Fa/API-Client?node-id=3897-251&t=ZFRbKVxKGJ3RaPRn-1) - proportion looks good tho!
> 
> just wondering about the open one and how it'll looks since it's not using the same off space - do you have an example?

Oh nice! Yeah having them align is good, I updated to use the Figma icon. :)

antlio (approved):
awesome! i'm wondering if we should remove the dot from one or another to increase the state difference wdyt?

PR #4754, opened 2025-02-12
fix(api-reference): example request footer

antlio:
**Problem**
currently on `hideTestRequestButton` the example request still displays a footer which in some cases act as an extra padding [as seen on vercel](https://vercel.com/docs/rest-api/endpoints#tag/access-groups/create-an-access-group-project).

**Solution**
this pr updates the footer display logic to ensure that it gets used only in multiple example or test request button presence.

| before | after |
| -------|------|
| image | image |
| footer remains even without multiple examples | footer is not displayed | 

**Checklist**

### DIFF
--- a/.changeset/breezy-tips-hide.md
+++ b/.changeset/breezy-tips-hide.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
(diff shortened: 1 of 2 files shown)


hanspagel (approved):
quick fix, nice! :)

PR #4755, opened 2025-02-12
fix(api-client): modal address bar

antlio:
**Problem**
currently address bar in modal is not using the right font style.

**Solution**
this pr updates code input style according to layout.

| before | after |
| -------|------|
| image | image |
| font family and size is off | consistent font family and size | 

**Checklist**


amritk (approved):
more for cam, but sure why not

cameronrohani (approved):

PR #4759, opened 2025-02-13
feat(proxy-server): always return wildcard (*) for `Access-Control-Allow-Origin`

hanspagel:
**Problem**
There’s an issue that should only affect us: When hitting the proxy with the same request from various origins (e.g. http://localhost:5137 and http://localhost:5065), there’s an error in the console:

Screenshot 2025-02-13 at 10 41 43

Currently, the proxy adds the `Origin` header as the `Access-Control-Allow-Origin`, so in the example above it would for example be explicitly `http://localhost:5137`. And when we do a request from a different origin, the browser blocks it, because - according to the explicit header - the request won’t go through.

I’ve added this behaviour a while ago, because an explicit value (instead of `*` as a wildcard) was required to make cookies work in certain scenarios (cross subdomain requests).

BUT, we’ve built a custom solution to make the cookies work, so that’s not needed anymore. 

**Solution**

With this PR, the `Access-Control-Allow-Origin` is set to `*` - always. That’s what basically every other CORS middleware does, too, and that’s what we did until a few months ago.

**Checklist**

No changeset needed for the proxy, it’s not hooked up to our release process.

### DIFF
--- a/examples/proxy-server/main.go
+++ b/examples/proxy-server/main.go
@@ -194,14 +194,8 @@ func (ps *ProxyServer) executeProxyRequest(w http.ResponseWriter, r *http.Reques
 	}
 
 	// Add CORS headers here, after the response headers are copied
-	allowOrigin := "*"
-
-	if r.Header.Get("Origin") != "" {
-		allowOrigin = r.Header.Get("Origin")
-	}
-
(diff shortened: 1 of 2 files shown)


amritk (approved):
Makes sense as it is a proxy

marclave (approved):
👁️ 👄 👁️

PR #4760, opened 2025-02-13
feat(express-api-reference): render the content only once

hanspagel:
**Problem**

Currently, we’re rendering embedded content twice in the Express integration. First as part of the configuration object and then as the content of the script tag.

This makes the server response roughly twice as big as needed.

**Solution**

With this PR we’re rendering it only once! I’ve used the opportunity to extract a few of the functions into a lib.

As a next step, I’d like to move the lib to a global place and use the same lib in all JS-based integrations. 👀  

**Checklist**

### DIFF
--- a/integrations/express/src/lib/static-rendering.ts
+++ b/integrations/express/src/lib/static-rendering.ts
@@ -54,7 +54,11 @@ export function getConfiguration(givenConfiguration: ApiReferenceOptions) {
     ...givenConfiguration,
   }
 
-  delete configuration.spec
+  if (!configuration.spec?.url) {
+    delete configuration.spec
+  } else if (configuration.spec?.content) {
+    delete configuration.spec?.content
+  }
 
   return JSON.stringify(configuration).split('"').join('&quot;')
 }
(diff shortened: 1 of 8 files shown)


integrations/express/src/lib/static-rendering.ts
```diff
@@ -0,0 +1,84 @@
...
+
+/**
+ * The script tags to load the @scalar/api-reference package from the CDN.
+ */
+export function getScriptTags(configuration: ApiReferenceOptions) {
+  return `
+      <script
+        id="api-reference"
+        type="application/json"
+        data-configuration="${getConfiguration(configuration)}">${getScriptTagContent(configuration)}</script>
```
amritk, line 43: So we have more and more config options that cannot be passed through this way, should we instead push users towards passing config through JS instead of html?

amritk (approved):
My question is unrelated to this PR but I think it is important

integrations/express/src/lib/static-rendering.ts
hanspagel: That said, it’s complex. :)) we’ll chat, got some thoughts

amritk: we do have the esm api we use in a few integrations, maybe we can swindle that into the standalone. That would actually be nice because then we won't need to update the integrations in partner repos!

marclave (approved):

PR #4766, opened 2025-02-13
fix(api-client): sidebar icon state

antlio:
**Problem**
currently sidebar icon state are not being styled properly on hover.

**Solution**
this pr updates sidebar icon state style.

| before | after |
| -------|------|
| image | image |

**Checklist**


hanspagel (approved):
better! let’s get this in

PR #4767, opened 2025-02-13
fix(api-client): enum long value

antlio:
**Problem**
currently enum with long value breaks the layout.

**Solution**
this pr sets overflow on enum value.

| before | after |
| -------|------|
| image | image |

**Checklist**


hanspagel (approved):
yess

PR #4769, opened 2025-02-13
fix(api-client): ensure we preserve lowercase authentication headers

### LINKED ISSUE #4763: Latest version of scalar 1.25.118 breaks authentication
### What happens?

Authentication header is no longer passed/sent to server in [1.25.118](https://www.npmjs.com/package/@scalar/api-reference/v/1.25.118) version even if its specified in auth config. This works on [1.25.116](https://www.npmjs.com/package/@scalar/api-reference/v/1.25.116).

When authentication is specified, send auth headers as well

### OpenAPI Document

_No response_

amritk:
**Problem**
If you have a lowercase only auth header it will get deleted

**Solution**
Instead of adding and deleting we just lowercase them.

fixes #4763

**Checklist**

### DIFF
--- a/packages/api-client/src/libs/send-request/create-request-operation.ts
+++ b/packages/api-client/src/libs/send-request/create-request-operation.ts
@@ -114,13 +114,15 @@ export const createRequestOperation = ({
     const security = buildRequestSecurity(selectedSecuritySchemes, env)
 
     // For securityheaders, we lowercase them so they can be uppercased later (in normalizeHeaders)
-    Object.keys(security.headers).forEach((key) => {
-      security.headers[key.toLowerCase()] = security.headers[key] ?? ''
-      delete security.headers[key]
-    })
+    const normalizedSecurityHeaders = Object.entries(security.headers).reduce<
+      Record<string, string>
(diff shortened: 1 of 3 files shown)


packages/api-client/src/libs/send-request/create-request-operation.ts
```diff
@@ -114,13 +114,15 @@ export const createRequestOperation = ({
     const security = buildRequestSecurity(selectedSecuritySchemes, env)
 
     // For securityheaders, we lowercase them so they can be uppercased later (in normalizeHeaders)
-    Object.keys(security.headers).forEach((key) => {
-      security.headers[key.toLowerCase()] = security.headers[key] ?? ''
-      delete security.headers[key]
-    })
+    const normalizedSecurityHeaders = Object.entries(security.headers).reduce<
```
amritk, line 117: @marclave sometimes reduce is the way

marclave: mfw i see a reduce

👁️ 👄 👁️

marclave (approved):
🖖

PR #4771, opened 2025-02-14
feat(api-client): remove reliance on `activeEntities` injection and switch to prop drilling

amritk:
**Problem**
We inject active entities, but when we want to consume these components elsewhere with different values for these entities it doesnt work.

**Solution**
With this PR we prop drill all the active entities around 🥵. not the best solution but it works and is easy to understand.

One alternate strategy would be to allow injection of any entity at any level which would clean up the code alot. The downside is you would never know which entities are currently inject + plus it would be harder to test.

This needs a fine tooth review! + testing!

I put this in draft so we can get this in first for extra protection

https://github.com/scalar/scalar/issues/4773

**Checklist**

### DIFF
--- a/.changeset/violet-garlics-rhyme.md
+++ b/.changeset/violet-garlics-rhyme.md
@@ -1,0 +1,6 @@
+---
+'@scalar/api-reference': patch
+'@scalar/api-client': patch
+---
+
+chore: remove active entities for prop drilling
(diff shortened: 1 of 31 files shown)


hanspagel (commented):
looks good, but only did a quick review/test

marclave (approved):
🫡

PR #4775, opened 2025-02-14
feat(components): add sidebar nesting indicator lines

hwkr:
This PR adds theme-able nesting lines with a customizable offset / spacing.

https://github.com/user-attachments/assets/3e33c03b-6f26-4ed1-b1ca-cfdc504f154e

**Checklist**

### DIFF
--- a/packages/components/src/components/ScalarSidebar/ScalarSidebarGroup.vue
+++ b/packages/components/src/components/ScalarSidebar/ScalarSidebarGroup.vue
@@ -46,7 +46,7 @@ defineOptions({ inheritAttrs: false })
 const { cx } = useBindCx()
 </script>
 <template>
-  <li class="contents">
+  <li class="group/item contents">
     <slot
       :level="level"
       name="button"
@@ -73,7 +73,7 @@ const { cx } = useBindCx()
     <component
       :is="is"
       v-if="open"
-      v-bind="cx('flex flex-col relative')">
+      v-bind="cx('flex flex-col gap-px')">
       <slot
         name="items"
         :open="open" />
--- a/packages/components/src/components/ScalarSidebar/ScalarSidebarGroup.vue
+++ b/packages/components/src/components/ScalarSidebar/ScalarSidebarGroup.vue
@@ -47,6 +47,7 @@ const { cx } = useBindCx()
 <template>
   <li class="contents">
     <slot
+      :level="level"
       name="button"
       :open="open">
       <ScalarSidebarButton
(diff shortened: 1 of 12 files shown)


antlio:
love the active state! should we keep the gap as currently set in client? (or move off of it?) cc @cameronrohani @hwkr 

| before | after |
| -------|------|
| image | image |

cameronrohani:
@hwkr @antlio let's keep gap as long as the lines can still connect :p

cameronrohani:
> @hwkr @antlio let's keep gap as long as the lines can still connect :p

good catch anthony 🧠

hwkr:
Okay it was a little complicated but I managed to add a 1px gap between items and get the indents to look nice.

(This has become a very fancy sidebar)

https://github.com/user-attachments/assets/591fd5e2-ec5e-42ef-867f-8b88bf20d18d

packages/components/src/components/ScalarSidebar/ScalarSidebarGroup.vue
```diff
@@ -45,25 +46,34 @@ defineOptions({ inheritAttrs: false })
...
+            :open="open" />
         </template>
         <slot :open="open" />
       </ScalarSidebarButton>
     </slot>
     <component
       :is="is"
       v-if="open"
-      v-bind="cx('flex flex-col')">
+      v-bind="cx('flex flex-col gap-px')">
```
antlio, line 76: should we keep this .5? cc @cameronrohani 

```suggestion
      v-bind="cx('flex flex-col gap-1/2')">
```

hwkr:
> This is cool, we using it on the otherside? Should we use this in the client/ref as well

Yep! The plan is to use it everywhere

packages/components/src/components/ScalarSidebar/ScalarSidebarGroup.vue
hwkr: We could switch to that if we want, I actually messed up in the client config though and it shouldn't be `1/2` (`1/2` is supposed to be `50%` in tailwind spacing). But we could create `.5px` or `1/2px` as a value.

cameronrohani (approved):

packages/components/src/components/ScalarSidebar/ScalarSidebarGroup.vue
hwkr: Going to merge for now and we can update later if we want 🫡

amritk: merge it?

amritk:
Mergin!

PR #4778, opened 2025-02-14
chore: remove “spec” wording

hanspagel:
**Problem**
Currently, we’re using spec to refer to OpenAPI documents a lot. But the OpenAPI "spec” is the actual standard, the set of rules that makes the OpenAPI standard.

**Solution**
With this PR we’re reducing it, and use the word “document” where possible.

**Checklist**


marclave (approved):
proper 😎

PR #4793, opened 2025-02-17
feat(build-tooling): only lint when vue files are there

hanspagel:
**Problem**
The new lint script uses ESLint to lint Vue files, but it tries it even in packages that don’t have Vue files.

**Solution**
Let’s just check whether we have Vue files before running ESLint.

**Checklist**


antlio (approved):
oui!

PR #4795, opened 2025-02-17
feat(api-reference): use Zod to parse the configuration 

hanspagel:
**Problem**

Currently, the `ReferenceConfiguration` is passed to the `<ApiReference />` and `<ApiReferenceLayout />` components and we do some things with it (setting defaults, showing deprecation warnings, migrating theme variables, migrating deprecated attributes) in various places.

This makes me nervous.

And we have few things that we need to/want to change in the configuration, but I’m afraid to do this without migrations and tests.

**Solution**

With this PR we’re establishing a single pipeline that the configuration goes through, using Zod.

* Zod Schema in the `@scalar/types` package
* Getting the TypeScript type from the Zod schema, instead of raw TS
* Tests for the configuration parsing
* Attached the migration of proxy -> proxyUrl and legacy CSS variables to the Zod schema

I also added a few things 
- type check kept dieing so I removed turbo, now it actually runs much faster (using all of cpu)
- moved some of the integrations to tsconfig.node.json since I needed the stricter types

**Checklist**


packages/types/src/packages/api-reference-configuration.ts
```diff
@@ -0,0 +1,265 @@
...
+  theme: z.union([ThemeIdEnum, z.undefined()]).optional(),
+  /** The layout to use for the references */
+  layout: z.enum(['modern', 'classic']).optional(),
+  /** The Swagger/OpenAPI spec to render */
+  spec: SpecConfigurationSchema.optional(),
+  /**
+   * URL to a request proxy for the API client
+   * @deprecated Use proxyUrl instead
+   */
+  proxy: z.string().optional(),
```
amritk, line 107:
```suggestion
  proxy: z.string().optional(),
```
I think we can use a transform to convert proxy => proxyUrl

https://zod.dev/?id=transform

packages/types/src/packages/api-reference-configuration.ts
```diff
@@ -0,0 +1,265 @@
...
+
+/** Configuration for path-based routing */
+const PathRoutingSchema = z.object({
+  /** Base path for the API reference */
+  basePath: z.string(),
+})
+
+export const ApiReferenceConfigurationSchema = z.object({
+  /** A string to use one of the color presets */
+  theme: z.union([ThemeIdEnum, z.undefined()]).optional(),
```
amritk, line 98:
```suggestion
  theme: ThemeIdEnum.nullable().optional(),
```
I think optional will allow the paramter to be optional (undefined). If we want to explicitly allow an empty value, lets use null instead.

packages/types/src/packages/api-reference-configuration.ts
```diff
@@ -0,0 +1,265 @@
...
+  /** The layout to use for the references */
+  layout: z.enum(['modern', 'classic']).optional(),
```
amritk, line 100:
```suggestion
  layout: z.enum(['modern', 'classic']).optional().default('modern').catch('modern'),
```
lets set a default wherever we can

packages/types/src/packages/api-reference-configuration.ts
```diff
@@ -0,0 +1,265 @@
...
+  spec: SpecConfigurationSchema.optional(),
+  /**
+   * URL to a request proxy for the API client
+   * @deprecated Use proxyUrl instead
+   */
+  proxy: z.string().optional(),
+  /** URL to a request proxy for the API client */
+  proxyUrl: z.string().optional(),
+  /** Whether the spec input should show */
+  isEditable: z.boolean().optional(),
```
amritk, line 111:
```suggestion
  /** 
    Whether the spec input should show 
    @default false
  */
  isEditable: z.boolean().optional().default(false).catch(false),
```

packages/types/src/packages/api-reference-configuration.ts
```diff
@@ -0,0 +1,265 @@
...
+  /** Whether to show the sidebar */
+  showSidebar: z.boolean().optional(),
```
amritk, line 113:
```suggestion
  /** 
    Whether to show the sidebar 
    @default true
  */
  showSidebar: z.boolean().optional().default(true).catch(true),
```
not sure what we do right now but it feels like this should be defaulted to true?

packages/types/src/packages/api-reference-configuration.ts
```diff
@@ -0,0 +1,265 @@
...
+  /**
+   * Whether to show models in the sidebar, search, and content.
+   * @default false
+   */
+  hideModels: z.boolean().optional(),
```
amritk, line 118:
```suggestion
  hideModels: z.boolean().optional().default(false).catch(false),
```

packages/types/src/packages/api-reference-configuration.ts
```diff
@@ -0,0 +1,265 @@
...
+  /**
+   * Whether to show the "Download OpenAPI Document" button
+   * @default false
+   */
+  hideDownloadButton: z.boolean().optional(),
```
amritk, line 123:
```suggestion
  hideDownloadButton: z.boolean().optional().default(false).catch(false),
```

packages/types/src/packages/api-reference-configuration.ts
```diff
@@ -0,0 +1,265 @@
...
+  /**
+   * Whether to show the "Test Request" button
+   * @default false
+   */
+  hideTestRequestButton: z.boolean().optional(),
```
amritk, line 128:
```suggestion
  hideTestRequestButton: z.boolean().optional().default(false).catch(false),
```

packages/types/src/packages/api-reference-configuration.ts
```diff
@@ -0,0 +1,265 @@
...
+  /**
+   * Whether to show the sidebar search bar
+   * @default false
+   */
+  hideSearch: z.boolean().optional(),
```
amritk, line 133:
```suggestion
  hideSearch: z.boolean().optional().default(false).catch(false),
```

amritk (changes requested):
pretty much just add defaults

packages/types/src/packages/api-reference-configuration.ts
```diff
@@ -0,0 +1,265 @@
...
+  operationsSorter: z
+    .union([z.literal('alpha'), z.literal('method'), z.function().args(z.any(), z.any()).returns(z.number())])
+    .optional(),
+  /** Integration type identifier */
+  _integration: IntegrationEnum.optional(),
+  /** Whether to hide the client button */
+  hideClientButton: z.boolean().optional(),
+})
+
+export type ApiReferenceConfiguration = z.infer<typeof ApiReferenceConfigurationSchema>
```
amritk, line 265: Also we should have two types here.
z.input is for the user, it's before zod validation.

z.infer is for us. It's after zod validation

packages/themes/src/index.ts
```diff
@@ -14,7 +14,6 @@ import purpleTheme from './presets/purple.css?inline'
 import saturnTheme from './presets/saturn.css?inline'
 import solarizedTheme from './presets/solarized.css?inline'
 
-export { migrateThemeVariables } from './utilities/legacy'
```
hwkr, line 17: I think this is being used in `scalar-org`

packages/types/src/api-reference/api-reference-configuration.ts
```diff
@@ -0,0 +1,297 @@
...
+    'react',
+    'rust',
+    'vue',
+  ])
+  .nullable()
+
+/** Configuration for the OpenAPI/Swagger specification */
+const SpecConfigurationSchema = z.object({
+  /** URL to an OpenAPI/Swagger document */
+  url: z.string().url().optional(),
```
amritk, line 83:
```suggestion
  url: z.string().optional(),
```
Probably stay away from the url validation as it will fail

packages/types/src/api-reference/api-reference-configuration.ts
```diff
@@ -0,0 +1,297 @@
...
+/** Configuration for path-based routing */
+const PathRoutingSchema = z.object({
+  /** Base path for the API reference */
+  basePath: z.string(),
+})
+
+export const ApiReferenceConfigurationSchema = z
+  .object({
+    /** A string to use one of the color presets */
+    theme: ThemeIdEnum.optional().catch('default'),
```
amritk, line 101:
```suggestion
    theme: ThemeIdEnum.optional().default('default').catch('default'),
```

packages/types/src/api-reference/api-reference-configuration.ts
```diff
@@ -0,0 +1,297 @@
...
+     * Path to a favicon image
+     * @default undefined
+     * @example '/favicon.svg'
+     */
+    favicon: z.string().optional(),
+    /**
+     * List of httpsnippet clients to hide from the clients menu
+     * By default hides Unirest, pass `[]` to show all clients
+     */
+    hiddenClients: z.union([z.record(z.union([z.boolean(), z.array(z.string())])), z.array(z.string())]).optional(),
```
amritk, line 166: Should we transform to the string array?

packages/types/src/api-reference/api-reference-configuration.ts
```diff
@@ -0,0 +1,297 @@
...
+        targetKey: z.custom<TargetId>(),
+        clientKey: z.string(),
+      })
+      .optional(),
+    /** Custom CSS to be added to the page */
+    customCss: z.string().optional(),
+    /** onSpecUpdate is fired on spec/swagger content change */
+    onSpecUpdate: z.function().returns(z.void()).optional(),
+    /** Prefill authentication */
+    authentication: z.any().optional(), // Using any for Partial<AuthenticationState>
```
amritk, line 179: We should add a typed authentication schema

packages/types/src/api-reference/api-reference-configuration.ts
```diff
@@ -0,0 +1,297 @@
...
+      .args(
+        z.object({
+          name: z.string(),
+          method: z.string().optional(),
+        }),
+      )
+      .returns(z.string())
+      .optional(),
+    /** Callback fired when the reference is fully loaded */
+    onLoaded: z.union([z.function().returns(z.void()), z.undefined()]).optional(),
```
amritk, line 247: Do we need to allow undefined?

packages/types/src/api-reference/api-reference-configuration.ts
```diff
@@ -0,0 +1,297 @@
...
+    /** Base URL for the API server */
+    baseServerURL: z.string().optional(),
+    /** List of OpenAPI server objects */
+    servers: z.array(z.any()).optional(), // Using any for OpenAPIV3_1.ServerObject
+    /** Whether to include default fonts */
+    withDefaultFonts: z.boolean().optional(),
```
amritk, line 253: Is this defaulted to true?

packages/types/src/api-reference/api-reference-configuration.ts
```diff
@@ -0,0 +1,297 @@
...
+    const configuration = originalConfiguration
+
+    // Migrate legacy theme variables
+    if (configuration.customCss) {
+      configuration.customCss = migrateThemeVariables(configuration.customCss)
+    }
+
+    // Migrate proxy URL
+    if (configuration.proxy) {
+      migrateProxyUrl(configuration)
```
amritk, line 283: This should return the value, never mutate an object in a function!

amritk (approved):
Sweet PR! Let's zod all the things! Looks good for the most part! Mostly nit picks. But some important schema changes.  I will approve just so you don't need to re review. Did not test

packages/types/src/api-reference/api-reference-configuration.ts
hanspagel: Not in this pr

packages/types/src/api-reference/api-reference-configuration.ts
hanspagel: Rookie mistake

packages/types/src/api-reference/api-reference-configuration.ts
amritk: ah doesn't matter

packages/api-client/src/store/store.ts
```diff
@@ -126,7 +127,7 @@ export const createWorkspaceStore = ({
   // Set some defaults on all workspaces
   Object.values(workspaces).forEach(({ uid }) => {
     if (proxyUrl) workspaceMutators.edit(uid, 'proxyUrl', proxyUrl)
-    if (themeId) workspaceMutators.edit(uid, 'themeId', themeId)
+    if (themeId) workspaceMutators.edit(uid, 'themeId', themeId as ThemeId)
```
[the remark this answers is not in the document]
amritk: yup we were exporting the wrong type!

integrations/express/tsconfig.json
```diff
@@ -1,5 +1,5 @@
 {
-  "extends": "../../tsconfig.json",
+  "extends": "../../tsconfig.node.json",
```
amritk, line 2: I needed the stricter types, might as well use node while we're here

packages/types/src/api-reference/api-reference-configuration.ts
amritk: actually going to redo it anyway so I'll do it as a fast follow

.github/workflows/ci.yml
```diff
@@ -164,7 +164,7 @@ jobs:
         with:
           node-version: ${{ matrix.node-version }}
       - name: Check types
-        run: pnpm turbo types:check
+        run: pnpm types:check
```
amritk, line 167: turbo type check kept dieing

amritk:
@hanspagel whats left to get this in?

amritk:
bottoms up

PR #4802, opened 2025-02-18
chore(tooling): use node slim base image

tmastrom:
**Problem**
Currently, we use a non-slim base image for the scalar base docker image. 

The image `node:20-bullseye` is built on `buildpack-deps` which includes a lot of tooling that is unnecessary for our use case. This image contains a full python installation and `gcc` toolchain for building node modules from C++. This massive package bloat leads to a large image and also opens the door to many more vulnerabilities. 

More info available [here](https://labs.iximiuz.com/tutorials/how-to-choose-nodejs-container-image) 

NOTE: we should be aware of this and if our Dockerfiles are having issues with the npm install step then this is the likely culprit. 

**Solution**
With this PR, we change the base image from `node:20-bullseye` to `node:20-slim` and save almost 1GB in image size.  These are still official, debian based node docker images with lots of standard compiled binaries.`slim` is debian bookworm (current stable release) whereas `bullseye` is "old stable" release

### Before (3.39GB)
Screenshot 2025-02-18 at 12 20 47 PM
### After (2.56GB)
Screenshot 2025-02-18 at 12 28 40 PM

**Checklist**

### DIFF
--- a/tooling/base-docker-builder/Dockerfile
+++ b/tooling/base-docker-builder/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:20-bullseye AS base
+FROM node:20-slim AS base
 
 # process init system
 RUN apt-get update && apt-get install -y --no-install-recommends dumb-init


hanspagel (approved):
nice

amritk:
how about alpine?

tmastrom:
> how about alpine?

@amritk Alpine is a bit of a footgun that I don't want to use on our base image. We could consider using alpine or distroless on our deploy images though! 

The savings of slim vs. alpine are pretty small compared to regular vs. slim. 
Screenshot 2025-02-19 at 8 22 20 AM
[source](https://labs.iximiuz.com/tutorials/how-to-choose-nodejs-container-image)

PR #4808, opened 2025-02-18
feat(fastify-api-reference): pass `logLevel` to registered routes

### LINKED ISSUE #4783: passthrough logLevel to fastify routes
### What happens?

In fastify, when logging is enabled at request level, by default it logs all requests, which is useful. However for specific routes it can be disabled by passing {logLevel:'silent'}. This is not supported for scalar routes. Please allow passthrough configuration for logging.

https://github.com/scalar/scalar/blob/11d873d0be2b3fb82c1a1a8416425be8483014db/packages/fastify-api-reference/src/fastifyApiReference.ts#L248

When setting {logLevel:'silent'} when registering scalar routes, it will passthrough to all registered routes.

### OpenAPI Document

_No response_

hanspagel:
**Problem**
Currently, we just use the default logger settings. If you want to override the setting for the logLevel for our routes, there’s no way to do this.

**Solution**
With this PR we’re passing an (optional) `logLevel` to all registered routes.

Fixes #4783

**Checklist**

### DIFF
--- a/.changeset/cyan-camels-exist.md
+++ b/.changeset/cyan-camels-exist.md
@@ -1,0 +1,5 @@
+---
+'@scalar/fastify-api-reference': patch
+---
+
+feat: pass logLevel to all registered routes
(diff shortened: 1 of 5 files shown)


amritk (approved):
Looks decent

marclave (approved):
!!

Issue #4757, opened 2025-02-12
Just show download button if response is too big to be serialized


amritk:
### What happens?

https://discord.com/channels/1135330207960678410/1334843202569506836

> Sorry delayed response again. So I can see that virtualization is happening on one endpoint that returns 7.9MB. It does take a short while but the text loads and looks like your second screenshot
However there is another endpoint that returns 39MB and after a short while the page displays an "error code: out of memory" page. 
mrhockeymonkey-7869673610 — Yesterday at 3:21 AM
Watching the process in task manager it looks like the chrome proc memory hits ~4700MB before it is killed

PR #4816, opened 2025-02-19
fix(api-client): virtualization issues causing tab to be unresponsive

DemonHa:
**Problem**
Currently, on small screens virtualization was not working since the container did not have a fixed height so it was taking up all the space. That was causing all the response body to be rendered which was making the tab unresponsive

See  #4757 

**Solution**
With this PR we set a max height on the container in order to make virtualization work on smaller screens.

**Checklist**

### DIFF
--- a/packages/api-client/src/views/Request/ResponseSection/ResponseBodyVirtual.vue
+++ b/packages/api-client/src/views/Request/ResponseSection/ResponseBodyVirtual.vue
@@ -37,7 +37,7 @@ const { mimeType, attachmentFilename, dataUrl } = useResponseBody({
       This response body is massive! Syntax highlighting won’t work here.
     </div>
     <ScalarVirtualText
-      containerClass="custom-scroll scalar-code-block border-1/2 rounded-b flex flex-1 max-h-[100vh]"
+      containerClass="custom-scroll scalar-code-block border-1/2 rounded-b flex flex-1 max-h-screen"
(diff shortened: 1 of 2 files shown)


packages/api-client/src/views/Request/ResponseSection/ResponseBodyVirtual.vue
```diff
@@ -37,7 +37,7 @@ const { mimeType, attachmentFilename, dataUrl } = useResponseBody({
       This response body is massive! Syntax highlighting won’t work here.
     </div>
     <ScalarVirtualText
-      containerClass="custom-scroll scalar-code-block border-1/2 rounded-b flex flex-1"
+      containerClass="custom-scroll scalar-code-block border-1/2 rounded-b flex flex-1 max-h-[100vh]"
```
antlio, line 40: there's a utility class that should do it - `max-h-screen` or could use `max-h-dvh` too ([see reference](https://v3.tailwindcss.com/docs/max-height))

```suggestion
      containerClass="custom-scroll scalar-code-block border-1/2 rounded-b flex flex-1 max-h-screen"
```

antlio:
great call! added one nitpick comment - can you add a changeset?

amritk (approved):
Wicked atomic PR @DemonHa !

antlio (approved):

PR #4818, opened 2025-02-19
fix(api-client): request body file upload

antlio:
**Problem**
currently the request body file cta might not seen very understandable only displaying `File` and no default empty row is added on upload.

**Solution**
this pr is a proposal to use `Upload File` as already used for binary file + fix the row management.

| before | after |
| -------|------|
| image | image |
| default row not added on upload | default row added on upload + cta updated | 

**Checklist**

### DIFF
--- a/packages/api-client/src/views/Request/RequestSection/RequestTable.vue
+++ b/packages/api-client/src/views/Request/RequestSection/RequestTable.vue
@@ -1,13 +1,14 @@
 <script setup lang="ts">
+import { ScalarButton, ScalarIcon, ScalarTooltip } from '@scalar/components'
+import type { RequestExampleParameter } from '@scalar/oas-utils/entities/spec'
+import { computed } from 'vue'
+import type { RouteLocationRaw } from 'vue-router'
+
 import CodeInput from '@/components/CodeInput/CodeInput.vue'
 import DataTable from '@/components/DataTable/DataTable.vue'
 import DataTableCell from '@/components/DataTable/DataTableCell.vue'
 import DataTableCheckbox from '@/components/DataTable/DataTableCheckbox.vue'
 import DataTableRow from '@/components/DataTable/DataTableRow.vue'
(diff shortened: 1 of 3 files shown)


packages/api-client/src/views/Request/RequestSection/RequestTable.vue
```diff
@@ -175,26 +176,26 @@ const flattenValue = (item: RequestExampleParameter) => {
...
         </template>
         <template v-else>
           <ScalarButton
-            class="bg-b-2 hover:bg-b-3 border-0 py-px text-c-2 shadow-none"
+            class="bg-b-2 hover:bg-b-3 text-c-2 border-0 py-px shadow-none"
             size="sm"
             variant="outlined"
             @click="handleFileUpload(idx)">
-            <span>File</span>
+            <span>Upload File</span>
```
hanspagel, line 198: to be precise, you just *select* a file here, the upload only happens when you send a request

```suggestion
            <span>Select File</span>
```

amritk: but you are telling it to upload a file 😉 @cameronrohani what say you

antlio: for now 😏 <br>— as we already use that wording + give the action context - i guess matter of taste

hanspagel: but the *button* doesn’t upload, a button label should describe what the button does 😶

amritk (approved):

packages/api-client/src/views/Request/RequestSection/RequestTable.vue
amritk: tomato potato I say lol. Normally I would agree with you hans but in this case, even though its technically wrong, it feels right 😉

marclave (approved):
lets merge away 🫡

PR #4822, opened 2025-02-19
feat(api-client): add brands to entity UIDs

amritk:
# Branding PR 1/2

**Problem**
In another PR, claude was tabbing a long and accidentally added an operation.uid into a requestExampleMutator. Before merging that one I want to ensure that cannot happen through type safety.

**Solution**
I stumbled across zod brand which should do exactly that! This PR just did the minimum(😅) to add brands to the UIDs, with some `as` sprinkled in just to get it working.

The follow up PR will add the branded types to the mutators just because this PR is so large already. No need to test this PR as it would be better to test [that one](https://github.com/scalar/scalar/pull/4825) then merge em back to back.

For reviewing: alot of it is just formatting as we go from prettier -> biome, also ignore the test files and it becomes much more palatable 

**Checklist**

### DIFF
--- a/packages/api-client/src/components/Form/Form.vue
+++ b/packages/api-client/src/components/Form/Form.vue
@@ -1,4 +1,7 @@
 <script setup lang="ts">
+import type { Cookie } from '@scalar/oas-utils/entities/cookie'
+import type { Path, PathValue } from '@scalar/object-utils/nested'
+
 import DataTable from '@/components/DataTable/DataTable.vue'
 import DataTableInput from '@/components/DataTable/DataTableInput.vue'
 import DataTableRow from '@/components/DataTable/DataTableRow.vue'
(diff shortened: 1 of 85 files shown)


marclave (approved):
tested and awesome 🫡 lets wait for @hanspagel to review this since such a big one!

hanspagel (approved):
OMG, feels so good

amritk:
> I just renamed request to operation, hope that’s OK.

Ya I wanted to do that but also reduce the amount of changes :shrug:

packages/api-client/src/components/Form/Form.vue
```diff
@@ -12,7 +15,10 @@ defineProps<{
     placeholder: string
   }[]
   data: Record<string, any>
-  onUpdate: (key: string, value: any) => void
+  onUpdate: <P extends Path<Cookie>>(
```
amritk, line 18: @antlio this is generally how we type the mutators

PR #4825, opened 2025-02-20
feat(api-client): accept the correct uids in the mutators

amritk:
# Branding PR 2/2 - DO NOT REBASE

**Problem**
The problem is explained in the other PR. Review both PRs but only test this one, merge the other one first!

**Solution**
- we type the mutators (edit + delete) so we cannot accidentally do `workspaceMutators.delete(collection.uid)`
- we allow `Brand | null | undefined` in the above and handle with a warning in the console so we don't need to do checks everwhere like `?? ''` or `if (entity.value) ...`

**Checklist**


hanspagel (approved):
works like a charm!!

PR #4828, opened 2025-02-20
fix: Biome removes React import

hanspagel:
**Problem**
Currently, Biome removes “unused” React imports. But (I don’t know why) they are necessary for those files to work.

**Solution**
Luckily, there’s a Biome setting to tell Biome it’s React, so it’ll keep the “unused” import.

I’ve also looked for `eslint-disable` comments and was able to remove a lot of them. And I applied the new formatting to more files.

```json
"javascript": {
  "jsxRuntime": "reactClassic"
}
```

**Checklist**

### DIFF
--- a/.changeset/fresh-rivers-leave.md
+++ b/.changeset/fresh-rivers-leave.md
@@ -1,0 +1,20 @@
+---
+'@scalar/api-reference-react': patch
+'@scalar/api-client-react': patch
+'@scalar/aspnetcore': patch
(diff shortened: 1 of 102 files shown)


hwkr (approved):
Looks good to me!

amritk (approved):
Not a big fan of the useless react import but yolo lets go

marclave (approved):

PR #4830, opened 2025-02-20
feat(api-client): schema based validation

antlio:
**Problem**
currently when using a different type or expected value in a request parameter, no visual cue is given.

**Solution**
this pr is adding in order to incentive the user regarding the expected type or value:
- lib function that return warning message according to the expected type/value
- visual alert on the given row receiving a non valid type/value
- keeps allowing request to be sent

| before | after |
| -------|------|
| image | image |
| no visual cue is used | row is highlighted along displaying a warning message | 

**Checklist**

### DIFF
--- a/packages/api-client/src/components/CodeInput/CodeInput.vue
+++ b/packages/api-client/src/components/CodeInput/CodeInput.vue
@@ -63,7 +63,7 @@ const emit = defineEmits<{
   (e: 'curl', v: string): void
 }>()
 
-const attrs = useAttrs()
+const attrs = useAttrs() as { id?: string }
 const uid = (attrs.id as string) || `id-${nanoid()}`
 
 const isFocused = ref(false)
@@ -258,7 +258,7 @@ export default {
       :id="uid"
       v-bind="$attrs"
       ref="codeMirrorRef"
-      class="font-code peer relative w-full overflow-hidden whitespace-nowrap text-xs leading-[1.44] -outline-offset-1 has-[:focus-visible]:rounded-[4px] has-[:focus-visible]:outline"
+      class="group-[.alert]:outline-orange font-code peer relative w-full overflow-hidden whitespace-nowrap text-xs leading-[1.44] -outline-offset-1 has-[:focus-visible]:rounded-[4px] has-[:focus-visible]:outline"
       :class="{
         'flow-code-input--error': error,
       }"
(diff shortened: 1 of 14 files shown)


packages/api-client/src/components/CodeInput/CodeInput.vue
```diff
@@ -257,7 +257,7 @@ export default {
       :id="uid"
       v-bind="$attrs"
       ref="codeMirrorRef"
-      class="font-code peer relative w-full overflow-hidden whitespace-nowrap text-xs leading-[1.44] -outline-offset-1 has-[:focus-visible]:rounded-[4px] has-[:focus-visible]:outline"
+      class="group-[.alert]:outline-orange font-code peer relative w-full overflow-hidden whitespace-nowrap text-xs leading-[1.44] -outline-offset-1 has-[:focus-visible]:rounded-[4px] has-[:focus-visible]:outline"
```
hwkr, line 260: Any reason to set the group like this vs a custom tailwind group?

https://tailwindcss.com/docs/hover-focus-and-other-states#differentiating-nested-groups

hwkr (approved):
Looks good! Just left one comment about tailwind group :)

packages/api-client/src/components/CodeInput/CodeInput.vue
antlio: as i relied on an existing `group` utility on the row i was just afraid it would break things to change it 😂 but will do!

amritk (approved):
Send it! Did not test but code looks good

packages/api-client/src/views/Request/libs/request.ts
```diff
@@ -10,11 +11,63 @@ import type { RequestExampleParameter } from '@scalar/oas-utils/entities/spec'
...
-      item.maximum,
-  )
+  Boolean(item.description || item.type || item.default || item.format || item.minimum || item.maximum)
+
+/**
+ * Checks if the value of a RequestExampleParameter is the expected type or format
+ * Returns an alert message if the value is not in the correct type or format, otherwise false
+ */
+export const parameterIsInvalid = (item: RequestExampleParameter) => {
+  return computed(() => {
```
amritk, line 21: why is this computed? Cant we just use it without the computed? I think this will be slower unless computed caches by input and not by instance 🤔

marclave (approved):

PR #4833, opened 2025-02-20
feat(api-client): request required field validation

antlio:
**Problem**
currently request can be sent out with empty required field and no visual cue is given.

**Solution**
this pr adds visual cue when a required field value is empty and prevents the request to be sent.

| before | after |
| -------|------|
| image | image |
| no required field visual cue nor error | required field visual cue + toast | 

**Checklist**

### DIFF
--- a/.changeset/many-balloons-smash.md
+++ b/.changeset/many-balloons-smash.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
+
+feat: adds request required empty parameter error
--- a/.changeset/mean-brooms-decide.md
+++ b/.changeset/mean-brooms-decide.md
@@ -1,0 +1,5 @@
+---
+'@scalar/components': patch
+---
+
+feat: adds alert icon
--- a/.changeset/mighty-eggs-do.md
+++ b/.changeset/mighty-eggs-do.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
+
+feat: adds request parameter value validation alert
--- a/.changeset/wise-actors-tease.md
+++ b/.changeset/wise-actors-tease.md
@@ -1,0 +1,5 @@
+---
+'@scalar/themes': patch
+---
+
+feat: updates danger variable + add background alert variable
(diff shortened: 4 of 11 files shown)


packages/api-client/src/views/Request/Request.vue
```diff
@@ -93,6 +95,26 @@ const executeRequest = async () => {
...
+        toast(`Required ${paramType} parameter ${param.key} is empty`, 'error')
+        events.requestStatus.emit('abort')
+        return true
+      }
+      return false
+    })
+  })
+
+  if (hasEmptyRequiredParameters) {
+    events.cancelRequest.on(cancelRequest)
```
amritk, line 114: wait what this? cancel request? We don't want to cancel

amritk (changes requested):
Not sure exactly what is happening there

packages/api-client/src/views/Request/Request.vue
antlio: cancels the request if a required field is empty as this would throw an error anyway?

amritk: we don't want to cancel the request, we want to let it go

antlio: we talked that we would let it go regarding type and value validation but thought we would prevent it for required one? cc @hanspagel

antlio: i'll update accordingly

amritk: nope just let em all go! use case: what if someone is testing what happens when you miss required params 😄

hwkr (commented):
Colors look good! Seems like the group `.alert` class is back

antlio:
> Colors look good! Seems like the group `.alert` class is back

you mean `group` vs `group/row`? i didn't change it, will here!

hanspagel (commented):
We shouldn’t stop a user from sending a request with invalid (for the purpose of testing an API), so I think it’s not an error, but a warning and it should then be orange (and not red). If that makes sense. :)

amritk (approved):

amritk:
Going to merge as all the comments are minor and can be changed after the fact! Just so we can format

PR #4839, opened 2025-02-20
chore: disable `vue/singleline-html-element-content-newline` ESLint rule

amritk:
**Problem**
Eslint fixes it, then prettier breaks it.

**Solution**
We unfortunately need to comment the eslint rule, its much nicer but there's no way to configure it into prettier.

Lets remove prettier?

**Checklist**


hanspagel (approved):
omg, thanks

Issue #4480, opened 2025-01-16
validate command not using  the source document named in the `scalar.config.json` file


DavidBiesack:
### What happens?

I'm using:
```bash
$ scalar --version
0.2.286
```

(installed with `npm install -g @scalar/cli`)

I created a `scalar.config.json` file and entered the file name `openapi.yaml` when prompted.
The contents of the file:
```
$ cat scalar.config.json
{
  "subdomain": "example.apidocumentation.com",
  "references": [
    {
      "name": "API Reference",
      "path": "openapi.yaml"
    }
  ],
  "guides": []
}
```

However, when trying to validate, I get an error 

```bash
$ scalar validate
No file provided.

Try `scalar init` or add the file as an argument. Read `scalar --help` for more information.

```

Passing the file name on the command line works but this is not what the tool's documentation implies with `scalar init`

```bash
$ scalar validate openapi.yaml
Matches the OpenAPI specification (OpenAPI 3.1)

File validated in 187 ms
```

Here is how I created the config file:
```bash
$ scalar init
◇  What’s the name of your project? We’ll use that to create a custom subdomain
for you.
│  Example
✔ Subdomain: example.apidocumentation.com
│
◇  Where is your OpenAPI file? (Add a path to the file)
│  openapi.yaml
✔ Configuration stored.

scalar.config.json

  {
    "subdomain": "example.apidocumentation.com",
    "references": [
      {
        "name": "API Reference",
        "path": "openapi.yaml"
      }
    ],
    "guides": []
  }

What to do next:
  scalar format [options] [file|url] to format your OpenAPI file
  scalar validate [file|url] to validate your OpenAPI file
  scalar bundle [options] [file] to bundle your OpenAPI file
  scalar serve [options] [file|url] to serve your OpenAPI file

Run scalar --help to see all available commands.
$
```

### What did you expect to happen?

```bash
$ scalar validate
````
should work when the `scalar.config.json` exists and names an OpenAPI document.

DemonHa:
ok closing this issue since this is not how that command is intended to work...

now on the new version of the cli the validate command has been movied under the document domain [Reference](https://scalar.com/tools/cli/commands#validate)

`scalar document validate [options] [file|url]`

PR #4842, opened 2025-02-21
fix(cli): load configuration references correctly

DemonHa:
**Problem**
Currently, when user does not provide a file name we don't load references correctly

See #4480 

**Solution**
With this PR, it loads references from configuration file correctly when the user does not specify a file name

**Checklist**

### DIFF
--- a/.changeset/new-carpets-bake.md
+++ b/.changeset/new-carpets-bake.md
@@ -1,0 +1,5 @@
+---
+'@scalar/cli': patch
+---
+
+fix: load references correctly from configuration file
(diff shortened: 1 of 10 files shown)


hanspagel:
Oh, can you create a detailed issue? Because it seems to work in CI: https://github.com/scalar/scalar/blob/main/.github/workflows/validate-openapi-file.yml

That said, the CLI is in a kind of transition phase, so we might not be able to merge PRs against the current state easily. Sorry, we should note this somewhere.

PR #4854, opened 2025-02-21
fix(api-client): remove header update if initial body is set to raw

marclave:
**Problem**
Currently, we remove content-type header if its added in the spec

**Solution**
With this PR we only set the content-type header if the body is selected

**Checklist**

### DIFF
--- a/.changeset/twelve-points-raise.md
+++ b/.changeset/twelve-points-raise.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
+
+fix: remove header update if initial body is set to raw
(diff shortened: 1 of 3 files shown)


amritk (approved):
yolo

amritk (changes requested):
Just needs tests

marclave:
> Just needs tests

my favourite test, is the test of production. (lol jk ill add some tonight)

antlio:
@marclave allowed myself to add test coverage + also a commit that maintain the content type removal on request body `none`

PR #4856, opened 2025-02-22
feat(api-client): merge sidebar hooks and inject sidebar state

amritk:
**Problem**
Currently, the `useSidebarToggle` hook is causing a warning as `useLayout` is used outside of a setup function or composeable. This means its just using the default value and not actually working.

Ultimately since its just the sidebar its not the biggest deal BUT its good to get into the habit of doing things correctly, the testing improvements are far more important!

**Solution**
With this PR we:
- add a vite setup function to do some global mocks for useLayout and useSidebar + fail on error or warnings by default.
- merge the `useSidebar` and `useSidebarToggle` hooks
- use a factory function to create the state and inject it at the app creation level, this ensures we have a separate state per client instance
- moved the watcher into its respective layout components, now the sidebar will no longer accidentally open in the modal view when you resize the screen

Update: turns out a lot of our tests actually have errors  and warnings :( so will change this to opt in instead of opt out for now

**Checklist**


hanspagel (approved):
thanks for cleaning up my mess 🫡

PR #4857, opened 2025-02-22
feat(oas-utils): Set example request body if Content-Type header exists

inyourtime:
**Problem**
Currently, when the Content-Type header value exists but there is no requestBody, the API client sets the body to None.

**Solution**
With this PR, the API client will set an example request body if the Content-Type header exists.

**Checklist**

before:

https://github.com/user-attachments/assets/8d684c27-4fa0-4d9e-ab5c-3c278a7fc528

after:

https://github.com/user-attachments/assets/59826d4a-01c1-485f-b6bb-448271d1e2ab

### DIFF
--- a/packages/oas-utils/src/spec-getters/getRequestBodyFromOperation.ts
+++ b/packages/oas-utils/src/spec-getters/getRequestBodyFromOperation.ts
@@ -170,7 +170,7 @@ export function getRequestBodyFromOperation(
 
     return {
       mimeType,
-      text: typeof body === 'string' ? body : JSON.stringify(body, null, 2),
+      text: body ? typeof body === 'string' ? body : JSON.stringify(body, null, 2) : undefined,
     }
   }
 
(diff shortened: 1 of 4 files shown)


hanspagel (approved):
Awesome work! LGTM!

packages/oas-utils/src/spec-getters/getRequestBodyFromOperation.ts
```diff
@@ -155,7 +155,7 @@ export function getRequestBodyFromOperation(
 
     return {
       mimeType,
-      text: typeof body === 'string' ? body : JSON.stringify(body, null, 2),
+      text: body ? typeof body === 'string' ? body : JSON.stringify(body, null, 2) : undefined,
```
antlio, line 158: 👍

antlio (approved):
great work, tested + approved

PR #4864, opened 2025-02-22
fix: request history

marclave:
**Problem**
Currently, our request history doesnt work, any updates to the request update each history item

**Solution**
With this PR we clone each result before appending to the history, so we can properly handle the history

**Checklist**


amritk (approved):
This will probably cause other issues but it's a quick hot fix for now, we can fix it next week CC: @antlio

PR #4865, opened 2025-02-22
fix(api-reference): show */* mimetype in example response

marclave:
**Problem**
Currently, we don't handle `*/*` mimeTypes

**Solution**
With this PR we render it :) 

before:
image

after:
image

**Checklist**


hanspagel (approved):
nice, just added a test

PR #4867, opened 2025-02-22
feat(api-reference): add multiple openapi document support

marclave:
**Problem**
Currently we only support one "spec"

**Solution**
With this PR we allow for a new configuration type, called "specs" which means you can load multiple versions of an API, different APIs and it shows in the header above the sidebar.

currently we save it as a query parameter, or load the first one in the array. but in the future we can load all these in our workspace store and select them without a refresh.

image

image

**Checklist**

### DIFF
--- a/packages/api-reference/index.html
+++ b/packages/api-reference/index.html
@@ -11,10 +11,7 @@
   <body>
     <!-- Add your own OpenAPI/Swagger specification URL here: -->
     <!-- Note: The example is our public proxy (to avoid CORS issues). You can remove the `data-proxy-url` attribute if you don’t need it. -->
-    <script
-      id="api-reference"
-      data-proxy-url="https://proxy.scalar.com"
-      data-url="https://cdn.jsdelivr.net/npm/@scalar/galaxy/dist/latest.json"></script>
+    <script id="api-reference"></script>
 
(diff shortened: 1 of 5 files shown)


marclave:
some feedback from @hanspagel was to make the general config an array so we can configure each openapi document, ill do this tomorrow

xC0dex:
This is sooo great. I think the community will love this. 🥳 

For the .NET integration, it _could_ be a bit challenging to have one configuration per OpenAPI document. But we'll find a way once this is merged 👀.

hanspagel (changes requested):
great initiative! a few things that come to mind:

* we should make it possible to pass just an array of API definitions, but also an array of whole configuration objects = to have full control over the rendering of every single API defintion
* when we promote this, it’s a great opportunity to get rid of the unnecessary `spec` prefix, #4795 helps with supporting the old structure
* the ApiReference component is huge already, we should make this a dedicated component
* the URL logic could live in a separate tested file
* the API selector could be a Vue component
* let’s not use spec as word, not even in the code base
* we need tests

if you don’t mind I submit a PR with the described changes

hanspagel:
Closing in favor of #4867

PR #4871, opened 2025-02-24
ci: run `format:check` again

hanspagel:
**Problem**
We skipped the `format:check` script while we still had the old formatting.

**Solution**
This PR adds it to CI again.


hanspagel:
I’ll skip the queue here, I don’t see  a reason to wait. It’ll just cause more conflicts when we add more PRs without forcing the new formatting.

PR #4872, opened 2025-02-24
feat(api-reference): multiple api definitions

### LINKED ISSUE #1851: Multiple OpenAPI documents
### Discussed in [https://github.com/scalar/scalar/discussions/1379](https://github.com/scalar/scalar/discussions/1379)

Originally posted by **rexvel** April 4, 2024
What do you think about supporting multiple API specifications to display on a single page? The files could be merged into one to fit the current scalar component's API (React, in my case).

For instance, if I have APIs X, Y, and Z, I'd like to display them on a single page, where each API would have its own separate section (similar to the pet, store, and user sections in the Petstore example).

Here's how the navigation might look. Additionally, such changes would require support for nested navigation for endpoints, in my opinion.

image

hanspagel:
WIP

**Problem**
Currently, we support a single API definition only. This doesn't work well for people with multiple APIs. :(

**Solution**
This PR takes #4867 and adds support for multiple API definitions, but:

* in separate components
* allowing to pass in different configurations
* with tests

Fixes #1851

**Tasks**

- [x] we should make it possible to pass just an array of API definitions, but also an array of whole configuration objects = to have full control over the rendering of every single API defintion
- [ ] when we promote this, it’s a great opportunity to get rid of the unnecessary spec prefix, feat(api-reference): use Zod to parse the configuration  #4795 helps with supporting the old structure
- [x] the ApiReference component is doing a lot already, we should make this a dedicated component
- [ ] the URL logic could live in a separate tested file
- [x] the API selector could be a Vue component
- [ ] let’s not use spec as word, not even in the code base
- [ ] we should add support to the HTML API, and then in all the integrations, too
- [x] we need tests
- [ ] we need styling :))

**Checklist**


cameronrohani:
image

here's my first pass

antlio:
great work so looking forward to get this in! i noticed the following things:
- topbar menu on mobile not displayed anymore
- anchor are removed while sharing a link
- loading while navigating between apis is a bit wriggling

packages/types/src/legacy/reference-config.ts
```diff
@@ -69,305 +67,7 @@ export type ThemeId =
...
-    | 'vue' // ✅
-    | undefined
-  /**
-   * Whether to show the client button from the reference sidebar and modal
-   *
-   * @default false
-   */
-  hideClientButton?: boolean | undefined | unknown
-}
+export type ReferenceConfiguration = Partial<ApiReferenceConfiguration>
```
amritk, line 70: @hanspagel we have type parity between the two now

hanspagel: Do we need the old type though? If we modify the new type, we still want to support the old syntax, right?

amritk (changes requested):
After merging everything else it might be a little busted

image

amritk:
I wouldn't mind a little loading state as well when you switch docs, just to get rid of that jump, what you think @cameronrohani

antlio:
> I wouldn't mind a little loading state as well when you switch docs, just to get rid of that jump, what you think @cameronrohani

def! let's bring skeleton in

cameronrohani:
> I wouldn't mind a little loading state as well when you switch docs, just to get rid of that jump, what you think @cameronrohani

usually we stick loading on the h1 title

can we do the same here? 

Loading... (animate the ...'s) cc @hwkr

packages/types/src/legacy/reference-config.ts
amritk: Sorry I dont quite understand? Its only there in case something is using it. The syntax is the same between both now?

xC0dex:
I wondered if it makes sense to provide an option to preselect a document. Imagine you have 2 API versions `v1` and `v2`. Probably `v1` is already marked as deprecated, so it would be neat if it's possible to set `v2` as the default, preselected document.

amritk (approved):

PR #4874, opened 2025-02-24
fix(api-reference): doesn’t clear state properly

hanspagel:
**Problem**

Currently, the API reference doesn’t clear it’s state properly when the OpenAPI document changes. You can see this in the sidebar, that it keeps models and webhooks, even if you replace the document with an empty OpenAPI document.

This is especially a problem when used with an input or when switching between multiple API definitions.

**Solution**

With this PR we’re clearing the state properly. :)

**Checklist**

### DIFF
--- a/.changeset/tall-bikes-crash.md
+++ b/.changeset/tall-bikes-crash.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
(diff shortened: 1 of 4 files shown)


hanspagel:
Minor addition, with tests, unblocks other PRs, so I skip the queue here. 🙊

PR #4876, opened 2025-02-24
feat(api-client): request duplication, improved top nav 

bgrcs:
**Problem**
Just noting that this is a small experiment with request duplication and better top nav tab handling; completely happy for this to be closed if there were other things in the works or if a better approach is needed, but just wanted to have a go at implementing it.

At the moment, we can't duplicate requests (which I often find that I need to do, and it's a mess to create a new request and copy over all of the info).

When looking into request duplication, I also noticed that the nav bar was quite limited in a few aspects, so I moved the state to a store to allow some extra goodies.

**Solution**
This PR adds support for duplicating requests specifically (and only shows the dropdown menu item for duplication for requests only). I don't think there would be much need to implement it for other nav items (apart from examples, so let me know if you would like that added as well). The duplication works my deep copying the store request item and creating a new one with updated UID references.

This solution also addresses some current limitations with the top nav bar, and attempts to address them by making top nav bar items 1-1 with router navigation. I.e. instead of the tab holding custom state driven by route logic all over the place, it just stores the router destination data, and then we derive the title, icon and anything else from it.

I'm aware I could've created a new tab with the request summary, icon and then set it as default + navigate, but that seemed quite backwards to me, so I experimented with simplifying the nav.

Here's a quick demonstration of both:

https://github.com/user-attachments/assets/c325dfe7-4ac9-45d3-be00-2b08801ea6d3

If we are happy with this solution and the nav changes, just lmk and I'll add some tests to the duplication logic and nav behaviour before we merge :)

**Checklist**

### DIFF
--- a/packages/api-client/src/store/requests.ts
+++ b/packages/api-client/src/store/requests.ts
@@ -5,6 +5,7 @@ import {
   type RequestPayload,
   type Tag,
   createExampleFromRequest,
+  requestExampleSchema,
   requestSchema,
 } from '@scalar/oas-utils/entities/spec'
 import { LS_KEYS, iterateTitle, schemaModel } from '@scalar/oas-utils/helpers'
@@ -12,6 +13,8 @@ import { mutationFactory } from '@scalar/object-utils/mutator-record'
 import { reactive } from 'vue'
 
 import type { StoreContext } from './store-context'
+import { nanoid } from 'nanoid'
+import type { BRAND } from 'zod'
 
 /** Create top level request handlers for a workspace */
 export function createStoreRequests(useLocalStorage: boolean) {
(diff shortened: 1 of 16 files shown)


marclave:
@brunogrcsada we have this concept of "examples" which are different types of requests per a path, to keep it OpenAPI compliant.

what is your intention with duplication, do you want the path the same `/request` (not query param)

bgrcs:
> @brunogrcsada we have this concept of "examples" which are different types of requests per a path, to keep it OpenAPI compliant.
> 
> what is your intention with duplication, do you want the path the same `/request` (not query param)

Yeh good point. I think when using the client there are two ways you can think about using it; 1. from an exploration/playground stance and 2. from the client collections being fully populated by your specs.

With 1., I usually just need to actually duplicate the full request but to make a completely different route or path and then play around with that.

Even with 2 it's likely I would need duplicate a request if I'm trying out changes to my API quickly without generating a new spec (unless I have sync)

Examples are more geared towards 2 and less experimentation.

hanspagel (changes requested):
LOVE IT, what a great PR.

pls rebase/resolve the conflicts and one request:
can we make it so the request is added below the one that was duplicated and not at the end of the list?

would love @amritk to review of the code and @antlio review the ux

amritk (changes requested):
First of all, wicked PR! About time we implemented some proper state for the tabs. I did come across one issue while testing, when duplicating some requests with examples:

[duping.webm](https://github.com/user-attachments/assets/6d29f3fe-c938-45cb-beca-55463261570a)

Do you think you could split this PR up into one for the tab state and one for the request duping? Would just make it easier to test.

That being said I think the issue may have to do with the hard coded `requestUid` in the examples (gross I know but examples were always supposed to stick with a request). We are planning to migrate the store closer to the openapi spec some day so when we get to that we can get rid of this `requestUid` but for now wanna just update it and see if it fixes it?

amritk:
Actually hold up, doesn't this break openapi spec? since we have a few things that depend on unique path + method. cc: @hanspagel @marclave

marclave:
> Actually hold up, doesn't this break openapi spec? since we have a few things that depend on unique path + method. cc: @hanspagel @marclave

https://github.com/scalar/scalar/pull/4876#issuecomment-2680004296 ye

bgrcs:
@amritk Cool I'll pick up the nav state changes and we can get that in separately.

With the duplicate request spec compatibility, I would probably say we could maybe auto append a /1 at the end of the request or a quick modal? I think that we should definitely support these use cases because API experimentation would always be a common use case for the client.

packages/api-client/src/store/requests.ts
```diff
@@ -118,9 +121,108 @@ export function extendedRequestDataFactory(
...
+
+    /**
+     * Assign a new UID to the request.
+     */
+    newRequest.uid = nanoid()
+
+    /**
+     * Make it easier to see that it's a duplicate request.
+     */
+    if (newRequest.summary) newRequest.summary += ' #1'
```
antlio, line 141: what do you think about sticking to append `Copy` here insted of `#1` since its not incrementing anyway?

hanspagel: ideally, add ` (Copy)`, but also increment, so:

* ` (Copy)`, and if that exists:
* ` (Copy 2)`, and if that exists:
* ` (Copy 3)` …

hanspagel:
@bgrcs will split this up in two new PRs 🏄

bgrcs:
Thanks for all the thoughts, I have a good idea on how to sort out a good UX for the duplication too 👍 

Closing now and will split those up.

PR #4878, opened 2025-02-24
refactor(api-reference): move HTML data-attribute API to a lib

hanspagel:
**Problem**

Currently, most of our traffic goes through the `standalone.ts` file that reads data attributes from the HTML and mounts a reference. It has 3 tests only, so there’s a lot potential to improve this.

**Solution**

With this PR, we’re moving the logic to a lib, that has a bunch more tests.

I’m doing this before adding a new feature: multi document support.

**Checklist**


marclave (approved):
:) nice

PR #4884, opened 2025-02-25
fix(api-reference): improve search button and modal a11y

hwkr:
Improves the search dialog to better support NVDA vs Voiceover. Sorry if the video is a little much, should be much more verbose now though ✨ 

https://github.com/user-attachments/assets/950d1bee-3712-42e6-9887-f30220cc8e2e

**Checklist**

### DIFF
--- a/packages/components/src/components/ScalarModal/ScalarModal.stories.ts
+++ b/packages/components/src/components/ScalarModal/ScalarModal.stories.ts
@@ -83,7 +83,8 @@ export const Search: Story = {
               <ScalarSearchResultItem
                 v-for="result in results"
                 :key="result.title"
-                icon="Search">
+                icon="Search"
+                href="#">
                 {{ result.title }}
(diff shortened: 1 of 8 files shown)


packages/components/src/components/ScalarModal/ScalarModal.stories.ts
```diff
@@ -83,7 +83,8 @@ export const Search: Story = {
               <ScalarSearchResultItem
                 v-for="result in results"
                 :key="result.title"
-                icon="Search">
+                icon="Search"
+                href="#">
```
antlio, line 87: non blocking/not coming from this but seems like we have a double modal in the storie

antlio (approved):
great work and nice fix on the webhook 💖

hanspagel (approved):
thanks for the great screen recording! great work

PR #4889, opened 2025-02-25
chore: ignore some aspnetcore folders

hanspagel:
We’ve got some problems with Biome and the ASP.NET Core integration, this should fix it.


xC0dex (approved):
Thanks!

PR #4890, opened 2025-02-25
feat(api-reference): multiple documents in the HTML API

hanspagel:
**Problem**

Currently, we only support passing a single document/configuration through the HTML API. But we need a way to pass more than one very soon.

See #4872

**Solution**

With this PR we’re adding support for multiple documents/configurations.

Until now, we expected an `id="api-reference"`, but you can only have one ID with the same name. So we might want to switch to a data-attribute? Excited to hear your feedback!

This PR adds the support, but only passes the first configuration, until #4872 is done.

**Example**

```html
  <script data-scalar-api-reference data-url="/openapi-1.json"></script>
  <script data-scalar-api-reference data-url="/openapi-2.json"></script>
```

You can even mix the formats:

```html
  <script data-scalar-api-reference data-url="/openapi-1.json"></script>
  <script data-scalar-api-reference type="application/json">{"openapi":"3.1.0"}</script>
```

**Checklist**

### DIFF
--- a/.changeset/witty-humans-promise.md
+++ b/.changeset/witty-humans-promise.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
+feat: support multiple configurations through the HTML API
--- a/packages/api-reference/index.html
+++ b/packages/api-reference/index.html
@@ -12,7 +12,11 @@
     <!-- Add your own OpenAPI/Swagger specification URL here: -->
     <!-- Note: The example is our public proxy (to avoid CORS issues). You can remove the `data-proxy-url` attribute if you don’t need it. -->
     <script
-      id="api-reference"
+      data-scalar-api-reference
+      data-proxy-url="https://proxy.scalar.com"
+      data-url="https://petstore.swagger.io/v2/swagger.json"></script>
+    <script
+      data-scalar-api-reference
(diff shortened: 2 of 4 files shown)


amritk (approved):
Personally just not a fan of using data attributes in general, is this time to switch to a global var?
```html
<script>
var scalarConfig = {
  url: '...',
  proxyUrl: '...'
}
</script>
```

hanspagel:
Yeah, I prefer that, but it’s a bigger lift. I’d do that once we go all-in on the store.

hanspagel:
Yeah, I think if we want to switch to a new approach, it should be ESM. But the ESM CDN isn’t working well:

```ts
// esm.sh works, but it’s 6.7 MB (looks like tree-shaking doesn’t work):
import { createScalarReferences } from 'https://esm.sh/@scalar/api-reference'
```

```ts
// esm.run doesn’t work:
import { createScalarReferences } from 'https://esm.run/@scalar/api-reference'
```

Sooo … let’s fix the ESM CDN setup, and then promote a new JS-based API.

hanspagel:
would love to see one or two more ✅ here, though it might be intermediate, we have to live with that API for a while

hanspagel:
Let’s go with #5013

PR #4892, opened 2025-02-25
feat(api-reference): HTML rendering library

hanspagel:
**Problem**

We’re about to introduce multi document support (#4872) in all our integrations. This changes how the HTML needs to be rendered. But it’s tedious and error-prone to do this in every single integration.

**Solution**

With this PR, we’re adding a new HTML rendering lib to `@scalar/api-reference`. This can be used to render the HTML in all JS/TS-based integrations. This gives us a central place to update the HTML rendering.

This PR adds it to Express, Hono and Fastify already.

The PR adds a lot of new tests for the lib, but also for the integrations.

This adds @scalar/api-reference to those packages as a dependency.

Screenshot 2025-02-25 at 16 40 56

**Checklist**


amritk (approved):
Looks like this already existed and was just renamed, beefed up, tests added. So should be fine 👍🏾

PR #4895, opened 2025-02-25
fix(api-client): parameters are validated before the request is sent 

antlio:
**Problem**

currently empty required parameters are displaying error state by default.

**Solution**

this pr only sets the error cue on request in empty required parameter presence.

**Checklist**


amritk (commented):
~hmm feels a bit strange handling them in the requestroot. Traditionally for things like this we would use something along the lines of vue form, react hook form, or the home baked solution on org which would manage the dirty/clean states of the form elements~

on the other hand it seems to work quite well, will defer to @marclave for some QA.

packages/api-client/src/views/Request/RequestSection/RequestTable.vue
```diff
@@ -12,11 +12,7 @@ import DataTableCheckbox from '@/components/DataTable/DataTableCheckbox.vue'
 import DataTableRow from '@/components/DataTable/DataTableRow.vue'
 import type { EnvVariable } from '@/store/active-entities'
 
-import {
-  hasEmptyRequiredParameter,
-  hasItemProperties,
-  parameterIsInvalid,
-} from '../libs/request'
+import { hasItemProperties, parameterIsInvalid } from '../libs/request'
```
hanspagel, line 15:
```suggestion
import { hasItemProperties, parameterIsInvalid } from '@/libs/request'
```

antlio: seems like it cannot import it

hanspagel (changes requested):
it’s working perfectly fine! but I’m afraid this will drown in more and more logic in the component eventually. can we pull this out into a `useParameterValidation` hook and add some tests?

amritk:
I like what hans said 🧠, thats what I wanted but did not know it haha

marclave (approved):

amritk (approved):
@antlio I moved some things around, give it a test and merge if its good!

packages/api-client/src/views/Request/RequestRoot.vue
```diff
@@ -57,7 +56,7 @@ const executeRequest = async () => {
   if (!activeRequest.value || !activeExample.value || !activeCollection.value)
     return
 
-  validateParameters(activeExample.value, invalidParams.value)
+  invalidParams.value = validateParameters(activeExample.value)
```
antlio, line 59: 👍 @amritk

PR #4898, opened 2025-02-26
fix(api-reference): fix headings and add / organize regions 

hwkr:
More a11y! This PR cleans up our headings so users get a proper tree with the heading names and labels our regions correctly so you can navigate the page sections. Models and webhooks still need a bit more work but this is still an improvement.

I also added hidden headings so the full heading tree structure is available to screen readers even if the "Show More" buttons are collapsed. 

Here's a before and after video with NVDA, it also works well with voiceover but I can't take a screen recording of that because running the screen recorder confuses voiceover.

https://github.com/user-attachments/assets/06300a38-3c11-4cb0-b265-a2e23584bd30

**Checklist**

### DIFF
--- a/.changeset/spotty-queens-pretend.md
+++ b/.changeset/spotty-queens-pretend.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
+fix(api-reference): fix headings and add / organize regions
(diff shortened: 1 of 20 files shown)


hanspagel (approved):
great!!

PR #4901, opened 2025-02-26
fix(api-client): address bar background

antlio:
**Problem**

Currently some background css variable are not being properly set within the address bar, causing some glitch on scrolling fade.

image

**Solution**

this pr updates the address bar background style.

**Checklist**


packages/api-client/src/components/AddressBar/AddressBar.vue
```diff
@@ -315,26 +318,21 @@ function updateRequestPath(url: string) {
...
 }
 .addressbar-bg-states:has(.cm-focused) {
-  background: var(--scalar-background-1);
+  --addressbar-bg: var(--scalar-background-1);
   border-color: var(--scalar-border-color);
   outline: 1px solid var(--scalar-color-accent);
 }
+.addressbar-bg-states:has(.cm-focused) .fade-left,
+.addressbar-bg-states:has(.cm-focused) .fade-right {
+  --addressbar-bg: var(--scalar-background-1);
```
hanspagel, line 336: should we prefix this? or don’t we need this because it’s scoped to a CSS class?
but we should add a dash between adress and bar.

```suggestion
  --scalar-address-bar-bg: var(--scalar-background-1);
```

packages/api-client/src/components/AddressBar/AddressBar.vue
```diff
@@ -141,11 +141,11 @@ function updateRequestPath(url: string) {
     <div class="m-auto flex flex-row items-center">
       <!-- Address Bar -->
       <div
-        class="addressbar-bg-states text-xxs p-0.75 group relative order-last flex w-full max-w-[calc(100dvw-24px)] flex-1 flex-row items-stretch rounded-lg lg:order-none lg:min-w-[580px] lg:max-w-[580px] xl:min-w-[720px] xl:max-w-[720px]">
+        class="address-bar-bg-states text-xxs p-0.75 group relative order-last flex w-full max-w-[calc(100dvw-24px)] flex-1 flex-row items-stretch rounded-lg lg:order-none lg:min-w-[580px] lg:max-w-[580px] xl:min-w-[720px] xl:max-w-[720px]">
         <div
           class="pointer-events-none absolute left-0 top-0 block h-full w-full overflow-hidden rounded-lg border">
           <div
-            class="bg-mix-transparent bg-mix-amount-90 z-context absolute left-0 top-0 h-full w-full"
+            class="bg-mix-transparent bg-mix-amount-90 absolute left-0 top-0 z-[1002] h-full w-full"
```
amritk, line 148: why z1002? shouldn't we have set values for z-index?

amritk (approved):
Looks great!

packages/api-client/src/components/AddressBar/AddressBar.vue
antlio: exception, otherwise the fader are going over - we'll dig this more

hwkr: There is a z-context-plus if you want to try that

hwkr: Oh wait it's being used below, z-context-plus-2? 😓

hwkr (approved):
Code looks good! :)

packages/api-client/src/components/AddressBar/AddressBar.vue
antlio: yes - because the way we display address server and history menu - can be enhanced!

PR #4907, opened 2025-02-26
fix(integrations): express and fastify examples not running

amritk:
**Problem**

We recently added the html builder to api-reference.
pnpm will only replace the `workspace:*` dependencies on publish, here we are using the monorepo packages so it could not find the api-reference in the docker containers

**Solution**

We could either replace all `workspace:*` dependencies before pnpm i OR just copy over the dependencies into the docker container. Since there was only one dependency in this case I just copied it and it worked. Open to the other method as well. 

**Checklist**

### DIFF
--- a/.changeset/cool-brooms-add.md
+++ b/.changeset/cool-brooms-add.md
@@ -1,0 +1,6 @@
+---
+'@scalar/express-api-reference': patch
+'@scalar/fastify-api-reference': patch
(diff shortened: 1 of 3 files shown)


marclave (approved):
HES COOKING

hanspagel (approved):
U so smart

PR #4910, opened 2025-02-26
feat(scalar_fastapi): Add authentication configuration for FastAPI

g-airey:
**Problem**

Currently, it is not possible to set the `authentication` configuration options for Scalar from the `scalar-fastapi` integration as it is not a supported argument.

**Solution**

With this PR it is possible to add the authentication options in the FastAPI integration, it defaults to an empty dictionary if no argument is passed.

**Checklist**

- [ ] I’ve added a changeset (pnpm changeset).

### DIFF
--- a/integrations/fastapi/README.md
+++ b/integrations/fastapi/README.md
@@ -43,6 +43,7 @@ Currently available [configuration options](https://github.com/scalar/scalar/blo
 - `hidden_clients` (default `[]`)
 - `servers` (default `[]`)
 - `default_open_all_tags` (default `false`)
+- `authentication` (default `{}`)
 
 ## Example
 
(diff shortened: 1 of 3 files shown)


g-airey:
Sounds like changesets don't work for the python integration so I haven't generated one. Also noticed that older PRs against this integration don't bump the version in setup.py but I thought I would since I'm adding a feature, feel free to revert that if it won't work!

marclave (approved):
awesome @g-airey !! thanks so much for this pr ✨

g-airey:
Hi @marclave! Thanks for merging, what's the expected delay on the update propagating to pypi? Current version looks to still be v1.0.3 (From August last year)

PR #4911, opened 2025-02-27
fix(api-reference-react): ensure react packages work on both react versions

amritk:
**Problem**

We have issues running the packages on react 18

**Solution**

With this PR we should be able to run on both react 18 as well as 19.

**Checklist**


gregberge (approved):

marclave (approved):
based @amritk

PR #4915, opened 2025-02-27
fix(api-reference): improve high zoom level readability

hwkr:
Before / after:

https://github.com/user-attachments/assets/ae531ffb-4dc4-497e-b504-ce4834442200

**Checklist**

### DIFF
--- a/packages/api-reference/src/components/Card/Card.vue
+++ b/packages/api-reference/src/components/Card/Card.vue
@@ -18,6 +18,15 @@
   );
   position: relative;
 }
+/*
+ * Don't constrain card height on mobile
+ * (or zoomed in screens)
+ */
+@media (max-width: 600px) {
+  .scalar-card {
+    max-height: unset;
+  }
+}
 
 .scalar-card-sticky {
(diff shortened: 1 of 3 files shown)


packages/api-reference/src/components/Card/Card.vue
```diff
@@ -18,6 +18,15 @@
   );
   position: relative;
 }
+/*
+ * Don't constrain card height on mobile
+ * (or zoomed in screens)
+ */
+@media (max-width: 600px) {
```
antlio, line 25: should we use `640px` here? ([ref](https://tailwindcss.com/docs/responsive-design))

antlio (approved):
loving this! tested and approved

amritk (approved):
LGTM, just anthonys question

packages/api-reference/src/components/Card/Card.vue
antlio: but that's what we used elsewhere actually i saw we updated it so nevermind!

PR #4919, opened 2025-02-27
fix(fastify):  location of standalone file

amritk:
**Problem**

Currently, we moved the fasify integration but did not update the path to the standalone.js, this was breaking cloudbuild.

**Solution**

We update the path back into the packages, we also copy the reference package into hono as its required there as well.

**Checklist**


hanspagel (approved):
Omg yes

PR #4923, opened 2025-02-27
feat(api-reference): add redirect

amritk:
**Solution**

With this PR we add a front-end redirect.

To test:
```zsh
cd packages/api-reference
pnpm playground:esm
```
then hit [http://localhost:5173/#tag/planets/GET/planets/{planetId}](http://localhost:5173/#tag/planets/GET/planets/%7BplanetId%7D) and you should land on the path version

**Checklist**


marclave (approved):
awesome @amritk !

lets fast follow w docs & tests

PR #4926, opened 2025-02-28
fix(api-reference): improve references focus management

hwkr:
**Checklist**


amritk (approved):
Seems to be working, nice work! 

Do we have a way to set the featured clients?

hwkr:
> Do we have a way to set the featured clients?

@amritk didn't seem like it when I was in there, it wouldn't be too hard to make it configurable though

hanspagel:
> Do we have a way to set the featured clients?

not yet

packages/api-reference/src/components/Content/ClientLibraries/useFeatured.ts
```diff
@@ -0,0 +1,46 @@
+import { useHttpClientStore, type HttpClientState } from '@/stores'
+import type { ClientId, Target } from '@scalar/snippetz'
+
+export type FeaturedClient = { targetKey: Target['key']; clientKey: ClientId<Target['key']> }
+
+export function useFeatured() {
```
hanspagel, line 6: nice

can we call this `useFeaturedHttpClients` as a more descriptive name?
and add a comment describing what it does?
and add one or two tests?

👀

hanspagel (dismissed):

amritk:
@hwkr https://github.com/scalar/scalar/issues/4971

packages/api-reference/src/components/Content/ClientLibraries/useFeatured.ts
hwkr: Yeah sure, I was just consolidating the logic while doing the a11y stuff

hanspagel (approved):
supercool

PR #4928, opened 2025-02-28
feat(api-client): create requests without a name

antlio:
**Problem**

currently when creating a request from the client, cmd +k or from the sidebar we are shown a modal that require a name.

this is not ideal as in term of experience as we want to increase the time to request call and remove any barriers when possible.

**Solution**

this pr is a proposal to create untitled request creation from:
- command palette
- empty state
- request sidebar

and removes the request command palette accordingly.

**Checklist**


amritk:
I'll leave this one for product @marclave

packages/api-client/src/components/CommandPalette/TheCommandPalette.vue
```diff
@@ -52,30 +53,30 @@ import type { HotKeyEvent } from '@/libs'
...
-        icon: 'ExternalLink',
-      },
       {
         name: 'Import from OpenAPI/Swagger/Postman',
         icon: 'Import',
       },
       {
-        name: 'Add Tag',
-        icon: 'Folder',
+        name: 'Create Draft Request',
```
hanspagel, line 69:
```suggestion
        name: 'Create Request',
```

I don’t think we have to explicitly mention that a new PR will land in the drafts collection

hanspagel:
Let’s merge this before it becomes stale! If @marclave wants to chime in, we can iterate over the feature. 🏄

PR #4938, opened 2025-02-28
feat: add server change callback

marclave:
**Problem**

Currently when a server changes we cant tell the outside world its changed

**Solution**

With this PR we allow a callback for when the active server changes

**Checklist**

### DIFF
--- a/packages/api-reference/src/features/BaseUrl/BaseUrl.vue
+++ b/packages/api-reference/src/features/BaseUrl/BaseUrl.vue
@@ -26,7 +26,6 @@ const updateServerVariable = (key: string, value: string) => {
 const { onServerChange } = useConfig()
 
 const updateServer = (server: string) => {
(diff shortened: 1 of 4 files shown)


packages/api-reference/src/features/BaseUrl/BaseUrl.vue
```diff
@@ -18,6 +20,13 @@ const updateServerVariable = (key: string, value: string) => {
   variables[key] = { ...variables[key], default: value }
 
   serverMutators.edit(activeServer.value.uid, 'variables', variables)
+
+  /** Trigger the onLoaded event when the component is mounted */
```
amritk, line 24: copy and paste? what loaded event?

amritk (approved):
I think you copy and pasted that comment, but the rest of the code looks good 👍🏾 

did not test

PR #4956, opened 2025-03-03
chore: upgrade turbo

amritk:
**Problem**

I thought upgrading turbo would fix this next.js issue, it did not but I found a workaround

https://github.com/vercel/turborepo/issues/3340

Also turbo sometimes randomly hangs so maybe that'll get fixed

**Solution**

With this PR we upgrade the turbo package

**Checklist**


hanspagel:
I tried upgrading turbo and found a blocking issue a while ago. Let me review this

hanspagel (approved):
all good, nice update

PR #4958, opened 2025-03-03
fix(api-client): find correct request for route change

amritk:
**Problem**

Currently, every time you open the modal it goes to the same request

**Solution**

With this PR we boot out the function into its own util so we can test it, also fixed it.

**Checklist**


marclave (approved):
🔥

PR #4959, opened 2025-03-04
chore: enable a few biome flags

amritk:
**Problem**

Currently, we are missing import sorting!! I have been sorting my own imports like the 90s

**Solution**

With this PR we re-enable that plus a few more flags for perf and code consistency

**Checklist**


hanspagel (approved):
FINALLY, TIME HAS COME

PR #4960, opened 2025-03-04
fix(api-client): improve api client sidebar and address bar a11y

hwkr:
Improves the client sidebar toggle, search and address bar. I moved the toggle to an overlay in a single location so it doesn't break focus when opening the sidebar. As a bonus we get the toggle animation back ✨ @cameronrohani if you want to double check I didn't break the sidebar.

https://github.com/user-attachments/assets/12d266a5-79b3-44bc-a4da-4912dcbc205b

**Checklist**

### DIFF
--- a/.changeset/grumpy-flowers-know.md
+++ b/.changeset/grumpy-flowers-know.md
@@ -1,0 +1,6 @@
+---
+'@scalar/api-reference': patch
+'@scalar/api-client': patch
+---
+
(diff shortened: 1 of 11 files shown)


hanspagel (approved):
great PR, LGTM but leaving it to @cameronrohani

amritk (approved):
Yolo

hwkr:
@cameronrohani seems to be busy so I'ma merge and fast follow if we need fixes

PR #4966, opened 2025-03-04
feat(core): hello `@scalar/core`

hanspagel:
**Problem**

Currently, we’re importing a central lib from `@scalar/api-reference` to render the static HTML in our integrations (fastify, express, hono). But this package comes with a huge dependency tree.

**Solution**

With this PR we’re moving the lib to a new `@scalar/core` package, that (as of now) only has the HTML rendering lib.

This package should have no/as few as possible dependencies and provide shared libs for our packages.

It even comes with a test that’ll fail when we add a dependency.

This is the dependency tree of our Fastify integration with this PR:

```bash
dependencies:
@scalar/core link:../../packages/core
@scalar/openapi-parser link:../../packages/openapi-parser
├── ajv 8.17.1
├── ajv-draft-04 1.0.0
├── ajv-formats 3.0.1
├── jsonpointer 5.0.1
├── leven 4.0.0
└── yaml 2.4.5
@scalar/openapi-types link:../../packages/openapi-types
fastify-plugin 4.5.1
github-slugger 2.0.0
```

There’s a plan to remove all dependencies but `yaml` from `@scalar/openapi-parser`.

**Checklist**


hanspagel:
Thanks for the approval, would love to have a thumbs up by @amritk before merging.

amritk (changes requested):
cool idea, scalar core is here. we should make sure tree shaking works in this one!

requesting changes because without types as a prod dependency, we cannot export this type

`export type HtmlRenderingConfiguration = ReferenceConfiguration & CdnConfiguration & PageTitleConfiguration` 

either we add types to the dependencies, or we export this type from the types package. Either way the integration will need types as a dependency

hanspagel:
omg, yes

> or we export this type from the types package.

let’s do this

mcollina (approved):
lgtm

This is great.

packages/types/src/api-reference/html-rendering-configuration.ts
```diff
@@ -0,0 +1,29 @@
...
+   */
+  pageTitle: z.string().optional().default('Scalar API Reference'),
+})
+
+/**
+ * The configuration for the static HTML rendering using the CDN.
+ *
+ * It’s the ApiReferenceConfiguration, but extended with the `pageTitle` and `cdn` options.
+ */
+export type HtmlRenderingConfiguration = ApiReferenceConfiguration & z.infer<typeof htmlRenderingConfigurationSchema>
```
amritk, line 29:
```suggestion
export type HtmlRenderingConfiguration = Partial<ApiReferenceConfiguration> & z.infer<typeof htmlRenderingConfigurationSchema>
```
that should be a partial I believe as its for user input

amritk (approved):
Looks good 👍🏾, I will probably just test everything all at once with multi doc

PR #4967, opened 2025-03-04
refactor(snippetz): move types to `@scalar/types`

hanspagel:
**Problem**

It just hit me: It’s super strange that we import the types from `@scalar/snippetz` in `@scalar/types`. It’s the opposite of what we do with the `@scalar/api-reference` types, that live in the types package and are then used in the api-reference package.

**Solution**

With this PR we’re moving the snippetz types to the types package, which simplifies the dependency tree.

Why does it have to live in a central place? Because we need it in the `@scalar/api-reference` configuration.

**Checklist**


amritk (approved):
Seems like a lot of changes, but really wasn't bad

PR #4968, opened 2025-03-04
feat(api-client): post-response scripts 🏃‍♂️‍➡️

hanspagel:
**Problem**

Currently, you can’t execute any code after a response was received. 😭 Is this even a pro tool?

**Solution**

This PR is 90 % done, and it’s less than day of work to finish it I think. But I don’t know how long it’ll take until I do that day, so instead of rebasing it every few days, I’d prefer to get this PR in already, but not use it yet.

This PR:

* ✅ introduces the same plugin API we have for @scalar/api-reference for @scalar/api-client
* ✅ it adds a new @scalar/scripts package that has all the post-response scripts logic
* ⚠️ it doesn’t use the new post-response scripts plugin yet (see above)
* ⚠️ it doesn’t publish the package yet

This is great, though:

* None of the dependencies are loaded, when we don’t need the scripting capabilities (in @scalar/api-reference for example)
* The scripts API can be used outside the API client, for example in the CLI. 👀

There are some tasks left, to actually use it, though:

**Tasks**

- [ ] pass environment to script
- [ ] pass down the response duration
- [ ] most important: use `postman-runtime` to actually support Postman scripts 1:1 (instead of mocking it partially)

**Preview**

| Request | Response |
| -------|------|
| image | image |

**Test**

```js
pm.test("Status code is 200", async () => {
  await new Promise(r => setTimeout(r, 100))
  pm.response.to.have.status(200)
})

pm.test("Response time is less than 200ms", function () {
  pm.expect(pm.response.responseTime).to.be.below(200)
})
```

**Checklist**

### DIFF
--- a/packages/api-reference/src/components/ApiReferenceLayout.vue
+++ b/packages/api-reference/src/components/ApiReferenceLayout.vue
@@ -64,12 +64,9 @@ defineEmits<{
   (e: 'toggleDarkMode'): void
 }>()
 
-const configuration = computed(() => {
-  // The plugins for @scalar/api-reference and @scalar/api-client are different (as of now, doesn’t have to be).
-  const { plugins, ...rest } = props.configuration
-
-  return apiReferenceConfigurationSchema.parse(rest)
-})
+const configuration = computed(() =>
+  apiReferenceConfigurationSchema.parse(props.configuration),
+)
 
 // Configure Reference toasts to use vue-sonner
 const { initializeToasts, toast } = useToasts()
--- a/packages/api-reference/src/features/ApiClientModal/ApiClientModal.vue
+++ b/packages/api-reference/src/features/ApiClientModal/ApiClientModal.vue
@@ -14,7 +14,8 @@ import { useExampleStore } from '@/legacy/stores'
 import { useApiClient } from './useApiClient'
 
 const { configuration, parsedSpec } = defineProps<{
-  configuration: Partial<ApiClientConfiguration>
+  // The plugins for @scalar/api-reference and @scalar/api-client are different (as of now, doesn’t have to be).
+  configuration: Partial<Omit<ApiClientConfiguration, 'plugins'>>
   parsedSpec: Spec
 }>()
 


packages/api-client/src/views/Request/RequestSection/RequestSection.vue
```diff
@@ -232,10 +234,15 @@ const handleRequestNamePlaceholder = () => {
...
 
+      <ScalarErrorBoundary>
+        <PostResponseScripts
+          v-show="activeSection === 'All' || activeSection === 'Scripts'"
+          :operation="operation" />
+      </ScalarErrorBoundary>
+
       <!-- Spacer -->
-      <div class="-my-0.25 flex flex-grow" />
+      <div class="-mt-0.25 z-1 flex flex-grow border-b" />
```
hanspagel, line 244: shouldn’t we use the z-index presets instead of a specific number?

amritk:
THEY ARE BACK!

marclave (approved):
😮‍💨 youre insane

packages/api-client/src/libs/create-client.ts
```diff
@@ -110,6 +111,11 @@ export const createApiClient = ({
   // Create the sidebar state
   const sidebarState = createSidebarState({ layout })
 
+  // Create the plugin manager
+  const pluginManager = createPluginManager({
```
amritk, line 115: do you want to pass in the pluginManager from references? Only create this one if it doesn't exist like we do for store? Or does it not matter?

packages/scripts/src/components/ViewLayout/ViewLayoutCollapse.vue
```diff
@@ -0,0 +1,83 @@
+<script setup lang="ts">
+// TODO: This is a copy of the ViewLayoutCollapse component in the API client.
+// I’ve copied it here, so we can move the scripts to a separate package.
+// But maybe we can move the component to a shared package?
```
amritk, line 4: Yes we can!

amritk (approved):
Its a big one! I got through about half way but it looks decent!

PR #4969, opened 2025-03-04
fix: base docker image was missing git

amritk:
**Problem**

Currently, my fork of a package will not download as there's no git on the slim docker image.

**Solution**

With this PR we add git, which also adds quite a bit of disk space but 🤷🏾 

**Checklist**


amritk:
just gunna put this one in for some prod testing

hanspagel (approved):
Oops

PR #4972, opened 2025-03-04
fix(oas-utils): parameter examples array support

antlio:
**Problem**

currently parameter examples of type array or not well supported as seen in [sf compute api documentation](https://docs.sfcompute.com/api-reference#tag/orders/POST/v0/orders)

**Solution**

this pr updates the parameter entity to support examples of type array.

| before | after |
| -------|------|
| image | image |
| endpoint is not rendering properly | endpoint is not rendered | 

**Checklist**

### DIFF
--- a/packages/oas-utils/src/entities/spec/request-examples.ts
+++ b/packages/oas-utils/src/entities/spec/request-examples.ts
@@ -287,12 +287,18 @@ export function convertExampleToXScalar(example: RequestExample) {
 export function createParamInstance(param: RequestParameter) {
   const schema = param.schema as any
   const keys = Object.keys(param?.examples ?? {})
-  const firstExample =
-    keys.length && !Array.isArray(param.examples)
-      ? param.examples?.[keys[0]!]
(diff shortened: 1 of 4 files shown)


packages/oas-utils/src/entities/spec/request-examples.ts
```diff
@@ -287,7 +287,12 @@ export function convertExampleToXScalar(example: RequestExample) {
 export function createParamInstance(param: RequestParameter) {
   const schema = param.schema as any
   const keys = Object.keys(param?.examples ?? {})
-  const firstExample = keys.length ? param.examples?.[keys[0]!] : null
+  const firstExample =
```
amritk, line 290: This is getting a bit crazy, I almost think we should turn this rule on

https://biomejs.dev/linter/rules/no-nested-ternary/#_top

amritk (changes requested):
I would bump that ternary into a function with if statements and returns to make it easier to read :)

hanspagel (approved):
nice fix!

PR #4973, opened 2025-03-04
feat(api-reference): add version selector above server select

amritk:
Just built a "version" selector that can be used for anything really. Open to different naming

image

To test:
```zsh
cd packages/api-reference
pnpm playground:esm
```

**Checklist**


amritk:
Apparently we can just add multiple api doc support instead

PR #4980, opened 2025-03-05
refactor(api-reference): use store for title and stability

hanspagel:
Just another refactor to use the store (instead of our legacy transformed operation) for:

* the operation title and
* the stability level

tiny step towards the future

EDIT: lol, 19 files changed. those refactors escalate quickly 🙈


amritk (approved):
Seems to be working, for the record I think we should make our store improvements BEFORE we move references to use the store, then we'll have less work to do

PR #4981, opened 2025-03-05
fix(api-client): add keyboard instructions for code inputs

hwkr:
Adds visual and screen reader instructions for exiting the code mirror input. I tried to add a screen reader label to the code mirror input but wasn't able to get it working so this just uses an alert for now.

Google Chrome-2025-03-05-03-45-20@2x

**Checklist**


hanspagel (approved):
niiice

marclave (approved):

PR #4982, opened 2025-03-05
fix(api-client): multi form upload

antlio:
**Problem**

currently the multi form upload style is broken.

**Solution**

this pr updates the multi form styles + its logic in order to remove the file only if the key have been updated, otherwise the all row.

| before | after |
| -------|------|
| image | image |
| overflowing row creating a gap | row height is preserved | 

**Checklist**

### DIFF
--- a/packages/api-client/src/views/Request/RequestSection/RequestBody.vue
+++ b/packages/api-client/src/views/Request/RequestSection/RequestBody.vue
@@ -367,12 +367,22 @@ const removeBinaryFile = () =>
 function handleRemoveFileFormData(rowIdx: number) {
   const currentParams = formParams.value
   const updatedParams = [...currentParams]
+  const param = currentParams[rowIdx]
+  const file = param?.file as File | undefined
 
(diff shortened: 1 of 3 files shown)


packages/api-client/src/views/Request/RequestSection/RequestBody.vue
```diff
@@ -367,12 +367,22 @@ const removeBinaryFile = () =>
...
-    // Removes the row if not the last one
+  const param = currentParams[rowIdx]
+  const file = param?.file as File | undefined
+
+  // Empty key value or non updated file name then remove the row
+  if (
+    currentParams.length > 1 &&
+    ((!param?.key && !param?.value) ||
+      (file && param?.key === file.name && param?.value === file.name))
+  ) {
```
amritk, line 378: I would move some of these to a variable just to make it more readable, that or add comments to each line

amritk (approved):
I'm not sure what exactly it does, did not test

PR #4994, opened 2025-03-06
ci: build Docker base image

hanspagel:
**Problem**

Currently, we build the Docker images only in main and we break them every few days.

**Solution**

This PR builds the base Docker image in PRs.

I failed to build the other Docker images, but at least we’re building the one. Tiny step.

**Checklist**


amritk:
The base image didn't have issues though, it was the other ones. Or is this the first step?

hanspagel:
Yeah, I wanted to do all, but I failed. And I didn't want to spend more time. So yes, a first step I guess. 😅

PR #4995, opened 2025-03-06
feat(api-reference): content type select

antlio:
**Problem**

currently the content type selector is using browser native dropdown.

**Solution**

this pr is a proposal to use the scalar dropdown component.

| before | after |
| -------|------|
| image | image |

**Checklist**

### DIFF
--- a/packages/api-reference/src/features/Operation/components/ContentTypeSelect.vue
+++ b/packages/api-reference/src/features/Operation/components/ContentTypeSelect.vue
@@ -65,7 +65,7 @@ const contentTypeSelect = cva({
 <template>
   <ScalarListbox
     v-if="prop?.requestBody && contentTypes.length > 1"
-    :modelValue="selectedOption"
+    v-model="selectedOption"
     :options="options"
     placement="bottom-end"
     @update:modelValue="handleSelectContentType"
--- a/packages/api-reference/src/features/Operation/components/ContentTypeSelect.vue
+++ b/packages/api-reference/src/features/Operation/components/ContentTypeSelect.vue
@@ -1,6 +1,14 @@
 <script setup lang="ts">
+import {
+  cva,
+  ScalarButton,
+  ScalarIcon,
+  ScalarListbox,
+} from '@scalar/components'
 import type { ContentType, RequestBody } from '@scalar/types/legacy'
 import { computed, ref } from 'vue'
+
+import ScreenReader from '@/components/ScreenReader.vue'
 
 const prop = defineProps<{
   requestBody?: RequestBody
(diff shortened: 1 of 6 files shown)


hwkr (changes requested):
I think you want ScalarListbox not ScalarDropdown unless I'm missing something 🤔

antlio:
> I think you want ScalarListbox not ScalarDropdown unless I'm missing something 🤔

or! bad old habit, also thinking about bringing a specific select component asp, could be a good reka usecase too?

hwkr:
> > I think you want ScalarListbox not ScalarDropdown unless I'm missing something 🤔
> 
> or! bad old habit, also thinking about bringing a specific select component asp, could be a good reka usecase too?

The listbox is basically just a Javascript select, is there something you want that's select specific?

hwkr (approved):
Looks good! Do we need to indent the listbox outline?

antlio:
> Looks good! Do we need to indent the listbox outline?

i did + added your great screen reader

packages/api-reference/src/features/Operation/components/ContentTypeSelect.vue
```diff
@@ -30,101 +35,58 @@ const contentTypes = computed(() => {
...
 <template>
-  <label
-    class="content-type-select"
-    :class="{ 'content-type-no-select': contentTypes.length <= 1 }">
+  <ScalarListbox
+    v-if="prop?.requestBody && contentTypes.length > 1"
+    v-model="selectedOption"
+    :options="options"
+    placement="bottom-end"
+    @update:modelValue="handleSelectContentType"
```
hanspagel, line 71: seems dangerous to use `v-model` (alias for `@update:modelValue` and `:modelValue`) and `@update:modelValue`

```suggestion
    :modelValue="selectedOption"
    :options="options"
    placement="bottom-end"
    @update:modelValue="handleSelectContentType"
```

hanspagel (changes requested):
just one minor nitpick

and any chance we can add a `ContentTypeSelect.test.ts` with 3 tests or so?

packages/api-reference/src/features/Operation/components/ContentTypeSelect.vue
antlio: oops good catch

amritk (approved):
Tests are in!

amritk:
Actually it doesn't seem to work! I will fix then merge

amritk (changes requested):
Alright I fixed the missing v-model but this still needs some work.

I'm not sure if it even does anything as the section keeps collapsing. We probably want to prevent propogation on there somewhere

packages/api-reference/src/features/Operation/components/ContentTypeSelect.vue
```diff
@@ -65,7 +65,7 @@ const contentTypeSelect = cva({
 <template>
   <ScalarListbox
     v-if="prop?.requestBody && contentTypes.length > 1"
-    :modelValue="selectedOption"
+    v-model="selectedOption"
```
antlio, line 68: oh my bad updated it after review but didn't test it again, tysm!

amritk (approved):
Yes much better 👨🏾‍🍳

PR #4996, opened 2025-03-06
chore(workspace): default vite css name

geoffgscott:
Workspace updates:

bumps pnpm to align with org
Adds vite6->vite5 alignment in build tooling


hanspagel (approved):
sure, why not

PR #4999, opened 2025-03-06
feat(api-reference): remove `spec` prefix 🌶️

hanspagel:
**Problem**

Currently, we add a `spec` prefix to the content sources (for some reason):

```diff
const configuration = {
-  spec: {
-    url: '/openapi.json',
-  }
}
```

It might make sense to nest a few of the settings, but the content and the URL is so important it should be top-level.

Every time I see code that deals with it, it feels strange to tap into the nested object.

**Solution**

With this PR we’re removing the `spec` prefix. We still understand the old syntax, warn about it in the browser console and in TypeScript and migrate to the new syntax using the new Zod pipeline for the configuration.

```diff
const configuration = {
+  url: '/openapi.json',
}
```

**Checklist**

### DIFF
--- a/.changeset/mighty-mayflies-buy.md
+++ b/.changeset/mighty-mayflies-buy.md
@@ -1,0 +1,21 @@
+---
+'@scalar/api-reference-editor': minor
+'@scalar/api-reference-react': minor
+'@scalar/api-client-react': minor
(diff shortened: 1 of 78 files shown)


marclave (approved):
@hanspagel this is incredible work, tested and 🔥

PR #5001, opened 2025-03-06
fix(api-reference): hidden clients zod schema update

amritk:
**Problem**

Currently, we missed allowing true for `hiddenClients`

**Solution**

With this PR we allow `true`

**Checklist**


hanspagel (approved):
That’s literally true

PR #5002, opened 2025-03-06
fix(components): scalar modal position

antlio:
**Problem**

Currently the modal is centered when opening the client from an api reference.

**Solution**

this pr removes the x axis margin from the scalar modal component.

| before | after |
| -------|------|
| image | image |

**Checklist**


amritk (approved):
gogogog

PR #5010, opened 2025-03-07
fix(api-client): incorrect body being selected

amritk:
**Problem**

Currently, the body selected would be none due to an early return of mime type

**Solution**

With this PR we remove that early return and things seem to just work?

**Checklist**

### DIFF
--- a/integrations/nuxt/package.json
+++ b/integrations/nuxt/package.json
@@ -25,7 +25,7 @@
     "node": ">=18"
   },
   "scripts": {
-    "build": "pnpm dev:prepare && nuxt-module-build build",
+    "build": "NUXT_TELEMETRY_DISABLED=false pnpm dev:prepare && nuxt-module-build build",
     "dev": "pnpm dev:prepare && nuxi dev playground",
     "dev:build": "nuxi build playground",
     "dev:prepare": "nuxi prepare && nuxt-module-build build --stub && nuxt-module-build prepare && nuxi prepare playground",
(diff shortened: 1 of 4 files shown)


integrations/nuxt/package.json
```diff
@@ -25,7 +25,7 @@
     "node": ">=18"
   },
   "scripts": {
-    "build": "pnpm dev:prepare && nuxt-module-build build",
+    "build": "NUXT_TELEMETRY_DISABLED=false pnpm dev:prepare && nuxt-module-build build",
```
amritk, line 28: bonus fix of that damn interactive command in nuxt

packages/api-client/src/views/Request/RequestSection/RequestBody.test.ts
```diff
@@ -64,7 +64,7 @@ describe('RequestBody.vue', () => {
 
   it('renders correctly with no body', async () => {
     const wrapper = mount(RequestBody, props)
-    expect(wrapper.text()).toContain('No Body')
+    expect(wrapper.findComponent({ name: 'ScalarListbox' }).text()).toContain('None')
```
amritk, line 67: wasn't able to test this issue but these are a bit better so they look at the listbox

packages/api-client/src/views/Request/RequestSection/RequestBody.vue
```diff
@@ -68,16 +68,16 @@ const contentTypeOptions = (
 /** Match the activeBody to the contentTypeOptions */
 const activeExampleContentType = computed(() => {
   const { activeBody, formData, raw } = example.body
-  if (raw?.mimeType) return raw.mimeType
+
```
amritk, line 71: this line was causing the issues cc: @marclave

marclave (approved):

PR #5013, opened 2025-03-07
feat(api-reference): new JS API 🤖

hanspagel:
**Problem**

Currently, we’re using data-attributes for the most popular way to use Scalar. But this has a few issues:

* ❌ The configuration object has to fit in HTML attributes, you can’t pass functions (e.g. `tagsSorter`).
* ❌ It becomes harder to add multiple documents/configurations. See:
  * #4872
  * #4890

```diff
<!doctype html>
<html>
  <head>
    <title>Scalar API Reference</title>
    <meta charset="utf-8" />
    <meta
      name="viewport"
      content="width=device-width, initial-scale=1" />
  </head>
  <body>
-    <script
-      id="api-reference"
-      data-url="https://cdn.jsdelivr.net/npm/@scalar/galaxy/dist/latest.yaml"></script>
-
-    <!-- Optional: You can set a full configuration object like this: -->
-    <script>
-      var configuration = {
-        theme: 'purple',
-      }
-
-      document.getElementById('api-reference').dataset.configuration =
-        JSON.stringify(configuration)
-    </script>

    <script src="https://cdn.jsdelivr.net/npm/@scalar/api-reference"></script>
  </body>
</html>
```

**Solution**

✅ This PR proposes a new JS-based API, while still supporting the old API, all through our famous CDN script.
✅ Updating our integrations to add multi document support becomes way easier.
✅ Supporting JS functions in the configuration object becomes possible.

```html
<!doctype html>
<html>
  <head>
    <title>Scalar API Reference</title>
    <meta charset="utf-8" />
    <meta
      name="viewport"
  </head>

  <body>
    <div id="app"></div>

    <!-- Load the Script -->

    <script>
        // Avoid CORS issues
      })
    </script>
  </body>
</html>
```

Or you can create the app and mount it later:

```ts
const app = Scalar.createApiReference({
})

// Mounting
app.mount('#app')
```

And you can update the configuration like this:

```ts
const app = Scalar.createApiReference('#app', {
})

// Updating the configuration
app.updateConfiguration({
  url: 'https://petstore.swagger.io/v2/swagger.json',
})
```

⚠️ The example doesn’t have the `spec` prefix, PR:

* #4999

**Future Tasks**

- update the documentation
- update the html-rendering lib to use it in the integrations
- expose this new function through the package, not just the CDN build

**Checklist**


packages/api-reference/index.html
```diff
@@ -9,28 +9,23 @@
...
+    <div id="app"></div>
 
+    <!-- Our script -->
     <script
       src="src/standalone.ts"
       type="module"></script>
+
+    <!-- Initialize the Scalar API Reference -->
+    <script type="module">
+      Scalar.createApiReference('#app', {
```
geoffgscott, line 21: I think we can improve the user control over this a bit. Mounting should be explicit not implicit and we should also expose an updateConfig method. 
```
const config = {}

const apiReference = createApiReference(`#app`, config)
apiReference.mount()
apiReference.updateConfig({...})
```

hanspagel: cool, added this

amritk (approved):
Let's go

PR #5022, opened 2025-03-07
fix(fastify-api-reference): ensure we copy over the standalone file into the correct directory

amritk:
**Problem**

Currently, the standalone file seems to be missing

**Solution**

With this PR we find it using resolve so its relative to the current file. ALSO as a bonus we remove undefined selected security schemes.

**Checklist**


hanspagel (approved):
Omg, nice

any idea how we can catch this in CI next time?

amritk:
@hanspagel we would need to build the docker images

PR #5025, opened 2025-03-07
chore: update eol to LF

xC0dex:
**Problem**

Windows users (✋) had issues with prettier and biome during a commit. The formatters replaced every EOL to `crlf`.

image

**Solution**
I added a `.gitattributes` file to the root of the project and updated the prettier config. Other big projects also have the `* text=auto` setting in the configuration. It looks like this works on my (windows) machine. It would be awesome if someone with a mac could verify that `pnpm format` works as expected.

With this PR …

**Checklist**

- [ ] ~I’ve added a changeset (`pnpm changeset`).~
- [ ] ~I’ve added tests for the regression or new feature.~
- [ ] ~I’ve updated the documentation.~

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+* text=auto
--- a/.prettierrc
+++ b/.prettierrc
@@ -10,6 +10,7 @@
   "singleQuote": true,
   "singleAttributePerLine": true,
   "bracketSameLine": true,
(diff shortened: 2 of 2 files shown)


hanspagel (approved):
LGTM 🫡

hanspagel:
More context: https://prettier.io/docs/options#end-of-line

PR #5027, opened 2025-03-07
fix(fastify): missing script to build playground

amritk:
**Problem**

Currently, we were missing the build:playground script.

**Solution**

With this PR we add it back

**Checklist**


hanspagel (approved):
Never ending

PR #5041, opened 2025-03-10
refactor(oas-utils): separate Zod OpenAPI schemas from the rest

hanspagel:
**Tasks**

- [ ] Fix the migrations
- [ ] Extract all OpenAPI extensions
- [ ] Align naming, types, exports and all that
- [ ] Write so many tests

**Problem**

Currently, we try to stick with the Zod schemas as close as possible to the specification, but there are a few problems:

* We merge extensions on various levels and in various way.
* We’ve accidentally added custom attributes here and there along the way.

**Solution**

This PR tries to separate the schemas that we create from the OpenAPI specification, from the custom stuff.

This is just one step towards having super strict Zod schemas actually sticking to the OpenAPI specification.

And it’ll help to generate documentation for all the custom extensions that we use, because they live in their own place and are all added in the same way.

It’s also easier to cover them all with tests.

**Checklist**


hanspagel:
Great attempt, but I have to close this. I want to apply new formatting rules, it already has conflicts, and I want to rebase to remove the malicious GitHub Action. Quickest is to just close the PR for now.

PR #5042, opened 2025-03-10
feat(nextjs): use HTML rendering lib from `@scalar/core`

hanspagel:
With this PR we’re using the central HTML rendering library from `@scalar/core` in the Next.js integration.

This will make it so easy to add the multi document feature.

### DIFF
--- a/integrations/nextjs/src/types.ts
+++ b/integrations/nextjs/src/types.ts
@@ -1,0 +1,6 @@
+import type { HtmlRenderingConfiguration } from '@scalar/core/libs/html-rendering'
+
+/**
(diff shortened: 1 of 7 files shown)


integrations/nextjs/src/types.ts
```diff
@@ -0,0 +1,6 @@
+import type { HtmlRenderingConfiguration } from '@scalar/core/libs/html-rendering'
+
+/**
+ * The configuration for the Scalar API Reference for Hono
```
amritk, line 4:
```suggestion
 * The configuration for the Scalar API Reference for Next.js
```

amritk (approved):
lets go

PR #5045, opened 2025-03-10
fix(api-client): complex auth indicator is not retrieved

### LINKED ISSUE #4933: Authentication card subtitle for "Required" schemes
### What happens?

When multiple auth schemes are defined as "required", the card subtitle only indicates that one of the schemes is required (e.g. "Client ID required").

Image

The dropdown correctly shows the group of schemes as a single entry under "Required authentication"

Image

Relevant section of Open API doc:

```json
{
  "components": {
    "securitySchemes": {
      "Client ID": {
        "type": "apiKey",
        "description": "The 'Client ID', passed as a HTTP header",
        "name": "x-api-client-id",
        "in": "header"
      },
      "Client Key": {
        "type": "apiKey",
        "description": "The 'Client Key' API key, passed as a HTTP header",
        "name": "x-api-client-key",
        "in": "header"
      }
    }
  },
  "security": [
    {
      "Client ID": [ ],
      "Client Key": [ ]
    }
  ],
}
```

I expected the card subtitle to reflect that both the "Client ID" and "Client Key" schemes were required.

### OpenAPI Document

https://sandbox.scalar.com/e/-9tqg

antlio:
**Problem**

currently when using complex auth within the client the indicator is not displaying the right value.

**Solution**

this pr updates the auth indicator function to support complex auth and fixes #4933.

| before | after |
| -------|------|
| image | image |
| auth is only required for one requirement | auth required for both requirements | 

**Checklist**

### DIFF
--- a/.changeset/thin-seahorses-tickle.md
+++ b/.changeset/thin-seahorses-tickle.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
+
+fix: adds auth indicator complex auth support
--- a/packages/api-client/src/views/Request/RequestSection/RequestAuth/RequestAuth.test.ts
+++ b/packages/api-client/src/views/Request/RequestSection/RequestAuth/RequestAuth.test.ts
@@ -49,6 +49,20 @@ vi.mock('@/store/store', () => ({
         password: 'pass',
         token: '123456',
       },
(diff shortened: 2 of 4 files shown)


amritk (approved):
Very safe

PR #5046, opened 2025-03-10
feat(client-scalar-com): add OpenGraph tags

hanspagel:
We don’t have OpenGraph tags, no description and nothing on https://client.scalar.com since forever.

Screenshot 2025-03-10 at 16 56 17

Screenshot 2025-03-10 at 17 06 30

Let’s fix this!


amritk (approved):
yup

PR #5047, opened 2025-03-10
refactor(api-reference-react): move to integrations/ folder

hanspagel:
Let’s move the React integration over, too.


amritk:
Taking over this since I need to add multi config to the react config

PR #5053, opened 2025-03-10
fix(oas-utils): allow null as an oauth2 scope

amritk:
**Problem**

Currently, having null a scope breaks everything

**Solution**

With this PR we add a catch for null values

**Checklist**


hanspagel (approved):
True

PR #5070, opened 2025-03-12
feat(components): create ScalarTextInput and ScalarTextArea

hwkr:
Removes `ScalarTextField` and adds the following components to the components library:

- `ScalarForm`: A wrapper around the `<form>` better suited to JS forms (prevents default for the submit event)
- `ScalarFormSection`: Sections for groups of inputs
- `ScalarFormField`: A `<label>` to wrap an input
- `ScalarTextInput`: Provides a styled `<input>` element aligned with how we've been using them in the dashboard, support prefixes and suffixes
- `ScalarTextArea`: Provides a styled `<textarea>` element that supports auto resizing 

Check out the storybook preview to play with them :)

Where possible I tried to keep things as simple as possible and as close to the underlying html elements with the expectation that they'll be extended as needed. While I was in there I also added a tests for the new components and some docs. Also I update `useBindCx` to support putting the merged `class` on one element while putting the remaining attributes on another element.

I wanted to also add the `useForm` hook and integrate it's submission control and state with `ScalarForm` but it was pretty complicated so I figured it's better to start with this and we can add `useForm` down the road.

Finally I updated the two uses of `<ScalarTextField>` in the client to use `<ScalarTextInput>` so now they better align with our internal use 🥳

Video of client:

https://github.com/user-attachments/assets/7c3733af-ebad-4988-9931-59368ea69c6b

**Checklist**


antlio (approved):
great work and tests!

PR #5072, opened 2025-03-12
fix(api-reference): path parameter in operation 

### LINKED ISSUE #5052: Router does not resolve a direct link to an endpoint.
### What happens?

When I link to a specific section of the API, the link does not resolve.   This specific issue is on client side rendered Nuxt.

Can be reproduced a few ways:
- copy the link using the hash next to the header, and paste in another window.
- view a specific endpoint and refresh

The behavior:
The URL remains correct, but the content that renders is the root route.  This can be seen in the screenshot or by following this link to POST scoped variable:

https://cycle.io/docs/api/platform/tag/scoped-variables/POST/v1/environments/%7BenvironmentId%7D/scoped-variables

This is an issue since we often refer users to a specific section for interlinking with our other documentation.

I spun up a React hosted application and experienced the same issue, so my thought is that this is a Scalar level issue and not an issue with the integration.

If I link to the section header and not the individual endpoint, it resolves correctly. 

Example:
https://cycle.io/docs/api/platform/tag/memberships

Is this a known issue, and is there a workaround to get routing working properly?

Image

I expect to be able to link a user to a specific endpoint and for that route to resolve properly.

### OpenAPI Document

_No response_

antlio:
**Problem**

currently operations with path parameters maintain curly braces in the operation id, while it gets encoded in the url, leading to issue when sharing a url.

**Solution**

this pr is a proposal to normalize path parameter at the nav state level to keep matching id and url. fixes #5052 

**Checklist**

### DIFF
--- a/packages/api-reference/src/hooks/useNavState.test.ts
+++ b/packages/api-reference/src/hooks/useNavState.test.ts
@@ -78,6 +78,21 @@ describe('useNavState', () => {
     it('should generate webhook ID', () => {
       expect(navState.getWebhookId({ name: 'Test Webhook', method: 'POST' })).toBe('webhook/POST/test-webhook')
     })
+
+    it('should normalize operation ID with path parameters', () => {
+      const operation = {
+        httpVerb: 'GET',
+        path: '/environments/{environmentId}/scoped-variables',
+      } as const
+      const parentTag = {
+        name: 'Test Tag',
+        description: 'Test Description',
+        operations: [],
+      } satisfies Tag
+      expect(navState.getOperationId(operation, parentTag)).toBe(
+        'tag/test-tag/GET/environments/environmentId/scoped-variables',
+      )
+    })
   })
 
   describe('custom slug generation', () => {
--- a/.changeset/itchy-rats-search.md
+++ b/.changeset/itchy-rats-search.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
+fix: normlizes nav state operation id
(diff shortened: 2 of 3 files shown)


packages/api-reference/src/hooks/useNavState.test.ts
```diff
@@ -78,6 +78,21 @@ describe('useNavState', () => {
...
+        httpVerb: 'GET',
+        path: '/environments/{environmentId}/scoped-variables',
+      } as const
+      const parentTag = {
+        name: 'Test Tag',
+        description: 'Test Description',
+        operations: [],
+      } satisfies Tag
+      expect(navState.getOperationId(operation, parentTag)).toBe(
+        'tag/test-tag/GET/environments/environmentId/scoped-variables',
```
hanspagel, line 93: kind of strange that we just ditch the brackets 🤔 I don’t know a better solution though

antlio: felt it was easier to use within url but open to ideas!

hanspagel: This works fine in Chrome: https://galaxy.scalar.com/#tag/planets/GET/planets/{planetId}

How can I reproduce the issue? Can we just not encode the brackets?

antlio: we could but does it make sense to maintain brackets in url?

antlio: i'll update this way then

amritk: lets close and re-open this PR if needed

amritk:
Closing as it doesn't seem to be an issue

PR #5073, opened 2025-03-12
ci: merge deploy-examples jobs

hanspagel:
For some reason, we have one job to deploy the examples in the main branch (deploy-examples.yml), and one job to deploy examples in PRs (ci.yml).

I removed deploy-examples.yml and made ci.yml do both.

This should leverage the build cache and help to avoid build issues.


amritk:
Thanks! but also why did you get rid of the start page? I like that one

PR #5074, opened 2025-03-12
fix(api-client): icon positions

antlio:
**Problem**

currently some icon and label are not perfectly aligned in the client request section.

**Solution**

this pr updates icon and label alignement to be perfectly aligned and heal alignment OCD. label like required and clear action keep a safe margin to maintain visual aligment with icons.

| before | after |
| -------|------|
| image | image |

**Checklist**


amritk (approved):
The 🪄

PR #5075, opened 2025-03-12
chore(worksapce): adds tailwindcss to workspace

antlio:
**Problem**

currently the `prettier-plugin-tailwindcss` dependency is available to the workspace level, where `taildwindcss` is not, which is creating some formatting issue and then breaks the format CI.

**Solution**

this pr adds `tailwindcss` dev dependency at the workspace level.

**Checklist**


amritk (approved):
Nice one haha!

PR #5081, opened 2025-03-12
refactor(void-server): strict TS config

hanspagel:
Let’s use the stricter TS config for @scalar/void-server. :)


amritk (approved):
always welcome

PR #5082, opened 2025-03-12
test(api-reference): add a CDN example

hanspagel:
WIP

Adds new CDN tests for the new JS API and the old data attributes API

Will fail until we release

**Problem**

Currently, …

**Solution**

With this PR …

**Checklist**


hanspagel:
Will make a fresh attempt

Issue #5077, opened 2025-03-12
show deprecated (query) parameters


hanspagel:
> I used to indicate when a query parameter is deprecated, as we usually give our customers a several-months prior notice. As of today, there is no specific styling for deprecated query parameters with Scalar.

Source: https://github.com/scalar/scalar/discussions/5076

PR #5083, opened 2025-03-12
feat(api-reference): show deprecated parameters

hanspagel:
**Problem**

Currently, we don’t show whether a parameter is deprecated.

See #5077

**Solution**

We’ve had the styles already, with this PR we’re passing down the `deprecated` attribute to show whether a parameter is deprecated.

**Preview**

Screenshot 2025-03-12 at 14 57 58

**Checklist**

### DIFF
--- a/.changeset/happy-bees-search.md
+++ b/.changeset/happy-bees-search.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
+feat: show if a parameter is deprecated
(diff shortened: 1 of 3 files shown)


antlio:
would it make sense to use also a `property-deprecated` item to it?
image

antlio (approved):
awesome call! left non blocking comment

cameronrohani:
I think strike through is a good idea :) 

> would it make sense to use also a `property-deprecated` item to it? also i'm wondering if we should avoid using line through as it might decrease readability even though i get the reason behind it.
> 
> v1	v2
> image	image
> with line through	without line through

PR #5084, opened 2025-03-12
fix(api-reference): sidebar arrows on all tags configuration

antlio:
**Problem**

currently when using `defaultOpenAllTags` configuration, sidebar arrows are still display even though they cannot be folded.

**Solution**

this pr avoids sidebar arrows from being displayed when the configuration is on.

| before | after |
| -------|------|
| image | image |

**Checklist**

### DIFF
--- a/.changeset/giant-icons-pay.md
+++ b/.changeset/giant-icons-pay.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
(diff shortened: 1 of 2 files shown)


amritk:
Do we want them to be able to close them though? it just opens by default not forced open?

amritk (approved):
sure then

PR #5085, opened 2025-03-12
feat(api-client): address bar server deselection

antlio:
**Problem**

currently a server cannot be deselected from the client address bar when set.

**Solution**

this pr introduces deselection capability while keeping the possibility to select back server.

| after |
| -------|
| image |

**Checklist**

### DIFF
--- a/packages/api-client/src/store/active-entities.ts
+++ b/packages/api-client/src/store/active-entities.ts
@@ -156,14 +156,14 @@ export const createActiveEntitiesStore = ({
   /** The currently selected server in the addressBar */
   const activeServer = computed(() => {
     // Request has a selected server
-    if (isDefined(activeRequest.value?.selectedServerUid)) {
+    if (activeRequest.value?.selectedServerUid) {
       // Return server if selected
       const server = servers[activeRequest.value.selectedServerUid]
       if (server) return server
     }
 
     // Collection has a selected server
-    if (isDefined(activeCollection.value?.selectedServerUid)) {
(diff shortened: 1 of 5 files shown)


packages/api-client/src/store/active-entities.ts
```diff
@@ -154,15 +154,31 @@ export const createActiveEntitiesStore = ({
...
+        if (server) return server
+      } else {
+        return undefined
+      }
+    }
+
+    // Collection has a selected server
+    if (isDefined(activeCollection.value?.selectedServerUid)) {
+      // Return server if selected
+      if (activeCollection.value.selectedServerUid !== null) {
```
amritk, line 172: isDefined should check for null so you do not need this line

packages/api-client/src/store/active-entities.ts
```diff
@@ -154,15 +154,23 @@ export const createActiveEntitiesStore = ({
...
-      servers[
-        (activeRequest.value?.selectedServerUid ||
-          activeCollection.value?.selectedServerUid ||
-          activeCollection.value?.servers[0]) ??
-          ''
-      ],
-  )
+  const activeServer = computed(() => {
+    // Request has a selected server
+    if (isDefined(activeRequest.value?.selectedServerUid)) {
```
amritk, line 159:
```suggestion
    if (activeRequest.value?.selectedServerUid) {
```
sorry one more change, we should just check like this otherwise an empty string might fall through!

PR #5091, opened 2025-03-12
fix(import): importing of unescaped configuration urls

amritk:
**Problem**

Currently, we aren't importing URLs from the config attribute

**Solution**

With this PR we grab them

**Checklist**


hanspagel (approved):
Pls address my comment, but I’ll approve it already

PR #5092, opened 2025-03-12
fix: treat png files as binary

amritk:
**Problem**

Currently, we added a new line to git attributes which seems to be editing our snapshots

**Solution**

With this PR we ignore png files (may need to add other file types down the road)

Also added a changeset to bump references from a previous fix

**Checklist**


amritk:
Going to yolo merge to get the changeset in

xC0dex:
You're so right. I totally forgot about images.
Thanks for the fix.

PR #5096, opened 2025-03-13
chore: Cursor rules

hanspagel:
**Problem**

Currently, we’re using Cursor, but only have individual Cursor rules (@marclave) or non at all (me). This makes Cursor write in various styles or maybe even delete code that we want, or introduce new patterns/styles.

**Solution**

With this PR we’re adding Cursor rules that we all share. This could help to streamline the output for us, and even for external contributors.

It’s also a perfect place for humans to look up coding style related things, or even learn from what we consider best practices.

They are probably not perfect, but it’s a start. @amritk feel free to add to it! I quickly jotted down a few points and asked ChatGPT to improve it for the purpose of using it as a Cursor rule.

**Summary**

#### Vue Components
- Keep components small, focused, and composable.
- Move logic out of templates into `computed`, `methods`, or composables.
- Components should be fault-tolerant: handle missing or malformed data gracefully.
- Keep components clean—avoid putting too much logic in them. Extract logic to separate files (`/composables` or utilities).
- Use Tailwind CSS for styling. Keep class lists clean, readable, and consistent.
- Don’t rely on markup or Tailwind classes in tests or component logic.
- Prioritize readability, accessibility, and testability.

#### Tests
- Use Vitest (no globals). Import `describe`, `it`, and `expect` explicitly.
- Test files live **next to** the file under test and end with `.test.ts`.
- Each test file starts with a `describe()` matching the file name.
- Minimize mocking—prefer refactoring to avoid unnecessary mocks.
- Test happy paths, error handling, and edge cases. Think like a QA engineer.
- Playwright tests are limited and live in `/playwright/tests/*.spec.ts`.
- Optimize tests for clarity and readability. Comments should explain *why*, not *what*.

#### TypeScript
- Prefer `type` over `interface` for all type definitions.
- Be explicit with return types on exported functions and composables.
- Avoid `any`; use `unknown` if necessary, with proper type checks.
- Use string literal unions instead of enums.
- Type everything in Vue (`defineProps`, `defineEmits`, `refs`, composables).
- Keep naming consistent and descriptive.
- Favor clarity and simplicity over cleverness.

#### Comments (for everything)
- Explain *why*, not *what*. Comments should clarify reasoning or intent.
- Write friendly, human-sounding comments—clear and approachable, not robotic.
- Avoid contractions in comments, but if you use them, make sure they have proper apostrophes.
- Use JSDoc style for exported functions and types to improve tooling support.
- Add `TODO` comments for known limitations or future improvements.

**Checklist**

### DIFF
--- a/.cursor/rules/tests.mdc
+++ b/.cursor/rules/tests.mdc
@@ -1,0 +1,29 @@
+---
+description: Write awesome tests
+globs: *.test.ts,*.spec.ts
+alwaysApply: false
+---
+# Write tests 
+
+## Setup
+
+- Most tests just use Vitest.
+- We don’t have globals enabled, so all tests files should import describe, it and expect from vitest.
+- All test files end with *.test.ts and match the name of the file that contains the code we’re testing.
+- We should try to only mock if it’s really necessary. Sometimes we can refactor the actual code to make mocking unnecessary.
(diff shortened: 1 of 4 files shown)


amritk (approved):
Good start lets try it

PR #5103, opened 2025-03-13
fix(api-reference): doesn’t switch the collection

amritk:
**Problem**

closes #5102

Currently, we have collection hard coded to [0] in a few places so we werent getting things like operation lists
image

**Solution**

With this PR we add proper multi collection(doc) support by using the slug from the selector as the uid.
image

To test just keep switching specs with the console open and see if there's any errors

**Checklist**

### DIFF
--- a/.changeset/curly-plants-bow.md
+++ b/.changeset/curly-plants-bow.md
@@ -1,0 +1,7 @@
+---
+'@scalar/api-reference': patch
+'@scalar/api-client': patch
+'@scalar/oas-utils': patch
+---
+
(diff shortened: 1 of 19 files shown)


hanspagel (changes requested):
This is awesome! I just found out the API client sidebar shows all collections (kind of expected for the code we have, but confusing for the user).

Screenshot 2025-03-14 at 09 39 21

amritk:
Nice catch, fixed

amritk:
Going to yolo merge as its currently busted

PR #5104, opened 2025-03-14
feat(react): use the new JS API

amritk:
This moves over react to use the new JS api and enables multi config. Currently its pointing at the multi doc PR as that needs to go in first.

**Checklist**


hanspagel (approved):
RAD

PR #5106, opened 2025-03-14
fix(components): improve ScalarCombobox screen reader a11y

hwkr:
Substantially improves screen reader accessibility for the ScalarCombobox. I had thought that I had added the aria attributes back when I built the component but I guess not 😅

Before / After with NVDA:

...

**Checklist**


antlio (approved):
great work! guess we could ff with an empty search state message eventually?

hwkr:
> great work! guess we could ff with an empty search state message eventually?

Yeah I wasn't sure about adding one because W3 APG, Headless UI and Linear don't use one. AriaKit does have one though 🤔

PR #5107, opened 2025-03-14
fix(themes): hide summary caret in reset

hwkr:
We have an issue with a caret showing up in Safari on `<summary>` elements. It seems like we had a more specific fix in our markdown code but this moves it to the `reset.css`.

Safari-2025-03-13-22-34-56@2x

**Checklist**


hanspagel (approved):
cool

PR #5111, opened 2025-03-14
fix(api-reference): style fixtures

antlio:
**Problem**

currently introduction badges are lacking gap when multiple and show more button is not optically centered.

**Solution**

this pr adds gap to introduction badges:
| before | after |
| -------|------|
| image | image |

optically correct show more content position and update style:
| before | after |
| -------|------|
| image | image |

**Checklist**


amritk (approved):
Alt Text

looks the same to me!

PR #5115, opened 2025-03-14
fix(api-client): focus the client modal sidebar button when modal opens

hwkr:
Turns out this wasn't working in the references only in the playground. Now it works in the references.

https://github.com/user-attachments/assets/d1dcc864-482e-4b4c-982a-b851d0662014

**Checklist**


antlio:
might lack of context but would it make sense to instead focus on submit button?

PR #5118, opened 2025-03-14
fix(api-reference): booleans incorrectly allowing null

amritk:
**Problem**

Currently, boolean values can be null regardless of the spec

**Solution**

With this PR we only allow null when nullable is true. 

Also bonus fix added back in a lost request body description

**Checklist**


marclave (approved):
✨

PR #5119, opened 2025-03-15
chore(api-client): add tests for create client

amritk:
**Problem**

Currently, we have no testing on create-client.

**Solution**

With this PR we add  basic testing to create-client. This is just part one of a series to improve live store updates by using sync tech. Currently we just destroy the store everytime but that will soon change

**Checklist**


marclave (approved):
nice 😎

PR #5124, opened 2025-03-15
feat(api-reference-react): re-export config types from the react integration + fix ci

amritk:
**Problem**

Currently, we only export the props for the component

**Solution**

With this PR we export a few of the key config types for ease of use.

Also fixed the ci cuz its broke

**Checklist**


amritk:
yolo merging

Issue #5127, opened 2025-03-16
ClientId is no longer prefilled


xC0dex:
### What happens?

The following configuration no longer prefills the `clientId` for OAuth Flows:

```json
 "authentication": {
    "preferredSecurityScheme": "OAuth",
    "oAuth2": {
      "clientId": "my-client-id"
    }
  }
```

The related security scheme:

```json
"securitySchemes": {
      "OAuth": {
        "type": "oauth2",
        "flows": {
          "clientCredentials": {
            "tokenUrl": "https://your-authorization-server.com/connect/token",
            "scopes": { }
          }
        }
      }
    }
```

![Image](https://github.com/user-attachments/assets/1f463e35-3b9f-4979-9e98-efd7b0f08694)

### What did you expect to happen?

I'd expect that `my-client-id` is used as clientId by default,

Wolf-Syndrome:
This was raised from a discord thread. I use aspnet but I assume this is happens in other clients too, the main cause is the 
`packages\oas-utils\src\transforms\import-spec.ts`::`securitySchemes`
```
if (authentication.oAuth2.scopes) {
     flow.selectedScopes = authentication.oAuth2.scopes
     flow['x-scalar-client-id'] = authentication.oAuth2.clientId
   }
```
If you define the clientId but have not defined scope, it will not display. 
A temporary work around is to define a list of empty scopes.

xC0dex:
Fixed with https://github.com/scalar/scalar/pull/5128

PR #5128, opened 2025-03-16
fix: Fix client not being set without scope

Wolf-Syndrome:
### Problem
In response to: #5127 

When scope is missing from authentication -> oauth, client id will not be set.
E.g. since the following doesn't have a `oauth.scopes`, therefore it will not set the clientid
```
{
    options.WithOAuth2Authentication(oauth =>
    {
        oauth.ClientId = "my-client-id";
    });
});
 ``` 
image

### Solution 
Ensure the client id is set regardless of whether or not scopes is set.

### CODE REVIEW

**Comment by amritk** [comment]
ah nice find! Hilarious mistake. Could you just add a changeset and we'll get this in? `pnpm changeset`

File: packages/oas-utils/src/transforms/import-spec.ts [hunk 1]
```diff
@@ -206,14 +206,14 @@ export async function importSpecToWorkspace(
               flow.token = authentication.oAuth2.accessToken
             }
 
+            if (authentication.oAuth2.clientId) flow['x-scalar-client-id'] = authentication.oAuth2.clientId
+
+            if (authentication.oAuth2.scopes) flow.selectedScopes = authentication.oAuth2.scopes
```
amritk, line 211:
```suggestion
            if (authentication.oAuth2.clientId) {
              flow['x-scalar-client-id'] = authentication.oAuth2.clientId
            }

            if (authentication.oAuth2.scopes) {
              flow.selectedScopes = authentication.oAuth2.scopes
            }
```

amritk (approved):
Nice work!

hanspagel (approved):
thanks for taking the time, let’s get this in!

Wolf-Syndrome:
Thanks for you help.

PR #5130, opened 2025-03-17
chore(types): new authentication config types

### LINKED ISSUE #3887: Can't fill a single Auth schema if there are multiple with same `type`
### What happens?

I'm using the scalar ui from my angular application and i want to show and autofill api tokens for an openapi spec that has two mandatory authorization schemas. Both of them are of type 'apiKey' but with different names. Id like to autofill just one of them and keep the other one empty, but im only able to fill both of them or none. How do I configure it?

example schema:
```
securitySchema:
 X-Domain-ApiKey:
  in: header
  name: apikey1
  type: apiKey
 X-Domain-Host:
  in: header
  name: apikey2
  type: apiKey
```

example configuration:
```
authentication: {
 apiKey: {
 token: "xyz"
 }
}
```

I expected to fill just one of them, and not both of them

### OpenAPI Document

_No response_

amritk:
WIP: still messing around to see exactly what we even need

**Problem**

closes #3887

Our previous authentication config only allows for setting some values for auth by type. If you have more than one type you cannot set the individual values.

**Solution**

With this PR we lay the groundwork for authentication v2 which basically allows setting any value we support.
In this PR we:
- moved security scheme entitiy to the types package (later we may move the rest here but we only need security scheme for now)
- moved some helpers that the scheme used to types
- fixed the types in the migration so they won't break in the next upgrade
- added a clean function which removes default and catches from a schema, we only 
- added a new authentication schema which is basically `{ name: securityScheme }`
- added a transform from the old auth config to the new one

**Checklist**


amritk:
closed in favor of 

https://github.com/scalar/scalar/pull/5236

PR #5132, opened 2025-03-17
feat(client.scalar.com): add more meta tags

hanspagel:
This PR adds just a few more meta tags to https://client.scalar.com. 

Not necessary, but nice to have. :)


hanspagel:
Doesn’t hurt, let’s skip the queue.

PR #5153, opened 2025-03-17
ci: test integrations and packages in parallel

hanspagel:
lol, our vitest.workspace.ts in root just had `packages/*`, not `integrations/*`, which means those tests weren’t executed in CI.

I just found out, because I had the idea to split the test job into multiple jobs to shave some CI time off.

Soo, what this PR is doing:

* adding `integrations/*` and `projects/*` (doesn’t have tests yet, though) to the Vitest workspace configuration
* splitting the tests in three jobs: packages, integrations, proxy-server.

I don’t expect this to speed up CI, though. The 78 tests for integrations weren’t executed before, so it’s not really saving time to run them in parallel. And the proxy server tests are really fast, so no win expected here, too.

Anyway, with that PR we’re executing 78 more tests and splitting the tests up feels better, too.

Screenshot 2025-03-17 at 11 49 10


amritk (approved):
Nice! We can toss our e2e tests here as well

PR #5154, opened 2025-03-17
chore: use pnpm catalog

hanspagel:
**Problem**

Currently, we’re having the same dependencies in many packages and I think it’s our package formatting script that’s ensuring we have the same version anywhere. This works OK, but every time we update a dependency we have a big changeset for all the package.json’s.

**Solution**

With this PR we’re using the pnpm 10 catalog feature, where we specify the version for heavily used dependencies in the pnpm workspace configuration. This should make update PRs a bit easier to check and avoid the conflicts we have for those PRs all the time.

https://pnpm.io/catalogs

**Checklist**


amritk (approved):
Sick!

PR #5155, opened 2025-03-17
chore(api-reference): remove wildcard exports

hanspagel:
**Problem**

Currently, we’re doing this in the root index of `@scalar/api-reference`:

```ts
export * from '@/stores'
export * from '@/helpers'
export * from '@/types'
export * from '@/hooks'
```

This exposes all stores, helpers, types and hooks from the package. But we want to be able to move them around, rename them, drop them or change their API. Also, tree-shaking doesn’t work well in all environments.

**Solution**

Let’s trim down the list of exports to the stuff that we’re actually importing somewhere. Ideally, we’d move most of the remaining exports to other entrypoints, but that’s for a separate PR.

**Checklist**


amritk (approved):
Yes get rid of all the *!

marclave (approved):

PR #5156, opened 2025-03-17
feat(api-reference): Vue SSR playground

hanspagel:
**Problem**

Currently, we don’t properly test SSR environments, but we want to make the SSR support phenomenal.

**Solution**

With this PR we finally got a Vue SSR playground for `@scalar/api-reference`.

And we’re fixing the recently added `useMultipleDocuments` composable for SSR environments.

**Issues**

* It’s working with the `<ApiReference />` component, but not with the `createApiReference` method.
* The CSS isn’t preloaded, at least not in dev.
* We should have a few basic tests for it.
* We still have a bunch of hydration mismatches, so we might want to fix this here or in a separate PR, but it’s running!
Screenshot 2025-03-17 at 13 04 45

**Checklist**


hanspagel:
This is just a playground and it works okay-ish with the component already. Let’s merge it and take it from there.

PR #5157, opened 2025-03-17
docs: Update link in changelog

xC0dex:
I noticed that I used a dead link into the changelog 👀. I updated the link to point to the correct file.


hanspagel (approved):
oops, thanks

PR #5161, opened 2025-03-17
fix(nest.js): incorrect type import in nest.js fastify playground

amritk:
**Problem**

Currently, our cloudbuild is borked

**Solution**

With this PR we import the app.service normally.

**Checklist**


hanspagel:
Ah, I wonder why this didn’t fail in CI, though?

PR #5162, opened 2025-03-17
fix(nest.js): incorrect type import in nest.js express playground 

amritk:
Same fix but for nest + express

**Checklist**


hanspagel (approved):
love it

PR #5163, opened 2025-03-17
refactor(api-reference): fetch documents outside the components

hanspagel:
WIP

**Issues**

- [ ] If we want to fetch documents outside the components, we need to refactor the multi document logic.
- [ ] Currently, we don't fetch document URLs at all
- [ ] Currently, you can’t use ApiReferenceLayout anymore

**Problem**

It feels wrong that useReactiveSpec fetches URLs, we want to move away from useReactiveSpec. Let’s move the fetching out.

**Solution**

With this PR …

**Checklist**

### DIFF
--- a/examples/web/src/pages/ApiReferencePage.vue
+++ b/examples/web/src/pages/ApiReferencePage.vue
@@ -54,11 +54,10 @@ watch(
   },
 )
 
+// TODO:
 const { parsedSpec } = useReactiveSpec({
   proxyUrl: () => configuration.proxyUrl ?? '',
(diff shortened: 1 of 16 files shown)


amritk:
Can we move the fetching outside of the components completely and toss it into the html-api? That would be real sweet

hanspagel:
Yess, let’s do it

hanspagel:
Okok, we need to do this somehow. But it’s too much work for now. We’d need to refactor the multi document code a lot to make it work.

PR #5164, opened 2025-03-17
docs(Scalar.AspNetCore): more detailed examples and explanations

xC0dex:
We had many questions regarding about .NET, OpenAPI and Scalar. I hope that these docs help to understand everything a bit better. My plan is to gradually enhance and improve these documentation/examples.

Currently, the docs are located here `/integrations/aspnetcore/docs`. Do you know a better place for multiple related doc files?

This documentation contains:

- Scalar.AspNetCore
- Authentication
- Multiple OpenAPI Documents
- Multiple Scalar API References
- SubPath Deployment

Later, I'd like to

- enhance Authentication → Even more examples.
- add Metadata → Describing endpoints (summaries, comments etc.).
- improve the existing integration documentation by adding docs for **all** configuration options.

**Checklist**

I’ve gone through the following:

- [x] I’ve added an explanation _why_ this change is needed.
- [ ] I’ve added a changeset (`pnpm changeset`).
- [ ] I’ve added tests for the regression or new feature.
- [x] I’ve updated the documentation.

PR #5165, opened 2025-03-17
feat(Scalar.AspNetCore): Dynamic baseServerUrl

xC0dex:
**Problem**

Currently, if a .NET API is deployed under a sub path, the devs have to update the OpenAPI server 

**Solution**

With this PR, I add the missing `baseServerURL` configuration to the `ScalarOptions`. In addition, I introduce a new property `DynamicBaseServerUrl` which dynamically updates the `baseServerURL`  (in JS) based on the request. So the user doesn't have to configure anything to get the API and Scalar running under a sub path. 

**TODO**

- Update the docs (maybe after https://github.com/scalar/scalar/pull/5164 was merged)

**Checklist**


hanspagel (approved):
Let’s gooo

PR #5166, opened 2025-03-17
feat(api-reference): expose isLoading in config & "fix" config reactivity

amritk:
**Problem**

Currently, if you change the servers in the reference config, it will not get updated as it is not active.

**Solution**

With this PR we don't make it reactive, but instead do a quick fix of just re-creating the store whenever anything changes. This is much slower, but should be ok for now. If we run into performance issues we can bump up priority of using sync for the config.

Also exposed the `isLoading` prop to control loading of references
image

**Checklist**

### DIFF
--- a/.changeset/soft-pets-look.md
+++ b/.changeset/soft-pets-look.md
@@ -1,0 +1,6 @@
+---
+'@scalar/api-reference': patch
+'@scalar/types': patch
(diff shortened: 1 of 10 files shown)


hanspagel (approved):
ok, cool

PR #5172, opened 2025-03-18
feat(api-reference): explicitly set the default source

### LINKED ISSUE #5171: Add ability to set primary API configuration, when use multiple OpenAPI documents
### What happens?

Thank you very much for adding support for multiple API documents. Now Scalar can really replace Swagger UI.

But you missed one small feature: when you have multiple API documents, in Swagger UI you can specify the main document that should be opened by default. For example, you have several API versions, for example, several releases, the latest stable and snapshot (the most recent by release time). So, the current API allows you to open the desired API by default - the first one in the list. BUT! I need the latest stable API to open by default, but the snapshot to be the top one in the API selection list. In the current implementation, this cannot be done, unlike in Swagger UI.

My suggestion: add the primary flag to the configuration, which will indicate that this API should be opened by default if the name of the API to be opened is not specified

Need to add configuration flag for multiple openAPi documents like this:
```json
"primary": true
```

### OpenAPI Document

_No response_

hanspagel:
**Problem**

Currently, if you configure multiple OpenAPI documents, the first one will be the default. In some cases, this list is auto-generated, though and you might want - let’s say - v5 to be the default and not v1 (which might be the first entry).

**Solution**

With this PR, we’re looking for a `default: true` attribute to determine the default source. :)

```js
// Multiple documents
{
  sources: [
    { url: '/v1' },
    { url: '/v2' },
    { url: '/v3', default: true },
  ]
}
```

Or with multiple configurations:

```js
// Multiple configurations
[
  { url: '/v1' },
  { url: '/v2' },
  { url: '/v3', default: true },
]
```

Fixes #5171

**Checklist**


amritk:
So this only works for sources but not multiple configs?

hanspagel:
Yeah, I just assume if you have individual configs it might not be auto generated and it’s not necessary to explicitly pick one.

But if someone asks for it, we might add a default: true as a top level attribute?

wdyt?

altro3:
@hanspagel What difference between multiple sources and multiple configs? Swagger UI has ability to set only multiple URI's of openapi documents (like sources here, as I understnad). But I don't unserstand, in which case you need multiple configurations and what differences with multiple sources

hanspagel:
@altro3 What is Swagger UI? 😋

We support multiple documents, but also multiple configurations. Some examples where that might make sense:

* Prefilling authentication in different ways
* Overwriting server URLs in different ways
* Having an alternative color scheme for a specific API

altro3:
Okay, I get it. Then it would indeed be nice to be able to set a default configuration.

hanspagel:
Done! You can set a default configuration now, too.

altro3:
Thanks! After next release I'll integrate mutli documents feature with https://github.com/micronaut-projects/micronaut-openapi

amritk (approved):
Very good sir

hanspagel (commented):
code looks good

hanspagel:
works great!

Issue #5174, opened 2025-03-18
callbacks


hanspagel:
Let’s add callbacks:

**Specification**

https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.1.md#callback-object

**Example**

https://learn.openapis.org/examples/v3.0/callback-example.html

Krasav4k1:
Hi @hanspagel 
Do you have any ETA when Callback feature will be done, I see that you've added already the code example of Callback into `planet schema` but it still doesn't render, do you have any roadmap of releasing callback UI?

Krasav4k1:
@hanspagel do you have any updates on this?

LenKlose:
+1

hanspagel:
Wait! We added callbacks just recently :)

https://galaxy.scalar.com/#tag/planets/post/planets

<img width="524" height="216" alt="Image" src="https://github.com/user-attachments/assets/f1892031-d5ff-4dbc-93f2-1c207526ccd8" />

PR #5185, opened 2025-03-19
feat(galaxy): more of everything

hanspagel:
I got side tracked, and did a bigger update to Scalar Galaxy.

* added more data to the planet schema
* split the user schema in user and credentials again (cosmetic change)
* fixed the invalid examples for the image upload
* added a callback (see #5174, not rendered yet)
* add an externalDocs link (not rendered yet)
* added some missing responses

new minor version

### DIFF
--- a/.changeset/brown-elephants-burn.md
+++ b/.changeset/brown-elephants-burn.md
@@ -1,0 +1,5 @@
+---
+'@scalar/galaxy': minor
+---
+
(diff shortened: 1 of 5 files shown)


amritk (approved):
Sweet! Would be nice if we can get one of everything in it eventually, maybe we can ask claude nicely

PR #5186, opened 2025-03-19
fix(openapi-parser): Swagger 2.0 `basePath` is ignored when there’s no `host`

### LINKED ISSUE #5126: base url is missing
### What happens?

I have following docs, but in the client, it will not include the base path, unless I specify schema and server. but the server url is different base on the deployment. it would be spend too much time manually change the api is generated with `swag` or updated it with command.
```json
{
    "swagger": "2.0",
    "info": {
        "description": "This is the backend API for the project",
        "title": "Backend API",
        "contact": {},
        "version": "1.0"
    },
    "basePath": "/api"
// content omit here.
}
```

I found a similar issue #2806 but it is kind of different for my case

It should include the base path after the url.

### OpenAPI Document

_No response_

hanspagel:
**Problem**

Currently, we just ditch the `basePath` of a Swagger 2.0 document when there’s no `host`. That’s so wrong.

**Solution**

We’re keeping the lovely `basePath` with this PR here, isn’t it wonderful?

Fixes #5126 

**Checklist**

### DIFF
--- a/.changeset/swift-mirrors-battle.md
+++ b/.changeset/swift-mirrors-battle.md
@@ -1,0 +1,5 @@
+---
+'@scalar/openapi-parser': patch
+---
+
+fix: swagger 2.0 basePath is ignored, if there’s no host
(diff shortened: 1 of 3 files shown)


amritk (approved):
Noice

PR #5190, opened 2025-03-19
fix(api-client): don’t send authorization header for public client

jwangnz:
AWS Cognito will return 400 when authorization header is provided.

**Problem**

When using scalar with AWS Cognito for public oauth2 client, we got 400 when exchanging the code.

**Solution**

With this PR, it won't send the Authorization header when it is a public client.

**Checklist**

### DIFF
--- a/packages/api-client/src/views/Request/libs/oauth2.ts
+++ b/packages/api-client/src/views/Request/libs/oauth2.ts
@@ -260,7 +260,9 @@ export const authorizeServers = async (
     }
 
     // Add client id + secret to headers
-    headers.Authorization = `Basic ${btoa(`${flow['x-scalar-client-id']}:${flow.clientSecret}`)}`
+    if (flow.clientSecret) {
+      headers.Authorization = `Basic ${btoa(`${flow['x-scalar-client-id']}:${flow.clientSecret}`)}`
+    }
 
     // Check if we should use the proxy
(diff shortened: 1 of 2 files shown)


packages/api-client/src/views/Request/libs/oauth2.ts
```diff
@@ -260,7 +260,9 @@ export const authorizeServers = async (
     }
 
     // Add client id + secret to headers
-    headers.Authorization = `Basic ${btoa(`${flow['x-scalar-client-id']}:${flow.clientSecret}`)}`
+    if (flow.clientSecret) {
```
amritk, line 263: should we check for clientId as well?

jwangnz: @amritk Thank you for the review. While I'm not familiar with this codebase, I discovered this fix works for my use case.

Regarding clientId: it appears this will always be present, as line 222 unconditionally sets client_id in formData from `flow['x-scalar-client-id']`. This contrasts with line 230, which checks if flow.clientSecret exists before adding it to formData. 

If my understanding is incorrect, please let me know and I'll happily update the condition.

amritk: oh sure sorry I didn't look at the rest of the file, guess I missed the context

amritk (approved):

PR #5194, opened 2025-03-19
fix(api-reference): reduce some jank on multi doc

amritk:
**Problem**

Currently, when switching spec's the window was jumping all over the place, this was due to some remaining global state we have. 

**Solution**

With this PR we provide/inject the state from useNavState with defaults (so it still works in ORG) and we clear it every time we change documents (which should reduce the jank)

There may be some remaining in useSidebar but lets try this first

**Checklist**


hanspagel (approved):
a man has to do what a man has to do

PR #5202, opened 2025-03-20
feat(api-reference): add callback `onDocumentSelect`

amritk:
**Solution**

With this PR we add a callback which gets triggered when you select a version

**Checklist**


hanspagel (approved):
Cool! I renamed it to to `onDocumentSelect`, because it’s triggered when you select a document in the document selector. 

Though the list of documents can come from different configurations, it doesn’t have to come from different configurations.

PR #5203, opened 2025-03-20
feat(api-reference): "Freeze" elements on page when switching clients

### LINKED ISSUE #5196: Changing clients causes change of current method or triggers scroll
### What happens?

Steps to reproduce:
1. Visit https://galaxy.scalar.com/#tag/planets/PUT/planets/{planetId}
2. In the request preview window change client to any other one

Results in :
- ~~page scrolling up/down~~ not scrolling, the items are shrinking/expanding
- sometimes changes the current focused method, for example from DELETE to PUT (path changes)

It should not trigger scroll.

### OpenAPI Document

https://galaxy.scalar.com/openapi.yaml

amritk:
**Problem**

Currently, when you change clients, all of snippets change size and the content shifts

closes #5196

**Solution**

With this PR we "Freeze" the current element such that it doesn't move when you switch clients.

We can use this tech to improve lazy loading as well!

TO test:
- change clients and the content should not move!

**Checklist**

### DIFF
--- a/.changeset/cool-points-love.md
+++ b/.changeset/cool-points-love.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
+feat: freeze element in window when switching clients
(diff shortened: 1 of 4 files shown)


hanspagel (approved):
big brain PR

Issue #5199, opened 2025-03-20
express integration broken after refactor from #4795


gizero:
### What happens?

Removing braces with [this change](https://github.com/scalar/scalar/commit/89d8b7504759c5bb3eae687c49b3363e8890828b#diff-e48952aee58806120778d949837355e99e3da2c19b98c0d7e4814d1446be68a1R123) as part of https://github.com/scalar/scalar/pull/4795 caused the value returned by the express middleware to change from `void`/`undefined` to whatever `res.type().send()` will return (arrow function implied `return`). This in turn broke the interface for an express middleware which is expected to have a `void` return type. With Typescript this led to raising the following error at compile time, even with the very minimal example from [README.md](https://github.com/scalar/scalar/tree/main/integrations/express#usage):

```
error TS2769: No overload matches this call.
  The last overload gave the following error.
    Argument of type '(_: Request<ParamsDictionary, any, any, ParsedQs, Record<string, any>>, res: Response<any, Record<string, any>>) => Response<...>' is not assignable to parameter of type 'Application<Record<string, any>>'.
      Type '(_: Request<ParamsDictionary, any, any, ParsedQs, Record<string, any>>, res: Response<any, Record<string, any>>) => Response<...>' is missing the following properties from type 'Application<Record<string, any>>': init, defaultConfigu
ration, engine, set, and 63 more.

 59       apiReference({
          ~~~~~~~~~~~~~~
 60         spec: {
    ~~~~~~~~~~~~~~~
...
 62         }
    ~~~~~~~~~
 63       })
    ~~~~~~~~

  node_modules/@types/express-serve-static-core/index.d.ts:164:5
    164     (path: PathParams, subApplication: Application): T;
            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    The last overload is declared here.
 ```

If it sound like the appropriate approach, I can take care to open a PR to revert that specific chunk of https://github.com/scalar/scalar/pull/4795 's changeset.

### What did you expect to happen?

I'd expect the value returned by `apiReference()` from `@scalar/express-api-reference` to be a type-valid express.js middleware.

gizero:
[...]

PR #5222, opened 2025-03-23
fix(express): @types/express@5 TypeScript error

gizero:
fixes https://github.com/scalar/scalar/issues/5199.

Commit 89d8b7504759c5bb3eae687c49b3363e8890828b as part of [1], incorrectly changed the return value type for express middleware function returned by apiReference(). Removing braces from the arrow function implementation turned its body type from "block" to "expression". In an expression body, only a single expression is specified, which becomes the implicit return value.

This commit reverts that change to restore the original return type which was consistent with express's guidelines.

Although the above change did not impact runtime execution of the middleware (express does not care about the return value type at all), it did trigger a build error on Typescript projects using @types/express 5.x for type validation.

[1] https://github.com/scalar/scalar/pull/4795

**Problem**

Currently, …

**Solution**

With this PR …

**Checklist**

### DIFF
--- a/.changeset/green-panthers-change.md
+++ b/.changeset/green-panthers-change.md
@@ -1,0 +1,5 @@
+---
+'@scalar/express-api-reference': patch
+---
+
+fix(express): restore express middleware return value type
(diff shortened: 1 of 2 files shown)


hanspagel:
Thanks for putting together this fix! Great work

PR #5224, opened 2025-03-24
fix(scalar-app): toDesktop can’t install pnpm catalog dependencies

hanspagel:
**Problem**

We moved a few dependencies to the pnpm workspace catalog, and while we use pnpm@10 in the toDesktop cloud, we don’t push the whole monorepository there, but only the scalar-app/ folder. This breaks the build.

**Solution**

With this PR we’re ignoring the pnpm catalog dependencies in the toDesktop build cloud.

**Checklist**

### DIFF
--- a/.changeset/tricky-pots-cry.md
+++ b/.changeset/tricky-pots-cry.md
@@ -1,0 +1,5 @@
+---
+'scalar-app': patch
+---
+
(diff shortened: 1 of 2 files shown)


amritk (approved):

hanspagel:
Tested, and the toDesktop build passes again. ✌️

PR #5230, opened 2025-03-26
feat(types): add security schemes

amritk:
**Solution**

With this PR we move security schemes into the types package. Just so we can the type for auth config v2.

Also has various type fixes I stumbled upon while working on auth config v2.

**Checklist**


packages/types/src/utils/nanoid.ts
```diff
@@ -0,0 +1,25 @@
+import { nanoid } from 'nanoid'
+import { z } from 'zod'
+
+/** Generates a default value */
+export const nanoidSchema = z
+  .string()
+  .min(7)
+  .optional()
+  .default(() => nanoid())
```
hanspagel, line 9: there’s also `z.nanoid()`, would that work, too?

packages/types/tsconfig.json
```diff
@@ -2,6 +2,7 @@
   "extends": "../../tsconfig.node.json",
   "include": ["src", "test"],
   "compilerOptions": {
+    "rootDir": "./src",
```
hanspagel, line 5: ahh, everytime I’ve had to add this, something else was actually off. what’s wrong if we remove that?

packages/api-reference/src/hooks/useMultipleDocuments.ts
```diff
@@ -141,18 +140,18 @@ export const useMultipleDocuments = ({
   const selectedConfiguration = computed(() => {
     // Multiple sources
     if (configuration.value && isConfigurationWithSources(configuration.value)) {
-      return apiReferenceConfigurationSchema.parse({
+      return {
         ...configuration.value,
         ...configuration.value?.sources?.[selectedDocumentIndex.value],
         ...availableDocuments.value[selectedDocumentIndex.value],
-      })
+      }
```
hanspagel, line 147: why do we need this change?

hanspagel (dismissed):
omg, great work! I’m glad I didn’t touch the security scheme schemas in my PR yet.

just a few questions, but we need to make CI pass

packages/types/src/utils/nanoid.ts
amritk: yea lets try it!

packages/types/tsconfig.json
amritk: really? I thought this puts all the types in the correct folders?

packages/api-reference/src/hooks/useMultipleDocuments.ts
amritk: we don't want to parse here, we want to only parse in one place so we currently do it in the apireferencelayout (just because of org) but eventually we will bring it back out here. Otherwise we get defaults and what not

amritk: Now I can't remember exactly what issue it was solving but there was one!

amritk: I can remove this change for now and we can bring it back if we see it, this is what its breaking the tests

packages/types/src/utils/nanoid.ts
amritk: alright I looked into it. Found two issues:
- we don't always use a nanoid so things like `default` will fail validation
- I couldn't find a way to generate a default using only zod

we can re-visit it in the future if we move away from the default uid

packages/types/tsconfig.json
amritk: removed it lets see

hanspagel (approved):
nicee

I’ll move the security schemes to the openapi-types package probably, that’s where I put all the other OpenAPI Zod schemas

hanspagel (approved):

PR #5236, opened 2025-03-26
feat(api-reference): preconfigure multiple authentication methods 🔐🔐

### LINKED ISSUE #3887: Can't fill a single Auth schema if there are multiple with same `type`

amritk:
**Problem**

closes #3887

Currently, we can only set one type of auth at a type

**Solution**

With this PR we can set multiple auths of the same type + we support more oauth2 specs and any auth we add in the future will be supported.

This is much better than the previous attempt because it doesn't have that crazy zod deep partial function. It does not even need a transform as it just supports the old auth (until we no longer want to)

**Checklist**

### DIFF
--- a/.changeset/small-poems-hope.md
+++ b/.changeset/small-poems-hope.md
@@ -1,0 +1,6 @@
+---
+'@scalar/api-reference': patch
+'@scalar/types': patch
+---
+
+feat: added new auth config (v2)
--- a/documentation/configuration.md
+++ b/documentation/configuration.md
@@ -4,10 +4,10 @@ You can pass a — what we call — universal configuration object to fine-tune 
 
 ## Universal Configuration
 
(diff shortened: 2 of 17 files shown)


hanspagel (approved):
OMG, PR of the week. Tested it, and it works so great.

PR #5240, opened 2025-03-27
fix(nestjs): configuration requires too many attributes

### LINKED ISSUE #5228: nestjs doesn’t use partial config type
https://discord.com/channels/1135330207960678410/1135330208627564597/1353228028813115483
> hans I think its just using the wrong type! We need to update it to use Partial<ApiReferenceConfiguration> now which should make it all optional

hanspagel:
**Problem**

When using the NestJS integration, we expect too many attributes. We forgot to make the type Partial.

**Solution**

With this PR we add `Partial<>` and now you can freely pass any configuration while still having a nice auto-complete.

Fixes #5228

**Checklist**

### DIFF
--- a/.changeset/brown-owls-brush.md
+++ b/.changeset/brown-owls-brush.md
@@ -1,0 +1,5 @@
+---
+'@scalar/nestjs-api-reference': patch
+---
+
+fix: configuration has too many required attributes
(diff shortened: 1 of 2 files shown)


marclave (approved):
nice @hanspagel

PR #5241, opened 2025-03-27
chore(oas-utils): upgrade tsconfig in oas-utils to point to the strict node one

amritk:
**Problem**

Currently, I was having an issue where I wasn't receiving the correct types inside of import-spec.

**Solution**

- Upgraded to the strict node tsconfig
- fixed the type issues in import spec
- wrote a script to rename the files to match our style guide since we had to change the extensions anyway
- got rid of export * and moved to explicit exports for better tree shaking
-  tests all passing ✅

**Checklist**


packages/oas-utils/src/diff/diff.test.ts
hanspagel, line 1: Moved or deleted?

hanspagel (approved):
you're the GOAT

PR #5242, opened 2025-03-27
feat(api-reference): copy example/examples  to clipboard

cameronrohani:
we currently don't have a copy to clipboard button for examples + it gets overwhelming when you have too many examples, so this aims to fix that :) 

I also updated an example in scalar galaxy that have multiple examples

before:
image

after:
image

### DIFF
--- a/packages/api-reference/src/components/Content/Schema/SchemaPropertyHeading.vue
+++ b/packages/api-reference/src/components/Content/Schema/SchemaPropertyHeading.vue
@@ -184,12 +184,14 @@ const { copyToClipboard } = useClipboard()
           class="property-example-value group"
           @click="
             copyToClipboard(
-              formatExample(
-                value?.example ||
-                  (discriminatorType &&
-                    value?.items &&
-                    typeof value.items === 'object' &&
-                    value.items[discriminatorType]),
+              String(
+                formatExample(
+                  value?.example ||
+                    (discriminatorType &&
+                      value?.items &&
+                      typeof value.items === 'object' &&
+                      value.items[discriminatorType]),
+                ),
               ),
             )
           ">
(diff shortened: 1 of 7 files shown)


packages/api-reference/src/components/Content/Schema/SchemaPropertyHeading.vue
```diff
@@ -1,28 +1,50 @@
...
   },
 )
 
+const discriminatorType = discriminators.find((r) => {
+  if (!props.value || typeof props.value !== 'object') {
+    return false
+  }
+
+  return (
+    r in props.value ||
```
amritk, line 34: wow cam's fancy

amritk (approved):
Fixed a minor issue, looks good to me now

packages/api-reference/src/components/Content/Schema/SchemaPropertyHeading.vue
cameronrohani: **c**l**a**ude **ron**

amritk: Your secrets safe with me 😉

hanspagel (changes requested):
OMG, nice. I’ve moved it to a separate component.

I removed this code and just passed the example:

```diff
formatExample(
  optimizedValue?.example ||
-    (discriminatorType &&
-      optimizedValue?.items &&
-      typeof optimizedValue.items === 'object' &&
-      optimizedValue.items[discriminatorType]),
)
```

I don’t understand why we’d pass the items to the `formatExample` function.

But we still have a z-index issue it seems.

Screenshot 2025-03-28 at 13 38 06

cameronrohani:
should be good to merge now :)

hanspagel (approved):
good to merge now! awesome work

PR #5243, opened 2025-03-27
fix(api-reference): allow promise returns in onDocumentSelect

amritk:
**Problem**

Currently, we can only use sync functions as `onDocumentSelect`

**Solution**

- allow using async functions aka returning promises
- useReactiveSpec not clearing spec between document changes, switched to a ref from reactive for this
- ensure interaction observer is always disabled before opening up sections

**Checklist**


hanspagel (approved):
nice qol update

PR #5244, opened 2025-03-27
style(api-client): add class names for targetting client

cameronrohani:
I need to target the api client via CSS classes :) 

this adds classes to everything I'll need to target


hanspagel (changes requested):
if we _need_ them, we should add tests for them, I’ll take this one over ✌️

hanspagel (approved):
* changed the one class to say "code-example" not "code-snippet"
* added tests
* removed the yooo class

let’s go

PR #5246, opened 2025-03-28
feat(api-reference): multiple configurations with multiple sources

hanspagel:
**Problem**

We support multiple configurations and a single configuration with multiple sources, but can you image: we don’t support multiple configurations *with* multiple sources. Let’s tackle this.

**Solution**

This PR adds support for multiple configurations with multiple sources. 💥 

**Checklist**

### DIFF
--- a/packages/types/src/api-reference/api-reference-configuration.ts
+++ b/packages/types/src/api-reference/api-reference-configuration.ts
@@ -501,6 +501,7 @@ export type AnyApiReferenceConfiguration =
   | Partial<ApiReferenceConfiguration>
   | Partial<ApiReferenceConfigurationWithSources>
   | Partial<ApiReferenceConfigurationWithDefault>[]
+  | Partial<ApiReferenceConfigurationWithSources>[]
 
 /** Typeguard to check to narrow the configs to the one with sources */
 export const isConfigurationWithSources = (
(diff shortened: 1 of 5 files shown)


packages/types/src/api-reference/api-reference-configuration.ts
```diff
@@ -498,6 +498,7 @@ export type AnyApiReferenceConfiguration =
   | Partial<ApiReferenceConfiguration>
   | Partial<ApiReferenceConfigurationWithSources>
   | Partial<ApiReferenceConfigurationWithDefault>[]
+  | Partial<ApiReferenceConfigurationWithSources>[]
```
amritk, line 501:
```suggestion
  | Partial<ApiReferenceConfiguration>
  | Partial<ApiReferenceConfiguration>[]
```
should we just add sources and default as optional to the main schema?

amritk (commented):
Yea when we did this I wasn't sure if this was going to be a use case or not. But I suppose it could work?

packages/types/src/api-reference/api-reference-configuration.ts
hanspagel: I don‘t know, I think I’d prefer to be more explicit there. I’d actually love the configuration to be more explicit in a few details (like requiring content OR url, currently everything is optional).

I’d tend to leverage TypeScript more, even if it’s just a detail.

hanspagel:
Yeah, this feature seems to be perfect for our upcoming Aspire integration, that’s why @xC0dex requested it.

amritk (approved):
kk cleared up, send it

PR #5250, opened 2025-03-28
feat(api-client): add auth to the collection page

amritk:
**Problem**

Currently, in the client we can only set auth at the request level

**Solution**

With this PR we are able to set auth at the collection level. Also added a `useCollectionSecurity` flag to the collection so we can flick it on and off.

image

**Checklist**


hanspagel:
OMG, so cool! Worked on the copy a bit:

Screenshot 2025-03-31 at 12 22 16

hanspagel (approved):

PR #5260, opened 2025-03-31
fix(api-reference): proxy is skipped for localhost, even if the proxy is local

### LINKED ISSUE #5258: Make Proxy usable in local environments
### Discussed in https://github.com/scalar/scalar/discussions/5257

<sup>Originally posted by **xC0dex** March 30, 2025</sup>
Currently, the `proxyUrl` will not be used if the current URL is `localhost`. For the Aspire integration, it would be _awesome_ if this would also work for `localhost` URLs. 

**Context:**

Aspire starts multiple services with different ports. Example:

Book API: `https://localhost:55555`
User API:  `https://localhost:55556`
Scalar Aspire Integration: `https://localhost:55557`

Without a proxy, Scalar cannot load the OpenAPI documents or make requests to the APIs because they are running on a different Port. This will cause CORS issues.</div>

hanspagel:
**Problem**

Our API reference can’t fetch OpenAPI documents from other (local) domains, if CORS isn’t set up properly (because we can’t use the remote proxy then).

**Solution**

Initially, I thought we’re just missing something in the fetching logic, turns out: Not really.  BUT two changes:

* The @scalar/api-reference fetching logic is closer to what we do for API requests.
* If the proxyUrl is relative, we’ll use it even for local domains (fixes #5258)

I used the opportunity to streamline the fetching a bit, though:

* Renamed `fetchSpecFromUrl` to `fetchDocument` (we’re talking about documents, not about the standard/specification)
* Updated comments
* Removed the isValidUrl/isRelativeUrl/whatever logic from `useReactiveSpec` and just used `fetchDocument` (to make it work the same in all places)
* Added a few more tests
* Catched a few more exceptions to simplify the usage of the related helpers

**Checklist**

### DIFF
--- a/packages/oas-utils/src/helpers/redirect-to-proxy.ts
+++ b/packages/oas-utils/src/helpers/redirect-to-proxy.ts
@@ -1,25 +1,37 @@
 import { isLocalUrl } from './is-local-url.ts'
 import { REGEX } from './regex-helpers.ts'
 
-/** Redirects the request to a proxy server with a given URL. */
+/**
+ * Redirects the request to a proxy server with a given URL. But not for:
+ *
+ * - Relative URLs
+ * - URLs that seem to point to a local IP
+ * - URLs that don't look like a domain
+ **/
 export function redirectToProxy(proxy?: string, url?: string): string {
-  if (!shouldUseProxy(proxy, url)) {
+  try {
+    if (!shouldUseProxy(proxy, url)) {
+      return url ?? ''
+    }
+
+    // Create new URL object from url
+    const newUrl = new URL(url as string)
+
+    // Rewrite the URL with the proxy
+    newUrl.href = proxy as string
+
+    // Add the original URL as a query parameter
+    newUrl.searchParams.append('scalar_url', url as string)
+
+    return newUrl.toString()
+  } catch {
     return url ?? ''
   }
-
-  // Create new URL object from url
-  const newUrl = new URL(url as string)
-
-  // Rewrite the URL with the proxy
-  newUrl.href = proxy as string
-
-  // Add the original URL as a query parameter
-  newUrl.searchParams.append('scalar_url', url as string)
-
-  return newUrl.toString()
 }
 
-/** Check if the URL is relative or if it's a domain without protocol */
+/**
+ * Check if the URL is relative or if it's a domain without protocol
+ **/
 export const isRelativePath = (url: string) => {
   // Allow http:// https:// and other protocols such as file://
   if (REGEX.PROTOCOL.test(url)) {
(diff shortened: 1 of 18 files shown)


packages/oas-utils/src/helpers/redirect-to-proxy.ts
```diff
@@ -1,25 +1,37 @@
...
-  }
+  try {
+    if (!shouldUseProxy(proxy, url)) {
+      return url ?? ''
+    }
 
-  // Create new URL object from url
-  const newUrl = new URL(url as string)
+    // Create new URL object from url
+    const newUrl = new URL(url as string)
```
amritk, line 18:
```suggestion
    const newUrl = new URL(url)
```
why the type cast here, just do an early return above to check for undefined

amritk: actually we can turn `shouldUseProxy` into a type guard for proxy then just check for url

packages/oas-utils/src/helpers/redirect-to-proxy.ts
```diff
@@ -35,22 +47,28 @@ export const isRelativePath = (url: string) => {
...
 
-  // Relative URLs
-  if (isRelativePath(url)) {
-    return false
-  }
+    // Relative URLs
+    if (isRelativePath(url)) {
+      return false
+    }
+
```
amritk, line 64:
```suggestion
    }

  // Requests to a local proxy
  if (isLocalProxy(proxy) {
    return true
  }

```

amritk (changes requested):
So this doesn't actually fix the issue stated in #5258, I added a dummy function below

packages/oas-utils/src/helpers/redirect-to-proxy.ts
hanspagel: ahh, type guarding works for one parameter only, but still a great idea

packages/oas-utils/src/helpers/redirect-to-proxy.ts
amritk: Yea if its deployed then they should not pass in a localhost proxy. Its only for local dev

amritk: why can't we just allow the user to pass it in? If they pass in a localhost on a domain it won't work

xC0dex (approved):
LGTM. This will work for Aspire

hanspagel:
Waiting for @amritk to re-review here. It’s core functionality, so I don’t want to rush it.

hanspagel:
We won’t need this for the Aspire integration, and I don’t see the use case, but I’ve just added this, too. :v:

amritk:
Its come up a couple times, helpful for local testing!

amritk (approved):
I don't even need to re-test, code looks 👌🏾

PR #5266, opened 2025-03-31
fix(api-reference): empty search results on references

amritk:
**Problem**

Currently, we had a watcher on a value to generate search results. .value is not responsive.

**Solution**

With this PR we move the watcher to the ref instead so its reactive.

**Checklist**


hanspagel (approved):
Great find

PR #5270, opened 2025-04-01
feat(snippetz): new php/guzzle plugin

hanspagel:
**Problem**

We’re still using the httpsnippet-lite converter for PHP/Guzzle. But most of those converters lack features and we just want to get rid of them. 

**Solution**

I wrote PHP plugin yesterday anyway, so I just replaced this one, too. This PR adds our own PHP/Guzzle plugin to the snippetz package.

**Checklist**


amritk (approved):
Probably fine

Issue #5273, opened 2025-04-01
show the request details sent on UI


WeihanLi:
Could I inspect the raw request detail when sending a request or request sent

_Originally posted by @WeihanLi in https://github.com/scalar/scalar/discussions/3986_

Currently, we can not get the http request details on the UI, which is not easy to share the request with others, and when there's variable on the request path, it would be harder to share the request detail

hanspagel:
Yess, I love this idea!

hanspagel:
I’ve quickly made something. It shows the response headers *and* the request headers. But it has two issues:

1. It’s not all headers, the browser adds a few that we can’t access.
2. It’s only the headers, not the rest.

I think I like it though. Wdyt?

<img width="597" alt="Image" src="https://github.com/user-attachments/assets/c8992525-a284-4f0a-99f5-72b836335e31" />

spaasis:
In Swagger UI I've gotten used to being able to copy the request as cURL - do you think that would be reasonable to include here too?

antlio:
@spaasis greta feedback! do you mean as cURL including the request values vs the today generated code snippet?

antlio:
@WeihanLi thank you so much for the thorough feedback, we are working on this and will share more here asap!

antlio:
@spaasis what are you using as of today to generate your reference? do you seen currently the Code Snippet (as shared in the issue description) below your body section?

spaasis:
@antlio Oh sorry, I missed this tiny tool in the bottom left of the send dialog. The stuff I need is already implemented 🎉

![Image](https://github.com/user-attachments/assets/3dd971a4-0205-4250-92c7-a18b65909482)

antlio:
@spaasis we'll keep working to enhance this so feel free if you have any feedback toward this :) in a sharing context it might make sense to update its position at some point!

taoneill:
👍 on this request.

Following up on @spaasis, having the each URL {variable} get swapped in the Test Request dialog's cURL Code Snippet would make copy-pasting the actual Request URL easier. Currently, only the Query Parameters are replaced with their current test values in the snippet.

Going back to @WeihanLi's original request: being able to inspect the final Request URL (without having to dig into the Chrome Dev Tools > Network > Headers  > General section) would be a great convenience.

Thanks for your great work!

PR #5275, opened 2025-04-01
feat(api-client): show actual request headers

hanspagel:
**Problem**

Sometimes, I’d wish to see the actual request headers that were sent, or even the whole request (see #5273). But it’s only in the browser console (as of now).

**Solution**

With this PR we’re adding a “Request Headers” section to the response pane, so you can see the headers that are sent.

Just two things to note:

1. It’s just the headers, not the whole HTTP request yet.
2. It’s not showing the headers that the browser adds (we can’t access those).

**Checklist**

### DIFF
--- a/packages/api-client/src/libs/send-request/create-request-operation.ts
+++ b/packages/api-client/src/libs/send-request/create-request-operation.ts
@@ -30,6 +30,8 @@ type SendRequestResponse = Promise<
   }>
 >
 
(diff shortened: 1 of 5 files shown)


amritk (approved):
saweet

hanspagel:
Waiting for product approval from @cameronrohani or @marclave.

And maybe I’ll add tests, too.

marclave (approved):
LOVE it <3

nice work @hanspagel lets get this in

PR #5280, opened 2025-04-01
feat(openapi-types): more OpenAPI extensions

hanspagel:
**Problem**

We have Zod schemas for a few OpenAPI extensions already, but not even all that we use in the code base. 

**Solution**

This PR adds more of the OpenAPI extensions that we use in our code base.

**Checklist**


amritk (approved):
What's with the mix of camel case and kebab case, should we just support one and have transforms?

PR #5289, opened 2025-04-01
fix(api-reference): ensure pathRouting works on initial load

amritk:
**Problem**

Currently, if you refresh the page while using pathRouting it does not route to the correct spot.

**Solution**

It turns out config was undefined as it is provided in `ApiReferenceLayout` so if we use `useNavState` in that component we don't get config. So we pass in config here as we are computing it above.

**Checklist**


hanspagel (approved):
great explanation, great PR

PR #5291, opened 2025-04-01
fix(api-reference): prevent the store from crashing when we have exponential body properties

amritk:
**Problem**

Currently, we have a hard limit on depth for circular bodies, however nothing stops it from expanding horizontally.

**Solution**

With this PR we add a cap on properties after 3 levels deep to 10 (can be adjusted as we go) in the client request body.
image

Added a show additional properties button as well when it has 12 properties
image

**Checklist**


cameronrohani (approved):

hanspagel (approved):
OMG, awesome PR

PR #5292, opened 2025-04-02
fix(api-reference): Open in API Client button is missing

amritk:
**Problem**

Currently, the open in api client button is missing because we were still looking at the old spec.url

**Solution**

With this PR we also check for config.url for the `documentUrl` which brings the button back!

**Checklist**


hanspagel (approved):
🙈 thanks

PR #5297, opened 2025-04-02
feat(hono): rename export from `apiReference` to `Scalar`

hanspagel:
**Problem**

Currently, we’re exporting `apiReference` as the middleware for Hono.

But people refer to our solution as Scalar and when they [come from Swagger UI](https://hono.dev/examples/swagger-ui) it might be more familiar to replace swaggerUI with Scalar in your code.

**Solution**

With this PR we deprecate the ~~`apiReference`~~ export and add a `Scalar` export.

```ts
import { Hono } from 'hono'
import { Scalar } from '@scalar/hono-api-reference'

const app = new Hono()

// Use the middleware to serve the Scalar API Reference at /scalar
app.get('/scalar', Scalar({ url: '/doc' }))

export default app
```

**Checklist**

### DIFF
--- a/.changeset/flat-ladybugs-explain.md
+++ b/.changeset/flat-ladybugs-explain.md
@@ -1,0 +1,5 @@
+---
+'@scalar/hono-api-reference': patch
+---
+
+chore: deprecated apiReference export
--- a/.changeset/slow-icons-share.md
+++ b/.changeset/slow-icons-share.md
@@ -1,0 +1,5 @@
+---
+'@scalar/hono-api-reference': minor
+---
+
+feat: rename apiReference to scalar (please update your import)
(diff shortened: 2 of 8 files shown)


amritk:
I think we may still need something in there to differentiate from the scalar api client and whatever other tools we have

amritk:
cc: @marclave

hanspagel:
* Hono-style: `ScalarAPIReference`
* Our style: `ScalarApiReference`
* Or just keep it, actually Hono-style, too: `apiReference` (just not what people use to refer to our solution)

marclave (approved):
oooo nice @hanspagel 😎

amritk (approved):
Its all good tbh, I leave the choice up to hans

PR #5298, opened 2025-04-02
docs(hono): CORS middleware

hanspagel:
I’ve added a few words about CORS to the Hono README.

Adding the CORS middleware, makes it easier to import the OpenAPI document (from localhost, where we can’t use the remote proxy) into our web client.


marclave (approved):
some fresh handwritten docs by our co-founder 🤤

PR #5299, opened 2025-04-02
fix(api-client): doesn't show a download button anymore

### LINKED ISSUE #5208: Scalar.AspNetCore: Download button disappeared in new version
### What happens?

I've updated Scalar.AspNetCore version from 2.0.1 to 2.1.1 and download button disappered for files.

Version="2.1.1"

Image

Version="2.0.1"
Image

You can download demo project of this issue:

[ScalarDownloadFileIssueDemo.zip](https://github.com/user-attachments/files/19382458/ScalarDownloadFileIssueDemo.zip)

I expect download button as it used to be.

### OpenAPI Document

_No response_

hanspagel:
**Problem**

In #4864 we started to clone the request result, but we lost the `Blob` for binary responses. And then we couldn’t and didn't show a Download button anymore.

**Solution**

With this PR we’re improving the cloning of the request result, so we keep the `Blob`.

Fixes #5208 

Note: Restoring from the request history doesn’t work at all (and never did). We need to fix that in a separate PR.

**Checklist**

### DIFF
--- a/.changeset/red-pans-battle.md
+++ b/.changeset/red-pans-battle.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
+
+fix: download doesn’t appear anymore
--- a/packages/api-client/src/components/AddressBar/AddressBarHistory.vue
+++ b/packages/api-client/src/components/AddressBar/AddressBarHistory.vue
@@ -8,10 +8,8 @@ import {
 import type { Operation, RequestEvent } from '@scalar/oas-utils/entities/spec'
 import { httpStatusCodes } from '@scalar/oas-utils/helpers'
 import { computed } from 'vue'
-import { useRouter } from 'vue-router'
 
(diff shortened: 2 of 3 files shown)


marclave (approved):
🖖

amritk (approved):
yolo didn't test

hanspagel:
I’d love to, but I think this would require a bit more restructuring, and I already worked a little bit too long on that seemingly small issue, so let’s just roll with the patch and do the history properly soon.

KrzysztofPajak:
@hanspagel 
Thank you! In the new version of Scalar.AspNetCore (v2.1.6), the Download button has appeared and it correctly returns the file on the first request. However, when I make additional requests (e.g., by changing some request parameters or modifying a field), the downloaded file remains the same as the initial one — the updated file is not generated or returned correctly.

PR #5301, opened 2025-04-02
feat(api-client): add support for json api mime type

malinbranduse:
**Problem**

Currently, any response of `Content-Type: application/vnd.api+json` will be parsed as a Binary file, when it should be treated as a plain text json response.

See https://sandbox.scalar.com/e/bKl1f to reproduce.

**Solution**

With this PR, we map the json api content type to the right format.

**Checklist**

### DIFF
--- a/.changeset/rude-suits-roll.md
+++ b/.changeset/rude-suits-roll.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
+
+Adds support for JSON:API media type on responses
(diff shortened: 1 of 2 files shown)


hanspagel:
Oops, thanks for the PR! Can you just add a changeset? `pnpm changeset` (in root)

malinbranduse:
@hanspagel that was fast!
changeset pushed

malinbranduse:
Context: this is quite an issue for us @ Snyk so we would appreciate this getting merged in a timely manner.
We use the `application/vnd.api+json` content type for all our APIs, meaning none of the Scalar instances work properly on https://docs.snyk.io

hanspagel (approved):
Cool, thank you! :)

PR #5304, opened 2025-04-02
fix(api-reference): client selector and sidebar scroll listener

amritk:
**Problem**

Currently, the sidebar scroll listener is disabled by default. Also when we have a custom client selected, we cannot switch to the globally selected client (shell + curl by default).

**Solution**

With this PR we enable the scroll listener by default.
We also toss in a quick fix for the custom client selector.

**Checklist**


hanspagel (approved):
this is good

PR #5306, opened 2025-04-03
fix(openapi-parser): migrate file type, basic auth, parameters and responses headers correctly

matthyk:
**Problem**

Currently, some required migrations are missing:

### File type
```yaml
type: file
```
to
```yaml
type: string
format: binary
```

### Basic Auth
```yaml
type: basic
```
to
```yaml
type: http
scheme: basic
```

### Parameters
```yaml
 in: query
name: size
required: false
type: integer
multipleOf: 2,
minimum: 1,
maximum: 100,
```
to
```yaml
in: 'query',
name: 'size',
required: false,
schema: {
      type: 'integer',
      multipleOf: 2,
      minimum: 1,
      maximum: 100,
}
```

and the last one same for responses headers.

**Solution**

With this PR these required migrations are implemented and tested.

**Checklist**

### DIFF
--- a/.changeset/angry-singers-listen.md
+++ b/.changeset/angry-singers-listen.md
@@ -1,5 +1,0 @@
----
-'@scalar/snippetz': patch
----
-
-feat: new php/guzzle plugin
--- a/.changeset/brown-glasses-marry.md
+++ b/.changeset/brown-glasses-marry.md
@@ -1,5 +1,0 @@
----
-'@scalar/api-reference': patch
----
-
-fix: open in api client button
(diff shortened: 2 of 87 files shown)


hanspagel (approved):
OMG, this is awesome, you’re the GOAT

matthyk:
> You wanna fix the TS error?
> 
> > Error: src/utils/upgradeFromTwoToThree.ts(302,34): error TS2559: Type 'ParameterObject' has no properties in common with type 'ItemsObject'.

Sure! Will have a look

Issue #4481, opened 2025-01-16
Scalar bundle fails to properly bundle


DavidBiesack:
### What happens?

I'm using:
```bash
$ scalar --version
0.2.286
```

(installed with `npm install -g @scalar/cli`)

I have a source file in `apis/example/openapi.yaml` that $ref components from in `apis/common/openapi.yaml`
(source shown below)
and `scalar validate` shows both are valid.

```bash
$ scalar validate apis/example/openapi.yaml
scalar validate apis/example/openapi.yaml
Matches the OpenAPI specification (OpenAPI 3.1)

File validated in 143 ms

$ scalar validate apis/common/openapi.yaml
scalar validate apis/common/openapi.yaml
Matches the OpenAPI specification (OpenAPI 3.1)

File validated in 137 ms
```

However, the `scalar bundle` warns that the file doesn’t match the OpenAPI specification, then emits a bunch of garbage output

```bash
scalar bundle apis/example/openapi.yaml -o apis/example/bundled.yaml
[WARN] File doesn’t match the OpenAPI specification.

[WARN] {
  has: [ 33 ],
  keys: [ { open: '\x1B[33m', close: '\x1B[39m', rgx: /\x1b\[39m/g } ],
  reset: [Function: bound ],
  bold: [Function: bound ],
  dim: [Function: bound ],
  italic: [Function: bound ],
  underline: [Function: bound ],
  inverse: [Function: bound ],
  hidden: [Function: bound ],
  strikethrough: [Function: bound ],
  black: [Function: bound ],
  red: [Function: bound ],
  green: [Function: bound ],
  yellow: [Function: bound ],
  blue: [Function: bound ],
  magenta: [Function: bound ],
  cyan: [Function: bound ],
  white: [Function: bound ],
  gray: [Function: bound ],
  grey: [Function: bound ],
  bgBlack: [Function: bound ],
  bgRed: [Function: bound ],
  bgGreen: [Function: bound ],
  bgYellow: [Function: bound ],
  bgBlue: [Function: bound ],
  bgMagenta: [Function: bound ],
  bgCyan: [Function: bound ],
  bgWhite: [Function: bound ]
} (undefined)

OpenAPI Schema bundled in 135 ms
```

This does generate the output file `apis/example/bundled.yaml` and it is valid

```bash
$ scalar validate apis/example/bundled.yaml
Matches the OpenAPI specification (OpenAPI 3.1)

File validated in 136 ms
```

however

[...]

PR #5311, opened 2025-04-03
feat(api-reference): external references ↩️

### LINKED ISSUE #5176: $refs to external URLs don’t work
I tried this, and it’s not working.

Pretty sure the new store can’t resolve the references.

```ts
  // Works:
  // url: 'https://petstore31.swagger.io/api/v31/openapi.json',
  // Doesn’t work:
  content: `{"openapi": "3.1.0",
    "info": {
      "title": "Scalar API",
      "version": "1.0.0"
    },
    "paths": {
      "/pet": {
        "post": {
          "$ref": "https://petstore31.swagger.io/api/v31/openapi.json#/paths/~1pet/post"
        }
      }
    },
    "components": {
      "schemas": {
        "Pet": {
          "$ref": "https://petstore31.swagger.io/api/v31/openapi.json#/components/schemas/Pet"
        }
      }
    }
  }`,
  // Avoid CORS issues
})
```

hanspagel:
## Problem

External references weren’t working (anymore).

Our new store never really had support for external `$ref`’s.
And `@scalar/openapi-parser` never had support for relative URLs in `$ref`’s.

## Solution

Sooo … this PR adds those features (and more!) and makes the reference and the client work with external references.

I’ve had to rewrite how we fetch/read files into memory. Seems to simple, but was days of work. 😬 Old:

```ts
const filesystem = [
  {
    filename: 'foobar.yaml',
    dir: 'https://example.com',
    references: ['foobar.yaml']
  }, {
    // oops, the same filename as above, but in a another directory? caused conflicts
    filename: 'foobar.yaml',
    // omg, super complex to calculate relative URLs based on those scattered information
    dir: 'https://example.com/components/',
    references: [],
  },
]
```

New: 
```ts
const filesystem = [
  {
    // unique identifier
    uri: 'https://example.com/foobar.yaml',
    references: {
      // map references to unique identifiers
      'foobar.yaml': 'https://example.com/components/foobar.yaml',
    },
  }, {
    uri: 'https://example.com/components/foobar.yaml',
    references: {},
  },
]
```

**API Reference**
* local examples with external references
* logging for the fetching of external references

**Store**
* make the store fetch external references
* pass down the `proxyUrl` to fetch external references

**Parser**
* add `resolveInternalRefs` to bundle documents
* make `isFilesystem` a type guard
* resolve relative URLs

The `resolveInternalRefs` is also what we were missing for the bundling in the CLI, see #4481.

Fixes #5176.

## Checklist


packages/api-client/src/store/import-spec.ts
```diff
@@ -10,7 +10,7 @@ import { toRaw } from 'vue'
 /** Maps the specs by URL */
 export const specDictionary: Record<string, { hash: number; schema: OpenAPIV3.Document | OpenAPIV3_1.Document }> = {}
 
-type ImportSpecFileArgs = ImportSpecToWorkspaceArgs & Pick<ApiReferenceConfiguration, 'servers'>
+type ImportSpecFileArgs = ImportSpecToWorkspaceArgs & Pick<ApiReferenceConfiguration, 'servers' | 'proxyUrl'>
```
amritk, line 13: we weren't using the proxy before at all? Oh my. I guess from references we passed down the spec

amritk: wait the proxy is working from before

amritk (approved):
Glanced through the code, just the one question about proxy. But I did test and didn't get a double proxy so I guess it works

packages/api-client/src/store/import-spec.ts
hanspagel: the proxy is used for requests, but wasn’t used to fetch external references, that’s all 👌

PR #5321, opened 2025-04-03
fix(api-client): fix random extra borders + missing borders

cameronrohani:
before response had no borders + some extra borders existed in code snippet: 
image

now we have borders:
image

Also we're going to try defaulting to code snippets open by default for a few weeks and see how it sits :)


amritk (approved):

hanspagel (approved):
love it

PR #5322, opened 2025-04-03
fix(api-reference): strange jump on initial load

amritk:
Turns out this was a mistake, @cameronrohani can we confirm that the sidebar still scrolls on this PR?

**Checklist**


amritk:
Yolo merging this one for clerk. if it causes any issues I'll do a quick fix

PR #5323, opened 2025-04-03
fix(api-reference): fix some minor type issues with properties

cameronrohani:
fixing some very minor type stuff that's been driving me crazy

1. when there's no heading on a property there's double padding 
before:
image

after:
image

2. the delineator I used in enum's can't really be used with the fonts we have so had to just do it with CSS

before
image

after:
image

Last thing is we didn't have margin a proper between description + child attribute button so added one


amritk (approved):
yolo

hanspagel (approved):
omg, finally

PR #5324, opened 2025-04-03
fix(api-reference): this should fix the unselectable auth bug

amritk:
**Problem**

Currently, auth is unselectable if we use `updateConfiguration` from the htmlApi

**Solution**

With this PR this is another bandaid, we may need that new store sooner rather than later.

A fully reactive store makes me 🤤 

**Checklist**


hanspagel (approved):
okay, cool

PR #5332, opened 2025-04-04
fix(api-reference): schema padding and anyof

cameronrohani:
fixed padding on this + styled anyof to look like default tab styles

image


hanspagel (approved):
I didn’t do a proper review yet, but if you want it in today, it’s approved already. ✌️

hanspagel:
On a second glance, just awesome

PR #5334, opened 2025-04-04
fix(api-reference): disable intersection on update

amritk:
**Problem**

Currently, we may have some jumpiness on config update.

**Solution**

With this PR we disable the intersection observer in case it was causing the jumpiness.

**Checklist**


hanspagel (approved):
ok, makes sense

PR #5342, opened 2025-04-07
feat(api-client): show custom code examples

hanspagel:
**Problem**

Currently, we show custom code examples (`x-codeSamples`) in the API reference, but not in the API client.

**Solution**

This PR adds custom code examples to the store, adds them to the dropdown in the API client, and renders them in the code snippet, when they are selected.

If they are selected, the selected HTTP client isn’t updated in the store, because no one knows whether other operations have custom examples, too.

**Checklist**

### DIFF
--- a/.changeset/new-rules-beam.md
+++ b/.changeset/new-rules-beam.md
@@ -1,0 +1,7 @@
+---
+'@scalar/api-reference': patch
+'@scalar/api-client': patch
+'@scalar/oas-utils': patch
+---
+
+feat: show custom code examples in the API client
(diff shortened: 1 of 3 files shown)


marclave (approved):
NICE 🔥

amritk (commented):
The only issue I found was that the custom code sample is selected by default in references but not in the client. So I will just change that

amritk (approved):
Alright I made that change. It would also be nice if the references and client were synced in client selected but thats a problem for another day

PR #5345, opened 2025-04-07
feat(api-reference): new lazy loading without fake floating container

amritk:
So this is some cool tech, but I don't think it will make the cut.

Just wanted to make the PR in case I ever want to look back on it


amritk:
Closing in favor of the real fix, this is just here for fun

https://github.com/scalar/scalar/pull/5348

PR #5348, opened 2025-04-07
fix(api-reference-react): only lazyload once

amritk:
**Problem**

Currently, we show the lazy loading fake container on every config change, we don't need to do that.

**Solution**

With this PR we only show the lazy loading on initial load.

**Checklist**


hanspagel (approved):
ok

PR #5349, opened 2025-04-08
fix(api-reference): references search a11y round 2 (3?)

hwkr:
Adds a number of a11y fixes to the references search and a few smaller fixes to the rest of the references.

- Updates the search dialog to have a label and stopped the entire dialog content being used as the description for all dialogs.
- Updates the search dialog to use a combobox / aria-activedescendent pattern (reads the number of results).
- Return to the search button if the search is opened via click
- Search results need to either wrap to another line at high zoom levels
- Fixed the tab count being incorrect for the client libraries tabs, also looked into the tab order but decided to keep the order as-is since the more button is sort of a pseudo tab.
- Add unique screen reader names for the 'Show/Hide Child Attributes' buttons.
- Swapped an `<li>` for a `<div>` in the response model since it only ever has one item (as far as I can tell).

**Checklist**

### DIFF
--- a/packages/api-reference/index.html
+++ b/packages/api-reference/index.html
@@ -21,12 +21,13 @@
       Scalar.createApiReference('#app', {
         sources: [
           {
-            url: 'https://petstore.swagger.io/v2/swagger.json',
-          },
-          {
             title: 'Scalar Galaxy', // optional, would fallback to 'API #1'
             slug: 'scalar-galaxy', // optional, would be auto-generated from the title or the index
             url: 'https://cdn.jsdelivr.net/npm/@scalar/galaxy/dist/latest.json',
+          },
+          {
+            title: 'Swagger Petstore',
+            url: 'https://petstore.swagger.io/v2/swagger.json',
           },
         ],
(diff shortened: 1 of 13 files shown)


packages/api-reference/index.html
```diff
@@ -20,14 +20,15 @@
...
-          },
           {
             title: 'Scalar Galaxy', // optional, would fallback to 'API #1'
             slug: 'scalar-galaxy', // optional, would be auto-generated from the title or the index
             url: 'https://cdn.jsdelivr.net/npm/@scalar/galaxy/dist/latest.json',
           },
+          {
+            title: 'Swagger Petstore',
+            url: 'https://petstore.swagger.io/v2/swagger.json',
+          },
```
hanspagel, line 31: lol I did this in so many branches already, but it never made it to main

hanspagel (approved):
based

packages/api-reference/index.html
hwkr: I swear I've done it before too 😂

PR #5341, opened 2025-04-07
feat(api-reference): custom install instructions for SDKs

hanspagel:
**Problem**

We have our awesome code examples, but some people have SDKs, that are even cooler. They can provide custom code examples already, but there’s no obvious way to tell users how to install their SDK.

**Solution**

With this PR we’re introducing a new OpenAPI extension to render custom setup instructions for SDKs.

Fixes #4065

**Usage**

```diff
openapi: 3.1.0
info:
  title: Val Town API
  version: 1.0
paths:
  '/v1/eval':
    post:
+      x-codeSamples:
+      - label: ValTown JS SDK
+        lang: JavaScript
+        source: |-
+          import ValTown from '@valtown/sdk';
+
+          const valTown = new ValTown();
+
+          async function main() {
+            const valRunAnonymousResponse = await valTown.vals.runAnonymous({ code: 'console.log(1);' });
+
+            console.log(valRunAnonymousResponse);
+          }
+
+          main();
```

**Checklist**

I’ve gone through the following:

- [x] I’ve added an explanation _why_ this change is needed.
- [ ] I’ve added a changeset (`pnpm changeset`).
- [ ] I’ve added tests for the regression or new feature.
- [ ] I’ve updated the documentation.

PR #5350, opened 2025-04-08
fix(oas-utils): `omitUndefinedValues` (for Zod schemas) doesn’t handle Arrays

hanspagel:
**Problem**

We have this helper we use for Zod schemas, that removes undefined values from the data. But it transforms arrays to object, that’s bad.

**Solution**

With this PR it’s all good again, because now we handle Arrays and we even have tests for it.

Got this from #5341, but need this in sooner.

**Checklist**

### DIFF
--- a/packages/oas-utils/src/helpers/omit-undefined-values.ts
+++ b/packages/oas-utils/src/helpers/omit-undefined-values.ts
@@ -1,0 +1,24 @@
+/**
+ * Removes undefined values from an object.
+ *
+ * Can be used as a transform function for any Zod schema.
+ */
+export const omitUndefinedValues = <T extends object>(data: T): T => {
+  // Handle arrays specially
+  if (Array.isArray(data)) {
+    return data.map((item) =>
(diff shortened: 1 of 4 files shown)


packages/oas-utils/src/helpers/omit-undefined-values.ts
```diff
@@ -0,0 +1,24 @@
+/**
+ * Removes undefined values from an object.
+ *
+ * Can be used as a transform function for any Zod schema.
+ */
+export const omitUndefinedValues = <T extends object>(data: T): T => {
+  // Handle arrays specially
+  if (Array.isArray(data)) {
+    return data.map((item) =>
```
amritk, line 9: I thought isDefined might work here but we would mod it to allow null.

https://github.com/scalar/scalar/blob/main/packages/oas-utils/src/helpers/is-defined.ts

🤷🏾

amritk (approved):
It has tests, it gets approved

PR #5352, opened 2025-04-08
chore: ensure consistent order of package.json exports keys

xC0dex:
**Problem**

Currently, the generated order of the export entries is different on Windows. This could be annoying as it would always generate a gitt diff.

**Solution**

With this PR the keys are sorted, so its always the same order.

**Checklist**


hanspagel (approved):
nice, we need this for Windows support

PR #5357, opened 2025-04-08
fix(api-client): improve request table header a11y

hwkr:
Before / After:

https://github.com/user-attachments/assets/11af1bb0-cc58-4368-913a-423f095e620f

**Checklist**


hanspagel (approved):
a11y all the things

PR #5358, opened 2025-04-08
fix(api-client): don't set html background from client

hwkr:
This was causing issues if `--scalar-background-1` was set to transparent (which it shouldn't be but we probably don't want to do this anyway).

It seems like this was [added along with the client import flow](https://github.com/scalar/scalar/commit/bd8e253c4126216fabd1a8d504f61026e2373690#diff-fda38c80e7cf76a6c78e3caf45248ede32f55bad34d29fb4517aa46260d40b47) at some point but I did some testing and the import flow works fine without it so I'm guessing it's just stale code. We probably don't want to do it this way anyway because we use the `.dark-mode` class for the contrast cards in the references meaning `html:has(.dark-mode)` applies even in light mode, we just didn't notice because it was hidden behind a background of `--scalar-color-1`  <sup>so<sup>cursed</sup></sup>.

**Checklist**

### DIFF
--- a/.changeset/forty-flowers-impress.md
+++ b/.changeset/forty-flowers-impress.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
+
+fix(api-client): don't set html background from client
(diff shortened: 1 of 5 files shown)


amritk (approved):
when it comes to css vars, Brynn knows best

hwkr:
> when it comes to css vars, Brynn knows best
😅

PR #5371, opened 2025-04-09
fix(api-client): make useResponseBody reactive

marclave:
**Problem**

Currently useResponseBody is not reactive 🫨 

**Solution**

With this PR it is 

**Checklist**

### DIFF
--- a/packages/api-client/src/hooks/useResponseBody.ts
+++ b/packages/api-client/src/hooks/useResponseBody.ts
@@ -1,5 +1,5 @@
 import { extractFilename } from '@/libs/extractAttachmentFilename'
-import { computed, isRef, ref } from 'vue'
+import { computed, isRef } from 'vue'
 import MimeType from 'whatwg-mimetype'
 import type { Ref } from 'vue'
 
--- a/packages/api-client/src/hooks/useResponseBody.ts
+++ b/packages/api-client/src/hooks/useResponseBody.ts
@@ -8,10 +8,8 @@ import type { Ref } from 'vue'
  * Extracts MIME type, attachment filename, and generates a data URL.
  */
 export function useResponseBody(props: {
-  data: Ref<unknown> | unknown
-  headers:
-    | Ref<{ name: string; value: string; required: boolean }[]>
-    | { name: string; value: string; required: boolean }[]
+  data: Ref<unknown>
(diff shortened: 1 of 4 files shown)


cameronrohani (approved):
didnt know u knew how to code, but good fix 😹

packages/api-client/src/hooks/useResponseBody.ts
```diff
@@ -1,37 +1,44 @@
...
  * Processes the response body of an HTTP request.
  * Extracts MIME type, attachment filename, and generates a data URL.
  */
 export function useResponseBody(props: {
-  data: unknown
-  headers: { name: string; value: string; required: boolean }[]
+  data: Ref<unknown> | unknown
+  headers:
+    | Ref<{ name: string; value: string; required: boolean }[]>
+    | { name: string; value: string; required: boolean }[]
```
amritk, line 14:
```suggestion
  data: MaybeRef<unknown> 
  headers: MaybeRef<{ name: string; value: string; required: boolean }[]>
```

got some utility types for this

packages/api-client/src/hooks/useResponseBody.ts
```diff
@@ -1,37 +1,44 @@
...
 }) {
   const isBlob = (b: any): b is Blob => b instanceof Blob
 
+  // Handle both Ref and direct values
+  const dataRef = computed(() => (isRef(props.data) ? props.data.value : props.data))
+  const headersRef = computed(() => (isRef(props.headers) ? props.headers.value : props.headers))
```
amritk, line 20:
```suggestion
  const dataRef = toRef(props.data)
  const headersRef = toRef(props.headers)
```
I believe this should work as well

amritk (approved):
Good stuff, requested some code style changes but left approved for easy merge after

packages/api-client/src/hooks/useResponseBody.ts
amritk: on second thought I would just force them to be refs, in which case do we ever not want refs?

hanspagel:
Let me add tests

hanspagel (approved):
looks okay, seems to work, has tests. what issue do we fix here?

marclave:
> looks okay, seems to work, has tests. what issue do we fix here?

we fix an issue where the download button doesnt properly get the latest response :)

PR #5378, opened 2025-04-10
feat(components): scalar form text field improvements

hwkr:
Adjusts the ScalarForm fields to allow changing the label component (in case you don't want to automatically tie to an input) and tweaks the layout / stories for the text input. Also adds a test.

**Checklist**


hanspagel (approved):
rad

PR #5382, opened 2025-04-11
refactor(components): cleanup old icons in Scalar components

hwkr:
We had a bunch of icons that either weren't aligned with our current icon spec or just weren't being used.

Before / After

Google Chrome-2025-04-10-20-54-08

**Checklist**


packages/components/src/components/ScalarIcon/icons/Ellipses.svg
hwkr, line 1: This icon actually looks identical with our default stroke width of 2

packages/components/src/components/ScalarIcon/icons/Error.svg
hwkr, line 1: For this icon I tried to match our Help icon as best I could.

hanspagel (approved):
great work, needs approval by @cameronrohani

packages/components/src/components/ScalarIcon/icons/Error.svg
antlio: there's already the `Alert` one that does the same job, should we ditch one or another?

marclave (approved):
image

PR #5396, opened 2025-04-11
fix: make code snippet default closed

marclave:
**Problem**

Currently we hit some large codemirror perf bugs on api client cause we load BODY + SNIPPET

image

**Solution**

Lets not overwhelm the user and have code snippets hidden by default

**Checklist**


amritk (approved):
huge W

PR #5398, opened 2025-04-11
fix(api-reference): show sidebar if not set

hwkr:
It seems like at some point the defaults for the config got moved into the `ApiReferenceLayout` (which is a child of `ModernLayout`) meaning that that if `showSidebar` was undefined it didn't get defaulted to true in `ModernLayout`.

Also updated our tests so it will hopefully be harder to break it next time.

@hanspagel let me know if this seems like a good solution or if we should do something else 🫡 

**Checklist**

### DIFF
--- a/.changeset/metal-experts-explode.md
+++ b/.changeset/metal-experts-explode.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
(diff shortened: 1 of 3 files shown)


hanspagel (approved):
Perfect solution

Issue #5399, opened 2025-04-11
CURL examples show incorrect Content-Type when using custom media types with multiple examples


AnnaKudriasheva:
### What happens?

When defining a `POST` endpoint with a custom media type (e.g. `application/vnd.api+json`) and multiple request examples, Scalar's generated curl examples incorrectly default to `Content-Type: application/json`.

### What did you expect to happen?

Expected behavior:
- Curl example should show correct `Content-Type` value - `application/vnd.api+json`.
- When you click `Test Request`, the `Content-Type` should default to `application/vnd.api+json`.

### OpenAPI Document

https://sandbox.scalar.com/e/NTQrF

antlio:
hey @AnnaKudriasheva thank you for bringing this up!
side note: examples doesn't seem to be displayed also when using vendor specific mime type.

PR #5402, opened 2025-04-12
fix(api-client): vendor specific content type examples

antlio:
**Problem**

currently when using vendor specific content type (e.g. `application/vnd.api+json`), the type gets changed to `application/json` + the examples selector is not getting displayed as seen in #5399.

**Solution**

this pr updates the request body logic to avoid updating the content type and bring back the example selector + fix the body update on change.

| before | after |
| -------|------|
| image | image |
| unsupported vendor specific content type + no examples | vendor specific content type support + examples | 

**Checklist**

### DIFF
--- a/.changeset/bright-mirrors-build.md
+++ b/.changeset/bright-mirrors-build.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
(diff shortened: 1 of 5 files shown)


hanspagel (approved):
rad, we have tests, we won’t break it again

PR #5406, opened 2025-04-14
fix(api-client): mask api key value in authentication section

SebastianBienert:
**Problem**

Currently, when using [apiKey security schema ](https://swagger.io/docs/specification/v3_0/authentication/api-keys/) value of the api key is displayed as plain text. API keys are considered secrets the same way passwords are, they should not be displayed as plain text.

**Solution**

With this PR I change displayed value in api key input to password type the same way it is already handled for password in basic HTTP authentication

**Checklist**

### DIFF
--- a/.changeset/few-dogs-brake.md
+++ b/.changeset/few-dogs-brake.md
@@ -1,0 +1,5 @@
+---
+"@scalar/api-client": patch
+---
+
+feat: mask api key value in authentication section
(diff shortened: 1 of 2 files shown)


amritk (approved):
I added a changeset, wicked PR!

hanspagel (approved):
cool! thanks for taking the time to submit this PR 🤟

PR #5415, opened 2025-04-15
fix(api-reference): align document selector chevron

hwkr:
Before / After

Google Chrome-2025-04-15-16-53-48@2x

Google Chrome-2025-04-15-16-53-34@2x


antlio:
seems like the padding is back somehow:

image

PR #5425, opened 2025-04-17
feat(icons): add Phosphor icons to @scalar/icons

hwkr:
Brings all the icons in [`@phosphor-icons/core`](https://github.com/phosphor-icons/core) into a new export in `@scalar/icons`. It's loosely based on [`@phosphor-icons/vue`](https://github.com/phosphor-icons/vue) and the assembler that they use there.

- Tree shaking now works for the icons generated in `@scalar/icons` so we can have as many icons as we want ✨ 
- I took the `@scalar/icons` package for the new phosphor icons and moved the icon library that was there to `@scalar/icons/library` (we can replace the icons in there with Phosphor too at some point)
- I moved `useBindCx` and `cva` to `@scalar/use-hooks/useBindCx`
- I have the generator set up to only generate [`ScalarIconMagnifyingGlass.vue`](https://github.com/scalar/scalar/pull/5425/files#diff-48ba4c32f79bb482e4e0aac362df23b3fb7ff0e164a4b6d149a68b3936a46f2d) rather than all ~1500 icons to make the PR easier to read, when we're ready to go ahead I'll generate the rest


packages/api-reference/src/features/Search/SearchButton.vue
```diff
@@ -63,11 +64,9 @@ function handleClick() {
     role="search"
     type="button"
     @click="handleClick">
-    <ScalarIcon
+    <ScalarIconMagnifyingGlass
```
hwkr, line 67: Added this here to test tree shaking and see if the bundle explodes

hwkr:
I'm sorry this PR break the GitHub file view...

packages/api-reference/src/features/Search/SearchButton.vue
hwkr: image


Google Chrome-2025-04-16-22-07-19@2x

marclave (approved):
lets goooo @hwkr 

(cam also approves this)

PR #5427, opened 2025-04-17
feat(api-reference): new string fallback for schema examples

hanspagel:
**Problem**

Currently, we’re using `"…"` as the fallbacks for strings when we generate examples from schemas.

 Screenshot 2025-04-17 at 15 40 28

**Solution**

I thought I’d be smart when I built that, but a few people are confused by it, and actually, just using `"string"` seems easier to understand (it’s also what Swagger UI does).

Screenshot 2025-04-17 at 15 38 17

Note: We can’t really do this for numbers, booleans and others, because we’d need to make them a string then.

**Checklist**

### DIFF
--- a/.changeset/proud-forks-fly.md
+++ b/.changeset/proud-forks-fly.md
@@ -1,0 +1,6 @@
+---
+'@scalar/api-reference': patch
+'@scalar/mock-server': patch
+---
+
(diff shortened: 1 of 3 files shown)


antlio (approved):
digging this

marclave (approved):
😮‍💨 love it

PR #5432, opened 2025-04-17
feat(api-reference): register scalar instances globally

amritk:
**Problem**

Currently, we register a window.Scalar object but all it can do is create instances

**Solution**

With this PR we save a reference to the instance globally so we can perform actions on it. Also updated the deprecation message as it did nothing before.

**Checklist**

I’ve gone through the following:

- [x] I’ve added an explanation _why_ this change is needed.
- [x] I’ve added a changeset (`pnpm changeset`).
- [x] I’ve added tests for the regression or new feature.
- [ ] I’ve updated the documentation.

PR #5434, opened 2025-04-17
fix(docusaurus): move docusaurus to the CDN + HTML api

### LINKED ISSUE #2169: docusaurus: doesn’t use full height
References with just a little bit of content aren’t using the full height in Docusaurus.

**Configuration**

```js
[
  '@scalar/docusaurus',
  {
    id: 'yaml-string',
    label: 'yaml-string',
    route: '/yaml-string',
    configuration: {
      spec: {
        content: `openapi: 3.1.0
info:
title: Hello World
version: 1.0.0
paths: {}
`,
      },
    },
  } as ScalarOptions,
],
],
```

**Screenshot**

Screenshot 2024-06-19 at 14 13 10

### LINKED ISSUE #4976: Module Not Found Error in Docusaurus/Scalar Project: Export Path Issue
### What happens?

Hi, I've found an error when starting my Docusaurus/scalar project, it shows an error `ERROR in ./node_modules/@scalar/docusaurus/dist/ScalarDocusaurus.js 1:642-698` and `Module not found: Error: Package path . is not exported from package /Users/macintosh/done/work/chainx/chainx-docs/node_modules/@scalar/api-reference-react (see exports field in /Users/macintosh/done/work/chainx/chainx-docs/node_modules/@scalar/api-reference-react/package.json)`

Image
Image

Able to run `yarn start` perfectly, I'm also using `npm` but the results are the same.

### OpenAPI Document

https://sandbox.scalar.com/e/ZSjN5

### LINKED ISSUE #5313: deprecated scalar events on docusaurus plugin
### What happens?

Using the scalar docusaurus plugin we get this warning:

- `scalar:destroy-references event has been deprecated, please use window.Scalar.destroy instead`

And the scalar component doesn't appears in the DOM.

- The OpenAPI scalar document should appear;
- Move to the new specs.

### OpenAPI Document

_No response_

### LINKED ISSUE #5411: Authentication card title looks wrong in Docusaurus
### What happens?

Image

Expected the title style to match the ones below and above.

### OpenAPI Document

_No response_

### LINKED ISSUE #5416: @scalar/docusaurus plugin bug: switching back and forth makes scalar page to not render
### What happens?

If I open the scalar page for the first time within Docusaurus it renders okay but if I navigate away from it and come back, it shows a blank page. This is what I see in the console log:

Image

I think the issue is similar to #5043, however the underlying bug causing it might be different 

Scalar page to render properly even when I navigate between pages

### OpenAPI Document

_No response_

amritk:
# DO NOT REBASE
~wait till we merge the [html api global PR](https://github.com/scalar/scalar/pull/5432) first~
closes #5416
closes #5313
closes #4976
closes #2169
closes #5411

**Problem**
Currently, docusaurus is busted when switching pages

**Solution**

With this PR we create a new instance of the app on every page switch. Unfortunately you can only mount a vue app once so there was no point in keeping the same instance.

We also removed the dependency on the api-reference-react package so it is MUCH lighter now.

There is one downside, it adds a new script tag for every call to the plugin. But it only makes one network call so it should be ok. Will open a docusaurus issue for this.

image

**Checklist**


packages/api-reference/src/standalone/lib/register-globals.ts
```diff
@@ -15,10 +16,8 @@ export const registerGlobals = () => {
     return
   }
 
-  const globalWindow = window
```
amritk, line 18: not sure what that was for  @hanspagel ?

amritk:
Alright @hanspagel I killed the global registration PR and removed it from this one. All I kept was moving the types to the types package then we can get rid of the `api-reference-react` package in docusaurus and all of the dependencies

marclave (approved):

packages/api-reference/src/standalone/lib/register-globals.ts
hanspagel: Left-over

integrations/docusaurus/package.json
```diff
@@ -38,7 +38,6 @@
     "CHANGELOG.md"
   ],
   "dependencies": {
-    "@scalar/api-reference-react": "workspace:*",
```
hanspagel, line 41: 🎉

hanspagel (approved):
Works like a charm ✨  PR of the week already

PR #5442, opened 2025-04-18
fix(api-client): prevents paste event from code input

antlio:
**Problem**

currently when pasting a specification within a code input, it fires the document paste event and open the import collection modal.

**Solution**

this pr updates the handle paste function to prevent event from being fired from a code mirror instance by targeting `cm-content` class.

**Checklist**

### DIFF
--- a/.changeset/few-days-matter.md
+++ b/.changeset/few-days-matter.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
+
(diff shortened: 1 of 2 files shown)


marclave (approved):
nice fix @antlio <3

dad strength energy coming in with these PRs fr fr 💪

amritk:
This is better than my fix lol

PR #5446, opened 2025-04-18
fix(api-client): import modal transition

antlio:
**Problem**

color mode utility classes being currently set to the body element - while background variables are scoped to them (dark | light) - do not make background variables available at the html element level (as wrapping the body one).

this is causing some background color mismatch during the closing the import collection modal transition.

**Solution**

this pr updates the import collection modal style to target the element and maintain the background.

| before | after |
| -------|------|
| image | image |
| background color is not set  | background color set according to theme | 

**Checklist**

### DIFF
--- a/.changeset/chilled-vans-join.md
+++ b/.changeset/chilled-vans-join.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
+
+fix: updates import collection modal transition style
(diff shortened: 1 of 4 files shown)


cameronrohani:
I'd honestly be slightly scared to take this risk on the API References when it's just to fix 1 thing in the client 😶 

maybe @hwkr has more guidance here though

hwkr:
As an alternative maybe we could change how we do the import modal in the client so we don't need to show the HTML element?

marclave (approved):
🖖

PR #5450, opened 2025-04-19
style(schema-discriminator): update padding in discriminator panel

### LINKED ISSUE #5428: Missing a bit of padding on property level 0
### What happens?

Either the level is wrong or we are missing some padding

Image

The padding to match the one above

### OpenAPI Document

https://api.fabric.so/v2/openapi/public.yaml

Jannchie:
**Problem**

- resolved #5428

Currently:

The discriminator panel has zero padding, which doesn't look like an intentional style. 

image

**Solution**

In this PR, I have set the padding for level 0 to `8px`.

After:

image

**Checklist**

### DIFF
--- a/.changeset/brave-feet-know.md
+++ b/.changeset/brave-feet-know.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
+style: Update padding in discriminator panel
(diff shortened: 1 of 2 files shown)


marclave (approved):
nice work @Jannchie ✨ !

antlio:
thank you so much for that contribution @Jannchie 

i started fixing this but adding padding there is actually breaking some other discriminator panel in some case. 

will add a spec for further test in another comment!

hanspagel (approved):
oops, thanks!

PR #5454, opened 2025-04-20
docs(Scalar.AspNetCore): add docs for legacy .NET versions and FastEndpoints

xC0dex:
This PR adds documentation for legacy .NET versions that are not supported by our official `Scalar.AspNetCore` package. Specifically, I included a simple guide in the integration docs explaining how to use the HTML integration in existing .NET applications.

Additionally, I’ve added `FastEndpoints` to the list of supported OpenAPI document providers. While it uses `NSwag` under the hood, its setup differs slightly from the standard configuration.

### DIFF
--- a/documentation/integrations/dotnet.md
+++ b/documentation/integrations/dotnet.md
@@ -71,6 +71,21 @@ if (app.Environment.IsDevelopment())
 }
 ```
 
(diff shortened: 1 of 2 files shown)


marclave (approved):
AWESOME work @xC0dex <3

PR #5456, opened 2025-04-21
fix(components): `<ScalarTooltip />` base font size

antlio:
**Problem**

currently the default text size used in the tooltip has broken and is now too big according to its context.

**Solution**

this pr adds a text size variant with a text size default.

| before | after |
| -------|------|
| image | image |
| image | image |

**Checklist**


hanspagel (approved):
yess

hwkr (commented):
Thank you!

PR #5458, opened 2025-04-21
fix(api-client): environment collection

### LINKED ISSUE #5384: Selecting environment with short name throws an error
### What happens?

Environment does not change.
```
ZodError: [
  {
    "code": "too_small",
    "minimum": 7,
    "type": "string",
    "inclusive": true,
    "exact": false,
    "message": "String must contain at least 7 character(s)",
    "path": [
      "uid"
    ]
  }
]
```

I used environment name `owner` which is less than 7 characters. There is no visual indication about this error, also why is this validation even here?

Environment should change.

### OpenAPI Document

_No response_

antlio:
**Problem**

currently collection environment usage is not properly usable due to issues:
- environment selector position is overflowing once selected
- environment collection routes needs to be updated
- pill plugin lacks reactivity on environment change
- draft collection if shown in environment page
- environment selector is displaying uid instead of name
- environment collection name cannot be updated
- zod schema prevent short name usage for environment

**Solution**

this pr fixes #5384 #4349 + listed above issues +:
- updates default environment variable pill color for better accessibility
- adds icon on global variable pill

`environment selector`
| before | after |
| -------|------|
| image | image |
| selector overflowing once selected | selector is teleported | 

`environment active state`
| before | after |
| -------|------|
| image | image |
| lacks active state and shows draft | active state shown and draft filtered out | 

**Checklist**

### DIFF
--- a/packages/api-client/src/components/CodeInput/CodeInput.test.ts
+++ b/packages/api-client/src/components/CodeInput/CodeInput.test.ts
@@ -1,12 +1,13 @@
 import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest'
 import { mount, type VueWrapper } from '@vue/test-utils'
 import CodeInput from './CodeInput.vue'
-import { useCodeMirror, type Extension } from '@scalar/use-codemirror'
+import { useCodeMirror } from '@scalar/use-codemirror'
 import { useClipboard } from '@scalar/use-hooks/useClipboard'
 import { enableConsoleError, enableConsoleWarn } from '@/vitest.setup'
-import { ref } from 'vue'
+import { ref, toValue } from 'vue'
 import { environmentSchema } from '@scalar/oas-utils/entities/environment'
 import { workspaceSchema } from '@scalar/oas-utils/entities/workspace'
+import { nanoid } from 'nanoid'
 
 // Mock dependencies
 vi.mock('@scalar/use-codemirror', async (importOriginal) => {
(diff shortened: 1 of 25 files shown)


packages/api-client/src/components/CodeInput/CodeInput.test.ts
```diff
@@ -53,13 +54,25 @@ describe('CodeInput', () => {
     }
   })
 
+  const defaultProps = {
+    modelValue: 'test value',
+    environment: environmentSchema.parse({
+      uid: nanoid(),
```
amritk, line 60:
```suggestion
```
uids are automatically added!

packages/api-client/src/components/CodeInput/CodeInput.test.ts
```diff
@@ -53,13 +54,25 @@ describe('CodeInput', () => {
...
+      name: 'Test Environment',
+      value: 'test',
+      color: '#000000',
+    }),
+    envVariables: [],
+    workspace: workspaceSchema.parse({
+      uid: nanoid(),
```
amritk, line 67:
```suggestion
```

amritk (approved):
Code looks good! Did not test but as its broken now it should only get better eh 🙂

packages/api-client/src/components/CodeInput/CodeInput.test.ts
antlio: aaah not anymore as i get rid of the nanoid schema

antlio: does it still after the last commit update?

hanspagel:
I just tested it and we don’t seem to replace the environment variables with their values.

See how it’s detected in the address bar, but still sent as {{foo}} in the request.

Screenshot 2025-04-25 at 15 17 27
be" />

antlio:
> I just tested it and we don’t seem to replace the environment variables with their values.
> 
> See how it’s detected in the address bar, but still sent as {{foo}} in the request.

yes it's currently broken - i fixed this in another PR

hanspagel:
let’s just get the other PR in then ✌️

hanspagel (approved):
works (again), great!

PR #5460, opened 2025-04-22
feat(api-reference): omit deprecated fields from generated examples

hanspagel:
**Problem**

If a schema/property is flagged as deprecated, it’s still in the generated examples.

**Solution**

We can omit it from the generated examples. I’ve found this great addition here: https://github.com/GitbookIO/gitbook/pull/3085

**Checklist**


amritk (approved):
gj gitbook

PR #5461, opened 2025-04-22
feat: enforce file names

hanspagel:
**Problem**

Currently, we have different approaches to name files in the repository. This feels very inconsistent.

**Solution**

With this PR we’re introducing https://ls-lint.org/ to lint file names. Unfortunately, it doesn’t fix file names. So I’ve added a tiny script that tries to fix file names.

It’s only configured for the mock server, the void server and the core package for now. Let’s start easy.

I’ve also added a Git commit hook.

**Checklist**


amritk:
whooooooooooooooooooooooooo

.ls-lint.yml
```diff
@@ -0,0 +1,18 @@
+ls:
+  packages/{void-server,mock-server,core}/src:
+    .ts: kebab-case
+    .d.ts: kebab-case
+    .test.ts: kebab-case
+    .vue: PascalCase
+  packages/{void-server,mock-server,core}/hooks:
+    .ts: PascalCase
```
amritk, line 8:
```suggestion
  use*.ts:
    .ts: PascalCase
```
not sure the syntax but hooks should be camel case? What do you think? Or just kebab case since its a typescript file. Not really thinking it should be pascal case.

amritk (commented):
Wicked script, just one question about hooks

.ls-lint.yml
hanspagel: Just tested it, couldn’t make it work like this!
Unfortunately, the tool seems a bit limited.

amritk:
just moved to draft so it doesn't get merged by accident, lets nail it down

hanspagel:
Actually, we should do one script that does the linting and the fixing. We’ll run into problems if those are two separate things.

amritk:
Maybe we just do all kebab case. It's actually much safer since different file systems handle case differently.

hanspagel:
Nobody got time to do that

PR #5462, opened 2025-04-22
feat(galaxy): use `default` theme for playground

hanspagel:
**Problem**

Currently, https://galaxy.scalar.com/ shows the Hono custom theme. But this is our default example, that’s not really related to Hono.

**Solution**

Let’s just use the default theme.

**Checklist**


antlio (approved):
oui !

cameronrohani:
we should also update our elysia / hono themes to use p3 colors :P

PR #5471, opened 2025-04-23
feat(api-reference): single `text-base` wrap for specification extensions

hanspagel:
**Problem**

When registering two (or more) specification extensions (through the plugin API), each component is wrapped in a `div.text-base` wrapper to reset the CSS `font-size`. This can lead to problems, when styling the group of rendered elements.

Screenshot 2025-04-23 at 13 44 55

**Solution**

With this PR we’re rendering only one `div.text-base` around all components.

**Checklist**


antlio (approved):
dope

PR #5472, opened 2025-04-23
feat(api-reference): render top-level specification extensions

hanspagel:
**Problem**

Currently, we render specification extensions (`x-custom-property`) already, but we ignore them in the root (because the specification doesn’t explicitly allows them). But a lot of people actually have top-level specification extensions (`x-logo` is a super popular one).

**Solution**

With this PR we allow to register custom components for those top-level specification extensions, too.

Turns out, the rendering wasn’t reactive, so I fixed that, too. (They didn’t show up when the content changed.)

**Checklist**

### DIFF
--- a/.changeset/giant-boxes-drop.md
+++ b/.changeset/giant-boxes-drop.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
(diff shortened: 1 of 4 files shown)


amritk:
Code looks good, did not test!

PR #5476, opened 2025-04-23
fix(api-reference): types + reactivity for auth config

amritk:
**Problem**

Currently, authentication config was typed to any and didn't work for empty string values (falsy).

**Solution**

With this PR we add an explicit check for defined and null and fixed types.

**Checklist**


packages/oas-utils/src/transforms/import-spec.ts
```diff
@@ -215,21 +217,30 @@ export async function importSpecToWorkspace(
           flow.type = key
 
           // Prefill values from authorization config - old deprecated config
+          // @ts-expect-error - deprecated
```
amritk, line 220: @hanspagel how long should we support the old config? we can do a minor release and just remove it soon I thnk

marclave (approved):

PR #5484, opened 2025-04-24
fix(api-client): request input pill

antlio:
**Problem**

currently when using pill in an input the padding is off.

**Solution**

this pr updates padding on input with pill element.

| before | after |
| -------|------|
| image | image |

**Checklist**


hanspagel (approved):
send it

PR #5487, opened 2025-04-24
chore(api-client): export a few components

amritk:
**Problem**

Would like to use these components outside of the client.

**Solution**

With this PR we export a few more components. Also added an actions slot to auth

As per hans's request I moved the component to components package instead and added some build tooling. 

**Checklist**


hanspagel (commented):
Should we exclude zod from the builds? That might fix it, too?

hanspagel:
Seems stale, I just close it for now, but feel free to reopen.

amritk:
I forgot about this one lol

PR #5429, opened 2025-04-17
fix(api-reference): overflow scrolling on anyOf discriminator

amritk:
**Problem**

image

**Solution**

image

**Checklist**

I’ve gone through the following:

- [x] I’ve added an explanation _why_ this change is needed.
- [x] I’ve added a changeset (`pnpm changeset`).
- [ ] I’ve added tests for the regression or new feature.
- [ ] I’ve updated the documentation.

Issue #795, opened 2024-01-16
Options could be clearer when using a discriminator


mrchimp:
### Expected Behavior

When using a [discriminator](https://spec.openapis.org/oas/v3.0.3#discriminator-object) a user should be able to tell which of the  oneOf/AnyOf/allOf options are which, and what property should be used to discriminate between them.

### Actual Behavior

Currently each of the available options is deserialised and listed in full and each section doesn't have a title, so while the options are different the user might not know which one they should be choosing. The name of the discriminator is also not displayed anywhere.

### Reproduction steps

```bash
1. Create a spec such as:

  
  {
    [...]
    "paths": {
      "/foo": {
        "put": {
          [...]
          "requestBody": {
            "content": {
              "application/json": {
                "schema": {
                  [...]
                  "discriminator": {
                    "propertyName": "type",
                    "mapping": {
                      "foo": "#/components/schemas/foo",
                      "bar": "#/components/schemas/bar"
                    },
                    "oneOf": [
                      {
                        "$ref": "#/components/schemas/foo"
                      },
                      {
                        "$ref": "#/components/schemas/bar"
                      }
                    ]
                  }
                }
              }
            }
          }
        }
      }
    },
    "components": {
      "schemas": {
        "foo": {
          [Some fields go here...]
        },
        "bar": {
          [Some fields go here...]
        }
      }
    }
  }  
  ```
2. Each of the fields within `#/components/schemas/foo` and `#/components/schemas/bar` will be listed but it won't show which is `foo` and which is `bar`.
3. The propertyName `type` will also not be diaplyed
```


### Link to site / OpenAPI Spec

_No response_

### Version

1.13.4

### What browsers are you seeing the problem on?

Chrome

hanspagel:
Thanks for reporting! I thought this is rendered properly, will check.,

hanspagel:
Ah, it’s something else. I’ve just learned about the discriminator. I’m leaving this example spec here:

https://sandbox.scalar.com/e/fUEFn

cpbr-dev:
Adding to this as I've just stumbled upon this issue within the documentation I'm working on.
Here is how Redoc handles it https://redocly.com/docs/resources/discriminator/

![image](https://github.com/scalar/scalar/assets/30962666/77f25ab3-5403-4e08-b624-b54943a4a259)

Selecting one of the options will change the rendering to show only its parameters:

[...]

PR #5491, opened 2025-04-25
fix(api-reference): title and higher level discriminator support

### LINKED ISSUE #5447: anyOf in array in nested object does not show type definitions
### What happens?

I have an array where mixed items are allowed. When this array is in an object, then the type is not showed.

The type should be showed.

### OpenAPI Document

https://sandbox.scalar.com/e/y6zb3

### LINKED ISSUE #5488: api-reference does not support `anyOf[number].title`
### What happens?

Scalar takes the name of the first key in the object

Image

In Redocly, as an example where it works, value of "anyOf[number].title" changes the tab's label; on the screenshot where you see: Any of and two  "title" "title" tabs

### OpenAPI Document

https://sandbox.scalar.com/e/LaRjM

antlio:
**Problem**

currently when any of discriminator includes a title with value, it is not retrieved within the tab.

**Solution**

this pr adds:
- support of title value that fixes #5488 
- update schema discriminator logic that fixes #5447
- dropdown for long discriminator content usage follow up #5429
- style enhancement finalizing #795

`title`
| before | after |
| -------|------|
| image | image |

`deeper level`
| before | after |
| -------|------|
| image | image |

`long content support`
| before | after |
| -------|------|
| image | image |

**Checklist**

### DIFF
--- a/packages/api-reference/src/components/Content/Schema/SchemaProperty.vue
+++ b/packages/api-reference/src/components/Content/Schema/SchemaProperty.vue
@@ -334,7 +334,7 @@ const displayPropertyHeading = (
   color: var(--scalar-color-1);
   display: flex;
   flex-direction: column;
-  padding: 12px 8px;
+  padding: 8px;
   font-size: var(--scalar-mini);
   position: relative;
 }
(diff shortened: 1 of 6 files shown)


packages/api-reference/src/components/Content/Schema/SchemaProperty.vue
```diff
@@ -314,7 +314,7 @@ const displayPropertyHeading = (
           typeof optimizedValue.items === 'object' &&
           discriminator in optimizedValue.items &&
           Array.isArray(optimizedValue.items[discriminator]) &&
-          level < 3
+          level < 5
```
antlio, line 317: any cons in passing directly optimizedValue here?

hanspagel: Idk, give it a try!

hanspagel (approved):
great improvement

PR #5503, opened 2025-04-28
feat(api-client): select all option to the auth scopes

antlio:
**Problem**

currently it is not possible to select all auth scope at once.

**Solution**

this pr adds a select all button when there are more than 4 scopes to make it easier to quickly catch them all™.

| after |
| -------|
| <video src="https://github.com/user-attachments/assets/06cb1655-e700-45c2-bebc-bf4179fbd682" /> |

**Checklist**


hanspagel (approved):
nice qol feature

PR #5505, opened 2025-04-28
fix(nuxt): don’t use default layout

### LINKED ISSUE #5018: Nuxt: Docs automatically inherit default layout
### What happens?

When using the `@scalar/nuxt` module, the [default layout](https://nuxt.com/docs/guide/directory-structure/layouts#default-layout) is automatically inherited on the docs pages. I believe this to be undesirable, since the docs pages have their own themes, which may clash with an styles inherited in the layout. For a simple header, it may not be so much of an issues, but in more complex apps, it can really affect the docs pages.

Minimal reproduction: https://stackblitz.com/edit/nuxt-scalar-layout-issue?file=layouts%2Fdefault.vue

Image

I don't think the docs pages should automatically inherit the default layout. I'm not sure how the pages are implemented, but Nuxt allows the [layout to be overwritten per route](https://nuxt.com/docs/guide/directory-structure/layouts#overriding-a-layout-on-a-per-page-basis). 

Ideally, there should be an option to choose a layout for the docs pages to inherit (if desired by the developer). 

### OpenAPI Document

N/A

luc122c:
Fixes #5018

**Problem**

Currently, the Scalar page inherits the default Nuxt layout.

**Solution**

This PR turns off the layout, leaving all the styling to Scalar.

Reference: https://nuxt.com/docs/api/utils/define-page-meta#defining-layout

**Checklist**

### DIFF
--- a/.changeset/dull-tigers-count.md
+++ b/.changeset/dull-tigers-count.md
@@ -1,0 +1,5 @@
+---
+'@scalar/nuxt': patch
+---
+
+don't inherit the default layout
--- a/integrations/nuxt/playground/layouts/default.vue
+++ b/integrations/nuxt/playground/layouts/default.vue
@@ -1,0 +1,25 @@
+<template>
+  <div>
+    <header>
+      <p>This is an example header</p>
+    </header>
+    <NuxtPage />
+  </div>
+</template>
+
+<style>
+header {
+  width: 100%;
+  height: 50px;
+  color: white;
+  background-color: black;
+  font-family: sans-serif;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  border: hotpink 1px solid;
+  p {
+    margin: 0;
+  }
+}
+</style>
--- a/integrations/nuxt/playground/nuxt.config.ts
+++ b/integrations/nuxt/playground/nuxt.config.ts
@@ -1,0 +1,14 @@
+// https://nuxt.com/docs/api/configuration/nuxt-config
+export default defineNuxtConfig({
+  compatibilityDate: '2025-05-11',
+
+  devtools: { enabled: false },
+
+  modules: ['@scalar/nuxt'],
+
+  nitro: {
+    experimental: {
+      openAPI: true,
+    },
+  },
+})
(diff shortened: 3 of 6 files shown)


hanspagel (approved):
omg, yes, thanks!

luc122c:
I haven't had chance to test this yet as I can't get the playground running 😢  It *should* work.

amritk (changes requested):
The rest should be good

amritk:
Alright it shouldn't be imported but we do need to sort something out as it isn't aware its a page

marclave (approved):
awesome work @luc122c !

@amritk @hanspagel are we good to get this merged? :)

amritk:
The types don't quite work like that. This is the issue with "magic" imports

marclave:
@luc122c can you investigate why the types are failing for one of the tests :)

Saganic:
Could you go one step further and allow for `nuxt.config` overrides for the layout value?

e.g 

```ts
//nuxt.config.ts

scalar: {
   layout: 'my-custom-layout' || false
}

```

luc122c:
> Could you go one step further and allow for `nuxt.config` module option overrides for the layout value?

@Saganic This would certainly be possible if we can figure out the issue with the virtual imports.

luc122c:
According to this [Vue School video](https://vueschool.io/lessons/extend-and-alter-nuxt-pages?friend=nuxt) at 5:03, we should be manually importing the utils we need from `#imports`.

Perhaps the issue is with the type check? Maybe `#imports` is not being generated correctly before the typecheck?

amritk:
Found a similar issue here

https://discord.com/channels/473401852243869706/1042756264205230081/threads/1047109438961627167

Should we try the same thing? Does layout exist in the meta property on the NuxtPage objects you add in extendPages,

luc122c:
@amritk I did see that, but I wasn't sure if it was the same 'meta' property. Worth a try though!

amritk:
@Saganic alright I just modified it so you can pass in a layout!

PR #5506, opened 2025-04-28
chore: don’t rely on void.scalar.com for tests

hanspagel:
**Problem**

Past me used https://void.scalar.com in the tests. https://void.scalar.com is down. Main is broken. No good.

**Solution**

Present me removed https://void.scalar.com from the tests.

**Checklist**


antlio (approved):
wow

PR #5507, opened 2025-04-28
chore: remove cloudbuild and example servers files

geoffgscott:


hanspagel (commented):
we still need the nextjs example (but as a playground in the package)

also this also has the changes from the #5504 PR, eh?

PR #5513, opened 2025-04-29
fix(api-client): mask password input but don't define it as a password

cameronrohani:
currently we use input type="password" to mask password credentials 

this causes browsers to trigger their password autofill's which is _generally_ intended for user passwords and not quite intended for our use cases of passwords in our data tables.

now when someone uses type="password" on password inputs it will apply change the type to text and apply the  class `scalar-password-input` which overrides the characters to be discs, giving it the effect of a password input. 

there's a better way to go about doing this that maybe @hwkr or @antlio can help me with in the future :) but we need to get this in for now 

before: 
image

after:
image

### DIFF
--- a/packages/api-client/src/components/DataTable/DataTableInput.vue
+++ b/packages/api-client/src/components/DataTable/DataTableInput.vue
@@ -59,11 +59,7 @@ const handleBlur = () => {
 }
 
 const inputType = computed(() =>
-  props.type === 'password'
-    ? mask.value
-      ? 'password'
-      : 'text'
-    : (props.type ?? 'text'),
+  props.type === 'password' ? 'text' : (props.type ?? 'text'),
 )
 
 // If not an enum nor read only, focus the code input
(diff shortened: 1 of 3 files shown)


packages/api-client/src/components/DataTable/DataTableInput.vue
```diff
@@ -118,13 +115,14 @@ const handleLabelClick = () => {
...
             description && 'pr-8',
+            type === 'password' && 'scalar-password-input',
           ]"
           :description="description"
+          :lineWrapping="Boolean(lineWrapping)"
           disableCloseBrackets
           disableTabIndent
           :envVariables="envVariables"
           :environment="environment"
-          :lineWrapping="Boolean(lineWrapping)"
```
antlio, line 127: won't this remove long content display for table input?

cameronrohani: cursor removed that for some reason, I added it back

geoffgscott (approved):

antlio (approved):
•••••

marclave (approved):

PR #5517, opened 2025-04-29
feat: move to ESBuild for most packages

geoffgscott:
- Moves to esbuild from rollup for most things
- Line count increase is from the conversion of .json schemas to .ts files using a codegen script
- Moved away from nodeNext for most packages. In our case nodeNext was the worst of both worlds. We had to have explicit path imports but we were using .ts so it still required build path transforms


hanspagel (commented):
I’m pretty sure we want NodeNext, but I’m ooo today. Pls don’t merge.

amritk (changes requested):
I'm with hans, i'm pretty sure we need nodeNext for when we do SSR. we have been slowly moving everything over

> NodeNext is a resolution mode designed for Node’s specific implementation of co-existing ESM and CJS

hanspagel:
**TL;DR**

* It’s fine, 
* but we still want to use the stricter TSConfig, 
* though we can move away from NodeNext again,
* thanks to ESBuild and tsc-alias 

**Explanation**

We used Rollup with:

```bash
scalar-build-rollup
```

And `tsc-alias`:

```json
  "compilerOptions": {
    "paths": {
      "@/*": ["./src/*"],
      "@test/*": ["./test/*"]
    }
  },
```

And when we had:

```
    "module": "ESNext",
    "moduleResolution": "Bundler"
```

The build had:

```diff
// no file extension, breaks userland projects with `NodeNext`
-import type { MockServerOptions } from './types';
```

But with esbuild we get:

```diff
// with file extension, doesn’t break userland projects with `NodeNext`
+import type { MockServerOptions } from './types.js';
```

And for Vite projects, we can use:

```json
"resolveFullPaths": true
```

which will work great, too.

Case closed, we can leave `NodeNext` behind us and go back to:

```json
    "module": "ESNext",
    "moduleResolution": "Bundler"
```

amritk (approved):
big changes always suck! hopefully we don't have to go back (should be fine)

PR #5520, opened 2025-04-29
chore(prettier): add aspnetcore build directories to prettierignore

xC0dex:
During `pnpm format` I noticed that prettier also formats generated files inside `bin` and `obj` folder. We already have this rule for biome, let's add it to prettier.


hanspagel (approved):
yess!

PR #5521, opened 2025-04-29
feat(api-reference): add classes to intro cards

amritk:
**Problem**

Currently, we cannot target the intro cards with css in case we want to hide them.

**Solution**

With this PR we add a unique class to each of the intro cards.

**Checklist**


antlio (approved):
`introduction-card-item` not enough? or could prefix it? but still nice!

amritk:
they are all `introduction-card-item`, needed classes for specific targetting. Should we do
`introduction-card-item-xxxx` ?

PR #5528, opened 2025-04-30
feat(api-reference): CSS class for HTTP methods

jlnslv:
Added HTTP method-specific CSS classes for custom styling

**Problem**

Currently, there's no possibility to customize http methods of endpoints in the sidebar. They all have to same css class.

**Solution**

With this PR I added a specific css class per http method so that they can be customized.

**Checklist**

### DIFF
--- a/.changeset/poor-apricots-chew.md
+++ b/.changeset/poor-apricots-chew.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
+feat: CSS class for HTTP methods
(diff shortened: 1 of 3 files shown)


hanspagel (changes requested):
great addition! if we/you rely on those classes for styling, we should add tests to confirm they are actually there. you can add them, or I do it soon-ish. ✌️

jlnslv:
Unfortunately, I wouldn't know how to add such tests. Therefore, I would be very grateful if you could add them. Thanks :)

hanspagel (approved):
added tests ✅ 🤝

PR #5529, opened 2025-04-30
Feat/store

hanspagel:
**Tasks/Ideas**

Not all are necessary to get the store in. We need to identify the critical ones.

- [ ] less dependencies (maybe, we can just have `@vue/reactivity` eventually?)
  - `api-client` is just there for the benchmarks comparing it to the old store
  - `openapi-parser` is just there for upgrading
- [ ] watch `Ref`’s
- [ ] add more hooks, and:
- [ ] add a plugin to get performance metrics
- [ ] add a logging plugin (for the browser console)
- [ ] add an IndexDB plugin (use idb?)
- [ ] write a test that loads up 100 real-world examples to verify
  - we’re fast (requires the perf plugin)
  - they all pass, nothing gets lost
- [ ] `workspace.create('myCollection', data)` (data can be a string, object, ref, function, (async) function )
- [ ] `workspace.delete(myCollection)`
- [ ] manually update a collection: `workspace.update(myCollection, newValue)`
- [ ] persist changes (currently, we only persist the initial state)
- [ ] find a way to run the old store and the `helper/parse.ts` in parallel
- [ ] proof of concept for migrating @scalar/api-reference
- [ ] add “views” (a `computed()`) on top of the data store: to retrieve all tags, with a default tag, with tag groups, tagsSorter …
- [ ] the OpenAPI Zod schemas are a 1:1 translation of the specification, but too strict for real-world documents, how do we deal with that? copy the schemas and add `optional()` and `catch()` everywhere?
- [ ] We may want a `createWorkspace({ readOnly: true })`
- [ ] do we want mutations again?

Not sure, but I think we said we have to (= it’s required) to:

- [ ] rewrite the sidebar to the new data structure:

Super early ideas:

- [ ] `createPartialCollection<Operation>({ summary: 'This is the same reactive collection, but just for a specific OpenAPI object' })` (helps with the transition? might be cool for Scalar Blocks?)

**Problem**

We’re using a reactive data store (mostly in @scalar/api-client), which helped us to learn a lot. There are few limitations we hit, though:

* Updates to the document (with multi documents, watch mode and other features) isn’t easy
* Exporting data isn’t easy (we lose the structure of the original document)
* All the heavy work is done before rendering, which adds a waiting time for big documents before rendering
* Rendering can become tricky, as we need to take the whole workspace structure into account everytime we interact with the store (e.g. showing all servers requires you to filter for the ones for the “active” collection).

**Solution**

With this PR we’re introducing a new, fully OpenAPI-based reactive data store. :) Here are the main ideas:

* It’s resolving `$ref`’s on the fly (not all at once)
* The data structure is 1:1 the OpenAPI 3.1 specification, exporting is easy now and updating is easy, too (no UUIDs)
* It has a few additional cache layers to speed things up

**Checklist**


amritk:
new store whooooooooooooooooooooooooo

packages/store/README.md
```diff
@@ -0,0 +1,223 @@
...
+    },
+  },
+})
+
+// Both paths access the same data
+const person = collection.document.components.schemas.Person
+const user = collection.document.components.schemas.User
+
+// Changes through either path update the source
+user.properties.age = { type: 'number' }
```
amritk, line 80: okay so we probably don't want to do this, make it immutable and use mutators instead

amritk (commented):
Didn't go too deep but glanced over and left some comments. Pretty sweet!. We should iron a couple things out.

- we should have tests for full reactivity for live sync, handle the whole spec + some conflict resolution
- figure out how we want to do additional data like auth
- figure out if we want to support [overlays](https://www.openapis.org/blog/2024/10/22/announcing-overlay-specification)

amritk (approved):
Alright sick PR, looks pretty good as far as I can tell. I just made some minor improvements:

- switched to stricter tsconfig (should be our default moving fwd)
- removed a lot of the type casting (no need most of the time and it kills type safety)
- moved to instance based maps to prevent memory leaks
- added another lookup cache for even more perf
- didn't wanna bother hans too much but I would really like to type it further and get rid of all those unknowns since we are working with openapi objects after all, unless we want to support all objects

in the future if we needed we can probably improve the queue to avoid the chunking (claude provided an example) and use web workers for true multi threading, but we don't really need it rn since its all async calls and the processing is minimal.

amritk (changes requested):
Spoke to soon, I found a file resolution issue

PR #5539, opened 2025-05-01
feat(themes): Unify theme interface

geoffgscott:
Currently we have a fragmented theme approach between org and oss. This aligns all the theme logic + styles into the @scalar/themes package.  Some unused methods are removed and any style extension from scalar-org were integrated into the base themes.


packages/themes/src/index.ts
```diff
@@ -35,12 +72,7 @@ export const themeIds = [
   'none',
```
antlio, line 72: this doesn't seem to do much and is breaking theme when used, we might want to remove it?

antlio (approved):
great pull! tested throughout the apps and didn't catch any issue - just one nitpick on the `none` id but lgtm!

packages/themes/src/index.ts
geoffgscott: Yeah @hwkr @hanspagel is there a reason for this? I just left it as is from before.

hwkr: This is used to override the default option for the references but it probably only actually needs to be in the refs

PR #5555, opened 2025-05-02
feat(platform): Adding local repo CLI and improving tests

geoffgscott:
Adds a local CLI used for composing repo tasks (and avoids bash scripts)


bgrcs (approved):

geoffgscott:
Merging as the JSDelivr failure looks related to mobile sidebar and nothing in here. 
No changeset as the catalog name was changed but not the contents. Everything else is CI changes.

PR #5556, opened 2025-05-02
feat(Scalar.AspNetCore): support for custom js modules

xC0dex:
This PR brings support for custom JS configuration modules. This allows the user to use all JS configuration options like the `generateOperationSlug` method. The .NET integration itself cannot serialize a C# method or a string to a JS function. Moreover, the DX would be bad. 

Now the use can provide a path to a JS file (which contains a default module) and Scalar will use it to overwrite the existing config with the given one. The setup is straightforward (checkout the changes in the playground project).

**Todo**
- Docs
- Changeset

**Checklist**


xC0dex:
Does `ModulePath` sound ok? Or should we name it `CustomModulePath` or so?

hanspagel:
What if we call it …

```csharp
{
    options.WithJavaScriptConfiguration("/scalar/config.js");
    // or
    options.JavaScriptConfiguration = "/scalar/config.js";
});
```

xC0dex:
> What if we call it …
> 
> ```cs
> app.MapScalarApiReference(options =>
> {
>     options.WithJavaScriptConfiguration("/scalar/config.js");
>     // or
>     options.JavaScriptConfiguration = "/scalar/config.js";
> });
> ```

Awesome, I'll rename it!

xC0dex:
Done 👍

PR #5564, opened 2025-05-02
fix(api-reference): config servers change

amritk:
**Problem**

Currently, we moved to diffing on config changes. However servers is an override so we want to remove the old ones and just add the new ones.

**Solution**

With this PR we remove then re-add servers on change. (we can diff this later but need to be smarter on addition)

**Checklist**

### DIFF
--- a/.changeset/sour-eels-whisper.md
+++ b/.changeset/sour-eels-whisper.md
@@ -1,0 +1,5 @@
+---
+'@scalar/oas-utils': patch
+---
+
+chore: export getServersFromOpenapi function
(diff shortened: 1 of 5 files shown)


marclave:
why do we need the parsed spec?

amritk:
@marclave servers is actually an override so we never import the spec servers if this config is present. 

Now when we remove it, I'm adding the servers back from the spec. We could actually just pass the parsedSpec.servers instead

hanspagel (approved):
image

Issue #5391, opened 2025-04-11
Asp.Net Core Integration: Couldn’t find a [data-spec], [data-spec-url] or <script id="api-reference" /> element


HHaoWang:
### What happens?

After updating the Scalar package for ASP.NET Core from version 2.0.x to 2.1.x, the Scalar API documentation page cannot be opened properly. The page displays a blank screen, and the console shows the following error: scalar.js:27250  
Couldn’t find a [data-spec], [data-spec-url] or <script id="api-reference" /> element. Try adding it like this: <div data-spec-url="https://cdn.jsdelivr.net/npm/@scalar/galaxy/dist/latest.yaml" />

In programs.cs:

builder.Services.AddOpenApi(options => { options.AddDocumentTransformer<BearerSecuritySchemeTransformer>(); });
WebApplication app = builder.Build();
if (app.Environment.IsDevelopment())
{
    app.UseDeveloperExceptionPage();
    app.MapOpenApi();
    app.MapScalarApiReference(options =>
    {
        options.WithPreferredScheme("Bearer");
        options.DefaultFonts = false;
    });
}

### What did you expect to happen?

The Scalar API documentation page can be opened properly.

HHaoWang:
The screenshot is shown below:
![Image](https://github.com/user-attachments/assets/c78c3981-6ffa-4748-822d-704098f5f720)

HHaoWang:
# Here is original HTML content
```
<!doctype html>
<html>
<head>
    <title>Scalar API Reference</title>
    <meta charset="utf-8" />
    <meta name="viewport" content="width=device-width, initial-scale=1" />
    
</head>
<body>
    
    <div id="app"></div>
    <script type="module" src="scalar.aspnetcore.js"></script>
    <script type="module" src="scalar.js"></script>
    <script type="module">
        import { initialize } from './scalar.aspnetcore.js'
        initialize(
        '/scalar/v1',
        false,
        {"showSidebar":true,"hideModels":false,"hideDownloadButton":false,"hideTestRequestButton":false,"darkMode":true,"hideDarkModeToggle":false,"defaultHttpClient":{"targetKey":"shell","clientKey":"curl"},"authentication":{"preferredSecurityScheme":"Bearer"},"withDefaultFonts":false,"defaultOpenAllTags":false,"theme":"purple","layout":"modern","_integration":"dotnet","hideClientButton":false,"sources":[{"title":"v1","url":"openapi/v1.json"}]})
    </script>

[...]

PR #5568, opened 2025-05-04
fix(Scalar.AspNetCore): improve module script loading

xC0dex:
**Problem**

Currently, the API Reference script is loaded as `type=module` even if we don't use any module feature here. This _may_ cause some issues in browsers. 

**Now**

This PR removes the `type=module` attribute, so the script is loaded like a default JS script.

**Checklist**

### DIFF
--- a/.changeset/pink-dolphins-rhyme.md
+++ b/.changeset/pink-dolphins-rhyme.md
@@ -1,0 +1,5 @@
+---
+'@scalar/aspnetcore': patch
+---
+
+fix(Scalar.AspNetCore): improve script module loading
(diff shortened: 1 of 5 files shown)


worldofchu:
> **Preview Examples**
> 
> https://7fb7aed3-a137-410d-8a0e-7a8e868caab4--scalar-deploy-preview.netlify.app

Firefox 138.0.1 x64  #5391    you can test https://github.com/worldofchu/scalartest.git

hanspagel (commented):
but we still use `<script type="module">`, does that PR actually fix an issue then? 🤔

worldofchu:
> but we still use `<script type="module">`, does that PR actually fix an issue then? 🤔
> 
> EDIT: Ah, just saw the above comment.

It's  not  ok!

PR #5569, opened 2025-05-05
refactor(Scalar.AspNetCore): compress embedded resource to reduce the Scalar.AspNetCore.dll size

### LINKED ISSUE #5604: Compression support
### What happens?

Currently, the static assets are not compressed, which means that the browser has to load the raw file instead of a pre-compressed Brotli (`.br`) or GZIP (`.gz`) file.

stratosblue:
**Problem**

Currently, the size of `Scalar.AspNetCore.dll` is about 3000KB. The reason why the file is so large is that the embedded resources are relatively large.

**Solution**

Using gzip to compress the `StaticAssets/*` at build time. And decompress it before response it.
This change can compress the size of the`Scalar.AspNetCore.dll` from 3000KB to 1000KB.

Closes #5604

**Checklist**

### DIFF
--- a/.changeset/empty-plants-own.md
+++ b/.changeset/empty-plants-own.md
@@ -1,0 +1,5 @@
+---
+'@scalarapi/api-reference': patch
+---
+
+feat(Scalar.AspNetCore): Support GZip compression for embedded ressources
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -575,6 +575,10 @@ jobs:
         working-directory: integrations/aspnetcore
         run: pnpm copy:standalone
 
+      - name: Compress static assets
+        working-directory: integrations/aspnetcore
+        run: pnpm compress
+
       - name: Restore dependencies
         working-directory: integrations/aspnetcore
         run: dotnet restore
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -653,6 +653,11 @@ jobs:
         run: pnpm copy:standalone
 
       - if: steps.changed-files.outputs.aspnetcore_package_any_changed == 'true'
+        name: Compress static assets
+        working-directory: integrations/aspnetcore
+        run: pnpm compress
+
+      - if: steps.changed-files.outputs.aspnetcore_package_any_changed == 'true'
         name: Restore dependencies
         working-directory: integrations/aspnetcore
         run: dotnet restore
--- a/integrations/aspnetcore/package.json
+++ b/integrations/aspnetcore/package.json
@@ -16,6 +16,7 @@
   },
   "scripts": {
     "copy:standalone": "cp ../../packages/api-reference/dist/browser/standalone.js ./src/Scalar.AspNetCore/StaticAssets/scalar.js",
+    "compress": "cd ./src/Scalar.AspNetCore/StaticAssets && gzip --keep --verbose --force scalar.js scalar.aspnetcore.js",
     "format": "scalar-format",
     "format:check": "scalar-format-check",
     "lint:check": "scalar-lint-check",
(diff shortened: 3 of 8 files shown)


hanspagel:
OMG, so cool! Waiting for @xC0dex to review.

xC0dex:
Hey @stratosblue,

Thank you for the PR and for looking into reducing the size of dll!

We generally try to keep our dependencies to a minimum wherever possible. Additionally, runtime performance is our main priority for this package.
Out of curiosity, have you encountered any significant runtime issues or drawbacks specifically related to large .dll sizes? It would be great to hear if there’s a concrete scenario where this change makes a noticeable difference.

For reference, here are some benchmark results comparing the default and GZIP-decompressed stream approaches:

| Method                         | Mean       | Error    | StdDev   | Ratio | RatioSD | Allocated | Alloc Ratio |
|------------------------------- |-----------:|---------:|---------:|------:|--------:|----------:|------------:|
| DefaultReadStreamSync          |   638.9 ns | 10.76 ns |  9.54 ns |  1.00 |    0.02 |    2.7 KB |        1.00 |
| GZipDecompressedReadStreamSync | 6,993.4 ns | 29.48 ns | 26.13 ns | 10.95 |    0.17 |   3.04 KB |        1.12 |

As you can see, decompression adds quite a bit of _unnecessary_ overhead.

Anyway, this PR brings our attention to the topic of compression in general, thanks!

stratosblue:
> We generally try to keep our dependencies to a minimum wherever possible. Additionally, runtime performance is our main priority for this package. Out of curiosity, have you encountered any significant runtime issues or drawbacks specifically related to large .dll sizes? It would be great to hear if there’s a concrete scenario where this change makes a noticeable difference.

Hi @xC0dex ,

Thank you for your reply.

The starting point for this requirement is a single file publishing scenario, such as `PublishSingleFile` or `PublishAot`, the difference in file size of 2MB is a considerable size for single file distribution. For other distribution scenarios, the smaller the file is better too.

As for the performance issue, at the beginning of writing the code, I believed that the main usage scenario of this project was during development, and the related files included client-side caching processing, so performance may not need to be considered here. So for convenience, I directly performed GZip decompression every time I returned. If performance needs to be considered, there are some methods to alleviate it: 
- Cache decompressed content: This will only decompress on the first request, and there will be no performance impact in the future 
- Directly writing gzip data: When the client supports gzip (or the server enforces compression), the compressed content is directly written into the response, so there is no need to decompress and performance may be better. Currently, mainstream browsers support gzip. The code is like (of course, we still need to do some work to ensure compatibility):
  ```C#
  httpContext.Response.Headers.ContentEncoding = "gzip";

  return ifNoneMatch == etag ? Results.StatusCode(StatusCodes.Status304NotModified) : Results.Stream(resourceFile.CreateReadStream(), MediaTypeNames.Text.JavaScript, entityTag: new EntityTagHeaderValue(etag));
  ```

stratosblue:
Now when the client supports gzip, it will directly use gzip response without any additional decompression overhead (and this is the effect in most cases), and it will also have a smaller response size.

xC0dex:
> Now when the client supports gzip, it will directly use gzip response without any additional decompression overhead

This is awesome!


I think for the compression itself we'll probably use the built-in gzip CLI feature of Ubuntu (our GitHub Runner). And we have to fully cover this with tests. 

Have you noticed any performance improvement by using `[MethodImpl(MethodImplOptions.AggressiveInlining)]`?

xC0dex:
This is what I'd push to your PR later https://github.com/scalar/scalar/commit/3eda84d37665670266165ae1e08cdc4de456b9b4 Once I fixed the tests I'd say your PR is ready! 🚀

stratosblue:
> I think for the compression itself we'll probably use the built-in gzip CLI feature of Ubuntu (our GitHub Runner). And we have to fully cover this with tests.

Except for some additional operations that may be required to obtain compressed files when developing on Windows, everything else is fine

> Have you noticed any performance improvement by using `[MethodImpl(MethodImplOptions.AggressiveInlining)]`?

It can eliminate some method call overhead (although there is not much improvement in the current scenario, there are also no side effects)

xC0dex (approved):
I know I'm repeating myself, but this PR is awesome! Thanks! 🚀 

I think later we may improve the JS setup to use vite and TS, so the way how to compress will probably change.

hanspagel (approved):

PR #5590, opened 2025-05-06
chore: remove unused scripts

hanspagel:
Some scripts were left over, but are broken now. Let’s remove them.


amritk (approved):
Sure if you like! We don't use docker base anymore?

PR #5591, opened 2025-05-06
feat(openapi-to-markdown): Markdown API Reference

hanspagel:
**Problem**

People use AI today, but LLMs prefer Markdown over beautiful HTML.

**Solution**

This PR adds the new `@scalar/openapi-to-markdown` which generates plain Markdown from OpenAPI files.

The goal is to have an easy tool to add `/llms.txt` to your backend.

**Tasks**

Pretty early, but it should be pretty digestible for LLMs already. 

**Preview**

Comes with a Hono playground that serves a plain HTML page on `/` and a `/llm.txt` with Markdown.

Screenshot 2025-05-06 at 17 23 05

**Checklist**


packages/openapi-to-markdown/src/components/XmlOrJson.vue
```diff
@@ -0,0 +1,23 @@
+<script setup lang="ts">
+import type { OpenAPIV3_1 } from '@scalar/openapi-types'
+// @ts-expect-error Doesn’t come with types
```
amritk, line 3: 🫨 in this day and age?

amritk (approved):
I dunno when you build these things!? nice work as always (didn't test but its new)

hanspagel:
Got lots of ideas, but it’s new, it’s okay if it’s not perfect right-away. Let’s merge it and take it from there.

packages/openapi-to-markdown/src/components/XmlOrJson.vue
hanspagel: yeah, I should look for an alternative. but it’s used in another package already, so that was easy.

PR #5594, opened 2025-05-07
feat: arbitrary query parameters for OAuth authorize + handle Deny request

Ma-ve:
I'm having a harder time overseeing all of the places / occurrences where this can/should happen. I _think_ I've got them all. I could use some fresh set of eyes!

**Problem**

Currently, there is no support for the `prompt=` query parameter. A lot of OAuth2 servers implement this query parameter, with the following values being common:

- `&prompt=login`
- `&prompt=consent`
- `&prompt=none`

The current OAuth implementation also can't handle "Deny" authorization situations.

Other issues/feedback have been spotted saying they'd like to use "audience" too. 

**Solution**

With this PR there's an extension, `x-scalar-security-query`, for setting additional query parameters for the Authorize request to the OAuth2 server:

```
{
  ...
  'x-scalar-security-query': {
    prompt: 'consent',
    audience: 'scalar'
  }
}
```

Per [RFC 6749 4.1.2.1](https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.2.1), the PR also implements handling the Deny authorizations, by checking for `error` as a query or hash parameter, and console warning that error (with the `error_description` if filled too)

**Checklist**


Ma-ve:
> This is so awesome! I was thinking about a more generic approach, like key value based query parameters. 👀 I saw some requirements for `audience` parameters etc.

Yeah, was thinking about that too. Either/or, but that requires more UI work, and I'm not good at that haha.

I'm having a hard time discerning what should be in the scheme, and what should be user input through the UI - as per my, if x-prompt, set prompt to default to x-prompt, but user can override haha.

Ma-ve:
@xC0dex Would it doing this way be OK to get in first, before moving to an `x-additionalQueryParameter` (or whatever) setup, or would it be a definitive no, and we'll have to wait until who implements just that?

amritk:
First of all, awesome PR! tests and all 😍 

Yea I like the idea of making this customize-able to support other query params as well. Also agree with not having minimal UI adjustments, we can start with making it spec only then add the UI if needed.

```ts
export const XScalarSecurityQuery = z.record(z.string(), z.string())
```

this way you can add any/as many query params as you like.

Ma-ve:
> @ [amritk](https://github.com/amritk) 
> First of all, awesome PR! tests and all 😍
> 
> Yea I like the idea of making this customize-able to support other query params as well. Also agree with not having minimal UI adjustments, we can start with making it spec only then add the UI if needed.
> 
> ```ts
> export const XScalarSecurityQuery = z.record(z.string(), z.string())
> ```
> 
> this way you can add any/as many query params as you like.
> 
> Also we like to keep the scalar prefix until it becomes a standard openapi extension

Thanks for the insight! I've reworked x-prompt to x-scalar-security-query! 😎

hanspagel (commented):
just wanted to say, I love this PR

Ma-ve:
> @hanspagel just wanted to say, I love this PR

Thanks! Me too, haha. Who would need to review the code / give approval?

amritk (approved):
And you added error handling as well, amazing! Could be PR of the week

Ma-ve:
Thanks for the help @amritk @hanspagel! I'm loving it 😎 

image

amritk:
So good!

Ma-ve:
@amritk seems like the Scalar Mock server hasn't been updated, the current back button is still a `javascript:history.back()`

https://galaxy.scalar.com/oauth/authorize?response_type=code&redirect_uri=https%3A%2F%2Fdocs.scalar.com%2F&client_id=&state=e5s3l6pv

is that something that's still to be done manually?

amritk:
Oh yea I'll make an issue

Issue #5553, opened 2025-05-02
scalar handles array type incorrectly specifically with multipart


ekristen:
### What happens?

This might happen in other places but I have for sure confirmed this with multipart/form. 

I have an endpoint that has a property `files` with `type` as `array`. This gets rendered as `files[0]` in the scalar UI which is wrong. 

![Image](https://github.com/user-attachments/assets/c2fd9a5a-19ca-4f8c-899c-24faad1f9f43)

It should be `files`. 

The curl statement is even generated incorrectly beyond getting the field wrong, it sends a JSON payload with a multipart/form-data which is not right. It should use the `-F` option on curl.

**However** it does seem to send multipart/form-data PROPERLY when clicking send, so it's just a curl generation that's wrong.

```
curl http://localhost:4242/api/v1/files/upload \
  --request POST \
  --header 'Content-Type: multipart/form-data' \
  --data '{
  "files[0]": "test.txt",
  "owner_id": "01JT76N7T4PSSE646454MHDSCM",
  "owner_type": "case"
}'
```

### What did you expect to happen?

I expect the multipart form to be `files` not `files[0]`. Swagger UI gets this right. 

Swagger generates the CURL statement correct as well 

```
curl -X 'POST' \
  'http://localhost:4242/api/v1/files/upload' \
  -H 'accept: */*' \
  -H 'x-api-key: test' \
  -H 'Content-Type: multipart/form-data' \
  -F 'files=@text1.txt;type=text/plain' \
  -F 'files=@test2.txt;type=text/plain' \
  -F 'owner_id=01JT76N7T4PSSE646454MHDSCM' \
  -F 'owner_type=case'
```

![Image](https://github.com/user-attachments/assets/cda1d396-5cb3-4792-820e-a27a141d3526)

reslear:
By the way, it would be cool to be able to select multiple like in postman


![Image](https://github.com/user-attachments/assets/f30a0afe-40a4-4190-8f5b-2d5165d27fff)
![Image](https://github.com/user-attachments/assets/87ad6d4d-f877-4ca6-8159-6ffa0540918b)

ekristen:
BTW thanks for this library. It's great! I've got it embedded in my API and it's so much better than others.

Issue #4834, opened 2025-02-20
Scalar misinterprets multipart/form-data with part that contains structured JSON


hniksic:
### What happens?

We have an endpoint that accepts a multipart/form-data body with two parts:

* an unstructured file, named `file`
* a structured JSON document, named `props`

The OpenAPI document describes multipart/form-data as `requestBody` with two properties: `file` of type `string` (and format `binary`), and `props` of type `object`, with its own properties. The minimal schema that reproduces this is [attached here](https://github.com/user-attachments/files/18891209/minimal-schema.json), and also [at the sandbox](https://sandbox.scalar.com/e/2Hwa2).

Scalar 0.3.150, which we currently deploy, generates the following request:

```sh
curl URL \
  --request POST \
  --header 'Content-Type: multipart/form-data' \
  --form 'file=' \
  --form 'props[created_at]=null' \
  --form 'props[description]=' \
  --form 'props[name]='
```

It is incorrect because it splits up `props` into multiple form parts instead of as a single parts encoded in JSON. It seems to use [qs](https://github.com/ljharb/qs)-like syntax, which was not requested by the schema and won't be accepted by the endpoint.

The scalar deployed on https://docs.scalar.com/swagger-editor generates the following:

```sh
curl URL \
  --request POST \
  --header 'Accept: */*' \
  --header 'Content-Type: multipart/form-data; boundary=----WebKitFormBoundarybtmt1870GXquBkll' \
  --data "{"file":"","props[name]":"","props[description]":"","props[created_at]":""}"
```

This is also incorrect, and a step back from the older version because:

* it manually specifies form data boundary instead of leaving encoding up to curl
* it sends the whole form in a single JSON-encoded part instead of in two parts
* the keys of `props` are again encoded like with qs

### What did you expect to happen?

Correct request might look like this:

```sh
curl URL \
  --request POST \
  --header 'Content-Type: multipart/form-data' \
  --form 'file=@filename' \
  --form 'props={"name": "", "created_at": "", "description": ""}'
```

It is generating two form parts, one with the raw contents of a file (but `'file=whatever` would also be fine), and the other with `props` initialized to JSON valid for the schema.

While there are already several issues about multipart/form-data, they don't seem to cover the exact same bug. In case it's relevant, [swagger's editor](https://editor-next.swagger.io/) interprets the document as intended.

### OpenAPI Document

https://sandbox.scalar.com/e/2Hwa2

hanspagel:
Great feedback, thanks!

[...]

Issue #4125, opened 2024-12-08
Appropriate Body type neither applies nor prefills in case of multipart/form-data in FastAPI Docs.


vazome:
### What happens?

In OpenAPI spec the `Body` is type of `multipart/form-data` which has a key `images`

<img src="https://github.com/user-attachments/assets/96dbbf84-43b8-47e8-afbe-93f432500e52" width=50% height=50%>

1. When I click on "Test Request" the `Body` is assigned default type `JSON`.
<img src="https://github.com/user-attachments/assets/ecb72e41-a3b1-4e30-89fd-83b39159df32" width=50% height=50%>

2. When select `Multipart Form` in `Body` it is not prefilled with the `images` key, but empty. 
<img src="https://github.com/user-attachments/assets/e888ae87-8764-4ecb-86a8-6a9512255778" width=50% height=50%>

This error is only specific to FastAPI Scalar integration, the behavior is as expected in sandbox.scalar.com

But the issue appears to be fixed in https://github.com/scalar/scalar/issues/3222

### What did you expect to happen?

I expected `Body` to default to the type defined in OpenAPI spec and the keys to be prefilled.
This error is only specific to FastAPI Scalar integration, the behavior is as expected in sandbox.scalar.com it selects correct type and prefills keys.

amritk:
Hey this should work now! Feel free to re-open if it does not!

Issue #3316, opened 2024-09-30
multipart/form-data Body Not Rendering as Expected


dominicmallo:
### What happens?

When rendering a request that expects "multipart/form-data", the body in the Web Client still defaults to "JSON". Likewise, the expected/default fields are not displayed.

### What did you expect to happen?

Configuring a request to consume multipart/form-data should trigger Scalar to use the "Multipart Form" for the request.

### OpenAPI Document

https://petstore.swagger.io/v2/swagger.json

marclave:
hi @dominicmallo this has been fixed in the latest version of the Scalar API Client :) can update it or use client.scalar.com to see it in action

https://github.com/user-attachments/assets/b4c886c8-259e-4b93-920c-b46b30d92233

aoaostar:
multipart/form-data
scalar web ui don't work

aoaostar:
![image](https://github.com/user-attachments/assets/adffa5f9-8c5e-4d32-82d1-cc5da1e8ab01)
![image](https://github.com/user-attachments/assets/0cff0d08-bd50-46aa-be57-583769a4bd65)
![image](https://github.com/user-attachments/assets/a5b55ff7-2bba-4665-b316-a257e3fa112c)

But swagger ui is ok

eo33:
In Scalar, when the `Content-Type` of the request is `multipart/form-data`, the field values are not shown. 
When rendered using Swagger UI, the field values are shown, as pointed out by @aoaostar.

1. Here's what it looks like when rendered with Scalar (via GitBook):
![image](https://github.com/user-attachments/assets/c487201e-99a8-40cf-939f-e204ba07f4b6)

Note: The field values in the body are not shown, and it defaults to JSON. 

2. Here's what it looks like when rendered with Swagger UI:
![image](https://github.com/user-attachments/assets/c82f21ea-5177-43a8-88d8-e8ec63e792ed)

Note: The field values in the body are shown.

Teyik0:
I see this closed but having the same problem here

pfo-omicsstudio:
Same problem here! My multiform data worked fine in Swagger UI but not in Scalar.

antlio:
@Teyik0 @pfo-omicsstudio do you mind sharing a specification document so I can replicate on my end?
thank you so much!

pfo-omicsstudio:
[...]

PR #5595, opened 2025-05-07
fix: render correct queries with form data

nsychev:
**Problem**

Currently, endpoints with `multipart/form-data` body (and likely `application/x-www-form-urlencoded`) are rendered incorrectly: for some reason, all requests send JSON payload.

Details are in #5553, #3316 and #4125.

**Solution**

With this PR, I used `PostDataParams` type from `har-format` when body has form data. Some `snippetz` plugins (for example, `curl`) already support this type and render it correctly:

image

Also, I fixed the way how multipart/form-data body is build:
* before this PR, when array was in the request, field names were assigned with an index, e.g. `field[0]=value0&field[1]=value1&...` (for `application/x-www-form-urlencoded`)
* after this PR, all fields should be named the same, e.g. `field=value0&field=value1`. This matches default HTML forms behavior.

There is currently no standard about expected behavior when field contains some complex object (see #4834), I think this issue should be addressed separately.

**Checklist**

### DIFF
--- a/.changeset/sweet-chicken-rule.md
+++ b/.changeset/sweet-chicken-rule.md
@@ -1,0 +1,7 @@
+---
+'@scalar/api-client': patch
+'@scalar/oas-utils': patch
+'@scalar/types': patch
+---
+
+fix: render correct queries with form data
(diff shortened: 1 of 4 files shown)


nsychev:
Hi!

As I see in test run, there are already tests to check all of that (but they expect JSON in multipart requests). I'll fix them today and add test for multiple files.

nsychev:
All required tests were already implemented, fixed them.

hanspagel (approved):
niiice

hanspagel (approved):
Niiice

PR #5600, opened 2025-05-08
feat(api-reference): persist auth in local storage

### LINKED ISSUE #5320: Provide the ability to persist Authentication credentials in localStorage / cookies
### Discussed in https://github.com/scalar/scalar/discussions/4704

<sup>Originally posted by **risen228** February  6, 2025</sup>
Now the only way to save the auth token between page refreshes is to specify it in options on server side.

This is inconvenient and adds various problems: you either need to set a common token during development and add logic that does not add this token in other environments, or each developer should, for example, set their own token through env variables so that it gets into scalar ui

There is a very simple solution to this problem, which allows you not to touch the server code at all - when the developer has specified the token, it can be saved in localStorage / cookie and after reloading the page, we just take it from there and fill in the field. You can also make this behavior optional and provide a checkbox with text like "Remember the value".

However, if we add compatibility with the server token value and support for multiple authentication schemes, the logic becomes a bit more complicated:
- Check if there is a locally saved value, if there is - use it, if not - check if there is a token value from the server, if there is - use it, otherwise - empty field
- For multiple authentication schemes, you need to store values with prefixes. The resulting key could be, for example, `scalar/authentication/Bearer/token`, where `Bearer` is the name of the authentication scheme.

This way, you don't have to add any logic to the server, and each developer can use their own token by simply adding it to the interface.</div>

amritk:
**Problem**

Currently, we do not save auth in references. Users must enter it everytime if they refresh.

**Solution**
- added auth persistance for security schemes + selection
- updated docs with all missing properties including the new persist auth

closes #5320
closes #5598
closes #4704

**Checklist**

### DIFF
--- a/packages/api-client/src/views/Request/RequestSection/RequestAuth/RequestAuthTab.vue
+++ b/packages/api-client/src/views/Request/RequestSection/RequestAuth/RequestAuthTab.vue
@@ -7,7 +7,9 @@ import type {
   Server,
 } from '@scalar/oas-utils/entities/spec'
 import type { Workspace } from '@scalar/oas-utils/entities/workspace'
+import { isDefined } from '@scalar/oas-utils/helpers'
 import type { Path, PathValue } from '@scalar/object-utils/nested'
+import type { Entries } from 'type-fest'
 import { capitalize, computed, onMounted, ref } from 'vue'
 
 import { DataTableCell, DataTableRow } from '@/components/DataTable'
@@ -43,7 +45,8 @@ const {
 }>()
 
 const storeContext = useWorkspace()
-const { securitySchemes, securitySchemeMutators } = storeContext
+const { collectionMutators, securitySchemes, securitySchemeMutators } =
+  storeContext
 const security = computed(() =>
   securitySchemeUids.map((uid) => ({
     scheme: securitySchemes[uid],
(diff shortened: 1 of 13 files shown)


cameronrohani:
this PR going to get over 9000 likes on twitter when i post it for this week at scalar 

image

packages/api-client/src/views/Request/RequestSection/RequestAuth/RequestAuthTab.vue
```diff
@@ -113,8 +180,8 @@ const dataTableInputProps = {
         :aria-label="scheme.description"
         class="text-c-2 auth-description-container group/auth -mb-0.25 flex items-center whitespace-nowrap outline-none hover:whitespace-normal">
         <ScalarMarkdown
-          :value="scheme.description"
-          class="auth-description z-1 bg-b-1 text-c-2 outline-b-3 top-0 line-clamp-1 h-full w-full px-3 py-1.5 group-hover/auth:line-clamp-none" />
+          class="auth-description z-1 bg-b-1 text-c-2 outline-b-3 top-0 line-clamp-1 h-full w-full px-3 py-1.5 group-hover/auth:line-clamp-none"
```
amritk, line 183: just lint, unrelated

marclave (approved):
awesome 😎

PR #5619, opened 2025-05-12
feat(api-client): add fallback handling for +json MIME types

### LINKED ISSUE #5586: Add support for 'application/fhir+json' mime types
### What happens?

Fhir json responses are still being rendered as binary file.
I don't see any reason why they can't be rendered as a json file.

Since these files follow the normal json rules, they should be rendered in the UI

### Documentation
The MIME-type for this format is `application/fhir+json`
Source: https://build.fhir.org/json.html

The 'Media Type (MIME Type) Definitions' list should probably be extended with this line:
```
'application/fhir+json': { extension: '.json', raw: true, language: 'json' }
```

### OpenAPI Document

_No response_

inyourtime:
This PR introduces fallback support for MIME types that end in `+json`, which are not explicitly listed in the `mediaTypes` map. These types now correctly return a default JSON configuration via the `getMediaTypeConfig` function.

closes #5586

### DIFF
--- a/.changeset/funny-toys-peel.md
+++ b/.changeset/funny-toys-peel.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
+
+feat: support fallback for +json MIME types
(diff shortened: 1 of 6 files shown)


hanspagel (approved):
great PR, as always 👏

PR #5620, opened 2025-05-12
chore: drop Node 18 File polyfill

hanspagel:
There were a few leftovers to support Node 18, this PR removes them:

* We skipped one test for Node 18
* We had a polyfill for the File class

But hey, Node dropped support for version 18, so should we.


amritk (approved):
Good riddance

marclave (approved):

PR #5626, opened 2025-05-12
docs: add the integration docs for gofiber 

yokeTH:
**Problem**

Currently, it lack of gofiber example and integration docs

**Solution**

With this PR i added the document and example of my module [gofiber-scalar](https://github.com/yokeTH/gofiber-scalar) it's [gofiber](https://github.com/gofiber/fiber) middleware that integrated the scalar

**Checklist**

I’ve gone through the following: (I checked only 2 because it just docs)


amritk:
Tagging in our gopher @marclave

hanspagel (changes requested):
OMG SO COOL

We actually want to get rid of the /examples folder eventually, and I don’t think we need to have/maintain an example in our repository. So I think we can just add to the documentation BUT we should also add it to the list of integrations in the root README.

Wdyt?

yokeTH:
I removed the example from project

hanspagel:
You’re so right 🧠

hanspagel (approved):

yokeTH:
It appears to be “5 workflows awaiting approval.

PR #5630, opened 2025-05-13
fix(api-client): add aria label to request table code inputs

hwkr:
Just adds some labels to the code inputs.

**Checklist**


antlio (approved):
yess

PR #5639, opened 2025-05-13
fix(components): set search result item to use an anchor tag

### LINKED ISSUE #5589: Search result links are <li> with href which is not working on Brave (in SSR context I believe)
### What happens?

Image
Searches are returning `<li>` elements with href directly on them. While it seems to work and be supported on Chrome, it isn't on Brave leading to clicking an element just closing the search modal without doing anything.

Also I believe that it could be SSR related but I have no strong clue because it's not working on a Nuxt-based server but this playground seems to work https://codepen.io/scalarorg/pen/VwOXqam

Link to our integration (under Nuxt) https://github.com/kestra-io/docs/pull/2526.
I'm aware of the Nuxt native integration but as don't want to have a raw openapi page I found it easier to use the api-reference Vue module directly

Have a `<a>` nested with a proper `href` so that it's supported everywhere and especially in a SSR context?

### OpenAPI Document

_No response_

hwkr:
Semantically it makes more sense for this to be an `<a>` tag and we're managing the aria semantics already with the `role="option"`.

Fixes #5589

**Checklist**

### DIFF
--- a/packages/components/src/components/ScalarSearchResults/ScalarSearchResultItem.vue
+++ b/packages/components/src/components/ScalarSearchResults/ScalarSearchResultItem.vue
@@ -12,9 +12,10 @@ defineOptions({ inheritAttrs: false })
 const { cx } = useBindCx()
 </script>
 <template>
-  <li
+  <a
     :aria-selected="selected"
     role="option"
+    tabindex="-1"
     v-bind="
       cx(
         'group flex cursor-pointer gap-2.5 rounded px-3 py-1.5 no-underline hover:bg-b-2',
@@ -52,5 +53,5 @@ const { cx } = useBindCx()
         <slot name="description" />
       </div>
     </div>
-  </li>
+  </a>
 </template>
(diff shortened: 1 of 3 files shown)


packages/components/src/components/ScalarSearchResults/ScalarSearchResultItem.vue
```diff
@@ -12,9 +12,10 @@ defineOptions({ inheritAttrs: false })
 const { cx } = useBindCx()
 </script>
 <template>
-  <li
+  <a
```
antlio, line 15: would set `a` within `li` vs replace it achieve the same? so it keeps the ul > li markup relationship

marclave (approved):

packages/components/src/components/ScalarSearchResults/ScalarSearchResultItem.vue
hwkr: Updated!

antlio: yes make sense! was just a nit pick but not against ul > div then!

antlio (approved):
let's go

brian-mulier-p:
Thanks a lot for tackling this :+1:

PR #5644, opened 2025-05-14
perf(nuxt): replace $fetch with useFetch in component script setup

ahmedrangel:
**Problem**

Currently, the component uses `$fetch`, using this in the script setup causes it to fetch twice (server and client), using `useFetch` solve this problem by ensuring that if something is fetched on server, the data is forwarded to the client in the payload.

**Solution**

With this PR we benefit from SSR with `useFetch`

References:
https://nuxt.com/docs/api/composables/use-fetch
https://nuxt.com/docs/getting-started/data-fetching#the-need-for-usefetch-and-useasyncdata

**Checklist**

### DIFF
--- a/.changeset/thin-flowers-fold.md
+++ b/.changeset/thin-flowers-fold.md
@@ -1,0 +1,5 @@
+---
+'@scalar/nuxt': patch
+---
+
+perf: replace $fetch with useFetch in component setup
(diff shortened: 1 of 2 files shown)


hanspagel (approved):
OMG yes

PR #5646, opened 2025-05-14
refactor(api-reference): faster sidebar

hanspagel:
⚠️ Requires #5660, so we can feed it something else (`dereferencedDocument`) than the `parsedSpec`.

Say hello to a (partially) revamped sidebar:

* Moved all related files to `@/feature/Sidebar`
* Get rid of the deprecated `parsedSpec` data structure
* Added benchmarks to make it faster (currently 600×, but that’ll vary a bit until the PR is done)
* More tests!

What this PR isn’t changing:

* The API to consume it is mostly the same (`const { items } = useSidebar({ content: { … } })`) to make it easier to integrate
* I tried to not touch the components in this PR

**Tasks**

- [ ] decouple from useNavState (we don't want this inside useSidebar)
- [ ] add supporting for expanding/collapsing sidebar items
* [ ] `pnpm test features/Sidebar` should pass
* [ ] `pnpm vitest bench create-sidebar` should show the new sidebar is faster
- [ ] remove old sidebar code

**Checklist**

### DIFF
--- a/packages/api-reference/src/features/Sidebar/hooks/useSidebar.ts
+++ b/packages/api-reference/src/features/Sidebar/hooks/useSidebar.ts
@@ -1,0 +1,45 @@
+import type { SortOptions } from '@/features/Sidebar/types'
+import type { OpenAPIV3_1 } from '@scalar/openapi-types'
+import { type InjectionKey, inject, provide } from 'vue'
+import { createSidebar } from '../helpers/create-sidebar'
+
+/**
+ * Injection key for the sidebar instance.
+ * This ensures type safety when injecting the sidebar.
+ */
+export const SIDEBAR_SYMBOL: InjectionKey<ReturnType<typeof createSidebar>> = Symbol('sidebar')
+
+/**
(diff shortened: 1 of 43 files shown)


amritk:
lets goooooo

packages/api-reference/src/features/Sidebar/hooks/useSidebar.ts
```diff
@@ -0,0 +1,45 @@
...
+ *
+ * When called with a collection, it creates and provides a new sidebar instance.
+ * When called without parameters, it returns the injected sidebar instance.
+ */
+export function useSidebar(
+  content?: OpenAPIV3_1.Document,
+  sortOptions?: SortOptions,
+): ReturnType<typeof createSidebar> {
+  // If collection is provided, create and provide a new sidebar instance
+  if (content) {
```
amritk, line 23: wow now this is an interesting way to do it

hanspagel: interesting in a good way?

PR #5652, opened 2025-05-14
fix(api-reference): copy path without hash when pathRouting is enabled

### LINKED ISSUE #5638: Anchor copy while `pathRouting` should copy full path and not href + hash
### What happens?

When you click an anchor to copy to the clipboard while pathRouting is configured, it's copying the window.location.href + hash id, resulting in this behavior:

Let's suppose that my current path is:
`http://localhost:5173/tag/users/GET/api/user/%7Bid%7D`

Copied result after clicking anchor for this section:
`http://localhost:5173/tag/users/GET/api/user/%7Bid%7D#tag/users/GET/api/user/{id}`

This problem gets a bit annoying if your screen size allows you to click on a different anchor than the current path resulting in something like this.
`http://localhost:5173/tag/users/GET/api/user/%7Bid%7D#tag/data/GET/api/data/{slug}`

What I would expect while pathRouting is that the anchor should copy the path of the section and not the hash.

Maybe a way to solve this problem is to modify the [getHashedUrl](https://github.com/scalar/scalar/blob/cd0722d74e9d1a1d0fe0ed72a797b94f7e7dd419/packages/api-reference/src/hooks/useNavState.ts#L91-L96) method like this.

```ts
const getHashedUrl = (replacementHash: string, url = window.location.href, search = window.location.search, origin = window.location.origin) => {
    if (config.value.pathRouting) {
      return combineUrlAndPath(origin + config.value.pathRouting.basePath, replacementHash)
    }
    const newUrl = new URL(url)
    newUrl.hash = hashPrefix.value + replacementHash
    newUrl.search = search
    return newUrl.toString()
  }
```

### OpenAPI Document

_No response_

amritk:
**Problem**

Currently, we combine the path and hash when we hit the copy button

**Solution**

With this PR we check for path routing then correctly copy the path or hash

closes #5638

**Checklist**

### DIFF
--- a/.changeset/real-buckets-destroy.md
+++ b/.changeset/real-buckets-destroy.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
+fix: copy correct path when pathRouting enabled
--- a/packages/api-reference/src/components/Anchor/Anchor.vue
+++ b/packages/api-reference/src/components/Anchor/Anchor.vue
@@ -6,7 +6,7 @@ import { useNavState } from '@/hooks/useNavState'
 
 import ScreenReader from '../ScreenReader.vue'
 
-defineProps<{
+const { id } = defineProps<{
   id: string
 }>()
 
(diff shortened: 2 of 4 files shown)


hanspagel (approved):
oops, thanks

PR #5653, opened 2025-05-14
fix(api-client): array query params only showing last value

### LINKED ISSUE #5642: Array parameters not pulling through into example snippet in some languages in API Client
### What happens?

My API endpoint has an array parameter, where multiple IDs can be submitted. If I open up the API client and set a few of these, the example code snippets at the bottom seem quite variable as to whether they include all the parameters or not!

I'm testing on our developer site, here https://developer.churchsuite.com/bookings#tag/resources/GET/bookings/resources/{id}.

If I set up the following parameters:

Image

I expect to see examples like this:

PHP cURL (working as expected)
`$ch = curl_init("https://api.churchsuite.com/v2/bookings/booked_resources?page=2&booking_ids%5B%5D=1&booking_ids%5B%5D=4");`

But for a number of languages, it's very variable:

JS fetch (missing one of the booking_ids)
`fetch('https://api.churchsuite.com/v2/bookings/booked_resources?page=2&booking_ids%5B%5D=4')`

JS Axios (no parameters at all)
```
import axios from 'axios';

const options = {method: 'GET', url: 'https://api.churchsuite.com/v2/bookings/booked_resources'};

try {
  const { data } = await axios.request(options);
  console.log(data);
} catch (error) {
  console.error(error);
}
```

### OpenAPI Document

_No response_

amritk:
**Problem**

Currently, we only show the last query param when there's multiple

**Solution**

This was due to using an object which can only have one key, with this PR we convert it to a query string which can support multiple.

closes #5642

Unfortunately I couldn't fix it for axios as it still uses httpSnippetLite. When we convert that we can fix axios as well.

I also fixed an annoying scrollbar issue 😄 

**Checklist**


packages/api-client/src/components/CodeInput/CodeInput.vue
```diff
@@ -342,7 +342,7 @@ export default {
   </div>
   <div
     v-if="$slots.icon"
-    class="centered-y group-has-[.cm-focused]:z-1 -right-0.25 absolute flex h-full items-center p-1.5">
+    class="centered-y group-has-[.cm-focused]:z-1 absolute right-0 flex h-full items-center p-1.5">
```
amritk, line 345: this was causing an overflow due to the negative position 
cc: @antlio

antlio: good catch

antlio: and not needed as no scroll no more

hanspagel (changes requested):
Tests are failing, with:

```diff
- ofetch('https://example.com/', {
-   query: {
-     foo: 'bar',
-     bar: 'foo'
-   }
- })
+ ofetch('https://example.com?foo=bar&bar=foo')
```

I’d prefer the first syntax for the code example (`query: {…}`), so I’m with the test.

amritk:
@hanspagel that first way doesn't support multiple query params with the same name. Like we cannot do 
```ts
  query: {
-     foo: 'bar',
-     foo: 'baz'
-   }
```
open to other ways but string was the easiest.

amritk:
Alright @hanspagel I dove deep in the code and it looks like we can pass an array

https://github.com/unjs/ufo/blob/main/src/query.ts#L17

hanspagel (approved):
hello array query params

PR #5654, opened 2025-05-14
feat(scalar-app): send cross-origin cookies

hanspagel:
**Problem**

The Scalar App doesn't send cookies without the proxy. The cookies are omitted in cross-origin requests for security reasons, that’s a browser limitation.

**Solution**

We can safely work around this and use the same approach that we use for requests that go through the proxy. We’re setting a `X-Scalar-Cookie` header if it’s going through the proxy or the request is sent in Electron. And right before the request goes out, we’re renaming the header to `Cookie`.

**Checklist**


projects/scalar-app/src/main/index.ts
```diff
@@ -84,14 +84,19 @@ function createWindow(): void {
     return { action: 'deny' }
   })
 
-  // Disable CORS
+  // Allow cross-origin cookies
   mainWindow.webContents.session.webRequest.onBeforeSendHeaders((details, callback) => {
     const { requestHeaders } = details
 
-    upsertKeyValue(requestHeaders, 'Access-Control-Allow-Origin', ['*'])
```
hanspagel, line 91: I don’t know how this ended up in the *REQUEST* headers, that’s a *RESPONSE* header.

packages/api-client/src/libs/send-request/create-request-operation.ts
```diff
@@ -133,8 +133,16 @@ export const createRequestOperation = ({
     const cookieHeader = replaceTemplateVariables(getCookieHeader(cookieParams, headers['Cookie']), env)
 
     if (cookieHeader) {
+      /**
+       * If we are running in Electron, we need to add a custom header
+       * that’s then forwarded as a `Cookie` header.
+       */
+      const isElectron = typeof window !== 'undefined' && 'electron' in window
```
amritk, line 140: this could move to utils, I can see is using it elsewhere

projects/scalar-app/src/main/index.ts
```diff
@@ -84,14 +84,19 @@ function createWindow(): void {
...
+    if (requestHeaders['X-Scalar-Cookie']) {
+      requestHeaders['Cookie'] = requestHeaders['X-Scalar-Cookie']
+      delete requestHeaders['X-Scalar-Cookie']
+    }
+
     callback({ requestHeaders })
   })
 
+  // Disable CORS
```
amritk, line 99: whats this comment for?

amritk (approved):
Sweet!

marclave (approved):

projects/scalar-app/src/main/index.ts
hanspagel: it’s stating what we do below, adding the CORS headers on the fly 😅

I should up my comment game

hanspagel: probably better to see when you look at the whole file though

amritk: 100% didn't look at the rest haha. @claude write awesome comments

PR #5657, opened 2025-05-15
feat(openapi-parser): faster upgrade (50ms → 15ms)

hanspagel:
**Problem**

We’re upgrading the OpenAPI document on the fly for every user fetching a Swagger 2.0 or OpenAPI 3.0 document, which still seems to be the biggest chunk. It only takes ~50ms for large documents (on my machine). But what if we'd make this even faster?

**Solution**

This PR improves the performance for the upgrade from OpenAPI 3.0 to OpenAPI 3.1. I wasn’t able to optimize the upgrade from Swagger 2.0 to OpenAPI 3.0, but they use the other upgrader, too. So everyone benefits.

The biggest gain comes from not recursively traversing the whole document multiple times, but only once.

**Results**

On my machine, the upgrade for Stripe is down from ~50ms to ~15ms. 💥

Screenshot 2025-05-15 at 15 15 18

**Checklist**


marclave (approved):
unreal @hanspagel ❤️  🔥 ✨

PR #5658, opened 2025-05-15
chore(openapi-parser): deprecate pipeline syntax

hanspagel:
**Problem**

In the README, we promote a pipeline syntax for the parser:

```ts
import { openapi } from '@scalar/openapi-parser'

const specification = …

// New pipeline …
const result = openapi()
  // loads the specification …
  .load(specification)
  // upgrades to OpenAPI 3.1 …
  .upgrade()
  // removes all internal operations …
  .filter((schema) => !schema?.['x-internal'])
  // done!
  .get()
```

But this adds a lot of complexity to the code base, and it hindered me a few times to change the utilities.

**Solution**

I’d like to get rid of it. People can just use the individual utilities. With this PR we’re marking it as @deprecated and removing it from the README.

**Checklist**

### DIFF
--- a/.changeset/cyan-meals-fly.md
+++ b/.changeset/cyan-meals-fly.md
@@ -1,0 +1,5 @@
+---
+'@scalar/openapi-parser': patch
+---
+
+chore: mark pipeline syntax as deprecated


amritk (approved):
+1 for functional!

PR #5660, opened 2025-05-15
feat(api-reference): single content pipeline

### LINKED ISSUE #5407: api-reference-react@0.6.0 -- can not execute structuredClone regression
### What happens?

api-reference-react@0.6.0 introduced a regression leading to the `can not execute structuredClone` error message when directly passing in JSON/YAML content 

```
import { ApiReferenceReact } from '@scalar/api-reference-react'

...
    <ApiReferenceReact
      configuration={{
        content: content,
...
```

Image

The api-reference-react package is able to support directly passing in JSON/YAML content without error

### OpenAPI Document

_No response_

hanspagel:
⚠️ This changes how we ingest content, it affects *all* users. We should test a lot of user pages (before and after release).

**Problem**

To get and dereference OpenAPI documents, we have a few approaches that we use:

* The configuration (what most people use)
* The old `parsedSpec` we want to get rid of (required for `ApiReferenceLayout`)
* The `rawSpec` (required for `ApiReferenceLayout`)
* The `@scalar/api-client` store (used in a few places)

Some of the hard work is done twice (dereference), or only in the client, but not in the reference (upgrade). The configuration for retrieving documents varies (no support for external references in the client).

Eventually, we want to have one store that does all that, but that won’t happen over night and eventually we’ll still need to ways to do that:

* For SSG/SSR, you can prepare all the data and directly pass it (as a store, soon™)
* For the self-hosted version you’ll need to fetch the data in the browser
* We don’t plan to replace the API client store soon.

**Solution**

This PR introduces a dedicated `DocumentSource` feature, that does all the OpenAPI document work in a central place. This allows us to do the hard work only once (dereference, upgrade) and benefit from it everywhere.

No matter how we’ll proceed, we can pass/create whatever data structure in that component and then use it everywhere.

* Good bye `useReactiveSpec` 👋
* Hello smaller `ApiReferenceLayout` 👋
* We dereference documents once now (did it in `parse` & in the API Client store)
* We only updated the API client store once (we did it in `ApiReferenceLayout.vue` & `ApiClientModal.vue`)
* `parse` works with an dereferened + upgraded document now -> we can remove all Swagger 2.0 and OpenAPI 3.0 types and hacks!

For more details, there’s a README in the feature folder!

Fixes #5407.

**Checklist**


packages/api-reference/src/features/DocumentSource/hooks/useDocumentSource.ts
```diff
@@ -0,0 +1,183 @@
...
+import { createEmptySpecification } from '@/libs/openapi'
+
+import { measure } from '@/helpers/measure'
+import type { Spec } from '@scalar/types/legacy'
+import { useDocumentFetcher } from './useDocumentFetcher'
+
+/**
+ * Pass any data source, retrieve all the data stores and whatever we need to render the API reference.
+ */
+export function useDocumentSource({
```
amritk, line 18: The issue with putting these things in a hook is if you call the hook again it will duplicate everything? Thats why its nice to have these things as high up as possible in a component, you don't need to worry about someone calling it again

packages/api-reference/src/helpers/measure.ts
```diff
@@ -0,0 +1,32 @@
...
+export function measure<T>(name: string, fn: () => T): T
+export function measure<T>(name: string, fn: () => T | Promise<T>): T | Promise<T> {
+  const start = performance.now()
+
+  const result = fn()
+
+  const end = performance.now()
+  const duration = Math.round(end - start)
+
+  console.log(`${name}: ${duration} ms`)
```
amritk, line 29:
```suggestion
  console.info(`${name}: ${duration} ms`)
```
lets use info when its something that we want permanent, logs should just be for debugging

amritk (commented):
So there's a lot of good stuff in this PR, especially the type improvements and the new measure function 👍🏾 

What I'm unsure about is just moving evertything from one hook + component into another hook. Having heavy code like this meant to be called in one place in a hook is always dangerous because it can be called elsewhere then you are doubling the computation. Hooks are composeable. 

The issue were trying to avoid is introducing bugs when we are just going to delete this hook later anyway with the new store.  Or maybe I'm missing the point of it.

packages/api-reference/src/components/ApiReferenceLayout.vue
```diff
@@ -344,32 +342,33 @@ const themeStyleTag = computed(
       '--scalar-y-offset': `var(--scalar-custom-header-height, ${yPosition}px)`,
     }"
     @scroll.passive="debouncedScroll">
+    <div>deref:{{ !!dereferencedDocument }}</div>
     <!-- Header -->
     <div class="references-header">
       <slot
-        v-bind="referenceSlotProps"
+        v-bind="{ ...referenceSlotProps }"
```
amritk, line 349:
```suggestion
        v-bind="referenceSlotProps" 
```
was it not the same before?

packages/api-reference/src/components/Content/Lazy/Loading.vue
```diff
@@ -218,7 +218,7 @@ onMounted(() => {
               <Anchor :id="getModelId({ name })">
                 <SectionHeaderTag :level="2">
                   {{
-                    (getModels(parsedSpec)?.[name] as OpenAPIV3.SchemaObject)
+                    (getModels(parsedSpec)?.[name] as OpenAPIV3_1.SchemaObject)
```
amritk, line 221: one step up would be type it in getModels, but also we will remove this component anyway

amritk (changes requested):
Alright it was looking good for the most part BUT I did find one issue on the 
http://localhost:5050/api-reference page in the web example.

Operations and what not seem to be missing.
Main:
image

This PR:
image

Also lets debounce the change detection

image

packages/api-reference/src/components/ApiReferenceLayout.vue
hanspagel: oh yeah, that’s a left over

amritk (approved):
Could test a few more specs but I dare say its ready

packages/api-reference/src/components/Content/Introduction/Introduction.test.ts
```diff
@@ -100,7 +96,7 @@ describe('Introduction', () => {
     })
 
     const downloadLink = wrapper.findComponent(DownloadLink)
-    expect(downloadLink.props('filename')).toBe('hello-world-api')
+    expect(downloadLink.props('title')).toBe('Hello World API!')
```
amritk, line 99: @hanspagel why was this changed?

amritk: rolling it back

marclave (approved):

amritk:
OMG here we go

PR #5662, opened 2025-05-15
feat(nuxt): SSR-friendly color mode preference implementation

ahmedrangel:
**Problem**

Currently, the nuxt module does not save the user's color mode preference when using the toggle switch, it is tied to starting with the one defined in the config.

**Solution**

With this PR we'll add support for saving the user's color mode preference and handle their states.

I look forward to your feedback.

**Checklist**

### DIFF
--- a/.changeset/lazy-roses-wave.md
+++ b/.changeset/lazy-roses-wave.md
@@ -1,0 +1,5 @@
+---
+'@scalar/nuxt': patch
+---
+
+feat(nuxt): ssr-friendly color mode preference implementation
--- a/integrations/nuxt/package.json
+++ b/integrations/nuxt/package.json
@@ -53,7 +53,8 @@
     "@nuxt/kit": "^3.12.3",
     "@scalar/api-client": "workspace:*",
     "@scalar/api-reference": "workspace:*",
-    "@scalar/types": "workspace:*"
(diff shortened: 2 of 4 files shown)


marclave (commented):
thanks for this PR @ahmedrangel ✨ 

going to add @amritk to take a look at this as they did a ton of the nuxt work <3

amritk:
So this does work as long as you don't mind the flash of incorrect color in the beginning as its coming from local storage. Since we are on nuxt, ideally we could set a cookie then read it on the server to get the correct color from the beginning. 

If you don't mind the flash we can merge this then add the cookie solution when we do SSR.

hanspagel (dismissed):
sick PR

I’d tend to avoid the color flash (flashing colors aren’t great for some people), but not a super strong opinion here.

amritk:
I actually did not test it, but I know that we cannot use localStorage for SSR. That being said I don't think the SSR is quite working anymore anyway so maybe its fine. I'll test it and see 👍🏾

amritk (approved):
Yep just tested and there is no flash because its not doing SSR. We'll take it for now thanks for the PR! I'll do a proper cookie fix when we fix SSR

ahmedrangel:
Good!

PR #5665, opened 2025-05-16
feat(components): scalar tooltip overhaul

hwkr:
Overhauls the `ScalarTooltip` component with a few notable changes:

* We now only render a single tooltip and swap out the content (should be more performant)
* There's a hook you can use in addition to the Vue component to create a tooltip
* Tooltips can now only contain plain text, if it's so complicated it needs rich text it probably shouldn't be in a tooltip
* Tooltips are now dark in light mode and vice versa for better contrast

A quick note from MDN on tooltip usage:

> Instead of using tooltips and hiding important information, consider writing clear, succinct, always visible descriptions. If you have space, don't use tooltips or toggletips. Just provide clear labels and sufficient body text.

The new tooltips are compliant with the [MDN recommendations](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/tooltip_role#required_javascript_features):

- [x] The tooltip is displayed and disappears via keyboard focus and removal of focus and by the mouse events - mousing over and mousing out.
- [x] The tooltip never receives focus. Focus stays on the owning element.
- [x] The tooltip can be hidden with the Escape key.
- [x] The tooltip stays open when hovered.

Changes to usage of the tooltip:

* Tweaked the wording for the watch mode tooltip
* Swapped the tooltip for a popover for the request table because the info is pretty complicated and probably shouldn't be in a tooltip

Still to do for Brynn:

- [x] Extract some of the types / constants that are in the tooltip hook file right now
- [x] Add some tests for the hook
- [x] Any color / style tweaks cc @antlio @cameronrohani 
- [x] Possibly something else I'm forgetting

**Checklist**


hanspagel (commented):
🤯

Just something I noticed, without the border they seem to not elevated (visually):

Before

Screenshot 2025-05-16 at 11 01 43

After

Screenshot 2025-05-16 at 11 01 16

And one note: I think it’s nice to be able to use inline styles, specifically `<strong>` and `<code>` (e.g. for keyboard shortcuts). But I didn’t check whether we actually do that and I don’t if this costs performance or something.

antlio:
great work! i feel like bold in such small font is harder to read that smaller weight. a first pass on the overall UI

**request tip**
| before | after |
| -------|------|
| image | image |
| previous tooltip | style broke here for the header infos | 


**warning tip**
| before | after |
| -------|------|
| image | image |
| warning tooltip | warning is not displaying infos anymore | 

**workspace dropdown**
| before | after |
| -------|------|
| image | image |
| delete button aligned | the delete button icon/text are not aligned anymore |

packages/components/src/components/ScalarTooltip/useTooltip.ts
```diff
@@ -0,0 +1,229 @@
...
+}
+
+/**
+ * Handle the escape key
+ *
+ * If the escape key is pressed we need to hide the tooltip
+ *
+ * @see https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/tooltip_role#keyboard_interactions
+ */
+function handleEscape(e: KeyboardEvent) {
```
amritk, line 183: You might also want to stop propogation or it might close the modal for example

packages/components/src/components/ScalarTooltip/useTooltip.ts
```diff
@@ -0,0 +1,229 @@
...
+/**
+ * Use the tooltip
+ *
+ * If the tooltip is not initialized it will be initialized
+ */
+export function useTooltip(opts: TooltipConfiguration) {
+  initializeTooltipElement()
+
+  const showThisTooltipAfterDelay = (e: Event) => showTooltipAfterDelay(e, opts)
+  const showThisTooltip = (e: Event) => showTooltip(e, opts)
```
amritk, line 206: Not sure what the point of this bit is, you could just move those functions into here, or just use the global ones directly in the listeners below

packages/components/src/components/ScalarTooltip/useTooltip.ts
```diff
@@ -0,0 +1,229 @@
...
+
+  watch(
```
amritk, line 208: We should also see what the cost of this watcher is when its used in a tonne of components I know watchers used to be heavy (not sure what the current state is in latest vue version). 

We should do some benchmarking here by testing the sidebar on a big spec with and without this watcher

amritk: would it be cheaper to just do a @hover on the element which needs the tooltip then show it? questions to ponder

amritk (commented):
I only reviewed the code for the new tooltip, leaving the ux stuff to the other reviews.

I just got some questions about perf as this was done to improve perf, can we run some benchmarks to see what sort of improvements we get? Mostly worried about adding a bunch of watchers but seems like a solid learning experience

packages/api-client/src/components/TopNav/TopNavItem.vue
```diff
@@ -34,46 +35,34 @@ defineEmits<{
   <ScalarContextMenu triggerClass="overflow-hidden w-full">
     <template #trigger>
       <ScalarTooltip
-        class="scalar-client"
-        :delay="500"
-        :disableClosingTrigger="true"
-        resize
-        :sideOffset="4">
-        <template #trigger>
+        :content="`${isMacOS() ? '⌘' : '^'} ${hotkey}`"
```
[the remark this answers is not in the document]
antlio: we can export the modifier key symbol from hot key component maybe

packages/components/src/components/ScalarTooltip/useTooltip.ts
hwkr: We need a reference to a specific function that runs with that config but you're right we could move it into the hook that probably makes more sense

packages/components/src/components/ScalarTooltip/useTooltip.ts
amritk: @hanspagel is the master benchmarker.

But we can just do a visual test of where the tooltip issue was found in the first place

packages/api-client/src/components/TopNav/TopNavItem.vue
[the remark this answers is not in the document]
hwkr: Yeah I was thinking it should be in a general tips package but we don't have one of those? I'm not sure where it should like but probably not in use-tooltip (which we should probably get rid of anyway)

packages/components/src/components/ScalarTooltip/useTooltip.ts
hwkr: Sure, probably want to release for that though since it's in the other repo (and I changed a bunch of packages here)

hwkr:
@antlio 

**request tip**

The fonts seem to be working for me? It has a bit more padding (which I can remove) and I adjusted the alignment to be centered.

Google Chrome-2025-05-16-12-42-17@2x
Google Chrome-2025-05-16-12-42-03@2x

**warning tip**

Seems to be working? It's worth note you have to click it now since it's a toggletip not a tooltip.

Google Chrome-2025-05-16-12-43-24@2x

**workspace dropdown**

The delete button seems to be broken on prod but I can fix it :)

Google Chrome-2025-05-16-12-49-38@2x

packages/components/src/components/ScalarTooltip/useTooltip.ts
amritk: kk thats fine, iterative enhancements

amritk (approved):
You can test with pnpm link as well

antlio:
@hwkr awesome! for the workspace dropdown, the delete text/icon seem still off compare to the rename one

hwkr:
> @hwkr awesome! for the workspace dropdown, the delete text/icon seem still off compare to the rename one

Yeah it's definitely off it just seems to be a pre-exisiting issue, just fixing it :)

antlio:
can we add a hover hint or something for warning/error? it doesn't seem obvious that it is clickable, especially in presence of enums, the row has already a pointer cursor

packages/components/src/components/ScalarTooltip/useTooltip.ts
hwkr: Moved into the hook

hwkr:
Okay fixes / suggestions implemented 🙏

hwkr:
Added tests 🫡

antlio (approved):

cameronrohani (approved):

marclave (approved):
omg lets get this in once @cameronrohani approves

marclave:
oops ill wait for @hanspagel now 😛

hanspagel (approved):
omg, let’s git this in _**now**_

PR #5672, opened 2025-05-17
feat(snippetz): add multipart/form-data support for several converters

nsychev:
**Problem**

Currently, most of the converters in `snippetz` ignore `multipart/form-data` body or try to encode it in JSON, which is the wrong way to handle such requests. 

**Solution**

With this PR, I supported `multipart/form-data` and `application/x-www-form-urlencoded` body types for some converters, and fixed the relevant bugs:

* C (libcurl): add support, fix skipped tests
* Go (net/http): add support, fix tests
* Java, Kotlin (okhttp): add support, add tests
* JS, Node.JS (fetch): add support, add tests
  * These two converters were fully equal — I don't know if it was intended. But I decided to generate different code examples: in JS I just put `Blob` into file upload fields, and in Node.js I read file from the filesystem.
* Python (requests): add support for multiple fields with the same name
* shell (okhttp): add support, fix skipped tests

**Checklist**

### DIFF
--- a/packages/snippetz/src/httpsnippet-lite/esm/targets/go/native/client.ts
+++ b/packages/snippetz/src/httpsnippet-lite/esm/targets/go/native/client.ts
@@ -19,7 +19,7 @@ export const native = {
     description: 'Golang HTTP client request',
   },
   convert: ({ postData, method, allHeaders, fullUrl }, options = {}) => {
-    const { blank, push, join } = new CodeBuilder({ indent: '\t' })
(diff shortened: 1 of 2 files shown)


marclave:
wow awesome work @nsychev ! im going to lean on @hanspagel for this since he made the package ✨ 

but great PR ✨

hanspagel (approved):
PR of the week, great work 🫶

Issue #5314, opened 2025-04-03
Group tagged model(s) together with APIs


amritk:
### Discussed in https://github.com/scalar/scalar/discussions/5290

<div type='discussions-op-text'>

<sup>Originally posted by **shimonenator** April  1, 2025</sup>
Currently, if you tag a few APIs, they are neatly nested under the collapsible heading in the side nav. However, models, tagged or not, are always displayed under the Models heading at the bottom of the side nav. 

It would be great if there was an option to enable grouping models together with APIs, as it would make it much easier to understand the domain object and its relationship with APIs. 

Stripe has a good example of such approach:

![Screenshot 2025-04-01 192717](https://github.com/user-attachments/assets/eb0233cc-421d-4523-98aa-c6a9c2dbd54d)
</div>

shimonenator:
@amritk Thank you! Do you have an example of how to use this feature?

PR #5674, opened 2025-05-18
feat(api-reference): group tagged models together with APIs

men232:
**Problem**

Currently, if you tag a few APIs, they are neatly nested under the collapsible heading in the side nav. However, models, tagged or not, are always displayed under the Models heading at the bottom of the side nav.

Related issue: #5314

**Solution**

This PR introduces support for grouping models alongside APIs in the side navigation by using `x-tags` on component schemas. With this change, models can be displayed above the same collapsible tag sections as the related APIs.

**Checklist**

### DIFF
--- a/packages/api-reference/src/components/Content/Tag/TagList.vue
+++ b/packages/api-reference/src/components/Content/Tag/TagList.vue
@@ -6,6 +6,7 @@ import type { Spec, Tag as TagType } from '@scalar/types/legacy'
 import { computed } from 'vue'
 
 import { Lazy } from '@/components/Content/Lazy'
+import { Models } from '@/features/Models'
 import { Operation } from '@/features/Operation'
 import { useNavState } from '@/hooks/useNavState'
 import { useSidebar } from '@/hooks/useSidebar'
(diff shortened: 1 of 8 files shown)


hanspagel:
Thanks for the PR! While the feature is great, we are about to drop the parse.ts (as stated at the top of the file), so this won’t be there for long. :(

men232:
> Thanks for the PR! While the feature is great, we are about to drop the parse.ts (as stated at the top of the file), so this won’t be there for long. :(

Yeah, I saw the note about deprecating parse.ts, but since I didn’t see the new schema processing in action yet, I figured I’d just build on the current setup for now 🙂

This feature was super important for our project, so I thought it might be helpful to share our solution anyway. Thanks!

amritk:
@men232 I'll add it as I needed to get a feature in before the parse function gets dropped 😄

packages/api-reference/src/components/Content/Tag/TagList.vue
```diff
@@ -59,6 +60,11 @@ const isLazy = (index: number) =>
       :collection="collection"
       :spec="spec"
       :tag="tag">
+      <Models
```
amritk, line 63: wont' be lazy here but we can worry about that later

amritk:
Alright we have added this to our new sidebar so it should be done soon!

PR #5676, opened 2025-05-19
Add schema titles in ui

DmitriiKhudiakov:
**Problem**

Currently,

1. In **SchemaDiscriminator.vue**, `getModelNameFromSchema` currently prioritizes `schema.name` over `schema.title`, but we want the human-readable `title` to come first.
2. In **SchemaPropertyHeading.vue**, `getModelNameFromSchema` doesn’t even consider `schema.title`, causing UI labels to fall back directly to the raw JSON type.
3. For non-array schemas, the UI renders `object` (the JSON type) instead of falling back to `title → name → type`.

**Solution**

With this PR:

1. **SchemaDiscriminator.vue**

   * Reorder `getModelNameFromSchema` to prefer `schema.title` first, then `schema.name`, then deep-match component keys.
2. **SchemaPropertyHeading.vue**

   * Extend its `getModelNameFromSchema` helper to include a `title` fallback.
3. **Type display expression**

   * Replace the inline logic with:

     ```vue
     {{ Array.isArray(value.type) ? value.type.join(' | ') : (value?.title ?? value.name ?? value.type) }}
     ```
   * This ensures:

     1. Union types render as `A | B`
     2. Objects show `schema.title` (if present), then `schema.name`, then the raw type.

**Checklist**


hanspagel:
Amazing PR! I’ll hand the review to @antlio. 🤙

packages/api-reference/src/components/Content/Schema/SchemaDiscriminator.vue
```diff
@@ -64,14 +64,14 @@ const getModelNameFromSchema = (schema: any): string | null => {
     return null
   }
 
-  if (schema.name) {
-    return schema.name
-  }
-
   if (schema.title) {
```
antlio, line 67: good call! prob got change within some previous changes by mistake, def should prioritize human-readable first!

packages/api-reference/src/components/Content/Schema/SchemaPropertyHeading.vue
```diff
@@ -124,7 +128,11 @@ const constValue = computed(() => {
           {{ getModelNameFromSchema(value.items) || value.items.type }}[]
         </template>
         <template v-else>
-          {{ Array.isArray(value.type) ? value.type.join(' | ') : value.type }}
+          {{
+            Array.isArray(value.type)
+              ? value.type.join(' | ')
+              : (value?.title ?? value.name ?? value.type)
```
antlio, line 134: might want to extract this at some point to make it easier to understand?

packages/api-reference/src/components/Content/Schema/SchemaDiscriminator.vue
antlio: can you update also the test in `SchemaDiscriminator.test.ts` to swap precedence?

DmitriiKhudiakov:
In OpenAPI you can include an `xml.name` override to control XML element names for this schema—even if your JSON model is called “Tag”, the XML serializer will wrap it in `<Tag>…</Tag>`. Because most of our users consume JSON-only APIs, this is a valid but very uncommon case—so our UI doesn’t currently display the XML-specific name.

image

hanspagel (approved):
niiice, let’s get it in

Issue #4232, opened 2024-12-16
Add support for arrays in query parameters


MarvinKlein1508:
### What happens?

Parameters are being displayed as type array but you cannot add more than one value. See #2835 

### What did you expect to happen?

I expect Scalar to allow the user to add n amount of values.

### OpenAPI Document

https://sandbox.scalar.com/e/wyd2k

DenuxPlays:
Hey,
I am also interested in this feature.
Are they any updates on this?

rektdeckard:
Hey all! I see some work was put into this in #4512 but I don't think this is complete until the client respects the `explode: false` option, which [per the spec](https://swagger.io/docs/specification/v3_0/serialization/#query-parameters) should serialize as comma-separated values. Currently, array parameters are always serialized as exploded (i.e. `type=foo&type=bar&type=baz`), though it should be CSV when the option is `false`, (e.g. `type=foo,bar,baz`).

OpenAPI doc:

```json
{
  "name":"category",
  "in":"query",
  "description":"Filter search results by one or more comma-separated icon categories.",
  "required":false,
  "schema":{
    "type":"array",
    "items":{
      "$ref":"#/components/schemas/Category"
    }
  },
  "style":"form",
  "explode":false
}
```

Generated URL (see devtools panel):

<img width="1721" alt="Image" src="https://github.com/user-attachments/assets/1055c539-1848-42c5-9820-c7a33f4c8aa8" />

PR #5678, opened 2025-05-20
fix(api-client): support unexploded array query parameters

rektdeckard:
**Problem**

Per #4232, the API client only has rudimentary support for array query params and can only correctly handle `style: 'form'` and `explode: true`. Unexploded array params are quite common, and though it would eventually be nice to see [all styles](https://spec.openapis.org/oas/v3.1.1.html#style-values) supported in paths, headers, queries, and cookies, this seems like a good start.

**Solution**

This patch updates `createFetchQueryParams` to take in an additional optional`Request` object, and uses that to locate the schema corresponding to each param to be serialized. If the schema has the `explode` property set to `false` (the default is `true`) then the parameters are serialized as CSV. Otherwise, the existing behavior of serializing the same key multiple times is used.

There is some awkward type gymnastics due to some duplication of zod schemas, and [this unresolved TODO](https://github.com/scalar/scalar/blob/main/packages/oas-utils/src/entities/spec/request-examples.ts#L18-L24) about including the request schema in the request itself. If there's a preferred approach to my workaround, I'm happy to implement it.

**Checklist**

### DIFF
--- a/.changeset/forty-toes-build.md
+++ b/.changeset/forty-toes-build.md
@@ -1,0 +1,6 @@
+---
+'@scalar/api-client': patch
+'@scalar/oas-utils': patch
+---
+
+support unexploded serialization of array query params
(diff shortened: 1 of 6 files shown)


hanspagel:
YESS

The tests are failing, can you fix them up?

rektdeckard:
@hanspagel done!

amritk (approved):
Looks good to me! Awesome PR!

PR #5681, opened 2025-05-20
feat(hono): support for dynamic configuration

inyourtime:
This PR introduces a flexible middleware for `hono` that accepts dynamic configuration as either a synchronous or asynchronous function with hono context, which is useful in runtimes like [Cloudflare Workers](https://hono.dev/docs/getting-started/cloudflare-workers#bindings) where environment values can be accessed via `c.env`

**Checklist**


hanspagel (approved):
Babe, wake up, @inyourtime did it again.

PR #5688, opened 2025-05-20
feat(api-reference): extracts schema composition components

antlio:
**Solution**

in parallel of the current work around discriminator, this pr separates composition from discriminator in their own files to prep the incoming updates in order to make things clearer.

**Checklist**


hwkr (approved):
Cool cool

amritk (approved):
Looks mostly like renames 🤷🏾

PR #5689, opened 2025-05-20
fix(api-reference): layout slot propogation

amritk:
**Problem**

Currently, the sidebar-start slot would override the document-selector. Now we can use both.

**Solution**

With this PR
- removed an old helper file from components
- fixed the sidebar-start slot so that its usable along with document-selector

**Checklist**


packages/api-reference/src/components/Layouts/Layouts.vue
```diff
@@ -41,14 +32,11 @@ const layouts = {
     @updateContent="$emit('updateContent', $event)">
     <!-- Expose all layout slots upwards -->
     <template
-      v-for="(_, name) in referenceLayoutSlots"
+      v-for="(_, name) in slots"
```
amritk, line 35: @hanspagel this does the same thing?

hwkr: Should do

hwkr (approved):
Lgtm 🙏

PR #5693, opened 2025-05-21
fix(api-client): server variable form border

antlio:
**Problem**

currently on reference border are missing from the server variable form.

**Solution**

this pr sets the missing style for server variable form when using reference layout.

| before | after |
| -------|------|
| image | image |
| form border are missing | form border gets displayed | 

**Checklist**

### DIFF
--- a/packages/api-client/src/components/Server/ServerVariablesForm.vue
+++ b/packages/api-client/src/components/Server/ServerVariablesForm.vue
@@ -6,11 +6,17 @@ import type {
   ServerVariableValues,
 } from '@/components/Server/types'
 
-const props = defineProps<{
+const {
+  values,
+  variables,
+  layout = 'client',
+} = defineProps<{
   variables?: ServerVariables | undefined
   values?: ServerVariableValues
   /** The ID of the input controlled by the variables form */
   controls?: string
+  /** The layout of the server variables form */
+  layout?: 'client' | 'reference'
 }>()
 
 const emit = defineEmits<{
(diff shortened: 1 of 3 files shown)


packages/api-client/src/components/Server/ServerVariablesForm.vue
```diff
@@ -22,19 +28,17 @@ function setVariable(name: string, value: string) {
 }
 
 const getVariable = (name: string) => {
-  return (
-    props.values?.[name] ??
-    props.variables?.[name]?.default ??
-    ''
-  ).toString()
+  return (values?.[name] ?? variables?.[name]?.default ?? '').toString()
```
amritk, line 31: This should be the other way around? obecause it theres a default then this will be 
`{ default: 'xxxx' }`.toString()

how about
```suggestion
  return (values?.[name]?.default ?? variables?.[name] ?? 'variable').toString()
```
threw an optional default name in there, though it should never get hit

amritk (changes requested):
Unless I'm misreading it again, what you think about that?

packages/api-client/src/components/Server/ServerVariablesForm.vue
antlio: that's how it's been i just destructured prop here, but i believe you are right!

amritk: yup it was like that before, but wanna fix it anyway 😄

amritk: we can just commit this suggestion actually

amritk: Whoops  I misread it LOL my eyes! This is why I gotta review in the editor lol

amritk (approved):
I was blind yet again

PR #3405, opened 2024-10-08
chore: add `default` to package.json exports fields

geoffgscott:
Adds fallback exports for require syntax

closes #2994

PR #5694, opened 2025-05-21
chore: add `default` to `api-reference-react` package.json export field

mpminardi:
**Problem**

Currently, attempting to import newer versions of `api-reference-react` into commonJS projects fails with:

`Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: No "exports" main defined in /node_modules/@scalar/api-reference-react/package.json`

Previously this was addressed in https://github.com/scalar/scalar/pull/3405 for other packages by adding a `default` export.

**Solution**

With this PR a default export is added in a similar fashion for `api-reference-react`.

**Checklist**

### DIFF
--- a/.changeset/curvy-onions-thank.md
+++ b/.changeset/curvy-onions-thank.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference-react': patch
+---
+
+Add default export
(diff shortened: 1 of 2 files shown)


amritk (approved):
Sure lets do it!

PR #5699, opened 2025-05-22
chore(themes): migrate to tailwind v4

hwkr:
Upgrades Tailwind to v4 and ~~refactors our Tailwind packages to consume each others Tailwind configs rather than the compiled Tailwind substantially reducing our CSS duplication (and making it so dev tools works again)~~. Update: See [DOC-3440: Dedupe Tailwind CSS](https://linear.app/api-docs/issue/DOC-3440/dedupe-tailwind-css).

This still has some CSS duplication (similar to before), but I cleaned up our configs a fair bit and centralized styles where I could.

Todo:

- [x] ~~Put a copy of the Tailwind vars on `.scalar-app` so they don't get overridden by other peoples Tailwind configs~~ Did a little investigating and it looks like we probably won't actually need to do this. Will keep an eye on it when we release though
- [ ] Add a changeset and mark breaking changes

**Checklist**


amritk:
One of the top priorities is we gotta make sure our tailwind is de-duped: 
we either output two style.css files, with and without tailwind then add the packages to sources of the consuming package
OR
we just de-dupe the css on build since its in vite now, we can probably just make a plugin.
OR 
whatever else!

hwkr:
@amritk yep! That's on the todo list! Need to get all the packages moved to v4 first though

amritk:
hey so now in the references css, do we need to import components or does that come from the client css?

hwkr:
> hey so now in the references css, do we need to import components or does that come from the client css?

We need to import components as well

.vscode/settings.json
```diff
@@ -30,10 +30,7 @@
     "editor.defaultFormatter": "esbenp.prettier-vscode"
   },
   "[css]": {
-    "editor.defaultFormatter": "biomejs.biome",
-    "editor.codeActionsOnSave": {
-      "source.fixAll.biome": "explicit"
-    }
+    "editor.defaultFormatter": "esbenp.prettier-vscode"
```
hanspagel, line 33: wait, don’t we use Biome for CSS now? I think we want to use Biome where possible.

packages/api-reference/src/components/Content/Schema/Schema.vue
```diff
@@ -61,7 +61,7 @@ const handleClick = (e: MouseEvent) =>
       :class="[
         `schema-card--level-${level}`,
         { 'schema-card--compact': compact, 'schema-card--open': open },
-        { 'border-t-1/2': additionalProperties && open },
+        { 'border-t': additionalProperties && open },
```
hanspagel, line 64: you sure we don’t want the 1/2 border anymore? disclaimer: I know nothing about the borders

hanspagel (commented):
amazing work! I’m not the best one to review, but left a few comments that came up while skimming through it

packages/api-client/vite.config.ts
```diff
@@ -25,7 +27,36 @@ export default defineConfig({
...
+      'src/components/ViewLayout/index.ts',
+      'src/components/Sidebar/index.ts',
+      'src/components/Server/index.ts',
+      'src/components/ImportCollection/index.ts',
+      'src/components/HttpMethod/index.ts',
+      'src/components/DataTable/index.ts',
+      'src/components/CommandPalette/index.ts',
+      'src/components/CodeInput/index.ts',
+      'src/components/AddressBar/index.ts',
+    ],
```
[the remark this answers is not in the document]
hwkr: Yeah I need a way to not overwrite the additional export of `tailwind.css` I'll probably extend the build package.

packages/api-reference/src/components/Content/Schema/Schema.vue
hwkr: Our default border size is now `0.5px`. This is legacy from before @cameronrohani changed the default border width.

.vscode/tailwind.json
hwkr, line 1: This file stops vscode from complaining about Tailwind directives in the CSS

packages/api-client/src/index.ts
```diff
@@ -1,3 +1,5 @@
+import './style.css'
```
hwkr, line 1: I moved this import out of the individual layouts because it was importing everything three times.

packages/api-client/src/style.css
```diff
@@ -0,0 +1,20 @@
+/**
+ * API Client Styles
+ */
+
+@import "@scalar/themes/style.css"; /* Theme Base Styles and Reset */
+@import "@scalar/components/style.css"; /* Component Library Styles */
+@import "@scalar/themes/tailwind.css"; /* Tailwind Theme + Config */
+
+.scalar-client {
```
hwkr, line 9: I tried turning this into `.scalar-app` but we end up with some ordering issues. I'm going to do it when I do the deduping PR.

packages/api-client/src/views/Collection/components/MarkdownInput.vue
```diff
@@ -48,19 +48,19 @@ watch(mode, (newMode) => {
...
     <div
-      class="has-[:focus-visible]:bg-b-1 z-1 group relative flex flex-col rounded-lg">
+      class="has-[:focus-visible]:bg-b-1 group relative z-1 flex flex-col rounded-lg">
       <div class="flex h-full min-h-[calc(1rem*4)] flex-col">
         <!-- Preview -->
         <template v-if="mode === 'preview'">
           <template v-if="modelValue && modelValue.trim().length">
             <ScalarMarkdown
               v-if="modelValue"
-              class="hover:border-b-3 h-full flex-1 rounded border border-transparent p-1.5"
```
hwkr, line 58: `border-b-3` can also mean `border-bottom: 3px` and Tailwind v4 doesn't like that

packages/api-reference/src/style.css
```diff
@@ -0,0 +1,16 @@
+/**
+ * API Reference Styles
+ */
+
+@import "@scalar/themes/style.css"; /* Theme Base Styles and Reset */
+@import "@scalar/themes/tailwind.css"; /* Tailwind Theme + Config */
+
+.scalar-api-reference {
```
hwkr, line 8: See comment above about `.scalar-client`

PR #5701, opened 2025-05-22
feat(openapi-parser): bundle openapi documents

DemonHa:
Bundles an OpenAPI specification by resolving all external references.
This function traverses the input object recursively and replaces any external $ref references with their actual content. External references can be URLs or local files.

**Checklist**

- [X] I’ve added tests for the regression or new feature.


hanspagel (changes requested):
OMG, PR of the week already 🤟

super clean code, great solution, awesome tests, works just great! way better than everything I ever wrote.

I allowed myself to add a few things, though:

* we’re using `it()` instead of `test()` in all our tests
* we make shorter test names `it('does something')` vs. `it('should do something')`
* I’ve added an example to the README
* unrelated: I removed an unused dev dependency
* while the whole package still uses `type AnyObject` (my bad), I’d use `type UnknownObject` for new stuff

One edge case, that I think isn’t covered (I think it works, it’s just not tested):

- [ ] A document (http://example.com/foobar/openapi.json), that uses a relative path (../components.json), that then uses another relative path (./foobar.json) - does this resolve to http://example.com/foobar.json (and not http://example.com/foobar/foobar.json)?

One thing that might be a bigger change: The promise of the parser package was to work in Node *and* in the browser. 

With the old `load` util, which we might even want to drop completely, I solved this with plugins (fetchUrls(), readFiles()) that are only accessible to a path export. (Otherwise webpack failed to tree shake it).

- [ ] Can we do something similar so bundle will work in a browser environment, too?

And somehow related:

- [ ] Can we stick to vitest instead of `node:assert` and `node:test`?

And something for bigger documents:

- [ ] I found that servers are blocking requests if you run too many in parallel, can you try https://github.com/digitalocean/openapi/blob/main/specification/DigitalOcean-public.v2.yaml and see if we need to chunk requests? (I needed to do that for the store)

packages/openapi-parser/src/utils/bundle.test.ts
```diff
@@ -0,0 +1,308 @@
+import assert from 'node:assert'
+import { randomUUID } from 'node:crypto'
+import fs from 'node:fs/promises'
+import { afterEach } from 'node:test'
```
inyourtime, line 4: Not sure, but should this be imported from vitest?

inyourtime, line 4: Not sure, but should this be imported from vitest?

packages/openapi-parser/src/utils/bundle.ts
```diff
@@ -0,0 +1,219 @@
...
+        }
+
+        if (typeof value === 'object' && '$ref' in value && typeof value['$ref'] === 'string') {
+          const ref = value['$ref']
+
+          if (!isExternalRef(ref)) {
+            return
+          }
+
+          const [prefix, path] = ref.split('/#', 2)
```
handrews, line 195: I might be missing something (my JavaScript skills are very rusty so maybe this is a JS regexp thing?), but if this is splitting on `/#` instead of `#` I am confused.

packages/openapi-parser/src/utils/bundle.ts
```diff
@@ -0,0 +1,219 @@
...
+
+          if (!cache.has(prefix)) {
+            cache.set(prefix, resolveRef(prefix))
+          }
+
+          // Resolve the remote ref
+          const result = await cache.get(prefix)
+
+          if (result.ok) {
+            root[key] = getNestedValue(result.data as any, getSegmentsFromPath(path))
```
handrews, line 205: This does not appear to handle any adjacent fields, correct?  (3.0 and 3.1 Path Item Object field combination, 3.1 Reference Object field overwriting, and 3.1 Schema Object adjacent fields which are complex to handle).

I am guessing it also does not support plain name fragments (declared by `$anchor` or `$dynamicAnchor` in OAS 3.1 Schema Objects.

Note that `$dynamicAnchor` anchors are usable as normal fragments with `$ref`, they just _also_ have special `$dynamicRef` behavior.  I am not suggesting that you can handle `$dynamicRef` here (it's theoretically possible, but very complex and can increase the size of your document exponentially).

I'm assuming no support for `$id` either?

hanspagel:
@handrews OMG, wonderful feedback, thank you so much 🙏

packages/openapi-parser/src/utils/bundle.ts
```diff
@@ -0,0 +1,219 @@
...
+ *   console.log('Failed to fetch JSON')
+ * }
+ * ```
+ */
+export async function fetchJson(url: string): Promise<ResolveResult> {
+  try {
+    const result = await fetch(url)
+
+    if (result.ok) {
+      const body = await result.json()
```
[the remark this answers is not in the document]
DemonHa: I updated the code to resolve both json and yaml content

packages/openapi-parser/src/utils/bundle.ts
```diff
@@ -0,0 +1,253 @@
...
+    }
+
+    return {
+      ok: false,
+    }
+  } catch {
+    return {
+      ok: false,
+    }
+  }
```
geoffgscott, line 76: Promise chain is cleaner here. 
```suggestion
    return await fetch(url).then(res => res.ok ? {
      ok: true,
      data: normalize(await res.text()),
    } : {
      ok: false,
    }
  ).catch(() => ({
    ok: false,
  }))
```

packages/openapi-parser/src/utils/bundle.ts
```diff
@@ -0,0 +1,253 @@
...
+ * await resolveRef('./schemas/user.json')
+ * // Invalid reference returns { ok: false }
+ * await resolveRef('#/components/schemas/User')
+ */
+async function resolveRef(ref: string): Promise<ResolveResult> {
+  if (isRemoteUrl(ref)) {
+    return fetchUrl(ref)
+  }
+
+  return readFile(ref)
```
geoffgscott, line 125:
```suggestion
  return isRemoteUrl(ref)) ? fetchUrl(ref) : readFile(ref)
```

packages/openapi-parser/src/utils/bundle.ts
```diff
@@ -0,0 +1,253 @@
...
+    const mergedPath = path.join(path.dirname(url.pathname), relativePath)
+    return new URL(mergedPath, base).toString()
+  }
+
+  return path.join(path.dirname(base), relativePath)
+}
+
+/**
+ * Bundles an OpenAPI specification by resolving all external references.
+ * This function traverses the input object recursively and replaces any external $ref
```
geoffgscott, line 179: @hanspagel @DemonHa should we be bundling by dereferencing? I think it would be more performant to put the $ref object into the components section (or an extension like `x-external`) and then replace the $ref. Right now if a bunch of paths share the same external $ref (pretty likely use case for external schemas) then we will be ballooning the size of the bundled document a lot.

geoffgscott (changes requested):

packages/openapi-parser/src/utils/bundle/bundle.ts
```diff
@@ -0,0 +1,284 @@
...
+ * @param input - The OpenAPI specification object to bundle
+ * @returns A promise that resolves when all references have been resolved
+ */
+export function bundle(input: UnknownObject, config: Config) {
+  // Cache for storing promises of resolved external references (URLs and local files)
+  // to avoid duplicate fetches/reads of the same resource
+  const cache = new Map<string, Promise<ResolveResult>>()
+
+  // Initialize storage for external references using a custom OpenAPI extension key
+  const EXTERNAL_KEY = 'x-external'
```
hanspagel, line 210: nice, I’d prefer to include "reference", so maybe:

* `x-externalReference`

packages/openapi-parser/src/utils/bundle/create-limiter.test.ts
```diff
@@ -0,0 +1,28 @@
...
+        return id
+      })
+
+    const tasks = [1, 2, 3, 4, 5].map(makeTask)
+    const results = await Promise.all(tasks)
+
+    expect(results).toEqual([1, 2, 3, 4, 5])
+    expect(Math.max(...maxObserved)).toBeLessThanOrEqual(2)
+  })
+})
```
hanspagel, line 28: 🔥

packages/openapi-parser/README.md
```diff
@@ -55,6 +55,30 @@ const specification = `{
...
+    }
+  }
+}
+
+// This will resolve and inline all external $refs (like user-schema.json)
+await bundle(document)
+
+console.log(document)
+```
+
```
hanspagel, line 81: can you add examples here?

* using the fetchUrls plugin (could be in a browser)
* using the fetchUrls & readFiles plugin (only in Node.js)
* changing the limit of concurrent requests

PR #5706, opened 2025-05-22
feat(api-reference): specification download button

antlio:
**Problem**

currently it is not possible to access both specification document extension. (yaml | json) on download.

**Solution**

this pr updates the download component style and introduces both extension to be downloaded at ease.

| after |
|------|
| image |

**Checklist**

### DIFF
--- a/.changeset/curly-roses-flow.md
+++ b/.changeset/curly-roses-flow.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
+feat: updates download link component
--- a/packages/api-reference/src/components/ApiReferenceLayout.vue
+++ b/packages/api-reference/src/components/ApiReferenceLayout.vue
@@ -156,8 +156,8 @@ onMounted(() => {
   history.scrollRestoration = 'manual'
 
   // Enable the spec download event bus
-  downloadEventBus.on(({ filename }) => {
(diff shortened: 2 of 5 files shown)


hanspagel (changes requested):
Cool!

The JSON I downloaded contains YAML, though:

https://1d63bccc-cc69-44ee-8e97-597587d5f180--scalar-deploy-preview.netlify.app/standalone-api-reference

And let’s add tests!

hwkr (changes requested):
Okay did some testing and it seems like the links aren't included in the tab order at all. `a` tags need to have an `href` to be considered interactive elements so if we're not able to set an href on them they should probably be `button`s. 

I think in terms of keyboard accessibility it's probably fine if you can just tab through the two buttons as though they're consecutive links on the page (e.g. they become visible once there is `:focus-visible` on one of the elements).

Let me know if you need any help with the implementation :)


https://github.com/user-attachments/assets/bc3e782f-1e53-4c9f-9ef0-8f8430535f59

hwkr (approved):
Okay! Rebased and pushed a fix for keyboard a11y :)

https://github.com/user-attachments/assets/13218877-a625-4db1-b297-da216ad6acd2

antlio:
great catch and addition tysm @hwkr

PR #5714, opened 2025-05-22
chore(Scalar.AspNetCore): remove obsolete license URL property

xC0dex:
This PR removes the deprecated license URL property. This is some preparation for the prefix reservation.


amritk:
🫡

PR #5721, opened 2025-05-23
fix: theme laserwave string quotes

rolldeep-stepmerrily:
**Problem**

Currently, the `laserwave` theme option in the documentation is missing single quotes, which makes it inconsistent with other theme options and could cause confusion for developers trying to use this theme.

**Solution**

With this PR, I've added the missing single quotes around `laserwave` in the theme options list to maintain consistency with other theme string values.

**Checklist**

- [x] I've added an explanation *why* this change is needed.


xC0dex (approved):
Great catch! Thank you 🥳

hanspagel:
nobody but you saw this 🤟

hanspagel (approved):

PR #5723, opened 2025-05-23
feat(api-reference): discriminator support

### LINKED ISSUE #795: Options could be clearer when using a discriminator

antlio:
**Problem**

currently in discriminator presence reference is not retrieving mapping nor resolving schemas in its ui.

**Solution**

this pr introduces schema discriminator in order to be able to select schema accordingly and fixes #795 .

| after |
|------|
| image |

**Test**
[Open API spec doc](https://sandbox.scalar.com/files/FRNqy/openapi.yaml)

**Checklist**

### DIFF
--- a/packages/api-reference/src/components/Content/Schema/Schema.vue
+++ b/packages/api-reference/src/components/Content/Schema/Schema.vue
@@ -2,8 +2,14 @@
 import { Disclosure, DisclosureButton, DisclosurePanel } from '@headlessui/vue'
 import { ScalarIcon, ScalarMarkdown } from '@scalar/components'
 import type { OpenAPIV2, OpenAPIV3, OpenAPIV3_1 } from '@scalar/openapi-types'
-import { computed } from 'vue'
+import { computed, ref, watch } from 'vue'
 
+import {
+  getDiscriminatorMapping,
+  getDiscriminatorPropertyName,
+  mergeDiscriminatorSchemas,
+} from '@/components/Content/Schema/helpers/schema-discriminator'
+import SchemaDiscriminator from '@/components/Content/Schema/SchemaDiscriminator.vue'
 import ScreenReader from '@/components/ScreenReader.vue'
 
 import SchemaHeading from './SchemaHeading.vue'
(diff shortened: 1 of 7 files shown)


packages/api-reference/src/components/Content/Schema/Schema.vue
```diff
@@ -36,9 +42,58 @@ const props = withDefaults(
...
 
+const selectedDiscriminatorType = ref<string>('')
+
+/** Returns the discriminator mapping */
+const discriminatorMapping = computed(() => {
+  if (!props.value) {
+    return undefined
+  }
+
+  return getDiscriminatorMapping(props.value as OpenAPIV3_1.SchemaObject)
```
amritk, line 56: The `as` should be in the `getDiscriminatorMapping` method if its a type guard

amritk: technically here we are overwriting undefined which is bad

amritk (approved):
SO we been adding to this quite a bit to fix each issue and its getting a bit bloated. I think we write extensive tests, and once we switch to new store we can re-write this much more efficiently and type safe. Approving for now though!

cameronrohani (approved):

PR #5713, opened 2025-05-22
fix(api-reference): sets schema name property heading

antlio:
**Problem**

currently schema name of type object are not retrieved in the reference property level.

**Solution**

this pr adds a schema name function to handle both array and object type schema name retrieval and fixes #5711 .

| before | after |
| -------|------|
| image | image |
| only type object displayed | type object and model schema name displayed |

**Checklist**

I’ve gone through the following:

- [x] I’ve added an explanation _why_ this change is needed.
- [x] I’ve added a changeset (`pnpm changeset`).
- [x] I’ve added tests for the regression or new feature.
- [ ] I’ve updated the documentation.

PR #5724, opened 2025-05-23
fix(api-reference): schema property heading type

antlio:
**Solution**

following up with #5713 type were both handled the same way, this pr updates the logic to add array suffix accordingly.

| before | after |
| -------|------|
| image | image |

**Checklist**


hanspagel (approved):
cool!

PR #5725, opened 2025-05-23
fix(api-reference): schema composition style and value

antlio:
**Problem**

following recent refactoring some style got lost and also some properties value fallback is not happening anymore.

**Solution**

this pr updates the schema composition style along displaying value in some cases.

| before | after |
| -------|------|
| image | image |

**Checklist**


hanspagel (approved):
better every day

cameronrohani (approved):

PR #5729, opened 2025-05-23
chore(use-codemirror): remove unused packages

hanspagel:
**Problem**

I saw there is an unused dependency in @scalar/use-codemirror and a dependency that we just use for a super small type.

**Solution**

I removed them. :)

**Checklist**


amritk (approved):
Remove all the things

geoffgscott (approved):
image

PR #5733, opened 2025-05-23
fix(api-reference): schema references name in property heading

### LINKED ISSUE #5552: Response type for record is object
### What happens?

If an action method in an ASP.NET Core controller returns a [record](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/record), the `Responses` will contain `object` instead of the actual model.

```
<PackageReference Include="Scalar.AspNetCore" Version="2.2.3" />
```

I expect the actual record type to be displayed in the `Responses`.

### OpenAPI Document

_No response_

antlio:
**Problem**

currently schema references name does not get displayed in the property heading.

**Solution**

this pr update schema property heading component logic and fixes #5552.

| before | after |
| -------|------|
| image | image |

**Checklist**

### DIFF
--- a/.changeset/yellow-games-own.md
+++ b/.changeset/yellow-games-own.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
+fix: displays schema references name in property heading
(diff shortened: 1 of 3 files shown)


amritk (approved):
I see tests, I merge simple as that

PR #5734, opened 2025-05-23
refactor(openapi-parser): no more wildcard exports, no more pipeline syntax

hanspagel:
**Problem**

Our ESM CDN and our Nuxt integration has issues with ajv, though it doesn’t even use ajv (it’s used in the parser for the validation). I think it’s because of our wildcard exports and/or because of the pipeline syntax (`openapi().load()…`) that just has to load all the code. 

**Solution**

With this PR we’re removing wildcard exports, a few barrel files, some unused code, limit the exports to the stuff we promote and we use, and removes the pipeline syntax (was deprecated already).

Tagged as a minor version.

**Checklist**


geoffgscott:
The barrel files should remain for all folders in ./packages
 
IMO we should also be using them in FE projects but it's less important there. They provide proper scoped exports for directories and work easily with out build tooling auto exports.

amritk:
@geoffgscott barrel files have fallen out of favor lately in js land so it's best to stay away from them if possible! Use only sparingly and never use internally.

I _think_ if you don't use export * it's much better as it could get tree shaken. BUT typescript will still load every module in the barrel file to determine which one you need.

marclave (approved):

PR #5740, opened 2025-05-25
fix(oas-utils): do not add content-type for multipart requests

nsychev:
**Problem**

Currently, when building code examples for multipart requests, `Content-Type: multipart/form-data` header is added to the request. It is wrong behavior, because the correct header should contain **required** parameter `boundary` as per https://datatracker.ietf.org/doc/html/rfc2046#section-5.1.1. Often this value is desired by the HTTP client libraries because boundary should not be contained inside any of the parts.

I found out that if one explicitly passes header without boundary, both Python (requests) and curl do not overwrite the header and thus send malformed requests.

**Solution**

With this PR, header is removed.

**Checklist**

### DIFF
--- a/.changeset/angry-icons-applaud.md
+++ b/.changeset/angry-icons-applaud.md
@@ -1,0 +1,6 @@
+---
+'@scalar/api-client': patch
+'@scalar/oas-utils': patch
+---
+
+fix(oas-utils): do not add content-type for multipart requests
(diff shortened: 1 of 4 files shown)


nsychev:
I think I fixed the relevant test here: https://github.com/scalar/scalar/blob/8af87d5b69f7c0ef4fb23579cb9f00aac71b4d07/packages/oas-utils/src/entities/spec/request-example.test.ts. If you feel there is a need for more, I may try to add something else.

PR #5746, opened 2025-05-26
chore(Scalar.AspNetCore): make copy command windows compatible

xC0dex:
**Problem**

Currently, I have to copy the `standalone.js` manually to the target folder during local development.

**Solution**

Now I can use the existing `copy:standalone` command.

**Checklist**


amritk (approved):
Doing the windows work!

PR #5756, opened 2025-05-27
feat(api-reference): discriminator support 2/2

antlio:
**Solution**

this pr adds the last part for the discriminator support including:
- ui update to set the discriminator from the property field
- example request update on change

| after |
| -------|
| image |

**Checklist**


marclave (approved):

packages/api-reference/src/components/Content/Schema/Schema.vue
```diff
@@ -1,23 +1,23 @@
...
+import type { Schemas } from '@/features/Operation/types/schemas'
+import { DISCRIMINATOR_CONTEXT } from '@/hooks/useDiscriminator'
 
 import SchemaHeading from './SchemaHeading.vue'
 import SchemaProperty from './SchemaProperty.vue'
 
 const props = withDefaults(
   defineProps<{
     value?:
+      | OpenAPIV2.DefinitionsObject
```
amritk, line 17: @antlio we are upgrading now so we only need V3_1 everywhere. We should actually alias it somewhere as the current OpenAPI version in references then use that everywhere. I will do it in my sidebar PR 😄

packages/api-reference/src/components/Content/Schema/Schema.vue
```diff
@@ -32,58 +32,69 @@ const props = withDefaults(
...
+    const mergedSchema = discriminatorContext.value.mergedSchema
+    const originalSchema = props.value
 
-  // If no discriminator or no selected type, return original value
-  if (!discriminatorMapping.value || !selectedDiscriminatorType.value) {
-    return props.value
+    // Only use resolved schema if both original and resolved are objects
+    // This ensures we don't mix incompatible schema types
+    if (
+      originalSchema &&
```
amritk, line 84: this is wild lol

packages/api-reference/src/components/Content/Schema/Schema.vue
```diff
@@ -124,16 +128,19 @@ watch(
       <!-- Schema description -->
       <div
         v-if="
-          resolvedSchema?.description &&
-          typeof resolvedSchema.description === 'string' &&
-          !resolvedSchema.allOf &&
-          !resolvedSchema.oneOf &&
-          !resolvedSchema.anyOf &&
+          schema &&
```
amritk, line 131: I'd probably use a helper function here with some docs that explains what its looking for

amritk: or just a var

amritk (changes requested):
Probably just clean up some of the crazy IF statements in the templates, the rest looks good 👍🏾 

things are getting complicated!

packages/api-reference/src/components/Content/Schema/Schema.vue
antlio: lgmt!

packages/api-reference/src/components/Content/Schema/Schema.vue
antlio: done in #5782

PR #5759, opened 2025-05-28
fix(api-reference): reference header index

antlio:
**Problem**

currently in reference header presence, on scroll the sidebar is going above.

**Solution**

this pr increases reference header z index to stay on top.

| before | after |
| -------|------|
| image | image |

**Checklist**

### DIFF
--- a/packages/api-reference/src/components/ApiReferenceLayout.vue
+++ b/packages/api-reference/src/components/ApiReferenceLayout.vue
@@ -440,7 +440,7 @@ const themeStyleTag = computed(
   grid-area: header;
   position: sticky;
   top: var(--scalar-custom-header-height, 0px);
-  z-index: 10;
+  z-index: 1000;
 
   height: var(--scalar-header-height, 0px);
 }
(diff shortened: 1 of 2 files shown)


packages/api-reference/src/components/ApiReferenceLayout.vue
```diff
@@ -440,7 +440,7 @@ const themeStyleTag = computed(
   grid-area: header;
   position: sticky;
   top: var(--scalar-custom-header-height, 0px);
-  z-index: 10;
+  z-index: 1000;
```
amritk, line 443: is this in line with our zIndex strategy @hwkr @cameronrohani ?

antlio: what do you mean? didn't find much about that so as a quick fix since it's in prod i increased it

marclave (approved):
🔥

packages/api-reference/src/components/ApiReferenceLayout.vue
hwkr: Well because of stacking contexts things can still get covered pretty easily unfortunately but yes we do have a general strategy. This should probably actually only be 1 or maybe 2

PR #5762, opened 2025-05-28
fix(build-tooling): exclude test files from esbuild entry points

inyourtime:
This PR updates the `entryPoints` configuration for the `esbuild` build process to ensure that test and spec files (`.test.ts`, `.spec.ts`, `.test.js`, `.spec.js`) are excluded from the final bundle.


amritk (approved):
Sounds good to me! Will see what @geoffgscott says.

Nice PR!

PR #5766, opened 2025-05-28
feat(Scalar.AspNetCore): add support for code samples

xC0dex:
This PR adds a new `[CodeSample]` attribute and one extension `CodeSample()` method to the `Scalar.AspNetCore` package to support custom code samples. The OpenAPI generator extensions are also updated to make use of the new attribute and extension method.
I also updated the related documentation to be easier to follow.

Usage:
```csharp
// Extension methods
app.MapPost("/orders", CreateOrder)
    .CodeSample("fetch('/orders', { method: 'POST' })", ScalarTarget.JavaScript, "Create Order")
    .CodeSample("curl -X POST /orders", ScalarTarget.Shell, "Create Order with cURL")

// Attribute
[CodeSample("fetch('/products').then(r => r.json())", ScalarTarget.JavaScript)]
public IActionResult GetProducts() { }
```

What do you think about the name `CodeSample`?

**Checklist**

### DIFF
--- a/.changeset/cold-camels-smell.md
+++ b/.changeset/cold-camels-smell.md
@@ -1,0 +1,5 @@
+---
+'@scalar/aspnetcore': patch
+---
+
+feat(Scalar.AspNetCore): add support for code samples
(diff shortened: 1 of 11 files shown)


marclave (approved):
ncie work @xC0dex ✨

PR #5777, opened 2025-05-28
feat(themes): use .scalar-app as the base class for all tailwind styles

hwkr:
Switches the components, references and the client to all use the .scalar-app scoping class so the styles can be deduped.

Unfortunately the deduping only seems to happen in production environments so we still have some duplication when running the dev servers.
 
**Checklist**


marclave (approved):
cam & i approve!

PR #5779, opened 2025-05-29
fix(api-reference): fix error when using complex type definitions

Borisich:
Some complex types in model definitions causes crash of appropriate section:
image

And instead of method description we can see following:
image

example file in attachment. Strange, but it normally works at https://docs.scalar.com/swagger-editor

[doc-dev-en.json](https://github.com/user-attachments/files/20499861/doc-dev-en.json)


antlio:
> > great catch @Borisich! do you mind adding a changeset? can do if need
> 
> Please help me with it :)

you can run from the root repository `pnpm changeset` and here you can skip major/minor selection so we can patch it - then add your change title (no need to add api reference since it is scoped) - let me know if that helps!

PR #5794, opened 2025-05-30
chore: move to esbuild for integration packages

### LINKED ISSUE #4009: Generated TypeScript types does not work on ESM

inyourtime:
**Problem**

After #5517 was merged, it broke NodeNext compatibility for the following integrations: `fastify`, `express`, `hono`, because these packages still use `rollup`

By switching to `esbuild`, we can restore NodeNext compatibility.

This resolves #4009 

**Checklist**


amritk (approved):
Looks fine to me! Thanks for the contribution as always! Will let @geoffgscott merge as its his domain.

PR #5802, opened 2025-05-30
fix: handle none theme

marclave:
**Problem**

Currently theme set to none breaks rendering

**Solution**

With this PR we handle the none theme

**Checklist**


cameronrohani (approved):
how does he do it

PR #5811, opened 2025-05-30
window shopping

amritk:
# Update

We will probably just fix the types in the new store, I have just been switching this on and off while I'm working so I got proper types.

**Problem**

Currently, the openapi types allow you to write whatever you want, due to the `& anyAttribute` they basically act like `any`

| before | after |
| -------|------|
| image | image |
| anything goes | we are restricted |

**Solution**

With this PR we remove that bit so the types are actually working. This obviously has knock on consequences in our repo. I tried to deal with em as safely as possible but in some places I couldn't discern what we were trying to do, other places I didn't wanna mess it up. I will tag the relevant authors in the files so we can clean up the last few type casts and ignores.

I would like to get this in before new sidebar + parsedSpec removal so we can be safe

- added a littler helper to narrow down `ReferenceObject` unions
- tried to centralize some of our http-method helpers as they were duplicated and everywhere, will convert it to a lib
- still WIP as I fix up the types and remove the `@ts-nochecks`
- can probably add some more helpers to clean it up but im cooked

**Checklist**


integrations/fastify/src/fastifyApiReference.ts
```diff
@@ -198,20 +197,19 @@ const fastifyApiReference = fp<
...
       method: 'GET',
       url: openApiSpecUrlJson,
-      // @ts-ignore We don’t know whether @fastify/swagger is loaded.
       schema: schemaToHideRoute,
       ...hooks,
       ...(options.logLevel && { logLevel: options.logLevel }),
       async handler(_, reply) {
         const spec = normalize(specSource.get())
-        const filename: string = await getSpecFilenameSlug(spec)
+        const filename: string = await getSpecFilenameSlug(spec as OpenAPI.Document)
```
amritk, line 212: not a fan of the type cast, but its tested. We can add a type guard later

packages/api-reference/src/components/Content/Schema/helpers/merge-all-of-schemas.ts
```diff
@@ -1,3 +1,4 @@
+// @ts-nocheck
```
amritk, line 1: @antlio again, maybe we can work on these together

amritk:
The new store is coming, will probably just keep this up for reference for now. I will start seeing how the components behave without parsedSpec though

packages/workspace-store/src/create-server-workspace-store.ts
```diff
@@ -319,8 +272,8 @@ export function createServerWorkspaceStore(workspaceProps: CreateServerWorkspace
 
       // add the assets
       assets[meta.name] = {
-        components: documentV3.schema?.components,
-        operations: documentV3.schema?.paths && escapePaths(filterHttpMethodsOnly(documentV3.schema.paths as any)),
+        components: documentV3.components,
```
amritk, line 275: @DemonHa was this working before?

DemonHa: it was

amritk: 😕

amritk: where does schema come from?

amritk:
Farewell friend, types will be fixed in the new store 😿

PR #5812, opened 2025-05-30
feat(proxy): harden cidr ranges

bgrcs:
We need to ensure that the proxy does not allow any traffic to private ip ranges (including GCP loopback and metadata cidr) on both initial request and any redirects. 

The proxy has been isolated, but to prevent some easy mistakes in the future, I'm just adding these for now. 

### DIFF
--- a/packages/api-client/src/libs/send-request/create-request-operation.test.ts
+++ b/packages/api-client/src/libs/send-request/create-request-operation.test.ts
@@ -213,30 +213,6 @@ describe('create-request-operation', () => {
     })
   })
 
-  // TODO: this doesn't actually hit the proxy due to 127.0.0.1
-  it('reaches the echo server *with* the proxy', async () => {
-    const [error, requestOperation] = createRequestOperation(
-      createRequestPayload({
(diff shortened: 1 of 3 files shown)


marclave (approved):

amritk (approved):

ramsey:
This change breaks use of the proxy when running a local development environment or hosting local development environments behind a VPN.

bgrcs:
@ramsey Thanks for flagging 🙏; we'll add an extra env flag that will include these optionally.

bgrcs:
@ramsey So just to check, you are just running an instance of our proxy locally?

Would this work well for your use case: https://github.com/scalar/scalar/pull/5959?

ramsey:
@bgrcs Yes! I think that will work well. For the meantime, I've just changed that to `true` and have been using it that way locally, so this would be a perfect solution to do that with configuration.

bgrcs:
@ramsey Great, will get that merged.

Apologies for the inconvenience, we typically just use it internally and don't track breaking changes too closely 😄

PR #5813, opened 2025-05-31
fix(api-client): move global styles out of client layout

hwkr:
Since all the client styles are bundled into a single `styles.css` we can't have global styles that will affect any package that consumes the client (e.g. the references).

**Checklist**

### DIFF
--- a/.changeset/blue-pears-explode.md
+++ b/.changeset/blue-pears-explode.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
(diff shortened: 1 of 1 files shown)


amritk (approved):

antlio:
this seems to bring back the following:

| before | after |
| -------|------|
| image | image |

hwkr:
@antlio that might just be the playground but I'll have a look! Did you run the dev server to see that?

PR #5816, opened 2025-06-01
fix(api-client): override Electron default User-Agent header

### LINKED ISSUE #5515: User-Agent header won't take effect in API Client
### What happens?

I just downloaded the desktop client (Version 0.1.184 (250424tbptkvii3)) for some api testing because of huge post body that I have to test.
There I noticed that the User-Agent header field that I've added to my request is overwritten by the Electron default header.

The header field entry should be respected-

### OpenAPI Document

_No response_

inyourtime:
**Problem**

Electron automatically sets a default `User-Agent` header on requests, so the user-defined `User-Agent` is not used.

**Solution**

With this PR
- A custom `User-Agent` is passed via an `X-Scalar-User-Agent` header when running in Electron.
- `onBeforeSendHeaders` is used to replace the default `User-Agent` header with the value from `X-Scalar-User-Agent`.

This ensures that the final `User-Agent` reflects the user-defined value rather than Electron’s default behavior.

I'm not sure if this is the best approach — open to suggestions or alternative ideas 🤔.
fixes #5515 

before:
Screenshot 2568-06-01 at 10 23 42

after:
Screenshot 2568-06-01 at 10 20 12
  
**Checklist**

### DIFF
--- a/.changeset/real-sloths-fold.md
+++ b/.changeset/real-sloths-fold.md
@@ -1,0 +1,6 @@
+---
+'@scalar/api-client': patch
+'scalar-app': patch
+---
+
+fix: override Electron default User-Agent header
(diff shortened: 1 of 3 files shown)


antlio (approved):
great pr @inyourtime - should we add a test?

approach is consistent with others - looks good to me for the time being :)))

geoffgscott (approved):
Love it.

PR #5833, opened 2025-06-02
feat(helpers): add new helpers package

amritk:
**Problem**

Currently, we just slapped all helpers in OAS-utils but they were unrelated and forced dependencies.

**Solution**

With this PR we create a new package with 0 deps and individual exports for max performance! I tried using the barrel files but couldn't get it to tree shake out functions even using @pure, with this way there's no need as its built in! Add all the files you want!

I kept the old exports from OAS-utils but added @deprecated so we can move em over as we go. Also categorized them, not sure if we wanted them categorized or flat in one folder. I'm good with either.

Also added more tests for some decent coverage

image

**Checklist**


hwkr:
Yooooo let's goooo, this is 🔥

marclave (approved):
hes helpin!

amritk:
Mergin!

PR #5834, opened 2025-06-03
just lookin

amritk:


amritk:
This has been replaced

https://github.com/scalar/scalar/pull/5869

PR #5837, opened 2025-06-03
feat(workspace-store, openapi-parser): incremental bundle for openapi documents

DemonHa:
**Solution**

With this PR …
- **Support for external references and chunking in the workspace store**
Adds support for handling external references within the workspace store, in conjunction with a chunking strategy. This allows large or modular documents to be resolved and stored more efficiently.
- **Hook into the reference resolution lifecycle**
Adds the ability to tap into the $ref resolution process, enabling custom behavior or tracking during reference resolution.
- **Partial bundle generation**
Supports generating a partial bundle for a given document, including all its dependencies. This is useful for modular or incremental document handling.
- **Custom fetch function support**
Adds support for passing a custom fetch function. This allows flexibility for client-specific requirements, such as proxying.
> Note: Proxying is not handled directly within the plugin to avoid enforcing a specific proxy implementation. This keeps the plugin more flexible and adaptable.
- **Fix `getHash` function to produce alphanumeric values**
Ensures that the getHash function generates alphanumeric strings only. This prevents issues where numeric-like strings are misinterpreted as array indices by `setValueAtPath` function.
- **Preserve non-operation properties in the server workspace store**
The server-side workspace store now retains original non-operation properties in their original object form, preserving structure and metadata as needed.

**Checklist**


marclave (approved):
👀

PR #5839, opened 2025-06-03
fix(openapi-parser): Migrate serialization style

matthyk:
**Problem**

Currently, the `collectionFormat` keyword is just ignored.

**Solution**

With this PR the `collectionFormat` keyword is migrated to the two new keywords `style` and `explode`. If the old behavior described by `collectionFormat` cannot be described with the new keywords, it is simply removed and the default behavior for the respective parameter types is used.

**Checklist**

### DIFF
--- a/.changeset/fresh-buckets-refuse.md
+++ b/.changeset/fresh-buckets-refuse.md
@@ -1,0 +1,5 @@
+---
+'@scalar/openapi-parser': patch
+---
+
+fix: collectionFormat is not migrated to new style and explode keywords
--- a/packages/openapi-parser/src/utils/upgrade-from-two-to-three.test.ts
+++ b/packages/openapi-parser/src/utils/upgrade-from-two-to-three.test.ts
@@ -634,4 +634,202 @@ describe('upgradeFromTwoToThree', () => {
       },
     })
   })
+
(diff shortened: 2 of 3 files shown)


matthyk:
Don't know why the CI failed and the logs are kinda buggy for me and do not show any errors

antlio:
thank you for that great pr! the logs are def off, my way around is to check at the very end [from the files](https://github.com/scalar/scalar/pull/5839/files). i just re-ran the test - same issue as had before on bundle test file cc @DemonHa

DemonHa:
> thank you for that great pr! the logs are def off, my way around is to check at the very end [from the files](https://github.com/scalar/scalar/pull/5839/files). i just re-ran the test - same issue as had before on bundle test file cc @DemonHa

About the failing tests, running again should fix it for now. I have a fix on [this PR](https://github.com/scalar/scalar/pull/5837) explained in more details

marclave (approved):
awesome PR @matthyk !!

Gunna wait for @amritk to take a look at this one ✨

matthyk:
I used [this](https://swagger.io/docs/specification/v3_0/serialization/) as reference 👍

PR #5845, opened 2025-06-03
chore(openapi-parser): bring back pipeline syntax

DemonHa:
**Problem**

**Reintroduce Pipeline Syntax for Backwards Compatibility**

In PR #5734, we removed the pipeline syntax. However, that change was released as a minor version, which violates semantic versioning since it introduced a breaking change.

To address this, this PR restores the pipeline syntax to ensure backwards compatibility until the next major release.

> ⚠️ Note: While the syntax is temporarily available, we strongly discourage its use, as it will be permanently removed in the next major version.

Documentation for the pipeline syntax has not been updated intentionally, to discourage adoption going forward.

**Checklist**

### DIFF
--- a/.changeset/sour-planes-protect.md
+++ b/.changeset/sour-planes-protect.md
@@ -1,0 +1,6 @@
+---
+'@scalar/workspace-store': minor
+'@scalar/openapi-parser': minor
+---
+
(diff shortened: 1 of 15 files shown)


marclave (approved):
we are so back

PR #5846, opened 2025-06-03
feat(api-reference): new spec traversal

amritk:
**Problem**

Currently, our sidebar doesn't allow us to tag webhooks. Also contains tonnes of loops again and again.

**Solution**

With this PR we only traverse the spec once and gather all of the data we need for various features. In this PR it grabs what we need for the new sidebar (but doesn't put it into use yet so its a safe merge). The next PR will be integration.

Post release we can add tagged schema/models  and pull the search indexing into this document traversal as well 😄 

**Checklist**


marclave (approved):
👁️ 👄 👁️

PR #5847, opened 2025-06-03
feat(api-references): Flattens layouts and add client workspace store

geoffgscott:
Initial bootstrap work for the client workspace store. 

This also flattens the layout code to prevent passing the store down multiple layers.


amritk (approved):
Seems reasonable, didn't test as its not passing yet but the architecture seems good! I'm all for reduction in component depth

PR #5850, opened 2025-06-04
feat(openapi-parser, workspace-store): handle hash collisions

DemonHa:
**Problem**

We currently hash source URLs or file paths to reduce bundle size. However, there is no mechanism to handle hash collisions, which can lead to different documents being assigned the same hash and potentially overwriting each other.

> ⚠️ Note: While hash collisions are very unlikely, especially with good hash functions, our implementation uses short hash strings for compactness. This slightly increases the risk of collisions, so we're handling this edge case proactively.

**Solution**

This PR introduces logic to ensure that generated hash values are unique, even in the event of a collision. If a collision occurs, the system will automatically resolve it by generating an alternative unique hash. This guarantees that different documents will always be assigned distinct hashes and prevents unintended overwrites.

**Checklist**

- [X] I’ve added tests for the regression or new feature.


marclave (approved):

geoffgscott:
Closed in favor of a different strategy.

PR #5852, opened 2025-06-04
ci: do not run cron job for forks

matthyk:
**Problem**

Currently, the `Test CDN Build (jsdelivr)` workflow is run every hour, even for forked repositories. I think for forked repositories the workflow does not have to be run every hour either.

**Solution**

With this PR, the workflow is only run if the trigger event is not `schedule` or if it is run for the `scalar/scalar` repository.

**Checklist**


geoffgscott (approved):

matthyk:
The failing CI is not related to my change, is it?

marclave:
@matthyk im not entirely sure! let me rebase ✨

PR #5863, opened 2025-06-04
feat(api-reference): use live sync for document updates to client store

### LINKED ISSUE #4290: Use live-sync to update the store on spec changes

amritk:
Waiting on #5869 then this needs to be re-tested and possibly fixed 😄 

Special request for @bgrcs 
closes #4290

**Problem**

Previously, we used to re-create the store, currently we are just adding a new store on top any time theres a change to the document.

**Solution**

With this PR we leverage our live sync tech to update the store instead of re-create it. I know it currently doesn't cover every case but its a big improvement over what we have now and it will be easy to add cases.

Test reactivity thoroughly

**Checklist**

### DIFF
--- a/.changeset/dull-cups-carry.md
+++ b/.changeset/dull-cups-carry.md
@@ -1,0 +1,6 @@
+---
+'@scalar/api-reference': patch
+'@scalar/api-client': patch
+---
+
+feat: use live sync on document change for store updates


bgrcs (approved):

amritk:
closing in favor of 
https://github.com/scalar/scalar/pull/5945/

PR #5865, opened 2025-06-04
fix(openapi-parser): multi entry build for bundler plugins

DemonHa:
**Problem**

Currently, we are a default export for all bundler plugins. We can't do that because it will cause compatibility issues.

**Solution**

With this PR we are doing multi entry build to fix this issue

**Checklist**


bgrcs (approved):
needs updating consumers but looks good 👍

marclave (approved):
🔥

geoffgscott (approved):

PR #5868, opened 2025-06-04
feat(workspace-store): create a sync client store

DemonHa:
**Problem**

Currently, on the `api-client` package we don't support top level awaits

**Solution**

With this PR we introduce a sync store so we can use it on the `api-client` package

**Checklist**


marclave (approved):
🔥 🔥 🔥 🔥 🔥

PR #5870, opened 2025-06-04
fix(api-reference): fix markdown styles pt2

cameronrohani:
updated markdown styles

image

note reduced header margin from figmas as we decided this was the play 

(bonus also changed document selector UI + download OpenAPI file badge styles)


marclave (approved):

antlio:
:wicked:

PR #5872, opened 2025-06-05
feat(components): deprecate ScalarIcon component

hwkr:
**Problem**

Currently, …

**Solution**

With this PR …

**Checklist**


hwkr:
I wasn't sure which checkmark to use for the checkboxes, Phosphor has several check options and I think both the normal check and the fat check could be good options but the normal check is closer to what we currently have.

packages/components/src/components/ScalarIcon/ScalarIcon.vue
```diff
@@ -1,54 +1,32 @@
+<script lang="ts">
+/**
+ * Scalar Icon Wrapper
+ *
+ * @deprecated Use the icons from the `@scalar/icons` package instead.
```
amritk, line 5: 👍🏾

amritk (approved):
Just did code check, will let @cameronrohani check the appearance. But very nice deprecation strategy with the legacy adapter!

Issue #4356, opened 2025-01-02
Cannot pull the official docker image


kiwamizamurai:
### What happens?

```shell
❯ docker pull scalarapi/api-reference:latest
Error response from daemon: manifest for scalarapi/api-reference:latest not found: manifest unknown: manifest unknown
```

https://github.com/scalar/scalar/tree/main/packages/api-reference/docker

### What did you expect to happen?

Pulled successfully

hanspagel:
True! It’s not published yet. We’ll do in the next days I assume. Sorry for the inconvenience.

kiwamizamurai:
@hanspagel 
Thanks!
If there’s anything I can do to help, please don’t hesitate to let me know.
I use your amazing product in Japan^^

hanspagel:
Still not published

https://hub.docker.com/r/scalarapi/api-reference

hanspagel:
Finally!

https://hub.docker.com/repository/docker/scalarapi/api-reference

PR #5892, opened 2025-06-07
feat(docker): improve docker integration

xC0dex:
Context: I wanted to build an MVP Aspire Scalar integration based on the docker image. That's why I touched the docker related code. 

This PR improves the docker integration by:

- migrating to the new JS configuration
- updating the `Caddyfile`
- providing a local (and GZIP compressed) version of the API Reference instead of the CDN version
- moving to a dedicated `integration/docker` folder (I'd expect it there)

**Off-Topic**
- We should consider switching to blacksmith docker workflows https://docs.blacksmith.sh/blacksmith-caching/docker-builds

**Todos**
- it looks like the image was never published yet https://github.com/scalar/scalar/issues/4356

**Checklist**


xC0dex:
Please let me know what you think. Moving to a dedicated `integrations/docker` folder isn't necessary. I can revert this.

amritk:
this ones for @geoffgscott

xC0dex:
(Maybe) this will change soon. I'm currently building a MVP for .NET Aspire based on .NET API wrapped in a docker image.

geoffgscott:
@xC0dex very happy to see this move to integrations. Can you delete the old directory so we are fully moved over? Once that is done I am happy to merge. 

I would prefer use to `node:22-bookwork-slim` for the base image to move from Alpine (which can be a bit quirky) to Debian. We would drop caddy in favour of a simple node HTTP server in that case. Can be done in a separate PR (or this one) but do you have any objections to moving away from Alpine linux?

xC0dex:
Hey @geoffgscott, 

Thanks for the feedback! I’ve always considered Caddy to be lightweight and performant compared to a node-based HTTP server, which is why I chose to stick with it. That said, I haven’t had any negative experiences with alpine-based images so far, but I’m totally fine with switching to a base image that offers better compatibility. The main reason I had in mind when choosing alpine was the smaller image size.

I'll remove the old docker folder later today!

geoffgscott (approved):
Approved pending file cleanup. Looks great.

hanspagel (approved):
love it

PR #5896, opened 2025-06-08
feat(snippetz): add python/httpx plugin

ClanEver:
**Problem**

Add python [httpx](https://github.com/encode/httpx/) as client.

**Solution**

Since httpx and requests have almost same usage patterns, I extracted the common code into `requestsLike.ts`. I also added conversion from `true/false/null` to `True/False/None` and related tests.

**Checklist**


amritk (approved):
Absolute 🔥

PR #5899, opened 2025-06-09
feat(openapi-parser): introduce parseJson and parseYaml plugins for bundler

DemonHa:
**Problem**

Currently, the bundler does not support raw YAML or JSON strings as input, which limits flexibility when consuming APIs from dynamic or in-memory sources.

**Solution**

This PR adds support for parsing both YAML and JSON strings as valid input to the bundler, allowing them to be processed just like file-based documents.

**Checklist**


marclave (approved):
nice!! :)

PR #5906, opened 2025-06-09
fix(api-client): address bar bg

antlio:
**Problem**

recent changes in #5795 and #5833 changed the approach that was used to display the background loading state, as highlighted in #5904

**Solution**

this pr updates the address bar style to bring bar the background on load.

| state | preview |
| -------|------|
| before | image | 
| after | image |

we might be able to remove the `backgroundColor` from the helper after this cc @amritk?

**Checklist**

### DIFF
--- a/.changeset/famous-pears-jog.md
+++ b/.changeset/famous-pears-jog.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
+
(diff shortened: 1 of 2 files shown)


amritk (approved):
Wicked fix! I think its still used in a few places though

PR #5910, opened 2025-06-09
feat(api-reference): Re-organize workspace initialization

geoffgscott:
Shifts state between components.

### DIFF
--- a/biome.json
+++ b/biome.json
@@ -172,7 +172,7 @@
       "performance": {
         "noDelete": "off",
         "noReExportAll": "warn",
-        "noBarrelFile": "off"
+        "noBarrelFile": "warn"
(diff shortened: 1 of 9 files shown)


biome.json
```diff
@@ -172,7 +172,7 @@
       "performance": {
         "noDelete": "off",
         "noReExportAll": "warn",
-        "noBarrelFile": "off"
+        "noBarrelFile": "warn"
```
amritk, line 175: turned this back on?

amritk (approved):
Should be fine

biome.json
geoffgscott: Yeah. Need to think about it a bit more.

PR #5943, opened 2025-06-12
fix(api-reference): scroll to top to remove hash

amritk:
**Problem**

Currently, our scroll listener was on a dom element

**Solution**

With this PR we move it to the window. I tested pathRouting but it doesn't seem like we even use hashPrefix anymore. Opened up another ticket to look into that one.

**Checklist**


marclave (approved):
tested and works :)

PR #5946, opened 2025-06-13
feat(workspace-store): improve workspace store validation

DemonHa:
**Problem**

There is currently no validation or type coercion for the input document. This means any arbitrary input is accepted, and we implicitly assume it to be a valid OpenAPI document. This can lead to runtime issues and unreliable behavior when the input is malformed or doesn't conform to the expected structure.

**Solution**

This PR introduces a coercion step for the input document, transforming it into a structure that aligns with the OpenAPI schema. By doing so, even loosely structured or partially invalid inputs are normalized before validation. The result is that the document consistently passes strict OpenAPI schema checks, improving reliability and robustness.

**Checklist**


geoffgscott (approved):
Looks good. We should discuss type and schema namespaces outside of this PR

PR #5947, opened 2025-06-13
feat(api-client): implement deleteRow function for RequestBody form data

inyourtime:
**Problem**

Currently, deleting rows in the request body form-data is not yet implemented.

before:

https://github.com/user-attachments/assets/93385b82-427a-4070-be04-27fd79d5ad1c

after:

https://github.com/user-attachments/assets/ccc18063-a335-430f-a435-71818a0b3248

**Checklist**

### DIFF
--- a/.changeset/chilled-roses-tell.md
+++ b/.changeset/chilled-roses-tell.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
+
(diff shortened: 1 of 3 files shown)


marclave (approved):
awesome work @inyourtime ! I hit this bug before and it bothered me a lot, so thank you ✨

antlio:
@inyourtime great catch and work!

PR #5949, opened 2025-06-13
feat(openapi-parser): ensure unique hashes and support custom compression

DemonHa:
**Problem**

To reduce bundle size, we currently hash source URLs or file paths using SHA-1 and generate short hash strings. However, the existing implementation does not account for potential hash collisions. In rare cases, different documents may be assigned the same hash, which can result in one document unintentionally overwriting another.

> ⚠️ Note: While hash collisions are very unlikely, especially with good hash functions, our implementation uses short hash strings for compactness. This slightly increases the risk of collisions, so we're handling this edge case proactively.

**Solution**

This PR introduces a collision-handling mechanism to ensure all generated hashes are guaranteed to be unique. If a collision is detected, the system will automatically resolve it by computing a new unique hash. This guarantees that each document is safely and reliably mapped to its own hash, preventing overwrites.

Additionally, this update allows for a pluggable compression function to be used when processing external inputs. This function can be:
- a random value generator,
- a different hashing algorithm,
- or any other method to generate a unique, compact key.

This adds flexibility and future-proofs the system for alternative collision-avoidance strategies.

----

**Checklist**


amritk (approved):
looks like this should work fine. any reason for object cache instead of map? as long as theres one non digit map should come out ahead

DemonHa:
> looks like this should work fine. any reason for object cache instead of map? as long as theres one non digit map should come out ahead

@amritk The reason for that is that we wanna store the mappings on the document itself.

When we are doing partial bundle we always store the mapping on the document itself to prevent user from providing us with an external cache.

For the normal bundle of full document we are only saving the mappings if the user specifies urlMap option to true

hanspagel (approved):
amazing work!

PR #5951, opened 2025-06-13
feat(workspace-store): server store input document from different sources

DemonHa:
**Problem**

The server-side store currently only supports loading documents passed as in-memory objects. This limits flexibility, especially when working with external sources like remote URLs or local files.

**Solution**

This PR extends the server-side store to support asynchronously loading documents from:

- a URL
- a local file path

in addition to directly passing an object.

To support this, the store API has been updated to work asynchronously:
- addDocument() now returns a Promise
- the store itself is initialized in an async context

> ⚠️ Breaking Change: These changes affect any code expecting addDocument or the store setup to be synchronous. Callers will now need to await these operations.

This improvement enables greater flexibility and unlocks use cases like lazy-loading large schemas or integrating external sources without prefetching them manually.

**Checklist**

### DIFF
--- a/packages/workspace-store/src/server.ts
+++ b/packages/workspace-store/src/server.ts
@@ -14,7 +14,7 @@ export const WORKSPACE_FILE_NAME = 'scalar-workspace.json'
 type CreateServerWorkspaceStoreBase = {
   documents: {
     name: string
-    document: Record<string, unknown> | string
+    document: Record<string, unknown>
     meta?: WorkspaceDocumentMeta
   }[]
   meta?: WorkspaceMeta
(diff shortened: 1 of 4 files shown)


hanspagel (approved):
Looks good! Can we keep a README for the package up to date with a few of those features and examples? I think that would be great an great overview for everyone (and even external people).

amritk (approved):
Looks good, I didn't see the proxy fetcher function though, might wanna add it to the docs? OR is that in another PR? lol

DemonHa:
@amritk that's for the client store and it's on [this PR](https://github.com/scalar/scalar/pull/5991)

amritk:
Yea saw that lol

packages/workspace-store/src/server.ts
```diff
@@ -176,62 +184,142 @@ export function externalizePathReferences(
...
+   * 3. Externalizing references based on the workspace mode (SSR or static)
+   * 4. Adding the processed document to the workspace with its metadata
+   *
+   * The resulting document contains minimal information with externalized references
+   * that will be resolved on-demand through the workspace's get() method.
+   *
+   * @param document - The OpenAPI document to process and add
+   * @param meta - Document metadata containing the required name and optional settings
+   */
+  const addDocumentSync = (document: Record<string, unknown>, meta: { name: string } & WorkspaceDocumentMeta) => {
```
geoffgscott, line 266: For the server I don't think we ever need a synchronous addDocument function. Maybe it doesn't hurt to have though.

geoffgscott (approved):

packages/workspace-store/src/server.ts
DemonHa: we are not actually exposing this function, just for internal use only for now

PR #5957, opened 2025-06-13
fix(api-reference): client libraries style

antlio:
**Problem**

currently the client libraries style appears inconsistant + doesn't offer copy capability (e.g. in SDK usage)

**Solution**

this pr updates the style accordingly and re-introduce the copy capability.

| state | preview |
| -------|------|
| before | image |
| after | image | 

**Checklist**

### DIFF
--- a/.changeset/large-stingrays-decide.md
+++ b/.changeset/large-stingrays-decide.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
+fix: updates client libraries footer style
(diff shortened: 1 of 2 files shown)


amritk:
should we add code highlighting as well?

antlio:
> should we add code highlighting as well?

yes i was actually bringing this up too in dms lets 💅 it

marclave:
how does this look for the typical install command which is most likely what people will put.

npm install @scalar/sdk for example

antlio:
> how does this look for the typical install command which is most likely what people will put.
> 
> npm install @scalar/sdk for example

updated

marclave (approved):
awesome fix @antlio ✨

PR #5960, opened 2025-06-13
fix(api-reference): restore breadcrumb into the slot props

amritk:
**Problem**

Currently, I didn't check the other side before removing the breadcrumb from the slot props

**Solution**

With this PR we bring it back jack

**Checklist**


packages/api-reference/src/components/ApiReferenceLayout.vue
```diff
@@ -367,6 +368,7 @@ watch(hash, (newHash, oldHash) => {
     <!-- Header -->
     <div class="references-header">
       <MobileHeader
+        :breadcrumb="referenceSlotProps.breadcrumb"
```
marclave, line 371: :)

marclave (approved):

PR #5963, opened 2025-06-14
fix(workspace-store): reactivity 

amritk:
**Problem**

Currently, the workspace store is not reactive

**Solution**

With this PR we bring back what was once removed. Not sure if this is what you guys want, but it works. @DemonHa @geoffgscott 

**Checklist**


DemonHa:
For now let's bring back the reactivity to the store. In the future we might want to pull it outside the store and make the store state framework independent

PR #5969, opened 2025-06-16
test(api-reference): updates playwright snapshots

antlio:
**Solution**

this pr updates api reference playwright snapshots to match with recent default theming ui changes.

**Checklist**


hanspagel (changes requested):
It looks like we’re showing models that we shouldn’t show now (`x-scalar-ignore: true`).

antlio:
> It looks like we’re showing models that we shouldn’t show now (`x-scalar-ignore: true`). This is probably not the right place, but it seems to be ok on https://galaxy.scalar.com (so it’s ok in main?) or is https://galaxy.scalar.com outdated?

now that we released it impacted it - #5977 is fixing this

cdn has been purged - waiting for #5978 release to get @amritk fix in and update the snapshots

PR #5976, opened 2025-06-16
feat(api-reference): export references useSidebar

bgrcs:
**Problem**

Need useSidebar separately in org for now to handle some item actions

**Solution**

Export current useSidebar; it was exported previously but from hooks instead, and we got rid of that

**Checklist**


amritk:
You should be able to find it here! We are moving away from exporting from the base!

```ts
import { useSidebar } from '@scalar/api-reference/features/sidebar'
```

PR #5981, opened 2025-06-17
fix(workspace-store): relative paths for ssg mode

DemonHa:
With this PR we correctly reference the chunks

**Checklist**


hanspagel (approved):
ok!

amritk:
OOps I saw the approvals @hanspagel missed ur comment

>the tests passed before, can we add a test that would fail without the fix in this PR?

Maybe you can add this later @DemonHa

PR #5986, opened 2025-06-17
style: only quote properties as needed in biome

amritk:
**Problem**

Currently, when you paste in json, I have been manually removing quotes like a sucker.

Before:
```ts
  'javascript': {
    'formatter': {
      'quoteProperties': 'asNeeded',
    }
  },
```

After:
```ts
  javascript: {
    formatter: {
      quoteProperties: asNeeded,
    }
  },
```
**Solution**

With this PR we do what we had before with eslint/prettier and properties are only quoted as needed ie `'/paths'`. I will run this lint on the weekend to as not to  be distruptive

**Checklist**

### DIFF
--- a/biome.json
+++ b/biome.json
@@ -65,7 +65,7 @@
     "formatter": {
       "quoteStyle": "single",
       "trailingCommas": "all",
-      "quoteProperties": "preserve",
+      "quoteProperties": "asNeeded",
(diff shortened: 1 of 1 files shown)


amritk:
Actually I'll just fold this into the biome 2 upgrade

PR #5989, opened 2025-06-17
fix(api-reference): show content type description in schema property

amritk:
**Problem**

Currently, we don't use the content encoding from version 3.1 of the spec to show descriptions, it used to be called format in 3.0.

https://spec.openapis.org/oas/latest.html#migrating-binary-descriptions-from-oas-3-0

**Solution**

With this PR we fix this and show the content type.

@antlio so I removed the code which checks for additional properties. Not sure if it was meant to match all additional properties but either way I don't think that is safe. What was that part specifically fixing?

Then I added the content encoding to be displayed so it matches the old style.

**Checklist**

### DIFF
--- a/.changeset/sour-mirrors-move.md
+++ b/.changeset/sour-mirrors-move.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
+fix: show content encoding description in schema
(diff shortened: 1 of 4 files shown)


antlio:
it was retrieving the schema name for some edge case context (i'll retrieve some spec), but it should be tested! anyway we should make that name retrieval easier from the parser.

amritk:
Ya also for all these kind of cases we need a test!

PR #5990, opened 2025-06-18
fix(api-reference): complex auth in client store at the operation level

### LINKED ISSUE #5926: Scalar only selects the first of multiple API keys
### What happens?

After selecting both authentication methods, `endpoint1` only has `publicApiKey` and `endpoint2` only has `privateApiKey`.

Image

Both endpoints have both API keys. This is what should happen according to [the documentation](https://swagger.io/docs/specification/v3_0/authentication/api-keys/#multiple-api-keys)

### OpenAPI Document

https://sandbox.scalar.com/e/okPIm

amritk:
**Problem**

closes #5926

Currently, complex auth is not working at the operation level.

**Solution**

This is my second PR today where I'm just deleting some code to make it work 🤷🏾 

**Checklist**

### DIFF
--- a/packages/oas-utils/src/transforms/import-spec.test.ts
+++ b/packages/oas-utils/src/transforms/import-spec.test.ts
@@ -829,6 +829,30 @@ describe('importSpecToWorkspace', () => {
       expect(res.collection.selectedSecuritySchemeUids).toEqual(selectedSecuritySchemeUids)
     })
 
+    it('handles complex security on the operation level', async () => {
+      const specWithComplexSecurity = {
+        ...galaxy,
+        security: [{ apiKeyHeader: [], basicAuth: [] }],
+        paths: {
+          '/test': {
+            get: {
+              security: [{ apiKeyHeader: [], basicAuth: [] }],
+            },
+          },
+        },
+      }
+
+      const res = await importSpecToWorkspace(specWithComplexSecurity)
+      if (res.error) {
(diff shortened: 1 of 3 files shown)


packages/oas-utils/src/transforms/import-spec.ts
```diff
@@ -410,25 +410,6 @@ export async function importSpecToWorkspace(
         delete requestPayload.examples
       }
 
-      // Add list of UIDs to associate security schemes
```
amritk, line 413: as far as I can see, all this code seems to do is break complex auth

hanspagel: lol

hanspagel (approved):
I don’t know how to test this, but the changes make sense to me and you’ve got a a test. So let’s ship it?

packages/oas-utils/src/transforms/import-spec.test.ts
```diff
@@ -829,6 +829,30 @@ describe('importSpecToWorkspace', () => {
       expect(res.collection.selectedSecuritySchemeUids).toEqual(selectedSecuritySchemeUids)
     })
 
+    it('handles complex security on the operation level', async () => {
+      const specWithComplexSecurity = {
```
hanspagel, line 833: never call it spec pls 😁 it’s a document or a definition

PR #5991, opened 2025-06-18
feat(api-reference): load documents on the new store

DemonHa:
**Solution**

With this PR we mirror and keep in sync the old state and the new store

**Checklist**


geoffgscott (approved):
Looks good. @hanspagel Do we want to keep the active state in an object or flat at the top level?

geoffgscott:
Just needs a changeset @DemonHa

amritk (changes requested):
Looks like we are loading all documents here. We only want to load the currently selected document

DemonHa:
> Looks like we are loading all documents here. We only want to load the currently selected document

@geoffgscott Do we wanna preload all documents or only the selected one and we load the others when we change the selection?

hanspagel:
@DemonHa my vote for

> only the selected one and we load the others when we change the selection

marclave:
@DemonHa load currently selected :)

PR #5996, opened 2025-06-18
fix(oas-utils): doesn’t create example data for items without a type

hanspagel:
**Problem**

If a schema has an array, where the item has `properties` or `items`, but no `type: 'array|object'`, we don’t generate an example.

**Solution**

We can be clever and assume it’s supposed to be an object if it has `properties` and it’s supposed to be an array if it has `items`.

**Checklist**


hanspagel:
this function is well tested, so how bad can it be? let’s get this fix in before we release

PR #6004, opened 2025-06-18
fix(api-reference): reduce incorrect schema names

### LINKED ISSUE #6001: referenced schema shows wrong type
ReferencesUsers shows "FOOBAR" as the type, should show "string"

https://sandbox.scalar.com/e/Drucf

Image

amritk:
**Problem**

closes #6001

Currently, we are matching the schema name based on a matching type.

**Solution**

That is unsafe as we can easily have multiple schemas with different names and the same types, specially for non-object or arrays. Previously it was matching strings which are quite common. Now it will only match if all array or object properties match which is a bit safer but not perfect.

## Update

Alright so after some digging I realize what @antlio meant. Once we de-reference we lose the name of the schema. I just made it a little safer by only property matching on objects or arrays, this should be slightly less error prone.

We should come back at this one in the new workspace store and keep some reference to the name @DemonHa as long as we have access to the ref string we should be good!

**Checklist**

### DIFF
--- a/.changeset/lovely-ducks-explain.md
+++ b/.changeset/lovely-ducks-explain.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
+fix: display correct schema type instead of false name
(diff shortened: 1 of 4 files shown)


DemonHa:
@amritk With the new store it should be easy to preserve information for the $ref

PR #6007, opened 2025-06-19
fix(api-reference): looks up wrong name for schemas

hanspagel:
**Problem**

shows `array description[]`

Screenshot 2025-06-19 at 15 16 05

**Solution**

don’t just blindly compare schemas to guess a name, for simple schemas we might retrieve the wrong name

`array string[]`

Screenshot 2025-06-19 at 15 14 11

**Checklist**

### DIFF
--- a/packages/api-reference/src/components/Content/Schema/SchemaComposition.test.ts
+++ b/packages/api-reference/src/components/Content/Schema/SchemaComposition.test.ts
@@ -85,7 +85,9 @@ describe('SchemaComposition', () => {
       expect(tab.text()).toBe('object')
     })
 
-    it('displays schema name from components when matching schema is found', () => {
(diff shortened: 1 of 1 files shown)


packages/api-reference/src/components/Content/Schema/SchemaComposition.test.ts
```diff
@@ -85,7 +85,9 @@ describe('SchemaComposition', () => {
       expect(tab.text()).toBe('object')
     })
 
-    it('displays schema name from components when matching schema is found', () => {
+    // TODO: This would be nice to have, but we used to compare schemas and in some cases it returned the wrong name.
+    // Let’s find another approach and enable this test again.
+    it.todo('displays schema name from components when matching schema is found', () => {
```
antlio, line 90: IMO we shouldn't match at all and pass the name at the parsing/store level

PR #6012, opened 2025-06-19
feat(workspace-store): preserve the original ref information

DemonHa:
**Problem**

When using the `magic-proxy` to resolve internal references in the workspace store, we currently lose track of the original reference path. This makes it difficult to trace back where a resolved value came from.

**Solution**
This PR updates the proxy mechanism to preserve the original reference path. When accessing a resolved ref, the original ref information is now retained and accessible, allowing for better traceability.

**Checklist**

### DIFF
--- a/.changeset/moody-carrots-fly.md
+++ b/.changeset/moody-carrots-fly.md
@@ -1,0 +1,5 @@
+---
+'@scalar/workspace-store': minor
+---
+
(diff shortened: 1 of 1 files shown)


antlio:
great work!

amritk (approved):
This iz the one

PR #6019, opened 2025-06-20
fix(helpers): include src to tsconfig (closes #5936)

### LINKED ISSUE #5936: @scalar/api-reference v1.31.4 -> v1.31.6 introduced a breaking change
### What happens?

I believe there's a bug here, not sure why this was changed:

<https://github.com/scalar/scalar/commit/8bc9f20#diff-1aaf604e4d2bcca8c3d30f2f6d666ff22719e827734c9747a2233ef1ed93d49cL2>

`  "include": ["src"],` was removed for some reason (?) and this seems to break the standalone dist browser build

specifically you can't use `unassert` on this package anymore as it has an acorn AST parsed related error

not sure if this is the exact issue, but i'm curious why that line was removed as it seems unrelated to the commit?

changelog: <https://github.com/scalar/scalar/blob/main/packages/api-reference/CHANGELOG.md#1315>

No errors should throw when you run unassert or gulp-unassert on the dist standalone build

### OpenAPI Document

_No response_

titanism:
<https://github.com/scalar/scalar/issues/5936>

<https://github.com/titanism/gulp-unassert-reproduction>

### DIFF
--- a/packages/helpers/tsconfig.json
+++ b/packages/helpers/tsconfig.json
@@ -1,5 +1,6 @@
 {
   "extends": "../../tsconfig.json",
+  "include": ["src"],
   "compilerOptions": {
     "paths": {
       "@/*": ["./src/*"]


hanspagel:
Thanks for the PR! I think we’ll have another solution, let me check.

DemonHa:
Closing this PR as not related to the error.

As mentioned before, that change does not affect the build in any way. We are building using `tsconfig.build.json` file which adds the `src` field anyways.

Going to add a more detailed explanation on the issue

titanism:
This was a breaking change and should not have been a patch nor minor bump, it should have been a major.  Thank you.

PR #6021, opened 2025-06-20
refactor(api-reference): remove transformed operation from operation components

hanspagel:
**Problem**

Currently, we still use the deprecated `TransformedOperation` in all the operation components. We’d prefer to stick to the structure of the OpenAPI 3.1 specification, though.

**Solution**

With this PR we’re refactoring the relevant components to work with just a regular OpenAPI 3.1 Operation Object.

**Checklist**


marclave (approved):
nice work @hanspagel :)

PR #6023, opened 2025-06-20
chore: update some packages in the catalog (yaml, @types/node, ansis …)

hanspagel:
This updates a few of the packages in the catalog:

- @google-cloud/storage
- @types/express
- @types/node
- ansis
- semver
- type-fest
- yaml


hanspagel:
I’m not too afraid of those updates, it all looks to seem good, all tests pass, all type checks pass … Let’s get it in before the next release.

PR #6025, opened 2025-06-20
fix(nuxt): set config content after grabbing spec

### LINKED ISSUE #5891: nuxt: couple of issues since the flattens layouts update of api-reference
### What happens?

I've encountered a couple of issues after this update: https://github.com/scalar/scalar/pull/5847

1. The first problem I found is that with this latest update, if I use the `url` config property, it gets fetched twice (server and client). Before the update, it was working well, the nuxt module was fetching it on the server side only.
```js
scalar: {
  url: "/myopenapi.json"
}
```

2. ~The second problem is that if I use the `content` config propery, most of the paths appear duplicated in the API reference.
BTW: this doesn't happens with `yaml` format.~

Thanks in advance for the hard work!

1. Using the `url` config prop should fetch only server side.

2. ~Using the `content` config prop should not duplicate the api paths~

### OpenAPI Document

_No response_

ahmedrangel:
**Problem**

Resolves #5891

**Solution**

Maybe this could be the best way to prevent the `ApiReferenceLayout` component from fetching again on the client. Since we are getting the specs from the server side we will assign the grabbed spec to the config content.

**Checklist**

### DIFF
--- a/.changeset/lovely-candles-beg.md
+++ b/.changeset/lovely-candles-beg.md
@@ -1,0 +1,5 @@
+---
+'@scalar/nuxt': patch
+---
+
+fix(nuxt): set config content after grabbing spec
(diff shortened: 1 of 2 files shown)


marclave:
awesome work @ahmedrangel ! let's get this in ✨

PR #6026, opened 2025-06-20
refactor(api-reference): move code to features/ folder

hanspagel:
**Problem**

We want to encapsulate our code more, so we decided to move code to features/ folder where they have their own dependencies, versus having their dependencies organized just by type (components, helpers, hooks) on a global level.

**Solution**

This PR moves more code to the features/ folders:

* multiple-documents
* sidebar
* specification-extension

And I’ve seen the recently created folders where using kebab-case, so I renamed most features to kebab-case.

Issues might appear on developers machines if I just make something lowercase Search -> search, so I avoided that for now.

**Checklist**


hanspagel:
Just moving files around, how bad can it be? And if we keep it open for longer, there’ll be just more and more conflicts. Let’s merge it. :)

PR #6031, opened 2025-06-20
fix(api-reference): fix up search and document selector on classic layout

hwkr:
Prevents inconsistent external line heights from affecting the search button layout and set the width to full for the references search button export.

Normally it's preferred to use the parent to manage layout (e.g. flexbox, grid) but this maintains the current behaviour.

Also:

* Removes the spacing around the document selector on the classic layout
* Manages the width of the search on classic layout
* Makes sure classes are being passed the the search on the classic layout

**Checklist**

### DIFF
--- a/packages/api-reference/src/components/ApiReferenceLayout.vue
+++ b/packages/api-reference/src/components/ApiReferenceLayout.vue
@@ -458,12 +458,12 @@ watch(hash, (newHash, oldHash) => {
               <ClassicHeader v-if="configuration.layout === 'classic'">
                 <div
                   v-if="$slots['document-selector']"
-                  class="w-64 empty:hidden">
+                  class="w-64 *:!p-0 empty:hidden">
                   <slot name="document-selector" />
                 </div>
                 <SearchButton
                   v-if="!configuration.hideSearch"
-                  class="t-doc__sidebar"
+                  class="t-doc__sidebar max-w-64"
                   :searchHotKey="configuration.searchHotKey"
                   :spec="parsedDocument" />
                 <template #dark-mode-toggle>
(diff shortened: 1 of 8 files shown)


packages/api-reference/src/components/ApiReferenceLayout.vue
```diff
@@ -458,12 +458,12 @@ watch(hash, (newHash, oldHash) => {
               <ClassicHeader v-if="configuration.layout === 'classic'">
                 <div
                   v-if="$slots['document-selector']"
-                  class="w-64 empty:hidden">
+                  class="w-64 *:!p-0 empty:hidden">
```
hwkr, line 461: Ideally we wouldn't set this here and set it in the document selector instead but this is how we use it in org.

packages/api-reference/src/features/Search/SearchButton.vue
```diff
@@ -58,6 +58,8 @@ function handleClick() {
 <template>
   <ScalarSidebarSearchButton
     ref="button"
+    class="w-full"
+    :class="$attrs.class"
```
hwkr, line 62: I don't know why Vue class inheritance isn't working here but 🤷‍♀️

packages/components/src/components/ScalarListbox/ScalarListbox.vue
```diff
@@ -65,8 +65,7 @@ const { cx } = useBindCx()
       :placement="placement ?? 'bottom-start'">
       <ListboxButton
         :id="id"
-        as="template"
-        class="justify-start focus:outline-none focus-visible:ring-1 focus-visible:ring-c-accent">
```
hwkr, line 69: This probably shouldn't have ever been here

packages/api-reference/src/features/Search/SearchButton.vue
amritk: Probably because there's two components and the root and it doesn't know which one it goes to

amritk (approved):

PR #6032, opened 2025-06-21
fix(components): janky scroll on combobox autofocus

amritk:
**Problem**

Currently, when you have long list of combobox options, it does a janky scroll on open.

**Solution**

With this PR we toss a timeout on the autofucus and it seems to fix the issue. @hwkr this is all yours now.

**Checklist**

### DIFF
--- a/packages/components/src/components/ScalarCombobox/ScalarComboboxOptions.vue
+++ b/packages/components/src/components/ScalarCombobox/ScalarComboboxOptions.vue
@@ -128,6 +128,14 @@ function moveActive(dir: 1 | -1) {
     block: 'nearest',
   })
 }
+
+// Manual autofocus for the input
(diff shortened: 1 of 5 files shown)


hwkr:
Added some tests and tweaked the logic to use a `nextTick` :)

hwkr (approved):
Should be good to go!

packages/components/src/components/ScalarCombobox/ScalarComboboxOptions.vue
```diff
@@ -130,12 +138,8 @@ function moveActive(dir: 1 | -1) {
 }
 
 // Manual autofocus for the input
-const input = ref<HTMLInputElement | null>(null)
```
amritk: @hwkr we needed the timeout here actually, its janky right now. bringing it back!

PR #6033, opened 2025-06-21
feat(oas-utils): get operation to HAR helper working with new types

amritk:
**Problem**

Currently, operation to har was all theoretical

**Solution**

With this PR I started using it in the request example code block and fixed some things while adding a lot more tests. 

Also exported the types from workspace store so they can now be consumed.

**Checklist**

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1973,6 +1973,9 @@ importers:
       '@scalar/types':
         specifier: workspace:*
         version: link:../types
+      '@scalar/workspace-store':
+        specifier: workspace:*
+        version: link:../workspace-store
       '@types/har-format':
         specifier: ^1.2.15
         version: 1.2.15


packages/workspace-store/package.json
```diff
@@ -43,6 +43,16 @@
       "types": "./dist/schemas.d.ts",
       "default": "./dist/schemas.js"
     },
+    "./schemas/v3.1/*": {
```
amritk, line 46: Not sure if this is where we want to export, but I just followed the folder structure

marclave (approved):

hanspagel (approved):
cool, looks good

packages/workspace-store/src/schemas/v3.1/strict/oauth-flow.ts
```diff
@@ -2,19 +2,74 @@ import { compose } from '@/schemas/v3.1/compose'
...
+const OAuthFlowCommonSchema = Type.Object({
+  /** The URL to be used for obtaining refresh tokens. This MUST be in the form of a URL. The OAuth2 standard requires the use of TLS. */
+  refreshUrl: Type.Optional(Type.String()),
+  /** REQUIRED. The available scopes for the OAuth2 security scheme. A map between the scope name and a short description for it. The map MAY be empty. */
+  scopes: Type.Record(Type.String(), Type.String()),
+})
+
+/** Configuration for the OAuth Implicit flow */
+export const OAuthFlowImplicitSchema = compose(
+  Type.Intersect([
```
DemonHa, line 15: You can remove the Type.Intersect here and just use the compose.

packages/workspace-store/esbuild.ts
```diff
@@ -1,6 +1,12 @@
 import { build } from '@scalar/build-tooling/esbuild'
 
-const entries = ['./src/schemas.ts', './src/client.ts', './src/server.ts']
+const entries = [
+  './src/schemas.ts',
+  './src/client.ts',
+  './src/server.ts',
+  './src/schemas/v3.1/*.ts',
```
DemonHa, line 7: I think we can export the guard function and type in the schemas.ts entry point

Issue #6038, opened 2025-06-22
Model names in navigation ignore schema.Title from OpenAPI and display raw C# type names


zbrkic:
### What happens?

Despite setting `schema.Title` via `AddSchemaTransformer` in .NET OpenAPI generation, Scalar UI shows the raw C# class names (e.g., `ListDto`, `ItemDto`, `EntityDto`) in the left-hand navigation and model link references.

However, under the *Models* section, the correct titles (`List`, `Item`, `Entity`) are displayed — which confirms that `schema.Title` is correctly set in the OpenAPI output.

This results in inconsistency and confusion for consumers of the API documentation, especially when DTO naming conventions are stripped intentionally in the spec for clarity.

📷 Screenshot attached (highlighted):
- Left nav: `ListDto`, `ItemDto`, `EntityDto`
- Models below: `List`, `Item`, `Entity` (correct)

Minimal example:

```csharp
options.AddSchemaTransformer((schema, context, cancellationToken) => {
    var name = context.JsonTypeInfo.Type.Name;
    if (name.EndsWith("Dto"))
        schema.Title = name[..^3];
    return Task.CompletedTask;
});
```

![Image](https://github.com/user-attachments/assets/e01c4cf0-b1ab-49f9-b54f-8807d7ac2652)

### What did you expect to happen?

The UI should consistently use schema.Title for all model name references, especially in:

- Left-hand model navigation panel

- Model hyperlinks

- Breadcrumbs or references

### OpenAPI Document

https://sandbox.scalar.com/e/IuhlU

PR #6041, opened 2025-06-23
feat(api-reference): use schema title for the sidebar (if available)

hanspagel:
**Problem**

We list the schemas in the sidebar, but we always use the key and ignore the title (if the schema has one):

https://json-schema.org/draft/2020-12/json-schema-core#section-4.3.5

**Solution**

With this PR we’re using the schema title (if it’s available) and fall back to the key of the schema.

Closes https://github.com/scalar/scalar/issues/6038

**Checklist**

### DIFF
--- a/.changeset/famous-carrots-remember.md
+++ b/.changeset/famous-carrots-remember.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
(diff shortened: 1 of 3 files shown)


xC0dex:
Where have I seen this before? 👀

xC0dex (approved):

PR #6047, opened 2025-06-23
feat(api-reference): modify request before it goes out

hanspagel:
**Problem**

Sometimes, you want to modify the outgoing request from the API client, for example to modify the headers or inspect the request.

**Solution**

With this PR we’re adding a `onBeforeRequest` hook to the @scalar/api-reference configuration to modify the request.

It’s using the existing API Client Plugin API to achieve this.

```js
createApiReference('#app', {
  url: 'https://cdn.jsdelivr.net/npm/@scalar/galaxy/dist/latest.json'',
  onBeforeRequest: ({ request }) => {
    // Add a custom header to all requests
    request.headers.set('X-Custom-Header', 'test')
  }
})
```

**Checklist**

### DIFF
--- a/packages/api-client/src/libs/send-request/create-request-operation.test.ts
+++ b/packages/api-client/src/libs/send-request/create-request-operation.test.ts
@@ -1203,23 +1203,40 @@ describe('create-request-operation', () => {
   })
 
   it('does not execute onBeforeRequest hook when plugin manager is not provided', async () => {
-    const [error, requestOperation] = createRequestOperation(
-      createRequestPayload({
-        serverPayload: { url: 'https://api.example.com' },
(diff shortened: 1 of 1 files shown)


Ma-ve:
Great work! 🙏 

I was looking for something like this; but specifically for `onReceivedResponse`: I saw that plugins had some way of hooking into before request / after response, but am I correct in assuming this PR makes the hook available in the main configuration object? Any plans on adding a `onResponse` too?

hanspagel:
@Ma-ve Yes, we’ll definitely add this, too!

marclave (approved):
a much needed feature! nice work @hanspagel !!!

PR #6048, opened 2025-06-24
feat(workspace-store): add configuration mapping to the workspace store

DemonHa:
**Problem**
There is currently no mechanism to consume configuration values from the workspace store, making it difficult to dynamically access and apply user-defined or document-scoped configuration.

**Solution**
This PR introduces the ability to consume document configuration.

Document configuration is resolved in the following order:
1. document level configuration
2. workspace level configuration
3. default configuration values

**Checklist**


packages/workspace-store/src/client.ts
```diff
@@ -9,14 +9,33 @@ import { extensions } from '@/schemas/extensions'
...
+/**
+ * Configuration object for workspace documents.
+ */
+type Config = {
+  'x-scalar-reference-config'?: ReferenceConfig
+}
+
+const defaultConfig: DeepTransform<Config, 'NonNullable'> = {
+  'x-scalar-reference-config': defaultReferenceConfig,
+}
```
hanspagel, line 63: let’s stay closer to the package name, maybe this:

```suggestion
type Config = {
  'x-scalar-api-reference'?: ReferenceConfig
}

const defaultConfig: DeepTransform<Config, 'NonNullable'> = {
  'x-scalar-api-reference': defaultApiReferenceConfiguration,
}
```

packages/workspace-store/src/client.test.ts
```diff
@@ -560,4 +561,74 @@ describe('create-workspace-store', () => {
       ],
     })
   })
+
+  it('should correctly get the config #1', () => {
```
hanspagel: nitpick alert: any chance you can set up your AI to make the test names a bit shorter? it’s just a bit nicer to read in the console output.

    ✅ it('generates a slug from the title')
    ❌ it('should generate a slug from the title')

https://github.com/scalar/scalar/blob/main/.cursor/rules/tests.mdc#L26-L27

packages/workspace-store/src/client.ts
geoffgscott: This really needs to have the word config in it for clarity. we could the word API in there if we want but I don't think it adds any value.

marclave (approved):

PR #6057, opened 2025-06-25
feat(workspace-store): add workspace specification schema

DemonHa:
Create workspace specification layer as a layer above openAPI to define a scalar workspace

**Checklist**


packages/workspace-store/src/schemas/workspace-specification/config.ts
```diff
@@ -0,0 +1,10 @@
+import { ReferenceConfigSchema } from '@/schemas/reference-config'
+import { Type, type Static } from '@sinclair/typebox'
+
+export const ConfigSchema = Type.Partial(
+  Type.Object({
+    'x-scalar-reference-config': ReferenceConfigSchema,
```
hanspagel, line 6:
```suggestion
    'x-scalar-api-reference': ApiReferenceConfigurationSchema,
```

can we stick closer to the package name?

packages/workspace-store/src/schemas/workspace-specification/index.ts
```diff
@@ -0,0 +1,26 @@
+import { compose } from '@/schemas/v3.1/compose'
+import { OpenAPIDocumentSchema } from '@/schemas/v3.1/strict/openapi-document'
+import { WorkspaceMetaSchema } from '@/schemas/workspace'
+import { ConfigSchema } from '@/schemas/workspace-specification/config'
+import { InfoSchema } from '@/schemas/workspace-specification/info'
+import { Type, type Static } from '@sinclair/typebox'
+
+export const WorkspaceSpecificationSchema = compose(
+  Type.Object({
+    workspace: Type.Union([Type.Literal('1.0.0')]),
```
hanspagel:
```suggestion
    workspace: Type.Union([Type.Literal('draft')]),
```

maybe we start with `draft` as a version? just a cosmetic change

packages/workspace-store/src/schemas/workspace-specification/config.ts
geoffgscott: I feel very strongly this needs to be associated with the word `config`. If you really want you could do `x-scalar-api-reference-config` but I don't see a reason to be more verbose.

marclave: also agree it should include config in it :)

hanspagel: I thought it’s just everything related to the package (not read by anything else). Personally, I don’t see a reason for it to contain the word `config`, but anyway:

What about this then:

```suggestion
    'x-scalar-api-reference-config': ApiReferenceConfigSchema,
```

DemonHa:
@geoffgscott do we wanna commit `packages/workspace-store/src/schemas/workspace-specification/workspace-specification.json`?

packages/workspace-store/src/schemas/workspace-specification/workspace-specification.json
geoffgscott, line 1: Yeah let's not commit this. Eventually we can just put it into the registry.

packages/workspace-store/src/schemas/workspace-specification/config.ts
geoffgscott: @marclave and I aligned on `x-scalar-reference-config` and `x-scalar-client-config`

geoffgscott (approved):

PR #6058, opened 2025-06-25
feat(workspace-store): create extensions schemas for the api-client

DemonHa:
Add schema and types for the api-client extension keys

**Checklist**


packages/workspace-store/src/schemas/v3.1/strict/openapi-document.ts
```diff
@@ -29,6 +31,11 @@ const OpenApiExtensionsSchema = Type.Partial(
         TagObjectSchema,
       ),
     ),
+    'x-scalar-active-environment': Type.String(),
+    /** A custom icon representing the collection */
+    'x-scalar-icon': Type.String(),
+    'x-scalar-environments': xScalarEnvironmentsSchema,
+    'x-scalar-secrets': xScalarSecretsSchema,
```
hanspagel: is it a good moment to move them to a `x-scalar-api-client` prefixed object? or should we do this later?

marclave (approved):
wicked :)

PR #6060, opened 2025-06-25
fix(api-reference): `withDefaults` is used with destructured props

hanspagel:
**Problem**

```bash
│ [@vue/compiler-sfc] withDefaults() is unnecessary when using destructure with defineProps().
│ Reactive destructure will be disabled when using withDefaults().
│ Prefer using destructure default values, e.g. const { foo = 1 } = defineProps(...).
│
│ /Users/hanspagel/Documents/Projects/scalar/packages/api-reference/src/components/Content/Lazy/Loading.vue
│ 42 |   * - need to find an event for codemirror loaded, currently using timeout for models
│ 43 |   */
│ 44 |  const { document, collection, server, layout, parsedSpec } = withDefaults(
│    |                                                               ^^^^^^^^^^^^
│ 45 |    defineProps<{
│ 46 |      document: OpenAPIV3_1.Document
```

**Solution**

Do what the warning says.

**Checklist**

### DIFF
--- a/.changeset/cool-humans-taste.md
+++ b/.changeset/cool-humans-taste.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
+fix: withDefaults is used with destructured props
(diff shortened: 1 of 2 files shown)


hanspagel:
Just removing a warning, let’s get it in!

PR #6063, opened 2025-06-25
refactor: use more common straight apostrophe ' instead of the real apostrophe ’

hanspagel:
**Problem**

I’ve use the real apostrophe ~600 times in the code base already, but it’s annoying.

* LLMs replace them with straight apostrophes every time you touch a file, resulting in confusing changesets
* It’s highlighted as an invalid character in most editors
* Noone knows how to type the real apostrophes

**Solution**

I’ve replaced all apostrophes with their straight alternative:

# ’ → '

See: https://en.wikipedia.org/wiki/%27_(disambiguation)

A sad day for someone into micro typography, but a great day for everyone else.

**Checklist**


hanspagel:
Let’s get this in before we run into a thousand conflicts.

PR #6064, opened 2025-06-25
fix(api-reference): schema does not render without type attribute

hanspagel:
**Problem**

We recently fixed the example generation for schemas without a type attribute in #5996.

We’re able to guess the type if the schema has an `items` (array) or `properties` attribute (object).

**Solution**

With this PR, we’re doing the same in the schema components.

**Before**

See how the example payload has a nested structure and the body on the left doesn't show much.

Screenshot 2025-06-25 at 13 31 26
 
**After**

Screenshot 2025-06-25 at 13 31 49

**Checklist**


hanspagel:
This is just better, let’s get it in. :)

PR #6067, opened 2025-06-25
ci: generate code coverage report and upload to codecov

hanspagel:
**Problem**

We write more and more tests, but we want to have everything covered eventually.

**Solution**

I’m not the biggest fan of code coverage reports, but it might be a nice motivation and reminder to write tests, so let’s try it.

**Checklist**


hanspagel:
It shows most of the code is covered, I think because of the ApiReferenceLayout test, that basically runs everything. Not super helpful.

PR #6071, opened 2025-06-25
feat(api-client): use `documentUrl` as a fallback server

hanspagel:
**Problem**

When we import documents that don’t have a server, we just fallback to the URL of the API client (which will never be the actual API, because … it’s the API client).

Example: https://api.theirstack.com/

**Solution**

With this PR we’re using the `documentUrl` as a fallback, so if something was imported from https://example.com without a `servers` array, we just assume the API is on https://example.com (not on the URL of the API client).

I've also renamed the import helpers (kebab-case) and added tests.

**Checklist**

### DIFF
--- a/.changeset/smooth-books-exist.md
+++ b/.changeset/smooth-books-exist.md
@@ -1,0 +1,6 @@
+---
+'@scalar/api-client': patch
+'@scalar/oas-utils': patch
+---
+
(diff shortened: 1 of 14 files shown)


marclave (approved):
AWESOMEEEEE

PR #6075, opened 2025-06-25
fix(api-references): duplicate section ID for tags on page load

DemonHa:
**Problem**
The tags section is rendered on initial page load using the same section ID as the main page content.

This causes potential conflicts or confusion when targeting or navigating to specific sections of the page, as both the page and the tags section share the same ID.

**Solution**

With this PR we give a unique id for the tag section so when the page finish loading we scroll to the correct section of the page

**Checklist**

### DIFF
--- a/.changeset/pretty-carrots-cheer.md
+++ b/.changeset/pretty-carrots-cheer.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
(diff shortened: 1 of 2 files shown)


marclave (commented):
👁️  👄 👁️ 
   🦵 🦵

marclave (approved):

PR #6083, opened 2025-06-26
feat(workspace-store): add support for managing original document state with download and save APIs

DemonHa:
**Problem**

Currently, all updates are written directly to the document, making it impossible to track unsaved changes or revert to the original state. This limits our ability to discard edits or re-download the original version of the document.

**Solution**

This PR introduces a new API to manage the original document state separately from unsaved changes:

- download: Retrieves the original, unmodified version of the document.
- save: Commits the current changes back to the original document.

> Note: When saving, changes to external references or any non-core extension keys are intentionally discarded to preserve the purity of the original document.

**Checklist**


hanspagel (commented):
maybe we call it `.export()` (instead of download), because it’s not actually triggering a file download or something

DemonHa:
> maybe we call it `.export()` (instead of download), because it’s not actually triggering a file download or something

@hanspagel 

We may need to choose a clearer name for this method, as the current options could be confusing for end users.
Since we’re exposing this on the workspace store, using `download` or `export` might misleadingly suggest that the entire workspace will be downloaded. However, the actual intent is to download only the active original document — meaning the initial input, exactly as it was provided, with no $ref resolution or other processing.

Let’s make sure the name clearly communicates that it returns the unaltered source, not the workspace as a whole.

cc @geoffgscott

hanspagel:
What about `workspace.getDocument()` (and optional: you can pass a document identifier if you want something else than the active document)?

marclave (approved):
AWESOME @DemonHa !!!!

packages/workspace-store/README.md
```diff
@@ -283,4 +283,74 @@ To get the active document configuration you can use config getter\
...
+##### Export Document
+
+Export the specified document in JSON or YAML format:
+
+```ts
+// Export the specified document as JSON
+const jsonString = store.exportDocument('documentName', 'json')
+
+// Export the specified document as YAML
+const yamlString = store.exportDocument('documentName', 'yaml')
```
hanspagel, line 301: needs an update I think

DemonHa: that's the updated version

PR #6086, opened 2025-06-27
fix(snippetz): shell/curl does not add quotes for URLs with special character `$`

mvt147:
**Problem**

Currently, URL's with a `$` (dollar sign) special character are not being handled correctly by the curl shell package (i.e. not single quoting the URL) which results in the curl command not executing as expected when run in a terminal/shell (as it will interpret it as a variable). 

For further context on how this was discovered: I am implementing OpenAPI documentation (with Scalar) for a specification that uses `$` characters in their routes, e.g. [`/CodeSystem/$lookup`](https://www.hl7.org/fhir/R5/codesystem-operation-lookup.html). Example here https://sandbox.scalar.com/e/lCjAs

**Solution**

With this PR I have added the `$` to the special characters which are used to determine whether the URL needs to be single quoted (which resolves the issue when the curl command is run in the terminal).

**Checklist**

### DIFF
--- a/packages/snippetz/src/plugins/shell/curl/curl.test.ts
+++ b/packages/snippetz/src/plugins/shell/curl/curl.test.ts
@@ -482,4 +482,44 @@ describe('shellCurl', () => {
   "simple": "value"
 }'`)
   })
+
+  it('handles URLs with dollar sign characters', () => {
+    const result = shellCurl.generate({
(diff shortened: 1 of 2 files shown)


hanspagel (changes requested):
omg, yes! great PR, we just need to add a test.

hanspagel (approved):

PR #6087, opened 2025-06-27
feat(api-reference): update font styles across scalar

cameronrohani:
A whole bunch of very minor changes
Updated font sizes across from 13px -> 14px in:
- Api Client across the whole entire app
- API Ref in cards
- In Search
- Made card styles more consistent 
- Made active / inactive states on tabs more consistent 

More notably I did a slight refresh of Search beyond changing font sizes (ty @hwkr for the help :) )
image

### DIFF
--- a/.changeset/unlucky-pumas-walk.md
+++ b/.changeset/unlucky-pumas-walk.md
@@ -1,0 +1,9 @@
+---
+'@scalar/api-reference': patch
+'@scalar/api-client': patch
+'@scalar/components': patch
+'@scalar/scripts': patch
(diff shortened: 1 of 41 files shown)


antlio:
@cameronrohani potential fast follow nitpicks:

| before | after |
| -------|------|
| image | image|
| input height breaks width padding consistency + placeholder misalignment | decreased input height for width consistency + increased gap for placeholder alignment with element | 

also the focus seems broken (not introduced by that PR but to be looked into after)

| after |
|------|
| image |

PR #6090, opened 2025-06-27
fix(api-reference): search tests are failing

hanspagel:
**Problem**

Oops, we merged two search PRs without rebasing them, resulting in failing tests in main.

**Solution**

This PR fixes the tests. :)

**Checklist**


hanspagel:
We want main to be green, so I’ll skip the review queue here.

PR #6093, opened 2025-06-27
fix(api-reference): search modal enter

antlio:
**Problem**

currently on `enter` in the first search modal result, we are not redirecting to the element nor anything happen.

**Solution**

this pr favors the `isDefined` helper usage on enter action which is checking for null or undefined instead to ensure that 0 (index) passes the check instead of being returned when checking value.

**Checklist**


hanspagel (approved):
yes!

PR #6094, opened 2025-06-27
refactor(api-reference): schema types

hanspagel:
**Problem**

Currently, we have a few issues with how we render the schema types. The example that made me look into that:

* Nullable arrays containg objects were rendered as `array | null`.
* But we used to render `array object[] | null`, which is what we want.

**Solution**

This PR ended up being a refactor of how we render the schema types.

* moved the code to a separate file
* added lots of tests
* replaced a chunk of the other `getSchemaModelName` code

Note: It fixes the mentioned issues, but chances are I introduced other bugs. I’m short in time now, ideally we’d test this PR with a lot of documents, to spot bugs.

That said, my aim was to not write bugs.

**Before**

`firmMatterNumbers array | null`

Screenshot 2025-06-27 at 14 00 46

**After**

`firmMatterNumbers array object[] | null`

Screenshot 2025-06-27 at 14 05 13

**Checklist**

### DIFF
--- a/.changeset/lazy-rivers-grab.md
+++ b/.changeset/lazy-rivers-grab.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
+refactor: schema type rendering


packages/api-reference/src/components/Content/Schema/SchemaComposition.test.ts
```diff
@@ -136,7 +136,7 @@ describe('SchemaComposition', () => {
       })
 
       const tab = wrapper.find('.composition-selector-label')
-      expect(tab.text()).toBe('Array of string')
+      expect(tab.text()).toBe('array string[]')
```
antlio, line 139: 👍

antlio (approved):
great work, tested a lot of use cases

marclave (approved):
tested ! and works great ✨

PR #6097, opened 2025-06-29
feat(nestjs): use title option

ahmadi-akbar:
use missing title option instead `Scalar API Reference`

**Problem**

Currently, …

**Solution**

With this PR …

**Checklist**


marclave (approved):
nice fix @ahmadi-akbar ! appreciate the contribution ✨ 

want to add a changeset in the root of the repo `pnpm changeset` and make it a patch? ✨

PR #6100, opened 2025-06-30
refactor(api-reference): create parameter map from a plain operation object

hanspagel:
**Problem**

We’re using this `useOperation` hook, which was taking in the old `TransformedOperation`, that we want to get rid of.

**Solution**

I’ve refactored it to be a plain function, that’s taking in a plain OpenAPI Operation Object, and I’ve added tests.

I think we can actually delete it and just directly use the OpenAPI structure, but not today.

**Checklist**


marclave (approved):
🔥 🔥 🔥

PR #6101, opened 2025-06-30
refactor(oas-utils): create request examples without transformed operation

hanspagel:
**Problem**

There‘s probably funny story that I missed, but the API client creates request examples based on the outdated `TransformedOperation`, which is mocked in the API client, because we don’t have it.

**Solution**

I did a small refactor to get rid of the `TransformedOperation` there and just the data structure that we have (Operation entity from the API Client store).

**Checklist**


amritk (approved):
We should start thinking about how to handle all Path Item Objects as a whole

PR #6104, opened 2025-06-30
refactor(api-reference): use sidebar logic to render main content

hanspagel:
**Problem**

Currently, …

**Solution**

With this PR …

EDIT: There’s a better approach, I need to go back a step and try again.

**Checklist**


hanspagel:
new approach, new PR

PR #6105, opened 2025-06-30
feat(api-client): display auth combobox on indicator click

antlio:
**Problem**

currently within the api client, when an auth is required we get is displayed in the title section however it feels duplicated when selected, and also can be hard to read when joined, long named being used.

**Solution**

this pr is a proposal to:
- opens the auth combobox on auth indicator click

as we are already displaying the required/optional information from there, this feels like being the fastest way also to get the one selected when it is not vs using a popover or any other approaches.

| state | preview |
| -------|------|
| before | image |
| after | image |

this pr also fixes:
- authentication section clickable area by setting the combobox absolute to get the full section clickable when closed
- scalar combobox options log warning when open 

**Checklist**

### DIFF
--- a/.changeset/thick-mirrors-thank.md
+++ b/.changeset/thick-mirrors-thank.md
@@ -1,0 +1,5 @@
+---
+'@scalar/components': patch
+---
+
+fix: favors ref over useTemplateRef in scalar combobox option


packages/components/src/components/ScalarCombobox/ScalarComboboxOptions.vue
```diff
@@ -137,8 +129,8 @@ function moveActive(dir: 1 | -1) {
   })
 }
 
-// Manual autofocus for the inputAdd commentMore actions
-const input = useTemplateRef('input')
+// Manual autofocus for the input
+const input = ref<HTMLInputElement | null>(null)
```
antlio, line 133: @amritk seems to fix 
```console
[Vue warn] Set operation on key "value" failed: target is readonly.
```

marclave (approved):
nice :)

packages/components/src/components/ScalarCombobox/ScalarComboboxOptions.vue
amritk: cc: @hwkr

PR #6110, opened 2025-07-01
feat(api-reference): use xml.name as a fallback for schema titles

### LINKED ISSUE #5682: use `xml.name` when rendering schemas
OpenAPI also supports an `xml.name` override for XML element naming. Although valid per spec, it’s very rare in JSON-first APIs and isn’t used by me and not handled here

image

_Originally posted by @DmitriiKhudiakov in https://github.com/scalar/scalar/issues/5676#issuecomment-2892431645_

hanspagel:
**Problem**

Currently, we render the schema title if it’s available.

**Solution**

With this PR we’re also using the `xml.name` as a fallback, as suggested in https://github.com/scalar/scalar/pull/5676#issuecomment-2892431645

Fixes #5682

We’ll still look into the `title` attribute first.

**Preview**

Screenshot 2025-07-01 at 14 39 42

**Example**

```json
{
  "openapi": "3.1.0",
  "info": {
    "title": "xml.name",
    "version": "1.0"
  },
  "components": {
    "schemas": {
      "ExampleSchema": {
        "type": "array",
        "items": {
          "type": "object",
          "properties": {
            "id": {
              "type": "integer"
            }
          },
          "xml": {
            "name": "XmlTag"
          }
        },
        "xml": {
          "name": "XmlName"
        }
      }
    }
  }
}
```

**Checklist**

### DIFF
--- a/.changeset/pink-camels-sit.md
+++ b/.changeset/pink-camels-sit.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
+feat: use xml.name as a fallback for the schema title
--- a/packages/api-reference/src/components/Content/Models/Models.vue
+++ b/packages/api-reference/src/components/Content/Models/Models.vue
@@ -75,7 +75,11 @@ const models = computed(() => {
             <template #heading>
               <SectionHeaderTag :level="3">
                 <SchemaHeading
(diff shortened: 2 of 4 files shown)


marclave (approved):
😮‍💨 nice @hanspagel

Issue #5582, opened 2025-05-06
Schema references with overlapping sibling keywords not handled correctly


handrews:
### What happens?

This OAD:

```YAML
openapi: 3.1.1 
info:
  version: 1.0.0 
  title: Test schema $ref with siblings
paths:
  /foo: 
    get:  
      responses:
        "200":
          description: Ok
          content:
            application/json:
              schema:
                $ref: "#/components/schemas/Foo"
                pattern: "^foo"
components:
  schemas:
    Foo:  
      type: string
      pattern: "bar$"
```

uses JSON Schema draft 2020-12's ability to have keywords adjacent to `$ref`.  These _all_take effect, meaning that the response needs to be a string that both starts with `foo` and ends with `bar`.

However, Scalar only shows the `pattern: "^foo"` constraint.

### What did you expect to happen?

Both `pattern` constraints must be shown (and validated, used for code generation, etc.)

handrews:
BTW, I'm not sure what goes here vs scalar/openapi-parser.  If this should be filed over there, please feel free to close and I can re-file with the right info in that repo.

hanspagel:
Oh, I wasn’t aware that’s the case. I don’t really know how we’d deal with that. Maybe we should then keep an array of all those additional properties and apply all? Looks like that would make the whole rendering a lot more complex. Or is pattern just a special case, should everything else just overwrite (that part is easy to add)? 🤔 Any thoughts on this?

handrews:
@hanspagel it's not a special case, I'm afraid.  The model for JSON Schema draft 2020-12 referencing is one of delegation, not of textual replacement (because people really didn't like that adjacent keywords were ignored / dropped, and textual replacement has _never_ generally worked in JSON Schema because of the need to support cyclic references).

To fully support JSON Schema references, even back in draft-04, you can't just remove and replace them.  You have to track their effects.

handrews:
@hanspagel There are also complexities around the Path Item Object's `$ref` to consider (in all OAS versions), although that complexity is more amenable to textual replacement.

But in general, you can't quite do that in the OAS either, because the Link object's `operationRef` keyword and the URI form of the Discriminator Object's `mapping` are essentially `$ref`s that cannot be replaced.  In OAS 3.2, keys in the Security Requirement Object will also have a URI form analogous to how `mapping` works.

hanspagel:
Submitted a PR to our new OpenAPI-based data source: #6111

[...]

PR #6111, opened 2025-07-01
feat(workspace-store): keep sibling keywords when resolving $ref's

hanspagel:
**Problem**

JSON Schema draft 2020-12 allows keywords adjacent to $ref. Read more: #5582

**Solution**

With this PR we’re adding this behaviour to the new store that we’re about to migrate to.

**Checklist**

### DIFF
--- a/packages/workspace-store/src/helpers/proxy.ts
+++ b/packages/workspace-store/src/helpers/proxy.ts
@@ -53,13 +53,13 @@ function createProxyHandler(
             const siblingKeywords: Record<string, unknown> = {}
             for (const [key, val] of Object.entries(value)) {
               if (key !== '$ref') {
-                siblingKeywords[`x-sibling-${key}`] = val
+                siblingKeywords[key] = val
               }
             }
 
             // If there are sibling keywords, merge them with the resolved value
             if (Object.keys(siblingKeywords).length > 0) {
-              const mergedValue = isObject(resolvedValue)
+              const mergedValue: Record<string, unknown> = isObject(resolvedValue)
                 ? {
                     ...resolvedValue,
                     ...siblingKeywords,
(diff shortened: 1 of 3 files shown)


packages/workspace-store/src/helpers/proxy.ts
```diff
@@ -49,7 +49,47 @@ function createProxyHandler(
             const referencePath = parseJsonPointer(ref)
             const resolvedValue = getValueByPath(sourceDocument, referencePath)
 
-            // preserve the first $ref to maintain the original reference
+            // Extract sibling keywords (all properties except $ref)
+            const siblingKeywords: Record<string, unknown> = {}
+            for (const [key, val] of Object.entries(value)) {
+              if (key !== '$ref') {
```
DemonHa, line 55: We might wanna filter also the `ref` extension keys as: `$global` and `$status `

packages/workspace-store/src/helpers/proxy.ts
```diff
@@ -49,7 +49,47 @@ function createProxyHandler(
...
+              const mergedValue: Record<string, unknown> = isObject(resolvedValue)
+                ? {
+                    ...resolvedValue,
+                    ...siblingKeywords,
+                    'x-original-ref': originalRef ?? ref,
+                  }
+                : {
+                    ...siblingKeywords,
+                    'x-original-ref': originalRef ?? ref,
+                    value: resolvedValue,
```
DemonHa, line 71: I think we should spread the value here

DemonHa:
@hanspagel How do we handle property collisions when we merge ref siblings with the resolved values?

Also we have to think about the deep nested refs. Which properties do we keep? Top level one, deep level one, merge all of them?

hanspagel:
unclear what we even want to do :D I set it back to draft for now, until I made some deeper research on this

hanspagel:
Closing, because of #6486

PR #6126, opened 2025-07-02
fix(core): handle config properties that accept both string and function types

inyourtime:
**Problem**

Currently, the `getScriptTags` function assumes that certain configuration properties like `tagsSorter` and `operationsSorter` are always functions. This causes incorrect output or broken behavior when these properties are provided as strings (e.g., "alpha" or "method")

**Solution**

With this PR adding proper support for configuration properties that can be either a string or a function — such as `tagsSorter` and `operationsSorter`.

**Checklist**

### DIFF
--- a/.changeset/thirty-impalas-end.md
+++ b/.changeset/thirty-impalas-end.md
@@ -1,0 +1,5 @@
+---
+'@scalar/core': patch
+---
+
+fix: handle config properties that accept both string and function types


marclave (approved):
AWESOME work @inyourtime as usual :)

going to wait for @amritk & @hanspagel to give it a review before merging ✨

hanspagel (approved):
OMG, that’s great

PR #6131, opened 2025-07-02
feat(api-reference): improved rendering for consts

### LINKED ISSUE #5609: anyOf omitted value when type null, empty values when enum of type nullable, example is always 1 when nullable
### What happens?

When I have an `anyOf` type with one of the types of `{"type": "null"}`, then this option is omitted. When I have any type `{"const": null}` or const with other nullable type, then the option is shown as an option, but the panel is empty. When I have an example which is nullable, then `1` is shown as example. The description in the `property-detail-valui` is also shown only as `nullable`.

The description, the schema tabs for each type and the panels should show the correct information.

### OpenAPI Document

https://sandbox.scalar.com/e/uioeo

hanspagel:
**Problem**

`consts`, especially with falsy values weren’t rendered well.

**Solution**

This PR fixes a bunch of bugs for consts:

* const value 0, false, null aren’t rendered at all
* consts were rendered as “Schema” in the composition dropdown
* if the example value for a response is `null`, it wouldn’t render

Fixes #5609

**Preview**

Screenshot 2025-07-02 at 11 59 41

**Example**

```json
{
  "openapi": "3.1.0",
  "info": {
    "title": "FooAPI",
    "version": "1.0.0"
  },
  "paths": {
    "/foo": {
      "get": {
        "summary": "Get Foo",
        "operationId": "",
        "parameters": [],
        "responses": {
          "200": {
            "description": "Successful Response",
            "content": {
              "application/json": {
                "schema": {
                  "anyOf": [
                    {
                      "type": "number"
                    },
                    {
                      "const": 1
                    },
                    {
                      "type": "null"
                    },
                    {
                      "const": null
                    },
                    {
                      "const": false
                    },
                    {
                      "const": ""
                    },
                    {
                      "const": 0
                    }
                  ]
                },
                "example": null
              }
            }
          }
        }
      }
    }
  }
}
```

**Checklist**


packages/api-reference/src/components/Content/Schema/SchemaPropertyHeading.vue
```diff
@@ -180,11 +181,11 @@ const modelName = computed(() => {
       <Badge>deprecated</Badge>
     </div>
     <div
-      v-if="isDefined(constValue)"
+      v-if="constValue !== undefined"
```
hanspagel, line 186: would not render `null`

amritk (approved):
Yep this makes sense

PR #6132, opened 2025-07-02
fix(api-reference): deprecated operations aren’t striked in the sidebar anymore

### LINKED ISSUE #6129: Deprecated endpoints no longer displayed with strikethrough in navigation
### What happens?

After upgrading from version 1.28.23 to 1.28.31 we noticed that the depreciated endpoints are no longer formatted in strikethrough in the navigation bar.

The endpoint summaries in the navigation bar should be formatted in strikethrough as they used to be in older versions of Scalar.

### OpenAPI Document

https://sandbox.scalar.com/e/bcfL7#tag/default/get/old-endpoint

hanspagel:
**Problem**

With the recent sidebar refactor we introduced a regression: deprecated operations aren’t rendered with a strike-through style anymore.

**Solution**

Fixes #6129

**Checklist**

### DIFF
--- a/.changeset/pretty-timers-impress.md
+++ b/.changeset/pretty-timers-impress.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
+fix: deprecated operations aren’t striked in the sidebar anymore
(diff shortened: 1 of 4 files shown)


amritk (approved):
Good catch

PR #6136, opened 2025-07-02
ci: add renovate configuration

hanspagel:
**Problem**

Currently, we’re using Dependabot to keep our dependencies up to date. It feels a bit … outdated, though. It fails to update the lock file, for updates in subdirectories. If often struggles to show relevant information from the changelog. And sometimes it just hangs. Created PRs are often red for strange reasons.

**Solution**

This PR translates the Dependabot configuration to be used with [Renovate](https://docs.renovatebot.com/), a wonderful Dependabot alternative that I see in more and more open-source repositories.

Just one of the features that I like: Instead of creating tons of PRs and spamming the queue, it’s creating a single GitHub issue, that has a task list with all the pending updates. To create a PR you can just check a box. Example: https://github.com/unjs/openapi-renderer/issues/1

✅ I’ve installed the required Renovate GitHub (scalar/scalar only), with read + write access to the code.
✅ It passes the `npx --yes --package renovate -- renovate-config-validator` check.

**Checklist**


renovate.json
```diff
@@ -0,0 +1,41 @@
+{
+  "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+  "extends": ["config:recommended"],
+  "schedule": ["monthly"],
+  "prConcurrentLimit": 0,
+  "packageRules": [
+    {
+      "matchManagers": ["github-actions"],
+      "matchFileNames": ["/"]
```
xC0dex: I think this package rule doesn't have any effect.

renovate.json
```diff
@@ -0,0 +1,41 @@
...
+    },
+    {
+      "matchManagers": ["npm"],
+      "matchFileNames": ["/"],
```
xC0dex: `matchFileNames` shouldn't be necessary here.

renovate.json
```diff
@@ -0,0 +1,41 @@
...
+      "ignoreDeps": ["vite", "vitest", "rollup", "tsup"]
+    },
+    {
+      "matchManagers": ["nuget"],
+      "matchFileNames": ["/integrations/aspnetcore"],
```
xC0dex: This requires a wildcard pattern. otherwise it will not match a file.

hanspagel:
thanks for the feedback, handing it over to @xC0dex

PR #6142, opened 2025-07-02
fix(api-client): input focus post collection description update

### LINKED ISSUE #6103: api-client can not add server
### What happens?

**Steps to reproduce**

When changing the collection description and then I go to add a new server or edit the document title I can't type on the input box

Image

I should be able to write on the input box

### OpenAPI Document

_No response_

antlio:
**Problem**

currently when updating api client collection description and switching more to preview breaks the focus within input preventing usage of modal for instance.

**Solution**

this pr delays mode switch to avoid input typing issue. fixes #6103

**Checklist**

### DIFF
--- a/.changeset/violet-terms-draw.md
+++ b/.changeset/violet-terms-draw.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
+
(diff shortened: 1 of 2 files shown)


hanspagel (approved):
tested, seems to fix the issue

PR #6144, opened 2025-07-02
chore(snippetz): move @scalar/types to dependencies

DemonHa:
**Problem**

Currently, we are installing `@scalar/types` as a dev dependency, which is going to cause problems when we install this package.

**Solution**

With this PR we move `@scalar/types` to dependencies

**Checklist**


amritk (approved):
Yup this bites us all the time!

PR #6145, opened 2025-07-03
feat(api-reference): OpenAPI request example block

amritk:
Alright so after working on the blocks a bit, I ended up with a core component which we can use ourselves. And a block wrapper. The core component is the "dumb" one which has [almost] no internal state and depends on props and events.

This block wrapper should be able to:
- stand on its own and work with nothing more than the basic props (document url, method, path)
- take in a workspace store OR create its own global workspace store which is shared between the blocks
- have access to a global block store to share state between blocks, like selected client for example
- probably be exported 3 ways to start: vue, react/mdx, esm

TODO:
- [x] ensure it works with blocks from different documents (currently using activeDocument need to move away from that) on the same page
- [x] ensure each helper and component have tests
- [x] replace example picker with scalar dropdown
- [ ] consume the core component in references
- [x] save selected client to local storage

Punted:
- [ ] ~move blocks into their own package with no dependency on references or the client~
  - [ ] ~have all of the exports working~


amritk:
closed in favor of https://github.com/scalar/scalar/pull/6191

PR #6148, opened 2025-07-03
Hanspagel/doc 3697 nested additionalproperties do not display instead showing

hanspagel:
**Problem**

Currently, …

**Solution**

With this PR …

**Checklist**

### DIFF
--- a/.changeset/selfish-flowers-bow.md
+++ b/.changeset/selfish-flowers-bow.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
+feat: render additionalProperties and patternProperties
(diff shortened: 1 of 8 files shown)


antlio:
how does it feel to keep the value first and use regex | unknown as a property detail (e.g. string, required)?

hanspagel:
@antlio this is what we did before and it’s a bit strange, because everything that's after the property name is about the VALUE (string, maxLength…), not the PROPERTY

antlio:
> @antlio this is what we did before and it’s a bit strange, because everything that's after the property name is about the VALUE (string, maxLength…), not the PROPERTY

[non-blocking]: true i see it after reading the spec document, maybe we should handle this differently to keep it closer to the context and not as another property as before. cc @cameronrohani

cameronrohani (approved):

marclave (approved):
lets merge once we get the conflicts touched up @hanspagel <3

PR #6174, opened 2025-07-04
fix(api-reference): content order doesn’t match sidebar order

### LINKED ISSUE #6062: Using a ///<summary>test</summary> is affecting sort order of tags
### What happens?

Using Scalar UI 2.5.0 and .NET 9

If i use:
`config.WithTagSorter(TagSorter.Alpha);`
Then the sidebar get's sorted correctly. But the right side content will then not match the sidebar sort.

It's quite easy to test: 

Image

As you can see the Event is the first tag/endpoint displayed on the right. While Activity is the first on the sidebar. This is incorrect as Activity according to the sidebar should be on top (which is how I want it), not Event. 

As soon as i remove the `///<summary>some text</summary>` part of: 
Image
The sortorder instantly changes (and the next endpoint/tag with a` ///<summary></summary>`) appears on top instead:

Image

is `<summary></summary>` info stuff really supposed to influence sortorder? If so, why? This also has the consequence of totally messing up the sidebar tracking of "where you are" and it flies all over the place when i scroll down on the right.

The right-side should match the sort order of the sidebar tags and operations. In my case here Activity should be at the top, and not Event, ShiftSummary, etc. 

### OpenAPI Document

_No response_

hanspagel:
**Problem**

When passing `tagsSorter: 'alpha'` the order of the tags in the content doesn’t match the order in the sidebar.

**Solution**

This PR refactors how we loop through tags and operations in the content to re-use the logic from the sidebar, and those felt unavoidable:

* Refactored a bunch of components to *not* use the deprecated ~~`TransformedOperation`~~ anymore
* Moved the code for the `IntroductionSection`, `OperationsSection`, `WebhooksSection` and `ModelsSection` to separate components

Fixes #6062

**Checklist**


packages/api-reference/src/features/traverse-schema/index.ts
```diff
@@ -4,7 +4,10 @@ export type {
   TraversedEntry,
   TraversedDescription,
   TraversedTag,
+  TraversedWebhook,
```
amritk, line 7: We aren't going to use this whole feature section, the traversal will happen in the workspace store once we switch

hanspagel: but for now, we need it and we only want to import from `@/features/whatever`, not from any subdirectory, right?

amritk: yea yea, once we switch to workspace we will delete this feature

amritk:
Alright its good to go, once its merged lets hold release and we should do lazy loading + parameter links at the same time

amritk:
This will be conflicted until I revert my revert

amritk:
@marclave give it a go!

PR #6180, opened 2025-07-04
fix(openapi-parser): applyChangesToDocument undefined type check

antlio:
**Problem**

we are currently checking for the string 'undefined' within the applyChangesToDocument function instead of the value, which can lead to not setting schema type properly.

**Solution**

this pr updates the function to check the value is defined instead.

**Checklist**


hanspagel (approved):
Great catch my friend

PR #6185, opened 2025-07-07
feat(oauth2): add support for x-tokenName extension

### LINKED ISSUE #6179: Add x-tokenName extension to OAuth flows
### Discussed in https://github.com/scalar/scalar/discussions/6173

<sup>Originally posted by **tiaringhio** July  4, 2025</sup>
i was wondering if it's possibile to save in the UI the value of the ID Token instead of the value of the Access Token for an OAuth2 authentication, i am currently migrating from Swashbuckle (.NET) and i currently have this piece of code that does exactly what i need.

```csharp
s.AddSecurityDefinition(AuthConstants.EntraIdSecurityDefinition,
  new OpenApiSecurityScheme
  {
      Name = "Entra ID",
      Type = SecuritySchemeType.OAuth2,
      Description = "Entra ID Authentication to secure access to API.",
      Extensions = new Dictionary<string, IOpenApiExtension>
      {
          {"x-tokenName", new OpenApiString("id_token")}
      },
      Flows = new OpenApiOAuthFlows
      {
          AuthorizationCode = openApiOAuthFlow
      },
  });
```

that allows me to call my API using the Id Token instead of the access token.
is it possibile to have this functionality? is there any way i can work around that in the meantime?

i have found, looking here an there, this issues (not scalar's), they might be of help:

- https://github.com/swagger-api/swagger-ui/issues/7561
- https://github.com/domaindrivendev/Swashbuckle.AspNetCore/issues/1128

i have looked in the documentation and ave not found what i was looking for, if i missed it can you point in the direction of the info?

thank you for the help</div>

xC0dex:
This PR introduces a new extension `x-tokenName` at the OAuth2 Flow level. It allows users to specify which property from the OAuth2 token response should be used.
By default, `access_token` is assumed, but this can now be customized via the` x-tokenName` extension.

Is the flow level OK?

Closes #6179

**Checklist**


amritk (commented):
Hey @xC0dex wicked PR, could you just also add this to the extensions section so we can keep track

https://github.com/scalar/scalar/tree/main/packages/openapi-types/src/schemas/extensions

xC0dex:
@amritk, I added the extension 👍. I guess we could refactor the extensions a bit to make use of these dedicated extension types?

hanspagel (approved):
super cool, thanks for the initiative on this!

amritk:
we don't do it now, but yea we either have a system that adds the type where its used OR we just do a schema.parse when we are using an extension to get the type. In the new store all extensions are allowed and typed with `unknown` so we should probably just parse them

PR #6191, opened 2025-07-08
feat(api-reference): OpenAPI request example block 

amritk:
## New:
This is just the dumb component for now, as we are only using internally. Once we want to export blocks for users we can bring it back. This PR just updates the code and tests, next PR will be integration.

## OLD:

Alright so after working on the blocks a bit, I ended up with a core component which we can use ourselves. And a block wrapper. The core component is the "dumb" one which has [almost] no internal state and depends on props and events.

The block wrapper can:
- stand on its own and work with nothing more than the basic props (document url, method, path)
- take in a workspace store OR create its own global workspace store which is shared between the blocks
- have access to a global block store to share state between blocks, like selected client for example
- block store saves to/restores from local storage
- works with blocks with the same + different documents on the page
- fully tested!

Its still not used so no need to test, just do code review + check tests to see if I covered all the cases for feature parity.

Pushed to next PR:
- replace the request example in the operation 
- create a openapi blocks package
- setup the build system in the blocks package to export esm, vue, react/mdx
- add more blocks!

**Checklist**


hanspagel (approved):
brother, your work is outstanding

packages/api-reference/src/v2/blocks/helpers/block-store.ts
```diff
@@ -0,0 +1,75 @@
...
+/** The base global state of the block store */
+const state = reactive<BlockStoreState>({
+  selectedClient: null,
+})
+
+/** Local storage key for the block store */
+export const BLOCK_STORE_LS_KEY = 'scalar-blocks-state'
+
+/** Keeps track of whether the block store has been restored from localStorage */
+let restored = false
```
marclave: could we use a ref here instead of a let?

packages/api-reference/src/v2/blocks/helpers/get-workspace-store.ts
```diff
@@ -0,0 +1,78 @@
...
+
+/** Lets us keep track of loading documents by name so we only load them once */
+const documentMap = new Map<string, Promise<void>>()
+
+/**
+ * Creates or returns the existing singleton store instance.
+ * This ensures we only create one store per application lifecycle.
+ */
+const createStore = (): WorkspaceStore => {
+  store ||= createWorkspaceStore()
```
marclave: :o nice

packages/api-reference/src/v2/blocks/helpers/block-store.ts
geoffgscott, line 1: No external stores other than workspace. That's the whole point :wink: 

Anything that comes into the block needs to be a prop. For other frameworks we would just use a wrapping element that maps the framework reactivity to vue reactivity. (useEffect -> update reactive object -> component props)

packages/api-reference/src/v2/blocks/helpers/get-document-name.ts
geoffgscott, line 1: @DemonHa (cc @amritk ) Let's put this logic directly into the store if its needed.

In the workspace store it is required that documents have a unique slug (the `documents` field of the workspace is an object) so we can use `info.title` or fallback to that.

packages/api-reference/src/v2/blocks/helpers/get-workspace-store.ts
geoffgscott, line 1: This needs to go away as well. If someone is using blocks it is expected they have created an store independently and will pass it in as a prop. 

@marclave we will need to document this appropriately and probably expand the @scalar/react package to include some workspace helpers.

packages/api-reference/src/v2/blocks/scalar-request-example-block/components/ExamplePicker.vue
```diff
@@ -0,0 +1,66 @@
...
+
+/** Handle example selection */
+const selectExample = (option: ScalarComboboxOption) => {
+  selectedExampleKey.value = option.id
+}
+</script>
+
+<template>
+  <ScalarCombobox
+    class="scalar-example-picker max-h-80"
```
geoffgscott: `scalar-example-picker` what is this class for? If this is the supposed to be the use targetable class we should get the enum going in @scalar/themes. 

In this case would recommend `t-reference__example-picker` or `t-blocks__example-picker`

packages/api-reference/src/v2/blocks/scalar-request-example-block/components/RequestExample.vue
```diff
@@ -238,29 +248,33 @@ const id = useId()
...
       </div>
 
       <!-- Client picker -->
       <template
         #actions
-        v-if="!config.hideClientSelector">
+        v-if="!hideClientSelector">
         <ScalarCombobox
-          class="max-h-80"
+          class="scalar-client-picker max-h-80"
```
geoffgscott: Same comment as above for class names

packages/api-reference/src/v2/blocks/scalar-request-example-block/components/ScalarRequestExampleBlock.vue
```diff
@@ -0,0 +1,132 @@
...
+  ...getWorkspaceStoreProps
+} = defineProps<ScalarRequestExampleBlockProps>()
+
+defineSlots<{
+  header: () => unknown
+  footer: () => unknown
+}>()
+
+/** Either creates or grabs a global/prop store */
+const store = getWorkspaceStore(getWorkspaceStoreProps)
```
geoffgscott: Do we really need to have access to the global store here? Ideally we would pass in only the required parts of the openAPI document (operation/security schemes/...?)

packages/api-reference/src/v2/blocks/scalar-request-example-block/components/ScalarRequestExampleBlock.vue
```diff
@@ -0,0 +1,132 @@
...
+
+      if (pathItem) {
+        operation.value = pathItem
+      }
+    } catch (error) {
+      console.error('Failed to resolve operation:', error)
+    }
+  },
+  { immediate: true },
+)
```
geoffgscott: All of this should be replaced by just accepting atomic parts of the document and not the whole workspace store.

packages/oas-utils/src/helpers/operation-stability.ts
```diff
@@ -0,0 +1,36 @@
...
+export const isOperationDeprecated = (operation: OperationStability): boolean =>
+  operation.deprecated || operation['x-scalar-stability'] === XScalarStabilityValues.Deprecated
+
+/**
+ * Get operation stability from deprecated or x-scalar-stability
+ */
+export const getOperationStability = (operation: OperationStability): XScalarStabilityValues | undefined =>
+  operation.deprecated
+    ? XScalarStabilityValues.Deprecated
+    : (operation['x-scalar-stability'] as XScalarStabilityValues | undefined)
```
geoffgscott, line 20: @DemonHa @amritk let's move this to workspace store. @DemonHa we might want to consider what we do with these "augmented" properties with regards to saving the documents after the fact.

packages/oas-utils/src/helpers/operation-stability.ts
```diff
@@ -0,0 +1,36 @@
...
+    case XScalarStabilityValues.Deprecated:
+      return 'text-red'
+    case XScalarStabilityValues.Experimental:
+      return 'text-orange'
+    case XScalarStabilityValues.Stable:
+      return 'text-green'
+    default:
+      return ''
+  }
+}
```
geoffgscott, line 36: Once it moves to workspace store this can probably just live as a local computed in the component.

packages/api-reference/src/v2/blocks/scalar-request-example-block/components/ScalarRequestExampleBlock.vue
amritk: Actually this is the block which doesn't work like that. Its completely standalone and handles everything. What you are thinking of is the core component which is alongside here.

packages/api-reference/src/v2/blocks/helpers/block-store.ts
amritk: we don't need a ref here actually

packages/api-reference/src/v2/blocks/helpers/block-store.ts
amritk: Agreed on no external store, but this is internal state shared between blocks

packages/api-reference/src/v2/blocks/helpers/get-workspace-store.ts
amritk: not sure what you mean by this? Since users consuming the blocks don't know anything about the store

packages/api-reference/src/v2/blocks/scalar-request-example-block/components/ExamplePicker.vue
amritk: Actually isn't even used, can remove it

packages/api-reference/src/v2/blocks/scalar-request-example-block/components/RequestExample.vue
amritk: This is used to target for the tests since data attributes aren't working with headless teleports

packages/api-reference/src/v2/blocks/scalar-request-example-block/components/ScalarRequestExampleBlock.vue
amritk: I believe you are thinking about the core component

packages/oas-utils/src/helpers/operation-stability.ts
amritk: Currently all of our extensions live [here](https://github.com/scalar/scalar/tree/main/packages/openapi-types/src/schemas/extensions). Should I just move it there. The plan is probably to move them to workspace store eventually but we can just move them all at once.

DemonHa: @geoffgscott Agree! We should move all of them to the new store

packages/api-reference/src/v2/blocks/scalar-request-example-block/components/RequestExample.vue
```diff
@@ -167,7 +173,10 @@ const localSelectedClient = ref<ClientOption>(
 watch(
   () => selectedClient,
   (newClient) => {
-    localSelectedClient.value = findClient(clients.value, newClient)
+    const client = findClient(clients.value, newClient)
+    if (client) {
+      localSelectedClient.value = client
+    }
   },
 )
```
geoffgscott, line 181: Why are we keeping a locally selected client? Out of scope for this PR maybe but it's certainly a code smell.

packages/api-reference/src/v2/blocks/scalar-request-example-block/components/RequestExample.vue
geoffgscott: Can we instead be more explicit with the naming then. `test-scalar-client-picker`.

It's already a bit of a nightmare keeping track of what CSS classes are required/deprecated/totally unused.

packages/api-reference/src/v2/blocks/scalar-request-example-block/components/RequestExample.vue
amritk: because you can have custom clients which are specific to an operation. This is just copied from how it works now, we can change it later if we want. I didn't bother re-doing it

packages/api-reference/src/v2/blocks/scalar-request-example-block/components/RequestExample.vue
amritk: hey what do you know, its not used anymore since we deleted the other code lol

PR #6198, opened 2025-07-08
fix(oas-utils): spec url used as base url in local

antlio:
**Problem**

recent changes favored documentUrl fallback when no server is provided, in order to ensure import to client. however in some cases like local development, we want to maintain if local url, local host as fallback.

**Solution**

this pr updates the import spec `getServersFromOpenApiDocument` function in order to bring back localhost server fallback in local development.

**Checklist**


packages/oas-utils/src/transforms/import-spec.ts
```diff
@@ -570,9 +571,25 @@ export function getServersFromOpenApiDocument(
...
+
+    // If not localhost and we have a documentUrl, we use it as the fallback
+    if (documentUrl) {
+      return [serverSchema.parse({ url: getBaseUrlFromDocumentUrl(documentUrl) })]
+    }
+
+    // Last fallback to the current window origin
+    if (fallbackUrl) {
+      return [serverSchema.parse({ url: fallbackUrl })]
+    }
```
antlio, line 592: not sure if we really need this, but double guard

amritk (approved):
Makes sense to me 👍🏾

xC0dex:
We should merge this PR and create a release. Many people are experiencing this issue 👀

antlio:
> We should merge this PR and create a release. Many people are experiencing this issue 👀

@hanspagel might take another approach to this, waiting for his 🟢

hanspagel (changes requested):
I think we want to fix relative URLs and keep the other tests, let me have a look

hanspagel:
Thanks for putting this together, the issue is not really related to localhost, it’s for all URLs, though. Closing this in favor of #6207. 🤟

PR #6201, opened 2025-07-08
fix(components): scalar combobox options scroll on open

amritk:
**Problem**

Currently, the page jumps when you open a large combobox. It can be seen in:

https://github.com/scalar/scalar/pull/6191

**Solution**

With this PR we revert back to the timeout which does not cause a scroll (nextTick does).

**Checklist**

### DIFF
--- a/.changeset/tough-bugs-report.md
+++ b/.changeset/tough-bugs-report.md
@@ -1,0 +1,5 @@
+---
+'@scalar/components': patch
+---
+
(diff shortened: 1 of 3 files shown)


hwkr (approved):
Oh that's annoying, ty for the fix 🙏

hwkr:
Updated the focus tests to account for the `setTimeout` 🫡

marclave (approved):

PR #6202, opened 2025-07-08
feat(themes): align default theme accent color

bgrcs:
**Problem**

We previously updated the accent color to always match `--scalar-color-1`; this change was mainly intended for links, but now we are using `----scalar-link-color.`

This change mainly affects org as we have a few toggles sprinkled around, and they don't look great at the moment in dark mode as we are using the accent color as the background.

cc @cameronrohani 

**Solution**

Sets a new light/dark blue accent color, which will be used primarily by the toggle component.

**Checklist**


hwkr (approved):
Looks good!! Pushed two tweaks:
1. We also had the `--scalar-link-color`'s defined in `variables.css` so I removed those (I think having them with the other colors / button colors in default makes more sense).
2. `--scalar-color-accent` was already defined higher up in `default.css` so I moved the colors up there so we're not setting it twice.

PR #6203, opened 2025-07-09
refactor(components): update slot definitions to use unknown instead of any

hwkr:
Clean up the slot definitions so they all match and use unknown as the return type instead of any. Shouldn't have any downstream effects AFAIK.

**Checklist**


amritk (approved):
send itt

hwkr:
> is there a reason to drop the Slot type?

It doesn't actually do anything and is kinda misleading, it was my bad I thought it was doing something it wasn't when I used it. The [`defineSlots` docs](https://vuejs.org/api/sfc-script-setup#defineslots) recommend using `any` but I think Amrit is right that `unknown` is probably a bit better.

marclave (approved):

PR #6207, opened 2025-07-09
fix(api-client): do not use relative URL as a server URL

### LINKED ISSUE #6192: Spec URL used as Base URL
### What happens?

Image

This issue has appeared very recently, where Scalar suddenly uses the Spec path as a base URL for all requests.

Below you can see the HTML we're using to set up Scalar:
```
<!DOCTYPE html>
<html>
  <head>
    <title>{title}</title>
    <meta charset="utf-8" />
    <meta
      name="viewport"
    <style>
      body {
        margin: 0;
     }
    </style>
  </head>
  <body>


    <div id="app"></div>
    <script>
        url: '/docs/private/api.json'
      })
    </script>
  </body>
</html>
```

It should not use the spec path, as that does not make any sense. The default fallback used to be the URL that scalar itself is reachable at. e.g. "http://localhost:8080"

### OpenAPI Document

_No response_

hanspagel:
**Problem**

When a document without servers is imported from a relative URL, the relative URL is used as a server URL.

Which is not good.

image

**Solution**

With this PR we won’t use relative document URLs as server fallbacks anymore.

Fixes #6192

**Checklist**

### DIFF
--- a/.changeset/gorgeous-berries-pump.md
+++ b/.changeset/gorgeous-berries-pump.md
@@ -1,0 +1,5 @@
+---
+'@scalar/oas-utils': patch
+---
+
+fix: uses relative document URL as server URL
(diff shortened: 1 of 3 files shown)


markcda (approved):
(*just in case xd, I know that I'm not the code owner! *)

hanspagel:
We need this, I just take the reviews that I got and merge it :)

PR #6209, opened 2025-07-10
feat(components): sidebar component follow up

hwkr:
Two small follow up updates:

1. Changes the toggle icon to be 12x12 instead of 16x16
2. Sets the nested group active style via prop instead of with CSS

For the active style it would be nice to be able to do it with just CSS but it means that the contents always need to be rendered (instead of unmounted with a v-if) which will probably be untenable for large sidebars with many children. Alternatively we could use a `v-show` instead of a `v-if` for the subitems if we're okay with large dom structures being there even if they aren't being rendered.

**Checklist**


hanspagel (approved):
nice tweaks

marclave (approved):

PR #6212, opened 2025-07-10
feat(api-reference): moves operation response property type to list item

antlio:
**Solution**

this pr hides the property name in the operation responses to make the information easier to grasp.

| state | preview |
| -------|------|
| before | image |
| after | image |

**Checklist**

### DIFF
--- a/.changeset/dry-onions-kneel.md
+++ b/.changeset/dry-onions-kneel.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
+feat: moves operation response property type to list item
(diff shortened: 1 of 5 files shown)


hanspagel:
I think we want to show the title of the schema if we know it

antlio:
> I think we want to show the title of the schema if we know it

isn't it gonna end up the same as what @cameronrohani was highlighting regarding title vs description? cc galaxy

marclave:
dont we want to do the type alongside the title? but not as its own line?

hanspagel:
Awesome, I’ll review this once the release is unblocked and put it on hold for now. ✌️

hanspagel:
I think we’ll refactor the whole thing instead ✌️

PR #6218, opened 2025-07-10
feat(api-reference): implement the first openapi block for request example

### LINKED ISSUE #5009: Data not being displayed for application/x-www-form-urlencoded
### What happens?

When running API reference from CLI `scalar serve` I notice that application/x-www-form-urlencoded  data property is not being displayed. This works correctly when rolling back to version `1.25.103`

Image

Payload should be displayed in the code sample when type is application/x-www-form-urlencoded 

### OpenAPI Document

https://api-docs.bolt.com/api.bundle.yaml

### LINKED ISSUE #5279: defaultHttpClient with x-codeSamples doesn't work
### What happens?

When specifying x-codeSamples the defaultHttpClient is ignored. 

I explicitly set defaultHttpClient to:

`defaultHttpClient: {
  targetKey: "shell",
  clientKey: "curl",
},`

However the first x-codeSamples is shown by default.

Also, when changing the client in the dropdown, going from the default selected x-codeSamples to the defaultHttpClient does not update the client.

Take the following scenario:

Examples
- Code Sample 1 (selected)
- Shell Curl

When I click the Examples dropdown and select the Shell Curl client, the code and selected client do not change. This only happens when going from the first x-codeSamples to the defaultHttpClient.

Expectations are that:
- defaultHttpClient is respected when using x-codeSamples
- it's possible to navigate from x-codeSamples to the defaultHttpClient in the client dropdown

### OpenAPI Document

_No response_

amritk:
**Problem**

closes #5009
closes #5279

Currently, we use a mix of the client store and parsedSpec in the references operation 

**Solution**

With this PR we move the operation to the new workspace store ALMOST completely. We still use the client store for the selected server and auth, but I will move those over later as the blocks are created. The rest just uses the workspace store :). Then we consume our new openapi block for the request code example

Update: I just kept the old operation for discriminator stuff as I don't want to mess with it too much. ~Though if we want it to affect the code sample and other things then we should sort it out now~ Turns out it doesnt currently work anyway so we'll do it proper in a fast follow.

Needs testing on all the things related to operation + its children

**Checklist**


packages/api-reference/src/features/Operation/components/callbacks/Callbacks.vue
```diff
@@ -1,15 +1,40 @@
...
+} = defineProps<{
+  path: string
+  method: HttpMethod
+  callbacks: CallbackObject
   schemas?: Schemas
 }>()
+
+const { resolve } = useStore()
+
+// Ensure we resolve any external references
```
amritk, line 27: Will not need this if we resolve higher up, but it was here from before, can remove if needed

packages/api-reference/src/v2/ApiReferenceWorkspace.vue
```diff
@@ -177,34 +183,77 @@ onMounted(() => {
...
+  store.update('x-scalar-active-document', event.detail.value)
 })
 
 onCustomEvent(root, 'scalar-update-selected-client', (event) => {
-  store.update('x-scalar-default-client', event.data)
-  safeLocalStorage().setItem(REFERENCE_LS_KEYS.SELECTED_CLIENT, event.data)
+  store.update('x-scalar-default-client', event.detail)
+  safeLocalStorage().setItem(REFERENCE_LS_KEYS.SELECTED_CLIENT, event.detail)
+
+  // Temp to keep the old stuff working, remove when moved over to new store
```
amritk, line 211: Theres a bunch of syncing code between the old state and new in here

packages/api-reference/src/components/ApiReferenceLayout.vue
```diff
@@ -296,27 +292,6 @@ provide(
   }),
 )
 
-/** Helper utility to map configuration props to the ApiReference internal state */
```
amritk, line 299: moved to workspace

packages/api-reference/src/v2/ApiReferenceWorkspace.vue
```diff
@@ -263,33 +311,35 @@ useFavicon(favicon)
...
-    <template #document-selector>
-      <DocumentSelector
-        v-model="selectedDocumentIndex"
-        :options="availableDocuments" />
-    </template>
-    <template #sidebar-start>
-      <slot name="sidebar-start" />
-    </template>
-  </ApiReferenceLayout>
+  <div ref="root">
```
amritk, line 314: we were missing the root for events

marclave (approved):

PR #6220, opened 2025-07-11
feat(components): create scalar card component

hwkr:
I removed a bunch of unused code and I don't _think_ I broke anything but some thorough testing help would be appreciated 🙏 

**Checklist**


hanspagel (approved):
Looks awesome

hwkr:
> double check we dont use any of this in org? but good to go after ✨

Doesn't look like it! We use another card in Agent but it's not the same as this one (could maybe reconcile them in the future though)

PR #6224, opened 2025-07-11
feat(api-reference):  lay groundwork for block implementation

amritk:
This is just to break up https://github.com/scalar/scalar/pull/6218 into two PR's. This one lays the groundwork and actually shouldn't change anything as its mostly touching files/components that aren't being used. So its safe to merge and it brings the cognitive load down on reviewing the other one!

**Checklist**

### DIFF
--- a/packages/api-reference/src/components/ApiReference.vue
+++ b/packages/api-reference/src/components/ApiReference.vue
@@ -18,6 +18,11 @@ defineProps<{
  * TODO: Reference config should set the appropriate workspace properties
  */
 const workspaceStore = createWorkspaceStore()
+
+if (typeof window !== 'undefined') {
+  // @ts-expect-error - For debugging purposes expose the store
+  window.dataDumpWorkspace = () => workspaceStore
+}
 </script>
 
 <template>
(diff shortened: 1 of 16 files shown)


packages/api-reference/src/v2/ApiReferenceWorkspace.vue
```diff
@@ -184,23 +185,23 @@ onMounted(() => {
 // })
 
 onCustomEvent(root, 'scalar-update-dark-mode', (event) => {
-  store.update('x-scalar-dark-mode', event.data.value)
+  store.update('x-scalar-dark-mode', event.detail.value)
```
amritk, line 188: these events aren't being used yet

packages/api-reference/src/components/ApiReference.vue
```diff
@@ -18,6 +18,11 @@ defineProps<{
  * TODO: Reference config should set the appropriate workspace properties
  */
 const workspaceStore = createWorkspaceStore()
+
+if (typeof window !== 'undefined') {
```
amritk, line 22: super helpful for debugging!

PR #6226, opened 2025-07-11
fix(api-reference): prevents discriminator selector display within nested properties

antlio:
**Problem**

currently if amongst nested properties the targeted property name in discriminator presence is also existing, the discriminator get displayed making when it should only be displayed at the schema level set.

**Solution**

this pr updates the schema object properties component in order to add a null check for discriminator schema and avoid duplicated display.

| state | preview |
| -------|------|
| before | image |
| after |  image |

**Checklist**

### DIFF
--- a/packages/api-reference/src/components/Content/Schema/SchemaObjectProperties.vue
+++ b/packages/api-reference/src/components/Content/Schema/SchemaObjectProperties.vue
@@ -106,8 +106,8 @@ const getAdditionalPropertiesValue = (
         schema.discriminator?.propertyName || discriminatorPropertyName
       "
       :isDiscriminator="
-        property ===
-        (schema.discriminator?.propertyName || discriminatorPropertyName)
+        property === schema.discriminator?.propertyName &&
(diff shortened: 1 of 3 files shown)


packages/api-reference/src/components/Content/Schema/SchemaObjectProperties.vue
```diff
@@ -106,8 +106,8 @@ const getAdditionalPropertiesValue = (
         schema.discriminator?.propertyName || discriminatorPropertyName
       "
       :isDiscriminator="
-        property ===
-        (schema.discriminator?.propertyName || discriminatorPropertyName)
+        property === schema.discriminator?.propertyName &&
+        !!schema.discriminator
```
antlio, line 110: too restrictive, now it doesn't apply at all in nested level which can still hold discriminator

antlio:
will get back to this w/ @amritk once the new workspace store is set so we can rely on `x-original-ref` to fix schema retrieval in some circular reference use case (cc galaxy example).

PR #6227, opened 2025-07-11
feat(api-reference): lazy loading v2

amritk:
## Update
Will either pick this up after the hans refactor PR or just merge it into that one

**Problem**

Currently, there are some issues with lazy loading - its fragile and breaks often.

**Solution**

With this PR we introduce lazy loading v2. I got rid of the loading component which causes issues sometimes and instead we render directly to where we want first. Then "freeze" the dom. Then unfreeze once all elements have lazy loaded in. There is some slight movement with this strategy as the items are all loading in at different times and we try to correct the scroll position.

TODO (some of these we can push to after Hans PR):
- [ ] make it work without tags
- [ ] make it work with untagged models
- [ ] make it work with untagged webhooks 
- [ ] get rid of the slight bumpiness
- [ ] add some nice tests so it never breaks again

**Checklist**


amritk:
I will fix this right after hans PR is in

PR #6237, opened 2025-07-14
chore: update dependencies of `@scalar/api-reference`

hanspagel:
This PR
* updates all dependencies of `@scalar/api-reference`
* removes the broken `format:packages` script (uses internal CLI now)
* removes an unused dependency (`pkg-pr-new`)


amritk (approved):
Should we just link all of these to pnpm workspace? Then we only gotta update em once place

PR #6238, opened 2025-07-14
fix(api-reference): example responses show additional `value` key

### LINKED ISSUE #5064: Sample responses partially not correctly shown
### What happens?

The issue occures on the Scalar web UI on every shown request path. If you click on multiple different 4xx responses in a row the sample jsons are presented in a wrong way. The whole json is wrapped into an object called **value**.

For example the sample response
```json
{
  "type": "https://tools.ietf.org/html/rfc9110#section-15.5.2",
  "title": "Unauthorized",
  "status": 401,
  "detail": "An API key is required to access this resource.",
  "traceId": ""
}
```
turns after this bug into
```json
{
  "value": {
    "type": "https://tools.ietf.org/html/rfc9110#section-15.5.2",
    "title": "Unauthorized",
    "status": 401,
    "detail": "An API key is required to access this resource.",
    "traceId": ""
  }
}
```

Also after clicking multiple times between those response tabs the response sample should stay the same.

### OpenAPI Document

https://sandbox.scalar.com/e/kzT-0

### LINKED ISSUE #5435: Scalar add unnecessary json field named `value` to response example when using `components.responses`
### What happens?

In the response example, a json field named `value` appear when using `components.responses` with example.

Image

It shouldn't show the `value` field. For comparison I use Redoc.

Image

### OpenAPI Document

https://sandbox.scalar.com/e/00Agv

hanspagel:
**Problem**

In some cases, an additional `value` key shows up in the example responses after switching the tabs.

**Solution**

This PR fixes it. :)

Fixes #5435
Fixes #5064

I struggled to write a test for it, the code is ready to be refactored for sure. I’ll look into this on another day.

**Checklist**

### DIFF
--- a/.changeset/tall-comics-fry.md
+++ b/.changeset/tall-comics-fry.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
+fix: example response has an additional value key
--- a/packages/api-reference/src/features/example-responses/ExampleResponses.vue
+++ b/packages/api-reference/src/features/example-responses/ExampleResponses.vue
@@ -85,7 +85,10 @@ const getFirstExampleResponse = () => {
   }
 
   const firstProperty = Object.keys(currentJsonResponse.value.examples)[0]
-  return currentJsonResponse.value.examples[firstProperty]
+  const firstExample = currentJsonResponse.value.examples[firstProperty]
+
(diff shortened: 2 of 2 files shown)


antlio (approved):
nice!

amritk (approved):

PR #6247, opened 2025-07-15
chore: update vscode tailwind config

hwkr:
Updates the vs code config so we get tailwind support in functions and Vue `<transition>` classes.

**Checklist**

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -72,10 +72,17 @@
   "git.pullTags": false,
   "vue.complete.casing.props": "camel",
   "vue.complete.casing.tags": "pascal",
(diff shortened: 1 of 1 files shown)


.vscode/settings.json
```diff
@@ -72,10 +72,17 @@
   "git.pullTags": false,
   "vue.complete.casing.props": "camel",
   "vue.complete.casing.tags": "pascal",
-  "tailwindCSS.experimental.classRegex": [
```
hwkr, line 75: Sounds like this was removed from experimental

hanspagel (approved):
great!

PR #6248, opened 2025-07-15
fix(components): update add scalar classes to always watch for the headless root

hwkr:
It seems like sometimes Headless UI removes the portal root when the modal goes away which causes it to loose the `scalar-app` class. This updates the helper to watch for the portal root as long as the page is open and always add the `scalar-app` class.

I also removed `scalar-client` since we don't use it anymore for scoping and adds a bunch of tests to make sure the helper works.

**Checklist**


hanspagel (approved):
yess

PR #6255, opened 2025-07-15
fix(workspace-store): can’t load relative URLs

hanspagel:
**Problem**

Currently, the workspace store doesn’t load relative URLs. I think it’s because `isRemoteUrl` returns `false` for relative URLs.

**Solution**

This PR adds a relative URL example (`pnpm dev:reference`) to show the error.

⚠️ It doesn’t fix the issue yet.

Screenshot 2025-07-15 at 14 32 59

**Checklist**

### DIFF
--- a/.changeset/seven-onions-compete.md
+++ b/.changeset/seven-onions-compete.md
@@ -1,0 +1,5 @@
+---
+'@scalar/galaxy': patch
+---
+
+fix: broken internal link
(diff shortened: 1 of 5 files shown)


amritk:
@DemonHa @geoffgscott I tried to make the least invasive changes which hopefully doesn't affect anything else

hanspagel:
All tests pass, it’s desperately needed, so we should get it in. I just send it to @DemonHa to give it a look, though.

PR #6265, opened 2025-07-15
fix(api-reference): search bug with undefined operation on enter key

antlio:
**Problem**

currently hitting enter within api reference search modal in no result presence lead to an error getting displayed and breaking the experience.

**Solution**

this pr prevents navigation on enter in no result presence + updates style to enhance empty state element positioning.

| state | preview |
| -------|------|
| before | image |
| after | image |

**Checklist**

### DIFF
--- a/packages/api-reference/src/features/Search/SearchModal.vue
+++ b/packages/api-reference/src/features/Search/SearchModal.vue
@@ -167,6 +167,12 @@ function onSearchResultEnter() {
   }
 
   const results = searchResultsWithPlaceholderResults.value
+
+  // Prevents the user from navigating if there are no results
+  if (results.length === 0) {
+    return
+  }
+
   onSearchResultClick(results[selectedSearchIndex.value])
 }
 </script>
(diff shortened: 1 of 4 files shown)


packages/api-reference/src/features/Search/SearchModal.vue
```diff
@@ -167,6 +167,12 @@ function onSearchResultEnter() {
   }
 
   const results = searchResultsWithPlaceholderResults.value
+
+  // Prevents the user from navigating if there are no results
+  if (results.length === 0) {
+    return
+  }
+
   onSearchResultClick(results[selectedSearchIndex.value])
```
antlio, line 176: could also do 

```suggestion
  if (results.length) {
    onSearchResultClick(results[selectedSearchIndex.value])
  }
```

hanspagel: early returns (the first version) are great, it’s the only way to keep the structure flat :)

PR #6275, opened 2025-07-17
ci(docker): switch to original docker build push for aspire

xC0dex:
**Problem**

Currently, the `aspire-publish` randomly fails...

**Solution**

With this PR, we switch from blacksmiths action back to the "original" docker build push action.

**Checklist**


xC0dex:
Okay, this won't fix our issue https://github.com/docker/build-push-action/issues/621

PR #6278, opened 2025-07-17
feat(api-reference): request example block

hanspagel:
We reverted #6218 and now we’re good to get it in again.

So here‘s the 2nd attempt.


hanspagel:
Tested, and it’s awesome!

PR #6280, opened 2025-07-17
test(galaxy.scalar.com): `@scalar/mock-server` playground tests

hanspagel:
**Problem**

We shipped a broken bundle where you wouldn’t see operations if your document URL is a relative URL. It was visible on https://galaxy.scalar.com/, but none of our tests catched it.

**Solution**

With this PR we would have catched the broken release yesterday. It’s introducing two types of tests for the `@scalar/mock-server` playground (deployed to https://galaxy.scalar.com):

* tests for the Hono server
* e2e tests using the local JS bundle (local = what’s in the branch, not what’s on jsdelivr)

We run the same test expectations in a few environments (jsdelivr with json + yaml, local JS bundle with json + yaml and new: @scalar/mock-server playground), and I extended those expectations to look for the first tag and the first operation, too.

**Checklist**

- [  ] I've added a changeset (`pnpm changeset`).


amritk (approved):
hell ya

antlio (approved):
ready to go

PR #6282, opened 2025-07-17
chore: upgrade to Vite 6

hanspagel:
This PR upgrades Vite from 5 to 6. :) Not the latest, but still the greatest.

> there was a breaking change with the CSS file output names so be very careful to check ALL styles to make sure they are getting loaded. 

✅ packages/api-reference/dist/style.css
✅ packages/components/dist/style.css
✅ packages/themes/dist/style.css
✅ packages/api-client/dist/style.css
✅ packages/draggable/dist/style.css
✅ packages/api-reference-react/dist/style.css

~~I’ve had to force pnpm to use a newer esbuild version, because:~~

>  ~~pnpm-lock.yaml » esbuild@0.24.2 – esbuild enables any website to send any requests to the development server and read the response (moderate severity)~~

~~The security issue is not really relevant to us: No sensible user data is available through our development servers.~~

I’ve had to increase the level on which the vulnerability scan fails to high (instead of low or moderate). We don't have any sensible user data in our development environment, so those are not relevant to us.

### DIFF
--- a/examples/nextjs-api-reference/.eslintrc.cjs
+++ b/examples/nextjs-api-reference/.eslintrc.cjs
@@ -1,3 +1,7 @@
 module.exports = {
-  extends: ['plugin:@next/next/recommended'],
+  plugins: {
+    'eslint-plugin-react': 'latest',
+    'eslint-plugin-react-hooks': 'latest',
+    'eslint-plugin-next': 'latest',
+  },
 }


amritk:
what its working!?!

hanspagel:
yesss, needs a bit of testing, but should be good to go!

hanspagel:
too busy to merge this now and deal with potential build pipeline bugs, this is something for future Hans

packages/api-reference/index.html
```diff
@@ -19,6 +19,7 @@
     <!-- Initialize the Scalar API Reference -->
     <script type="module">
       Scalar.createApiReference('#app', {
+        hiddenClients: true,
```
hwkr, line 22: Did we mean to commit this? 😅

PR #6284, opened 2025-07-17
fix(api-reference): value in requeste example code block

amritk:
**Problem**

Currently, we fixed the value problem in response examples.

**Solution**

With this PR we fix it in the request examples. ALSO changed the example picker to a dropdown, no need for a combobox there :). 
| Before | After |
|--------|--------|
| image | image |
| image | image |

**Checklist**

### DIFF
--- a/.changeset/bright-numbers-report.md
+++ b/.changeset/bright-numbers-report.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
+fix: value in request example code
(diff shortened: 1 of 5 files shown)


antlio:
legit cook - great call on the combobox, how does it feel to keep the listbox to know what is selected?

amritk:
Yea I saw that but I think we should fix that in the components in a follow up. the `resize` prop should take the max width of any item, not just the size of the picker button

PR #6288, opened 2025-07-18
fix(api-reference): download link not reachable

antlio:
**Problem**

currently when trying to select second download link in some markdown description cases, click is not possible.

**Solution**

this pr increases the download link stacking to fix the issue and preserve click.

| state | preview |
| -------|------|
| before | image |
| after | image |

**Checklist**

### DIFF
--- a/.changeset/cyan-bears-prove.md
+++ b/.changeset/cyan-bears-prove.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
+fix: increases download link stackking to preserve clickability


hanspagel (changes requested):
gotta love a good z-index

you know what we should do? we have that tailwind z-index map, can we use that instead of the inline CSS value?

amritk:
@hwkr

antlio:
> gotta love a good z-index
> 
> you know what we should do? we have that tailwind z-index map, can we use that instead of the inline CSS value?
> 
> it’s `z-overlay` or something

as we are within a class, i didn't go that way but i can `@apply` at any moment

hanspagel:
I think the `@apply` method is gone in Tailwind 4, can we use the Tailwind class instead? I’m afraid the whole z-index system we came up with only works if we use it everywhere.

hwkr:
@hanspagel @antlio

> I think the `@apply` method is gone in Tailwind 4, can we use the Tailwind class instead?

[`@apply`](https://tailwindcss.com/docs/functions-and-directives#apply-directive) isn't gone in Tailwind 4 they've just changed it so you need to use it in conjunction with [`@reference`](https://tailwindcss.com/docs/functions-and-directives#reference-directive) when used in something like a Vue component. This is actually pretty nice because you can trace where the config is coming from rather than making it global.

For an example of it in action check out [ScalarTooltip](https://github.com/scalar/scalar/blob/main/packages/components/src/components/ScalarTooltip/ScalarTooltip.vue#L53-L67) which uses `@apply` to apply classes from our Tailwind config without them being scoped in `.scalar-app` like they normally would be.


> I’m afraid the whole z-index system we came up with only works if we use it everywhere.

Actually also no. Since`z-index`'s only take effect within their current [stacking context](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_positioned_layout/Stacking_context) it's actually pretty easy to have something with `z-index: 1` appear on top of something with `z-index: 9999` and using our Tailwind variables religiously won't help.

Here's a [useful example](https://codepen.io/web-dot-dev/full/JjErOXV) and some [docs](https://web.dev/learn/css/z-index).

Where our Tailwind variables come into play is for content we want to teleport to the root (usually `<body>`) like dropdowns, modals and popovers. We expose the variables in case we want to make dropdown like things outside of the component library.

**TL;DR; `z-index` is relative to the stacking context and not absolute to the page.**

**As a rule of thumb** I would recommend always using the lowest possible `z-index` (or ideally no `z-index`) to achieve the layering you want and to be aware of stacking context your setting that `z-index` in (for example, is it in the same stacking context as the sidebar? the header? is it going to overlap them?).

hwkr:
Looks like the culprit of this issue is actually a `position: relative;` on our markdown `<li>` (which was maybe being used to align the `::marker` but doesn't seem to be anymore).


Google Chrome-2025-07-22-13-54-45@2x


If we want to we can just remove it to fix the issue or we can set a `z-index: 1` on the download buttons (or both). IMO both are reasonable solutions.

antlio:
thank you so much for the thorough feedbacks!

> If we want to we can just remove it to fix the issue or we can set a z-index: 1 on the download buttons (or both). IMO both are reasonable solutions.

as discussed regarding this, as we do not have the hand on the description content (that can be callout, numbered list, etc) it won't be possible to update every markdown element to avoid the stacking issue i'm afraid

hwkr (approved):
I think `z-index: 1` makes sense here given the context :)

hanspagel:
Thanks @hwkr, awesome explanation! :) Sorry for making the one line change so hard. ✌️

hanspagel (approved):

PR #6294, opened 2025-07-18
fix(nuxt): update nuxt integration to use workspace store

amritk:
**Problem**

Currently, nuxt was usting ApiReferenceLayout directly.

**Solution**

With this PR we create the workspace store and use `ApiReferenceWorkspace` instead. Now this works but ssr and everything doesn't. We will add that when the `workspaceServerStore` is ready. I also removed the old `ssrState` (goodbye old friend) we will create something new when we do SSR week.

Another thing to note: any editors will still be broken but we should just not upgrade there until we sort out what we want to do with syncing the workspace store.

Also updated some packages as relative CI was reporting duplicates, I updated nuxt so may have caused something

**Checklist**


hanspagel (approved):
ok, cool

PR #6296, opened 2025-07-19
feat(api-reference): bring back lazy loading

amritk:
Just putting this up as draft as its nearly there. Once hans PR gets in I will probably make a new PR anyway as rebase will break. This is not really lazy loading v2, its more like v1 but getting rid of the buggy loading component.

TODO:
- [ ] scroll to markdown headings
- [ ] scroll to models (we will do proper lazy loading of models once we add parameter links)
- [ ] add tests

**Checklist**


amritk:
As previously stated will need to make a new PR!
image

PR #6303, opened 2025-07-21
feat(components): adds scalar copy button component

antlio:
**Changes**

this pr adds the `ScalarCopyButton` component as same behavior/style are used amongst reference/client and fixes in the meantime some style regarding background inheritance.

| state | preview |
| -------|------|
| before | image |
| after | image |

**Checklist**


hanspagel (commented):
nice! requires approval by @cameronrohani and @hwkr

cameronrohani (changes requested):
liking it a lot!

can we just remove the border as we chose to not have that on the copy buttons :)

packages/api-reference/src/v2/blocks/scalar-request-example-block/components/RequestExample.vue
```diff
@@ -254,7 +254,7 @@ const id = useId()
     class="request-card dark-mode"
     ref="elem">
     <!-- Header -->
-    <ScalarCardHeader class="pr-0.5">
```
hwkr: It think this was intentional because of the client selector @cameronrohani?

hwkr (changes requested):
Love the code consolidation 🙏  Thanks so much for taking a swing at this!!

First thing is a bugfix / tweak, it looks like the button doesn't have a background which means it can occlude the code snippet text. I chatted with @cameronrohani and he suggested making it `bg-b-2` to match how they were mocked up for docs.

Google Chrome-2025-07-22-15-09-18@2x


Second thing is for adding it to the component library. Normally a top level component should have a `*.stories.ts` associated with it but in this case I think this component is so closely associated with the code block it it should probably actually just live in the `ScalarCodeBlock` folder and maybe actually be named `ScalarCodeBlockCopy` or something like that. That way you can use the codeblock stories to demonstrate the functionality.

I also have a checklist for adding components to the library, not every component has all these things right now but I'm trying to make sure anything new that's added has everything on the list. I added a few notes specific to this PR


- [x] Exported from `@scalar/components`
- [x] Has JSDocs for all:
  - [x] Components (with examples) (see [`ScalarDropdown.vue`](https://github.com/scalar/scalar/blob/main/packages/components/src/components/ScalarDropdown/ScalarDropdown.vue#L1-L18) for a good example of a component doc block)
  - [x] Props
  - [x] Slots
  - [x] Emits
  - [x] Functions
  - [x] Types
- [ ] Allows overriding of Tailwind classes where applicable (see useBindCx)
- [ ] Has stories showcasing any applicable variants (can use the `ScalarCodeBlock` stories)
- [x] Has unit tests covering any applicable interactions (would be good to test both content being both a string and an object and mouse and keyboard interactions).

Thanks again for working on this! :)

packages/components/src/components/ScalarCopyButton/ScalarCopyButton.vue
```diff
@@ -0,0 +1,101 @@
...
+        :class="{ copied: showCopied }" />
+      <ScalarIconCheck
+        class="check-icon size-4"
+        :class="{ visible: showCopied }" />
+    </ScalarButton>
+  </div>
+</template>
+
+<style scoped>
+.scalar-code-block:hover .scalar-code-copy {
```
hwkr: This should probably be up in the `ScalarCodeBlock` component

packages/components/src/components/ScalarCopyButton/ScalarCopyButton.vue
```diff
@@ -0,0 +1,101 @@
+<script setup lang="ts">
+import { ScalarIconCheck, ScalarIconCopy } from '@scalar/icons'
+import { cva, cx } from '@scalar/use-hooks/useBindCx'
+import { useClipboard } from '@scalar/use-hooks/useClipboard'
+import { computed, ref } from 'vue'
+
+import { ScalarButton } from '../ScalarButton'
+
+const props = withDefaults(
```
hwkr: Vue encourages using [props destructuring](https://vuejs.org/guide/components/props#reactive-props-destructure) now instead of `withDefault`

packages/components/src/components/ScalarCopyButton/ScalarCopyButton.vue
```diff
@@ -0,0 +1,101 @@
...
+  defineProps<{
+    /** Content to copy to clipboard */
+    content: string | object
+    /** Optional aria label for accessibility */
+    ariaLabel?: string
```
hwkr: Elsewhere in the component library we just use `label` or `srLabel`. I've stopped writing `ariaLabel` in camelcase because then people use it on actual HTML elements and it doesn't work (it needs to be `aria-label`)

packages/components/src/components/ScalarCopyButton/ScalarCopyButton.vue
```diff
@@ -0,0 +1,101 @@
...
+    /** Size variant */
+    size?: 'sm' | 'md'
+  }>(),
+  {
+    ariaLabel: 'Copy content',
```
hwkr: Phrasing suggestion

```suggestion
    ariaLabel: 'Copy',
```

hwkr:
A couple of additional accessibility things I thought of while reviewing the code:

1. We probably need some sort of announcement when it's copied for non-sighted users. Could be an `sr-only` block with [`role="alert"`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/alert_role) or it could be a visible label with something similar.
2. If we really want to good about our a11y practices we should probably associate the button with the content (code block) that it copies via an [`aria-controls`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-controls) attribute.

packages/api-reference/src/v2/blocks/scalar-request-example-block/components/RequestExample.vue
antlio: i'll keep fixing it in #6310 since its not aligned

antlio:
thank you so much for the thorough review! diving into it!

>  looks like the button doesn't have a background which means it can occlude the code snippet text

went that way to inherit the background instead as we do not use the same one within reference/client cc @cameronrohani however it requires the codeblock to have one set to avoid transparency. should we just set one by default there instead?

packages/components/src/components/ScalarCopyButton/ScalarCopyButton.vue
antlio: good catch!

packages/components/src/components/ScalarCopyButton/ScalarCopyButton.vue
antlio: yes not against it!

antlio: actually this is not even enough for code input usage

antlio: removing copy button from code input from now on as not really accessible and only used in global environment within client

hwkr (changes requested):
Looks good! I pushed some tweaks for the id and stories (`nanoid` isn't stable for SSR I think)

It looks like the padding is breaking the scrollbars on the ResponseBodyRaw and also the button is getting hidden behind the codemirror instance so your can't click it

This PR:
Google Chrome-2025-07-23-17-11-01@2x


Prod:
Google Chrome-2025-07-23-17-11-16@2x

packages/api-client/src/views/Request/ResponseSection/ResponseBodyRaw.vue
```diff
@@ -2,8 +2,7 @@
 import { ScalarCodeBlockCopy } from '@scalar/components'
 import { prettyPrintJson } from '@scalar/oas-utils/helpers'
 import { useCodeMirror, type CodeMirrorLanguage } from '@scalar/use-codemirror'
-import { nanoid } from 'nanoid'
-import { ref, toRef } from 'vue'
+import { ref, toRef, useId } from 'vue'
```
antlio, line 5: good call ty!

antlio:
great catch + stacking issue, should be fix now let's go!

hwkr:
> looks good, but the copy button is above the scrollbar, is that what we want?

I think it's okay? cc @cameronrohani

packages/use-codemirror/src/hooks/useCodeMirror.ts
```diff
@@ -296,11 +296,11 @@ function getCodeMirrorExtensions({
     syntaxHighlighting(defaultHighlightStyle, { fallback: true }),
     EditorView.theme({
       '.cm-line': {
-        lineHeight: '20px',
+        lineHeight: '22px',
```
hwkr, line 299: This affects all the codeblocks yeah? Does it have any downstream effects?

antlio: this should follows recent font size increases, didn't catch any downstream yet

antlio: and only the codeblock using code mirror (so not in the response example)

antlio:
@hwkr good catch, fixed

cameronrohani (approved):

packages/api-client/src/components/CodeInput/CodeInput.test.ts
```diff
@@ -159,15 +158,6 @@ describe('CodeInput', () => {
...
 
-  it('copies content when copy button is clicked', async () => {
-    wrapper = createWrapper({ isCopyable: true })
-
-    const copyButton = wrapper.find('.copy-button')
-    await copyButton.trigger('click')
-
-    expect(vi.mocked(useClipboard).mock.results[0]?.value.copyToClipboard).toHaveBeenCalledWith(wrapper.vm.modelValue)
-  })
-
```
hanspagel, line 170: we should keep a test that makes sure there’s a way to copy the content, even if the actual button isn’t part of the component anymore

packages/api-reference/src/features/example-responses/ExampleResponses.test.ts
```diff
@@ -320,33 +320,6 @@ describe('ExampleResponses', () => {
...
-    })
-
-    const copyButton = wrapper.find('.copy-button')
-    expect(copyButton.exists()).toBe(true)
-
-    await copyButton.trigger('click')
-
-    // Since we can't test clipboard directly in jsdom, we can verify the click handler was called
-    expect(wrapper.emitted()).toBeDefined()
-  })
```
hanspagel: we should keep a test that makes sure there’s a way to copy the content, even if the actual button isn’t part of the component anymore

and for this component specifically, it would also be good to check **which** example response is copied

packages/api-client/src/components/CodeInput/CodeInput.test.ts
antlio: the test has been moved to the scalar code block test

antlio: but as it is not used anymore within code input, any reason why to keep it there?

hwkr:
Looks good! Going to move it to a draft pending release then we can get it in

amritk:
Lotta in depth reviews on this one! Whats the current status?

packages/api-client/src/components/CodeInput/CodeInput.test.ts
hanspagel: yeah, we still want to make sure there is _some way to copy the content_ when using the `CodeInput` 

we don’t need to test any edge cases or whatever, just to check that there’s a way to copy the content

hanspagel:
@amritk just one unresolved comment left 👍

packages/api-client/src/components/CodeInput/CodeInput.test.ts
antlio: there's no copy button within code input (removed in this PR)

antlio: as only being used for the client environment page - not really accessible in the view that will change and not needed now within collection environment

hanspagel: got it!

hanspagel:
Is that just a glitch on my machine? This is from the preview link (and it doesn’t happen on client.scalar.com):

https://ed028e40.client-staging-7m0.pages.dev/workspace/default/request/default

https://github.com/user-attachments/assets/9777054e-b5d8-464c-8653-a38f3e6e753d

antlio:
great catch @hanspagel i missed this - on it!

hanspagel (approved):
good to go 🫡

amritk:
@hwkr re-review?

PR #6304, opened 2025-07-21
feat(openapi-parser): partial bundle to a depth

DemonHa:
**Problem**

1. Currently, the bundling logic does not support limiting the depth of the bundle. This makes it impossible to perform a partial bundle of the document based on a desired level of nesting.

2. We are not doing any bundling on document ingress, which means we can have refs anywhere on the document

**Solution**

1. This PR introduces support for bundling up to a specified depth. This allows partial bundling of the document, which can be useful in cases where full resolution is unnecessary or too expensive.

Notes:
- Bundling with a depth limit means only a portion of the document will be resolved.

- **⚠️ Caution**: When using depth-limited bundling, previously visited parent nodes may cause issues in subsequent bundling operations. If you reuse the same node set and later attempt to bundle the same document without a depth limit, it may skip already visited nodes, resulting in incomplete resolution. If you're performing multiple bundle operations at different depths concurrently or sequentially, do not reuse the same hash set for tracking visited nodes, as this will cause partial or skipped bundling due to shared state.

2. We bundle the document on ingress if the input document is not preprocess by the server side store

**Checklist**


geoffgscott (approved):
LGTM

PR #6306, opened 2025-07-21
feat(api-reference): pragmatic lazy loading

amritk:
**Problem**

Currently, we just removed the lazy code from hans branch.

**Solution**

With this PR we add it back in WITHOUT the loading component, the way it works now is:
- we lazyily load all entries until the one we want + 2, this loads the one we want at the top of the screen super fast
- now we "freeze" the screen and let the rest of the elements load in, otherwise we would just be scrolled to the top and lose focus
- once everythings loaded, we unfreeze

TODO:
- [x] add some good tests
- [x] do something better with models/webooks

Punted:
- [ ] no time for virtual scroll as we got things to do but we should add that for large specs

Its not perfect, but we can fine tune it later. Lets get it out so we can get all the fixes out.

**Checklist**

### DIFF
--- a/packages/helpers/src/dom/freeze-at-top.test.ts
+++ b/packages/helpers/src/dom/freeze-at-top.test.ts
@@ -195,8 +195,8 @@ describe('freezeAtTop', () => {
     it('should handle empty string id', () => {
       const cleanup = freezeAtTop('')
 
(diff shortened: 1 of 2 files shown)


hanspagel (approved):
hell ya

works flawless

PR #6315, opened 2025-07-22
ci(hono): take screenshot

hanspagel:
**Problem**

Currently, …

**Solution**

With this PR …

**Checklist**


integrations/hono/package.json
```diff
@@ -49,7 +50,9 @@
     "@hono/zod-openapi": "^0.8.6",
     "@scalar/build-tooling": "workspace:*",
     "@scalar/openapi-to-markdown": "workspace:*",
-    "hono": "^4.6.5",
+    "hono": "^4.0.0",
```
hanspagel, line 53: wat

amritk:
oooo baby here we go, is this what I think it is

hanspagel:
would be nice, but I don’t finish that now

the required tooling is heavier than I would have liked it

PR #6317, opened 2025-07-22
fix(api-reference): can not render operation without responses

hanspagel:
EDIT: Tests are failing, need to look into that

**Problem**

```js
{
  openapi: '3.0.0',
  info: {
    title: 'Example',
    version: '1.0.0',
  },
  paths: {
    '/hello': {
      get: {
        summary: 'Hello World',
        description: 'Hello World',
        // Operation *without* responses
        // responses: {
        //   200: {
        //     description: 'Hello World',
        //   },
        // },
      },
    },
  },
}
```

Screenshot 2025-07-22 at 16 52 09

**Solution**

This PR deals with undefined responses.

And I’ve switched to the new workspace OpenAPI types. 👍 

We work with an upgraded document now, so we can remove all the code to support Swagger 2.0 or the old single `example` syntax from OpenAPI pre-3.1.

**Checklist**

### DIFF
--- a/packages/api-reference/src/features/example-responses/ExampleResponses.test.ts
+++ b/packages/api-reference/src/features/example-responses/ExampleResponses.test.ts
@@ -68,7 +68,7 @@ describe('ExampleResponses', () => {
     expect(wrapper.text()).toContain('Example 2')
   })
 
-  it('handles xml example response', () => {
+  it.only('handles xml example response', () => {
     const wrapper = mount(ExampleResponses, {
       props: {
         responses: {
@@ -76,7 +76,9 @@ describe('ExampleResponses', () => {
             description: 'XML response',
             content: {
               'application/xml': {
-                example: '<user><name>John</name><age>30</age></user>',
+                examples: {
+                  example1: { value: '<user><name>John</name><age>30</age></user>' },
+                },
               },
             },
           },
@@ -104,7 +106,9 @@ describe('ExampleResponses', () => {
             description: 'Plain text response',
             content: {
               'text/plain': {
-                example: 'Hello world',
+                examples: {
+                  example1: { value: 'Hello world' },
+                },
               },
             },
           },
@@ -132,7 +136,9 @@ describe('ExampleResponses', () => {
             description: 'HTML response',
             content: {
               'text/html': {
-                example: '<div>Hello <strong>world</strong></div>',
+                examples: {
+                  example1: { value: '<div>Hello <strong>world</strong></div>' },
+                },
               },
             },
           },
@@ -160,7 +166,9 @@ describe('ExampleResponses', () => {
             description: 'Success response',
             content: {
               'application/json': {
-                example: { status: 'success' },
+                examples: {
+                  example1: { value: { status: 'success' } },
+                },
               },
             },
           },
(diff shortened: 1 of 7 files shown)


amritk (changes requested):
We should use the new types, The old Openapi types = any

marclave (approved):
waiting on @DemonHa about types

DemonHa:
Yes, let's use the new types from the 'workspace-store' package. If there are some types missing the compilation will fail, which would be a good opportunity to improve missing types 😊

hanspagel:
That required a few more changes, pls review again 😅

packages/api-reference/src/features/example-responses/ExampleResponses.test.ts
```diff
@@ -68,15 +68,17 @@ describe('ExampleResponses', () => {
     expect(wrapper.text()).toContain('Example 2')
   })
 
-  it('handles xml example response', () => {
+  it.only('handles xml example response', () => {
```
DemonHa: you forgot to remove .only

hanspagel: right! a few tests still fail though, I wonder how it ever worked

packages/api-reference/src/features/example-responses/ExampleResponses.vue
```diff
@@ -26,17 +29,16 @@ import ExampleResponseTabList from './ExampleResponseTabList.vue'
  * TODO: copyToClipboard isn't using the right content if there are multiple examples
  */
 
-const { responses } = defineProps<{ responses: Operation['responses'] }>()
+const { responses } = defineProps<{
+  responses: Dereference<OperationObject>['responses']
```
amritk, line 33:
```suggestion
  responses: ResponsesObject
```

You can use the ResponsesObject directly here

packages/api-reference/src/features/example-responses/ExampleResponses.vue
```diff
@@ -45,12 +47,12 @@ const selectedResponseIndex = ref<number>(0)
 // Return the whole response object
 const currentResponse = computed(() => {
   const currentStatusCode =
-    orderedStatusCodes.value[selectedResponseIndex.value]
+    toValue(orderedStatusCodes)[toValue(selectedResponseIndex)]
```
amritk, line 50: is toValue better than .value?

packages/api-reference/src/features/example-responses/ExampleResponses.vue
```diff
@@ -67,42 +69,46 @@ const currentJsonResponse = computed(() => {
     normalizedContent?.['*/*'] ??
     // Take the first key - in the future we may want to use the selected content type
     normalizedContent?.[keys[0]] ??
-    // Swagger 2.0
-    currentResponse.value
+    undefined
```
amritk, line 72:
```suggestion

```
probably don't need that as it should already be undefined?

packages/api-reference/src/features/example-responses/ExampleResponses.vue
```diff
@@ -67,42 +69,46 @@ const currentJsonResponse = computed(() => {
...
   )
 })
 
+const hasMultipleExamples = computed<boolean>(
+  () =>
+    !!currentResponseContent.value?.examples &&
```
amritk, line 78:
```suggestion
    Boolean(currentResponseContent.value?.examples) &&
```
Do you like the way boolean looks?

amritk (approved):
Just code nit picks, nothing major, did not test!

marclave (approved):
nice : )

PR #6319, opened 2025-07-22
feat(workspace-store): support replacing the whole document

DemonHa:
**Problem**

In collaborative or live editing scenarios, we often receive a full version of a document without knowing which specific fields have changed. Currently, our store does not support updating the document in such a case without manually diffing or overwriting field by field.

**Solution**

This PR introduces a new method on the store: `replaceDocument`. This method replaces the current document with the provided version in an atomic operation. It ensures that the store remains consistent.
**Checklist**


hanspagel (approved):
outstanding

PR #6320, opened 2025-07-22
fix(code-highlight): cursor not visible in codeblock when first added

antlio:
**Changes**

this pr removes the inline block display utility set to `hljs` to prevent the caret from being hiddent when using codeblock input.

| state | preview |
| -------|------|
| before | image |
| after | image | 

**Checklist**


amritk:
Seems reasonable, @cameronrohani

hwkr (approved):

marclave (approved):

PR #6323, opened 2025-07-22
fix(api-reference): hiddenClients config + move clients to workspace store

amritk:
**Problem**

Currently, `hiddenClients` config option is not working correctly.

**Solution**

With this PR I do what I should have done before (but PR was too big) and just moved over to the workspace store for the client selector stuff.

**Checklist**


amritk:
closed in favor of 
https://github.com/scalar/scalar/pull/6324

PR #6324, opened 2025-07-22
fix(api-reference): hiddenClients config + move clients to workspace store

### LINKED ISSUE #6313: The configuration `hiddenClients` doesn't work
### What happens?

Currently it appears that the `hiddenClients` configuration is not working in the latest API Reference (`1.32.10`). Neither passing `true` nor specifying a particular client, such as `fetch`, appears to have any impact.

The configuration should work as before.

### OpenAPI Document

_No response_

amritk:
**Problem**

closes #6313

Currently, `hiddenClients` config option is not working correctly.

**Solution**

With this PR I do what I should have done before (but PR was too big) and just moved over to the workspace store for the client selector stuff.

| Scenario | Picture |
|--------|--------|
| x-scalar-sdk-installation | image |
| hiddenClients: true | image |
| hiddenClients: { ruby: true } | image | 

**Checklist**


hanspagel:
`hiddenClients` is so back

PR #6328, opened 2025-07-23
feat(api-reference): update the workspace store when the document changes

hanspagel:
I only added the failing test, the actual PR is authored by @amritk. 🤝 

**Problem**

When using the `'scalar:update-references-config'` event, or the `app.updateConfiguration` API, or updating the `configuration` prop of `<ApiReference />`, the new store doesn’t listen to changes to `content` or `url` anymore.

### Alright this is getting there, we can decide if we want the content property to be reactive and update automatically OR we want to only update when we explicitly call `app.updateConfiguration`

**Solution**

- [x] We listen to configuration changes and update the document in the store.
- [ ] Disable intersection observer before updating spec, maybe even freeze element as well

**Checklist**

### DIFF
--- a/packages/api-reference/src/v2/ApiReferenceWorkspace.vue
+++ b/packages/api-reference/src/v2/ApiReferenceWorkspace.vue
@@ -131,6 +131,9 @@ const addDocument = (config: Partial<ApiReferenceConfigurationWithSources>) => {
   // If the document is already in the store, we may want to update it
   if (store.workspace.documents[name]) {
     if (document) {
+      // Disable intersection observer to prevent url changing
+      isIntersectionEnabled.value = false
+
       /**
        * We need to clone the object as it gets mutated by the upgrader causing a stack overflow
        * todo: we can remove the deepClone once the upgrader becomes functional
@@ -140,6 +143,11 @@ const addDocument = (config: Partial<ApiReferenceConfigurationWithSources>) => {
 
       // Lets set it to active as well just in case the name changed
       store.update('x-scalar-active-document', name)
+
+      // Re-enable intersection observer
+      setTimeout(() => {
+        isIntersectionEnabled.value = true
+      }, 300)
     }
 
     return
--- a/packages/api-reference/src/v2/ApiReferenceWorkspace.vue
+++ b/packages/api-reference/src/v2/ApiReferenceWorkspace.vue
@@ -19,17 +19,19 @@ import {
   safeLocalStorage,
 } from '@scalar/helpers/object/local-storage'
 import { makeUrlAbsolute } from '@scalar/helpers/url/make-url-absolute'
-import { parseJsonOrYaml, redirectToProxy } from '@scalar/oas-utils/helpers'
-import type { AnyApiReferenceConfiguration } from '@scalar/types'
+import { redirectToProxy } from '@scalar/oas-utils/helpers'
+import type {
+  AnyApiReferenceConfiguration,
+  ApiReferenceConfigurationWithSources,
+} from '@scalar/types'
 import { useColorMode } from '@scalar/use-hooks/useColorMode'
 import { type WorkspaceStore } from '@scalar/workspace-store/client'
+import { deepClone } from '@scalar/workspace-store/helpers/general'
 import { useSeoMeta } from '@unhead/vue'
 import { useFavicon } from '@vueuse/core'
 import {
   computed,
   onBeforeMount,
-  onMounted,
-  onServerPrefetch,
   provide,
   ref,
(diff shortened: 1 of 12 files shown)


amritk:
This PR should enable this

https://github.com/scalar/scalar/pull/6319

amritk:
Alright I was messing around with this (nothing is final feel free to provide feedback). There's a stackoverflow right now with content in `replaceDocument => traverseDocument` somewhere in there, will sort it out tomorrow. BUT it generally works for URL's right now

packages/api-reference/src/v2/ApiReferenceWorkspace.vue
```diff
@@ -142,43 +154,19 @@ const addDocument = (config: (typeof configs.value)[number]) => {
...
-// props.configuration?.documents?.forEach((document) => {
-
-onServerPrefetch(() => {
-  // For SSR we want to preload the active document into the store
-  configs.value.forEach((config) => {
-    if (config.url) {
-      addDocument(config)
-    }
-  })
-})
```
geoffgscott, line 165: This should not be getting removed or we will have nothing rendered in SSR.

amritk: I don't believe we support SSR yet. Can you give me an example to test? 

Also since that function didn't have an await, there's no reason to use `onServerPrefetch`. It behaves the same way as the code we have now. Once we do SSR we would be using the server store so I think the code will be a bit different. We would probably handle all of the adding documents and what not elsewhere and just consume the store in this component. 

tldr: lets add test cases for whatever cases we need. Going to add tests to the ApiReferenceWorkspace.vue rn

packages/api-reference/src/v2/ApiReferenceWorkspace.vue
```diff
@@ -112,24 +118,45 @@ const configs = availableDocuments
...
+
+      /**
+       * We need to clone the object as it gets mutated by the upgrader causing a stack overflow
+       * todo: we can remove the deepClone once the upgrader becomes functional
+       */
+      const clonedDocument = deepClone(document)
+      store.replaceDocument(name, clonedDocument)
+
+      // Lets set it to active as well just in case the name changed
+      store.update('x-scalar-active-document', name)
```
hanspagel, line 148: wait, does this work well with having multiple documents and updating the content attribute? didn’t test, just thinking out loud

packages/api-reference/src/v2/helpers/get-document-name.ts
```diff
@@ -0,0 +1,45 @@
...
+    const info = document.info as InfoObject
+    if (info.title) {
+      return info.title
+    }
+  }
+
+  // If we cannot resolve a name, we iterate over the documens in case of a conflict
+  if (documents) {
+    const names = Object.keys(documents)
+    return iterateTitle('OpenAPI Document #1', (t) => names.some((name) => name === t))
```
hanspagel:
```suggestion
    return iterateTitle('API #1', (t) => names.some((name) => name === t))
```

When the dropdown is populated, and presented to the end user of the documentation, they probably think in APIs, not in documents.

packages/api-reference/src/v2/helpers/get-document-name.test.ts
```diff
@@ -0,0 +1,36 @@
...
+      )
+      expect(result).toBe('Test title')
+    })
+
+    it('should base the unknown name on the number of documents', () => {
+      const result = getDocumentName({}, {
+        'OpenAPI Document #1': {},
+        'OpenAPI Document #2': {},
+        'OpenAPI Document #3': {},
+        'OpenAPI Document #5': {},
```
hanspagel:
```suggestion
        'API #1': {},
        'API #2': {},
        'API #3': {},
        'API #5': {},
```

When the dropdown is populated, and presented to the end user of the documentation, they probably think in APIs, not in documents.

packages/api-reference/src/v2/ApiReferenceWorkspace.vue
geoffgscott: As long as we await document loading in the hooks we effectively support SSR at the workspace level at least. We can add a test with any document and just confirm that we get some HTML output from the document. @DemonHa is moving all the addDocument calls back to async pretty quickly anyways

packages/api-reference/src/v2/ApiReferenceWorkspace.vue
amritk: it will set the current one, currently thats decided by the useMultipleDoc hook

PR #6329, opened 2025-07-23
fix(fastapi): `pnpm dev` throws error

hanspagel:
**Problem**

I’ve recently added a `dev` and `test` script to the fastapi integration. The `dev` script is also executed when you run `pnpm dev` in root and fails if python or uvicorn isn’t available.

**Solution**

This PR checks whether python and uvicorn are available.

I’ve also had to increase the concurrency for `pnpm dev`, because I don’t have enough cores to run that many dev servers in parallel.

**Checklist**

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,7 +56,7 @@
     "clean:nuxt": "shx rm -rf \"**/.nuxt\"",
     "clean:dist": "shx rm -rf \"**/dist\"",
     "clean:turbo": "shx rm -rf \"**/.turbo\"",
-    "dev": "turbo dev --concurrency=100% --filter './examples/*' --filter './integrations/*' --filter=@scalar/nuxt --filter=@scalar/draggable --filter @scalar/components --filter @scalar/api-client --filter @scalar/nextjs-openapi",
+    "dev": "turbo dev --concurrency=20 --filter './examples/*' --filter './integrations/*' --filter=@scalar/nuxt --filter=@scalar/draggable --filter @scalar/components --filter @scalar/api-client --filter @scalar/nextjs-openapi",
     "dev:client": "turbo dev --filter @scalar/api-client",
     "dev:client:desktop": "turbo dev --filter scalar-app",
     "dev:client:app": "turbo playground:app --filter @scalar/api-client",
--- a/integrations/fastapi/package.json
+++ b/integrations/fastapi/package.json
@@ -16,8 +16,8 @@
     "node": ">=20"
   },
   "scripts": {
-    "dev": "cd playground && uvicorn main:app --reload",
-    "test": "python3 run_tests.py"
+    "dev": "command -v uvicorn >/dev/null 2>&1 && cd playground && uvicorn main:app --reload || echo '⚠️ uvicorn is not available, skipping the dev server.'",
+    "test": "command -v python >/dev/null 2>&1 && python run_tests.py || echo '⚠️ python is not available, skipping the tests.'"
   },
   "dependencies": {},
   "devDependencies": {}


package.json
```diff
@@ -56,7 +56,7 @@
     "clean:nuxt": "shx rm -rf \"**/.nuxt\"",
     "clean:dist": "shx rm -rf \"**/dist\"",
     "clean:turbo": "shx rm -rf \"**/.turbo\"",
-    "dev": "turbo dev --concurrency=100% --filter './examples/*' --filter './integrations/*' --filter=@scalar/nuxt --filter=@scalar/draggable --filter @scalar/components --filter @scalar/api-client --filter @scalar/nextjs-openapi",
+    "dev": "turbo dev --concurrency=20 --filter './examples/*' --filter './integrations/*' --filter=@scalar/nuxt --filter=@scalar/draggable --filter @scalar/components --filter @scalar/api-client --filter @scalar/nextjs-openapi",
```
amritk: the issue with 20 is it limits to 20 threads, what was wrong with 100%?

amritk: I think we had 20 before and it was breaking on machines that did not have 20 threads. Also 100% allows for running more apps

marclave: @hanspagel we should try to move these kinds of changes into another pr, cause the fix you did works! but this is a separate issue and we cant merge smth that isnt atomic.

geoffgscott (changes requested):
This approach to dev scripts is just messy. We don't need the fastapi integration to run from root dev. It should be tested on its own anyways. 

IMO the root dev script should go away. Packages need to be able to be run in isolation. This isn't like org where a whole bunch of services and databases need to be spun up.

package.json
hanspagel: I wrote it in the PR description:

> I’ve also had to increase the concurrency for pnpm dev, because I don’t have enough cores to run that many dev servers in parallel.

`20` is working fine if you have less than 20 cores, but `100%` is not working if you have less cores than scripts you try to run.

Sooo… without that it’s not working on my machine, so it’s OK to have it in the same PR I’d say.

hanspagel:
> IMO the root dev script should go away. Packages need to be able to be run in isolation. This isn't like org where a whole bunch of services and databases need to be spun up.

@marc uses it and asked us to fix it, that’s what I did. 🤡

amritk:
yea the thing its used for is QA, when you want to test everything at once!

package.json
amritk: ooooo really, I thought 100% was 100% of your threads, not 100% of the scripts
> Use 100% to use all available logical processors.

https://turborepo.com/docs/reference/run#--concurrency-number--percentage

amritk: Oh I got the error too, my poor 12 thread laptop

```
  × Invalid task configuration
  ╰─▶   × You have 17 persistent tasks but `turbo` is configured for concurrency of 12. Set --concurrency to at least 18
```

so maybe we split it into pnpm dev:packages, pnpm dev:integrations, pnpm dev:examples

amritk: either way lets boot it to another PR?

marclave:
> > IMO the root dev script should go away. Packages need to be able to be run in isolation. This isn't like org where a whole bunch of services and databases need to be spun up.
> 
> @marc uses it and asked us to fix it, that’s what I did. 🤡

yeah `pnpm dev` is in our contributing guides and its a fair thing to expect to run when going to a repo after `pnpm i`

so i think the fix is valid for getting it in but a different discussion on needing `pnpm dev` at root or what the scope of that is can be followed

hanspagel:
By popular demand: I’ve reset the concurrency setting. 😅

Note: It’s still broken for me (and for everyone with less than 18 cores).

PR #6330, opened 2025-07-23
fix(examples/web): Basic API Reference example does not load

hanspagel:
**Problem**

This examples uses the new store and the `configuration` to provide content, but it’s not actually wired up and the example doesn’t even load:

image

**Solution**

This PR just loads Scalar Galaxy from an URL (through the configuration and through the store).

Editing doesn’t work yet, it requires #6319. 😬 

**Checklist**


hanspagel:
not impacting users, but desperately needed to test PRs, so I skip the queue

PR #6292, opened 2025-07-18
feat(api-reference): updates heading anchor style

antlio:
**Changes**

this pr updates the api reference heading anchor to use latest icons.

| state | preview |
| -------|------|
| before | image |
| after | image |

**Checklist**

I've gone through the following:

- [x] I've added an explanation _why_ this change is needed.
- [x] I've added a changeset (`pnpm changeset`).
- [ ] I've added tests for the regression or new feature.
- [ ] I've updated the documentation.

PR #6331, opened 2025-07-23
feat(api-reference): updates heading anchor style

hanspagel:
New attempt, see #6292

PR is actually by @antlio 

**Changes**

this pr updates the api reference heading anchor to use latest icons.

`modern`
| state | preview |
| -------|------|
| before | image |
| after | image |

`classic`
| state | preview |
| -------|------|
| before | image |
| after | image |

**Checklist**

### DIFF
--- a/packages/api-reference/src/components/Anchor/Anchor.vue
+++ b/packages/api-reference/src/components/Anchor/Anchor.vue
@@ -1,4 +1,6 @@
 <script setup lang="ts">
+import { ScalarButton, useBindCx } from '@scalar/components'
+import { ScalarIconHash } from '@scalar/icons'
 import { useClipboard } from '@scalar/use-hooks/useClipboard'
 import { useId } from 'vue'
 
@@ -12,6 +14,8 @@ const { id } = defineProps<{
 
 const labelId = useId()
 
+const { cx } = useBindCx()
+
 const { copyToClipboard } = useClipboard()
 const { getHashedUrl } = useNavState()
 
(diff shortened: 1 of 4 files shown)


cameronrohani (approved):

hanspagel:
@antlio `pnpm dev:reference` classic layout looks like this for me:

Screenshot 2025-07-24 at 10 52 52

antlio:
> @antlio `pnpm dev:reference` classic layout looks like this for me:

i get this, can you re-build and confirm?

image

hanspagel:
Confirmed, looking great! I’ve had to empty the browser cache for whatever reason!?

packages/api-reference/src/hooks/useReferenceLayout.ts
```diff
@@ -0,0 +1,18 @@
...
+ */
+export type ReferenceLayout = 'modern' | 'classic'
+export const REFERENCE_LAYOUT_SYMBOL = Symbol() as InjectionKey<ReferenceLayout>
+
+/**
+ * Get the current API reference layout
+ *
+ * @see {@link ApiReferenceLayout}
+ */
+export const useReferenceLayout = () => ({
```
amritk: I believe the current practice we have for these stateful hooks is call with params for provide, then call with no params for inject. Right @hanspagel ?

But also I think we should do this sparingly as its annoying to test. Maybe we combine everything into the workspace store instead? Then we only need to worry about one store for testing instead of multiple things.

cc: @hanspagel @DemonHa 

tagging @DemonHa so we can come up with a strategy for adding these things into the workspace store. Probably just an extension on the document eh?

geoffgscott: Yeah this is a legacy way to do this. x-scalar-layout can be added to workspace JSON def and consumed. 

Props can be drilled down as needed. This scattered approach for state management is just not robust.

hanspagel: > Right @hanspagel ?

Right @amritk 👍

packages/api-reference/src/components/Anchor/Anchor.vue
```diff
@@ -21,60 +27,31 @@ const handleCopy = () => {
...
       <span>&ZeroWidthSpace;</span>
-      <button
+      <ScalarButton
         :aria-describedby="labelId"
-        class="anchor-copy"
-        type="button"
+        class="absolute left-0 inline-block h-fit px-1.5 py-1 opacity-0 group-hover/heading:opacity-100 group-has-focus-visible/heading:opacity-100"
+        :class="{
+          'top-1': layout === 'modern',
+        }"
```
hanspagel: ideally we get the CSS straight without relying on the `layout`, @antlio can you try if you see a generic solution that doesn’t require a conditional class?

the few places where we have those seems to have an even higher risk for UI regressions it seems

antlio: only ui regression or related to lack of testing both IMO so this won't change it but can look to rely on a class, that will still rely on the layout at the end of the day i think. will dig!

antlio (approved):

packages/api-reference/src/hooks/useReferenceLayout.ts
amritk: right!

amritk: move all things to the store

marclave (approved):

PR #6333, opened 2025-07-23
fix(api-reference): path routing not changing on scroll

amritk:
**Problem**

Currently, we don't re-enable the intersection observer

**Solution**

With this PR instead of using a timeout like we used to, we rely on the mutation observer to let us know when everything has loaded and we can re-enable the intersection observer.

**Checklist**


hanspagel (approved):
works great again! 🤟

amritk:
@hanspagel thats unrelated to this PR, but I can see where it comes from

amritk:
Fixed!

hanspagel (approved):
2 fixes in a single PR, doesn’t get better than this

PR #6337, opened 2025-07-23
fix(nuxt): passing the workspace store correctly to nuxt

amritk:
**Problem**

Couple things went wrong here:
- we were passing the workspace store the old way
- typescript was not working so it didnt catch it

**Solution**

- adding vue as a dependency fixed the type issue
- then it was trivial to fix the store issue

Will do a full revamp on nuxt once SSR is in play!

**Checklist**

### DIFF
--- a/.changeset/dirty-turtles-pull.md
+++ b/.changeset/dirty-turtles-pull.md
@@ -1,0 +1,5 @@
+---
+'@scalar/nuxt': patch
+---
+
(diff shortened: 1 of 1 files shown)


hanspagel:
Let me e2e test this though
Great work

i can do this in separate PR

hanspagel (approved):
works great!

I’m a bit too busy to add tests in this PR, but I’ve got it on my list

PR #6339, opened 2025-07-23
feat(api-reference): use path routing base path for relative document url resolution

### LINKED ISSUE #6335: Workspace store resolving relative URL to document when path routing
### What happens?

When you have a relative path to the document say `examples/openapi.json` and you have path routing enabled. If you start on a sub path like `/basePath/subPath1/subPath2` the document will try to resolve from the last slash. In this case it will resolve to `/basePath/subPath1/examples/openapi.json`

We expect it to start from the `pathRouting.basePath`, it should be `/basePath/examples/openapi.json`.

We just need to decide where we want to do this, I'm thinking ApiReferenceWorkspace as we need to know if we are pathRouting and the basePath. One caveat is we'll probably need to pass in the base server URL for SSR, we can deal with that later and just use window.location for now

### OpenAPI Document

_No response_

amritk:
**Problem**

closes #6335

Currently, we resolve a relative document from whatever path we are at, could be `/basepath/tags/endpoints/etc/openapi.json`. 

**Solution**

With this PR we use the `pathRouting.basePath` to resolve a relative document URL.

Also re-wrote make-url-absolute, previously it was doing: `http://localhost:3000` => `http://`

**Checklist**


hanspagel (commented):
this needs some good (manual) testing
everytime we change how we construct URLs we break some edge cases 😅

something for later-today-hans I guess

amritk:
@hanspagel yea it had tests so I assumed as long as it passed the test, we must make sure all cases are in the tests now!

hanspagel (approved):
it’s great, let’s get it in!

PR #6340, opened 2025-07-23
fix(api-reference): add custom-scrollbar style to external docs

amritk:
**Problem**

Currently, we forgot to add the `.custom-scroll` class

**Solution**

With this PR we add the class.

| Before| After |
|--------|--------|
| image | image |

**Checklist**


hanspagel (changes requested):
> Can we please detect if they overflow the container and change them to tooltips

@cameronrohani

PR #6341, opened 2025-07-23
fix(api-reference): set the default client lib to shell/curl

amritk:
**Problem**

Currently, we had the default set in the request example blocks but not the top.

**Solution**

With this PR we set the default at the top as well. It must be done this way because in the "unset" state we show the custom examples in the operations.

| Before | After | 
|--------|--------|
| image | image |

**Checklist**

### DIFF
--- a/.changeset/old-fishes-relax.md
+++ b/.changeset/old-fishes-relax.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
+fix: select shell/curl as default client lib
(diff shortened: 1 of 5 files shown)


hanspagel (changes requested):
are we using the `defaultHttpClient` option, too?

https://github.com/scalar/scalar/blob/main/documentation/configuration.md#defaulthttpclient-httpclientstate

hanspagel (approved):
AH, we do already! :) nice

PR #6342, opened 2025-07-24
fix(api-reference): path parameters missing from operation

amritk:
**Problem**

Currently, we do not show path parameters.

**Solution**

With this PR we combine path parameters with parameters.

Tests inc...

**Checklist**


hanspagel (approved):
nice

PR #6343, opened 2025-07-24
test(api-reference): retry flaky React tests 3 times

hanspagel:
**Problem**

The React tests fail every now and then, it seems to be a timing issue.

**Solution**

This PR allows Vitest to retry the tests 3 times, maybe that helps already.

**Checklist**


amritk:
Should we look into why its a timing issue?

marclave:
> Should we look into why its a timing issue?

yeah i think we should but at least for now let's add this and @hanspagel wanna make a ticket

PR #6345, opened 2025-07-24
fix(api-reference): does not render oneOf/allOf schemas with root properties

hanspagel:
**Problem**

If a schema has `oneOf` or `allOf` and some root properties, we only render the content of the composition schemas and omit the root properties.

Screenshot 2025-07-24 at 14 16 19

**Solution**

…

**Example**

```js
{
  openapi: '3.0.0',
  info: {
    title: 'Hello World',
    version: '1.0.0',
  },
  paths: {
    '/example': {
      post: {
        requestBody: {
          content: {
            'application/json': {
              schema: {
                type: 'object',
                properties: {
                  id: {
                    type: 'string',
                  },
                  name: {
                    type: 'string',
                  },
                },
                oneOf: [
                  {
                    required: ['id'],
                  },
                  {
                    required: ['name'],
                  },
                ],
              },
            },
          },
        },
      },
    },
  },
}
```

**Checklist**


packages/api-reference/src/components/Content/Schema/helpers/merge-all-of-schemas.ts
```diff
@@ -82,11 +90,11 @@ function mergeProperties(existingProps: SchemaObject, newProps: SchemaObject, de
       if (value.type === 'array' && value.items?.allOf) {
         merged[key] = {
           ...value,
-          items: mergeAllOfSchemas(value.items.allOf, depth + 1),
+          items: mergeAllOfSchemas(value.items.allOf, undefined, depth + 1),
```
amritk, line 93: nit, but usually when this happens I switch to   an object instead

PR #6346, opened 2025-07-24
fix(api-client): authentication description breaks container

antlio:
**Changes**

this pr sets clamp display to the request auth label to better handle long content display as seen below.

| state | preview |
| -------|------|
| before | image |
| after | image |

**Checklist**

### DIFF
--- a/.changeset/silly-zebras-hug.md
+++ b/.changeset/silly-zebras-hug.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
+
+fix: adds request auth tab label clamp display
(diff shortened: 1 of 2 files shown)


amritk (commented):
I like it but can we add hover over tooltip?

antlio:
> I like it but can we add hover over tooltip?

what do you mean? just replicated what we have in the current markdown usage, however we had some user using label as the description which was breaking as seen in the preview

marclave:
@antlio the question is when we hover it what happens? : ) we want the full description. but lets get this in for now

antlio:
> @antlio the question is when we hover it what happens? : ) we want the full description. but lets get this in for now

it shows the full content as done previously for auth description, server description, as seen below:

image

PR #6347, opened 2025-07-24
fix(components): broken markdown lists with headers

antlio:
**Changes**

this pr updates scalar markdown display in order to keep spacing between element and align `p` within `li`.

| state | preview |
| -------|------|
| before | image |
| after | image |

**Checklist**


amritk (approved):
LGTM - one question, is that bottom part duplicated?
image

antlio:
> LGTM - one question, is that bottom part duplicated? <img alt="image" width="541" height="903" src="https://private-

will check the spec but doesn't seem so, however that listing indent looks off

packages/components/src/components/ScalarMarkdown/ScalarMarkdown.vue
```diff
@@ -83,6 +83,7 @@ const html = computed(() => {
     --markdown-spacing-md: 16px;
     --markdown-line-height: 1.625;
 
+    display: contents;
```
hwkr: Are we sure this won't break things elsewhere? It makes sense but I think we're already using `ScalarMarkdown` lots of places and removing it from layout might cause unexpected side affects

packages/components/src/components/ScalarMarkdown/ScalarMarkdown.vue
```diff
@@ -136,7 +137,7 @@ const html = computed(() => {
   .markdown p {
     color: inherit;
     line-height: 22px;
-    display: block;
+    display: inline;
```
hwkr: `<p>` should probably still be `block` by default, maybe we should just target `li > p`?

I feel like in fact our parser shouldn't be inserting `<p>` tags in our `<li>`s but if it's going to we probs only want to target those.

On another note does anyone know where this styling comes from headings in `<li>`'s? It seems to me we should just be parsing it to an inline `<strong>` tag

hwkr: Okay so I put it in a gist to see how Github parsed it and they _do_ seem to insert a `<p>` tag. 

https://gist.github.com/hwkr/8652ec2a8b15a1b8478034a8262c530c

They keep their display as `block` though so let's maybe just look at how they're doing it and try to match that?

Google Chrome-2025-07-24-17-18-06@2x

hwkr (approved):
Let's try and switch the display back to block if we can?

packages/components/src/components/ScalarMarkdown/ScalarMarkdown.vue
hanspagel: agree, `p { display: inline }` seems super dangerous, let’s avoid this

packages/components/src/components/ScalarMarkdown/ScalarMarkdown.vue
antlio: i checked as much as i could and didn't find any issues, can you give this a try?

packages/components/src/components/ScalarMarkdown/ScalarMarkdown.vue
antlio: i took another way there in order to avoid having to do any of this and update the style accordingly

packages/components/src/components/ScalarMarkdown/ScalarMarkdown.vue
antlio: removed as it was indeed breaking some current styling within table

hwkr (approved):
Looks good! cc @cameronrohani on the spacing

Google Chrome-2025-07-25-10-27-15@2x

cameronrohani (approved):

PR #6353, opened 2025-07-24
fix(themes): don't style `<a>` in reset only in markdown

hwkr:
Somehow I missed this in the original PR it was in but we really just want the reset to strip away the base styles not apply style (we use `<a>`'s for all kinds of buttons and stuff).

**Checklist**


cameronrohani (approved):
you dropped this
image

marclave (approved):
<3

PR #6354, opened 2025-07-24
feat(api-reference): convert link list to icons when it scrolls

amritk:
**Problem**

Currently, the link list may need to scroll

**Solution**

With this PR we reduce it to icons only if scroll is needed.
| | Preview |
|--------|--------|
| Before | image |
| After | image | 

**Checklist**


cameronrohani:
you dropped this
image

Issue #6355, opened 2025-07-24
String descriptions overlap


ventocis:
### What happens?

The `min` and `max` field for strings can be misleading in two ways:
* When the field is a number that is formatted as a string (ex: `"2"`) the user may think that it's indicating a maximum value for the number, when it's actually indicating the maximum string length.
* Additionally, if both `maximum` and `maxLength` or both `minimum` and `minLength` are specified, the UI will show `min` twice and `max` twice, making it impossible for the user to know what each property is referring to without inspecting the actual OpenAPI spec.

<img width="1128" height="208" alt="Image" src="https://github.com/user-attachments/assets/395e2850-4384-41bb-90a1-dfab0bd7f01a" />

### What did you expect to happen?

`maxLength` and `minLength` should be displayed in a way that ensures the user doesn't conflate them with `minimum` and `maximum`.

### OpenAPI Document

https://sandbox.scalar.com/p/0EK0e

antlio:
hey @ventocis thank you for that great feedback, will update this today!

PR #6356, opened 2025-07-24
feat(api-reference): better schema `minLength`/`maxLength` labels

ventocis:
**Problem**
The `min` and `max` field for `minLength` and `maxLength` can be misleading in two ways:
* When the field is a number that is formatted as a string (ex: `"2"`) the user may think that it's indicating a maximum value for the number, when it's actually indicating the maximum string length.
* Additionally, if both `maximum` and `maxLength` or both `minimum` and `minLength` are specified, the UI will show `min` twice and `max` twice, making it impossible for the user to know what each property is referring to without inspecting the actual OpenAPI spec.

Currently, …
This property definition
```
"message": {
    "type": "string",
    "minLength": 2,
    "minimum": 3,
    "example": "Hello, world!",
    "maximum": 8,
    "maxLength":10
}
```

results in this being displayed

image

**Solution**
image

With this PR …

I updated `maxLength` to be labeled with `max length` and `minLength` be labeled with `min length`

**Checklist**

- [X] I've added a changeset (`pnpm changeset`).
- [X] I've added tests for the regression or new feature.
- [X] I've updated the documentation (I didn't see any docs to update)

### DIFF
--- a/.changeset/fifty-toys-remember.md
+++ b/.changeset/fifty-toys-remember.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': minor
+---
+
+feat: Make the min & max descriptions more descriptive for strings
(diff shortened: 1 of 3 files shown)


ventocis:
I've created https://github.com/scalar/scalar/issues/6355 for this issue as well

hanspagel (approved):
dude, you’re smart. never thought about this. thanks for submitting the PR!

PR #6358, opened 2025-07-25
feat(api-client): standalone browser version

bparks:
**Problem**

Currently, all the examples for api-client assume an environment other than the browser, which means that the examples for integrating api-reference into *any* app don't translate to api-client as well.

**Solution**

With this PR, you should be able to embed the api-client into any page the same way you can the api-reference. Ideally, you can also populate it with your API docs, similar to the [Scalar Editor](https://docs.scalar.com/editor).

**Checklist**

### DIFF
--- a/packages/api-client/src/layouts/Web/create-api-client-web.ts
+++ b/packages/api-client/src/layouts/Web/create-api-client-web.ts
@@ -1,5 +1,5 @@
 import { createApiClient } from '@/libs'
-import { createWebHistoryRouter, saveActiveWorkspace } from '@/router'
+import { createWebHashRouter, createWebHistoryRouter, createModalRouter, saveActiveWorkspace } from '@/router'
 import type { ApiClientConfiguration } from '@scalar/types/api-reference'
 
 import ApiClientWeb from './ApiClientWeb.vue'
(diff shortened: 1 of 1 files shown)


amritk:
Hey this is awesome! I _almost_ did this once 😁. I'll check it out tomorrow!

packages/api-client/vite.standalone.config.ts
```diff
@@ -0,0 +1,69 @@
...
+    // Source: https://github.com/highlightjs/highlight.js/blob/b9ae5fea90514b864f2c9b2889d7d3302d6156dc/tools/build_config.js#L58-L73
+    terserOptions: {
+      format: {
+        max_line_len: 80,
+      },
+    },
+    lib: {
+      entry: ['src/standalone.ts'],
+      name: '@scalar/api-client',
+      formats: ['umd'],
```
geoffgscott, line 57: Would like to avoid any UMD bundling (references will hopefully be moving away from this soon). We can do an ES build and still have a standalone.

bparks: No objections here, but — as I'm relatively new to vite — can you suggest an alternative? I just copied this from the standalone build for the `api-reference` package.

geoffgscott:
Appreciate the PR @bparks. We are currently overhauling the API client interface and internals with an estimated completion of Nov. 7th for the initial release. Will close this for now and we can revisit after those changes.

PR #6359, opened 2025-07-25
feat(components): add support for custom combobox items

hwkr:
Adds support for custom combobox items and fixes the undefined issue if options is `[]`.

Specifically `options` (and the `v-model`) are now a generic that you can extend with additional properties then the extended option type is handed back to you in a dynamic slot to use in custom styling / behavior.

I removed the `isDeletable` code which was only being used for the request auth combobox and used the new generics and option template to recreate the existing behavior on the consumer end without needing the additional props on the core component that we're only using in the one place.

The custom slots will also be used in the dashboard onboarding for the GitHub repo combobox so we can show the images for the repo org in the list ✨ I'll probably add generic item support for the listbox too but I figured this was a good start.

[Example of custom options and group labels](https://8a0050e3-20e4-42df-86ff-1cedf5e42146--scalar-components.netlify.app/?path=/story/components-scalarcombobox--with-slots)

Check out the API Client preview to see the auth dropdown (it should look almost exactly the same but I added a hover effect to the trash icon).

**Checklist**


hwkr:
The playwright tests seem to be passing locally for me?
Hyper-2025-07-25-18-16-44@2x

packages/api-client/src/views/Request/consts/new-auth-options.ts
```diff
@@ -1,17 +1,15 @@
 import type { SecurityScheme, SecuritySchemePayload } from '@scalar/oas-utils/entities/spec'
 import type { Entries } from 'type-fest'
+import type { ScalarComboboxOptionGroup } from '@scalar/components'
 
 export type SecuritySchemeOption = {
-  id: string
+  id: SecurityScheme['uid']
```
hwkr, line 6: We need to type this because the brand the type with zod cc @amritk

packages/api-client/src/views/Request/RequestSection/RequestAuth/RequestAuth.vue
```diff
@@ -118,7 +120,7 @@ const authIndicator = computed(() => {
  * Currently selected auth schemes on the collection, we store complex auth joined by a comma to represent the array
  * in the string
  */
-const selectedSchemeOptions = computed(() =>
+const selectedSchemeOptions = computed<SecuritySchemeOption[]>(() =>
```
hwkr, line 123: This needed to be typed so the generic type inference on the modelValue would work

packages/api-client/src/views/Request/RequestSection/RequestAuth/RequestAuth.vue
```diff
@@ -288,6 +289,24 @@ const openAuthCombobox = (event: Event) => {
...
               class="size-3 shrink-0 transition-transform duration-100 group-aria-expanded/combobox-button:rotate-180" />
           </ScalarButton>
+          <template #option="{ option, selected }">
+            <ScalarListboxCheckbox
+              :selected="selected"
+              multiselect />
+            <div class="min-w-0 flex-1 truncate">
+              {{ option.label }}
+            </div>
+            <ScalarIconButton
```
hwkr, line 299: This used to be a clickable icon but technically a button is more semantically correct

packages/components/src/components/ScalarCombobox/ScalarCombobox.vue
```diff
@@ -1,66 +1,71 @@
-<script setup lang="ts">
+<!-- prettier-ignore-attribute generic -->
```
hwkr, line 1: If we don't do this prettier wrap the generic attribute and breaks biome 🤦‍♀️

packages/components/src/components/ScalarCombobox/ScalarComboboxOptions.vue
```diff
@@ -37,37 +38,37 @@ function getOptionId(option: Option) {
...
 /** An list of all groups */
-const groups = computed<OptionGroup[]>(() =>
-  isGroups(props.options)
-    ? props.options
-    : [{ label: '', options: props.options }],
+const groups = computed<G[]>(
+  () =>
+    isGroups(props.options)
+      ? props.options // G extends OptionGroup<O>
+      : [{ label: '', options: props.options } as G], // G is OptionGroup<O>
```
hwkr: We know `G` is an  unextended `OptionGroup<O>` here because of the structure of the component props but I wasn't able to convince Typescript of that without using an `as` (not the end of the world just kinda annoying)

packages/components/src/components/ScalarCombobox/ScalarComboboxOptions.vue
```diff
@@ -79,32 +80,31 @@ watch(
...
-const filtered = computed<Option[]>(() =>
+const filtered = computed<O[]>(() =>
   query.value === ''
     ? options.value
     : options.value.filter((option) => {
         return option.label.toLowerCase().includes(query.value.toLowerCase())
       }),
 )
 
-const selected = computed<Option[]>({
```
hwkr, line 95: We don't need this anymore because `defineModel` is awesome

packages/components/src/components/ScalarListbox/ScalarListbox.vue
```diff
@@ -83,7 +83,7 @@ const { cx } = useBindCx()
                 v-for="option in options"
                 :key="option.id"
                 :option="option"
-                :style="multiple ? 'checkbox' : 'radio'" />
```
hwkr, line 86: I guess when I made this I forgot `style` is an html attribute 🤦‍♀️

hwkr:
Converting to draft until we release OSS

hwkr:
Undrafting! Should be ready for review

packages/components/src/components/ScalarCombobox/ScalarCombobox.vue
amritk: fix yer shit prettier

packages/components/src/components/ScalarCombobox/ScalarComboboxOptions.vue
geoffgscott: Maybe just add a code comment that clarifies this.

geoffgscott (approved):
Code looks good.

cameronrohani (approved):

PR #6368, opened 2025-07-25
feat(examples/web): add more sources to the deployed example

hanspagel:
Hey, I’ve got tons of ideas how to improve our manual and automatic testing setup, but here is a quick (and intermediate one):

It adds a few more sources to the deploy example, so you can quickly check popular OpenAPI files in the deploy example.


amritk (approved):
Its like I think about something, then @hanspagel already has a PR up for it.

PR #6373, opened 2025-07-25
fix(api-reference): update the hash on scroll for classic layout

amritk:
**Problem**

Currently, this actually never worked before either.

**Solution**

With this PR we now update the hash on scroll on the classic layout.

**Checklist**


marclave (approved):
im a huge fan of `safeLocalStorage`

PR #6374, opened 2025-07-25
chore(repo): cleanup tsconfigs

geoffgscott:
Removes the old secondary tsconfigs. 

Will enable stricter rules in a follow up.


amritk (approved):
The strictnessssss

PR #6376, opened 2025-07-25
feat: enable stricter tsconfig

geoffgscott:
Adds the rules from the previous strict config. Two rules require additional work before enabling in api-reference

### DIFF
--- a/packages/components/src/components/ScalarCombobox/ScalarCombobox.vue
+++ b/packages/components/src/components/ScalarCombobox/ScalarCombobox.vue
@@ -31,8 +31,6 @@ defineSlots<{
 }>()
 
 function handleUpdateModelValue(value: Option | undefined) {
-  close()
-
   if (value) {
     emit('update:modelValue', value)
   }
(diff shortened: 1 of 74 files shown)


packages/components/src/components/ScalarCombobox/ScalarCombobox.vue
hwkr, line 1: I'm going to create a bunch of conflicts in these files in my combobox PR so I'll turn the rules on over there and fix them up ahead of time 👍

packages/components/src/components/ScalarCombobox/ScalarCombobox.vue
```diff
@@ -29,6 +29,14 @@ defineSlots<{
   /** A slot for contents after the combobox options */
   after?(props: SlotProps): unknown
 }>()
+
+function handleUpdateModelValue(value: Option | undefined) {
+  close()
```
hwkr: Also I think this might be calling close on the window not the slot

hwkr: Indeed this seems to break closing (the window doesn't close though for whatever reason)

geoffgscott: Great save. Auto link got me here.

PR #6380, opened 2025-07-25
fix(components): apply tooltip shadow without tailwind variant

hwkr:
Turns out we we're getting weird css generated by using our `dark:` Tailwind variant in an `@apply` and it was creating a warning from our minifiers.

This cleans it up while doing the same thing :)

Hyper-2025-07-25-16-17-26@2x

Hyper-2025-07-25-16-17-50@2x

**Checklist**

### DIFF
--- a/packages/components/src/components/ScalarTooltip/ScalarTooltip.vue
+++ b/packages/components/src/components/ScalarTooltip/ScalarTooltip.vue
@@ -63,6 +63,9 @@ useTooltip({
 :where(body) > .scalar-tooltip:before {
   content: '';
   inset: var(--scalar-tooltip-offset);
-  @apply absolute rounded bg-b-tooltip -z-1 backdrop-blur dark:shadow-border;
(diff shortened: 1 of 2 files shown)


packages/components/src/components/ScalarTooltip/ScalarTooltip.vue
```diff
@@ -63,6 +63,9 @@ useTooltip({
 :where(body) > .scalar-tooltip:before {
   content: '';
   inset: var(--scalar-tooltip-offset);
-  @apply absolute rounded bg-b-tooltip -z-1 backdrop-blur dark:shadow-border;
+  @apply absolute rounded bg-b-tooltip -z-1 backdrop-blur;
+}
+:where(body.dark-mode) > .scalar-tooltip:before {
```
hwkr, line 68: This is actually pretty much exactly what the `dark:` variant is supposed to do it just can't generate the css properly in the `@apply`

hanspagel (approved):
omg, thanks for looking into this

PR #6381, opened 2025-07-26
fix(openapi-parser): depend on @scalar/openapi-types

not-my-profile:
E.g. validate and dereference from the parser return the OpenAPI.Document type from @scalar/openapi-types. Without the dependency if you install the parser and call one of these functions TypeScript will infer the type of their returned schema to be any.


hanspagel (approved):
Good catch, thanks!

PR #6385, opened 2025-07-28
ci: pin action digests to latest tags

xC0dex:
**Problem**

Currently, …

**Solution**

With this PR …

**Checklist**


hanspagel (changes requested):
Ah, this is great! And the comment is updated by renovate?

Updates to GitHub Actions are high risk (access to secrets), so we need to look through all changes of all updated packages before merging this.

xC0dex:
> Ah, this is great! And the comment is updated by renovate?

Exactly. Renovate will update the hash and the comment. `# v1` will not be touched until a new major release is available. `# v1.2.3` will be updated to `# v1.2.4` if a new patch is available.

xC0dex:
> Updates to GitHub Actions are high risk (access to secrets), so we need to look through all changes of all updated packages before merging this.

I checked most of them and didn't find anything suspicious. But I can go through them once again and check for release notes and issues. I think the official ones should be straightforward.

xC0dex:
I went through all the changes of all actions. They look good to me.

However, I was unable to compare the changes of the `relative-ci/agent-action` action due to a large diff tree https://github.com/relative-ci/agent-action/compare/26338b6cffd0b7eef7136ebed5ff3ddc22cbe113..1707825cbfcc7452b2913d273414705415ae64d4

hanspagel (approved):
as discussed, awesome work, thanks!

PR #6391, opened 2025-07-28
feat(Scalar.Aspire): support publish of scalar aspire

### LINKED ISSUE #6387: Scalar.Aspire: `API_REFERENCE_CONFIG` is not published
### What happens?

Currently, when publishing the `Scalar.Aspire` service, the `API_REFERENCE_CONFIG` environment variable is not published. This variable is essential as it contains the configuration.

I'd expected that the configuration was published.

### OpenAPI Document

_No response_

xC0dex:
**Problem**

Currently, it is not possible to use the Aspire integration in published manifests because the `API_REFERENCE_CONFIG` variable is not published.

**Solution**

With this PR we switched from the `AfterEndpointsAllocatedAsync` event (which is not invoked during the publish step) to the `BeforeStartAsync` event. I think this shouldn't cause any problems because we don't access the endpoint URLs, so we don't have to wait for them.

Closes #6387 

**Checklist**

### DIFF
--- a/.changeset/brave-balloons-love.md
+++ b/.changeset/brave-balloons-love.md
@@ -1,0 +1,5 @@
+---
+'@scalar/aspire': minor
+---
+
+feat: support publish of scalar aspire


lvde0:
I would love to test this, if you tell me how :)

xC0dex:
> I would love to test this, if you tell me how :)

You could clone this PR and checkout the `integrations/aspire/playground/Scalar.Aspire.AppHost` project 🙂. I tested it locally with the `Aspire.Hosting.Docker` publisher. However, you have to remove the `user-service` temporarily because NPM currently doesn't work.

lvde0:
Tested this with the 0.2.0 release, but It doesn't seem to publish the required environment variables:

```
  apispec:
    image: "scalarapi/aspire-api-reference:0.2.0"
    environment:
      services__api__http__0: "http://api:${API_PORT}"
    expose:
      - "8080"
    networks:
      - "aspire"
```

Which results in:

> API Reference configuration is not provided. Please set the API_REFERENCE_CONFIG environment variable.

**Important: I updated to Aspire 9.4 today (including `Aspire.Cli`). Maybe that broke something again :/**

Perhaps @davidfowl could provide some guidance.

xC0dex:
Hi @lvde0,

It works on my end even with Aspire 9.4. The publish Scalar Aspire Playground looks like this:

```yml
  scalar:
    image: "scalarapi/aspire-api-reference:latest"
    environment:
      services__keycloak__http__0: "http://keycloak:8080"
      services__keycloak__management__0: "http://keycloak:9000"
      services__book-service__http__0: "http://book-service:${BOOK_SERVICE_PORT}"
      API_REFERENCE_CONFIG: "[{\"proxyUrl\":\"/scalar-proxy\",\"showSidebar\":true,\"hideModels\":false,\"hideTestRequestButton\":false,\"darkMode\":true,\"hideDarkModeToggle\":false,\"servers\":[{\"url\":\"http://book-service\",\"description\":\"book-service\"}],\"defaultHttpClient\":{\"targetKey\":\"csharp\",\"clientKey\":\"httpclient\"},\"authentication\":{\"preferredSecurityScheme\":[\"oauth2\"],\"securitySchemes\":{\"oauth2\":{\"flows\":{\"authorizationCode\":{\"x-scalar-client-id\":\"admin-cli\"}}}}},\"withDefaultFonts\":true,\"defaultOpenAllTags\":false,\"theme\":\"saturn\",\"_integration\":\"dotnet\",\"hideClientButton\":false,\"sources\":[{\"title\":\"book-service | v1\",\"url\":\"http://book-service/swagger/v1.json\"}],\"persistAuth\":false}]"
      CDN_URL: "https://cdn.jsdelivr.net/npm/@scalar/api-reference"
      DEFAULT_PROXY: "true"
    expose:
      - "8080"
    networks:
      - "aspire"
```

lvde0:
I can't find this "publish" playground. Can you share the code?

xC0dex:
I pushed the code here for test purposes: https://github.com/scalar/scalar/tree/feat/aspire-update-publish/integrations/aspire/playground/Scalar.Aspire.AppHost.
In this folder you can run `dotnet aspire publish -o artifacts`. to reproduce it 🙂

lvde0:
@xC0dex Thanks for sharing. Turns out I was calling `AddDockerComposeEnvironment()` _before_ adding the Scalar resource. Putting it after fixes the problem. I wonder what the recommended way is. Should order matter? Not sure.

davidfowl:
The order should not matter no. The scaler resource might be mutating resources in a way that can;t be observed by the docker compose environment.

xC0dex:
Quick update: The PR (https://github.com/scalar/scalar/pull/6547) changes the implementation from a hook to an environment callback. Now the order doesn't matter anymore! 🎉

PR #6393, opened 2025-07-28
feat: initial support for dev containers

xC0dex:
This PR introduces the initial `.devcontainer/devcontainer.json`, enabling:

- the usage of dev containers in VS Code
- the usage of GitHub Codespaces

Currently, the container only supports Node, which means that packages like the proxy (Go) or `Scalar.Aspire` (.NET) won't work. We can consider adding them later to this config or add dedicated dev containers for them.

Tested with the `pnpm dev:reference` command 👍 

**Checklist**

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -22,7 +22,7 @@
     }
   },
   "hostRequirements": {
-    "memory": "4gb",
+    "memory": "6gb",
     "cpus": 4
   }
 }
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -22,7 +22,7 @@
     }
   },
   "hostRequirements": {
-    "memory": "6gb",
+    "memory": "8gb",
     "cpus": 4
   }
 }
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -3,30 +3,40 @@
 {
   "name": "Node.js & TypeScript",
   // Or use a Dockerfile or Docker Compose file. More info: https://containers.dev/guide/dockerfile
-  "image": "mcr.microsoft.com/devcontainers/typescript-node:1-22-bookworm",
+  "image": "mcr.microsoft.com/devcontainers/base:ubuntu-22.04",
   "features": {
+    "ghcr.io/devcontainers/features/node:1": {
+      "version": "22",
+      "pnpmVersion": "10.5.2"
+    },
     "ghcr.io/devcontainers/features/dotnet:2": {
-      "version": "latest"
+      "version": "9.0"
     },
     "ghcr.io/devcontainers/features/go:1": {
-      "version": "latest"
-    },
-    "ghcr.io/devcontainers-extra/features/pnpm:2": {
-      "version": "latest"
+      "version": "1.24.5"
     }
   },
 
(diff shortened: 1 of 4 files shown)


packages/api-reference/vite.config.ts
```diff
@@ -11,6 +11,9 @@ export default defineConfig({
     'process.env.NODE_ENV': '"production"',
     'process.env.SCALAR_API_REFERENCE_VERSION': `"${version}"`,
   },
+  server: {
+    host: true,
```
xC0dex: To ensure port forwarding works properly in dev containers, we need to add this to the configuration. Reference: https://vite.dev/guide/troubleshooting.html#dev-containers-vs-code-port-forwarding

Do you have better ideas than adding it to every configuration?

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,28 @@
...
+        "streetsidesoftware.code-spell-checker",
+        "biomejs.biome",
+        "esbenp.prettier-vscode",
+        "bradlc.vscode-tailwindcss",
+        "vitest.explorer"
+      ]
+    }
+  },
+  "hostRequirements": {
+    "memory": "6gb",
```
xC0dex: 6 GB is the minimum. I tested with 5  GB, but some processes were randomly killed during the turbo build.

packages/api-reference/vite.config.ts
hanspagel: This also means someone in the coffee shop (same Wifi) will be able to access the Vite server. I think that’s okay, we don’t serve any sensible user data in the open-source repository.

Still: Can we find an more restrictive setting? Or maybe enable this only in Docker envs?

And if we have to add a setting here, can you add a command explaining why we need it?

.devcontainer/devcontainer.json
hanspagel: Make it 8gb then, otherwise we’ll run into strange memory issues in 6 months

hanspagel:
Pretty sweat! Can I test the codespace in this PR already?

xC0dex:
> Pretty sweat! Can I test the codespace in this PR already?

Yes. Just switch to this branch on GitHub und create a codespace out of it!

packages/api-reference/vite.config.ts
xC0dex: Let me check that 👍

hanspagel:
> Codespace could not be created: SKU name 'basicLinux32gb' is not allowed for this repository

😂 whatever that means

AH, we need to enable it

> Prebuilds are currently disabled because scalar has reached its Codespaces spending limit for this period. Please contact an administrator to enable prebuilds.

I’ve set the spending limit to $1 and the error is gone, but I still get this:

> Codespace could not be created: SKU name 'basicLinux32gb' is not allowed for this repository

packages/api-reference/vite.config.ts
xC0dex: I updated the condition and added a comment

hanspagel (changes requested):
Can’t test yet, because of

> Codespace could not be created: SKU name 'basicLinux32gb' is not allowed for this repository

I don’t know how to fix that yet, but we’ll figure it out.

hanspagel (approved):
tested, works! great work :))

PR #6363, opened 2025-07-25
fix(api-client): missing request auth border and empty state

antlio:
**Changes**

this pr handles in a better way unselected request auth in order to rely on the empty state + refactors style for header/description to fix missing borders, following up of #6346 

| state | preview |
| -------|------|
| before | image |
| before | image |
| after | image |

**Checklist**

I've gone through the following:

- [x] I've added an explanation _why_ this change is needed.
- [x] I've added a changeset (`pnpm changeset`).
- [ ] I've added tests for the regression or new feature.
- [ ] I've updated the documentation.

PR #6395, opened 2025-07-29
fix(api-reference):  request auth card displays both header and description

antlio:
**Changes**

this pr is a follow up of #6363 that omitted to maintain conditioning regarding the header/description which was displaying them both at the same time, and re-introduced it along some style fixtures.

| state | preview |
| -------|------|
| before | image |
| after | image | 

**Checklist**


packages/api-client/src/views/Request/RequestSection/RequestAuth/RequestAuthTab.vue
```diff
@@ -165,17 +165,26 @@ const dataTableInputProps = {
...
-      ">
+    <!-- Header -->
+    <DataTableRow v-if="security.length > 1">
       <DataTableCell
-        :aria-label="generateLabel(scheme!) || scheme?.description"
+        :aria-label="generateLabel(scheme!)"
+        class="text-c-2 group/auth flex items-center leading-[22px] whitespace-nowrap outline-none hover:whitespace-normal">
+        <p
+          class="bg-b-1 text-c-2 outline-b-3 top-0 z-1 h-full w-full overflow-hidden px-3 py-1.25 text-ellipsis group-hover/auth:absolute group-hover/auth:h-auto group-hover/auth:border-b *:first:line-clamp-1 *:first:text-ellipsis group-hover/auth:*:first:line-clamp-none">
+          {{ generateLabel(scheme!) }}
```
hanspagel: I wonder why we need the `!`, maybe we should add a `v-if="scheme"` somewhere higher, or make the function deal with an `undefined` input

amritk (approved):
yolo merge

PR #6396, opened 2025-07-29
fix(api-reference): misalignments in sidebar element icon and title

antlio:
**Changes**

this pr updates the api reference sidebar element style to align back the icon/title.

| state | preview |
| -------|------|
| before | image |
| after | image | 

**Checklist**


packages/api-reference/src/features/sidebar/components/SidebarElement.vue
```diff
@@ -167,7 +167,7 @@ const onAnchorClick = async (ev: Event) => {
   font-size: var(--scalar-small);
   font-weight: var(--scalar-sidebar-font-weight, var(--scalar-regular));
   word-break: break-word;
-  line-height: 1.385;
+  line-height: 22px;
```
hanspagel: just so we’re aware, that’s a fixed line height no matter what the font size is, is that what we want?

cameronrohani (approved):

PR #6286, opened 2025-07-17
feat(themes): try setting the base line height and font-size

hwkr:
**Checklist**

I've gone through the following:

- [ ] I've added an explanation _why_ this change is needed.
- [ ] I've added a changeset (`pnpm changeset`).
- [ ] I've added tests for the regression or new feature.
- [ ] I've updated the documentation.

PR #6397, opened 2025-07-29
fix(api-reference): wrong paragraph line height within markdown

antlio:
**Changes**

this pr brings back #6286 and updates adds tw utility usage within reference for text stytling. this pr fixes line height within markdown usage + overhaul text - some changes might apply vs previous state which should match latest figma updates.

| state | preview |
| -------|------|
| before | image |
| after | image | 

**Checklist**

### DIFF
--- a/packages/api-reference/src/components/Content/ClientLibraries/ClientLibraries.vue
+++ b/packages/api-reference/src/components/Content/ClientLibraries/ClientLibraries.vue
@@ -160,9 +160,12 @@ const installationInstructions = computed(() => {
   </div>
 </template>
 <style scoped>
+@reference "tailwindcss";
+
 .selected-client {
+  @apply text-sm leading-none;
+
   color: var(--scalar-color-1);
-  font-size: var(--scalar-small);
   font-family: var(--scalar-font-code);
   padding: 9px 12px;
   border-top: none;
--- a/packages/themes/src/base/reset.css
+++ b/packages/themes/src/base/reset.css
@@ -16,7 +16,7 @@
 :where(.scalar-app) {
   /* Apply styles that are normally applied to `html`. */
   font-family: var(--scalar-font);
-  line-height: 1.15;
+  line-height: 1;
   color: var(--scalar-color-1);
 
   -webkit-text-size-adjust: 100%;
--- a/packages/themes/src/base/variables.css
+++ b/packages/themes/src/base/variables.css
@@ -40,10 +40,10 @@
   --scalar-font-size-7: 10px;
 
   --scalar-line-height-1: 32px;
-  --scalar-line-height-2: 24px;
-  --scalar-line-height-3: 20px;
-  --scalar-line-height-4: 18px;
-  --scalar-line-height-5: 16px;
+  --scalar-line-height-2: 26px;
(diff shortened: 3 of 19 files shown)


antlio:
> I’m not smart enough to know, but shouldn’t we try to avoid `@apply` and instead keep using variables or refactor the relevant parts to use Tailwind utility classes?

this is using the variables set in `themes` packages under the hood, this allows to rely both on size/line-height by using one utility which feels easier - open to discuss!

hanspagel:
`.text-sm` is `font-size: var(--scalar-font-size-4);`, and I’m just wondering if should do one of the following:

1. use the `text-sm` class in the markup
2. or use `font-size: var(--scalar-font-size-4);` in the CSS

which *feels* easier to me, but that might be just me

antlio:
> `.text-sm` is `font-size: var(--scalar-font-size-4);`, and I’m just wondering if should do one of the following:

`text-sm` is

```css
.scalar-app .text-sm {
    font-size: var(--scalar-font-size-4);
    line-height: var(--tw-leading, var(--scalar-line-height-4));
}
```

discussed in slack but can stick to your approach, however coupling both help to make sure we are as close as possible to the design

hanspagel:
Ah, got it. I missed that, it’s just the `font-size` in the last release. Then I’d personally go with the utility class in the markup, but whatever @hwkr @amritk and you decide on. ✌️

hwkr:
I think the utility class in the markup is good.

hwkr:
Since this touches themes and the tailwind config pretty heavily I'm going to convert it back to a draft for now and do a more targeted fix. We can revisit post release.

packages/themes/src/base/variables.css
```diff
@@ -40,10 +40,10 @@
...
-  --scalar-line-height-3: 20px;
-  --scalar-line-height-4: 18px;
-  --scalar-line-height-5: 16px;
+  --scalar-line-height-2: 26px;
+  --scalar-line-height-3: 22px;
+  --scalar-line-height-4: 20px;
+  --scalar-line-height-5: 18px;
 
   --scalar-font-normal: 400;
   --scalar-font-medium: 500;
```
hanspagel, line 49: let’s not touch the themes

packages/themes/src/base/reset.css
```diff
@@ -16,7 +16,7 @@
 :where(.scalar-app) {
   /* Apply styles that are normally applied to `html`. */
   font-family: var(--scalar-font);
-  line-height: 1.15;
+  line-height: 1;
   color: var(--scalar-color-1);
 
   -webkit-text-size-adjust: 100%;
```
hanspagel, line 22: let’s not touch the reset

packages/api-reference/src/components/Content/ClientLibraries/ClientLibraries.vue
```diff
@@ -160,9 +160,12 @@ const installationInstructions = computed(() => {
   </div>
 </template>
 <style scoped>
+@reference "tailwindcss";
```
antlio, line 163:
```suggestion
@reference "../../style.css";
```

PR #6402, opened 2025-07-29
fix(components): make paragraph line height variable

hwkr:
Updates the markdown `<p>` line height to match the figma (more or less). We use a 16px font size for most of refs and we were hard coding the line height to 22px (supposed to be `26px`).

This just puts it back to using `1.625` which means at 16px font size we get `26px` line-height, however at `14px` is `22.75px` which isn't technically right (supposed to be 22px) but I think let's clean it up and figure out a more cohesive markdown line height approach once we get OSS released.

It's worth note since the font size is variable for markdown (and technically the rest of our app since it's themed?) we might need to move away from using pixel based line heights all together.

**Checklist**

### DIFF
--- a/packages/components/src/components/ScalarMarkdown/ScalarMarkdown.stories.ts
+++ b/packages/components/src/components/ScalarMarkdown/ScalarMarkdown.stories.ts
@@ -8,6 +8,9 @@ import markdownContent from './test.md?raw'
  */
 const meta = {
   component: ScalarMarkdown,
+  argTypes: {
+    class: { control: 'text' },
+  },
   tags: ['autodocs'],
   render: (args) => ({
     components: { ScalarMarkdown },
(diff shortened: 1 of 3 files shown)


packages/components/src/components/ScalarMarkdown/ScalarMarkdown.stories.ts
```diff
@@ -8,6 +8,9 @@ import markdownContent from './test.md?raw'
  */
 const meta = {
   component: ScalarMarkdown,
+  argTypes: {
+    class: { control: 'text' },
```
hwkr, line 12: This just let's us test font sizes more easily

cameronrohani (approved):

antlio:
FYI this is applying the wrong line height where markdown is set to a lower size (server description, request auth etc.) which makes one liner flicker on hover

PR #6404, opened 2025-07-29
fix(api-reference): reference theme and style fixes

hwkr:
**Problem**

Currently, …

**Solution**

With this PR …

**Checklist**


packages/api-reference/src/features/Operation/components/ParameterListItem.vue
```diff
@@ -74,12 +74,12 @@ const shouldShowParameter = computed(() => {
         v-if="shouldCollapse"
         class="parameter-item-trigger"
         :class="{ 'parameter-item-trigger-open': open }">
-        <ScalarIconCaretRight
-          weight="bold"
-          class="parameter-item-icon size-3 transition-transform duration-100"
-          :class="{ 'rotate-90': open }" />
         <span class="parameter-item-name">
-          {{ parameter.name }}
+          <ScalarIconCaretRight
```
hwkr, line 78: This moves the caret to be centered to the status code instead of the whole block

Before:
Image

packages/api-reference/src/components/Anchor/Anchor.vue
```diff
@@ -36,7 +36,7 @@ const handleCopy = () => {
       <span>&ZeroWidthSpace;</span>
       <ScalarButton
         :aria-describedby="labelId"
-        class="absolute top-1 left-0 inline-block h-fit px-1.5 py-1 opacity-0 group-hover/heading:opacity-100 group-has-focus-visible/heading:opacity-100"
+        class="absolute top-1/2 left-0 inline-block h-fit -translate-y-1/2 px-1.5 py-1 opacity-0 group-hover/heading:opacity-100 group-has-focus-visible/heading:opacity-100"
```
antlio, line 39: seems like this is breaking alignment for modern theme

Image

hwkr: Yeah we can adjust it to be more baseline in the future but it was breaking the other usages more (and centred isn't the worst)

PR #6405, opened 2025-07-29
fix(api-reference): add failsafe types for schema

amritk:
**Problem**

Currently, we aren't showing anything in certain cases of an `allOf` schema.

**Solution**

With this PR we show the object expand button button and have it labeled as object.

As a follow up we should auto expand this object as its the only one.

**Checklist**


hwkr (approved):
Looks good! Did have a chance to test yet

PR #6406, opened 2025-07-29
fix(api-reference): block quote font size

amritk:
**Problem**

Currently, the block quotes were too big

**Solution**

With this PR we drop em to 16px aka font-size-2

**Checklist**


hwkr (approved):
Should be all good 👍

PR #6407, opened 2025-07-29
fix(api-reference): show schema description from allOf properties

amritk:
**Problem**

Currently, we don't show the description from level 0 in schema.

**Solution**

With this PR we take the first description from allOf properties. A specific fix that we can later make more generic

**Checklist**

### DIFF
--- a/packages/api-reference/src/components/Content/Schema/Schema.vue
+++ b/packages/api-reference/src/components/Content/Schema/Schema.vue
@@ -340,7 +340,6 @@ button.schema-card-title:hover {
 }
 .schema-card-description {
   color: var(--scalar-color-2);
-  margin-top: 8px;
 }
 .schema-card-description + .schema-properties {
   width: fit-content;
--- a/packages/api-reference/src/components/Content/Schema/Schema.vue
+++ b/packages/api-reference/src/components/Content/Schema/Schema.vue
@@ -123,8 +123,12 @@ const schemaDescription = computed(() => {
 
 /** Determines whether to show the schema description */
 const shouldShowDescription = computed(() => {
-  // For allOf compositions, find the first schema with a description
-  if (schema.value?.allOf && Array.isArray(schema.value.allOf)) {
+  // For allOf compositions, find the first schema with a description, only for the request body (for now)
+  if (
+    schema.value?.allOf &&
+    Array.isArray(schema.value.allOf) &&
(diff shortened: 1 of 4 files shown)


packages/api-reference/src/components/Content/Schema/Schema.vue
```diff
@@ -314,6 +340,7 @@ button.schema-card-title:hover {
 }
 .schema-card-description {
   color: var(--scalar-color-2);
+  margin-top: 8px;
```
hwkr: Yo so this is probably adding margin places we don't want it

Image

amritk: whats the fix!

hwkr: I'll push it up

packages/api-reference/src/features/Operation/components/RequestBody.vue
```diff
@@ -156,6 +156,12 @@ const handleDiscriminatorChange = (type: string) => {
   font-size: var(--scalar-small);
   width: 100%;
 }
+
+.request-body-header
+  + .request-body-schema:has(> .schema-card > .schema-card-description) {
+  /** Add a bit of space between the heading border and the schema description */
+  padding-top: 8px;
```
hwkr, line 167: Okay try this @amritk

hanspagel (changes requested):
nice!

allOf (and others) can have a root-level description. We want to show the root-level description, because that’s often the one describing the combination of the other schemas (and not the individual schemas).

Here is an example:

```yaml
openapi: 3.0.0
info:
  title: Example
  version: "1.0"
paths: {}
components:
  schemas:
    MyExample:
      # Look at this bad boy here:
      description: Shows this description
      allOf:
        - type: object
          description: first object description
          properties:
            foo:
              type: string
        - type: object
          description: second object description
          properties:
            bar:
               type: number
```

✅ This shows in Swagger UI like this:

Screenshot 2025-07-31 at 11 45 37

❌ But in Scalar like this:

Screenshot 2025-07-31 at 11 46 06

amritk:
thanks @hanspagel added that case. I merged the two functions since they were basically the same. I know this one is just a quick fix but maybe in the future we could just show all the descriptions.

PR #6410, opened 2025-07-30
fix(components): scalar tooltip style inconsistency

antlio:
**Changes**

this pr updates the scalar tooltip style to be consistent with other tooltip like element usage.

| state | preview |
| -------|------|
| before | image image |
| after | image image | 

**Checklist**

### DIFF
--- a/.changeset/eighty-days-greet.md
+++ b/.changeset/eighty-days-greet.md
@@ -1,0 +1,6 @@
+---
+'@scalar/components': patch
+'@scalar/themes': patch
+---
+
+fix: scalar tooltip style inconsistency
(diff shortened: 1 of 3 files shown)


hwkr:
Hey! So that was actually intentional to different tooltips from "toggle tips".

* Tooltips: appear on hover, only contains inline content and has no internal interactions (e.g. links).
* ToggleTips: appear on click, can contain rich text content or interactive content.

From the original PR: https://github.com/scalar/scalar/pull/5665#:~:text=Swapped%20the%20tooltip%20for%20a%20popover%20for%20the%20request%20table%20because%20the%20info%20is%20pretty%20complicated%20and%20probably%20shouldn%27t%20be%20in%20a%20tooltip

For the tooltips @cameronrohani and I decided to match GitHub and then to use our popover styles for "toggle" tips.

antlio:
keep feeling while using the client that the mix of background + font weight with that size doesn't feel accessible / hard to read - but noted

PR #6414, opened 2025-07-30
feat(json-magic): new JSON tooling package

DemonHa:
Move the json tooling to this new package called `json-magic`

Submodules:
- bundler
- dereference (new module)
- diff
- magic-proxy

**Checklist**


packages/json-magic/src/utils/isJson.ts
geoffgscott, line 1: prefer filename of `is-json.ts` (same for all others that are similar)

packages/json-magic/src/utils/isObject.ts
```diff
@@ -0,0 +1,4 @@
+/**
+ * Check if the given value is an object
```
geoffgscott: Let's expand this comment to clarify the actual purpose.

packages/json-magic/src/utils/is-json.ts
```diff
@@ -0,0 +1,9 @@
+export function isJson(value: string) {
+  try {
+    JSON.parse(value)
+
+    return true
+  } catch {
+    return false
+  }
+}
```
geoffgscott: Lets improve this and call is `isJsonObject`.

```ts
export function isJsonObject(value: string) {
  // If we do not have an opening array or object bracket the string is not a valid JSON objects
  if (!/^\s*(\{|\[)/.test(value.slice(0, 500))) {
    return false;
  }

  try {
    const val = JSON.parse(value);
    return isObject(val)

    return true;
  } catch {
    return false;
  }
}
```

packages/json-magic/src/utils/is-yaml.ts
```diff
@@ -0,0 +1,18 @@
...
+  try {
+    parse(value, {
+      maxAliasCount: 10000,
+    })
+
+    return true
+  } catch (_error) {
+    return false
+  }
+}
```
geoffgscott: I think we should remove this function. YAML parsing is VERY slow and the YAML parser will parse JSON without errors. Rather than checking for YAML we can just check for JSON and then attempt parsing a valid object as YAML whenever it's not JSON.

amritk:
Bit of duplication here, we should centralize
https://github.com/scalar/scalar/pull/6328/files#diff-58fd5b1ca986c47a4ca78c02e53e8ceed6ef0510325bcea363a40cdb515d191dR4

DemonHa:
@geoffgscott @hanspagel Should we release this as a minor or as a major?

marclave (approved):
minor :o

i got the verbal @geoffgscott approval 😈

Issue #6367, opened 2025-07-25
nuxt4 docs bug


InfernalAzazel:
### What happens?

Error: useStore() was called without a store and no store instance was found. Make sure to call useStore(store) in a parent component first.

<img width="1584" height="791" alt="Image" src="https://github.com/user-attachments/assets/18588e08-f633-4098-8b83-00207c7b018b" />

### What did you expect to happen?

Normal display

amritk:
Hello, thanks for the report. We got a fix in for it just waiting on a release. For now you can just roll back the version. The fix should be out early next week!

amritk:
Hello the fix is out! Please try again on the latest version and re-open this if its still an issue!

InfernalAzazel:
> Hello the fix is out! Please try again on the latest version and re-open this if its still an issue!

@amritk A new error has occurred

<img width="1709" height="853" alt="Image" src="https://github.com/user-attachments/assets/e055465a-d366-44c3-946b-f22c9ec5fe41" />

Configuration

<img width="1706" height="723" alt="Image" src="https://github.com/user-attachments/assets/50cf2ed7-0b8d-44fc-b3ae-37b0c4d3a582" />

print
```shell
[nuxt] error caught during app initialization H3Error: The requested module '/_nuxt/@fs/Users/kylin/work/code/github/nuxt-full-stack/node_modules/.pnpm/jsonpointer@5.0.1/node_modules/jsonpointer/jsonpointer.js?v=6e479cee' does not provide an export named 'default' 
```
package.json
```json
{
  "name": "nuxt-app",
  "private": true,
  "type": "module",
  "scripts": {
    "build": "nuxt build",
    "dev": "nuxt dev",
    "generate": "nuxt generate",
    "preview": "nuxt preview",
    "postinstall": "nuxt prepare"
  },
  "dependencies": {
    "@nuxt/eslint": "1.7.1",
    "@nuxt/image": "1.10.0",
    "@nuxt/ui-pro": "^3.3.0",
    "@scalar/nuxt": "0.4.35",
    "@tailwindcss/vite": "^4.1.11",
    "@vueuse/nuxt": "13.6.0",
    "eslint": "^9.32.0",
    "mongoose": "^8.16.5",
    "nuxt": "^4.0.2",
    "nuxt-auth-utils": "0.5.20",
    "tailwindcss": "^4.1.11",
    "typescript": "^5.8.3",
    "vue": "^3.5.18",
    "vue-router": "^4.5.1",
[...]

PR #6415, opened 2025-07-30
fix(nuxt): CommonJS issues

amritk:
**Problem**
clsoes #6367

Currently, the commonjs issue came back.

**Solution**

With this PR I uncommented the fix

**Checklist**

### DIFF
--- a/.changeset/tall-frogs-talk.md
+++ b/.changeset/tall-frogs-talk.md
@@ -1,0 +1,5 @@
+---
+'@scalar/nuxt': patch
+---
+
+fix: nuxt commonjs issues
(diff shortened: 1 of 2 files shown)


hanspagel:
I just tried to reproduce the issue in main, but the error doesn’t even appear in `examples/nuxt`. Only when I create a separate Nuxt project. 🤔

Probably because we’re ESM everywhere? Or what’s the difference?

hanspagel (approved):
it’s broken rn, so it can only get better

but I’d really love to have some way of testing for it

marclave (approved):

amritk:
@hanspagel Yea for some reason the issues only show up in an external repo, lets just get rid of all the commonjs packages and problem solved

PR #6417, opened 2025-07-30
chore(scripts): rename @scalar/scripts to @scalar/pre-post-request-scripts

geoffgscott:
**Problem**

This package has a very vague name based on the intent. This updates that.


geoffgscott:
@hanspagel @amritk Do we really feel this package needs to be separate? If we wanted to fork anything I would think a package that extracts the entire fetch wrapper used by the client would be more appropriate. Plugins like this can then we exported from the same package.

amritk:
I'm not even sure what this package does so I'll leave it to @hanspagel to answer

hanspagel (approved):

hanspagel:
needs to be separate, because it’ll require the actual postman scripting API package (if I find the day to add it), which we don’t want to install by default

marclave:
@geoffgscott @hanspagel can merge once conflicts are resolved : )

PR #6419, opened 2025-07-31
feat(workspace-store): rebase document origin with a remote origin

DemonHa:
**Problem**

We currently track the initial document HEAD, but there's no mechanism to rebase the local HEAD with the remote HEAD before pushing local changes. This can lead to inconsistencies or conflicts when syncing changes across users or systems.

> We use Git terminology here because the concepts (e.g., HEAD, rebasing, conflicts) closely align with Git’s version control model.

**Solution**

This PR introduces support for rebasing the local document HEAD against a remote HEAD prior to pushing local changes.
- If conflicts are detected during the rebase, we return an array of conflicting changes so users can resolve them manually.
- In case the conflict affects the activeDocument, we currently drop local changes and preserve the intermediate version as the source of truth.

```ts
// Example: Rebase a document with a new origin and resolve conflicts
const conflicts = store.rebaseDocument('api', newOriginDoc)
if (conflicts && conflicts.length > 0) {
  // User resolves conflicts here...
  store.rebaseDocument('api', newOriginDoc, userResolvedConflicts)
}
```

**Checklist**


hanspagel (approved):
wow, this is crazy

we just need to update the mocked workspace in the api-reference package, can you do that?

packages/workspace-store/src/client.ts
```diff
@@ -701,6 +708,81 @@ export const createWorkspaceStore = (workspaceProps?: WorkspaceProps): Workspace
...
+      resolvedConflicts?: Difference[],
+    ) => {
+      const newOrigin = coerceValue(OpenAPIDocumentSchema, upgrade(newDocumentOrigin).specification)
+
+      const originalDocument = originalDocuments[documentName]
+      const intermediateDocument = intermediateDocuments[documentName]
+      const activeDocument = getRaw(workspace.documents[documentName]!) // raw version without any overrides
+
+      if (!originalDocument || !intermediateDocument || !activeDocument) {
+        // If any required document state is missing, do nothing
```
geoffgscott, line 720: I think we may want to at least console log a big scary warning here because it's never expected.

packages/workspace-store/src/client.ts
```diff
@@ -701,6 +708,81 @@ export const createWorkspaceStore = (workspaceProps?: WorkspaceProps): Workspace
...
+      // TODO: In the future, implement smarter conflict resolution if needed
+      const changesetB = changesB.diffs.concat(changesB.conflicts.flatMap((it) => it[0]))
+
+      const newActiveDocument = apply(deepClone(newIntermediateDocument), changesetB)
+
+      // Update the active document to the new value
+      workspace.documents[documentName] = createOverridesProxy(
+        createMagicProxy({ ...newActiveDocument }),
+        overrides[documentName],
+      ) as (typeof workspace.documents)[number]
```
geoffgscott: Is there a way to avoid this typecast?

DemonHa: we can remote the typecast here and move it to the apply function return value, since in the `apply` function from `json-diff` we can not be sure that the output value is going to have the same type with the input since we are performing changes, but in this case we can safely cast it to the same type as the input since the changeset is generated from a document with the same type

PR #6432, opened 2025-08-01
feat(api-reference): parameter linking

hanspagel:
**Problem**

If you’d like to redirect a user to a specific operation parameter (e.g. in support), you’ve got no way to do this.

**Solution**

This PR adds an ID to all the parameters (operation.requestBody && operation.parameters), that you can copy as an URL share as an URL with others.

**Preview**

Screenshot 2025-08-01 at 11 31 37

**Checklist**


packages/api-reference/src/components/Anchor/WithBreadcrumb.vue
hwkr, line 1: This is kind of a nit pick but aren't UI breadcrumbs usually like `Home / Somewhere / Subpage` not like a link?

hwkr (approved):

PR #6433, opened 2025-08-01
fix(api-reference): test expects just one external request

hanspagel:
**Problem**

We’ve added e2e tests for the configuration, and one was expecting exact 1 external request to `fonts.scalar.com`. But actually, it’s 2 requests. Whether we see both requests in the test, is a matter of time.

**Solution**

I don’t think we need to wait for both requests to got out, it’s enough to know that at least one request goes out. So I’ve updated the test to just check whether there was any request to `fonts.scalar.com`.

**Checklist**


hanspagel:
Super annoying, I have to skip the queue. 😬

PR #6434, opened 2025-08-01
test(api-reference): snapshot testing important documents

hanspagel:
**⚠️ Remaining Tasks (before we can merge)**

- [ ] decide on a list of documents (@marclave @cameronrohani)
- [ ] put the examples in the registry, so the tests don’t fail just because the documents change
- [ ] update the snapshots, because they changed already
- [ ] copy a few of the smart ideas of #6480 to this PR here 🤠

**Problem**

Currently, we only have snapshot tests for the Scalar Galaxy example, so we might break other documents without catching it in CI.

**Solution**

This PR adds snapshot tests for a few more documents.

**Usage**

I’ve built a Docker proxy for Playwright, so you can run the tests and update the snapshots in an isolated environment.

If you’re on Linux, it’ll directly run Playwright - without Docker.

```bash
# Go to the package
cd packages/api-reference

# Run tests inside Docker, get a visual diff for any snapshots that don’t match
pnpm test:e2e

# Update the snapshots
pnpm test:e2e:update-snapshots
```

**Checklist**


hanspagel:
@geoffgscott This is what we used before and AFAIK it was never the reason the tests failed, so I’d just keep it for now. I’m just gonna increase it if it’s causing trouble. 👍

amritk:
converted to draft so you can merge in the ideas from the other snapshots

hanspagel:
I think we can delete this, we compare with the CDN already and copy the same stuff to compare with main

image

PR #6440, opened 2025-08-02
feat(json-magic): introduce type-safe `apply` function with tracked target type in `diff` results

DemonHa:
**Problem**

The current apply function returns a generic UnknownObject, since the original target type of the Difference[] list is not retained. This limits type safety and forces consumers to manually cast the result.

**Solution**

This PR introduces a new DifferenceResult object returned by the diff function. This object retains the target type that the differences were generated from. With this context preserved, the apply function can now infer and return the correct type, enabling full type safety when applying diffs.

⚠️ **Note:** The return type of apply is only accurate when the entire changeset is provided, including all resolved conflicts. Partial or unresolved diffs may lead to an incomplete or invalid cast, so full context is required for complete type safety.

**Checklist**

### DIFF
--- a/packages/json-magic/src/diff/apply.ts
+++ b/packages/json-magic/src/diff/apply.ts
@@ -76,7 +76,7 @@ export const apply = <T>(document: Record<string, unknown>, diff: DifferenceResu
     applyChange(document, d.path, d)
   }
 
-  // Safe to typecast since this function which apply changes to A to turn it into B
-  // where B is the type of the changelog target
+  // It is safe to cast here because this function mutates the input document
+  // to match the target type T as described by the diff changeset.
   return document as T
 }
--- a/packages/json-magic/src/diff/apply.ts
+++ b/packages/json-magic/src/diff/apply.ts
@@ -1,4 +1,4 @@
-import type { Difference } from '@/diff/diff'
+import type { Difference, DifferenceResult } from '@/diff/diff'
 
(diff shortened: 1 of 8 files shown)


hanspagel (commented):
Do you want to add a few type tests, too?

Here is a simple example:
https://github.com/scalar/scalar/blob/main/packages/openapi-parser/src/utils/upgrade.test-d.ts

Here are the docs:
https://vitest.dev/guide/testing-types

Just be aware that vitest doesn’t run them as they do with other tests, they are evaluated on build (or with `vitest --typecheck`).

marclave (approved):
🆒

packages/json-magic/src/diff/apply.ts
```diff
@@ -36,9 +36,12 @@ export class InvalidChangesDetectedError extends Error {
  * const updated = apply(original, changes)
  * // Result: original document with content added to the 200 response
  */
-export const apply = (document: Record<string, unknown>, diff: Difference[]): Record<string, unknown> => {
+export const apply = <T extends Record<string, unknown>>(
+  document: Record<string, unknown>,
```
geoffgscott, line 40: Would we want the document to be of type T as well? Or possible extended/partial of T

DemonHa: No, we don't care about the source, we only care about the target object type. The types of the source and target can be totally unrelated

PR #6444, opened 2025-08-02
chore(workspace-store): clean up client store

DemonHa:
**Checklist**

### DIFF
--- a/packages/workspace-store/src/helpers/overrides-proxy.ts
+++ b/packages/workspace-store/src/helpers/overrides-proxy.ts
@@ -1,8 +1,12 @@
 import { isObject, type UnknownObject } from '@/helpers/general'
+import type { DeepPartial } from '@/types'
 
 const isOverridesProxy = Symbol('isOverridesProxy')
 
-export const createOverridesProxy = <T extends object>(targetObject: T, overrides?: unknown): T => {
+export const createOverridesProxy = <T extends Record<string, unknown>>(
+  targetObject: T,
+  overrides?: DeepPartial<T>,
+): T => {
   if (!targetObject || typeof targetObject !== 'object') {
     return targetObject
   }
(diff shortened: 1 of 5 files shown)


packages/workspace-store/src/types.ts
```diff
@@ -6,13 +20,19 @@ export type DeepRequired<T> = T extends Function
...
+ * - If T is an object, recursively applies DeepPartial to each property, making them optional.
+ * - Otherwise, T is returned as-is.
+ *
+ * @template T - The type to make deeply partial (optional).
+ * @example
+ * type Example = { a: { b: number } }
+ * type PartialExample = DeepPartial<Example>
+ * // Result: { a?: { b?: number } }
+ */
+export type DeepPartial<T> = T extends object ? { [K in keyof T]?: DeepPartial<T[K]> } : T
```
amritk: You can just use type-fest for these, we use it already

https://github.com/sindresorhus/type-fest/blob/main/source/partial-deep.d.ts

geoffgscott: IMO we should be dropping type fest. There are only a couple things we need from it.

geoffgscott: Out of scope though.

amritk: type-fest is in as per slack convo!

amritk (approved):
We fixed up the issues on this one!  🚂 

Would like someone from the outside to test it once! 🙏🏾

marclave (approved):

packages/workspace-store/src/helpers/overrides-proxy.ts
```diff
@@ -2,55 +2,99 @@ import { isObject, type UnknownObject } from '@/helpers/general'
...
+ * - If T is an object, recursively applies DeepPartial to each property, making them optional.
+ * - Otherwise, T is returned as-is.
+ *
+ * @template T - The type to make deeply partial (optional).
+ * @example
+ * type Example = { a: { b: number } }
+ * type PartialExample = DeepPartial<Example>
+ * // Result: { a?: { b?: number } }
+ */
+export type DeepPartial<T> = T extends object ? { [K in keyof T]?: DeepPartial<T[K]> } : T
```
amritk, line 17: https://github.com/sindresorhus/type-fest/blob/main/source/partial-deep.d.ts

this one handles arrays, maps, sets etc. Unless you don't need those

amritk:
> I don’t really know what to look for TBH, anything specific I should test?

@hanspagel just generally things working haha, references, client etc

Actually the e2e tests caught a bunch of issues that we fixed!

packages/workspace-store/src/helpers/overrides-proxy.ts
DemonHa: The problem there is that I can not use a complex type because typescript will then infer the T as an union of both inputs. That's the reason why I have a custom type just for the proxy. This way typescript can understand to infer the type only from the first param and the second one is just derived from the first

amritk: yolo

PR #6454, opened 2025-08-04
fix(openapi-parser): can't import plugins

### LINKED ISSUE #6452: can’t import plugin from openapi-parser anymore
> the scalar openapi parser package can no longer import plugins, only `.` specifier left hence the docs examples are not working

hanspagel:
**Problem**

In #6414 we removed the plugin exports from `@scalar/openapi-parser` and broke the examples from the README. That broke the update for users.

See #6452

**Solution**

This PR adds the entries/exports again and fixes #6452.

**Checklist**

### DIFF
--- a/.changeset/good-donuts-mate.md
+++ b/.changeset/good-donuts-mate.md
@@ -1,0 +1,5 @@
+---
+'@scalar/openapi-parser': patch
+---
+
+fix: can’t import plugins from parser package anymore


DemonHa:
@hanspagel We remove the bundler and it's plugin from `openapi-parser` and move everything to the new package `json-magic`

PR #6456, opened 2025-08-04
refactor(playwright): rename `tests/` to `test/`

hanspagel:
We use a `test/` directory everywhere, but `tests/` in the playwright folder.

This PR renames the folder to `test/`, too.


amritk:
Yes one PR for just name change / moves, this is the way!

PR #6457, opened 2025-08-04
refactor(scalar-app): move electron e2e tests to scalar-app directory

hanspagel:
We used to put all playwright tests in the `playwright/` folder, but it feels a bit strange to have the test so far away from the code.

This PR moves the Electron e2e test to the `scalar-app/` directory.

I've also merged the separate Electron E2E GitHub Action workflow into the main CI workflow, so we can actually use the turbo build cache and not do another full build in parallel.


marclave (approved):
🔥

PR #6460, opened 2025-08-04
fix(api-reference): models with `x-scalar-ignore` show up again

antlio:
**Changes**

this pr updates the models component in order to use the getModels function to render schema models.

| state | preview |
| -------|------|
| before | image |
| after | image | 

**Checklist**

### DIFF
--- a/.changeset/kind-buckets-vanish.md
+++ b/.changeset/kind-buckets-vanish.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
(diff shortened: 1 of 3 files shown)


antlio:
> I think we should fix this in the `traverse` functions, or don’t we use those here yet?

we don't it seems - i just quickly patched this while relying on what we had - prob should tho? not sure about the state of things here

hanspagel (approved):
OK, cool! thanks for putting this up so quickly

PR #6461, opened 2025-08-04
feat(galaxy): publish in Scalar Registry

hanspagel:
**Problem**

Currently, we’re publishing @scalar/galaxy on NPM and promote the jsdelivr URLs, but don’t have it in the Scalar Registry™.

**Solution**

With this PR the OpenAPI document is pushed to the Scalar Registry on release. It’s using the version from the package.json (which is also in the document).

The resulting URLs:

* https://registry.scalar.com/@scalar/apis/galaxy
* https://registry.scalar.com/@scalar/apis/galaxy/latest?format=json
* https://registry.scalar.com/@scalar/apis/galaxy/latest?format=yaml

I’ll replace the jsdelivr URLs in a different PR.

**Checklist**

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -584,7 +584,7 @@ jobs:
         run: npx @scalar/cli auth login --token ${{ secrets.SCALAR_API_KEY }}
       - if: steps.changed-files.outputs.galaxy_any_changed == 'true'
         name: Push to Scalar Registry
-        run: npx @scalar/cli registry publish --force true --namespace scalar --slug galaxy --version ${{ env.VERSION }} packages/galaxy/dist/latest.yaml
+        run: npx @scalar/cli registry publish --namespace scalar --slug galaxy --version ${{ env.VERSION }} packages/galaxy/dist/latest.yaml
 
   deploy-examples:
     # Main Branch or PR from the same repository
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -485,6 +485,43 @@ jobs:
...
+        working-directory: packages/galaxy
+        run: |
+          VERSION=$(node -p "require('./package.json').version")
+          echo "VERSION=$VERSION" >> $GITHUB_ENV
+      - if: steps.changed-files.outputs.galaxy_any_changed == 'true'
+        name: Login to Scalar Registry
+        run: npx @scalar/cli auth login --token ${{ secrets.SCALAR_API_KEY }}
+      - if: steps.changed-files.outputs.galaxy_any_changed == 'true'
+        name: Push to Scalar Registry
+        run: npx @scalar/cli registry publish --force true --namespace scalar --slug galaxy --version ${{ env.VERSION }} packages/galaxy/dist/latest.yaml
```
geoffgscott: Why are we using `--force` here? That is used only when we don't want to override a version. If we have any changes we should really be doing an up-rev

packages/galaxy/README.md
```diff
@@ -15,6 +15,13 @@ npm install @scalar/galaxy
 
 ## Usage
 
+### Scalar Registry
+
+| Version | Format | URL                                                                |
+| ------- | ------ | ------------------------------------------------------------------ |
+| Latest  | JSON   | https://registry.scalar.com/@scalar/apis/galaxy/latest?format=json |
+| Latest  | YAML   | https://registry.scalar.com/@scalar/apis/galaxy/latest?format=yaml |
```
geoffgscott: We should add the preview here as well.

marclave (approved):
AWESOME :)

PR #6466, opened 2025-08-05
Do not encode placeholder curly braces

ledgersteve:
**Problem**

This can be demonstrated with the _Galaxy_ example api docs that come with Scalar.  Various of the methods have placeholders in them, e.g. `get/planets/{planetId}`  The curly brackets there are placeholders, saying "you should put your _planetId_ here."  The curly brackets set the parameters off from the method path.  The curly brackets shouldn't be included in your path when you submit it; they are there just to say "this isn't part of the method path - this is where you add your stuff."

It follows then that the curly brackets should not be URL escaped.  

Some of the code samples viewable in the Scalar API explorer get this right.  e.g. for snippets in _Dart Http, JavaScipt Fetch, JavaScript oFetch, Node.js Fetch, Node.js ofetch, Node.js undici, PHP cURL, PHP Guzzle, Python Requests, Python HTTPX, Rust reqwest, Shell Curl,_ you get an example looking something like this (this is _Node.js Fetch_)

```
fetch('https://galaxy.scalar.com/planets/{planetId}')
```

But for other snippets, including _JavaScript Axios, JavaScript jQuery, Python http.client, C#,_ the curly braces end up getting URL-encoded, which obscures their function as marking the placeholder for where your parameter should go. e.g., for _C#_:

```
RequestUri = new Uri("https://galaxy.scalar.com/planets/%7BplanetId%7D"),
```

This is wrong.  It hides that the curly brackets are there to show where your _planetId_ should go.

**Tests**

There are many unit tests that touch this area with names like 'handles special characters in URL', and they check that placeholders using square brackets get through without being URL-encoded.  But most API docs (including Scalar's example _Galaxy_ docs) use curly brackets for denoting placeholders.  So this PR also adds tests for this - where we had a test for 'handles special characters in URL', we now have tests for 'handles special characters in URL, square brackets' and 'handles special characters in URL, curly brackets'.

**Solution**

With this PR, this issue is fixed.  The snippetz code makes extensive use of the JavaScript _URL_ object, and gets the page from the `URL.pathname` which comes out URL-encoded where applicable, including encoding curly brackets.  The new `getPathnameWithRestoredBrackets.ts` file has helper methods to reinstate brackets unencoded when they are doing the work of a placeholder.

**Checklist**

- [✓] I've added an explanation _why_ this change is needed.
- [✓] I've added a changeset (`pnpm changeset`).
- [✓] I've added tests for the regression or new feature.
- [✓] I've updated the documentation.

### DIFF
--- a/.changeset/blue-dancers-hang.md
+++ b/.changeset/blue-dancers-hang.md
@@ -1,0 +1,5 @@
+---
+'@scalar/snippetz': patch
+---
+
+Do not encode placeholder curly brackets in code samples.
--- a/packages/snippetz/src/httpsnippet-lite/esm/targets/python/python3/client.ts
+++ b/packages/snippetz/src/httpsnippet-lite/esm/targets/python/python3/client.ts
@@ -10,6 +10,7 @@
  */
 import { CodeBuilder } from '../../../helpers/code-builder.js'
 import { escapeForDoubleQuotes } from '../../../helpers/escape.js'
+import { getStringPathnameWithRestoredBrackets } from '@/utils/getPathnameWithRestoredBrackets.js'
 
 export const python3 = {
   info: {
@@ -22,6 +23,9 @@ export const python3 = {
     { uriObj: { path, protocol, host }, postData, allHeaders, method },
     options = {},
   ) => {
+
+    const bracketedPath = getStringPathnameWithRestoredBrackets(path)
+
     const { insecureSkipVerify = false } = options
     const { push, blank, join } = new CodeBuilder()
     // Start Request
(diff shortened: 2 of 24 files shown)


hanspagel:
awesome PR, I think @amritk can tweak it a bit

amritk:
Hey first of all, thanks for the awesome contribution! 

I started looking at it but got side tracked working on something. As soon as I'm done I'll come back to this one as we actually used to have this working so I'll check what the difference is

hanspagel:
Aahhh, I think we need to take a deeper look into that, but are a bit busy right now. I’m closing this PR for now. I’m so thankful you brought this up 🙏

ledgersteve:
@hanspagel Can we cross-reference this to an issue please, so that we keep a track on this as a bug that would be good to fix? (I'm happy to raise the issue if you want.)

PR #6469, opened 2025-08-05
feat(api-reference): option for sidebar to display path instead of summary

ledgersteve:
**Problem**

Currently, in the sidebar of the API explorer, it shows the method summary.  e.g. with the Scalar example _Galaxy_ API yaml, for the `/planets/{planetId}` method, we show "Get a planet".  

This is fine with the concise _Galaxy_ dataset.  But we have to work with APIs where the writer of the API method summaries is not so concise, and the method summary can get quite long.  To see this sort of thing, try changing the Scalar example _Galaxy_ API yaml, for the `/planets/{planetId}` method, to have a summary of "Get a particular celestial body from the solar system currently classified as a planet" (made up example, but we have had third-party APIs that are worse).  The sidebar then gets very long (tall) and wordy, and we would be much better off showing the method path (e.g. `/planets/{planetId}`) rather than the method summary (e.g.  `Get a particular celestial body from the solar system currently classified as a planet`).  It makes the sidebar more concise, and it makes it easier to visually parse it for the API endpoint that you want.

**Solution**

With this PR we introduce a _sidebarUsePath_ configuration value.  This is a nullable bool.  When not _true_, the API explorer behaves as before, i.e. the sidebar shows the method summary.  When set to _true_, the API explorer sidebar shows the method path.

**Checklist**

- [x] I've added tests for the regression or new feature, in the area of checking default config values.  I've not added UI tests, through ignorance of how to do so.
- [x] I've updated the documentation - the relevant MD files describe this new setting.

### DIFF
--- a/.changeset/eleven-coats-dress.md
+++ b/.changeset/eleven-coats-dress.md
@@ -1,0 +1,8 @@
+---
+'@scalar/aspnetcore': minor
+'@scalar/api-reference': minor
+'@scalar/aspire': minor
+'@scalar/types': minor
+---
+
+Configurable option for sidebar to show method path instead of method summary.
--- a/.changeset/eleven-coats-dress.md
+++ b/.changeset/eleven-coats-dress.md
@@ -1,8 +1,8 @@
 ---
-'@scalar/aspnetcore': minor
-'@scalar/api-reference': minor
-'@scalar/aspire': minor
-'@scalar/types': minor
+'@scalar/aspnetcore': patch
+'@scalar/api-reference': patch
+'@scalar/aspire': patch
+'@scalar/types': patch
 ---
 
 Configurable option for sidebar to show method path instead of method summary.
--- a/documentation/configuration.md
+++ b/documentation/configuration.md
@@ -302,13 +302,13 @@ Whether the sidebar should be shown.
 }
 ```
 
-### sidebarUsePath?: boolean
+### operationTitleSource?: 'summary' | 'path'
 
-Whether the sidebar display text should use the method path instead of the method summary.
+Whether the sidebar display text and search should use the operation summary or the operation path.
 
 ```js
 {
-  sidebarUsePath: true
+  operationTitleSource: 'path'
 }
 ```
 
--- a/documentation/configuration.md
+++ b/documentation/configuration.md
@@ -308,13 +308,13 @@ Whether the sidebar display text and search should use the operation summary or 
 
 ```js
 {
-  operationTitleSource: 'summary' // Show method summary (default)
+  operationTitleSource: 'summary' // Use method summary (default)
 }
 ```
 
 ```js
 {
-  operationTitleSource: 'path' // Show method path
+  operationTitleSource: 'path' // Use method path
 }
 ```
 
--- a/documentation/configuration.md
+++ b/documentation/configuration.md
@@ -308,7 +308,13 @@ Whether the sidebar display text and search should use the operation summary or 
 
 ```js
 {
-  operationTitleSource: 'path'
+  operationTitleSource: 'summary' // Show method summary (default)
+}
+```
+
+```js
+{
+  operationTitleSource: 'path' // Show method path
 }
 ```
 
--- a/documentation/configuration.md
+++ b/documentation/configuration.md
@@ -302,6 +302,16 @@ Whether the sidebar should be shown.
 }
 ```
 
+### sidebarUsePath?: boolean
+
+Whether the sidebar display text should use the method path instead of the method summary.
+
+```js
+{
+  sidebarUsePath: true
+}
+```
+
 ### hideModels?: boolean
 
 Whether models (`components.schemas` or `definitions`) should be shown in the sidebar, search and content.
--- a/documentation/configuration.md
+++ b/documentation/configuration.md
@@ -308,13 +308,13 @@ Whether the sidebar display text and search should use the operation summary or 
 
 ```js
 {
-  operationTitleSource: 'summary' // Use method summary (default)
+  operationTitleSource: 'summary' // Use operation summary (default)
 }
 ```
 
 ```js
 {
-  operationTitleSource: 'path' // Use method path
+  operationTitleSource: 'path' // Use operation path
 }
 ```
 
--- a/integrations/aspire/src/Scalar.Aspire/Enums/TitleSource.cs
+++ b/integrations/aspire/src/Scalar.Aspire/Enums/TitleSource.cs
@@ -1,0 +1,23 @@
+using System.ComponentModel;
+using NetEscapades.EnumGenerators;
+
+namespace Scalar.Aspire;
+
+/// <summary>
+/// Represents the theme modes available in the application.
+/// </summary>
+[EnumExtensions]
+public enum TitleSource
+{
+    /// <summary>
+    /// Title comes from operation summary.
+    /// </summary>
+    [Description("summary")]
+    Summary,
+
+    /// <summary>
+    /// Title comes from operation path.
+    /// </summary>
+    [Description("path")]
+    Path
+}
--- a/packages/api-reference/src/features/sidebar/components/SidebarElement.vue
+++ b/packages/api-reference/src/features/sidebar/components/SidebarElement.vue
@@ -33,15 +33,11 @@ const { getFullHash, isIntersectionEnabled, replaceUrlState } = useNavState()
 const config = useConfig()
 
 const getPathOrTitle = (item: TraversedEntry): string => {
-  const itemPath = 'path' in item ? item.path : item.title
-  const escapedItemPath = itemPath
-    .replace(/&/g, '&amp;')
-    .replace(/"/g, '&quot;')
-    .replace(/'/g, '&#39;')
-    .replace(/</g, '&lt;')
-    .replace(/>/g, '&gt;')
-    .replace(/\//g, '&#47;<wbr>') // add a line break opportunity after every forward-slash
-  return escapedItemPath
+  if ('path' in item) {
+    // Insert zero-width space after every slash, to give line-break opportunity.
+    return item.path.replace(/\//g, '/\u200B')
+  }
+  return item.title
 }
 
 // We disable intersection observer on click
@@ -142,8 +138,8 @@ const onAnchorClick = async (ev: Event) => {
         <p class="sidebar-heading-link-title">
           <span
             v-if="config.operationTitleSource === 'path'"
-            class="hanging-indent"
-            v-html="getPathOrTitle(item)">
+            class="hanging-indent">
+            {{ getPathOrTitle(item) }}
           </span>
           <span v-else>
             {{ item.title }}
--- a/packages/api-reference/src/features/sidebar/components/SidebarElement.vue
+++ b/packages/api-reference/src/features/sidebar/components/SidebarElement.vue
@@ -32,6 +32,13 @@ const { getFullHash, isIntersectionEnabled, replaceUrlState } = useNavState()
 
 const config = useConfig()
 
+const getPathOrTitle = (item: TraversedEntry): string => {
+  if ('path' in item) {
+    return item.path
+  }
+  return item.title
+}
+
 // We disable intersection observer on click
 const handleClick = async () => {
   // wait for a short delay before enabling intersection observer
@@ -128,7 +135,14 @@ const onAnchorClick = async (ev: Event) => {
         :tabindex="hasChildren ? -1 : 0"
         @click="onAnchorClick">
         <p class="sidebar-heading-link-title">
-          {{ item.title }}
+          <span
+            v-if="config.sidebarUsePath"
+            class="hanging-indent">
+            {{ getPathOrTitle(item) }}
+          </span>
+          <span v-else>
+            {{ item.title }}
+          </span>
         </p>
         <p
           v-if="'method' in item && !hasChildren"
@@ -186,6 +200,14 @@ const onAnchorClick = async (ev: Event) => {
 .sidebar-heading-link-title {
   margin: 0;
 }
+
+.sidebar-heading-link-title .hanging-indent {
+  padding-left: 0.7em;
+  text-indent: -0.7em;
+  overflow-wrap: break-word;
+  word-wrap: break-word;
+}
+
 .sidebar-heading:hover {
   background: var(
     --scalar-sidebar-item-hover-background,
--- a/packages/api-reference/src/features/sidebar/components/SidebarElement.vue
+++ b/packages/api-reference/src/features/sidebar/components/SidebarElement.vue
@@ -33,10 +33,15 @@ const { getFullHash, isIntersectionEnabled, replaceUrlState } = useNavState()
 const config = useConfig()
 
 const getPathOrTitle = (item: TraversedEntry): string => {
-  if ('path' in item) {
-    return item.path
-  }
-  return item.title
+  const itemPath = 'path' in item ? item.path : item.title
+  const escapedItemPath = itemPath
+    .replace(/&/g, '&amp;')
+    .replace(/"/g, '&quot;')
+    .replace(/'/g, '&#39;')
+    .replace(/</g, '&lt;')
+    .replace(/>/g, '&gt;')
+    .replace(/\//g, '&#47;<wbr>') // add a line break opportunity after every forward-slash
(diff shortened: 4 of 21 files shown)


documentation/configuration.md
```diff
@@ -302,6 +302,16 @@ Whether the sidebar should be shown.
...
+### sidebarUsePath?: boolean
+
+Whether the sidebar display text should use the method path instead of the method summary.
+
+```js
+{
+  sidebarUsePath: true
+}
+```
+
```
hanspagel: It’s used in the sidebar and the search, and we might be more flexible if we make it an enum and not a boolean.

What do you think about the following example?

```suggestion
### defaultOperationTitle?: 'summary' | 'path'

Control whether the sidebar shows the summary or the path in the sidebar and in the search.

Defaults to `'summary'`, falls back to `'path'`.

```js
{
  defaultOperationTitle: 'path'
}
```
```

Comment (ledgersteve) [reply documentation/configuration.md]: I think that's much better, yes - more readable and makes our intention clearer.  I see that is a "requested change" - I will do it as requested.

Comment (ledgersteve) [reply documentation/configuration.md]: I've made that change, so it is now a `'summary'|'path'` enum instead of a boolean.  The suggested config item name was `defaultOperationTitle`, but I don't think that would be quite right.  We're not setting a _default_ of anything, we are setting what it _is_.  I have gone with `operationTitleSource` (i.e. from where do we get what we use to display as the operation title), but it is an easy global search and replace to change it if we think of something better.

**Comment by ledgersteve** [comment]
With my latest commit ("Fix mock for traverse-document-tests", b75824322), the _stats_ CI job is failing with this message:

```
Run relative-ci/agent-action@1707825cbfcc7452b2913d273414705415ae64d4
[dotenv@17.2.1] injecting env (0) from .env -- tip: ⚙️  write to custom object with { processEnv: myObject }
fatal: bad object b75824322add2fa797814ef535fb0e4ea3ef4ba6
Faild to read the commit message from git: Command failed: git show -s --format=%B b75824322add2fa797814ef535fb0e4ea3ef4ba6
fatal: bad object b75824322add2fa797814ef535fb0e4ea3ef4ba6
```

But running this locally:
`git show -s --format=%B b75824322`
works fine for me.

If anyone has a tip for how to get the  _stats_ CI job passing, that would be great.

(Note received from Hans: it is OK for the _stats_ job not to pass for me.)

**Review by geoffgscott — APPROVED** [review]
Interface looks good.

**Comment by hanspagel** [comment]
@ledgersteve the stats job just needs a secret from our repository, it’s OK if it’s red here

**Comment by ledgersteve** [comment]
Thanks @xC0dex for the tip re enum usage in the C# implementations.  Done.

**Comment by xC0dex** [comment]
Hey @ledgersteve,

Could you please allow me to add some commits to the branch? I'm currently unable to make any changes. Thanks!

**Comment by ledgersteve** [comment]
> Could you please allow me to add some commits to the branch? I'm currently unable to make any changes. Thanks!

@xC0dex I'd love to, but it a PR from an organisation-owned fork, not a user-owned fork, so I don't have the "[ ] Allow edits and access to secrets by maintainers" option.  If there's another magical way, I don't know it.  Otherwise point me at the changes and I'll do them, or I abandon this PR and I'll do another one (at the cost of losing the conversation above), and this time do it from a fork owned by me (ledgersteve) rather than the organization (ledgerscope).

**Comment by ledgersteve** [comment]
@xC0dex I've invited you into our fork.  Hopefully that does it.

**Comment by xC0dex** [comment]
This is what I did: https://github.com/scalar/scalar/commit/ec012e481bb4e036d2f888bd211f2d33ed8cb175

I used a string instead of the enum in the `ScalarConfiguration` class, fixed a typo, and removed the default value of the `WithOperationTitleSource` method.

**Comment by ledgersteve** [comment]
> This is what I did: [ec012e4](https://github.com/scalar/scalar/commit/ec012e481bb4e036d2f888bd211f2d33ed8cb175)

@xC0dex These fixes (thank you) apply in the `integrations/aspire` area.  Do they need applying to the `integrations/aspnetcore` area too?  e.g. `integrations/aspnetcore/src/Scalar.AspNetCore/Contract/ScalarConfiguration.cs` looks to  be doing much the same thing as `integrations/aspire/src/Scalar.Aspire/Contract/ScalarConfiguration.cs`, so would need the same changes.

**Comment by xC0dex** [comment]
> Do they need applying to the integrations/aspnetcore area too

Yes please! I just started with Aspire and noticed that I cannot push 😄

**Comment by ledgersteve** [comment]
> Yes please! I just started with Aspire and noticed that I cannot push 😄

@xC0dex  Have applied your fixes (thank you) to the `Scalar.Aspire` and `Scalar.AspNetCore` areas.

File: .changeset/eleven-coats-dress.md [hunk 2]
```diff
@@ -0,0 +1,8 @@
+---
+'@scalar/aspnetcore': minor
+'@scalar/api-reference': minor
+'@scalar/aspire': minor
+'@scalar/types': minor
```
xC0dex:
```suggestion
'@scalar/aspnetcore': patch
'@scalar/api-reference': patch
'@scalar/aspire': patch
'@scalar/types': patch
```

Personally, I'd use `patch` instead of `minor`. We don't break anything or introduce something that could cause a conflict.

integrations/aspire/src/Scalar.Aspire/Enums/TitleSource.cs
```diff
@@ -0,0 +1,23 @@
+using System.ComponentModel;
+using NetEscapades.EnumGenerators;
+
+namespace Scalar.Aspire;
+
+/// <summary>
+/// Represents the theme modes available in the application.
+/// </summary>
+[EnumExtensions]
+public enum TitleSource
```
xC0dex: I wondered if we should name this `OperationTitleSource`

ledgersteve: I considered calling the enum `OperationTitleSource`, but:

1. Then the name of the enum that goes into the property is the same as the name of the property that the enum goes into, which slightly gives me the ick. e.g. we get code like `OperationTitleSource = OperationTitleSource.Path`, where the two mentions of `OperationTitleSource` are for quite different things.
2. We're not so much choosing an _OperationTitleSource_ as choosing a _TitleSource_ that we then apply as the source for the operation title, so I think _TitleSource_ is OK as a description for that.

So, I considered _OperationTitleSource_ but in the end felt that _TitleSource_ was better.  But I don't feel strongly about it and am happy to make changes as felt useful by those more experienced with the project than me.  Just say...

.changeset/eleven-coats-dress.md
ledgersteve: Thanks for the suggestion.  I've implemented it.

integrations/aspire/src/Scalar.Aspire/Enums/TitleSource.cs
xC0dex: I get what you mean! If we ever added another title source for any reason, it could cause confusion. That’s why I figured it’s best to make things clear right from the start. 

But... I also don't have a clear opinion. 🤣

hanspagel: `TitleSource` sounds like we’re talking about the title of the OpenAPI document TBH

hanspagel: maybe something like
`OperationTitleSource = OperationTitleSourceAttribute.Path`

xC0dex: The suffix `Attribute` has a special meaning in .NET. 

I see your point. Maybe lets use `OperationTitleSource`.

ledgersteve: I've renamed the enum from `TitleSource` to `OperationTitleSource`.

packages/api-reference/src/features/sidebar/components/SidebarElement.vue
```diff
@@ -32,6 +32,18 @@ const { getFullHash, isIntersectionEnabled, replaceUrlState } = useNavState()
...
 
+const getPathOrTitle = (item: TraversedEntry): string => {
+  const itemPath = 'path' in item ? item.path : item.title
+  const escapedItemPath = itemPath
+    .replace(/&/g, '&amp;')
+    .replace(/"/g, '&quot;')
+    .replace(/'/g, '&#39;')
+    .replace(/</g, '&lt;')
+    .replace(/>/g, '&gt;')
+    .replace(/\//g, '&#47;<wbr>') // add a line break opportunity after every forward-slash
```
hanspagel: wait, what is that doing?

packages/api-reference/src/features/sidebar/components/SidebarElement.vue
```diff
@@ -127,7 +139,14 @@ const onAnchorClick = async (ev: Event) => {
         :tabindex="hasChildren ? -1 : 0"
         @click="onAnchorClick">
         <p class="sidebar-heading-link-title">
-          {{ item.title }}
+          <span
+            v-if="config.operationTitleSource === 'path'"
+            class="hanging-indent"
+            v-html="getPathOrTitle(item)">
```
hanspagel: we want to avoid v-html I think

packages/api-reference/src/features/sidebar/components/SidebarElement.vue
```diff
@@ -127,7 +139,14 @@ const onAnchorClick = async (ev: Event) => {
         :tabindex="hasChildren ? -1 : 0"
         @click="onAnchorClick">
         <p class="sidebar-heading-link-title">
-          {{ item.title }}
+          <span
+            v-if="config.operationTitleSource === 'path'"
```
hanspagel, line 140: we should use a computed that has the if logic, and we might even want to have this logic in the traverse functions cc @amrit

packages/api-reference/src/features/sidebar/components/SidebarElement.vue
```diff
@@ -185,6 +204,14 @@ const onAnchorClick = async (ev: Event) => {
...
   margin: 0;
 }
+
+.sidebar-heading-link-title .hanging-indent {
+  padding-left: 0.7em;
+  text-indent: -0.7em;
+  overflow-wrap: break-word;
+  word-wrap: break-word;
+}
+
```
hanspagel, line 211: why do we need to change the CSS here? what does this do?

hanspagel (changes requested):
sorry, still some nitpicks left. maybe @amritk can help to bring this over the finish line soon-ish.

ledgersteve:
Response to @hanspagel   The answers to these queries are all related.  `getPathOrTitle` method adds a line break opportunity `<wbr>` after every forward slash.  Without this, the line breaks happen wherever they fall, which is ugly (actually, Firefox treats `/` as a line break opportunity, but Chromium-based browsers do not), so it looks like this:

image

So we add a `<wbr>` after each `/` to say to the browser, "You can break a line here if you need to", so it looks like this:

image

We're injecting angle brackets into the title for the line-break opportunity, so we have to use `v-html` otherwise it would get escaped to a visible `&lt;wbr&gt;` instead of an invisible line-break opportunity.  Now that we are using `v-html` we have to do the standard work for html escaping so that we do not introduce an XSS vuln.  So that's what the other substitutions in the _getPathOrTitle_ method are about.

Last thing to do is to take another look at our emitted output, with paths breaking at our line-break opportunity.  Everything is aligned hard left, but it would be easier to visually parse if we apply a [hanging indent](https://www.montana.edu/writingcenter/writing_resources/hanging-indents.html#:~:text=Works%20cited%20pages%20in%20APA,the%20beginning%20of%20individual%20citations.)  So that's what the extra bit of CSS does.  

The end result with line-break opportunity and hanging indent looks like this, with line breaks after forward slashes where applicable, and hanging indents helping show how things belong together:

image

packages/api-reference/src/features/sidebar/components/SidebarElement.vue
ledgersteve: See [my comment](#issuecomment-3192216286) below.

packages/api-reference/src/features/sidebar/components/SidebarElement.vue
ledgersteve: See [my comment](#issuecomment-3192216286) below.

packages/api-reference/src/features/sidebar/components/SidebarElement.vue
ledgersteve: See [my comment](#issuecomment-3192216286) below.

amritk:
@hanspagel whats needed here? We already show this if there's no summary right so there should be minimal changes required.

hanspagel:
Thanks for the explanation! Maybe we can use the https://en.wikipedia.org/wiki/Zero-width_space instead of `<wbr>` and `v-html`, and I’m not sure about the indent, wdyt @cameronrohani?

ledgersteve:
> Maybe we can use the https://en.wikipedia.org/wiki/Zero-width_space instead of `<wbr>` and `v-html`

@hanspagel  Don't we end up with the same "how do we inject characters that would normally get escaped" problem if trying to add `&ZeroWidthSpace;` , `&#8203;`, or `&#x200B;` as we do when adding `<wbr>` after every slash?  i.e. without the use of `v-html`, the ampersands would get escaped to `&amp;` and then not work as indicating the start of an HTML entity.

hanspagel:
Nah, I mean the actual unicode character, not the HTML entitity.

I chat with @cameronrohani about the design changes and hope we can get this over the finish line quickly.

Give me a moment. No need to change anything. I can do the final tweaks.

xC0dex (approved):

ledgersteve:
> Nah, I mean the actual unicode character, not the HTML entitity.

@hanspagel Ah yes, very good.  Change done as suggested. (Your latest was emailed to me with just the first line - only saw the later 'change nothing' addition when I came here after the commit. Hope that's not caused a problem.)

packages/api-reference/src/features/sidebar/components/SidebarElement.vue
amritk: Hey @hanspagel what's this for exactly? Just a display version of the path?

hanspagel:
unfortunately, we ran into conflicts and I can’t add to your PR.
can you allow me to edit the PR or resolve the conflicts?

then we should be good to finally get it in 🙏

documentation/configuration.md
```diff
@@ -302,6 +302,22 @@ Whether the sidebar should be shown.
 }
 ```
 
+### operationTitleSource?: 'summary' | 'path'
+
+Whether the sidebar display text and search should use the operation summary or the operation path.
+
+```js
+{
+  operationTitleSource: 'summary' // Use method summary (default)
```
hanspagel:
```suggestion
  operationTitleSource: 'summary' // Use operation summary (default)
```

documentation/configuration.md
```diff
@@ -302,6 +302,22 @@ Whether the sidebar should be shown.
...
+}
+```
+
+```js
+{
+  operationTitleSource: 'path' // Use method path
```
hanspagel:
```suggestion
  operationTitleSource: 'path' // Use operation path
```

packages/api-reference/src/features/sidebar/components/SidebarElement.vue
hanspagel: yes, just display

documentation/configuration.md
ledgersteve: @hanspagel Have made those changes and rebased to latest and resolved conflict.

ledgersteve: This branch is now fully rebased on top of latest _main_.  I am getting test fail here on _test-e2e-api-reference_ and I don't know why.  Perhaps someone can help on that.  All the tests pass for me locally.

hanspagel: that’s an issue in main, not your PR 👍

hanspagel (approved):
Thanks for all the time you put into this PR, time to get it in! :)

ledgersteve:
> Thanks for all the time you put into this PR, time to get it in! :)

@hanspagel Thanks for your work too.  Considerably improved on what I originally submitted.  Great result.

PR #6470, opened 2025-08-05
feat(api-client): share header values option

ledgersteve:
**Problem**

With many APIs, there will be a header that is used in many or all of the endpoints.  For example, we work with APIs where the user has to supply a _tenantId_ header value with every call by every method.  

Currently the Scalar API client remembers header values _per method_.  So if you have a _Foo_ method with a _tenantId_ header, if you use the _Foo_ method and set the _tenantId_ for it and then come back to the _Foo_ method later in the same browser session, your _tenantId_ value will be remembered from last time you used the _Foo_ method.  But if you have a _Bar_ method which also takes a _tenantId_ header, that is treated as different to Foo's _tenantId_, so you will have to enter it separately for _Bar_, even though you most likely want it to be the same value as you used for _Foo_.

If you have an API with, say, 50 endpoints, it becomes tedious for the user to have to enter the same _tenantId_ again and again, once for each method.  

Another example for this could be the _Accept_ header - it could apply to all of your API endpoints, and you might want to be able to set it once for all of them.

**Solution**

With this PR, we introduce a new configuration setting: `sharedHeaders`.  This is a nullable string array.  Where method headers are named here, they share their values across all endpoints that have a header of that name.  e.g. in my example above, I set `sharedHeaders: ["tenantId"]`, and then if the user sets a _tenantId_ header value in the API client for one method, the _tenantId_ header value will be populated with that same value for all methods that have a _tenantId_ header.

The `sharedHeaders` config value supports wildcard character `*` at the start and end of the string.  So if I set `sharedHeaders` to `["fooB*"]`, then the value of `fooBar` header value will be shared across all endpoints that have a header of that name, and the `fooBuzz` header value will be shared (separately to `fooBar`) across all endpoints that have a header of that name.

**Checklist**

- [x] I've added tests for the regression or new feature. The tests are limited to defaults for configuration, etc.
- [x] I've updated the documentation - the relevant `.md` file describes what this setting does.

### DIFF
--- a/.changeset/seven-melons-chew.md
+++ b/.changeset/seven-melons-chew.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': minor
+---
+
+Share header values between method calls
--- a/packages/api-client/src/libs/share-header-values.ts
+++ b/packages/api-client/src/libs/share-header-values.ts
@@ -1,0 +1,24 @@
+/* We persist/share header values across all method calls, e.g. if you set the Accept header
+ * for one method using the API Client, then that value will be set for all other methods
+ * that have an Accept header too. This persistence and sharing is done with this class. */
+export const shareHeaderValues = (activeExample: any, allExamples: any) => {
+  const myUid = activeExample.uid
+
+  const otherExamples = allExamples.filter((e: any) => e[0] !== myUid)
+  console.log(`There are ${otherExamples.length} other examples.`)
+  if (otherExamples.length === 0) {
+    return
+  }
+
+  otherExamples.forEach((otherExample: any) => {
+    otherExample[1].parameters.headers.forEach((otherExampleHeader: any) => {
+      activeExample.parameters.headers.forEach((activeExampleHeader: any) => {
+        if (otherExampleHeader.key === activeExampleHeader.key) {
+          if (otherExampleHeader.value !== activeExampleHeader.value) {
+            otherExampleHeader.value = activeExampleHeader.value
+          }
+        }
+      })
+    })
+  })
+}
--- a/packages/api-client/src/views/Request/RequestRoot.vue
+++ b/packages/api-client/src/views/Request/RequestRoot.vue
@@ -12,6 +12,7 @@ import { useSidebar } from '@/hooks/useSidebar'
 import { ERRORS } from '@/libs'
 import { createRequestOperation } from '@/libs/send-request'
 import type { SendRequestResult } from '@/libs/send-request/create-request-operation'
+import { shareHeaderValues } from '@/libs/share-header-values'
 import { validateParameters } from '@/libs/validate-parameters'
 import { usePluginManager } from '@/plugins'
 import { useWorkspace } from '@/store'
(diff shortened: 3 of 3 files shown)


hanspagel (changes requested):
I think the OpenAPI-way is to add headers to the Path Item Object:

```json
{
  "openapi": "3.1.0",
  "info": {
    "title": "Hello World",
    "version": "1.0.0"
  },
  "paths": {
    "/example": {
      "get": {
        "summary": "Example Operation"
      },
      "parameters": [
        {
          "in": "header",
          "name": "X-Custom-Header",
          "schema": {
            "type": "string",
            "example": "foobar"
          }
        }
      ]      
    }
  }
}
```

Does that help already?

Preview: https://sandbox.scalar.com/e/KSdyZ
Related issue: #6472

ledgersteve:
I think we may be talking about different things here.  The Open API snippet supplied (thank you) is about how to add headers to the API description.  That's not what this PR is about.

This PR is saying, "Where multiple endpoints have a header of the same name" (as described in the yaml for this API), we want them to share their values - if the user enters a value for one of them, we want to take care of it so that value is now entered for all headers of that name.

I've created a Loom video of it in action, here:
https://www.loom.com/share/2eb6f15a0000458f928e5b9abb324568?sid=9e822f71-8f2c-4cdd-83e3-2278d8d2403b

hanspagel:
Ah, I think if it’s in the Path Item, we should make sure to keep it in sync for all operations having it. Without additional configuration, because the OpenAPI document tells us this, doesn’t it?

ledgersteve:
To clarify:  We are talking about sharing header values across various methods within an API.  If an API method definition says it uses a header called 'foo', and the person using the api-client uses that method and sets the header 'foo' to value 'bar', we are saying that if the _sharedHeaders_ config item contains the header name 'foo', then when the client user updates the header 'foo' value to be 'bar' in one method call, the 'foo' value should be updated to 'bar' for all method calls - when the user pops up the api client for another method later on, if it has a 'foo' header then it will already be populated with the value 'bar' because they have set that value for the 'foo' header previously, in some other method that has a header of that name.  I believe that is where we are up to so far.  

If I have understood correctly, what you are proposing in your latest comment is that this behaviour should happen for _all_ headers, without needing to specify in the config as to which headers share their values across methods.  If I have understood that correctly, then yes I would support that change - I cannot think of a use case in which the header value set for one particular method should not also be set for all other methods with a header of that name.

hanspagel:
> If I have understood correctly, what you are proposing in your latest comment is that this behaviour should happen for all headers, without needing to specify in the config as to which headers share their values across methods.

yep, if those headers are defined on the path level, not on the operation level (what you call method)

ledgersteve:
Change made per @hanspagel request - header value sharing now applies to all headers all of the time, with no config to change this behaviour.  Done as a squash commit to as to not litter the history with the earlier config stuff.

amritk:
we could also use the environment variables here? @hanspagel

hanspagel:
@amritk I don’t know what you mean by this, let’s chat.

amritk:
@ledgersteve  @hanspagel this is solved in the new store by using a $ref

hanspagel:
NICE

PR #6477, opened 2025-08-05
feat(build-tooling): Migrate bin scripts to .js. They should be dep free

geoffgscott:
Prevents downstream vite configs from impacting the shell script execution.

### DIFF
--- a/.changeset/hungry-tables-rescue.md
+++ b/.changeset/hungry-tables-rescue.md
@@ -1,0 +1,5 @@
+---
+'@scalar/build-tooling': patch
+---
+
+Move to .js bin scripts to avoid vite-node conflicts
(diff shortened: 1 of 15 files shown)


amritk:
doesn't node have TS support now? Still prefer TS over js :)

geoffgscott:
> doesn't node have TS support now? Still prefer TS over js :)

Yeah but only with certain flags and conditions. TBH I preferred these just being shell scripts but good enough for now.

amritk:
> > doesn't node have TS support now? Still prefer TS over js :)
> 
> Yeah but only with certain flags and conditions. TBH I preferred these just being shell scripts but good enough for now.

haha yea that was for windows I think right. classic windows. 

No more flags as of Node v23. But thats not LTS so when we upgrade to Node v24 we can just do `node script.ts`

PR #6478, opened 2025-08-05
chore(api-reference): add test for circular document

amritk:
**Problem**

On the weekend we had a circular document issue come up from a hotfix.

**Solution**

With this PR we add a test for this. We also enable the error console spy on the `ApiReference` tests to check for any new errors that may come up. We should start enabling these spies everywhere.

No need for changeset as its test only

**Checklist**

### DIFF
--- a/packages/api-reference/src/components/ApiReference.test.ts
+++ b/packages/api-reference/src/components/ApiReference.test.ts
@@ -160,7 +160,7 @@ describe('ApiReference', () => {
     })
   })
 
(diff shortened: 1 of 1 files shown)


hanspagel (commented):
I think the setup file is not included when you start the tests from the root, or is it?

amritk:
Yea it still gets hit. Tested it.

Then I realized I run all of my tests from the root lol

PR #6480, opened 2025-08-06
test(components): add snapshot testing to scalar components

hwkr:
Still to do:

- [x] Get tests to run in docker for consistency (I've just been running them on my computer)
- [x] Publish the playwright report somewhere better than as an artifact
- [x] Should add more tests but this seemed like a good starting point

**Checklist**

### DIFF
--- a/.changeset/tame-wombats-crash.md
+++ b/.changeset/tame-wombats-crash.md
@@ -1,0 +1,5 @@
+---
+'@scalar/components': patch
+---
+
+test(components): add snapshot testing to scalar components


packages/components/src/components/ScalarCard/snapshots/base.png
hanspagel, line 1: how cool are those isolated screenshots, eh?

amritk (commented):
Snapshots!

marclave (approved):

packages/components/src/components/ScalarCard/snapshots/base.png
hwkr: Very cool 😎

.github/workflows/ci.yml
```diff
@@ -607,37 +657,25 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
-      - name: Check which files were touched
-        id: changed-files
-        uses: step-security/changed-files@95b56dadb92a30ca9036f16423fd3c088a71ee94 # v46
```
hwkr, line 638: I realized changes in `@scalar/themes` and other dependencies can affect components too so we probably just want to run this all the time?

packages/components/playwright.config.ts
```diff
@@ -0,0 +1,66 @@
...
+      maxDiffPixelRatio: 0.001,
+    },
+  },
+
+  /**
+   * In CI we only need the storybook server because the CI container is the playwright server
+   *
+   * @see https://playwright.dev/docs/ci#via-containers
+   */
+  webServer: CI ? [storybookServer] : [playwrightServer, storybookServer],
```
hwkr, line 59: Because we're running the server (and docker container) from here instead of `pnpm` `Playwright` will automatically wait until there's a connection on both port `5000` and `5001` before trying to run tests.

pnpm-workspace.yaml
```diff
@@ -26,6 +26,7 @@ catalogs:
     '@google-cloud/storage': 7.16.0
     '@headlessui/tailwindcss': ^0.2.2
     '@headlessui/vue': ^1.7.23
+    '@playwright/test': 1.54.0
```
hwkr, line 29: This updates playwright from `1.49` → `1.54` which hopefully won't break anything, cc @hanspagel

hanspagel: Those updates used to break CI (for some super strange reasons that I can’t remember), but if it’s green here, I’ll just take it. 👍

PR #6481, opened 2025-08-06
Fix/does not use window url when the document url is relative

hanspagel:
**Problem**

With 1.32.4 we introduced a regression where the API client doesn’t have a configured server, when the document has a relative server URL *and* the OpenAPI document is a relative URL, too

```
// configuration
url: '/openapi.json'
```

```
// document
servers: [{ url: '/' }]
```

See https://github.com/platformatic/platformatic/issues/4173

**Solution**

This PR is adding a test for it and would have been just another `if` in the code.

But, I wasn’t able to check what the code does anymore. It grew a lot over the last months. So I did what I had to do and moved it to a separate file, added way more tests and refactored it.

The changeset is hard to read now, but the important thing is that `import-spec.test.ts` still passes, with the one additional tests that we added here.

**Checklist**


packages/oas-utils/src/transforms/import-spec.test.ts
```diff
@@ -1455,4 +1455,35 @@ describe('getServersFromOpenApiDocument', () => {
...
+
+    expect(result.servers).toMatchObject([
+      { url: 'http://localhost:3000/' },
+      { url: 'http://localhost:3000/v1' },
+      { url: 'http://localhost:3000/api/v2' },
+    ])
+
+    // Restore the original window.location
+    vi.stubGlobal('location', originalLocation)
+  })
```
hanspagel, line 1488: this is the one additional test that failed before the changes

packages/oas-utils/src/transforms/import-spec.ts
```diff
@@ -184,23 +184,14 @@ export async function importSpecToWorkspace(
...
 
-  // Fallback to the current window.location.origin if no servers are provided
-  if (!collectionServers.length) {
-    const fallbackUrl = getFallbackUrl()
-
-    if (fallbackUrl) {
-      collectionServers.push(serverSchema.parse({ url: fallbackUrl }))
-    }
-  }
-
```
hanspagel, line 203: moved this logic to `getServersFromDocument`

packages/oas-utils/src/helpers/servers.ts
```diff
@@ -0,0 +1,178 @@
+import { type Server, serverSchema } from '@/entities/spec/server'
+import { isDefined } from '@scalar/helpers/array/is-defined'
+import { combineUrlAndPath } from '@scalar/helpers/url/merge-urls'
+import type { ServerObject } from '@scalar/workspace-store/schemas/v3.1/strict/server'
```
hanspagel, line 4: new types while we’re at it anyway

marclave (approved):

PR #6486, opened 2025-08-07
feat(json-magic): change the way we resolve refs

DemonHa:
**Problem**

Our current implementation resolves `$ref` by replacing it with the referenced schema. This approach is incorrect under JSON Schema 2020+ semantics and can lead to loss of information. In the newer specification, `$ref` is not a special case—it behaves like any other keyword, with its own evaluation results, rather than being “resolved” into a single merged schema object.

See #5582 

**Solution**

This PR updates our `$ref` handling to align with JSON Schema 2020+ by treating `$ref` as a regular keyword. When a `$ref` is encountered, we now evaluate it in its own schema evaluation path, preserving its original structure and results. This avoids unintended schema merging and ensures annotations and assertions are combined according to the spec.

**Checklist**


hanspagel (commented):
I think I’ve seen the future. I have few questions how we use this to actually render it, but adding this, is really next level. 👍

hanspagel:
@DemonHa can you add the getter function that Geoff proposed and I take it for a spin next week? might answer some of the questions that I have

DemonHa:
> @DemonHa can you add the getter function that Geoff proposed and I take it for a spin next week? might answer some of the questions that I have

I have added a helper function which by default does not handle merging siblings, but you can provide custom function to handle conflicting siblings

geoffgscott (approved):

PR #6491, opened 2025-08-07
fix(Scalar.Aspire, Scalar.AspNetCore): missing clients and targets

Tholdrim:
**Problem**

Currently, both .NET packages are missing enum values for the Rust language and the Python HTTPX client. As a result, it is not possible to configure Scalar to prevent them from being displayed.

**Solution**

With this PR, I’ve added the missing enum values in both packages. I also updated the corresponding tests to reflect these changes.

**Checklist**

### DIFF
--- a/.changeset/unlucky-hotels-ring.md
+++ b/.changeset/unlucky-hotels-ring.md
@@ -1,6 +1,6 @@
 ---
-'@scalar/aspnetcore': minor
-'@scalar/aspire': minor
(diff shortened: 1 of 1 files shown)


hanspagel:
Thanks for the PR! We should wait for @xC0dex’s review here ✌️

xC0dex (approved):
Awesome PR, thanks!

PR #6492, opened 2025-08-07
refactor(api-reference): schema rendering 💥

amritk:
## 1. Removed discriminator logic
We removed all the discriminator code. We no longer do anything with discriminators EXCEPT indicate its a discriminator and sort it to the top of the list. We not longer have incorrectly highlighted discriminators either. Now we "discriminate" using the `oneOf` instead

| Before | After |
|--------|--------|
| image | image |
| image | image |

## 2. Grab name from `x-original-ref`
We are now grabbing the ref name from `x-original-ref` so we no longer need to prop drill schemas throughout. This is much more accurate with much less code!

| Before | After |
|--------|--------|
| image | image |

TODO: update this as soon as @DemonHa's new ref PR is in

## 3. Removed nested composition merging

Ideally we want to merge the compositions if there's only one property, but that wasn't quite working correctly anyway. Also it was incorrectly merging when we didn't want to, so I removed it. We can add back in the single child merging after.

| Before | After |
|--------|--------|
| image | image |

## 6. Complete revamp of SchemaComposition

Noticed some isssues with the way we were rendering compisitions (anyOf, allOf, oneOf) so I just re-did it in a way that makes more sense.

| Before | After |
|--------|--------|
| image | image |

## 5. Removals

I removed `schema.name` which doesn't seem to be part of the openapi spec or json schema. maybe @hanspagel can fill me on on where this is from. Moving from un-typed -> typed has surfaced a few things

Didn't make the cut for this PR:
- handle refs/sibling properties correctly
- add back in all hidden descriptions

**Checklist**


hanspagel:
I didn’t test, but I just leave this here, we need to test the following specification extensions:

* x-enum-varnames
* x-enum-descriptions
* x-scalar-ignore
* x-additionalPropertiesName

and their aliases

amritk:
Yea we can't really test this until we fix the ref stuff, and the ref stuff is blocked by the extension stuff, and the extension stuff is blocked by type box. redis made a PR!

hanspagel:
pls also adress #6640

amritk:
Closing in favor of rebased one

PR #6494, opened 2025-08-08
feat(components): add hotkey component and tooltip support

hwkr:
Moves `ScalarHotkey` to the component library and adds `ScalarHotkeyTooltip` which is a tooltip with a hotkey.

**Checklist**


geoffgscott:
I can't say this looking that good to me:
image

@cameronrohani I think we should revisit our UX here. What we use in org is much closer to the linear approach:

image

Linear:
image

The blocks are much clearer to read. I also really prefer explicitly callout out `Super` or `Cmd/Ctrl` instead of the caret.

PR #6503, opened 2025-08-11
fix(api-reference): search does not match when the query has a single character word

hanspagel:
**Problem**

I’ve configured the search to not match when a word has 2 or less characters. That seemed resonable to me, but made fuse not match an entry when the query has a 1- or 2-charactrer word, example:

* ❌ "Get **a** token" didn’t match "Get a token"
* ✅ "Get token" did match "Get a token"

image

**Solution**

This PR fixes two things:

* you can use 1- or 2-character in a search query now
* webhooks are indexed with their name

And I’ve added tests just to check the search quality.

**Checklist**

### DIFF
--- a/packages/api-reference/src/features/Search/helpers/create-fuse-instance.ts
+++ b/packages/api-reference/src/features/Search/helpers/create-fuse-instance.ts
@@ -40,10 +40,6 @@ export function createFuseInstance(): Fuse<FuseData> {
     // Include detailed match information showing which parts of the text matched
     includeMatches: true,
 
-    // Minimum number of characters that must match to be considered a result
-    // Prevents single-character matches that are usually noise
-    minMatchCharLength: 2,
-
     // Don't require matches to be at the beginning of strings
     // Makes search more flexible and user-friendly
     ignoreLocation: true,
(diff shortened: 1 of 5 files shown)


packages/api-reference/src/features/Search/helpers/create-fuse-instance.ts
```diff
@@ -40,10 +40,6 @@ export function createFuseInstance(): Fuse<FuseData> {
     // Include detailed match information showing which parts of the text matched
     includeMatches: true,
 
-    // Minimum number of characters that must match to be considered a result
-    // Prevents single-character matches that are usually noise
-    minMatchCharLength: 2,
-
```
hanspagel, line 46: this was causing it

packages/api-reference/src/features/Search/helpers/create-search-index.ts
```diff
@@ -55,7 +55,7 @@ function addEntryToIndex(entry: TraversedEntry, index: FuseData[]): void {
   if ('webhook' in entry) {
     index.push({
       type: 'webhook',
-      title: entry.title,
+      title: entry.name,
```
hanspagel, line 58: webhooks don’t have a title, but a name

marclave (approved):

PR #6510, opened 2025-08-11
fix(api-reference): sidebar links are prefixed with the path

hanspagel:
**Problem**

In environments where we don’t have a regular URL, but a file path (e.g. Electron, VS Code Extension…) our sidebar links might not work.

This is because we’re prefixing them with the path (note the prefixes before the `#`), e.g. 

* `/#tag/authentication/post/user/signup`
* `/path/#tag/authentication/post/user/signup` 
* `index.html#tag/authentication/post/user/signup`

This is mimicking the default browser behaviour for regular URLs, and works for most use cases. It even works with file urls, BUT it doesn’t seem to work with restrictive Content Security Policies and breaks the navigation then. :(

See https://github.com/42Crunch/vscode-openapi/pull/339

**Solution**

This PR removes the path prefix and makes those links just hashes.

* This should do the same for all regular URLs.
* `pathRouting` has a separate handling

Or do I miss anything here?

**Checklist**

### DIFF
--- a/.changeset/unlucky-years-behave.md
+++ b/.changeset/unlucky-years-behave.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
+fix: sidebar links are prefixed with the path


amritk (changes requested):
I believe this may break (needs test) if the references are hosted at a path but using hash routing
ex: `example.com/api-reference#tag/test/operation/get`

hanspagel:
> This looks like a breaking change for users referencing specific operations.

You mean they have a link to a specific operation? The link stays the same, it’s just a different href in the sidebar. The browser adds the hash to the path.

> I believe this may break (needs test) if the references are hosted at a path but using hash routing
ex: example.com/api-reference#tag/test/operation/get

Doesn’t seem to break this case:

* served on `/api-reference`
* `href="#tag/authentication/get/me"`
* browser makes it `/api-reference#tag/authentication/get/me`

Screenshot 2025-08-12 at 12 28 42

marclave (approved):

PR #6512, opened 2025-08-11
fix(api-reference): horizontal scrollbars in a VS Code WebView

hanspagel:
**Problem**

Currently, the API reference has horizontal scrollbars in a VS Code WebView.

Screenshot 2025-08-11 at 17 30 00

**Solution**

The changes in this PR make them go away.

I’ve patched the VS Code integration already, so there’s no rush to get this in here. But I thought we might want to have the fix upstream.

**Checklist**

### DIFF
--- a/.changeset/orange-ties-relax.md
+++ b/.changeset/orange-ties-relax.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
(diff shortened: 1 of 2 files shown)


geoffgscott (approved):
@marclave flagging this one as medium risk for regressions. Tough to tell with all the different device combinations

cameronrohani (approved):

PR #6518, opened 2025-08-11
feat(workspace-store): validate that the bundle produces valid openapi document

DemonHa:
**Problem**

Currently, referenced values within documents are not being validated. This gap can result in invalid references passing through the build process, potentially causing runtime errors.

**Solution**

This PR updates the bundling process to validate all internal references against the OpenAPI schema. By ensuring references are correct after bundling, we prevent runtime errors caused by invalid or unresolved references, and improve overall schema compliance.

**Checklist**


packages/workspace-store/src/navigation/helpers/traverse-paths.ts
```diff
@@ -44,6 +44,8 @@ const createOperationEntry = (
   }
 }
 
+const methods = new Set(['get', 'put', 'post', 'delete', 'patch', 'connect', 'options', 'head', 'trace'])
```
amritk, line 47: We actually have a set already if u want it

https://github.com/scalar/scalar/blob/main/packages/helpers/src/http/http-methods.ts#L8

amritk: should probably freeze it haha

DemonHa:
> Whats up with the loose vs strict? Strict is proxied and loose is not?

Ok so the differences are as follows:

Loose schemas:
- refs in the places that is not allowed
- looser version of the strict schema (no required fields)
- no ref validation

Strict schema:
- used on the document we work with
- schema follows openapi spec
- ref validation

geoffgscott:
@amritk basically the difference is:
- **loose**: Incoming user documents
- **strict**: Validated with defaults applied. Fully OAS complaint; every document in the store will match this. We will probably be slightly less than full OAS compliant when we add SSR support but it will just be the operation level $refs that violate.

PR #6522, opened 2025-08-12
feat(Scalar.Aspire): support multiple scalar resources

### LINKED ISSUE #6424: Aspire integration: Only the first scalar resource is initialized correctly
### What happens?

I'm registering two Scalar resources in my Aspire app host, but the only the first container is accessible as expected.
The second Scalar container app just shows `API Reference configuration is not provided. Please set the API_REFERENCE_CONFIG environment variable.`

I think the cause might be the [`ScalarHook`](https://github.com/scalar/scalar/blob/main/integrations/aspire/src/Scalar.Aspire/ScalarHook.cs#L21) that only configures the first resource.  
Are multiple Scalar resources not supported or is this an oversight?

I'm aware that I could just reuse a single Scalar resource, but because the apphost contains two _mostly_ separate applications I would like to have separate Scalar APIs for them as well, even if they are only for development anyways.

All registered Scalar resources are accessible and configured as expected.

### OpenAPI Document

_No response_

xC0dex:
**Problem**

Currently, it's not possible to run multiple scalar instances in .NET Aspire.

**Solution**

Now every resource will be configured, not only the first one.

Closes #6424 

**Checklist**

### DIFF
--- a/.changeset/honest-flies-burn.md
+++ b/.changeset/honest-flies-burn.md
@@ -1,0 +1,5 @@
+---
+'@scalar/aspire': patch
+---
+
+feat: support multiple scalar resources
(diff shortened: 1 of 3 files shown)


xC0dex:
I'll add comprehensive Aspire integration tests in a dedicated PR.

PR #6256, opened 2025-07-15
feat(fastapi): new theme option

hanspagel:
**Problem**

Currently, we don’t support themes in the FastAPI integration.

**Solution**

This PR adds support for themes, including a whole bunch of tests.

Fixes #4157

**Checklist**

I've gone through the following:

- [x] I've added an explanation _why_ this change is needed.
- [x] I've added a changeset (`pnpm changeset`).
- [x] I've added tests for the regression or new feature.
- [ ] I've updated the documentation.

PR #6526, opened 2025-08-12
Fix fastapi theme

zrquan:
**Problem**

After #6256, we don't have to use hardcoded theme styles anymore. Also, if we set the `theme` but not the `scalar_theme`, the hardcoded style will take priority and make the theme ineffective.

**Solution**

With this PR …

**Checklist**

### DIFF
--- a/.changeset/shaggy-chicken-refuse.md
+++ b/.changeset/shaggy-chicken-refuse.md
@@ -1,0 +1,5 @@
+---
+'scalar-fastapi': patch
+---
+
+Remove the hardcoded scalar_theme
--- a/integrations/fastapi/scalar_fastapi/scalar_fastapi.py
+++ b/integrations/fastapi/scalar_fastapi/scalar_fastapi.py
@@ -55,103 +55,6 @@ class Theme(Enum):
     NONE = "none"
 
 
-scalar_theme = """
-/* basic theme */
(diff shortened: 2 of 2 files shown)


hanspagel:
Wait, can you elaborate on the issue? We definitely want to keep a custom theme for FastAPI as the default. There must be another way to fix the issue.

zrquan:
> Wait, can you elaborate on the issue?

I've tried setting the `theme` option, but the theme doesn't change. It's still the default theme.
```python
@api.get("/scalar", include_in_schema=False)
async def scalar_html():
    return get_scalar_api_reference(
        openapi_url=api.openapi_url,
        title=api.title,
        theme=Theme.SOLARIZED,  # not work
    )
```

When I empty the `scalar_theme`, the theme changes like it should.
```python
@api.get("/scalar", include_in_schema=False)
async def scalar_html():
    return get_scalar_api_reference(
        openapi_url=api.openapi_url,
        title=api.title,
        scalar_theme="",
        theme=Theme.SOLARIZED,  # work now
    )
```

hanspagel:
Ah, got it! Yeah, there’s a lot of strange things going on. 

Fixing it…

hanspagel:
Closing in favor of #6552 ✌️

PR #6530, opened 2025-08-12
refactor(api-reference): use store for the introduction section

hanspagel:
**Problem**

We’re using the dereferenced document for the introduction section, but we want to use the store for everything.

**Solution**

This PR refactors the components to use the store as input. :)

**Future Tasks**

- Don’t use the API client store anymore.

**New Components**

```vue
<template>
  <!-- … -->

  <!-- Introduction -->
  <IntroductionSection :showEmptyState="!store.workspace.activeDocument">
    <ScalarDocumentInfo
      :id
      :document="store.workspace.activeDocument"
      :layout="config.layout"
      :oasVersion
      :isLoading="config.isLoading"
      :onLoaded="config.onLoaded">
      <template #selectors>
        <ScalarErrorBoundary>
          <ScalarServerSelector :config="config" />
        </ScalarErrorBoundary>
        <ScalarErrorBoundary>
          <ScalarAuthSelector :config="config" />
        </ScalarErrorBoundary>
        <ScalarErrorBoundary>
          <ScalarClientSelector
            :config="config"
            :clientOptions="clientOptions"
            :selectedClient="store.workspace['x-scalar-default-client']"
            :document="store.workspace.activeDocument" />
        </ScalarErrorBoundary>
      </template>
    </ScalarDocumentInfo>

    <template #empty-state>
      <slot name="empty-state" />
    </template>
  </IntroductionSection>

  <!-- … -->
</template>
```

**Checklist**


packages/api-reference/src/components/Content/Content.vue
```diff
@@ -34,17 +39,31 @@ const clientOptions = computed(() =>
...
     <!-- Introduction -->
-    <Introduction
-      v-if="document?.info?.title || document?.info?.description"
-      :document
-      :store
-      :clientOptions
-      :config />
+    <IntroductionSection :showEmptyState="!store.workspace.activeDocument">
+      <ScalarDocumentInfo
+        :document="store.workspace.activeDocument"
```
geoffgscott, line 64: We should be able to only pass in the info field here and not a whole document.

packages/api-reference/src/v2/blocks/scalar-document-info/ScalarDocumentInfo.vue
```diff
@@ -0,0 +1,36 @@
+<script setup lang="ts">
+import type { ApiReferenceConfiguration } from '@scalar/types'
+import type { OpenApiDocument } from '@scalar/workspace-store/schemas/v3.1/strict/openapi-document'
+import { computed } from 'vue'
+
+import { IntroductionCard } from '@/components/IntroductionCard'
```
geoffgscott, line 5: These components should be in the V2 folder if that is the only place they are used. They all looks like pure render elements for the Info.

packages/api-reference/src/v2/blocks/scalar-server-selector/ScalarServerSelector.vue
```diff
@@ -0,0 +1,54 @@
...
+
+import { IntroductionCardItem } from '@/components/IntroductionCard'
+import { BaseUrl } from '@/features/base-url'
+
+const { config } = defineProps<{
+  config?: ApiReferenceConfiguration
+}>()
+
+const { collections, servers } = useWorkspace()
+const { activeCollection: _activeCollection } = useActiveEntities()
```
geoffgscott, line 16: None of this logic that uses old state management should live in the V2 folder. @DemonHa is working on migrating the selectors but we can use this interim solution for now. Just gotta put it out of V2.

packages/api-reference/src/v2/blocks/scalar-document-info/IntroductionLayout.vue
```diff
@@ -14,43 +12,27 @@ import {
...
+import InfoVersion from './InfoVersion.vue'
+import OpenApiVersion from './OpenApiVersion.vue'
 
 const { document, config } = defineProps<{
-  document: OpenAPIV3_1.Document
+  document: OpenApiDocument
   config?: ApiReferenceConfiguration
 }>()
 
 const { getHeadingId } = useNavState()
```
geoffgscott: No external hooks in these components. Gotta be pure render. @amritk do we have a current solution for moving to something more functional? Feels like all IDs can be deterministic so we shouldn't need a hook.

@hanspagel for now maybe we just pass the ID in as a prop

packages/api-reference/src/v2/blocks/scalar-document-info/IntroductionLayout.vue
```diff
@@ -14,43 +12,27 @@ import {
...
 
-import Description from './Description.vue'
+import InfoDescription from './InfoDescription.vue'
+import InfoLinks from './InfoLinks.vue'
+import InfoVersion from './InfoVersion.vue'
+import OpenApiVersion from './OpenApiVersion.vue'
 
 const { document, config } = defineProps<{
-  document: OpenAPIV3_1.Document
+  document: OpenApiDocument
```
geoffgscott, line 23:
```suggestion
  info: OpenApiDocument['info]
  externalDocs: OpenApiDocument['externalDocs']
```

packages/api-reference/src/v2/blocks/scalar-auth-selector/ScalarAuthSelector.vue
geoffgscott, line 1: Same comment as the other selector; we should have anything using the old state management out of the V2 folder.

packages/api-reference/src/components/Content/Content.vue
```diff
@@ -34,17 +39,31 @@ const clientOptions = computed(() =>
...
 
     <!-- Introduction -->
-    <Introduction
-      v-if="document?.info?.title || document?.info?.description"
-      :document
-      :store
-      :clientOptions
-      :config />
+    <IntroductionSection :showEmptyState="!store.workspace.activeDocument">
+      <ScalarDocumentInfo
```
geoffgscott, line 62: Commented below as well but we shouldn't be passing the full config in to any blocks. Just the required options (in this case just layout)

packages/api-reference/src/components/Content/Content.vue
hanspagel: we need the following root level properties though:

* `info`
* `externalDocs`
* `x-*` (any custom specification extension attribute on the root level)

felt easier to pass the document here. wdyt?

packages/api-reference/src/v2/blocks/scalar-document-info/ScalarDocumentInfo.vue
hanspagel: the `IntroductionCard` and `IntroductionCardItem` is used in multiple blocks, that’s why I put them in the components folder

I’ll check if I can refactor it, so it’s not used in multiple blocks.

packages/api-reference/src/v2/blocks/scalar-server-selector/ScalarServerSelector.vue
hanspagel: yeah, so should we keep it outside or just ff with the @DemonHa’s PR?

packages/api-reference/src/v2/blocks/scalar-document-info/IntroductionLayout.vue
hanspagel: yeah, we need any custom specification extension on the root level too, though

packages/api-reference/src/components/Content/Content.vue
hanspagel: yeah, let’s fix this in another PR, kay?

packages/api-reference/src/v2/blocks/scalar-document-info/ScalarDocumentInfo.vue
hanspagel: not as easy actually

packages/api-reference/src/v2/blocks/scalar-server-selector/ScalarServerSelector.vue
hanspagel: Ah, or we move it out of the blocks and pass it as a prop? (as long as we need those)

packages/api-reference/src/components/Content/Content.vue
hanspagel: I just decoupled it, so we prop drill the relevant parts of the config now

geoffgscott:
@hanspagel Since you are out next week anyways we can let @DemonHa get his selector work in and then come back to this.

DemonHa:
Closing in regards to #6619

PR #6539, opened 2025-08-13
fix(openapi-parser): ensure upgrader doesn't lose schema properties

### LINKED ISSUE #6537: Upgrader removes schema properties
### What happens?

When using the upgrader

```json
{
  "type": "object",
  "properties": {
    "file": {
      "type": "string",
      "description": "Some fun things",
      "format": "binary"
    }
  }
}
```

becomes 
```json
{
  "type": "object",
  "properties": {
    "file": {
      "type": "string",
      "contentMediaType": "application/octet-stream"
    }
  }
}
```

we are removing the old schema properties, we should add a `...schema` spread to all of those returns

https://github.com/scalar/scalar/blob/main/packages/openapi-parser/src/utils/upgrade-from-three-to-three-one.ts

### OpenAPI Document

_No response_

amritk:
**Problem**

closes #6537

Currently, the upgrader will remove all of the old schema properties.

**Solution**

With this PR we only remove format and keep the rest.

**Checklist**

### DIFF
--- a/packages/openapi-parser/src/utils/upgrade-from-three-to-three-one.ts
+++ b/packages/openapi-parser/src/utils/upgrade-from-three-to-three-one.ts
@@ -121,9 +121,13 @@ const applyChangesToDocument = (schema: UnknownObject, path: string[]) => {
     return {}
   }
 
+  // 6. Handle older formats
+  const { format: _, ...rest } = schema
+
   if (schema.type === 'string') {
     if (schema.format === 'binary') {
       return {
+        ...rest,
         type: 'string',
         contentMediaType: 'application/octet-stream',
       }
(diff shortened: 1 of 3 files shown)


packages/openapi-parser/src/utils/upgrade-from-three-to-three-one.ts
```diff
@@ -121,16 +121,21 @@ const applyChangesToDocument = (schema: UnknownObject, path: string[]) => {
     return {}
   }
 
+  // 6. Handle older formats
+  const { format: _, ...rest } = schema
```
geoffgscott, line 125: Is this actually valid though? Are old formats able to be used 1:1? If we do proceed we should probably put a specific JSON schema version locally or at the top of the document.

amritk: All this does is keep the old properties instead of deleting them. Like description

geoffgscott (approved):

PR #6540, opened 2025-08-13
feat(openapi-parser): join multiple openapi documents

DemonHa:
**Summary**

This PR introduces a new `join` function that merges multiple OpenAPI documents into a single file. Unlike the existing bundle command—which resolves and embeds all external references into one document—the join command simply combines multiple OpenAPI files into a single, valid OpenAPI specification.

Key Points
- New Feature: join command for merging multiple OpenAPI documents.

Difference from bundle:
- bundle → consolidates a document and inlines all external $ref values.
- join → merges multiple separate OpenAPI documents into one.

**Use Case**
This is useful when maintaining modular API specifications across separate files but needing a unified OpenAPI document for distribution, documentation, or tooling.

**Checklist**

### DIFF
--- a/packages/openapi-parser/src/utils/join/join.test.ts
+++ b/packages/openapi-parser/src/utils/join/join.test.ts
@@ -35,10 +35,6 @@ describe('join', () => {
           summary: 'Some extra information about the API',
         },
         paths: {},
-        webhooks: {},
-        tags: [],
-        servers: [],
-        components: {},
       },
     })
   })
@@ -89,10 +85,6 @@ describe('join', () => {
             },
           },
         },
-        webhooks: {},
-        tags: [],
-        servers: [],
-        components: {},
       },
     })
   })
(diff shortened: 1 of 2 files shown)


packages/openapi-parser/src/utils/join/join.test.ts
```diff
@@ -0,0 +1,982 @@
...
+          title: 'Main title',
+          version: '1.1.1',
+          summary: 'Some extra information about the API',
+        },
+        paths: {},
+        webhooks: {},
+        tags: [],
+        servers: [],
+        components: {},
+      },
```
hanspagel: can we make it so, it doesn’t add webhooks, tags, servers and components even if they are empty?

according to the specification the required attributes are

* `openapi`
* `info.title`
* `info.version`
* `paths`

hanspagel (approved):
SICK

just left one nitpick and update one test title

marclave (approved):

PR #6541, opened 2025-08-13
fix(components): add scalar fonts to storybook

hwkr:
Adds the Scalar fonts to Storybook. I kind of can't believe I didn't notice they weren't in there; I think they maybe used to be and got removed by accident. If we'd had snapshot testing back then we would have caught it 😂

This makes the fonts available to Storybook and probably more importantly to Playwright.

CleanShot 2025-08-12 at 23 07 46@2x

**Checklist**

### DIFF
--- a/.changeset/rich-dogs-design.md
+++ b/.changeset/rich-dogs-design.md
@@ -1,0 +1,5 @@
+---
+'@scalar/components': patch
+---
+
+fix(components): add scalar fonts to storybook


hanspagel (approved):
lol, I was looking at the snapshots and just thought it’s the Linux font rendering, but the snapshots look way better now

hwkr:
> lol, I was looking at the snapshots and just thought it’s the Linux font rendering, but the snapshots look way better now

Yeah same here 😂

PR #6552, opened 2025-08-13
feat(fastapi): new HTML/JS API

hanspagel:
**Problem**

In #6526 @zrquan reported a bug, where you can’t overwrite the default FastAPI theme. I wanted to fix that, but was shocked we're still using the old JS API.

**Solution**

This PR refactors the integration to use the new HTML/JS API. This also fixes the mentioned bug, where the theme can’t be overwritten.

**Checklist**

### DIFF
--- a/.changeset/warm-teachers-end.md
+++ b/.changeset/warm-teachers-end.md
@@ -1,0 +1,5 @@
+---
+'scalar-fastapi': minor
+---
+
+feat: use new HTML/JS API


geoffgscott (approved):
Untested but the code looks good

hanspagel:
The current release has a few painful regressions, I’ve tested it manually and it has tests and I think @marclave is too busy to review rn, so I'll just merge it and queue it up for release.

PR #6556, opened 2025-08-13
feat(workspace-store): move extensions to typebox + store

amritk:
**Problem**

Currently, we had the extensions everwhere in our schemas typed as unknown

**Solution**

With this PR:
- we parse the extensions as needed and have them fully typed.
- converted all the extensions from zod to typebox
- use the extensions in the code where we are using the new store

**Checklist**


hanspagel (approved):
awesome!!

CI isn’t passing yet, though

PR #6557, opened 2025-08-13
feat(workspace-store): preprocess documents to make them compliant

DemonHa:
**Problem**

Currently, no cleanup is performed on incoming OpenAPI documents. As a result, some schemas are missing the "type": "object" property, which can lead to invalid or non-compliant JSON Schema documents.

**Solution**

This PR introduces a preprocessing step on document ingress to handle these cases. It ensures that schemas with properties but missing "type": "object" are updated to be fully JSON Schema compliant, improving validation and downstream processing.

**Checklist**


DemonHa:
@hanspagel Is there any other scenario we need to handle?

hanspagel (approved):
this is so good

I think there’s a two ways to define something as nullable

```
type: 'string'
nullable: true
```

or, basically a constant

```
type: 'null'
```

or a nullable string (same as the first example)

```
type: ['string', 'null']
```

I think we should stick to one of those. the `'null'` type was introduced in OpenAPI 3.1, so I consider this the modern way to do mark something as nullable.

I’ve never seen it in the wild, but if something has `pattern` and no `type`, we can assume it’s a string.

We can add more as we go.

marclave (approved):

amritk:
I think 
```ts
nullable: true
```
was deprecated

amritk (approved):
send it

PR #6558, opened 2025-08-14
feat(components): update scalar hotkey design

hwkr:
**Problem**

Currently, …

**Solution**

With this PR …

**Checklist**


cameronrohani:
looks good, can we stick to border colour for the borders tho? cc @hwkr 
image

cameronrohani (changes requested):
make border colours our default border colours instead of text colour

hwkr:
@cameronrohani Because the tooltips are always dark we would need to introduce another theme variable just for the tooltip hotkey border, or we can just make it a lower opacity version of the text color?

cameronrohani:
> @cameronrohani Because the tooltips are always dark we would need to introduce another theme variable just for the tooltip hotkey border, or we can just make it a lower opacity version of the text color?

we can't just force dark-mode similiar to code block

if not let's just do a color-mix with the text color 👀

cameronrohani (approved):

PR #6564, opened 2025-08-14
docs: reorder sidebar

hanspagel:
This PR is a proposal to reorder the sidebar of the deployed docs:

Screenshot 2025-08-14 at 14 27 22

- First, an introduction,
- then the Registry (where everything should start)
- then Scalar Docs (most popular product I guess)
- then SDKs
- then the CLI for a deeper integration
- then Access for advanced use cases
- and then the Scalar API Reference

And inside the API Reference:

* Getting Started
* Configure it
* OpenAPI extensions (moved the plugins for specification extensions to the bottom, as something more advanced)
* Customize it (themes)
* Markdown support
* And if that’s not enough: Plugins

⚠️ And I removed the Scalar Access API Reference. Or did we have that on purpose? It seemed a bit confusing to me.

### DIFF
--- a/documentation/openapi.md
+++ b/documentation/openapi.md
@@ -1,12 +1,8 @@
-# OpenAPI Specification
+# OpenAPI Extensions
 
 We're expecting the passed OpenAPI document to adhere to [the Swagger 2.0, OpenAPI 3.0 or OpenAPI 3.1 specification](https://github.com/OAI/OpenAPI-Specification).
 
 On top of that, we've added a few things for your convenience:
-
-## Custom Specification Extensions
-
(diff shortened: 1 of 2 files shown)


geoffgscott:
I like this. Not sure if its there already but having a page link to API References from the intro section would be good.

cameronrohani:
> I like this. Not sure if its there already but having a page link to API References from the intro section would be good.

almost done redoing this page, I think there's like 3 page links to that page incoming 😅

marclave (changes requested):
@hanspagel can you double check the dashboard and make sure we arent breaking any links :) 

will review monday

marclave:
I think we should re-order once we finish the landing page work <3 

can re-open later

PR #6569, opened 2025-08-15
feat(api-reference): server selector block

DemonHa:
**Checklist**

We are migrating the current components to consume and emit events from the new store


packages/api-reference/src/components/Content/Introduction/Introduction.vue
```diff
@@ -78,11 +84,14 @@ const { hash } = useNavState()
...
               class="scalar-reference-intro-server scalar-client introduction-card-item text-base leading-normal [--scalar-address-bar-height:0px]">
-              <BaseUrl
-                :collection="activeCollection"
-                :server="activeServer" />
+              <ServerSelector
+                :servers="store.workspace.activeDocument?.servers ?? []"
+                :xSelectedServer="
+                  store.workspace.activeDocument?.['x-scalar-active-server'] ??
+                  ''
+                " />
```
geoffgscott: Let's do this as a slot so all the store logic can live higher up. We may need to hoist the slot a couple levels.

packages/api-client/src/components/CommandPalette/CommandPaletteServer.vue
```diff
@@ -77,6 +78,13 @@ const handleSubmit = () => {
   // Select the server
   collectionMutators.edit(collection.uid, 'selectedServerUid', server.uid)
 
+  emitCustomEvent(wrapper.value?.$el!, 'scalar-add-server', {
```
amritk:
```suggestion
  emitCustomEvent(wrapper.value?.$el, 'scalar-add-server', {
```
nit: not a big fan of using the `!` everywhere, it should be used sparingly. Instead an we just make the element `| null` and `?.dispatchEvent`

packages/api-client/src/components/CommandPalette/CommandPaletteServer.vue
```diff
@@ -92,9 +100,12 @@ const handleSubmit = () => {
 const redirectToCreateCollection = () => {
   events.commandPalette.emit({ commandName: 'Create Collection' })
 }
+
+const wrapper = useTemplateRef('wrapper')
```
amritk: @hwkr was there an issue with using the same name here?

packages/api-reference/src/components/Content/Introduction/Introduction.vue
```diff
@@ -77,13 +76,7 @@ const { hash } = useNavState()
...
             class="introduction-card"
             :class="{ 'introduction-card-row': config?.layout === 'classic' }">
-            <div
-              v-if="activeCollection?.servers?.length"
-              class="scalar-reference-intro-server scalar-client introduction-card-item text-base leading-normal [--scalar-address-bar-height:0px]">
-              <BaseUrl
-                :collection="activeCollection"
-                :server="activeServer" />
-            </div>
+            <slot name="serverSelector"></slot>
```
amritk, line 74: Any reason to slot this in? I believe we were slotting it before then it got moved to inside the component to clean up `Content.vue`

cc: @hanspagel

packages/api-reference/src/components/ApiReferenceLayout.vue
```diff
@@ -300,6 +301,11 @@ watch(hash, (newHash, oldHash) => {
   }
 })
 
+const root = ref(window.document.body)
```
amritk:
```suggestion
const rootRef = useTemplateRef('root')
```
I wouldn't put anything on window.document.body in the case of multiple references/clients on a single page!

amritk: if you need to add it to the same element as the other listeners we can just pass the element into this component

packages/api-reference/src/features/api-client-modal/ApiClientModal.vue
```diff
@@ -132,19 +127,12 @@ watchDebounced(
           },
         )
 
-        // Add the new ones
-        newServers.forEach((server) => {
-          store.serverMutators.add(server, collection.uid)
+        emitCustomEvent(el.value!, 'scalar-replace-servers', {
```
amritk: We still need the old code here as well don't we? unless its synced below

packages/api-reference/src/v2/blocks/scalar-server-selector-block/components/Selector.vue
```diff
@@ -0,0 +1,88 @@
...
+const { target, servers, xSelectedServer } = defineProps<{
+  /** The selected server URL */
+  xSelectedServer?: string
+  /** Available servers */
+  servers: ServerObject[]
+  /** The id of the target to use for the popover (e.g. address bar) */
+  target: string
+}>()
+
+const emit = defineEmits<{
```
amritk, line 16: if we are on the new store we can skip the emit and just send the custom event no?

packages/api-reference/src/v2/blocks/scalar-server-selector-block/components/Selector.vue
```diff
@@ -0,0 +1,88 @@
...
+const serverOptions = computed(() =>
+  servers.map((server) => ({
+    id: server.url,
+    label: server.url,
+  })),
+)
+
+const server = computed(() => {
+  return servers.find((s) => s.url === xSelectedServer)
+})
```
amritk:
```suggestion
const server = computed(() => 
  servers.find((s) => s.url === xSelectedServer)
)
```
no need for returns!

packages/api-reference/src/v2/blocks/scalar-server-selector-block/components/Selector.vue
```diff
@@ -0,0 +1,88 @@
...
+
+const serverUrlWithoutTrailingSlash = computed(() => {
+  if (server.value?.url.endsWith('/')) {
+    return server.value?.url.slice(0, -1)
+  }
+  return server.value?.url || ''
```
amritk:
```suggestion
const serverUrlWithoutTrailingSlash = computed(() => 
  server.value?.url?.replace(/\/$/, '') || ''
)
```
nit: cleaner with regex

packages/workspace-store/src/events/definitions.ts
```diff
@@ -0,0 +1,145 @@
...
+
+export type ApiReferenceEvent = Prettify<keyof ApiReferenceEvents>
+
+/**
+ * Scalar blocks will use vanilla events to allow more flexibility in integrations
+ *
+ * Event can include typed payloads using the `data` property. A target for the dispatch must be provided.
+ */
+export function emitCustomEvent<E extends ApiReferenceEvent>(
+  target: HTMLElement,
```
amritk:
```suggestion
  target: HTMLElement | null,
```

packages/workspace-store/src/events/definitions.ts
```diff
@@ -0,0 +1,145 @@
...
+  detail: ApiReferenceEvents[E]['detail'],
+) {
+  const instance = new CustomEvent(event, {
+    detail: detail,
+    bubbles: true,
+    composed: true,
+    cancelable: true,
+  })
+
+  target.dispatchEvent(instance)
```
amritk:
```suggestion
  target?.dispatchEvent(instance)
```

packages/workspace-store/src/events/definitions.ts
```diff
@@ -0,0 +1,145 @@
...
+}
+
+/** Type helper for expanding complex types */
```
amritk: import from type-fest

https://github.com/sindresorhus/type-fest/blob/main/source/simplify.d.ts

packages/workspace-store/src/mutators/server.ts
```diff
@@ -25,11 +25,13 @@ export const serverMutators = (target?: ServerObject[]) => {
...
    */
-  const deleteServer = (index: number): boolean => {
+  const deleteServer = (url: string): boolean => {
     if (!target) {
       return false
     }
-    target.splice(index, 1)
+    const newTarget = [...target.filter((it) => it.url !== url)]
+    target.splice(0, target.length)
+    target.push(...newTarget)
```
amritk, line 34: We can't just re-assign the array?
```ts
target = target.filter((it) => it.url !== url)
```

if not save one step
```ts
const newTarget = [...target.filter((it) => it.url !== url)]
target.splice(0, target.length, ...newTarget)
````

packages/api-reference/src/features/api-client-modal/ApiClientModal.vue
amritk: I think its synced below

packages/api-reference/src/components/Content/Introduction/Introduction.vue
DemonHa: Store logic should be living higher up in the tree cc @geoffgscott

packages/api-reference/src/v2/blocks/scalar-server-selector-block/components/Selector.vue
DemonHa: we emit events only from the main file for the block, for the pure components we use vue events

packages/api-reference/src/features/api-client-modal/ApiClientModal.vue
DemonHa: useLegacyStore will listen to that event and update the old store

packages/workspace-store/src/mutators/server.ts
DemonHa: it will not update the reference, we should either do this or get the parent and assign the property

packages/api-reference/src/components/Content/Introduction/Introduction.vue
geoffgscott: @amritk each of these blocks is atomic. Someone should be able to use the Intro block without the selectors if they choose. 

And also the store logic only lives in the one file. Everything else is partial documents and events.

amritk: sure the reason I said that is thats how it was I think, then it changed to like this, now we change back 😆

DemonHa:
> Good to go on my end. @geoffgscott what did you mean by we don't want async events? Isn't the whole system based on async events now?

I think he means sync events

PR #6570, opened 2025-08-15
fix(api-reference): apply reference helper to all possible references

amritk:
**Problem**

Currently, we updated how the references work in the workspaces store but not the front-end.

**Solution**

With this PR we: 
- apply the new helper
- fix up some types
- ~added a deep ref resolver which we need any time we display code~
- got rid of the deep resolver and updated the example generator. The issue its also used by the client + markdown generator BUT the only thing affected is the types so I added a bunch of `// @ts-ignore` so we can go back and fix them whenever they get updated to the new store.

Needs thorough testing to make sure I didn't miss any refs

**Checklist**

### DIFF
--- a/packages/api-client/src/views/Request/RequestSection/helpers/filter-security-requirements.test.ts
+++ b/packages/api-client/src/views/Request/RequestSection/helpers/filter-security-requirements.test.ts
@@ -46,9 +46,14 @@ describe('filterSecurityRequirements', () => {
     expect(result).toEqual([])
   })
 
-  it('should return empty array when no security schemes are selected', () => {
+  it('should return the first security scheme when no security schemes are selected', () => {
     const result = filterSecurityRequirements([{ bearerAuth: [] }], [], mockSecuritySchemes)
-    expect(result).toEqual([])
+    expect(result).toEqual([mockSecuritySchemes.bearerAuthUid])
+  })
+
+  it('should return the first non-optional security scheme when no security schemes are selected', () => {
+    const result = filterSecurityRequirements([{}, { bearerAuth: [] }], [], mockSecuritySchemes)
+    expect(result).toEqual([mockSecuritySchemes.bearerAuthUid])
   })
 
   it('should filter single security requirement correctly', () => {
--- a/packages/api-client/src/views/Request/RequestSection/helpers/filter-security-requirements.ts
+++ b/packages/api-client/src/views/Request/RequestSection/helpers/filter-security-requirements.ts
@@ -1,5 +1,4 @@
 import type { Collection, SecurityScheme } from '@scalar/oas-utils/entities/spec'
-import { isDefined } from '@scalar/oas-utils/helpers'
 
 /**
  * Takes in selected security and filters it with the requirements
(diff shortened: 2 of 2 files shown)


packages/api-reference/src/features/Operation/helpers/combine-params.test.ts
```diff
@@ -44,18 +44,25 @@ describe('combineParams', () => {
   it('filters out unresolved references', () => {
     const pathParams: any[] = [
       { name: 'id', in: 'path', required: true },
-      { $ref: '#/components/parameters/UnresolvedParam' }, // Unresolved reference
+      {
+        $ref: '#/components/parameters/UnresolvedParam',
+        '$ref-value': { name: 'username', in: 'query', required: false },
+      },
     ]
```
hanspagel, line 51: we expected an unresolved reference first and now expect a resolved reference?

geoffgscott (approved):
I'm happy with this pending @DemonHa testing and approving. 

Great work!

packages/api-reference/src/features/Operation/helpers/combine-params.test.ts
amritk: this is an unresolved reference in this context actually

packages/workspace-store/src/helpers/get-resolved-ref-deep.ts
DemonHa, line 1: can we try to remove this and handle every place one by one?

packages/workspace-store/src/helpers/get-resolved-ref-deep.ts
```diff
@@ -0,0 +1,65 @@
...
+    // First resolve this level
+    const resolved = getResolvedRef(currentNode)
+
+    // If it's not an object, return as-is
+    if (typeof resolved !== 'object' || resolved === null) {
+      return resolved as DeepDereference<Node>
+    }
+
+    // Check if we've already processed this object (cache hit)
+    if (cache.has(resolved)) {
```
DemonHa: we need to working with the raw object here, otherwise cache will never detect circular references

packages/workspace-store/src/helpers/get-resolved-ref-deep.ts
amritk: yes

packages/api-reference/src/features/example-responses/helpers/get-resolved-ref-deep.ts
```diff
@@ -0,0 +1,66 @@
...
+
+/**
+ * Recursively resolves all $ref objects in a data structure to their actual values.
+ * Traverses through objects, arrays, and nested structures to find and resolve
+ * any $ref references at any depth level.
+ *
+ * Handles circular references gracefully by detecting them and returning '[circular]'
+ * to prevent infinite loops.
+ */
+export const getResolvedRefDeep = <Node>(node: NodeInput<Node>): any => {
```
DemonHa: the output type of this should be: Node | '[circular]'

DemonHa (approved):
let's get it in 🔥

packages/api-reference/src/features/document-source/hooks/useDocumentFetcher.ts
```diff
@@ -49,9 +48,7 @@ const getContent = async ({ url, content }: SpecConfiguration, proxyUrl?: string
   // Fetch from URL only if we do not already have the content
   if (url && !content) {
     try {
-      const result = await measure(`fetch(${url})`, async () => await fetchDocument(url, proxyUrl))
```
amritk, line 52: got rid of these measures as they are moving to the workspace store

cameronrohani:
great work 🫡

PR #6573, opened 2025-08-15
feat: Add typing marker

rhuanbarreto:
This will tell linters that the code is typed and should be used as is according to PEP-561 (https://peps.python.org/pep-0561/).

**Problem**

Currently, type checkers require type stubs from the package to be published.

**Solution**

With this PR type checkers like mypy or pylance won't require type stubs as the code is properly typed.

**Checklist**


hanspagel (approved):
nice

rhuanbarreto:
@hanspagel changeset fixed 😀

rhuanbarreto:
@hanspagel can you release the version to pypi?

PR #6578, opened 2025-08-15
feat(workspace-store): support relative external references

DemonHa:
**Problem**

In the client store, relative external references are currently unsupported. This occurs because the document is passed to the bundler as an already resolved object, causing the bundler to lose the original document origin. As a result, relative references are resolved incorrectly.

Additionally, the magic proxy throws an error when attempting to write to a `$ref-value`, as it is considered a virtual property. This behavior can be problematic when a `$ref` points to a primitive during document ingress and coercion to the correct value is required.

**Solution**

This PR implements proper tracking of document origins and passes this information to the bundler, ensuring relative references are resolved against the correct origin.

It also updates the proxy handling logic with the following changes:

- Prevents overriding a `$ref-value` that points to the root document (throws an error).
- Throws an error when attempting to override an invalid $ref-value.
- Allows writing to $ref-value in all other valid cases.

**Checklist**


amritk (approved):
Did not test but there's tests!

PR #6590, opened 2025-08-18
feat(workspace-store): performance improvements

DemonHa:
**Problem**

Document processing on ingress is currently too slow, causing noticeable delays in loading. For example, loading the `Zoom` document previously took ~2291ms, which negatively impacts user experience and system performance.

**Solution**

This PR introduces performance optimizations that significantly reduce processing time. After the improvements, loading a Zoom document now takes ~196ms, representing nearly a 90% reduction in latency.

**Checklist**

### DIFF
--- a/packages/workspace-store/src/helpers/deep-clone.ts
+++ b/packages/workspace-store/src/helpers/deep-clone.ts
@@ -31,7 +31,7 @@ export const deepClone = <T>(value: T, hash = new WeakMap()): T => {
 
   Object.keys(value).forEach((key) => {
     // @ts-expect-error: Index signature for generic object
-    result[key] = deepClone((value as any)[key], hash)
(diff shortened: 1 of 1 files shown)


packages/workspace-store/src/helpers/deep-clone.ts
```diff
@@ -0,0 +1,38 @@
...
+  if (hash.has(value)) {
+    return hash.get(value)
+  }
+
+  const result = Array.isArray(value) ? [] : {}
+  hash.set(value, result)
+
+  Object.keys(value).forEach((key) => {
+    // @ts-expect-error: Index signature for generic object
+    result[key] = deepClone((value as any)[key], hash)
```
amritk:
```suggestion
    result[key] = deepClone((value)[key], hash)
```

you can remove the as any if you have the ts expect error

PR #6597, opened 2025-08-19
feat(workspace-store): move measurements to workspace store

amritk:
**Problem**

Currently, our measurements are done on the references.

**Solution**

With this PR we move them to the workspace store so we can track what is taking a long time. This doesn't move all of them but some. Once we are in a good place we can remove the breakdown and just keep the addDocument time.

**Checklist**


DemonHa (approved):
LGTM 🔥

marclave (approved):

PR #6600, opened 2025-08-19
feat(api-reference): map reference configuration to the new store configuration

DemonHa:
**Problem**

The current implementation does not propagate the existing references configuration when creating a new store configuration. As a result, important reference settings are lost, preventing consistent behavior and blocking upcoming migration work.

**Solution**

This PR ensures that the references configuration is properly propagated to the new store configuration. This allows us to maintain consistency across store updates and enables the references to be used effectively during migration tasks.

**Checklist**

### DIFF
--- a/packages/api-reference/src/v2/helpers/map-configuration.ts
+++ b/packages/api-reference/src/v2/helpers/map-configuration.ts
@@ -1,0 +1,56 @@
+import type { ApiReferenceConfigurationWithSources } from '@scalar/types'
+import type { DocumentConfiguration } from '@scalar/workspace-store/client'
+
+/**
+ * Maps a partial ApiReferenceConfigurationWithSources object to a DocumentConfiguration
+ * for use in the API Reference. This function transforms configuration options
+ * into the expected structure for the 'x-scalar-reference-config' property.
+ *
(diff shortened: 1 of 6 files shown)


packages/api-reference/src/v2/helpers/map-configuration.ts
```diff
@@ -0,0 +1,56 @@
...
+
+/**
+ * Maps a partial ApiReferenceConfigurationWithSources object to a DocumentConfiguration
+ * for use in the API Reference. This function transforms configuration options
+ * into the expected structure for the 'x-scalar-reference-config' property.
+ *
+ * @param config - Partial configuration object for the API Reference
+ * @returns DocumentConfiguration object with mapped settings
+ */
+export const mapConfiguration = (config: Partial<ApiReferenceConfigurationWithSources>) => {
```
amritk: you probably want `ApiReferenceConfiguration` for a single config
OR
`AnyApiReferenceConfiguration` to take in all the configs

amritk: this should most likely be a single config I think which can be transformed outside of this. This should be 1:1 configs

amritk (approved):
Seems reasonable

PR #6603, opened 2025-08-20
feat(api-reference): use galaxy.scalar.com as the first entry in index.html

amritk:
**Problem**

Currently, we were pointing the first entry index.html in references to the registry. However galaxy.scalar.com uses another document and we normally compare the two.

**Solution**

With this PR we add the same document that galaxy.scalar.com uses as the first entry and move the registry document to the second entry.

No need for changeset as this only affects dev

**Checklist**

### DIFF
--- a/packages/api-reference/index.html
+++ b/packages/api-reference/index.html
@@ -21,9 +21,9 @@
       Scalar.createApiReference('#app', {
         sources: [
           {
-            title: 'Scalar Galaxy (CDN)', // optional, would fallback to 'API #1'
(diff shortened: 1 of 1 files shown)


packages/api-reference/index.html
```diff
@@ -23,6 +23,11 @@
           {
             title: 'Scalar Galaxy', // optional, would fallback to 'API #1'
             slug: 'scalar-galaxy', // optional, would be auto-generated from the title or the index
+            url: 'https://galaxy.scalar.com/openapi.yaml',
```
geoffgscott, line 26:
```suggestion
            url: 'https://galaxy.scalar.com/openapi.json',
```

PR #6605, opened 2025-08-20
chore(CODEOWNERS): add xC0dex as codeowner

xC0dex:
**Problem**

Currently, I don’t have permission to merge PRs for:

- Docker integration
- Documentation
- GitHub workflows

I would appreciate it if we could change this 🙂
**Checklist**

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -2,12 +2,11 @@
 * @hanspagel @amritk @geoffgscott @marclave
 
 # Root level code owners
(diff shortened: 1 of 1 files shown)


marclave (changes requested):
will have to chat internally! we usually leave infra work to a select few people (im not even on the list :) 

as for documentation we are probably going to keep it with Hans & myself for merge capabilities.

<3

PR #6606, opened 2025-08-20
chore(api-reference): add CDN dev page for api references

hwkr:
We can maybe make the implementation tidier here but this let's us test the CDN side by side with main using the same sources.

**Checklist**

### DIFF
--- a/packages/api-reference/cdn.html
+++ b/packages/api-reference/cdn.html
@@ -13,7 +13,7 @@
 
     <!-- Our Script -->
     <script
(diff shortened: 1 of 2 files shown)


amritk (approved):
Just add a dev:cdn command and were good to go

hwkr:
> Just add a dev:cdn command and were good to go

This is run when you run `pnpm dev:reference` you just have to go to `http://localhost:5173/cdn`

PR #6611, opened 2025-08-21
fix(oas-utils): skip readOnly properties in request examples

federicobond:
**Problem**

Request examples should not contain readOnly properties.

**Solution**

Pass mode: write to getExampleFromSchema when generating request examples.

**Checklist**


marclave:
nice : )

https://swagger.io/docs/specification/v3_0/data-models/data-types/

PR #6612, opened 2025-08-21
fix(api-reference): remove extra coerces and cleanUp plugin from bundler

DemonHa:
**Problem**

Currently, there are some problems with the extra coerces and with the clean up plugin

**Solution**

With this PR we remove the extra coerces and disable cleanup plugin to fix main branch

**Checklist**


amritk (approved):

DemonHa:
Tuns out the extra coerce with the proxy in place is acting really weird and it is slowing down a lot

We can get rid of coerces cuz it will happen on document ingress so we always have the correct types

PR #6615, opened 2025-08-21
feat(api-reference): client selector block 

DemonHa:
We are migrating the current components to consume and emit events from the new store

**Checklist**


DemonHa:
Depends on #6569

PR #6625, opened 2025-08-22
feat(Scalar.AspNetCore): switch to json converters

xC0dex:
With this PR, we are using JsonConverters for the enums instead of mapping enum values to strings in the mapper.

This is a cleaner approach, IMO, as we reduce some magic mapping.
In the future, we can just remove all converters and use the built-in `[JsonStringEnumMemberName]` attribute. However, this API is only available for .NET 9+.

**Checklist**


hanspagel (approved):
looks better to me, but I have no clue

xC0dex:
Forgot the changeset, re-review required

PR #6642, opened 2025-08-25
ci: use API Reference artifacts

xC0dex:
With this PR we reuse the already built `standalone.js` (artifact of the API Reference package) instead of building it multiple times. This is faster and more efficient.

Thoughts?

**Checklist**


hanspagel (approved):
I think that’s smart, and actually pretty fast (2s) vs the build (which is 30s I think)!

PR #6643, opened 2025-08-25
feat(snippetz): better indentation for arrays and objects

federicobond:
**Problem**

Currently, arrays and empty objects are formatted with weird indentation and non standard spacing.

**Solution**

With this PR the indentation is much more natural. Arrays are formatted horizontally if they contain only primitive values or empty objects and switch to vertical formatting if any of the values contain a newline.

I also added some tests to validate the functionality.

**Checklist**

### DIFF
--- a/.changeset/big-socks-check.md
+++ b/.changeset/big-socks-check.md
@@ -1,0 +1,6 @@
+---
+'@scalar/api-reference': patch
+'@scalar/snippetz': patch
+---
+
+feat(snippetz): improve indentation handling in objectToString


packages/snippetz/src/plugins/node/ofetch/ofetch.test.ts
```diff
@@ -95,13 +95,16 @@ ofetch('https://example.com', {
...
+  query: [
+    {
+      name: 'foo',
+      value: 'bar'
+    },
+    {
+      name: 'bar',
+      value: 'foo'
+    }
+  ]
```
federicobond, line 107: @hanspagel btw, this should be:
```
{
  'foo': 'bar',
  'bar': 'foo'
}
```

PR #6648, opened 2025-08-25
feat(workspace-store): don't throw on validation errors, add validation errors to the console

### LINKED ISSUE #6644: API Reference: operations no longer render
### What happens?

Since the last release, the operations of the OpenAPI document no longer render in the API Reference. The browser console shows the following error:

> Uncaught (in promise) Invalid document provided! Please check your input document. It has some invalid refs.

It should render like before.

Unfortunately, we cannot share the OpenAPI document in this issue directly.

### OpenAPI Document

_No response_

hanspagel:
**Problem**

Currently, when a referenced part of a document is invalid, validation fails completely and says something about invalid refs. This makes the whole rendering fail.

**Solution**

With this PR, we show the exact validation error messages in the console and still add the document so it can be rendered.

Fixes #6644

The actual issue seems to be in the coerce, we’ll try to fix that in another PR.

**Example Document**

```
openapi: 3.1.0
info:
  title: Test API
  version: 1.0.0
paths:
  /test:
    get:
      responses:
        '200':
          content:
            application/json:
              schema:
                # Only happens when we $ref to a schema with invalid properties
                $ref: '#/components/schemas/Foobar'
components:
  schemas:
    Foobar:
      type: object
      # Invalid, makes the validation fail
      required:
      properties:
        storage:
          type: string

```

**Console**

Screenshot 2025-08-25 at 18 34 13

**Checklist**

### DIFF
--- a/packages/workspace-store/src/client.ts
+++ b/packages/workspace-store/src/client.ts
@@ -59,6 +59,13 @@ type ExtraDocumentConfigurations = Record<
   }
 >
 
+export type ValidationError = {
+  message: string
+  path: string
+  schema: unknown
+  value: unknown
+}
+
 const defaultConfig: RequiredDeep<Config> = {
   'x-scalar-reference-config': defaultReferenceConfig,
 }
@@ -529,7 +536,10 @@ export const createWorkspaceStore = (workspaceProps?: WorkspaceProps): Workspace
   }
 
   // Add a document to the store synchronously from an in-memory OpenAPI document
-  async function addInMemoryDocument(input: ObjectDoc & { initialize?: boolean; documentSource?: string }) {
+  async function addInMemoryDocument(input: ObjectDoc & { initialize?: boolean; documentSource?: string }): Promise<{
+    errors: ValidationError[] | null
+    document: unknown
+  }> {
     const { name, meta } = input
     const cloned = measureSync('deepClone', () => deepClone(input.document))
     const inputDocument = measureSync('upgrade', () => upgrade(cloned).specification)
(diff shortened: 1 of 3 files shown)


packages/workspace-store/src/client.ts
```diff
@@ -593,7 +593,20 @@ export const createWorkspaceStore = (workspaceProps?: WorkspaceProps): Workspace
...
+        console.error(`Error #${index + 1}:`, {
+          message: error.message,
+          path: error.path,
+          schema: error.schema,
+          value: error.value,
+        })
+      })
+      console.groupEnd()
+
+      throw 'Invalid document provided! Please check your input document.'
```
amritk: Remove the throw as well and just type cast it! we don't want it to fail

amritk: and maybe return the error array so the consumer knows its an error and can do what they want

packages/workspace-store/src/client.ts
```diff
@@ -585,17 +595,30 @@ export const createWorkspaceStore = (workspaceProps?: WorkspaceProps): Workspace
...
+        path: error.path,
+        schema: error.schema,
+        value: error.value,
+      }))
     }
 
+    // Create a cleaned document that follows the strict schema
+    const strictDocument = isValid
+      ? temporaryDocument
+      : Value.Cast(OpenAPIDocumentSchemaStrict, Value.Default(OpenAPIDocumentSchemaStrict, temporaryDocument))
```
hanspagel: this is the magic to make an invalid document a valid document

DemonHa: doing it this way will loose the proxy

DemonHa: also we are doing this before so the document is supposed to be a valid document

packages/workspace-store/src/client.ts
```diff
@@ -585,17 +595,28 @@ export const createWorkspaceStore = (workspaceProps?: WorkspaceProps): Workspace
...
+      errors = validationErrors.map((error) => ({
+        message: error.message,
+        path: error.path,
+        schema: error.schema,
+        value: error.value,
+      }))
     }
 
+    // Type-cast and just try to render, even if the document is (partially) invalid
+    const strictDocument = isValid ? temporaryDocument : (temporaryDocument as WorkspaceDocument)
```
hanspagel: we typecast here

packages/workspace-store/src/client.ts
hanspagel: we just typecast for now

PR #6649, opened 2025-08-25
feat(workspace-store): make `required: null` an empty array

hanspagel:
**Problem**

Documents that have `required: null` don’t pass the validation, which makes the whole rendering fail.

**Solution**

This PR adds a bit of code to the `cleanUp` plugin, to replace `required: null` with `required: []`.

**Checklist**

### DIFF
--- a/packages/workspace-store/src/plugins.ts
+++ b/packages/workspace-store/src/plugins.ts
@@ -226,6 +226,11 @@ export const cleanUp = (): LifecyclePlugin => {
       if ('pattern' in node && !('type' in node)) {
         node['type'] = 'string'
       }
+
+      // Convert required: null to required: [] for object schemas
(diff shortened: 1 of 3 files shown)


packages/workspace-store/src/plugins.ts
```diff
@@ -226,6 +226,11 @@ export const cleanUp = (): LifecyclePlugin => {
       if ('pattern' in node && !('type' in node)) {
         node['type'] = 'string'
       }
+
+      // Convert required: null to required: [] for object schemas
+      if ('required' in node && node.required === null && 'properties' in node) {
+        node.required = []
```
amritk, line 232: we should also add null to the type array

amritk (changes requested):
just one more quick change

amritk (approved):
I was thinking of something else 😄

PR #6656, opened 2025-08-25
feat(documentation): observe dom and re initialize draggable script

hwkr:
Adds a MutationObserver so we know when the page changes and then re-initializes the script elements after a debounce period.

**Checklist**

### DIFF
--- a/documentation/assets/landing.js
+++ b/documentation/assets/landing.js
@@ -1,6 +1,19 @@
+/* Debounce helper */
+function debounce(func, timeout = 300) {
+  let timer
+  return (...args) => {
+    clearTimeout(timer)
+    timer = setTimeout(() => {
+      func.apply(this, args)
+    }, timeout)
+  }
+}
+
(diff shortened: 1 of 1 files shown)


cameronrohani (approved):

hanspagel:
What’s the bug that we’re fixing here? 👀 I can’t really test because I don’t know what to look for.
But if someone else knows, it’s fine, too.

amritk:
> What’s the bug that we’re fixing here? 👀 I can’t really test because I don’t know what to look for. But if someone else knows, it’s fine, too.

looks like its adding draggability to items which get added to the dom after the fact?

hwkr:
> looks like its adding draggability to items which get added to the dom after the fact?

Specifically when you change pages in the nav bar the elements get removed and then re-added (also vite-ssg does something weird on load where they get removed and re added)

amritk (approved):
send it

marclave (approved):

PR #6661, opened 2025-08-26
feat(api-client): use base64 library

xC0dex:
**Problem**

Currently, …

**Solution**

With this PR …

**Checklist**


packages/api-client/src/libs/send-request/build-request-security.test.ts
```diff
@@ -81,6 +82,13 @@ describe('buildRequestSecurity', () => {
       expect(result.headers['Authorization']).toBe('Basic username:password')
     })
 
+    it('should handle basic auth with Unicode characters', () => {
+      basic.username = 'żółć'
+      basic.password = 'тест'
+      const result = buildRequestSecurity([basic])
+      expect(result.headers['Authorization']).toBe(`Basic ${encode('żółć:тест')}`)
+    })
```
hanspagel: nice

but we should use the hard coded string instead of  `${encode('żółć:тест')}`

otherwise we test the library against itself

and if you do this, leave a comment with the decoded values, above the encoded string (that no one is able to read)

packages/api-client/src/views/Request/libs/oauth2.ts
```diff
@@ -42,10 +40,7 @@ export const generateCodeChallenge = async (verifier: string, encoding: 'SHA-256
   const digest = await crypto.subtle.digest('SHA-256', data)
 
   // Base64URL encode the bytes
-  return btoa(String.fromCharCode(...new Uint8Array(digest)))
-    .replace(/\+/g, '-')
-    .replace(/\//g, '_')
-    .replace(/=+$/, '')
```
hanspagel, line 48: wait, I don’t even know what the replace did!? you sure we don't need that anymore?

xC0dex: This code made the string URL compatible. The new method does this for us if we pass `true` as the second parameter.

xC0dex:
> Nice didn't even know btoa was falling out of favor. do you think you could do atob -> decode as well?

Sure! I also wanted to ask if I should replace all occurrences of `btoa`. It's also used in the API Reference, but my current changes only affect the API client.

packages/api-client/src/libs/send-request/build-request-security.test.ts
xC0dex: Done

PR #6663, opened 2025-08-26
docs: rename access to access control

hanspagel:
**Problem**

I think @geoffgscott asked to rename Access to Access Control, and I think that’s better.

**Solution**

Just did it.

Don’t know if the redirect works, locally it doesn’t. Need to check the preview.

Screenshot 2025-08-26 at 15 01 47

**Checklist**


geoffgscott (approved):
:+1:

marclave:
we are following the nomenclature for dashboard.

also lets pause on updating guides sidebar until we have our new nav! : ) cam has lots of changes

image

marclave (changes requested):
see above comment

PR #6670, opened 2025-08-27
feat(helpers): better xml generation

hanspagel:
**Problem**

I found a bunch of problems with the XML generation we do:

* the `operationTorHar` wasn’t even able to return XML, it just returned JSON
* we’re using an old library (object2XML) and a helper (json2xml) for the same thing
* the generated XML is just one line and not formatted at all
* we don’t support xml tags on the root level (e.g. if the schema is `type: object`), so basically never
* we manually add the xml schema declaration manually every time

**Solution**

This PR fixes those problems and the XML becomes beautiful.

**Checklist**

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1968,6 +1968,9 @@ importers:
 
   packages/mock-server:
     dependencies:
+      '@scalar/helpers':
+        specifier: workspace:*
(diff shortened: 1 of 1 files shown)


marclave (approved):
hell yeah

amritk (approved):
Seems legit

PR #6675, opened 2025-08-27
fix(api-reference): allow server variables dropdown to close on selection

hwkr:
So because the ScalarListbox was wrapped in a `<label>` weird click events were getting passed into the listbox and the listbox wasn't closing. 

This updates it so the `<label>` is adjacent to the form inputs and connected with an `id`.

**Checklist**


geoffgscott:
Can we add a regression test?

PR #6677, opened 2025-08-27
fix(integration-fastapi): export all from scalar_fastapi

roryschadler:
**Problem**

Currently, the following snippet fails type-checking with `pyright`:

```py
from scalar_fastapi import get_scalar_api_reference
```

Output from `pyright`:

```text
error: "get_scalar_api_reference" is not exported from module "scalar_fastapi"
    Import from "scalar_fastapi.scalar_fastapi" instead (reportPrivateImportUsage)
```

This arose after the `py.typed` file was added - without that, it seems that `pyright` doesn't report the issue.

**Solution**

This PR adds a `__all__` export for each symbol imported into `__init__.py`, which satisfies the type checker.

**Checklist**

### DIFF
--- a/.changeset/good-ads-rush.md
+++ b/.changeset/good-ads-rush.md
@@ -1,0 +1,5 @@
+---
+'scalar-fastapi': patch
+---
+
+Export objects from scalar_fastapi package root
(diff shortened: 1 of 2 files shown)


marclave (approved):
awesome :)

PR #6681, opened 2025-08-27
chore(platform): Use custom playwright container to fix cache issues

geoffgscott:
Attempts to speed up the CDN snapshot run time


packages/components/test/compose.yml
```diff
@@ -2,13 +2,14 @@ name: scalar-components-playwright
...
     ports:
       - 5001:5001
     init: true
-    working_dir: /home/pwuser
-    user: pwuser
     stop_signal: SIGINT
     stop_grace_period: 10s
-    image: mcr.microsoft.com/playwright:v1.54.0-noble
-    command: /bin/sh -c "npx -y playwright@1.54.0 run-server --port 5001 --host
+    image: scalarapi/playwright:1.55.0
```
hwkr: Where did we define this? And do we know that it's the same as what's running in CI?

hwkr (approved):
Everything was able to run locally on my machine and generate snapshots (after pulling our 1gb docker image 😅)

It seems like theres some slight rendering difference between the `mcr.microsoft.com/playwright:v1.54.0-noble` image and our new `scalarapi/playwright:1.55.0` image so I updated the snapshots, they all look visually the same but now they render out to the same file contents too.

marclave (approved):

PR #6684, opened 2025-08-27
feat(workspace-store, json-magic): correctly validate the schemas

DemonHa:
**Problem**

Currently, schemas with optional fields were not being properly validated. In particular, when a $ref was used, the referenced value was not validated against its schema, leading to incorrect or incomplete validation results.

**Solution**

This PR enforces the presence of a type property for all schemas, ensuring proper validation of both direct and referenced values. Additionally, the union scoring algorithm now correctly identifies the matching schema, allowing values to be cast accurately. As a result, we can remove the previous “fixer” plugin from the bundler, simplifying the build process.

**Checklist**


amritk (changes requested):
On hold pending the typebox fix as this will slow us down even more

DemonHa:
The whole store loads in ~20ms

image

hanspagel:
let’s get the conflicts fixed and merge this, eh?

geoffgscott (approved):
LGTM

PR #6685, opened 2025-08-27
fix(snippetz): use correct query format in ofetch snippet

federicobond:
**Problem**

Currently, ofetch snippet uses an incorrect format for query parameters.

**Solution**

With this PR ofetch now uses the correct format.

**Checklist**


marclave (approved):

hanspagel (approved):
absolutely!

PR #6692, opened 2025-08-28
fix: deployments aren't skipped for forks

hanspagel:
**Problem**

Forks and bots don’t have access to secrets, so we need to skip some steps. Looks like my if condition had an invalid syntax.

**Solution**

This PR should finally fix it, so CI is passing for forks again. 🤞 

**Checklist**


hanspagel:
I desperately want forks to pass CI again, so I skip the queue here.

hanspagel:
rebased a PR from a fork and it worked!

PR #6695, opened 2025-08-28
fix(api-reference): grab first tag from sidebar if not defined

amritk:
**Problem**

Currently, we try to grab the first tag from the document to open it. But if there's no tags defined on the document this doesn't work.

**Solution**

With this PR we use the side bar entries to grab a tag.

**Checklist**

### DIFF
--- a/.changeset/poor-steaks-watch.md
+++ b/.changeset/poor-steaks-watch.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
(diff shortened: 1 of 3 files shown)


marclave (approved):

hanspagel:
I think the failing test needs to be updated

hanspagel (approved):
nice

I updated the test

amritk:
wicked ty

PR #6697, opened 2025-08-28
feat(components)!: add "Add new option" slot to combobox

hwkr:
Previously we were using the `after` slot in a bunch of places to have an "Add new option" (in fact this is all the `before` and `after` slots were being used for). The problem was buttons added into those slots weren't handled as part of the options list and weren't accessible by keyboard as part of the options list (tl;dr; the UX wasn't very good).

This PR:

* Deprecates (removes) the `before` and `after` slots
* Adds and `add` slot which when filled adds an "Add new option" entry at the bottom of the list
* The "Add new option" entry is
  * Always shown regardless of search query
  * Accessible by keyboard like a normal entry
  * Automatically focused if there aren't any other matching filter results
* Adds and updates the tests 

https://github.com/user-attachments/assets/83692d67-233c-4c9c-9660-b5fdf0ceacf3

**Checklist**

### DIFF
--- a/packages/components/src/components/ScalarCombobox/ScalarCombobox.stories.ts
+++ b/packages/components/src/components/ScalarCombobox/ScalarCombobox.stories.ts
@@ -194,26 +194,31 @@ export const WithAddNew: Story = {
   args: { options: [] },
   render: (args) => ({
     components: {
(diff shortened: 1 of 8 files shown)


marclave (approved):

marclave:
@cameronrohani approves

PR #6717, opened 2025-09-02
feat(nestjs): pure ESM distribution 

inyourtime:
Now all LTS versions of Node.js already support `require(ESM)` (see [Node.js docs](https://nodejs.org/api/modules.html#loading-ecmascript-modules-using-require)).

Therefore, we can safely use `@scalar/core` for this package.

This change should likely be released as a major version bump.

**Checklist**

### DIFF
--- a/integrations/nestjs/src/nestJSApiReference.ts
+++ b/integrations/nestjs/src/nestJSApiReference.ts
@@ -1,16 +1,8 @@
 import type { ServerResponse } from 'node:http'
-import type { ApiReferenceConfiguration } from '@scalar/types/api-reference'
 import type { Request, Response } from 'express'
 import type { FastifyRequest } from 'fastify'
-
-export type NestJSReferenceConfiguration = Partial<ApiReferenceConfiguration> & {
-  withFastify?: boolean
-  cdn?: string
-}
-
-export type ApiReferenceOptions = Partial<ApiReferenceConfiguration> & {
-  cdn?: string
-}
+import type { ApiReferenceOptions, NestJSReferenceConfiguration } from './types'
+import { getHtmlDocument } from '@scalar/core/libs/html-rendering'
 
 /**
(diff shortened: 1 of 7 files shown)


integrations/nestjs/test/index.test.ts
```diff
@@ -109,6 +109,19 @@ describe('apiReference middleware (express)', () => {
 
     expect(res.text).toContain('Function API')
   })
+
+  it('should preserves function properties in configuration', async () => {
```
hanspagel:
```suggestion
  it('preserves function properties in configuration', async () => {
```

integrations/nestjs/src/nestJSApiReference.ts
```diff
@@ -84,62 +76,30 @@ export const customThemeCSS = `
...
-      </body>
-    </html>
-  `
+export function apiReference(givenConfiguration: NestJSReferenceConfiguration) {
+  const configuration = {
+    ...DEFAULT_CONFIGURATION,
+    ...givenConfiguration,
+  }
+
+  const content = () => getHtmlDocument(configuration, customThemeCSS)
```
hanspagel, line 92:
```suggestion
  const content = () => getHtmlDocument(configuration, customTheme)
```

hanspagel:
@inyourtime the PR is still a draft. you want to add something? or is it ready? I’m doing the final test when you’re ready

integrations/nestjs/src/nestJSApiReference.ts
inyourtime: because this
https://github.com/scalar/scalar/blob/a3cc390b49fd35722e257b5886af941236066f31/integrations/nestjs/src/nestJSApiReference.ts#L18

hanspagel (approved):
tested and works great! AWESOME work

PR #6721, opened 2025-09-02
fix(workspace-store): performance issues on script load

DemonHa:
**Problem**

Initializing our TypeBox types was causing significant delays during the first render, resulting in a prolonged white screen. The performance issue was traced to the combination of Type.Recursive and our custom PartialDeep type, which introduced heavy initialization overhead.

**Solution**

This PR optimizes type initialization, reducing the startup time from over 2000ms to around 200ms—an order-of-magnitude improvement. This dramatically shortens the initial render delay and eliminates the visible white screen lag.

**Checklist**

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -28,8 +28,8 @@ catalogs:
       specifier: 1.55.0
       version: 1.55.0
     '@scalar/typebox':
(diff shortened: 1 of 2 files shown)


hanspagel (approved):
awesome improvement!

PR #6724, opened 2025-09-02
feat(fastapi): cover more of the configuration

hanspagel:
**Problem**

We just support a few of the configuration options in the FastAPI integration.

**Solution**

This PR adds most configuration options.

We don’t cover those that expect a JS function yet, though.

**Checklist**


hanspagel:
has tests, just adds to the existing integration, let’s get this in

PR #6728, opened 2025-09-02
fix(api-reference): lazy loading again

amritk:
**Problem**

Currently, we have some race conditions which cause an operation be at the top of the window on load sometimes.

**Solution**

With this PR we use some tomfoolery to ensure that the operation is always at the top. Now this PR doesn't solve the quick flash of another tag that you see sometimes, that we can fix later. This is just to get it out the door.

**Checklist**


packages/api-reference/src/components/Content/Models/Models.vue
```diff
@@ -22,9 +22,9 @@ const schemas = computed(() => getModels(document))
 </script>
 <template>
   <Lazy
-    id="models"
     v-if="schemas && Object.keys(schemas).length > 0"
-    :isLazy="Boolean(hash) && !hash.startsWith('model')">
+    id="models"
```
amritk, line 26: keep models lazy

packages/api-reference/src/components/Content/Operations/TraversedEntryContainer.vue
```diff
@@ -69,22 +71,29 @@ const activeServer = computed(() => {
...
-/** Resume scrolling */
-const resume = () => {
+/** Resume scrolling on a debounce in case elements are still loading in */
+const debouncedResume = useDebounceFn(() => {
   unfreeze?.()
   hasLazyLoaded.value = true
   isIntersectionEnabled.value = true
-}
+  emit('allEntriesLoaded', true)
+}, 500)
```
amritk: this debounce is the sauce that fixes this PR

amritk: might need to play with this number

hanspagel:
Wait, I have an idea about the root cause

hanspagel:
NOT FIXING ANYYTHINGGG 😫😫😫

(losing my mind)

PR #6729, opened 2025-09-02
feat(agent): Add `agent` slot and button to `ApiReferenceLayout`

liamdscalar:
**Problem**

Currently, there is no slot for agent in `ApiReferenceLayout`

**Solution**

With this PR a new agent button will render if some content has been provided to the agent slot.

**Checklist**

### DIFF
--- a/packages/api-reference/src/components/ApiReferenceLayout.vue
+++ b/packages/api-reference/src/components/ApiReferenceLayout.vue
@@ -39,6 +39,7 @@ import { Content } from '@/components/Content'
 import GettingStarted from '@/components/GettingStarted.vue'
 import { hasLazyLoaded } from '@/components/Lazy/lazyBus'
 import MobileHeader from '@/components/MobileHeader.vue'
+import AgentButton from '@/features/agent/AgentButton.vue'
 import { ApiClientModal } from '@/features/api-client-modal'
 import { useDocumentSource } from '@/features/document-source'
 import { SearchButton } from '@/features/Search'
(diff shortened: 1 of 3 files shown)


packages/api-reference/src/components/ApiReferenceLayout.vue
```diff
@@ -312,6 +314,10 @@ useLegacyStoreEvents(store, workspaceStore, activeEntitiesStore, documentEl)
     :style="{
       '--scalar-y-offset': `var(--scalar-custom-header-height, ${yPosition}px)`,
     }">
+    <!-- Agent -->
+    <slot
+      breadcrumb="agent"
```
amritk, line 319: we will want to pass our stores and the document and everything down as props

amritk: unless you have them already

liamdscalar: The agent currently is only using the schema, is there a reason we want to pass everything to agent?

amritk: yea it needs the client store as well for auth right now

liamdscalar: PR here: https://github.com/scalar/scalar/pull/6730 which should fix the symbol not being resolved when calling `useWorkspace`

hanspagel (changes requested):
maybe there is way to not call this "agent" specifically, but think a little bit more generic

let me take a look

hanspagel (changes requested):
Personally, I’d prefer to have those slots generic, is that just me?

* `<slot name="pre-header">`

But we should definitely put the button where we use it (org) and add another slot instead

* move `<AgentButton />` to org
* add `<slot name="sidebar-footer">` (or `agent-scalar-button` if we don’t want to make it generic)

amritk:
Yep new strategy!

PR #6730, opened 2025-09-03
chore(agent): Change `WORKSPACE_SYMBOL` to string for portability

liamdscalar:
**Problem**

Currently,`WORKSPACE_SYMBOL` is a Symbol, this is causing a few issues when different instances of the symbol are instantiated.

**Solution**

With this PR `WORKSPACE_SYMBOL` is now a string.

**Checklist**

### DIFF
--- a/packages/api-client/src/store/store.ts
+++ b/packages/api-client/src/store/store.ts
@@ -14,7 +14,7 @@ import { useModal } from '@scalar/components'
 import type { RequestEvent, SecurityScheme } from '@scalar/oas-utils/entities/spec'
 import type { Path, PathValue } from '@scalar/object-utils/nested'
 import type { ApiReferenceConfiguration } from '@scalar/types/api-reference'
-import { type InjectionKey, inject, reactive, ref, toRaw } from 'vue'
+import { inject, reactive, ref, toRaw } from 'vue'
 
 export type UpdateScheme = <P extends Path<SecurityScheme>>(
   path: P,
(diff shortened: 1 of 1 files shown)


liamdscalar:
@hanspagel 
> isn‘t this actually a feature? can we somehow make sure to only have one instance of the symbol?
> 
> I don’t know enough about that to make a decision here.

I called it a chore, since it's more of an implementation detail than anything else.

I agree that in an ideal world we would ensure that we reference the same symbol. Currently this is a problem because of the way the symbols are instantiated across different packages.

For example:
```vue
  <ApiReferenceLayout // provides
    :configuration="{...}"
    :store>
    <template #agent>
      <Agent /> // injects
    </template>
  </ApiReferenceLayout>
```

- `ApiReferenceLayout` here provides the workspace store directly via `provide(WORKSPACE_SYMBOL, workspaceStore)` (installed from `@scalar/api-client/store`).

- `Agent` wants to use the workspace store by calling `useWorkspace` (installed from `@scalar/api-client/store`).

So while both techincally import the store from `@scalar/api-client/store`, `Agent` in this scenario actually instantiates its own instance of `WORKSPACE_SYMBOL` since its bundled separetely from `ApiReferenceLayout`.

So in essense Agents `WORKSPACE_SYMBOL` !== ApiReferenceLayouts `WORKSPACE_SYMBOL`.

I can think of a few ways to fix this, maybe by passing down the symbol to Agent, but converting the symbol to a const string instead seems to resolve these issues because we aren't relying on the same object, but instead the equality of the two strings.

amritk (approved):
so this is why I didn't think that provide/inject was passed into slots. Nice find!

PR #6731, opened 2025-09-03
feat(Scalar.AspNetCore): switch to json converter

xC0dex:
Follow-up of https://github.com/scalar/scalar/pull/6469. I noticed that we forgot to use a JSON converter for the recently introduced `OperationTitleSource` enum.

**Checklist**


hanspagel (approved):
thanks buddy

PR #6732, opened 2025-09-03
refactor(api-reference): lazy loading

hanspagel:
**Problem**

the initial scroll to id doesn’t work, that’s why the `redirect.e2e.ts` fails

**Solution**

* new approach to render all sections with their id right-away (fast), BUT NOT the content (slow)
* so we can scroll to the right position already
* but it’s still fast
* removes lazy loading from the introduction: it was disabled there and we don’t the first section to render fast
* the ApiReferenceLayout isn't responsible for all the scroll logic anymore (open first tag, removing the hash when scrolling to the top…)
* we drop support for browsers that don't have [`requestAnimationFrame`](https://caniuse.com/requestanimationframe) (support is great)

**Benefits**

- easier to use/understand, all the code is in the lazy loading folder
- tags open faster, because they render the operations one by one (vs. all at once)

**Tasks**

- [ ] lazy load in a smart order (current hash first, the 2 after that, then  everything before, then the rest)
- [ ] do we still need the event bus? let’s wire it up again, or remove it
- [ ] check other browsers
- [ ] Check search, make it work with sections that aren’t loaded
- [ ] Remove the lazybus entirely, see how it goes
- [ ] do we need a factory for the state?
- [ ] test SSR/SSG
- [ ] Rename component
- [ ] Run the tests
- [ ] Write tests
- [ ] does not jump to the top when switching documents
- [ ] scrolling is broken it seems ⚠️
- [ ] models heading doesn’t open

**Lazy Loading Highlighted**

Screenshot 2025-09-03 at 13 26 36

Screenshot 2025-09-03 at 13 26 43

**Checklist**

### DIFF
--- a/packages/api-reference/src/components/IntersectionObserver.vue
+++ b/packages/api-reference/src/components/IntersectionObserver.vue
@@ -37,6 +37,7 @@ onMounted(() => {
       intersectionObserverRef,
       ([{ isIntersecting }]) => {
         if (isIntersecting && id) {
+          console.log('INTERSECTING', id)
           emit('intersecting')
         }
       },
--- a/packages/api-reference/src/components/Section/Section.vue
+++ b/packages/api-reference/src/components/Section/Section.vue
@@ -19,11 +19,10 @@ function handleScroll() {
 
   if (id) {
     // Open parent section
-    setCollapsedSidebarItem(getSectionId(id), true)
-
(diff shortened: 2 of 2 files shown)


hanspagel:
This introduces as much regressions as it fixes, and it’s just too big already.

PR #6733, opened 2025-09-03
chore(workspace-store): delete loose schemas

DemonHa:
**Problem**

Currently, we are using loose schemas to coerce input of rebase

**Solution**

With this PR we remove the loose schemas to improve initial load time and keep the original input untouched

**Checklist**


DemonHa:
> I always wondered what these were for 🙂

The initial idea was to had like a looser version for the input documents, so we still had a structure even for the original documents. But it turns out we can totally get rid of it and keep the original inputs untouched, which would help in the overall performance as well

PR #6734, opened 2025-09-03
feat(oas-utils): upgrade performance of example generator

amritk:
**Problem**

Currently, while investigating lazy loading I came across this performance issue, `getExampleFromSchema` was bottlenecking

**Solution**

With this PR I pulled this fix out from the other PR (which is headed for the trash), it greatly improves the request example schema generation time. With some benchmarking we can improve it further, but this was good enough for some really quick gains.

| Header | Header |
|--------|--------|
| Before | image |
| After | image | 

**Checklist**

### DIFF
--- a/packages/oas-utils/src/spec-getters/get-example-from-schema.ts
+++ b/packages/oas-utils/src/spec-getters/get-example-from-schema.ts
@@ -112,7 +112,7 @@ export const getExampleFromSchema = (
     return undefined
   }
 
-  // Check if the result is already cached
+  // Check if we have a cached result for this schema
   if (resultCache.has(schema)) {
     return resultCache.get(schema)
   }
@@ -375,7 +375,7 @@ export const getExampleFromSchema = (
     }
 
     // If all items are null, return null
-    return null
+    return cache(schema, null)
   }
 
   // Check if schema has the `allOf` key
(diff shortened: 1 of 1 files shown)


packages/oas-utils/src/spec-getters/get-example-from-schema.test.ts
```diff
@@ -1509,17 +1509,7 @@ describe('getExampleFromSchema', () => {
...
-            foobar: {
-              foobar: {
-                foobar: {
-                  foobar: '[Circular Reference]',
-                },
-              },
-            },
-          },
-        },
+        foobar: '[Circular Reference]',
```
hanspagel: We need the 10 levels here, otherwise people will complain for sure

hanspagel (changes requested):
awesome gains!

this makes it output less levels for circular refs (see comment), I think a few people might dislike this

need to double check

packages/oas-utils/src/spec-getters/get-example-from-schema.test.ts
amritk: yea I went back and fourth on that one

amritk: maybe we just remove it and use the max depth instead

packages/oas-utils/src/spec-getters/get-example-from-schema.ts
```diff
@@ -112,20 +117,16 @@ export const getExampleFromSchema = (
     return resultCache.get(schema)
   }
 
-  // Check whether it's a circular reference
-  if (level === MAX_LEVELS_DEEP + 1) {
-    try {
-      // Fails if it contains a circular reference
-      JSON.stringify(schema)
```
amritk, line 119: this was the biggest perf gain actually

PR #6737, opened 2025-09-04
feat(workspace-store): change the way we declare schemas

DemonHa:
With this PR we use internal refs on the schemas which makes it more performant

**Checklist**


amritk (commented):
Oh no he didint

hanspagel (approved):
crazy gains

PR #6746, opened 2025-09-04
ci: replace pinned action job with repository setting

hanspagel:
**Problem**

We use an action to make sure we only use actions pinned to a commit (for security reasons).

**Solution**

[GitHub introduced that as a checkbox](https://github.blog/changelog/2025-08-15-github-actions-policy-now-supports-blocking-and-sha-pinning-actions/), so we can just use that now.

Screenshot 2025-09-04 at 12 26 36

**Checklist**

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,16 +14,6 @@ env:
   NODE_OPTIONS: '--max_old_space_size=8192'
   TURBO_CACHE_DIR: '${{ github.ref_name }}/.turbo'
 jobs:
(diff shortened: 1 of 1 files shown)


hanspagel:
doesn’t seem to work, it’s failing and saying we use an action that’s not pinned to a commit, but that must be inside blacksmith or inside an action that we use, not directly in this repository here

xC0dex:
https://github.com/scalar/scalar/pull/6743 Fixed the issue

PR #6749, opened 2025-09-04
feat(api-reference): render heavy components when they become visible

hanspagel:
**Problem**

Generating examples from schemas is heavy, especially when we do it for all operations and all their responses and their request examples in a tag at once.

**Solution**

This PR adds a `<WhenVisible>` component, that renders the slot only when the component is in the viewport (or it’s close = 500px before).

As an example, for `@scalar/galaxy` this makes opening the `Planets` tag in the sidebar much, much faster.

Note: The downside is, that user might see an empty spot, e.g. when they scroll faster than the examples are generated. On the other hand, they wouldn’t be able to scroll otherwise. 🤷 

**Tasks**

- [ ] there’s an extraneous prop somewhere
- [ ] ah, we lose the scroll position when content above the viewport loads

**Checklist**

### DIFF
--- a/packages/api-reference/src/components/Lazy/WhenVisible.vue
+++ b/packages/api-reference/src/components/Lazy/WhenVisible.vue
@@ -22,7 +22,7 @@ onMounted(() => {
       // Trigger when placeholder is 10% visible
       threshold: 0.1,
       // Start observing before the element enters viewport
-      rootMargin: '300px',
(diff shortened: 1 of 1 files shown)


hanspagel:
Nope, that doesn’t help, because we lose the scroll position when something loads above the viewport.
I tried freezing, and that helps if you don’t move at the same time, but that might just happen.

PR #6619, opened 2025-08-22
feat(api-reference): info block

DemonHa:
Migrate Info section to consume from the new store and move away from the old store

**Checklist**

I've gone through the following:

- [x] I've added an explanation _why_ this change is needed.
- [x] I've added a changeset (`pnpm changeset`).
- [x] I've added tests for the regression or new feature.
- [ ] I've updated the documentation.

PR #6750, opened 2025-09-04
fix(api-reference): doesn't scroll to id on load

hanspagel:
**Problem**

Currently, when hitting an ID on load, we scroll to a random position it seems. 

It looks like there's still content rendering **after** we unfreeze the DOM, and the browser lands somewhere.

This was probably introduced in #6619.

https://github.com/user-attachments/assets/0f7121bc-32ba-463c-8430-4e6a47f9e850

**Solution**

The tiny but important difference in this PR is, that we wait for a fixed amount of time (300ms) until nothing lazy loads anymore. If that’s the case, we can safely _unfreeze_.

I moved the freezing code to a hook, and it’s executed in the `Content.vue` component now, so on the level we actually think when scrolling/freezing?

> Is this good? No, it’s not. But it’s not horrible either. After spending close to 3 days on this issue, I’m convinced we need to DROP the whole scrolling code. It’s the greatest form of spaghetti code.

Does it fix the issue? Manual testing says yes, and the e2e tests finally passes again, too:

Screenshot 2025-09-04 at 15 47 20

🙏 

**Checklist**

### DIFF
--- a/packages/api-reference/src/components/Lazy/Lazy.vue
+++ b/packages/api-reference/src/components/Lazy/Lazy.vue
@@ -60,7 +60,6 @@ if (isLazy) {
     readyToRender.value = true
 
     if (id) {
-      // console.log('✅ LOADED', id)
       nextTick(() => lazyBus.emit({ loaded: id, save }))
     }
   })
--- a/packages/api-reference/src/hooks/useFreezing.ts
+++ b/packages/api-reference/src/hooks/useFreezing.ts
@@ -52,7 +52,6 @@ export const useFreezing = () => {
         if (lazyIds.value.size === 0) {
           // If we've been at size 0 for CHECK_TIMEOUT, resume
           if (now - lastCheckTime >= CHECK_TIMEOUT) {
-            // console.log('▶️ RESUME', lazyIds.value.size)
             resume()
             return
           }
--- a/packages/helpers/src/dom/freeze-at-top.ts
+++ b/packages/helpers/src/dom/freeze-at-top.ts
@@ -38,8 +38,6 @@ export const freezeAtTop = (id: string) => {
 
   // Return function to stop maintaining position
   return () => {
-    console.trace('▶️ UNFREEZING')
-
     if (rafId !== null) {
       cancelAnimationFrame(rafId)
     }


packages/api-reference/src/hooks/useFreezing.ts
```diff
@@ -0,0 +1,72 @@
...
+
+  // Use the lazybus to handle [un]freezing elements
+  lazyBus.on(({ loading, loaded, save }) => {
+    if (hasLazyLoaded.value) {
+      return
+    }
+
+    // Track the previous elements that are loading
+    if (loading && save) {
+      lazyIds.value.add(loading)
```
geoffgscott, line 38: do we need early returns in here?

packages/api-reference/src/components/Lazy/Lazy.vue
```diff
@@ -60,10 +60,12 @@ if (isLazy) {
     readyToRender.value = true
 
     if (id) {
+      // console.log('✅ LOADED', id)
```
liamdscalar: shall we remove these logs?

packages/api-reference/src/hooks/useFreezing.ts
```diff
@@ -0,0 +1,72 @@
...
+    // Track which elements have loaded
+    if (loaded && save) {
+      lazyIds.value.delete(loaded)
+    }
+
+    // We are empty! Unfreeze the page
+    if (lazyIds.value.size === 0) {
+      let lastCheckTime = Date.now()
+
+      const checkAndResume = () => {
```
amritk, line 50: Hmm so check and resume does the same thing as the debounce I put in the other PR no?

amritk (approved):
Not sure what the difference is with this PR and the other one. I'm leaning towards the other one just for the cleanup of the extra component and renaming. But if this one works then get it in

packages/api-reference/src/hooks/useFreezing.ts
hanspagel: In theory yes! I think it has to live on a higher level, too. But I’m not sure, I lost my mind on this.

amritk: we'll redo it later, just get it in

packages/api-reference/src/hooks/useFreezing.ts
hanspagel: adding a return here would break it, but anyway, I’d prefer to not touch that code, I just moved it there 1:1

in the other PR I tried to update everything I touched, it felt like playing whac-a-mole

PR #6754, opened 2025-09-04
fix(openapi-parser): migrate global and path item parameters

matthyk:
**Problem**

Currently, global as well as path item parameters are not migrated.

**Solution**

With this PR, global and path item parameters are migrated. The following migration strategy is implemented for global and path item parameters:

**Global Parameters**

* `body` and `formData` parameters: Migrated and placed under `components.requestBodies`. For the `body` parameters, the media types from the global `consumes` array are used and defaulted to `application/json`
* `header`, `query` and `path` parameters: Migrated and placed under `components.parameters`

**Path Item Parameters**

This is a bit trickier for `body' and `formData`. If these parameter types are declared at the path item level, they apply to all operations of this path item. So we transform these parameters and add the resulting `RequestBodyObject` to all operations. If an operation defines a request body by itself, this overrides the path item request body. 

The migration of `header`, `query`, and `path` parameters is again straightforward.

**Checklist**

### DIFF
--- a/packages/openapi-parser/src/utils/upgrade-from-two-to-three.test.ts
+++ b/packages/openapi-parser/src/utils/upgrade-from-two-to-three.test.ts
@@ -1,3 +1,4 @@
+import type { OpenAPIV3 } from '@scalar/openapi-types'
 import { describe, expect, it } from 'vitest'
 
 import { upgradeFromTwoToThree } from './upgrade-from-two-to-three'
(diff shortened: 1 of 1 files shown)


hanspagel (approved):
OMG YES! awesome that you found this. definitely open for this change.

hanspagel (commented):
do we need the same for 3.0 -> 3.1? I suppose I forgot it there too

matthyk:
> do we need the same for 3.0 -> 3.1? I suppose I forgot it there too

I don't think so. I did some tests and also the migration from 3.0 to 3.1 mainly involves the migration of JSON Schemas. These migrations do not rely on the structure of the document but more on specific keywords within the schemas.

matthyk:
I have also finalized the migration for body and formData parameters and marked the PR as ready for review 👍 See the PR description for more details

hanspagel (approved):
niceee

PR #6756, opened 2025-09-05
feat(api-reference): openapi auth selector block

DemonHa:
Create server selector block. This block is still using the old store but emits events and does not modify the store directly.

**Checklist**

- [] I've added tests for the regression or new feature.


hanspagel (approved):
rad! tested, LGTM

packages/api-reference/src/v2/blocks/scalar-auth-selector-block/helpers/update-scheme.ts
```diff
@@ -0,0 +1,42 @@
...
+    const scheme = securitySchemes[uid]
+
+    if (auth && scheme?.nameKey) {
+      const nameScheme = (auth[scheme.nameKey] ||= {} as Record<P, NonNullable<PathValue<SecurityScheme, P>>>)
+      nameScheme[path] = value
+      safeLocalStorage().setItem(CLIENT_LS_KEYS.AUTH, JSON.stringify(auth))
+    }
+  } catch (e) {
+    console.error(e)
+  }
```
geoffgscott, line 41: Would prefer this moves to a dedicated workspace localstorage layer but good enough for now.

PR #6769, opened 2025-09-08
chore: update to latest pnpm 10.15.1

xC0dex:
**Problem**

Currently, we don't use the latest and greatest pnpm.

**Solution**

This PR bumps pnpm to the latest version, 10.15.1. With that, we have some new arguments, like:

image

**Checklist**


amritk:
Cool new goodies, we gotta upgrade both sides at the same time though @geoffgscott

hanspagel (commented):
Oh, yes, love it! 

I think we even need to upgrade our internal code base first, so we don’t interrupt work.

hanspagel:
looks like we’re doing this in #6810

PR #6770, opened 2025-09-08
chore(pnpm): use catalog in more places

xC0dex:
**Problem**

Currently, we have some inconsistencies across our dependencies.

**Solution**

With this PR, we:

- Use `vite-node` only from the catalog.
- Use `shx` in our root `package.json`, so it's available across all packages.

**Checklist**


xC0dex:
This is just the beginning 👀

PR #6773, opened 2025-09-08
fix(json-magic, workspace-store): use hidden properties during validation

DemonHa:
**Problem**

When coercing values, the proxy logic was emitting hidden internal properties back into the document. This led to issues with strict validation, since these hidden properties were not showing in the validated document.

**Solution**

This PR updates the proxy logic to handle internal properties more gracefully:
- Internal (private) properties are now visible during coercion and validation, ensuring correctness.
- When assigning the final document, private properties are hidden, so the emitted document only contains the expected public fields.

This approach preserves strict validation while keeping internal proxy behavior intact.

**Checklist**

### DIFF
--- a/.changeset/ninety-terms-bake.md
+++ b/.changeset/ninety-terms-bake.md
@@ -1,0 +1,6 @@
+---
+'@scalar/workspace-store': patch
+'@scalar/json-magic': patch
+---
+
+fix: use hidden properties during validation


packages/json-magic/src/magic-proxy/proxy.ts
```diff
@@ -50,6 +50,7 @@ const REF_KEY = '$ref'
  */
 export const createMagicProxy = <T extends Record<keyof T & symbol, unknown>, S extends UnknownObject>(
   target: T,
+  options?: { showInternal?: boolean },
   root: S | T = target,
   cache = new Map<string, unknown>(),
   proxyCache = new WeakMap<object, T>(),
```
geoffgscott, line 56: Should probably just move to an object prop for everything

DemonHa: we have to discuss what would be a good api for this, making it as an object prop feels kind of off to me:

```
createMagicProxy({ source, target, options, cache, proxyCache }) // ?? would confuse end user
```

I believe end user should just know about source and options, maybe the thrid argument is all other internal stuff

PR #6776, opened 2025-09-09
refactor(api-reference): schema rendering

DemonHa:
- Removed discriminator logic

image

- Get the ref name from the ref string

Before:
image
After: 
image

**Checklist**


amritk:
@hanspagel planet comes from the xml name which has priority, I believe if you remove that it will use the ref name
```json
"xml": {
  "name": "planet"
}
````

DemonHa:
> just one tiny thing that we might want to add to this PR: can we show the same name that’s in the list of models (**P**lanet, not **p**lanet)?

Right now ref have the lowest priority. Do we want to prioritise ref?

amritk:
nah I think ref should be last resort, we should check the titles first

amritk (approved):
LGTM! Nice work

PR #6780, opened 2025-09-09
chore(api-client): emit authorized event from RequestAuth component

amritk:
**Problem**

Currently, we don't emit any events, only update the store.

**Solution**

With this PR we emit a vue event when we authorize with oauth2

**Checklist**

### DIFF
--- a/packages/api-client/src/views/Request/RequestSection/RequestAuth/OAuth2.vue
+++ b/packages/api-client/src/views/Request/RequestSection/RequestAuth/OAuth2.vue
@@ -40,6 +40,10 @@ const {
   workspace: Workspace
 }>()
 
+const emits = defineEmits<{
+  authorized: []
+}>()
+
 const loadingState = useLoadingState()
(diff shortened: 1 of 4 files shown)


amritk:
Bringing it back afterall events > watchers

amritk:
were back

packages/api-client/src/views/Request/RequestSection/RequestSection.vue
```diff
@@ -213,6 +213,7 @@ watch(
         :collection="collection"
         :envVariables="envVariables"
         :environment="environment"
+        :isReadOnly="layout === 'modal'"
```
DemonHa, line 216: should this be `clientLayout === 'modal' || layout === 'reference'`?

amritk: reference doesn't use this anymore right? Uses the new one you built

PR #6783, opened 2025-09-09
feat(java): let users override the `ScalarController`

hanspagel:
**Problem**

We make it hard for Spring Boot users to override our `ScalarController`.

**Solution**

This PR adds `@ConditionalOnMissingBean`, so the controller won’t register if there’s a controller with the same name already. User can then extend the controller easily and override parts of it.

```java
@RestController
public class CustomScalarController extends ScalarController {
    // Custom implementation
}
```

**Checklist**

### DIFF
--- a/.changeset/fuzzy-pumas-doubt.md
+++ b/.changeset/fuzzy-pumas-doubt.md
@@ -1,0 +1,5 @@
+---
+'@scalar/webjar': patch
+---
+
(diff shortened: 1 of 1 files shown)


xC0dex (approved):

marclave:
thank goodness I interned at IBM and know java

PR #6804, opened 2025-09-10
docs: updated Scalar Docs section

hanspagel:
* tweaking of the copy
* API Docs >  Scalar Docs (sounds more like a product offering)
* **The** Platform
* Make the point that the result is an always-up-to-date documentation
* reorder list
* more engaging call to action

**Before**

Screenshot 2025-09-10 at 12 16 00

**After**

Screenshot 2025-09-10 at 10 53 38

wdyt?


marclave (approved):
dig it, we gotta keep the call to actions the same everywhere tho

PR #6808, opened 2025-09-10
docs: move quotes down

hanspagel:
typically, you show your products first and once the users understood what you do and just need something to build more trust, you show testimonials eventually

this PR moves those quotes down.

the thing that I don’t like about it: with the letter from marc we have a lot of text there.

maybe we cut those quotes to half the length to address that problem.

wdyt?


marclave (changes requested):
meant to be logo wall + quotes/logos are higher, we did this intentionally

PR #6809, opened 2025-09-10
fix(api-reference): `a > img` uses full-width in Markdown

### LINKED ISSUE #3961: Postman link extends full width does not accurately depict clickable area
### What happens?

The clickable area to open postman is bigger than the actual button due to the length of the <a> tag.
image

The clickable are should only be as large as the button image.

### OpenAPI Document

https://bolt-docs-git-abdul-api-ref-by-scalar-technical-writing.vercel.app/api-bolt/#description/about

hanspagel:
**Problem**

When using images wrapped in a link, the `<a>` HTML element takes the full-width. The clickable area is bigger than the actual child element.

Screenshot 2025-09-10 at 14 04 02

**Solution**

This PR sets `` elements to `display: inline-block` (instead of `display: block`), which should behave the same.

Except that text can be on the same line. But if you put the text on a separate line in the Markdown, it’ll be on separate line in the rendered content, too.

Fixes #3961

Definitely requires @cameronrohani and @hwkr and the snapshots tests to pass.

**Checklist**

### DIFF
--- a/.changeset/shiny-yaks-sneeze.md
+++ b/.changeset/shiny-yaks-sneeze.md
@@ -1,0 +1,5 @@
+---
+'@scalar/components': patch
+---
+
+fix: a > img takes the full width


hwkr:
Looks like GitHub actually uses `inline` for `img` tags so this is probably fine?

PR #6810, opened 2025-09-10
feat: bump vite

geoffgscott:
Bumps vite and moves a few package to the catalog.


geoffgscott:
> Should announce the pnpm upgrade so we can all update locally. Did you create a new lockfile?

didn't fully recreate yet. Can do in a follow up.

examples/react-webpack/package.json
```diff
@@ -14,7 +14,7 @@
   "engines": {
     "node": ">=20"
   },
-  "packageManager": "pnpm@10.5.2",
+  "packageManager": "pnpm@10.15.1",
```
xC0dex: I think this is redundant with the root `package.json`

hanspagel:
I don’t think we should recreate the pnpm-lock file. Doing this would update most packages to their latest version (all that aren’t pinned = most) and might introduce subtle bugs that we might not even catch before release. Which happened in this PR here.

pnpm takes care of updating the structure of the lock file if necessary.

hanspagel:
I’ve reset the @apidevtools/swagger-parser version from 10.1.1 to 10.1.0 and the tests pass again. When we want to update, we can do it in a separate PR, I’d say.

Issue #6815, opened 2025-09-11
Setting dark_mode to false doesn't work


ashish-b-choudhary-db:
### What happens?

Scalar UI loads in dark mode whether the setting is set to true or false.

### What did you expect to happen?

When using dark_mode as false the UI should load in light mode by default. Used in FastAPI.

```python
@app.get("/scalar", include_in_schema=False)
async def scalar_html():
    return get_scalar_api_reference(
        openapi_url=app.openapi_url,
        title=app.title,
        layout=Layout.CLASSIC,
        dark_mode=False,
    )
```

hanspagel:
True! I fixed the fastapi integration, because the output was wrong. But it seems the actual package isn’t doing what we say in the docs it does. Need to look into that, too.

hanspagel:
Fixed in #6853 and #6823. :)

PR #6823, opened 2025-09-12
fix(fastapi): wrong output for `dark_mode`

hanspagel:
**Problem**

When setting `dark_mode=true` in FastAPI, it doesn’t add anything to the output.

**Solution**

This PR fixes this.

See #6815.

I’m not sure we do the right thing when this config is set (need to check), but the output in this integration is fixed.

Follow-up: #6853

**Checklist**


geoffgscott (approved):
LG. Untested

PR #6824, opened 2025-09-12
fix(api-reference): use operation/path servers for code examples

tltsutltsu:
**Problem**

When using multiple servers (root servers and operation-level servers), the “Examples” code samples ignore the operation/path server selection and always use the first server from the root of the OpenAPI document. The “Test Request” button already respects operation-level (and path-level) servers, so Examples and Test Request can disagree.

Steps to reproduce:
1. Define root servers: `[https://root.example.com, …]`
2. For an operation, define servers: `[{ url: 'https://op.example.com' }]`
3. “Test Request” sends to https://op.example.com, but “Examples” shows code using https://root.example.com

**Solution**

With this PR, Examples now use the same precedence as Test Request:
1. Prefer `operation.servers[0]`
2. Else `pathItem.servers[0]`
3. Else fall back to the provided/active collection server

**Checklist**

### DIFF
--- a/packages/api-reference/src/features/Operation/Operation.vue
+++ b/packages/api-reference/src/features/Operation/Operation.vue
@@ -16,6 +16,7 @@ import { combineParams } from '@/features/Operation/helpers/combine-params'
 import { convertSecurityScheme } from '@/helpers/convert-security-scheme'
 import type { ClientOptionGroup } from '@/v2/blocks/scalar-request-example-block/types'
 
+import { getFirstServer } from './helpers/get-first-server'
 import ClassicLayout from './layouts/ClassicLayout.vue'
 import ModernLayout from './layouts/ModernLayout.vue'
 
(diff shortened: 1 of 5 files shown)


packages/api-reference/src/features/Operation/Operation.vue
```diff
@@ -70,6 +73,29 @@ const selectedSecuritySchemes = computed(() =>
...
+  const pathServer =
+    pathItem.value?.servers && pathItem.value.servers[0]
+      ? (getResolvedRef(pathItem.value.servers[0]) as ServerObject)
+      : undefined
+
+  if (pathServer?.url) return pathServer
+
+  // Fallback to the server coming from the active collection selection
+  return server as unknown as ServerObject | undefined
+})
```
hanspagel: I think we’d want to put such logic in a separate file, with a dedicated test file, so we can test more (edge) cases easily.

hanspagel:
Great PR, thanks! can one of you guys have a look @DemonHa @amritk 🙏

packages/api-reference/src/features/Operation/Operation.vue
amritk: yep just this, move to a helper with this logic, add some tests and this should be good! Nice work

hanspagel:
Taking it over 🏄

packages/api-reference/src/features/Operation/Operation.vue
hanspagel: done

PR #6830, opened 2025-09-12
fix(json-magic): do not throw when we set on an invalid ref

DemonHa:
**Problem**

When setting a value on a $ref that does not exist, the operation fails because the referenced path cannot be resolved.

**Solution**

This PR improves the behavior by automatically creating the missing path when a $ref is invalid, ensuring the value can still be set successfully.

**Checklist**


DemonHa:
@geoffgscott Do we want to emit some kind of warning in this case?

It means that your document is invalid

PR #6842, opened 2025-09-14
fix(api-reference): webhook request example display

federicobond:
**Problem**

Currently, the display of webhook request examples has regressed and shows the same as normal operations.

**Solution**

With this PR webhook request examples are displayed correctly.

**Checklist**


hanspagel (approved):
oh, how did that disappear! thanks!

PR #6843, opened 2025-09-14
fix(build-tooling): check dirname exists before accessing

hwkr:
I'm not sure why I'm the only one having this issue but sometimes `import.meta.dirname` was undefined for me...

**Checklist**


hwkr:
Closing since we require node 22 anyway

PR #6850, opened 2025-09-15
chore: upgrade Hono to 4.9.7

hanspagel:
Moves Hono to the catalog, upgrades to 4.9.7 (security). :)


hanspagel:
security, let’s goo

PR #6854, opened 2025-09-15
feat(api-reference): documentDownloadType: 'direct'

### LINKED ISSUE #6847: Download spec from orignal url
### Discussed in https://github.com/scalar/scalar/discussions/6752

<sup>Originally posted by **YaredFall** September  4, 2025</sup>
Previously, you could download the spec using a direct download link if the `url` was defined in the config.
Unfortunately, this feature was removed with migration to v2 `DownloadLink` component.
It was the only way to copy the spec link and a real time-saver for opening the spec in a new tab.
Would like this feature back 😺 
</div>

hanspagel:
**Problem**

> Previously, you could download the spec using a direct download link if the url was defined in the config.
> Unfortunately, this feature was removed with migration to v2 DownloadLink component.
> It was the only way to copy the spec link and a real time-saver for opening the spec in a new tab.
> Would like this feature back 😺

**Solution**

This PR introduces a new configuration value:

```js
{
  documentDownloadType: 'direct',
}
```

which renders a regular `a[href]` with just the URL (if we have an URL).

Fixes #6847

**Checklist**

### DIFF
--- a/packages/api-reference/src/v2/blocks/scalar-info-block/components/DownloadLink.vue
+++ b/packages/api-reference/src/v2/blocks/scalar-info-block/components/DownloadLink.vue
@@ -50,7 +50,6 @@ const handleDownloadClick = (format: 'json' | 'yaml') => {
       "
       class="download-button"
       type="button"
-      variant="ghost"
       @click.prevent="handleDownloadClick('yaml')">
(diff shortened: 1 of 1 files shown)


marclave (approved):

xC0dex (approved):
I like it

Issue #6822, opened 2025-09-12
Custom plugin does not work with NestJS Scalar setup


phamhieu:
### What happens?

I tried to create a custom plugin for NestJS Scalar by following the official guide: https://guides.scalar.com/scalar/scalar-api-references/plugins

However, when I add the plugin to my apiReference configuration, I encounter the following Zod validation error:
```
ZodError: [
  {
    "code": "invalid_type",
    "expected": "function",
    "received": "null",
    "path": [
      "plugins",
      0
    ],
    "message": "Expected function, received null"
  }
]
    at get error (standalone.js:121:10)
    at $e.parse (standalone.js:140:64)
    at Bo.fn (standalone.js:31340:20)
    at va (standalone.js:1356:68)
    at get value (standalone.js:1533:9)
    at Po (standalone.js:1507:74)
    at $o (standalone.js:1508:34)
    at standalone.js:31185:17
    at qo (standalone.js:1564:48)
    at Ho (standalone.js:1565:40)
```

### What did you expect to happen?

I expected the custom plugin to run without errors, just like in the example from the guide.

### OpenAPI Document

https://github.com/phamhieu/nestjs-scalar-plugin-bug

hanspagel:
Wait, I thought this would work with the latest release! Need to check.

inyourtime:
I think it’s related to this #5634

hanspagel:
I don’t think this is all we need to actually fix this, but definitely required, too: #6856

We probably need the vanilla JS extensions, too.

PR #6856, opened 2025-09-15
fix(core): HTML rendering does not output functions in arrays

hanspagel:
**Problem**

When functions are put in an array (e.g. `plugins`), they are not stringified and added to the HTML output.

**Solution**

With this PR those functions in arrays are added to the HTML output, too.

See #6822

**Checklist**


hanspagel:
tests passing, it can only become better, let’s get this in

PR #6860, opened 2025-09-16
feat(json-magic): support for `$id` and `$anchor`

DemonHa:
**Problem**

Our implementation currently lacks support for the `$id` and `$anchor` keywords, which are part of the official JSON Schema specification. This limits schema resolution and makes it harder to work with external or anchored definitions.

**Solution**

This PR introduces full support for `$id` and `$anchor`, enabling proper resolution and navigation of schema references.

Key Changes
 - *Bundler*: skips bundling external dependencies when defined with the $id keyword.
 - *Proxy Interface*: updated signature for `createMagicProxy(target, options, args)`.
 - *Proxy Behavior*: now internally handles path routing to correctly resolve `$id` and `$anchor` references.

Before:
image

After:
image

**Checklist**


packages/json-magic/src/helpers/get-value-by-path.ts
```diff
@@ -0,0 +1,44 @@
...
+ *   foo: {
+ *     bar: {
+ *       baz: 42
+ *     }
+ *   }
+ * };
+ * // Returns: { context: '', value: 42 }
+ * getValueByPath(obj, ['foo', 'bar', 'baz']);
+ */
+export function getValueByPath(target: Record<string, any>, segments: string[]): { context: string; value: any } {
```
geoffgscott: Can we do value: unknown instead? Any is always dangerous

hanspagel (approved):
woahh, sick work @DemonHa

PR #6863, opened 2025-09-16
fix(openapi-parser): fix body and reference parameter migration

matthyk:
**Problem**

Two problems:
* If an operation specifies the consumed media types __AND__ no request body parameter is defined, the consumes array is not deleted during the migration process.
* If a reference object parameter is migrated, an empty schema object is added

**Solution**

With this PR these two issues are fixed. If the two issues should be addressed in separate PRs, I can do that as well.

**Checklist**


hanspagel (approved):
another great PR, thanks @matthyk!

just migrated the changes to the new package

PR #6864, opened 2025-09-16
feat(api-reference): sidebar block

DemonHa:
Moved sidebar component to its own block and update it to consume from the new store

**Checklist**


packages/api-reference/src/components/Section/SectionAccordion.vue
```diff
@@ -30,7 +30,11 @@ const handleScroll = () => {
 
   // Open models and webhooks on scroll
   if (id?.startsWith('model') || id?.startsWith('webhook')) {
-    setCollapsedSidebarItem(getSectionId(id), true)
+    const sectionId = items.value.entities.get(id)?.parent?.id
+
+    if (sectionId) {
+      setCollapsedSidebarItem(sectionId, false)
+    }
```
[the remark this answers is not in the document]
hanspagel: yeah, I think the value was wrong all the time. it’s great if we can correct it, we just need to check if it’s used right in all places

packages/api-reference/src/components/ApiReferenceLayout.vue
```diff
@@ -51,6 +50,7 @@ import type {
   ReferenceLayoutSlot,
   ReferenceSlotProps,
 } from '@/types'
+import { SidebarBlock, useSidebar } from '@/v2/blocks/scalar-navigation-block'
```
hanspagel: can we import the `SidebarBlock` from `scalar-sidebar-block` (not `scalar-navigation-block`)?

packages/api-reference/src/components/Content/Models/Models.test.ts
```diff
@@ -1,28 +1,19 @@
 import type { ApiReferenceConfiguration } from '@scalar/types'
+import { createWorkspaceStore } from '@scalar/workspace-store/client'
+import { coerceValue } from '@scalar/workspace-store/schemas/typebox-coerce'
+import { OpenAPIDocumentSchema } from '@scalar/workspace-store/schemas/v3.1/strict/openapi-document'
 import { mount } from '@vue/test-utils'
-import { describe, expect, it, vi } from 'vitest'
+import { beforeEach, describe, expect, it, vi } from 'vitest'
+
+import { useSidebar } from '@/v2/blocks/scalar-navigation-block'
+import { createSidebar } from '@/v2/blocks/scalar-navigation-block/helpers/create-sidebar'
```
hanspagel: I don’t know what we did with the other blocks, but I think we should only import from the root, so `'@/v2/blocks/scalar-navigation-block'` in that case. That way we can move code around inside the block, while keeping the exports as they are. Feels more encapsulated. Wdyt?

packages/api-reference/src/components/Content/Models/Models.vue
```diff
@@ -1,42 +1,45 @@
...
+  const modelEntries = modelEntry.value?.children ?? []
+
+  return modelEntries
+    .filter((it) => it.type === 'model')
+    .map((it) => ({
+      id: it.id,
+      name: it.name,
+      schema: getResolvedRef(schemas[it.name]),
+    }))
 })
```
hanspagel: can we still call them `schemas`? because in the OpenAPI specification, there are only schemas, not models

packages/api-reference/src/components/Content/Operations/TraversedEntry.vue
```diff
@@ -111,15 +108,15 @@ defineExpose({
           :config="config"
           :document
           :isWebhook="isWebhook(entry)"
-          :method="entry.method"
+          :method="entry.method as any"
```
hanspagel: any chance we can avoid type-casting here? 😳

packages/api-reference/src/components/Content/Tags/ClassicLayout.vue
```diff
@@ -26,7 +26,7 @@ defineProps<{
       </SectionHeader>
       <ScalarMarkdown
         class="tag-description"
-        :value="tag.tag?.description"
+        :value="tag?.description"
```
hanspagel, line 29: nice :D

hanspagel:
a few things that we need to test before merging, just from the top of my head:

- [x] configuration: [generateModelSlug](https://github.com/scalar/scalar/blob/main/documentation/configuration.md#generatemodelslug-model--name-string---string)
- [x] configuration: `showSidebar: false`
- [x] search
- [x] `x-scalar-ignore: true`
- [x] `x-internal: true`
- [x] ~~do webhooks still appear in the sidebar?~~ 
> They do appear, if there are untagged webhooks we put them in a tag called `webhooks`
- [ ] documents without any tag (we create a default tag then and tweak the rendering)
- [x] ~~collapse of sections (because we corrected the boolean?)~~
> That was some old code that needed to be deleted, I pushed the update
- [ ] a large document (stripe, zoom)

packages/api-reference/src/components/Content/Models/Models.test.ts
DemonHa: This is for testing only, we don't actually need to expose the createSidebar function

packages/api-reference/src/components/Section/SectionAccordion.vue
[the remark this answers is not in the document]
DemonHa: turns out that part, we don't even need it anymore

packages/api-reference/src/components/OperationsList/OperationsList.vue
```diff
@@ -20,16 +20,16 @@ const { tag } = defineProps<{
...
 
 const operationsAndWebhooks = computed(
-  (): (TraversedOperation | TraversedWebhook)[] => {
+  (): (TraversedOperation | TraversedWebhook)[] | undefined => {
     return tag.children?.filter(
-      (child) => 'operation' in child || 'webhook' in child,
+      (child) => child.type === 'operation' || child.type === 'webhook',
     )
   },
 )
```
amritk:
```suggestion
const operationsAndWebhooks = computed(
  (): (TraversedOperation | TraversedWebhook)[] => 
   tag.children?.filter(
      (child) => child.type === 'operation' || child.type === 'webhook',
   ) ?? []
)
```
when dealing with arrays like this you can just nullish coalescing to be a little cleaner

packages/api-reference/src/components/Content/Operations/TraversedEntry.vue
```diff
@@ -39,19 +39,16 @@ const {
  * Type guards for different entry types
  */
 const isTagGroup = (entry: TraversedEntry): entry is TraversedTag =>
-  'isGroup' in entry && entry.isGroup
+  entry['type'] === 'tag' && entry.isGroup === true
 
 const isTag = (entry: TraversedEntry): entry is TraversedTag =>
-  'tag' in entry && !isTagGroup(entry)
+  entry['type'] === 'tag' && !isTagGroup(entry)
```
amritk, line 49:
```suggestion
const isTagGroup = (
  entry: TraversedEntry,
): entry is TraversedTag & { isGroup: true } =>
  entry['type'] === 'tag' && entry.isGroup === true

const isTag = (
  entry: TraversedEntry,
): entry is TraversedTag & { isGroup: false } =>
  entry['type'] === 'tag' && !isTagGroup(entry)
```

they are both being typed as tags here, so we need to add the boolean for proper type inference below. Now we can remove the `as TraversedTag`

packages/api-reference/src/components/Content/Operations/TraversedEntry.vue
```diff
@@ -147,7 +144,7 @@ defineExpose({
         :clientOptions
         :config
         :document
-        :entries="entry.children || []"
+        :entries="(entry as TraversedTag).children || []"
```
amritk:
```suggestion
        :entries="entry.children || []"
```

packages/api-reference/src/components/OperationsList/OperationsListItem.vue
```diff
@@ -35,18 +33,9 @@ const pathOrTitle = computed(() => {
...
-  'operation' in operation &&
-  isOperationDeprecated(
-    operation.operation as OpenAPIV3_1.OperationObject<{
-      'x-scalar-stability': XScalarStability
-    }>,
-  )
-
-const isWebhook = (operation: TraversedEntry): operation is TraversedWebhook =>
-  'webhook' in operation
+const isWebhook = (
```
amritk, line 36: I see this in more than one place, should move to a helper in the traversed section

packages/api-reference/src/features/Search/components/SearchButton.vue
```diff
@@ -24,7 +26,12 @@ const handleHotKey = (e: KeyboardEvent) => {
   ) {
     e.preventDefault()
     e.stopPropagation()
-    modalState.open ? modalState.hide() : modalState.show()
+
```
amritk, line 29: oh you fancy huh

packages/api-reference/src/features/Search/helpers/create-search-index.ts
```diff
@@ -30,18 +32,20 @@ export function createSearchIndex(entries: TraversedEntry[]): FuseData[] {
 /**
  * Adds a single entry to the search index, handling all entry types recursively.
  */
-function addEntryToIndex(entry: TraversedEntry, index: FuseData[]): void {
+function addEntryToIndex(entry: TraversedEntry, index: FuseData[], document?: OpenApiDocument): void {
   // Operation
-  if ('operation' in entry) {
-    const requestBodyOrParameterMap = extractRequestBody(entry.operation) || createParameterMap(entry.operation)
+  if (entry.type === 'operation') {
+    const operation = getResolvedRef(document?.paths?.[entry.path]?.[entry.method as 'get']) ?? {}
```
amritk:
```suggestion
    const operation = getResolvedRef(document?.paths?.[entry.path]?.[entry.method]) ?? {}
```
don't think you need this

packages/api-reference/src/features/Search/helpers/create-search-index.ts
```diff
@@ -52,24 +56,27 @@ function addEntryToIndex(entry: TraversedEntry, index: FuseData[]): void {
   }
 
   // Webhook
-  if ('webhook' in entry) {
+  if (entry.type === 'webhook') {
+    const webhook = getResolvedRef(document?.webhooks?.[entry.name]?.[entry.method as 'get']) ?? {}
```
amritk:
```suggestion
    const webhook = getResolvedRef(document?.webhooks?.[entry.name]?.[entry.method]) ?? {}
```
whyyyy

packages/api-reference/src/v2/blocks/scalar-server-selector-block/index.ts
```diff
@@ -1,2 +1,2 @@
-// biome-ignore lint/performance/noBarrelFile: <explanation>
+// biome-ignore lint/performance/noBarrelFile: It's a block
```
amritk, line 1: ✔️

amritk: export from blocks is good

packages/api-reference/src/v2/blocks/scalar-sidebar-block/components/SidebarBlock.vue
```diff
@@ -165,7 +166,7 @@ const hasChildren = (
                       <SidebarElement
                         :id="`sidebar-${child.id}`"
                         :isActive="isItemActive(child.id)"
-                        :item="child" />
+                        :item="child as TraversedEntry" />
```
amritk:
```suggestion
                        :item="child" />
```

packages/api-reference/src/v2/blocks/scalar-sidebar-block/components/SidebarBlock.vue
```diff
@@ -196,7 +197,7 @@ const hasChildren = (
                     <SidebarElement
                       :id="`sidebar-${child.id}`"
                       :isActive="isItemActive(child.id)"
-                      :item="child" />
+                      :item="child as TraversedEntry" />
```
amritk:
```suggestion
                      :item="child" />
```

packages/api-reference/src/v2/blocks/scalar-sidebar-block/helpers/generate-reverse-index.ts
```diff
@@ -0,0 +1,23 @@
+import type { OpenApiDocument, TraversedEntry } from '@scalar/workspace-store/schemas/v3.1/strict/openapi-document'
+
+/**
+ * Generate a reverse index for fast lookups of entities by their IDs.
+ */
+export const generateReverseIndex = (navigation: OpenApiDocument['x-scalar-navigation'] = []) => {
```
amritk, line 6: another word for this is just dictionary 🙂

amritk (changes requested):
Wicked PR! Thanks for doing this one. Gets rid of a lot of the duplication that was killing me. 

Requesting changes just for the type casts, lets try to avoid those! Will test now

packages/api-reference/src/components/OperationsList/OperationsListItem.vue
amritk: or just get rid of it!

amritk:
> documents without any tag (we create a default tag then and tweak the rendering)

~I don't think this works, will check on main~ works the same as main actually! maybe we should get rid of that default tag thing @hanspagel what does it even do?

amritk:
Good to merge as soon as type assertions are out!

packages/api-reference/src/features/Search/helpers/create-search-index.ts
DemonHa: method was typed as string before

packages/api-reference/src/v2/blocks/scalar-sidebar-block/helpers/generate-reverse-index.ts
DemonHa: or reversed index 😂

PR #6867, opened 2025-09-17
feat(code-highlighting): expand available languages

gevann:
**Problem**

Currently, users are only able to highlight a small subset of code languages.

**Solution**

With this PR we expand the set of language available for code highlighting, including more specific exports for the more esoteric entries.

**Checklist**


packages/code-highlight/src/languages/standard.ts
```diff
@@ -46,35 +88,99 @@ const standardLanguages: Record<string, LanguageFn> = {
...
   scss,
   shell,
   sql,
   swift,
+  toml: ini,
   typescript,
   xml,
   yaml,
 }
+const mediumLanguages: Record<string, LanguageFn> = {
```
hanspagel, line 156: can u add comments so future us knows what’s supposed to be in mediumLanguages, specializedLanguages and so on and what they are used for? 🙏

packages/components/src/components/ScalarMarkdown/snapshots/base-codeblocks.png
hanspagel, line 1: the colors in the first example changed it seems, is that what we want?

packages/code-highlight/src/languages/standard.ts
gevann: will do

packages/components/src/components/ScalarMarkdown/snapshots/base-codeblocks.png
gevann: Yes, the lowlight lib in that example is guessing the language of the example. It does that by running the grammars available and picking the one that had the highest relevance score. This PR adds `properties` to the `standardLanguages`, which returns a higher relevance score than `elixir` (which was what it was guessing before), so it's being highlighted as `properties` now (it's actually `applescript`, but that's not in the standardLanguages!).

packages/code-highlight/src/languages/standard.ts
gevann: done!

hanspagel (approved):
AWESOME

PR #6872, opened 2025-09-18
fix(openapi-upgrader): traversing circular references causes infinite recursion

matthyk:
**Problem**

Currently, when calling `traverse` with an object that includes a circular reference, an infinite recursion is caused, resulting in a maximum call size error. I encountered the bug when I wanted to upgrade an OpenAPI document that I had previously dereferenced. However, the problem lies in the traverse function.

**Solution**

With this PR the `traverse` caches all objects. If it traverses an already seen object, it simply returns the cached result.

**Checklist**

### DIFF
--- a/.changeset/calm-turkeys-battle.md
+++ b/.changeset/calm-turkeys-battle.md
@@ -1,0 +1,5 @@
+---
+'@scalar/openapi-upgrader': patch
+---
+
+fix: traversing circular references causes infinite recursion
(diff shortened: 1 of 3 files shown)


hanspagel:
Yeah, @DemonHa will replace this helper with something new in the `@scalar/json-magic` package.
Thanks for bringing this up @matthyk

hanspagel:
Sorry, we’re so busy right now. It’ll be added sooner or later, though. ✌️

PR #6879, opened 2025-09-18
chore: `minimumReleaseAge` (requires pnpm 10.16)

hanspagel:
**Problem**

Sometimes, npm packages have security issues. They are mostly removed/fixed quickly, but we might install them before they are fixed/removed.

**Solution**

This PR uses pnpm’s `minimumReleaseAge` to wait at least a day until it installs a new version of something.

**Checklist**

### DIFF
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -9,6 +9,9 @@ packages:
   - '!**/dist/**'
   - '!**/.output/**'
 
+# To reduce the risk of installing compromised packages, we delay the installation of newly published versions.
+minimumReleaseAge: 1.440 # 1 day
+
 catalogs:
   '*':
     '@electron-toolkit/preload': ^3.0.0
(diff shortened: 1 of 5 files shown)


geoffgscott (approved):
Good idea.

xC0dex:
We could also consider setting [`minimumReleaseAge`](https://docs.renovatebot.com/configuration-options/#minimumreleaseage) in our renovate configuration for node packages.

pnpm-workspace.yaml
```diff
@@ -9,6 +9,9 @@ packages:
   - '!**/dist/**'
   - '!**/.output/**'
 
+# To reduce the risk of installing compromised packages, we delay the installation of newly published versions.
+minimumReleaseAge: 1.440 # 1 day
```
amritk: should we do a bit more than 1 day? I don't think we need bleeding edge and if theres any security fixes we should be able to bypass this. 1 week?

hanspagel: I don’t know what a good number is. I just thought npm seems to react in <24 hours for those security incidents. 

Let’s try 1 week and see how it goes!

amritk: yea I feel like they may not catch it in a day (but pretty close). a week should be safe!

PR #6881, opened 2025-09-18
fix(api-reference): does not download updated content

hanspagel:
**Problem**

Currently, when updating the content, the download button still downloads the initial version (not the updated) version.

**Solution**

This PR contains a reproduction (that we want to remove before merging):

1. `pnpm dev:reference`
2. http://localhost:5173/
3. click download button
4. downloaded document contains "Loading…" not "Hello World"

Looks like this is an issue in the workspace-store, that seems to export the initial version?

```js
const getOriginalDocument = () => store.exportActiveDocument('json') ?? '{}'
```

**Checklist**

### DIFF
--- a/packages/api-reference/index.html
+++ b/packages/api-reference/index.html
@@ -18,14 +18,30 @@
 
     <!-- Initialize the Scalar API Reference -->
     <script type="module">
-      import sources from './test/data/sources.ts'
-
-      Scalar.createApiReference('#app', {
-        sources,
-        persistAuth: true,
-        // Avoid CORS issues
-        proxyUrl: 'https://proxy.scalar.com',
+      // TODO: Restore file before merging
+      const scalar = Scalar.createApiReference('#app', {
+        content: JSON.stringify({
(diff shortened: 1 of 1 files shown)


geoffgscott:
@hanspagel the active document is the live state for client usage. I believe the initial document is what we want.

cc @DemonHa

hanspagel:
@geoffgscott it downloads an empty document with the title "Loading…" for Scalar Docs, so we need to change something there then

amritk:
I can see people wanting both cases, maybe its a config option?

hanspagel:
currently, it’s broken. this would fix it:

* developers want their users to download the document they pass (not upgraded)
* the users want to download what they see (the last version of the document that was passed, not the "loading…" document)

if we want an option, that’s an additional feature for a separate PR, but I don’t think we need the option.
if some developers want their users to download an upgraded file, they’d be able to use the CLI to upgrade the file before serving it.

hanspagel:
closing in favor of #7037

PR #6882, opened 2025-09-18
chore: pin @biomejs/biome and prettier to latest and exact versions

xC0dex:
This PR pins biome and prettier to exact versions, as this is recommended by both tools!

This PR also ignores generated `standalone.js` and `scalar.js` for all integrations.

**Checklist**

### DIFF
--- a/biome.json
+++ b/biome.json
@@ -35,7 +35,8 @@
       "!**/packages/openapi-parser/src/schemas/**/*.ts",
       "!**/.github/renovate.json",
       "!integrations/**/scalar.js",
-      "!integrations/**/standalone.js"
+      "!integrations/**/standalone.js",
+      "!**/rust/**/target"
     ]
   },
   "assist": {


packages/api-client/src/views/Request/ResponseSection/ResponseLoadingOverlay.vue
```diff
@@ -17,9 +17,9 @@ events.requestStatus.on((status) => {
   if (status === 'start') {
     timeout.value = setTimeout(() => loading.startLoading(), 1000)
   } else {
-    clearTimeout(timeout.value),
+    ;(clearTimeout(timeout.value),
       (timeout.value = undefined),
-      loading.stopLoading()
+      loading.stopLoading())
```
geoffgscott: Can we fix this properly please. (only a little bit out of scope :wink:  )

xC0dex: Done 👍

biome.json
```diff
@@ -34,9 +34,8 @@
       "!**/bin/**",
       "!**/packages/openapi-parser/src/schemas/**/*.ts",
       "!**/.github/renovate.json",
-      "!**/webjars/**/*.js",
-      "!**/rust/**/*.js",
-      "!**/rust/**/target"
+      "!integrations/**/scalar.js",
+      "!integrations/**/standalone.js"
```
hanspagel: we still want to ignore the target/ folder, it has compiled output

xC0dex: Done

PR #6886, opened 2025-09-18
feat(docs): subpages redesign for getting started docs/sdks/registry

cameronrohani:
redesign subpages for scalar-docs/getting-started, scalar-sdks/getting-started, scalar-registry/getting-started

image

missing some copy which @marclave will add


marclave:
added copy : ) gunna merge this and get it in ✨

PR #6888, opened 2025-09-18
feat: rename standalone.js to scalar.js for integrations

xC0dex:
This PR renames `standalone.js` to `scalar.js`.
The new name makes it clearer what the file contains and what will be loaded. We do this already for Java and .NET AspNetCore.

**Checklist**


amritk (commented):
Seems ok to me but lets let @hanspagel review this one

PR #6892, opened 2025-09-19
ci: test against Node 22 only (remove Node 20)

hanspagel:
**Problem**

We never had issues with Node 20 (but with Node 18 and the lack of the native `File` type), and we never had failing *tests* in Node 20. If we had Node issues, it was in the build stage.

Though, we’re running the tests against Node 20 and 22.

**Solution**

This PR removes Node 20 from the test matrix. This won’t make CI pass faster, because those tests are running parallel anyway, but it’ll save some trees and maybe a few dollar notes.

**Checklist**

---

> [!NOTE]
> Updates CI matrices to drop Node 20 and run jobs on Node 22 only.
> 
> - **CI**:
>   - Update GitHub Actions matrix in `.github/workflows/ci.yml` to `node-version: [22]` for:
>     - `build`
>     - `test-packages`
>     - `test-integrations`
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 51409a6e15f8b7f098b4ef8ab95b26291280cda5. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,7 +22,7 @@ jobs:
     timeout-minutes: 10
     strategy:
       matrix:
-        node-version: [20, 22]
+        node-version: [22]
 
     steps:
       - name: Checkout repository
@@ -218,7 +218,7 @@ jobs:
     timeout-minutes: 10
     strategy:
       matrix:
-        node-version: [20, 22]
+        node-version: [22]
     steps:
       - name: Checkout repository
         uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5


geoffgscott:
Because I hate trees can we please add Node 24 to offset any gains from dropping node 20 :wink: 

We are using it in org so would be good to have it in the matrix.

amritk:
should we add 24 to the matrix since we are running on 24 now?

amritk:
So it turns out there's a few tests to fix for node 24

One of which requires an update to jsdom (abort-signal) which in turn causes even more issues. We could mock around it but we probably should get on the latest testing dom env sooner or later


| Dom | Result |
|--------|--------|
| jsdom 22 (current) | image |
| jsdom 27 (latest) | image |
| happy-dom | image |


maybe we stick to node 22 for now in OSS? Do this upgrade when we got some down time. I switched back to node 22 locally for OSS until we sort this out!

edit: or we should just bite the bullet and update. I think different tests are currently failing on different computers

hanspagel:
It looks like we’re using jsdom 26 already (or vitest does), so I wonder why we even need the explicit dependency and if we need it we should probably use the same version as vitest does to avoid conflicts.

Just tried to remove jsdom from the package.jsons, let me try that. Anyway, I’ll continue this PR and see if we can make it work or just don’t use 24 for now.

EDIT: AH, the AbortSignal error is a Vitest error that was fixed just recently: https://github.com/vitest-dev/vitest/pull/8390 So we need to update vitest I guess.

hanspagel:
Okay, so we have the following two issues:

1. vitest didn’t release the patch yet, it’ll be in v4
  * so we would need to apply a patch with pnpm
  * or wait for v4
 2. tests are still failing then, because Node 24 has a few updates to how fetch works
  * stricter compliance with the fetch() specification and the removal of third-party implementations of `Blob`, `FormData`, and `AbortController`.
  * several [pre-built interceptors](https://github.com/nodejs/undici/blob/main/docs/docs/api/Dispatcher.md#pre-built-interceptors) for common tasks:
    * `redirect`: automatically follow HTTP 304 and Location headers.
    * `retry`: retry failed requests.
    * `dump`: dump the body of all requests.
    * `dns`: cache DNS lookups for an origin.
    * `cache`: cache HTTP responses according to the [RFC 9111 standard](https://www.rfc-editor.org/rfc/rfc9111.html).

I restored the initial PR to just remove Node 20 and use Node 22 only.

I want Node 24, too. But that’s too much now. We should at least wait until Vitest released the necessary patch.

hanspagel:
Just saves some trees now, we can kill some more trees (and add Node 24) in a separate PR when the vitest patch is released.

PR #6906, opened 2025-09-21
fix(workspace-store): schema does not allow custom formats

hanspagel:
**Problem**

For `format: 'custom-format'` we render `date` as the format. That’s because our schema is too strict. 

> It should be noted that format is not limited to a specific set of valid values or types. 

Source: https://json-schema.org/understanding-json-schema/reference/type

**Solution**

This PR allows any string to be passed.

**Before**

Screenshot 2025-09-21 at 17 35 14

**After**

Screenshot 2025-09-21 at 17 34 23

**Checklist**

### DIFF
--- a/packages/workspace-store/src/schemas/v3.1/strict/schema.ts
+++ b/packages/workspace-store/src/schemas/v3.1/strict/schema.ts
@@ -130,6 +130,8 @@ const StringValidationProperties = Type.Object({
       Type.Literal('sf-token'),
       Type.Literal('sf-binary'),
       Type.Literal('sf-boolean'),
+      // Allow any arbitrary string
+      Type.String(),
     ]),
   ),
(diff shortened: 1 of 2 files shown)


packages/workspace-store/src/schemas/v3.1/strict/schema.ts
```diff
@@ -130,6 +130,8 @@ const StringValidationProperties = Type.Object({
       Type.Literal('sf-token'),
       Type.Literal('sf-binary'),
       Type.Literal('sf-boolean'),
+      // Allow any arbitrary string
+      Type.String(),
```
DemonHa, line 134: we can simplify the union here

amritk:
we might as well remove the whole union actually because it won't do anything

PR #6908, opened 2025-09-22
fix(turbo): remove build warnings

xC0dex:
**Problem**

Currently, we have some turbo build warnings because certain integrations don't have any build output.

**Solution**

This PR updates the `turbo.json` and tells turbo that these packages don't have any output. 

**Checklist**


integrations/rust/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "scalar_api_reference",
+  "name": "@scalar/rust-api-reference",
```
hanspagel: ah, my convention is to use the name we publish the package under, in that case the crate is called scalar_api_reference. that way, we avoid to make up another name (@scalar/rust-api-reference)

turbo.json
```diff
@@ -35,6 +35,18 @@
...
+    },
+    "@scalarapi/docker-api-reference#build": {
+      "outputs": []
+    },
+    "proxy-scalar-com#build": {
+      "outputs": []
+    },
+    "@scalar/rust-api-reference#build": {
+      "outputs": []
     }
```
hanspagel, line 44: the webjar and the rust crate do have outputs, they are just not named dist. I think we should add those explicitly there, so turbo can cache the build for them too

xC0dex: We skip some build jobs (for instance, the Rust integration) because cargo is not available. IDK if caching the build will work

integrations/rust/package.json
xC0dex: This _could_ be confusing in the list of packages when using the changeset, IMO. But I'll revert it 👍

geoffgscott: Yeah I agree that the pnpm package name should remain the same convention as the rest. We can have a dedicated ecosystem package name elsewhere.

turbo.json
xC0dex: IMO there is no benefit in caching these build outputs, neither locally nor in the pipeline. For each project (like the Rust or Java integration), we have dedicated build jobs anyway.

hanspagel: there is: locally, when you have cargo (I do)

integrations/rust/package.json
xC0dex: I'll do this in a separate PR

turbo.json
xC0dex: Done for webjar and cargo

PR #6916, opened 2025-09-22
chore(workspace-store): remove other string formats

amritk:
**Problem**

Recently we added a string to the schema type string format. This actually negates the purpose of having the others in the union.

**Solution**

With this PR we comment out the other formats (in case we ever want to bring them back). This should improve the type inference performance a bit.

**Checklist**

### DIFF
--- a/packages/json-magic/src/bundle/bundle.test.ts
+++ b/packages/json-magic/src/bundle/bundle.test.ts
@@ -1329,35 +1329,32 @@ describe('bundle', () => {
       })
     })
 
-    it(
-      'does not modify external URLs when already defined by $id property',
-      async () => {
-        const url = `http://localhost:${port}`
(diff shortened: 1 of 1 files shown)


packages/workspace-store/src/schemas/v3.1/strict/schema.ts
```diff
@@ -90,49 +90,49 @@ const StringValidationProperties = Type.Object({
...
-      Type.Literal('sf-string'),
-      Type.Literal('sf-token'),
-      Type.Literal('sf-binary'),
-      Type.Literal('sf-boolean'),
-      // Allow any arbitrary string
-      Type.String(),
-    ]),
+    // Allow any arbitrary string, this negates the purpose of the other formats so they are commented
+    Type.String(),
+    // Type.Union([
```
DemonHa: you can delete them, since we want to support any string

PR #6918, opened 2025-09-22
docs(openapi-parser): fix deprecation message

matthyk:
**Problem**

Currently, the deprecation message points to the `bundle` function within the `openapi-parser` package. I guess it actually means the `bundle` function from the `json-magic` package.

**Solution**

With this PR the deprecation message shows the correct import statement for the `bundle` function.

**Checklist**


amritk (approved):
nice

marclave (approved):

PR #6920, opened 2025-09-23
fix(components): use border to table instead of box shadow

hanspagel:
Updates the the markdown tables to use border instead of box shadow to prevent it from being clipped by overflow restrictions.

**Before**

Screenshot 2025-09-23 at 18 21 20

**After**

Screenshot 2025-09-23 at 18 21 27

**Example**

```js
{
  openapi: '3.1.1',
  info: {
    title: 'Example,
    version: '2.1.3',
  },
  paths: {
    '/hello': {
      get: {
        responses: {
          200: {
            description: `**Success**

| Column 1 | Column 2 |
| -------- | -------- |
| Cell 1   | Cell 2   |`,
            content: {
              'application/json': {
                schema: {
                  type: 'string',
                },
              },
            },
          },
        },
      },
    },
  },
}
```

---

> [!NOTE]
> Switches markdown table styling to a solid border (with zero spacing) instead of box-shadow to avoid clipping.
> 
> - **Components**
>   - **`packages/components/src/components/ScalarMarkdown/ScalarMarkdown.vue`**:
>     - Markdown tables: replace `box-shadow` with `border: var(--scalar-border-width) solid var(--scalar-border-color)` and set `border-spacing: 0` (removes shadow-based outline).
> - **Changeset**
>   - Patch releases for `@scalar/api-reference` and `@scalar/components`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit b456ee8e3742f4662c9bb01a0b505d40a85556c5. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


amritk (dismissed):
Snapshot tests to the rescue

Image

hwkr:
Rather than using the CSS here it might be good to just clamp the markdown to 1 line when the thing is collapsed. I believe I've updated ScalarMarkdown to support that.

Alternatively we could use ScalarMarkdownSummary and maybe hide the show hide buttons?

(Would need to add a prop to ScalarMarkdownSummary to hide the default controls)

hanspagel:
@hwkr Yess! Can you add to the PR here?

hwkr:
Okay so I considered doing the summary change but it would pretty substantially change how we render descriptions (which maybe we want to do but don't have to do here) cc @cameronrohani. 

Instead I just updated the table styles to use border instead of box shadow so it's inside the bounding box instead of outside.

hanspagel:
you're so smart. we just need to update the snapshots again.

PR #6934, opened 2025-09-25
chore(json-magic): remove unneeded vue dependency

255kb:
**Problem**

The package @scalar/json-magic [includes `vue` as a dependency](https://github.com/scalar/scalar/blob/main/packages/json-magic/package.json#L63), but it doesn't seem to be used anywhere.

**Solution**

Remove the dependency

**Checklist**


255kb:
sorry, I forgot to push the lock file 🤦

Issue #6693, opened 2025-08-28
don't use the title to replace the type of primitives


hanspagel:
when someone has a schema:

```js
{ type: "string", title: "Name"}
```
we show `Name` as the type

that makes sense for objects, to e.g. show the model name, but that doesn’t make sense for primitives.

the title keyword is for a super short description of the data, not supposed to replace the type: https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-validation-00#section-9.1

soo .. we might want to render the title additionally to the description on a separate line?

but in all cases we should keep the type I'd say

amritk:
I will do this one in the schema PR, since it will conflict

tibbe:
Could we please have a status update on this? :)

amritk:
hey @tibbe the latest is in that PR, you can chime in on what you would like to see!

https://github.com/scalar/scalar/pull/6953

tibbe:
This is getting a bit problematic for us (and our customers) at this point:

The current documentation is hard to use. Consider e.g. these path parameter from [one of our endpoints](https://developer.benetics.io/reference#tag/documents/get/projects/{project_id}/documents/{document_id}):

<img width="397" height="214" alt="Image" src="https://github.com/user-attachments/assets/bb55eff9-83d8-4be6-aa5f-0bd2818b3599" />

Having the `project_id` type documented as "Project Id" isn't helpful to the user. It's a `string`.

Same applies to both request body attributes and query parameters.

The only saving grace is that the example request body we provide has some actual values in it so the user can guess the type (although this doesn't apply to path/query parameters, for which we can't provide examples in an OpenAPI spec).

---

I'm also worried that https://github.com/scalar/scalar/pull/6953 is going in the wrong direction and, if I may be so blunt, there's a fundamental confusion how typing works in JSON: JSON isn't a nominally typed language, meaning that there are no classes. All objects are of type `object`, not e.g. `MyClass`. I think trying to use the OpenAPI `title` property as a class name is a symptom of this confusion. It's not a class name, because JSON doesn't have classes.

hanspagel:
working on a PR: #8107

PR #6953, opened 2025-09-26
fix(api-reference): ranking explicit title declaration as highest friendly name

### LINKED ISSUE #6952: Anyof not showing title when provided
### What happens?

When using this simple schema:

```yml
openapi: 3.1.0
info:
  title: Toggle Test
components:
  schemas:
    NumericCellType:
      anyOf:
        - type: string
          const: number
          title: Number Cell
          description: A number without a qualifier (e.g. dollar, percent, multiple, etc.)
        - type: string
          const: dollar
          title: Dollar Cell
        - type: string
          const: percent
          title: Percent Cell
        - type: string
          const: multiple
          title: Multiple Cell
```

The title's are not being displayed in the dropdown, even though they're provided:

Image

Image

The title's `"Number Cell"`, `"Dollar Cell"`, etc. should be showing up in the "Any of" dropdown.

### OpenAPI Document

_No response_

mclean25:
When a `title` is given, it should be given the highest precedence for use in a friendly name since it's explicity stated and that is the purpose of `title`.

Closes #6952

**Problem**

Currently, even when you use `title` to give your dropdown a friendly name, it wont be used (see #6952).

**Solution**

`title` is given the highest precedence since it is explicitly used as a friendly name by the user.

**Checklist**

- [x] I've updated the documentation. (Looked through and don't see anything pertaining to this explicitly).

---

> [!NOTE]
> Updates `getSchemaType` to prefer `title` over array/union types, consts, and `xml.name`, with tests adjusted and expanded; adds changeset.
> 
> - **API Reference Helpers**:
>   - Update `get-schema-type.ts` to make `title` the top priority when deriving a friendly name, ahead of array/union handling, `const`, and `xml.name`.
>   - Preserve existing logic for arrays, unions, content encoding, `$ref`, and fallbacks.
> - **Tests**:
>   - Revise expectations in `get-schema-type.test.ts` to reflect `title` precedence (e.g., arrays/unions now return the `title` when present).
>   - Add cases for arrays whose `items` have `title`, ensuring `title` is used (e.g., `array Mixed Item[]`).
> - **Changeset**:
>   - Add minor release note: `title` now has highest precedence for friendly names (incl. `anyof`).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 4a86d63a89397bfa53d6cd0842dc1a22621c454b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/.changeset/wise-buses-hug.md
+++ b/.changeset/wise-buses-hug.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': minor
+---
+
+`title` is now given the highest precedence when choosing a friendly name for `anyof`
--- a/packages/api-reference/src/components/Content/Schema/helpers/get-schema-type.test.ts
+++ b/packages/api-reference/src/components/Content/Schema/helpers/get-schema-type.test.ts
@@ -513,7 +513,7 @@ describe('get-schema-type', () => {
         expect(result).toBe('CustomArray')
       })
 
-      it('ignores the title for an array type if items are defined', () => {
+      it('Does not ignore the title for an array type if items are defined', () => {
         const schema = coerceValue(SchemaObjectSchema, {
           title: 'CustomArray',
           type: 'array',
@@ -522,7 +522,7 @@ describe('get-schema-type', () => {
           },
         })
         const result = getSchemaType(schema)
-        expect(result).toBe('array string[]')
+        expect(result).toBe('CustomArray')
       })
     })
 
--- a/packages/api-reference/src/components/Content/Schema/helpers/get-schema-type.ts
+++ b/packages/api-reference/src/components/Content/Schema/helpers/get-schema-type.ts
@@ -39,9 +39,9 @@ const processArrayType = (value: Extract<SchemaObject, { type: 'array' }>, isUni
  * Computes the human-readable type for a schema.
  *
  * Priority order:
- * 1. const values
+ * 1. title property.
  * 2. Array types (with special handling for items)
- * 3. title property
+ * 3. const values
  * 4. xml.name property
  * 5. type with contentEncoding
  * 6. $ref names


amritk (commented):
Hey thanks for the quick PR! It works for me (@hanspagel ?)

also should we bump down `const` or just delete it? I'm not sure I see the value

amritk (approved):

mclean25:
> Thanks for taking the time to submit this PR! We should also keep in mind that fastapi users get a title as a default for all types:
> 
> * #6693
> 
> Sooo it might be a little bit more complex to serve everyone well.
> 
> Just a quick feedback from me, I’ll think about a good solution.

Got it, was thinking there had to be a reason for it to be this way originally.

Just a thought--shouldn't this be as tool-agnostic as possible? I haven't used fastapi in ages, but from a design philosophy isn't it more strange that fastapi provides titles that aren't meant for display purposes? IE: from my naive vantage point it seems more like a fastapi limitation than a scalar one.

hanspagel:
I agree in theory. In practive, I also have to accept that it’s the reality (and won’t change) and the FastAPI user base is huge. Everyone deserves a beautiful API reference.

tibbe:
> Just a thought--shouldn't this be as tool-agnostic as possible? I haven't used fastapi in ages, but from a design philosophy isn't it more strange that fastapi provides titles that aren't meant for display purposes? IE: from my naive vantage point it seems more like a fastapi limitation than a scalar one.

Titles are meant for display purposes but they are not the same as types (even in the OpenAPI spec). There are three pieces of data:
- field name (or query parameter name)
- type
- title

The former two are required to use an API correctly. The last one is a nice to have. It can't be used to replace either of the above two (especially not in a scalar case, such as for query parameters).

hanspagel:
@tibbe yep

hanspagel:
closing in favor of #8107

Issue #6348, opened 2025-07-24
Bug in UI when required GET parameter is not filled


nathanboor:
### What happens?

When you have the following routes:
GET /customer (returns a list of customers)
GET /customer/{key} (returns one customer for the given id)

where the {key} is a required property.

If you test the GET /customer/{key} endpoint, but you do not fill out the required {key} property, the UI will say that the property is required (it flags the variable with red text 'required'), but then returns the list of the endpoint GET /customer

### What did you expect to happen?

1. I would expect that the UI flags the variable as required when the variable is not provided. 
2. I would expect that the UI does NOT return ANY results (blank screen) because the required variable is not set

hanspagel:
I think of the API client as a testing tool, and we need a way to test invalid payloads, too. So I’d like to keep the possibility to send the request even if a required parameter is missing. Does that make sense?

nathanboor:
It could make sense on a Post or Patch, however .... in the case where a required parameter is part of the Path it NEVER wires to the correct handler.  IMHO if GET /customer/{Key} is the pattern it should be blocked by the front-end.
If it would be GET /customer?Key={value} then indeed the validation code server side could detect that the required parameter is NULL and raise an error.

amritk:
Maybe we make path parameters required. This came up recently on discord as well

hanspagel:
@amritk love it, let’s do this!

OdujokoD:
@hanspagel @amritk I can pick this up if you don't mind. I'm thinking of including the check in `createRequestOperation(...)` when the `pathVariables` is being parsed and evaluated:

```    
/** Parsed and evaluated values for path parameters */
const pathVariables = example.parameters.path.reduce<Record<string, string>>((vars, param) => {

  // check param.value here  
  
  if (param.enabled) {
    vars[param.key] = replaceTemplateVariables(param.value, env)
  }

  return vars
}, {})
```

I'd also include a new error key to throw an error for missing parameter, which will be displayed in the toast. Alternatively, maybe I change the border color of the `AddressBar` component if this is in line with the design system.

amritk:
@OdujokoD sure that sounds great, but I think we should show it in the UI as well, we could prevent them from even hitting send and outline the path parameter in red

OdujokoD:
Is this what you mean @amritk? In this case, the `invalidParams` will be used on the components since it's already available. Let me know if this is fine so I can work on the PR.

[...]

PR #6958, opened 2025-09-29
fix(api-client): enforce request parameters

OdujokoD:
**Problem**

Currently, request parameters are not validated, leading to an unexpected behaviour for certain endpoints as described in #6348 

**Solution**

With this PR request parameters are now validated, with a UI update that prevents sending a request with invalid parameters

Screenshot 2025-09-29 at 9 25 31 AM

**Checklist**

---

> [!NOTE]
> Adds reactive validation for required params, prevents sending requests with invalid path params, and updates UI/UX and tests accordingly.
> 
> - **Validation & Execution**:
>   - Add reactive `invalidParams` and `isPathInvalid` in `views/Request/RequestRoot.vue`; block request execution and toast `ERRORS.INVALID_PATH_PARAMETER` when invalid path params.
>   - Wire `isPathInvalid` through `Request.vue` → `RequestSubpageHeader.vue` → `AddressBar.vue`; disable Send button via `isSendButtonDisabled`.
> - **UI Components**:
>   - `AddressBar.vue`: compute `isSendButtonDisabled`; use it to disable Send during requests or invalid paths.
>   - `RequestPathParams.vue`: accept `invalidParams` for table validation display.
>   - `RequestSubpageHeader.vue`: pass `isPathInvalid` down to `AddressBar`.
> - **Errors**:
>   - Add `INVALID_PATH_PARAMETER` to `libs/errors.ts` and integrate with toasts.
> - **Tests**:
>   - Expand `RequestRoot.test.ts` with parameter validation cases (invalid/valid/reactive/missing example/all types).
>   - Update `RequestSubpageHeader.test.ts` for new props.
> - **Misc**:
>   - Refactor watchers to computed; minor robustness tweaks (cloning fallback, imports).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c77ff37d44d786d421b1770875f7245caf7e8269. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/api-client/src/views/Request/RequestRoot.test.ts
+++ b/packages/api-client/src/views/Request/RequestRoot.test.ts
@@ -1,4 +1,4 @@
-import { collectionSchema, requestExampleSchema } from '@scalar/oas-utils/entities/spec'
+import { type RequestExample, collectionSchema, requestExampleSchema } from '@scalar/oas-utils/entities/spec'
 import { mount } from '@vue/test-utils'
 import { type Mock, beforeEach, describe, expect, it, vi } from 'vitest'
 import { ref } from 'vue'
@@ -50,7 +50,7 @@ vi.mock('@/store/active-entities', () => ({
 const mockUseActiveEntities = useActiveEntities as Mock
 const mockActiveEntities = {
   activeRequest: ref(null),
-  activeExample: ref(null),
+  activeExample: ref<RequestExample | null>(null),
   activeCollection: ref(collectionSchema.parse({ watchMode: false })),
   activeEnvironment: ref(null),
   activeWorkspace: ref(null),


hanspagel:
I get where you’re coming from, but my thinking is: An API testing tool should allow to send invalid data and just inform you that’s in invalid. So allowing users to send invalid requests is actually considered a feature. ✌️

hanspagel:
Ah, sorry. It’s just for path parameters. That’s fine.

packages/api-client/src/libs/errors.ts
```diff
@@ -3,6 +3,7 @@ export const ERRORS = {
   DEFAULT: 'An unknown error has occurred.',
   INVALID_URL: 'The URL seems to be invalid. Try adding a valid URL.',
   INVALID_HEADER: 'There is an invalid header present, please double check your params.',
+  INVALID_PARAM: 'There is an invalid request param present, please double check your params.',
```
hanspagel: let’s be a little bit more explicit

```suggestion
  INVALID_PATH_PARAMETER: 'There is an invalid request param present, please double check your params.',
```

packages/api-client/src/components/AddressBar/AddressBar.vue
```diff
@@ -232,7 +233,7 @@ function updateRequestPath(url: string) {
       <ScalarButton
         ref="sendButtonRef"
         class="z-context-plus relative h-auto shrink-0 overflow-hidden py-1 pr-2.5 pl-2 font-bold"
-        :disabled="isRequesting"
+        :disabled="isRequesting || invalidParams.size > 0"
```
hanspagel: if it’s more than just one condition we should make it a computed

`const isSendButtonDisabled = computed(() => isRequesting || invalidParams.size > 0)`

packages/api-client/src/views/Request/RequestRoot.vue
```diff
@@ -63,6 +63,22 @@ const selectedSecuritySchemeUids = computed(
...
+ * made when values are not passed
+ */
+
+const getInvalidParameters = (): Set<string> => {
+ if (!activeExample.value) {
+    return new Set()
+  }
+
+  return validateParameters(activeExample.value) 
+}
```
hanspagel: I didn’t test, but that looks like it’s blocking the request for any invalid data.

We should only block the request for missing/invalid *path* paremters.

OdujokoD: I can do this but that means `required` query parameters are also not enforced for imported OpenAPI spec. Using the [Petstore schema](https://petstore3.swagger.io/api/v3/openapi.json) as an example, the `/pet/findByTags` has a required `tags` query parameter. Since a specification already provides a "contract" for the endpoint, should the tool not enforce it?

hanspagel: I don’t think we should enforce the query parameters!

With the path parameters, it’s just that it would hit a completely different route without those, so there’s no point to test that.

But some might want to test what happens when they omit a query parameter.

packages/components/src/components/ScalarButton/ScalarButton.vue
```diff
@@ -30,6 +30,7 @@ const { cx } = useBindCx()
 <template>
   <button
     :aria-disabled="disabled || undefined"
+    :disabled="disabled"
```
hwkr: We actually don't want to disable buttons since it makes them unfocusable for screen readers. If possible just add a check in the logic to make sure you're not running the logic if it's disabled.

hwkr (changes requested):

packages/api-client/src/components/AddressBar/AddressBar.vue
```diff
@@ -70,6 +71,8 @@ const remaining = ref(0)
 const isRequesting = ref(false)
 /** The loading interval */
 const interval = ref<ReturnType<typeof setInterval>>()
+/** Toggle the send button */
+const isSendButtonDisabled = computed(() => isRequesting.value || isPathInvalid)
```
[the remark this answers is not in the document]
hwkr: Cursor you're wrong destructured props are reactive https://vuejs.org/guide/components/props#reactive-props-destructure

packages/components/src/components/ScalarButton/ScalarButton.vue
OdujokoD: Ohh! I already covered the logic in the `RequestRoot` component. I'll remove this now.

hwkr:
@OdujokoD you can ignore Cursor about the button disabling stuff, what you have there is correct

hwkr (approved):
Looks good to me for the components stuff :)

hanspagel:
Sorry, I’ve had to start fresh. Closing in favor of #7146.

Thanks for helping me to get this together!

PR #6969, opened 2025-09-30
feat(Scalar.AspNetCore): support showToolbar configuration

xC0dex:
**Checklist**


hanspagel:
if #6970 gets in, we might consider to rename the option, too 👀

xC0dex:
> if #6970 gets in, we might consider to rename the option, too 👀

Awesome! I'll wait for this!

connorjs:
Thank you for putting together this pull request. I’m looking to use `showOperationId` in .NET.

connorjs:
Is this still on hold given #6970 merged? @hanspagel @xC0dex

xC0dex:
Hey @connorjs, I'll split the PR so the toolbar property doesn't block the operationId configuration 👍

PR #6979, opened 2025-09-30
feat(api-client): operation block

DemonHa:
Operation block dumb component

---

> [!NOTE]
> Introduce a new Operation Block with params/body/auth UI, validation helpers, and OpenAPI schema extensions, with comprehensive tests and minor config tweaks.
> 
> - **API Client (UI/Blocks)**:
>   - `v2/blocks/scalar-operation-block` new components: `OperationBlock.vue`, `OperationBody.vue`, `OperationParams.vue`, `OperationTable.vue`, `OperationTableRow.vue`, `OperationTableTooltip.vue`; export via `index.ts`.
>   - Features: request name handling, filterable sections (Auth/Variables/Cookies/Headers/Query/Body), auth selector wiring, params table with add/update/delete/clear, request body editor with content-type switch, multipart/form-data and binary upload support, schema-driven tooltips and validation.
>   - Helpers: `group-by`, `get-parameter-*` (content/schema/example), `get-request-body-example`, `files` (filename), `validate-parameter`.
>   - Tests: unit tests for all new components and helpers.
> - **Workspace Store (Schemas)**:
>   - Add extensions: `x-is-disabled` on `Example`, `x-is-readonly` and `x-is-global` on `Parameter`; integrate via `compose`.
> - **Config**:
>   - ESLint: disable `vue/custom-event-name-casing`.
>   - Changeset: minor bumps for `@scalar/workspace-store` and `@scalar/api-client`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit cca731df2dc5e7758c5976b62a8c310bce761a83. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/workspace-store/src/schemas/extensions/example/x-is-disabled.ts
```diff
@@ -0,0 +1,13 @@
+import { Type } from '@scalar/typebox'
+
+/**
+ * An OpenAPI extension to mark examples as enabled or disabled
```
hanspagel: Can you add to this comment so I understand why an example should be disabled or what the consequence would be?

Just saying disabled disables it, doesn’t help present me, and future me even less. 👀 

Same for all those comments in the added extensions.

packages/workspace-store/src/schemas/v3.1/strict/example.ts
```diff
@@ -1,20 +1,26 @@
...
+    summary: Type.Optional(Type.String()),
+    /** Long description for the example. CommonMark syntax MAY be used for rich text representation. */
+    description: Type.Optional(Type.String()),
+    /** Embedded literal example. The value field and externalValue field are mutually exclusive. To represent examples of media types that cannot naturally represented in JSON or YAML, use a string value to contain the example, escaping where necessary. */
+    value: Type.Optional(Type.Any()),
+    /** A URI that identifies the literal example. This provides the capability to reference examples that cannot easily be included in JSON or YAML documents. The value field and externalValue field are mutually exclusive. See the rules for resolving Relative References. */
+    externalValue: Type.Optional(Type.String()),
+  }),
+  XIsDisabled,
+)
```
hanspagel, line 23: In the api client store we added our custom specification extensions all over the place and it ended up being super confusing.

Also, eventually we might want to use those to validate any file, and an OpenAPI document is still valid when someone uses a specification with another structure.

So what I’m thinking is: Should we put the OpenAPI schemas in one place and do the `compose` in another file, so we have both (the original OpenAPI schema and the OpenAPI schema + our extensions).

Wdyt?

packages/workspace-store/src/schemas/v3.1/strict/parameter.ts
```diff
@@ -41,6 +47,10 @@ type ParameterObjectBase = {
   deprecated?: boolean
   /** If true, clients MAY pass a zero-length string value in place of parameters that would otherwise be omitted entirely, which the server SHOULD interpret as the parameter being unused. Default value is false. If style is used, and if behavior is n/a (cannot be serialized), the value of allowEmptyValue SHALL be ignored. Interactions between this field and the parameter's Schema Object are implementation-defined. This field is valid only for query parameters. Use of this field is NOT RECOMMENDED, and it is likely to be removed in a later revision. */
   allowEmptyValue?: boolean
+  /** An OpenAPI extension to mark parameters as read-only */
+  'x-is-readonly'?: boolean
+  /** An OpenAPI extension to mark parameters as global */
+  'x-is-global'?: boolean
```
hanspagel: soo the OpenAPI-way to name booleans would be just `x-readonly` and `x-disabled` (see `deprecated` in the Operation Object)

so it looks a bit foreign to me, looking at it with OpenAPI-eyes

But I leave it up to you

hanspagel (changes requested):
LOVE IT

just left some thoughts

packages/workspace-store/src/schemas/v3.1/strict/example.ts
amritk: So I thought about this one before, we should move these over to the openapi-types package. BUT I would rather keep these typed extensions in the schema as they don't really hurt anything. 

If you want cleaner types though we can do that easily. We can export the types as we do here + we can just export openapi versions. The types are hard coded now so it should be trivial

amritk: Or we can just add the wild card to our types. Like so:
image
best of all worlds?

packages/workspace-store/src/schemas/v3.1/strict/parameter.ts
amritk: sounds reasonable

PR #6981, opened 2025-10-01
fix(json-magic): hides properties starting with an underscore

### LINKED ISSUE #6931: Underscored schema properties do not appear?
### What happens?

Hi there, sorry if I'm missing something, but I couldn't find anything related to my issue. I have this spec:

```
components:
  schemas:
    ErrorResponse:
      type: object
      properties:
        __v:
          type: number
          example: 1
```

Unfortunately, it does not show as expected:

Image

If I rename the field to not start with `_` (doesn't matter single or double), it shows properly:

Image

---

Am I missing anything, is this by design, or a bug? Thank you!

To show all properties of a schema, regardless if they start with underscore or not.

### OpenAPI Document

_No response_

hanspagel:
**Problem**

We consider properties starting with an _underscore as internal, that hides legit properties starting with an _underscore, too.

**Solution**

Pragmatic solution: With this PR, we’re using `__scalar_` as a prefix, which is just more unlikely to interfere with real user data.

Open for alternative approaches to address the issue. :)

Fixes #6931

**Checklist**

### DIFF
--- a/.changeset/slimy-lemons-judge.md
+++ b/.changeset/slimy-lemons-judge.md
@@ -1,0 +1,5 @@
+---
+'@scalar/json-magic': patch
+---
+
+fix: schema properties starting with an underscore are hidden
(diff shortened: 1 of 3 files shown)


DemonHa:
Also you need to change the fake schema property for the "SchemaObjectDefinition" to use this new hidden property so we don't display to the user `_` property for the schema

amritk (approved):
Merge itttt

PR #6985, opened 2025-10-01
feat(api-reference): use the new schema example generation for response

amritk:
**Problem**

Currently, we used the new generation for the code sample but not the response

**Solution**

With this PR we use the new one + deprecate the old one since its still being used in a few other places that aren't on the new store yet.

**Checklist**


DemonHa (approved):

DemonHa:
there are other places where we use the old function, should remove the old one as soon as we migrate to the new store

hanspagel:
@amritk pls fix types

image

amritk:
This is the forgotten PR

packages/components/src/components/ScalarCodeBlock/ScalarCodeBlock.vue
```diff
@@ -5,33 +5,59 @@ import { computed, useId } from 'vue'
 
 import { ScalarCodeBlockCopy } from '../ScalarCodeBlock'
 
+type BaseProps = {
+  content?: string | object
```
hwkr, line 9: I think because you have content here it would be allowed alongside prettyPrintedContent which I'm not sure if is what you want

amritk: Yea we just want to make sure it has at least one, the other would be optional

PR #7001, opened 2025-10-04
Geoff/cleanup config types

geoffgscott:
**Problem**

The configurations should be agnostic of any document sources. In the future it is likely no sources will be provided in the configuration.

**Solution**

This PR removes the sources from the base `ApiReferenceConfig` type to discourage the access of the sources anywhere except at the application entrypoints. Follow up work will streamline the document ingress. 

**Checklist**

### DIFF
--- a/examples/web/src/pages/ApiReferencePage.vue
+++ b/examples/web/src/pages/ApiReferencePage.vue
@@ -1,8 +1,6 @@
 <script setup lang="ts">
-import {
-  ApiReferenceLayout,
-  type ApiReferenceConfiguration,
-} from '@scalar/api-reference'
+import { ApiReferenceLayout } from '@scalar/api-reference'
+import type { ApiReferenceConfigurationWithSource } from '@scalar/types'
 import { useColorMode } from '@scalar/use-hooks/useColorMode'
 import { createWorkspaceStore } from '@scalar/workspace-store/client'
 import { computed, reactive, ref, watch } from 'vue'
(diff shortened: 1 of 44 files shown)


hanspagel:
not ready for review yet

examples/web/src/pages/ApiReferencePage.vue
```diff
@@ -35,7 +33,7 @@ const DEFAULT_CONTENT = {
 
 const content = ref(JSON.stringify(DEFAULT_CONTENT, null, 2))
 
-const configuration = reactive<Partial<ApiReferenceConfiguration>>({
+const configuration = reactive<Partial<ApiReferenceConfigurationWithSource>>({
```
amritk, line 36: Do we want to allow the array of configs? If so then you might want `AnyApiReferenceConfiguration`

packages/api-client/README.md
```diff
@@ -97,7 +97,7 @@ open()
 /** Configuration options for the Scalar API client */
 export type ClientConfiguration = {
   /** The Swagger/OpenAPI document to render */
-  spec: SpecConfiguration
+  spec: SourceConfiguration
```
amritk, line 100: using spec is old but this will be gone anyway on the new store

packages/api-reference/src/features/document-source/hooks/useDocumentSource.ts
```diff
@@ -129,10 +129,16 @@ export function useDocumentSource({
     { immediate: true },
   )
 
+  const config = toValue(configuration) ?? apiReferenceConfigurationSchema.parse({})
```
amritk, line 132:
```suggestion
  const config = apiReferenceConfigurationSchema.parse(toValue(configuration) ?? {})
```
I don't get this, why not just parse the input?

packages/types/src/api-reference/api-client-configuration.ts
```diff
@@ -0,0 +1,8 @@
+import type z from 'zod'
+
+import { baseConfigurationSchema } from './base-configuration'
+import { sourceConfigurationSchema } from './source-configuration'
+
+export const apiClientConfigurationSchema = baseConfigurationSchema.extend(sourceConfigurationSchema.shape)
+
+export type ApiClientConfiguration = z.infer<typeof apiClientConfigurationSchema>
```
amritk, line 8: are we making a separate client config? because the references config should be a super set of the client config

PR #7005, opened 2025-10-04
feat: implement environment variables management in api reference

ankitfxz:
**Problem**

Currently, there is no way to manage environment variables in the api reference defined through `x-scalar-environments ` in the openapi spec. If we have a value bieng repeatedly used in various endpoints, we have no way to define it globally and access it as an variable, for example, like in postman.

**Solution**

With this PR, we will able to view, edit, delete the variables defined in the openapi spec.

**Checklist**


hanspagel:
ha, amazing work! we’re swapping out our data layer, this PR has to wait a bit, though 😳

hanspagel:
I have to close this for now. I like the idea, I think we want to add it, but we need to invest some design time here and focus on a few other things first. 😬  Will reopen once we’re ready.

PR #7025, opened 2025-10-07
chore: update playwright to 1.56.0

hwkr:
Updates docker and hopefully fixes the version issue by setting it in the docker file.

Fixes DOC-4235

**Checklist**

### DIFF
--- a/scripts/src/commands/playwright-docker/Dockerfile
+++ b/scripts/src/commands/playwright-docker/Dockerfile
@@ -10,7 +10,8 @@ ENV PORT=7001
 
 ENV PLAYWRIGHT_BROWSERS_PATH=/ms-playwright
 RUN npm install -g npm@latest
-RUN npx -y playwright@$PLAYWRIGHT_VERSION install --with-deps
+RUN npm install -g @playwright/test@$PLAYWRIGHT_VERSION
+RUN playwright install --with-deps
 RUN chmod -R 777 /ms-playwright
 
-CMD npx -y playwright run-server --port ${PORT} --host 0.0.0.0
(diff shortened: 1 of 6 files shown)


scripts/src/commands/playwright-docker/Dockerfile
```diff
@@ -13,4 +13,4 @@ RUN npm install -g npm@latest
 RUN npx -y playwright@$PLAYWRIGHT_VERSION install --with-deps
 RUN chmod -R 777 /ms-playwright
 
-CMD npx -y playwright run-server --port ${PORT} --host 0.0.0.0
+CMD npx -y playwright@$PLAYWRIGHT_VERSION run-server --port ${PORT} --host 0.0.0.0
```
hwkr: @geoffgscott I think this is why we keep getting errors since `npx` is pulling latest playwright whenever the tests are run but the browser / install is fixed to a specific version.

hwkr: But also I don't know if we can put this here since it's a docker build arg, does that exist in this context?

geoffgscott: Pretty sure this can just be 
```
pnpm playwright run-server --port ${PORT} --host 0.0.0.0
```
If not the above is fine.

hwkr: Yeah we might need to install it first though? I'll play with it

hwkr: Okay updated based on testing.

hanspagel (approved):
let’s try

PR #7026, opened 2025-10-07
fix(schema): correctly handle `additionalProperties` when set to false

xC0dex:
**Problem**

Currently, we don't handle the `additionalProperties: false` case correctly.

**Solution**

This PR explicitly checks when `additionalProperties` is set to `false` and ensures that additional properties are correctly restricted.

**Checklist**


packages/api-client/src/v2/blocks/operation-code-sample/helpers/get-example-from-schema.ts
```diff
@@ -242,7 +242,11 @@ const handleObjectSchema = (
     }
   }
 
-  if ('additionalProperties' in schema && schema.additionalProperties !== undefined) {
+  if (
+    'additionalProperties' in schema &&
+    schema.additionalProperties !== undefined &&
+    schema.additionalProperties !== false
+  ) {
```
xC0dex: I'm unsure if I can simplify the condition to:

```suggestion
  if ('additionalProperties' in schema && schema.additionalProperties) {
```

hanspagel: that should work!

hanspagel:
ah, I missed that we copied that code. yeah, we should keep both in sync

PR #7033, opened 2025-10-07
fix: unhead dep ^2.0.17

Saganic:
**Problem**

Currently, the scalar/api-reference package has a dependancy on @unhead/vue^1.x.x, this creates issues in Nuxt projects that are using @unhead/vue^2.x.x

**Solution**

With this PR we update scalar/api-reference to use @unhead/vue^2.0.17 and unhead^2.0.17 (useSeoMeta composable was moved to unhead).

**Checklist**

### DIFF
--- a/.changeset/twenty-laws-clap.md
+++ b/.changeset/twenty-laws-clap.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': major
+---
+
+Updates use of @unhead/vue and moved useSeoMeta to unhead. References upstream issue with peer packages dependant on versions < 2.0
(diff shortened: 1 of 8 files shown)


hanspagel:
@amritk can we use the `@unhead/vue/client` here? I mean, we don’t really support SSR yet, but last time I checked it worked somehow on the server-side at least.

I just wanted to try the SSR playground, but it’s broken anyway. 🤡

.changeset/twenty-laws-clap.md
```diff
@@ -0,0 +1,5 @@
+---
+'@scalar/api-reference': major
```
xC0dex, line 2: Just wanted to verify that major release is correct here.

Saganic: Yeah, I actually wasn't sure if it was to be honest

Saganic:
> Looks great. Thanks for the PR!

Anything I can do to help get this released? We need it for a bunch of projects.

Saganic:
It looks like this doesn't actually solve the issue with using both @nuxt/content and @scalar/nuxt at the same time! Back to the drawing board!

PR #7037, opened 2025-10-08
feat: Migrate document handling to the workspace store

geoffgscott:
This eliminates most of the legacy document handling. A mapping layer to the client store is created to handle keeping the client store aligned during the migration period.


packages/api-reference/src/v2/blocks/scalar-auth-selector-block/components/AuthSelector.vue
```diff
@@ -158,7 +159,9 @@ function updateSelectedAuth(entries: SecuritySchemeOption[]) {
 
 const editSelectedSchemeUids = (uids: SelectedSecuritySchemeUids) => {
   // Set as selected on the collection for the modal
-  if (collection.useCollectionSecurity) {
+  if (collection.value?.useCollectionSecurity) {
+    console.log('editSelectedSchemeUids', uids)
```
hanspagel: not needed, eh?

packages/api-reference/src/v2/blocks/scalar-auth-selector-block/components/AuthSelector.vue
```diff
@@ -185,6 +188,8 @@ const editSelectedSchemeUids = (uids: SelectedSecuritySchemeUids) => {
   }
   // Set as selected on request
   else if (operation?.uid) {
+    console.log('editSelectedSchemeUids', uids)
```
hanspagel: not needed, eh?

packages/api-reference/src/v2/blocks/scalar-info-block/components/DownloadLink.test.ts
```diff
@@ -244,73 +217,12 @@ describe('DownloadLink', () => {
...
-      // Create a mock event that can be prevented
-      const mockEvent = {
-        preventDefault: vi.fn(),
-        stopPropagation: vi.fn(),
-      }
-
-      await link.trigger('click', mockEvent)
-
-      expect(mockEvent.preventDefault).toHaveBeenCalled()
-    })
```
hanspagel, line 314: do we really want to delete those?

packages/api-reference/src/v2/blocks/scalar-info-block/components/DownloadLink.vue
```diff
@@ -49,21 +48,6 @@ const handleDownloadClick = (format: 'json' | 'yaml') => {
...
     </button>
   </div>
-  <template v-else-if="documentDownloadType === 'direct'">
-    <a
-      v-if="url"
-      class="download-link"
-      :href="url">
-      Download OpenAPI Document
-    </a>
-    <a
```
hanspagel, line 59: the purpose of 'direct' was to a have a[href] with the actual URL, so we need to add this again

packages/api-reference/src/v2/helpers/map-config-to-workspace-store.ts
```diff
@@ -0,0 +1,61 @@
+import { isClient } from '@scalar/api-client/v2/blocks/operation-code-sample'
```
hanspagel, line 1: can we have a few basic tests for it? I mean, we want to get rid of it anyway, but we constantly break untested things

packages/api-reference/src/v2/helpers/map-config-to-client-store.ts
```diff
@@ -0,0 +1,245 @@
+import { createApiClientModal } from '@scalar/api-client/layouts/Modal'
```
hanspagel, line 1: can we have a few basic tests for it? I mean, we want to get rid of it anyway, but we constantly break untested things

packages/api-reference/src/types.ts
```diff
@@ -1,50 +1,10 @@
...
-  /**
-   * @deprecated Use `originalDocument` instead.
-   */
-  rawSpec?: string
-  /**
-   * @deprecated this is just temporary until we switch to the new store, we prop drill it down
-   */
-  store: WorkspaceStore
-}
-
```
hanspagel, line 47: good bye old friend

packages/api-reference/vite.config.ts
```diff
@@ -17,6 +19,7 @@ export default defineConfig({
     ...(process.env.REMOTE_CONTAINERS && { host: '127.0.0.1' }),
     allowedHosts: ['localhost', 'host.docker.internal'],
   },
+  minify: false,
```
hanspagel, line 22: we don’t want to minify anymore? to speed up the build, I guess?

packages/api-reference/vitest.config.ts
```diff
@@ -13,4 +13,13 @@ export default mergeConfig(
...
   }),
+  {
+    resolve: {
+      alias: {
+        '@': fileURLToPath(new URL('./src', import.meta.url)),
+        '@v2': fileURLToPath(new URL('./src/v2', import.meta.url)),
+        '@test': fileURLToPath(new URL('./test', import.meta.url)),
+      },
+    },
+  },
```
hanspagel: u sure we need this here? should be coming from the vite.config.ts already (using mergeConfig here) 🤔

packages/openapi-parser/src/utils/dereference.ts
```diff
@@ -8,6 +9,8 @@ export type DereferenceOptions = ResolveReferencesOptions
 
 /**
  * Resolves all references in an OpenAPI document
+ *
+ * @TODO: Why is this async?
```
hanspagel, line 13:
```suggestion
/**
* Resolves all references in an OpenAPI document
 *
 * @TODO: this was needed for some reason for the strange pipeline syntax that we want to get rid of `openapi().load().dereference()` once that’s removed, we can make this synchronous only
*/
export async function dereference(
value: AnyApiDefinitionFormat | Filesystem,
@@ -25,3 +28,23 @@ export async function dereference(
...details(entrypoint.specification),
}
}
```

packages/types/src/api-reference/api-reference-configuration.ts
```diff
@@ -58,6 +58,7 @@ export const apiReferenceConfigurationSchema = baseConfigurationSchema.extend({
   /**
    * Sets the file type of the document to download, set to `none` to hide the download button
    * @default 'both'
+   * @comment 'direct' is deprecated and will be downloaded as JSON
```
hanspagel: nooo, why. this was requested multiple times

packages/api-reference/src/v2/blocks/scalar-info-block/components/InfoBlock.vue
```diff
@@ -26,18 +26,11 @@ const { options } = defineProps<{
...
   options: {
-    /** Indicates if the info block is in a loading state. */
-    isLoading?: boolean
     /** Determines the layout style for the info block ('modern' or 'classic'). */
     layout?: 'modern' | 'classic'
     /** The document download type. */
     documentDownloadType?: ApiReferenceConfiguration['documentDownloadType']
-    /** The URL of the OpenAPI document. */
-    url?: string | undefined
     /** Optional callback invoked when the component has finished loading. */
```
hanspagel, line 33:
```suggestion
```

stale comment

packages/api-reference/src/v2/blocks/scalar-info-block/components/DownloadLink.test.ts
geoffgscott: Yes. It does not behave in the same way anymore. There is no link just a custom event.

packages/api-reference/vite.config.ts
geoffgscott: We only want to minify the CDN and build projects. Otherwise the wrapping integration can minify and even then it obfuscates logs.

packages/api-reference/vitest.config.ts
geoffgscott: You are probably right.

packages/api-reference/src/v2/blocks/scalar-info-block/components/DownloadLink.test.ts
hanspagel: ah, got it. as commented below, this was added because people were requesting the feature, so I think we want to keep the feature and the tests for it

packages/types/src/api-reference/api-reference-configuration.ts
geoffgscott: Alright I will add it back in

geoffgscott:
> * [ ]  we render the loading skeleton now, do we want that? it’s not really reflecting the actual layout (anymore). let’s just delete it?

Documents are async loaded now by default.

PR #7039, opened 2025-10-08
draft: adds support for oai 3.2.0 in types

baywet:
This pull request is still work in progress, opening it for visibility and early feedback.

It adds the type definitions for OAI 3.2.0.

Here is what's "completed" (unless I messed something up):

- feat: adds xmlObject nodeType field
- feat: adds mediaType components
- feat: adds server name field
- feat: adds securityScheme deprecated field
- feat: adds path item query and additionalOperations fields
- feat: adds tag parent, kind, summary fields
- feat: adds discriminator default mapping field

TODO:

- [ ] https://github.com/BinkyLabs/openapi.net/issues/14
- [ ] https://github.com/BinkyLabs/openapi.net/issues/13
- [ ] https://github.com/BinkyLabs/openapi.net/issues/10
- [ ] https://github.com/BinkyLabs/openapi.net/issues/9
- [ ] https://github.com/BinkyLabs/openapi.net/issues/8
- [ ] https://github.com/BinkyLabs/openapi.net/issues/7
- [ ] https://github.com/BinkyLabs/openapi.net/issues/6

**Checklist**

### DIFF
--- a/packages/openapi-types/src/openapi-types.ts
+++ b/packages/openapi-types/src/openapi-types.ts
@@ -64,10 +64,207 @@ export namespace OpenAPI {
 }
 
 // biome-ignore lint/style/noNamespace: We want it to be a module here.
+export namespace OpenAPIV3_2 {
+  type Modify<T, R> = Omit<T, keyof R> & R
+
+  type PathsWebhooksComponents<T = {}> = Modify<
+    OpenAPIV3_1.PathsWebhooksComponents<T>,
+    {
+      mediaTypes?: Record<string, MediaTypeObject | ReferenceObject>
+    }
+  >
+
+  export type Document<T = {}> = Modify<
+    Omit<OpenAPIV3_1.Document<T>, 'paths' | 'components'>,
+    {
+      /**
+       * Version of the OpenAPI specification
+       * @see https://github.com/OAI/OpenAPI-Specification/tree/main/versions
+       */
+      openapi?: '3.2.0'
+      swagger?: undefined
+      info?: InfoObject
+      jsonSchemaDialect?: string
+      servers?: ServerObject[]
+    } & (
+      | (Pick<PathsWebhooksComponents<T>, 'paths'> & Omit<Partial<PathsWebhooksComponents<T>>, 'paths'>)
+      | (Pick<PathsWebhooksComponents<T>, 'webhooks'> & Omit<Partial<PathsWebhooksComponents<T>>, 'webhooks'>)
+      | (Pick<PathsWebhooksComponents<T>, 'components'> & Omit<Partial<PathsWebhooksComponents<T>>, 'components'>)
+    ) &
+      T &
+      AnyOtherAttribute
+  >
+
+  export type InfoObject = OpenAPIV3_1.InfoObject
+
+  export type ContactObject = OpenAPIV3_1.ContactObject
+
+  export type LicenseObject = OpenAPIV3_1.LicenseObject
+
+  export type ServerObject = Modify<
+    OpenAPIV3_1.ServerObject,
+    {
+      name?: string
+    }
+  >
+
+  export type ServerVariableObject = OpenAPIV3_1.ServerVariableObject
+
+  export type PathsObject<T = {}, P extends {} = {}> = Record<string, (PathItemObject<T> & P) | undefined>
+
+  export type HttpMethods = OpenAPIV3_1.HttpMethods | 'query'
+
+  export type PathItemObject<T = {}> = Modify<
+    OpenAPIV3_1.PathItemObject<T>,
+    {
+      [method in HttpMethods]?: OperationObject<T>
+    }
+  > & {
+    additionalOperations?: Record<string, OperationObject<T>>
+  }
+
+  export type OperationObject<T = {}> = Modify<
+    OpenAPIV3_1.OperationObject<T>,
+    {
+      parameters?: (ReferenceObject | ParameterObject)[]
+      requestBody?: ReferenceObject | RequestBodyObject
+      responses?: ResponsesObject
+      callbacks?: Record<string, ReferenceObject | CallbackObject>
+      servers?: ServerObject[]
+    }
+  > &
+    T
+
+  export type ExternalDocumentationObject = OpenAPIV3_1.ExternalDocumentationObject
+
+  export type ParameterObject = OpenAPIV3_1.ParameterObject
+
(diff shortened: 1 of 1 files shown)


hanspagel:
Amazing!

baywet:
@hanspagel this should be good for review now. A couple of notes as I was finalizing things:
- noticed that parameter location and style union types were not here at all (simply string). Added their definitions, and that lead to a few modifications in the converter package for things to keep building.
- noticed the parameter and header 3.1 content properties were missing, added them
- refactored the Oauth2SecuritySchem to use an OAuthFlows object to reduce duplication across versions and be closer to the specification

Let me know if you have any additional comments or questions.

packages/openapi-upgrader/src/2.0-to-3.0/upgrade-from-two-to-three.ts
```diff
@@ -364,16 +377,21 @@ function transformParameterObject(parameter: OpenAPIV2.ParameterObject): OpenAPI
   delete parameter.collectionFormat
   delete parameter.default
 
+  if (!parameter.in) {
+    throw new Error('Parameter object must have an "in" property')
+  }
```
baywet, line 390: This is causing a lot of unit tests to fail, but not having it is technically invalid in every version of the specification. So I'm not sure which way you'd like to go here?

packages/openapi-upgrader/src/2.0-to-3.0/upgrade-from-two-to-three.ts
```diff
@@ -352,9 +354,20 @@ function transformItemsObject<T extends Record<PropertyKey, unknown>>(obj: T): O
...
+  return location as OpenAPIV3.ParameterLocation
+}
+
+function transformParameterObject(
+  parameter: OpenAPIV2.ParameterObject | OpenAPIV2.ReferenceObject,
+): OpenAPIV3.ParameterObject | OpenAPIV3.ReferenceObject {
+  if (Object.hasOwn(parameter, '$ref') && '$ref' in parameter) {
+    return {
+      $ref: parameter.$ref,
+    }
```
baywet, line 378: cursor is complaining about this update. this is simply to better match the type definitions and the differences between a parameter object and a reference object. This method is called both for the components section and the path item/operations. I believe it'll be the responsibility of the caller to ensure it does not pass a reference in if it doesn't want one back. Let me know if you'd like to take a different approach.

packages/openapi-upgrader/src/2.0-to-3.0/upgrade-from-two-to-three.ts
```diff
@@ -352,9 +354,20 @@ function transformItemsObject<T extends Record<PropertyKey, unknown>>(obj: T): O
   }, {} as OpenAPIV3.SchemaObject)
 }
 
-function transformParameterObject(parameter: OpenAPIV2.ParameterObject): OpenAPIV3.ParameterObject {
-  if (Object.hasOwn(parameter, '$ref')) {
-    return parameter
+function getParameterLocation(location: OpenAPIV2.ParameterLocation): OpenAPIV3.ParameterLocation {
+  if (location === 'formData') {
+    return 'query'
+  }
```
baywet, line 365: We should ideally never fall into that condition, since the caller would already have filtered out body and form data parameters. I think this is potentially revealing a bug from the caller (like the location thing) and I'm tempted to throw an error if that condition is true (like I do in location). But I'd like some input.

packages/openapi-types/src/openapi-types.ts
```diff
@@ -354,7 +638,7 @@ export namespace OpenAPIV3 {
   // See https://swagger.io/specification/#path-item-object
   // You can use keys or values from it in TypeScript code like this:
   //     for (const method of Object.values(OpenAPIV3.HttpMethods)) { … }
-  export type HttpMethods = 'get' | 'put' | 'post' | 'delete' | 'options' | 'head' | 'patch' | 'trace' | 'connect'
+  export type HttpMethods = OpenAPIV2.HttpMethods | 'trace'
```
baywet, line 642: I removed connect here since it's invalid according to the specification. This is causing some vue component (packages\api-reference\src\features\test-request-button\TestRequestButton.vue#L17) to fail. And I have zero experience with vue. The caller should not try to assign connect since it's invalid according to the spec.
It should either:
- use an x-connect on the path item object for < 3.2.0
- use the additional operations field in 3.2.0

baywet:
I've also left comments where some of my changes are causing side effects in consumers of the package and I need some help.
Additionally to the comments I've left the **narrows it down to Swagger 2.0** test is failing because of a never type which I'm not able to track down and that was introduced by changes added with https://github.com/scalar/scalar/pull/7039/commits/a69ac364cd9855e79e52bbf98e3c56ee5aa358b6

packages/openapi-upgrader/src/2.0-to-3.0/upgrade-from-two-to-three.ts
baywet: VeteGIF

packages/openapi-upgrader/src/2.0-to-3.0/upgrade-from-two-to-three.ts
hanspagel: throw an error 👍

packages/openapi-upgrader/src/2.0-to-3.0/upgrade-from-two-to-three.ts
hanspagel: we have to fix the tests then :) I look into that

packages/openapi-upgrader/src/2.0-to-3.0/upgrade-from-two-to-three.ts
baywet: thanks! updated.

hanspagel:
Got some fixes on my machine already, can’t push for whatever reason. I’ll figure it out.

hanspagel:
Closing in favor of #7071

hanspagel:
We’re so back

packages/openapi-types/src/openapi-types.ts
hanspagel: let’s remove it

baywet:
Thanks for pushing it across the finish line! Can you let me know here once that actually releases please? I'd like to pull the latest and greatest changes in my WIP PR for TypeSpec.

PR #7047, opened 2025-10-08
fix(types): allow async callbacks for event handlers

### LINKED ISSUE #7046: async `onBeforeRequest` are not working due to a zod error accepting only sync void functions
### What happens?

I'm trying to implement an asynchronous `onBeforeRequest`.
On [documentation](https://guides.scalar.com/scalar/scalar-api-references/configuration#configuration__configuration-options__events__onbeforerequest) is stated that it can be both sync and async.
However when writing asynchronous code I encounter this error:

Image

<details><summary>Error stack</summary>
<p>

```text
scalar.js:87 Uncaught (in promise) $ZodError: [
  {
    "expected": "void",
    "code": "invalid_type",
    "path": [],
    "message": "Invalid input"
  }
]
    at scalar.js:87:61
    at scalar.js:564:23
    at scalar.js:563:52
    at scalar.js:29660:55
    at Array.map (<anonymous>)
    at Object.executeHook (scalar.js:29660:44)
    at Object.sendRequest (scalar.js:33570:23)
    at S (scalar.js:33610:21)
    at scalar.js:26375:33
    at Set.forEach (<anonymous>)
```

</p>
</details> 

The error is likey caused by the zod validation coming from here:
https://github.com/scalar/scalar/blob/63ff417e6bf5574280f8da2794dab81d5d05a842/packages/types/src/api-reference/api-reference-configuration.ts#L140-L142

### Reproduction:

1. Clone https://github.com/marcalexiei/fastify-type-provider-zod-playground
2. Run `pnpm i`
3. Run `pnpm --filter=multipart run dev`
4. Open http://localhost:3000/docs
5. Run the only available API
6. Check console for the error

I can write asynchronous code in `onBeforeRequest` without getting a zod validation error

### OpenAPI Document

_No response_

marcalexiei:
**Problem**

- closes #7046,\
  all relevant info can be found there

**Solution**

The validation for `onBeforeRequest`, `onShowMore`, and `onSidebarClick` was incorrectly enforcing a void return type, which prevented asynchronous methods from being used. 
According to the documentation, these handlers are intended to support both synchronous and asynchronous callbacks.

This update removes the output validation from Zod, 
allowing the handlers to properly support `Promise<void> | void` return types 
and align with the documented behaviour.

From my understanding, this approach is already used for the onDocumentSelect function 
to continue using parse instead of parseAsync. 
Let me know if you’d like me to add comments in the code to clarify the distinction 
between  type definitions and runtime behavior.

**Checklist**

- [x] I've updated the documentation. (no update needed)

---

> [!NOTE]
> Await the plugin onBeforeRequest hook before sending requests and update types to accept async callbacks for event handlers, with tests adjusted accordingly.
> 
> - **API Client**:
>   - Await `pluginManager.executeHook('onBeforeRequest', { request })` in `packages/api-client/src/libs/send-request/create-request-operation.ts` to ensure hooks complete before `fetch`.
> - **Types**:
>   - Relax Zod schemas to allow async callbacks by removing enforced `output` types for `onDocumentSelect`, `onLoaded`, `onBeforeRequest`, `onShowMore`, `onSidebarClick` in `packages/types/src/api-reference/api-reference-configuration.ts`.
>   - Similarly remove `output` constraints for plugin hooks in `packages/types/src/api-reference/api-client-plugin.ts`.
> - **Tests**:
>   - Update and add tests in `packages/api-client/...create-request-operation.test.ts` to verify `onBeforeRequest` is called and awaited (including header mutation).
>   - Extend `packages/types/...api-reference-configuration.test.ts` to validate sync/async handlers for the above callbacks.
> - **Changesets**:
>   - Patch releases for `@scalar/api-client` and `@scalar/types`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 231a5e90ca99237f146b541cb3b59787a5b69398. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/api-client/src/libs/send-request/create-request-operation.test.ts
+++ b/packages/api-client/src/libs/send-request/create-request-operation.test.ts
@@ -1241,12 +1241,61 @@ describe('create-request-operation', () => {
   })
 
   it('executes onBeforeRequest hook before making the request', async () => {
-    let hookExecuted = false
-
     const mockPluginManager = {
-      executeHook: vi.fn().mockImplementation(async () => {
-        hookExecuted = true
-        return Promise.resolve()
+      executeHook: vi.fn().mockResolvedValue(void 0),
+      getViewComponents: vi.fn().mockReturnValue([]),
+    }
+
+    // Store original fetch
+    const originalFetch = global.fetch
+    try {
+      // Mock fetch to return a successful response
+      const mockResponse = new Response('{"test": "data"}', {
+        status: 200,
+        headers: new Headers({
+          'content-type': 'application/json',
+        }),
+      })
+
+      global.fetch = vi.fn().mockResolvedValue(mockResponse)
+
+      const [error, requestOperation] = createRequestOperation({
+        ...createRequestPayload({
+          serverPayload: { url: 'https://api.example.com' },
+          requestPayload: {
+            path: '/test',
+          },
+        }),
+        pluginManager: mockPluginManager,
+      })
+
+      if (error) {
+        throw error
+      }
+
+      await requestOperation.sendRequest()
+
+      expect(mockPluginManager.executeHook).toHaveBeenCalled()
+      expect(mockPluginManager.executeHook).toHaveBeenCalledWith('onBeforeRequest', {
+        request: expect.any(Request),
+      })
+    } finally {
+      // Restore original fetch
+      global.fetch = originalFetch
+    }
+  })
+
+  // https://github.com/scalar/scalar/pull/7047#pullrequestreview-3319608975
+  it('executes onBeforeRequest and await hook execution before sending it', async () => {
+    const HEADER_MOCK_NAME = 'header-request-body'
+    const HEADER_MOCK_VALUE = 'yes'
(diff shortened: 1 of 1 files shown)


packages/api-client/src/libs/send-request/create-request-operation.ts
```diff
@@ -190,7 +190,7 @@ export const createRequestOperation = ({
       status?.emit('start')
 
       if (pluginManager) {
-        pluginManager.executeHook('onBeforeRequest', { request: proxiedRequest })
+        await pluginManager.executeHook('onBeforeRequest', { request: proxiedRequest })
```
marcalexiei, line 193: [This should resolve the problem where everything after an await inside `onBeforeRequest` was ignored.](https://github.com/scalar/scalar/pull/7047#pullrequestreview-3319608975)

hanspagel:
tested and it works great! awesome, thanks for fixing this

PR #7050, opened 2025-10-09
feat(api-reference): editable config in dev toolbar

hwkr:
Makes the dev toolbar config editable (and synced with the toggles). It was hard to get the config to sync with the toggles in a way that made sense but I think this is okay?

https://github.com/user-attachments/assets/c63b6136-8c10-4a16-b55e-d6d1b54236c4

**Checklist**

### DIFF
--- a/packages/api-reference/src/features/toolbar/ApiReferenceToolbarConfigEditor.vue
+++ b/packages/api-reference/src/features/toolbar/ApiReferenceToolbarConfigEditor.vue
@@ -1,4 +1,5 @@
 <script lang="ts" setup>
+import { ScalarCodeBlockCopy } from '@scalar/components'
 import { prettyPrintJson } from '@scalar/oas-utils/helpers'
 import type { ApiReferenceConfiguration } from '@scalar/types'
 import { useCodeMirror } from '@scalar/use-codemirror'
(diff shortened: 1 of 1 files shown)


hanspagel:
ha, that’s so cool!

hwkr:
@hanspagel 

> can we make it so, that if I hit return, the next lines start with the correct indentation, so I don’t have to hit space 2 times?

Seems like this is something with codemirror, the client has the same behavior unfortunately. I'd like to switch this to Monaco anyway so maybe we wait and that should fix it?

> we have some strange attributes there by default (slug, title, default), but not the actual document URL, which is probably coming from the recent changes to the configuration. I think it’s a bit confusing.

Yeah I tried to clean it up best I could, maybe there's a better source of truth for the generated config?

> It shows "esc then tab" but esc already closes the overlay, that’s not how it’s supposed to work, is it?

Good catch, I should probably capture the esc here

hanspagel:
Off-topic: I don’t think we can use Monaco in the self-hosted version. It requires some kind of sophisticated web worker setup. And it’s just way bigger than CodeMirror. If we want it - let’s say for the Scalar App - we could make Monaco a plugin for the API Client package.

geoffgscott:
Let's put this one on hold for right now. I have some uncertainty about config reactivity and this is low priority for right now.

hanspagel:
I close it for now, we can reopen any time. I just like the PR to be clean.
I still love the feature BTW

PR #7060, opened 2025-10-10
fix(api-client): create-request-operation - await from pluginManager onResponseReceived executeHook

marcalexiei:
**Problem**

- Followup to #7047.

While reviewing usages of `pluginManager.executeHook`, I found another instance of a non-awaited invocation in  
`packages/api-client/src/libs/send-request/create-request-operation.ts`.

---

**Solution**

As in **#7047**, I added an `await` before  `pluginManager.executeHook('onResponseReceived', /* ... */)`  
to ensure the hook completes before continuing execution.

---

**Additional improvement on `create-request-operation.test.ts`**

Refactored tests to use `vi.spyOn` for mocking `global.fetch`.  
Previously, each test wrapped its logic in a `try/catch` to restore the mock, which was repetitive.

This change simplifies test setup and teardown.  
It's committed separately so it can be easily reverted if the previous implementation is preferred.

---

**Checklist**

---

> [!NOTE]
> Await the pluginManager onResponseReceived hook and refactor tests to use a global fetch spy with added assertions around hook timing, empty bodies, and streaming.
> 
> - **api-client**:
>   - Await `pluginManager.executeHook('onResponseReceived')` in `packages/api-client/src/libs/send-request/create-request-operation.ts`.
> - **Tests** (`create-request-operation.test.ts`):
>   - Refactor to use `vi.spyOn(global, 'fetch')` with `afterEach` reset; remove per-test try/finally restores.
>   - Expand/adjust tests:
>     - Empty body (204) response handling and safe normalization.
>     - Response streaming using `ReadableStream` and reader verification.
>     - `onBeforeRequest` execution and ordering.
>     - `onResponseReceived` execution, awaiting behavior, and header mutation; updated call count expectations.
> - **Changeset**: Add patch changeset noting the fix for awaiting `onResponseReceived`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 85a7c4c3b86b13787b2f5ced2be48fff702584f1. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/api-client/src/libs/send-request/create-request-operation.test.ts
+++ b/packages/api-client/src/libs/send-request/create-request-operation.test.ts
@@ -9,7 +9,7 @@ import {
   serverSchema,
 } from '@scalar/oas-utils/entities/spec'
 import { encode } from 'js-base64'
-import { beforeAll, describe, expect, it, vi } from 'vitest'
+import { afterEach, beforeAll, describe, expect, it, vi } from 'vitest'
 import type { z } from 'zod'
 
 import * as electron from '../electron'
@@ -92,6 +92,11 @@ Try to run it like this:
 $ pnpm dev:void-server
 `)
   }
+})
+
+const globalFetchSpy = vi.spyOn(global, 'fetch')
+afterEach(() => {
+  globalFetchSpy.mockReset()
 })
 
 describe('create-request-operation', () => {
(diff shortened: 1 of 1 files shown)


packages/api-client/src/libs/send-request/create-request-operation.ts
```diff
@@ -238,7 +238,7 @@ export const createRequestOperation = ({
         })
 
         if (pluginManager) {
-          pluginManager.executeHook('onResponseReceived', { response: normalizedResponse, operation: request })
+          await pluginManager.executeHook('onResponseReceived', { response: normalizedResponse, operation: request })
```
marcalexiei, line 241: To help prevent floating promises, 
I’d like to suggest enabling the [`noFloatingPromises`](https://biomejs.dev/linter/rules/no-floating-promises/) rule.

Although it’s still part of the nursery group, 
it could help catch issues similar to #4047.

I ran a quick check, and there are currently 123 violations related to this rule.  
In most cases, these can likely be resolved by either awaiting or explicitly voiding the promises.

Let me know if this is something you’d like to enable,
I’d be happy to open a PR for it!

packages/api-client/src/libs/send-request/create-request-operation.test.ts
```diff
@@ -828,49 +833,44 @@ describe('create-request-operation', () => {
...
+      headers: new Headers(),
+      body: mockEmptyStream,
+      ok: true,
+      statusText: 'No Content',
+      // Add the methods your code uses
+      clone: () => mockResponse,
+      text: async () => '',
+      json: async () => ({}),
+      arrayBuffer: async () => new ArrayBuffer(0),
+    } as Partial<Response>
```
marclave, line 853: @amritk @geoffgscott @DemonHa i like that we use a partial vs cast to unknown at first, is there a reason why we did this? looks like below we need to cast it as a Response anyways tho @marcalexiei ?

packages/api-client/src/libs/send-request/create-request-operation.ts
marclave: i dig this, wdyt @geoffgscott @amritk @hanspagel @DemonHa ?

packages/api-client/src/libs/send-request/create-request-operation.test.ts
marcalexiei: > looks like below we need to cast it as a Response

I think performing the cast when providing it as a `Response` is better,
 as it makes it explicit that `mockResponse` isn’t a complete `Response` at runtime.

It’s a minor detail given we’re only dealing with a few lines of code, 
but it does make the intent a bit clearer 😅

packages/api-client/src/libs/send-request/create-request-operation.ts
hanspagel: Yes, pls! Enable the rule was a warning, and if you want to fix a few of those, that would be amazing!

packages/api-client/src/libs/send-request/create-request-operation.test.ts
hanspagel: I agree

Issue #6817, opened 2025-09-11
Form data which is binary should not default to dummy filename


ashish-b-choudhary-db:
### What happens?

For binary form data, the try it out feature puts defaults

<img width="700" height="172" alt="Image" src="https://github.com/user-attachments/assets/7c5f50bc-2356-4768-a407-5b02c9528076" />

To upload a file when you hover over filename, delete option pops up and after doing that you can upload a file

<img width="700" height="172" alt="Image" src="https://github.com/user-attachments/assets/6175a310-165d-4246-8f55-6fc7f92535c8" />

### What did you expect to happen?

Upload File button should be active by default to avoid hassle of having to delete the default first. Eg. swagger -

<img width="700" height="207" alt="Image" src="https://github.com/user-attachments/assets/705e3d99-b392-441f-84fc-1f510e12ae80" />

### OpenAPI Document

https://sandbox.scalar.com/p/hh081

OdujokoD:
Hey @hanspagel I can pick this up if no one isn't working on it yet.

ajay9415:
the problem still not fixed.

PR #7069, opened 2025-10-13
fix(api-client): remove form data binary default file

OdujokoD:
**Problem**
Currently, when a `string` parameter of the request body for a Form URL Encoded has a `binary` format, a default file is returned requiring the user to delete that file before uploading another as described in #6817 

**Solution**

With this PR, the default file has been removed allowing a direct file upload
Screenshot 2025-10-13 at 8 21 13 AM
s now been removed allowing a direct upload of a file

**Checklist**


OdujokoD:
Hey @hanspagel @amritk , I'd love feedback on this before switching to a completed PR, particularly around the **TODO** comments I left:
- The `handleFileUploadFormData` in the `RequestBody` component maintains the initial value (or example) set in the spec and doesn't change to the uploaded file name. Should this be fixed in this PR or raised as a separate issue?
- Should the check for the binary format be totally removed since the check is currently redundant or is there something else being consider for binary formats?

ajay9415:
the problem still not fixed.

PR #7092, opened 2025-10-14
feat(openapi-parser): adds OpenAPI 3.2.0 parsing support

baywet:
related #7039

- **tests: adds a test for 3.2.0 dereferencing**
- **chore: linting**
- **test: adds a test for media type reference resolution**
- **feat: adds function overload to match the return type to the upgrade method**
- **chore: updates upgrade tests to use 3.2 types**
- **tests: adds unit tests for xmlObject upgrade**
- **chore: refactors tags migration to its own method**
- **feat: adds support for migrating xmlNodes wrapped and attribute properties to 3.2.0**
- **tests: adds unit tests for query and additional operations parsing**

**Problem**

Currently, OpenAPI 3.2.0 parsing is not supported by the parser package. Or at least we don't have unit tests to confirm it does.

**Solution**

With this PR parsing of OpenAPI 3.2.0 documents is tested to work. This PR also adds migration of xmlNodeType to support the latest specification and avoid using deprecated fields. Since the internal use JSON.parse or equivalent for YAML, testing the other new fields of that version shouldn't be necessary.

**Notes**

This does not update the reference resolution parsing using $self. We've had zero demand for that (and $id) in OpenAPI.net so I figured the situation would be similar here. It can also be implemented at a later time if there's demand for it.

**Checklist**

---

> [!NOTE]
> Adds OpenAPI 3.2.0 schema/validation/dereferencing support and a 3.1→3.2 upgrader with XML and tag-group migrations, plus type updates and tests.
> 
> - **OpenAPI Parser**:
>   - Add 3.2.0 schema (`schemas/v3.2/schema.{json,ts}`) and register in `configuration` (`OpenApiSpecifications`, `OpenApiVersions`).
>   - Extend `Validator` to support `3.2` and apply `media-range` format for `3.1` and `3.2`.
>   - Tests: validation and dereferencing for `3.2.0`; media type reference resolution; query and `additionalOperations` parsing.
>   - Minor: improve YAML parse fallback; fix typo in error deduplication.
> - **Upgrader (3.1 → 3.2)**:
>   - Migrate `x-tagGroups` to tag `kind` (refactored into helper).
>   - Migrate `xml` properties: `wrapped` → `nodeType: 'element'`, `attribute` → `nodeType: 'attribute'`; error if both true.
>   - Update `upgrade` API with overloads and return types; switch to `OpenAPIV3_2` in tests.
> - **Types**:
>   - Re-export `OpenAPIV3_2` in `@scalar/types`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit d0f566305daa29a9b8904e5c0532e12b1e1c11f1. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/openapi-parser/src/lib/Validator/Validator.ts
+++ b/packages/openapi-parser/src/lib/Validator/Validator.ts
@@ -150,13 +150,8 @@ export class Validator {
     // https://ajv.js.org/packages/ajv-formats.html#formats
     addFormats(ajv)
 
-    // OpenAPI 3.1 uses media-range format
-    if (version === '3.1') {
-      ajv.addFormat('media-range', true)
-    }
-
-    // OpenAPI 3.2 uses media-range format
-    if (version === '3.2') {
+    // OpenAPI 3.1 and 3.2 uses media-range format
+    if (version === '3.1' || version === '3.2') {
       ajv.addFormat('media-range', true)
     }
 


marclave:
AWESOME work @baywet !! I think we merged a pretty big PR before this so there's a lot of conflicts unfortunately.

do we want to auto upgrade to 3.2? i think so eh @geoffgscott @hanspagel @amritk

hanspagel:
OMG again: amazing work

And sorry for all the merge conflicts, I didn’t expect you to submit this PR so quickly. 😬 

Ideally, this PR here would be focused on just `@scalar/openapi-parser` and leave the other packages for now. Not sure if that's feasible. That would also help with the conflicts I guess?

And eventually, we want to auto-update to OpenAPI 3.2 very soon, but ideally not in this PR here.

baywet:
Thank you for the additional information.

The reason why I "had to" go out of the parser package is because the front end components would not build anymore after the changes. I'm more than happy to try to leave it out if I can, or I could leave this PR in a state that doesn't build and you take it from here? let me know.

hanspagel:
Yeah, I’ll just take it from here.

packages/openapi-upgrader/src/upgrade.ts
```diff
@@ -1,15 +1,21 @@
...
+import type { UnknownObject } from '@scalar/types/utils'
 
 /**
- * Upgrade OpenAPI documents from Swagger 2.0 or OpenAPI 3.0 to OpenAPI 3.1.0
+ * Upgrade OpenAPI documents from Swagger 2.0 or OpenAPI 3.0 to the specified target version
  */
-export function upgrade(value: UnknownObject, targetVersion: '3.0' | '3.1' | '3.2') {
+export function upgrade(value: UnknownObject, targetVersion: '3.0'): OpenAPIV3.Document
+export function upgrade(value: UnknownObject, targetVersion: '3.1'): OpenAPIV3_1.Document
+export function upgrade(value: UnknownObject, targetVersion: '3.2'): OpenAPIV3_2.Document
```
hanspagel, line 14: nice

PR #7136, opened 2025-10-20
feat(workspace-store): persistence layer for client store

### LINKED ISSUE #3217: Using a big spec in the config causes a quota error
### What happens?

Uploading a collection via URL works fine. But when you load up a big spec through config.spec.url it throws some sort of quota error. Possibly related to the useDebounceFn from vueuse

image

no errors

### OpenAPI Document

_No response_

DemonHa:
This PR introduces an IndexedDB-backed persistence layer for the workspace store.
The new layer provides a functional interface for performing common operations such as get, set, and general querying of stored workspace data.

By using IndexedDB, the workspace store now supports persistent client-side storage with better performance for larger datasets.

Fixes #3217

**Checklist**

---

> [!NOTE]
> Introduces an IndexedDB-backed persistence layer with a client plugin that debounces and persists workspace changes, while refactoring workspace APIs and adding change-detection proxies.
> 
> - **Persistence (IndexedDB)**:
>   - Add `src/persistence/index.ts` and `src/persistence/indexdb.ts` with tables for `workspace`, `meta`, `documents`, `originalDocuments`, `intermediateDocuments`, `overrides`, `documentMeta`, `documentConfigs`.
>   - New export `@scalar/workspace-store/persistence`; add `fake-indexeddb` for tests.
> - **Client Plugins & Auto-persist**:
>   - Add plugin system (`src/workspace-plugin.ts`) and `plugins/client/persistence` to debounce writes and persist workspace state.
>   - Emit workspace state change events via `createDetectChangesProxy` and hook into `client.ts`.
> - **Store API changes**:
>   - `exportWorkspace()` now returns an object (`InMemoryWorkspace`); `loadWorkspace()` accepts object.
>   - Fire change events for `documents`, `originalDocuments`, `intermediateDocuments`, `documentConfigs`, `overrides`, `documentMeta`, and `meta`.
> - **Helpers**:
>   - New `debounce`, `detect-changes-proxy`, `unpack-proxy`; enhance `overrides-proxy` API/signature.
> - **Bundler & Schemas**:
>   - Move bundler plugins import to `@/plugins/bundler`.
>   - Update `InMemoryWorkspace`/schema shapes; minor `reference-config` typing fix.
> - **Build/Exports**:
>   - Include persistence and client plugins in `esbuild` entries and `package.json` exports.
> - **Tests**:
>   - Extensive tests for persistence, indexdb wrapper, debounce, change-detection proxy, and client API updates.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 41846a67b3df5530405585843bc7bc23676d38d7. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/workspace-store/src/persistence/indexdb.ts
+++ b/packages/workspace-store/src/persistence/indexdb.ts
@@ -6,33 +6,54 @@ type TableEntry<S extends TObject, K extends readonly (keyof Static<S>)[]> = {
 }
 
 /**
- * Creates a connection to an IndexedDB database for managing tables and persistence.
+ * Initializes and manages an IndexedDB database connection for table-based persistence.
  *
- * @param dbName - Optional database name (default 'scalar-workspace-store')
- * @returns An object with `createTable` and `closeDatabase` functions.
+ * @param name - The database name. Defaults to 'scalar-workspace-store'.
+ * @param tables - Table definitions: the tables to create and their key schemas.
+ * @param version - The database version. Bump this to trigger upgrades (default: 1).
+ * @param migrations - Optional migration steps to run for version upgrades.
+ * @returns An object with the following methods:
+ *   - `get(tableName)` — Get a wrapper to interact with the object store for the given table name.
+ *   - `closeDatabase()` — Closes the database connection.
  *
- * Example:
+ * Example usage:
  * ```ts
  * import { Type } from '@scalar/typebox'
  * import { createIndexDbConnection } from './indexdb'
  *
- * // Define a schema
+ * // Define a schema for a user
  * const UserSchema = Type.Object({
  *   id: Type.String(),
  *   name: Type.String(),
  *   age: Type.Number(),
  * })
  *
- * // Create the database connection
- * const { createTable, closeDatabase } = createIndexDbConnection()
+ * // Define tables in the database
+ * const dbConfig = {
+ *   users: {
+ *     schema: UserSchema,
+ *     index: ['id'] as const,
+ *   },
+ * }
  *
- * // Create (or open) a "users" table with "id" as key
- * const usersTable = await createTable('users', {
- *   schema: UserSchema,
- *   key: ['id'],
+ * // Open the database connection and get table API
+ * const { get, closeDatabase } = await createIndexDbConnection({
+ *   name: 'my-app-db',
+ *   tables: dbConfig,
+ *   version: 1,
  * })
  *
- * // Use usersTable methods for CRUD (see createTableWrapper)
+ * // Get a strongly-typed users table API
+ * const usersTable = get('users')
+ *
+ * // Add a user
+ * await usersTable.addItem({ id: 'user-1' }, { name: 'Alice', age: 25 })
+ *
+ * // Retrieve a user by id
+ * const user = await usersTable.getItem({ id: 'user-1' })
+ *
+ * // Don't forget to close the database when done!
+ * closeDatabase()
  * ```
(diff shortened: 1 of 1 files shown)


amritk (changes requested):
Hmm reactivity doesn't seem to work for me again. Maybe we should collapse all of our proxies into one proxy with plugins?

DemonHa:
> Hmm reactivity doesn't seem to work for me again. Maybe we should collapse all of our proxies into one proxy with plugins?

Reactivity seams to be working fine for me, maybe you didn't install/build properly.

I really love that idea but Idk how realistic it is. I will try to play around with it and see what we can do. We are clearly using way to many proxies 😁.

The other thing is that the magic proxy needs to way the document level because it needs the root of the document to resolve refs against. So even if we create a "pluginable proxy system" we would still need to use multiple instances to wrap our stuff.

amritk:
Yea even if we can't do a plugin system, we can at least combine into one proxy? because the multi proxies is messing up vue

DemonHa:
@amritk @geoffgscott Should we version our indexdb tables?

packages/workspace-store/src/helpers/detect-changes-proxy.ts
```diff
@@ -0,0 +1,124 @@
+import { isObject } from '@/helpers/general'
+
+const isDetectChangesProxy = Symbol('isDetectChangesProxy')
+const detectChangesProxyTarget = Symbol('detectChangesProxyTarget')
+
+const isPropsVue = (prop: string | Symbol) => {
+  return (
+    typeof prop === 'symbol' ||
```
amritk, line 8: still don't like this, but if its working

packages/workspace-store/src/helpers/unpack-proxy.ts
```diff
@@ -0,0 +1,13 @@
...
+import { unpackDetectChangesProxy } from '@/helpers/detect-changes-proxy'
+import { unpackOverridesProxy } from '@/helpers/overrides-proxy'
+
+/**
+ * Unpacks special vue reactivity & override & detect-changes & magic proxy from an input object or array,
+ * returning the "raw" plain object or array.
+ */
+export const unpackProxyObject = <T extends object | Array<unknown>>(input: T): T => {
+  return unpackDetectChangesProxy(toRaw(getRaw(unpackOverridesProxy(input))))
+}
```
amritk:
```suggestion
export const unpackProxyObject = <T extends object | Array<unknown>>(input: T): T => 
  unpackDetectChangesProxy(toRaw(getRaw(unpackOverridesProxy(input))))
```
unnecessary return

packages/workspace-store/package.json
```diff
@@ -93,6 +103,7 @@
     "@scalar/themes": "workspace:*",
     "@scalar/typebox": "catalog:*",
     "@scalar/types": "workspace:*",
+    "fake-indexeddb": "^6.2.3",
```
amritk:
```suggestion
```
should be in devDependencies as its only used for testing

amritk (approved):
Move the dep and its good to go! Nice work.

I drew a picture of all of our proxies

proxy

packages/workspace-store/src/helpers/detect-changes-proxy.ts
DemonHa: I don't like it either

PR #7146, opened 2025-10-21
feat(api-client): block requests with empty path parameters

### LINKED ISSUE #6348: Bug in UI when required GET parameter is not filled

hanspagel:
New attempt of #6958 by @OdujokoD

**Problem**

Requests with empty path parameters (e.g. `/posts/{postId}`) just hit `/posts/`, which is a different endpoint.

**Solution**

With this PR, we validate the path parameters, block requests with empty parameters and show a toast to inform the users.

I’ve also renamed "Variables" to "Path Parameters" (that’s how it’s called in the OpenAPI specification).

Fixes #6348

**Checklist**

---

> [!NOTE]
> Block sending requests when required path parameters are empty, returning structured validation results and updating UI to "Path Parameters."
> 
> - **Validation/Libs (`validate-parameters`)**:
>   - Return `ValidationResult` with `invalidParams` and `hasBlockingErrors`.
>   - Treat empty required `path` params as blocking; `query/headers/cookies` remain non-blocking but flagged.
>   - Add/adjust tests to cover blocking and non-blocking cases.
> - **Request Execution (`RequestRoot.vue`)**:
>   - Compute validation from active example; block `executeRequest` on `hasBlockingErrors`, show toast, emit abort.
>   - Pass `validation.invalidParams` to child via `RouterView`; remove old invalid param state/watch.
> - **UI (`RequestSection.vue`)**:
>   - Rename "Variables" to "Path Parameters" in filters, ids, and section titles.
> - **Meta**:
>   - Add changeset for `@scalar/api-client` patch release.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 89d53983eda8519a27a4a5a03e0c8f54ab57e007. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/api-client/src/libs/validate-parameters.ts
+++ b/packages/api-client/src/libs/validate-parameters.ts
@@ -13,18 +13,20 @@ export type ValidationResult = {
  * should prevent the request from being sent.
  */
 export const validateParameters = (example: Partial<RequestExample> | null): ValidationResult => {
-  const invalidParams = new Set<string>()
-  let hasBlockingErrors = false
+  const result = {
+    invalidParams: new Set<string>(),
+    hasBlockingErrors: false,
+  }
 
   if (!example) {
-    return { invalidParams, hasBlockingErrors }
+    return result
   }
 
   // Check path parameters separately - these are blocking
   example.parameters?.path?.forEach((param) => {
     if (param.enabled && param.value.trim() === '') {
-      invalidParams.add(param.key)
-      hasBlockingErrors = true
+      result.invalidParams.add(param.key)
+      result.hasBlockingErrors = true
     }
   })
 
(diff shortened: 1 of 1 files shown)


packages/api-client/src/views/Request/RequestRoot.vue
```diff
@@ -46,6 +46,7 @@ const element = ref<HTMLDivElement>()
 
 const requestAbortController = ref<AbortController>()
 const invalidParams = ref<Set<string>>(new Set())
+const hasBlockingErrors = ref(false)
```
amritk: Feels like this should be a computed instead of ref + watcher

hanspagel: you’re not wrong

hwkr: Updated!

hwkr:
Updated to use a computed property :)

PR #7091, opened 2025-10-14
refactor: enable `noFloatingPromises`

marcalexiei:
**Problem**

- https://github.com/scalar/scalar/pull/7060#discussion_r2420757796

**Solution**

This PR enables the `noFloatingPromises` rule with severity set to `warn`.

Currently, on the `main` branch, there are **114** issues reported by this rule.  
All warnings in JavaScript and TypeScript files have been fixed in this PR.  
The remaining **82** issues are located in `*.vue` files.

Most of these are related to the Vue router’s `push` method.

According to the [Vue documentation](https://router.vuejs.org/guide/advanced/navigation-failures), 
it’s recommended to handle the navigation result explicitly.  
Since I’m not very familiar with Vue, I’ve chosen to leave these cases as they are for now.

**Checklist**

I've gone through the following:

- [x] I've added an explanation _why_ this change is needed.
- [x] I've added a changeset (`pnpm changeset`).
- [x] I've added tests for the regression or new feature.
- [x] I've updated the documentation (Not needed).

---

> [!NOTE]
> Enable Biome `noFloatingPromises` (warn) and update code/tests to properly await or explicitly discard async calls across packages and examples.
> 
> - **Config**:
>   - Enable `linter.nursery.noFloatingPromises: "warn"` in `biome.json`.
> - **Examples/Playgrounds**:
>   - Use `void` on startup calls in `examples/nestjs/*/src/main.ts`, `projects/client-scalar-com/src/main.ts`, and `packages/api-client/playground/*/main.ts`.
>   - Await build in `packages/mock-server/playground/esbuild.ts` and use `void` for `main()` in mock-server playground.
> - **Packages**:
>   - `api-client-react`: prefix async effects with `void` in `ApiClientModalProvider`.
>   - OAuth helpers (`packages/api-client/src/v2/.../oauth.ts`, `packages/api-client/src/views/Request/libs/oauth2.ts`): ensure promise handling; add comments to ignore lints where promises intentionally float.
>   - Build tooling: prefix long-running `runCommand` calls with `void`.
> - **Tests/Integrations**:
>   - Convert Promise wrappers to `async/await` and await async APIs in tests: Fastify export, ASP.NET Core static assets, API client (App/Web/Modal), client creation, OAuth tests, OpenAPI parser, workspace-store, mock-server.
>   - Replace ad-hoc globals with spies (e.g., `fetch`) and improve setup/teardown in tests.
> 
[...]

PR #7147, opened 2025-10-21
fix(openapi-parser): make `validate` synchronous

marcalexiei:
**Problem**

[Another change related to the `useAwait` rule](https://github.com/scalar/scalar/pull/7091#discussion_r2429906517):

`Validator.getAjvValidator` is actually synchronous. 
Making it synchronous will remove the `await` requiremen from `Validator.validate` and `validate` (function exported by `@scalar/openapi-parser`)

> [!CAUTION]
> Even if this is actually is a breaking,
> I marked it as a minor change since the package has a `0.x` version.

**Solution**

Make all the above mentioned  function chain synchronous.

**Checklist**

- [x] I've updated the documentation (Not needed, I haven't found mentation on this function in the documentation).

---

> [!NOTE]
> Make `validate` synchronous across the parser, updating the validator, utilities, queue workflow, and tests accordingly.
> 
> - **openapi-parser**
>   - **Validator** (`packages/openapi-parser/src/lib/Validator/Validator.ts`):
>     - Make `validate` and `getAjvValidator` synchronous; remove `async/await` usage.
>     - Type `ajvValidators` as `Record<string, ValidateFunction>` and cast Ajv instance.
>     - Throw error using `errors[0].message`; only aggregate reference resolution errors in result.
>   - **Utils**:
>     - `validate` (`src/utils/validate.ts`): now synchronous; calls `new Validator().validate` directly.
>     - `workThroughQueue` (`utils/openapi/utils/workThroughQueue.ts`): call synchronous `validate` in command chain.
>   - **Tests**: Remove `async/await` from validation calls and adjust expectations across test suites to match synchronous API.
>   - **Changeset**: Minor version bump for `@scalar/openapi-parser`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 8b6e0fa82c66485c20d97209536c4d742ba3953b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/openapi-parser/src/utils/validate.ts
+++ b/packages/openapi-parser/src/utils/validate.ts
@@ -2,6 +2,7 @@ import type { OpenAPI } from '@scalar/openapi-types'
 
 import { Validator } from '@/lib/Validator/Validator'
 import type { AnyObject, Filesystem, ThrowOnErrorOption, ValidateResult } from '@/types/index'
+
 import { makeFilesystem } from './make-filesystem'
 
 export type ValidateOptions = ThrowOnErrorOption
@@ -9,14 +10,11 @@ export type ValidateOptions = ThrowOnErrorOption
 /**
  * Validates an OpenAPI document
  */
-export async function validate(
-  value: string | AnyObject | Filesystem,
-  options?: ValidateOptions,
-): Promise<ValidateResult> {
+export function validate(value: string | AnyObject | Filesystem, options?: ValidateOptions): ValidateResult {
   const filesystem = makeFilesystem(value)
 
   const validator = new Validator()
-  const result = await validator.validate(filesystem, options)
+  const result = validator.validate(filesystem, options)
 
   return {
     ...result,
--- a/.changeset/six-knives-remain.md
+++ b/.changeset/six-knives-remain.md
@@ -1,0 +1,5 @@
+---
+'@scalar/openapi-parser': minor
+---
+
+fix(openapi-parser): make`validate` synchronous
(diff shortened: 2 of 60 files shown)


packages/openapi-parser/src/utils/validate.ts
```diff
@@ -2,21 +2,19 @@ import type { OpenAPI } from '@scalar/openapi-types'
...
 export type ValidateOptions = ThrowOnErrorOption
 
 /**
  * Validates an OpenAPI document
  */
-export async function validate(
-  value: string | AnyObject | Filesystem,
-  options?: ValidateOptions,
-): Promise<ValidateResult> {
+export function validate(value: string | AnyObject | Filesystem, options?: ValidateOptions): ValidateResult {
```
marcalexiei, line 13: To avoid changing the public API we could consider to wrap the return value into a `Promise.resolve`. 
This would also help prevent potential breaking changes if asynchronous logic is reintroduced in the future.
WDYT?

marcalexiei:
Hi @hanspagel,  
~~Just an FYI: 
this is the final user-facing code change that needs to be merged before opening the PR to enable the [`useAwait` rule](https://github.com/scalar/scalar/pull/7091#discussion_r2429906517) across the entire repo.~~

~~I already have that branch ready, so I’ll open the PR right after this is merged.~~

---

UPDATE: I found another change worth merging before opening the useAwait PR: #7174

geoffgscott:
These were sync to allow chaining of the methods I believe. @hanspagel do we want to change the API?

marcalexiei:
No longer needed after #7196

PR #7152, opened 2025-10-21
refactor(sidebar): remove some divs from sidebar elements

hwkr:
Reduces the number of divs in sidebar elements somewhat.

**Checklist**

### DIFF
--- a/.changeset/stupid-ways-fix.md
+++ b/.changeset/stupid-ways-fix.md
@@ -1,0 +1,6 @@
+---
+'@scalar/components': patch
+'@scalar/sidebar': patch
+---
+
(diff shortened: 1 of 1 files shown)


packages/sidebar/src/components/SidebarItem.vue
```diff
@@ -178,15 +178,13 @@ const handleDragEnd = (
       class="text-left"
       :selected="selectedItems[item.id] ?? false"
       @click="() => emits('click', item.id)">
-      <div class="group/entry flex flex-1 items-center justify-center">
```
hwkr, line 180: This change might require a `self-end` on whatever is put in the aside slot

PR #7162, opened 2025-10-24
test(api-reference): add visual regression testing

hwkr:
The beginnings of adding visual regression testing for the references. Fairly minimal right now but gets it working so we can expand on it in the future.

Tests the following (in both classic and modern layouts):

- Request Body
- Request Query params
- Responses
- Callbacks
- Request examples 
- Models

Things it doesn't test (yet):

- Other specs besides Scalar Galaxy (the tests need fairly specific targeting so we'll need to tweak things to get this working, also has the potential to generate _a lot_ of screenshots).
- API Client Modal

Other cool things / changes:

- Tweaked a bunch of a11y stuff and regions to make test targeting work better
- The snapshots use the same test config as the other e2e tests so now we get fancy comments with test info for all references e2e tests (including a screenshot if they fail).

**Checklist**

- [x] I've added tests for the regression or new feature. ← So many


packages/api-reference/src/components/Content/Models/ModelTag.vue
```diff
@@ -34,14 +29,10 @@ const headerId = useId()
       v-if="options.layout === 'modern'"
       id="models">
       <Section
-        :aria-labelledby="headerId"
+        aria-label="Models"
```
hwkr, line 31: We don't really need to use an id here since it's always the same label. Slightly fewer dom attribute this way

packages/api-reference/src/components/Section/CompactSection.vue
```diff
@@ -18,7 +18,9 @@ const emit = defineEmits<{
 }>()
 </script>
 <template>
-  <div class="collapsible-section">
+  <section
+    :aria-label="label"
```
hwkr, line 21: Fun fact: `<section>` tags will be given the `region` role only if they have a accessible label.

packages/components/src/components/ScalarCard/ScalarCard.vue
```diff
@@ -46,13 +46,14 @@ defineOptions({ inheritAttrs: false })
 const { cx } = useBindCx()
 </script>
 <template>
-  <section
```
hwkr, line 49: This was my bad making all cards a `region`, in hindsight that's a bit overkill. Group makes more sense.

packages/api-reference/test/snapshots/reference/scalar-galaxy-classic-model.png
geoffgscott, line 1: Please rename the folder `fixtures` or `snapshots` or `images`. `fixtures` is best IMO

hwkr: Yeah I had it called `snapshots` but then it was `snapshots/snapshots`. I'm going to redo the folders to be per test file if that's cool so it should make more sense. You'll have `operation.e2e.ts` and `operation.snapshots` or something like that.

packages/api-reference/test/utils/is-classic.ts
geoffgscott, line 1: kebab case filenames please.

hwkr: Oh right, gets me everytime

hwkr: Fixed

PR #7166, opened 2025-10-24
feat: no cast bundle

### LINKED ISSUE #7150: OpenAPI - bundle method should have a return type that's better than "object"
https://github.com/scalar/scalar/blob/1fcc8d8a773634345ef7a6190ccdb991c50b2710/packages/json-magic/src/bundle/bundle.ts#L617

Is only "documenting" returning object. Is should instead document returning OpenAPI.Document, or provide a generic type parameter and return that so consumers of the method do not have to rely on obscure casts down the line.

Happy to submit a pull request if we have some kind of agreement on the direction to take here.

baywet:
fixes #7150

---

> [!NOTE]
> Add generic type parameters to `bundle` and `dereference` for strongly-typed results, with docs updated accordingly.
> 
> - **API (typed results)**:
>   - `bundle<T>` now generic; returns `Promise<BundleResult<T>>` with `x-ext`/`x-ext-urls` support.
>   - `dereference<Opts, T>` now generic; returns typed `DereferenceResult<T>` (sync or async), and uses typed `createMagicProxy<T, T>`.
>   - Minor internal typing refinements (casted `x-ext-urls` mappings).
> - **Docs**:
>   - README examples updated to demonstrate generic usage (e.g., `bundle<Document>(...)`).
> - **Release**:
>   - Changeset added for `@scalar/json-magic` patch release.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 6afff3e02fb59d92abb935f2ff8fbef8b49a0330. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/json-magic/src/bundle/bundle.ts
+++ b/packages/json-magic/src/bundle/bundle.ts
@@ -549,6 +549,11 @@ const extensions = {
   externalDocumentsMappings: 'x-ext-urls',
 } as const
 
+type BundleResult<T extends UnknownObject> = T & {
+  'x-ext'?: Record<string, unknown>
+  'x-ext-urls'?: Record<string, string>
+}
+
 /**
  * Bundles an OpenAPI specification by resolving all external references.
  * This function traverses the input object recursively and embeds external $ref
@@ -615,7 +620,10 @@ const extensions = {
  * // The function will first fetch the OpenAPI spec from the URL,
  * // then bundle all its external references into the x-ext section
  */
-export async function bundle<T extends UnknownObject = UnknownObject>(input: T | string, config: Config): Promise<T> {
+export async function bundle<T extends UnknownObject = UnknownObject>(
+  input: T | string,
+  config: Config,
+): Promise<BundleResult<T>> {
   // Cache for storing promises of resolved external references (URLs and local files)
   // to avoid duplicate fetches/reads of the same resource
   const cache = config.cache ?? new Map<string, Promise<ResolveResult>>()
@@ -912,5 +920,5 @@ export async function bundle<T extends UnknownObject = UnknownObject>(input: T |
     delete documentRoot[extensions.externalDocumentsMappings]
   }
 
-  return rawSpecification as T
+  return rawSpecification as BundleResult<T>
 }


baywet:
@hanspagel here I believe Cursor is correct, Essentially Config and ResolveResult should also be made generic. And I think because things were not "typed" before, it was masking conversion issues. But I didn't want to go too far down the rabbit hole without additional context.

packages/json-magic/src/bundle/bundle.ts
```diff
@@ -614,7 +615,7 @@ const extensions = {
  * // The function will first fetch the OpenAPI spec from the URL,
  * // then bundle all its external references into the x-ext section
  */
-export async function bundle(input: UnknownObject | string, config: Config) {
+export async function bundle<T extends UnknownObject = UnknownObject>(input: T | string, config: Config): Promise<T> {
```
DemonHa: the return type here should be `Promise<T & { 'x-ext': Record<string, unknown>  }>` but it can also contain `{ 'x-ext-urls': Record<string, string> }` if `urlMap` is set to true

packages/json-magic/src/dereference/dereference.ts
```diff
@@ -3,19 +3,20 @@ import { fetchUrls } from '@/bundle/plugins/fetch-urls'
...
       errors: string[]
     }
 
-type ReturnDereferenceResult<Opt extends { sync?: boolean }> = Opt['sync'] extends true
-  ? DereferenceResult
-  : Promise<DereferenceResult>
+type ReturnDereferenceResult<
+  Opt extends { sync?: boolean },
+  T extends UnknownObject = UnknownObject,
+> = Opt['sync'] extends true ? DereferenceResult<T> : Promise<DereferenceResult<T>>
```
DemonHa, line 19: for the DereferenceResult the return type is not just T but for every `$ref` encountered, a `$ref-value` field is added containing the resolved content.

packages/json-magic/src/bundle/bundle.ts
baywet: defined a type to cover that

packages/json-magic/src/dereference/dereference.ts
baywet: I'm not sure I understand what you mean here? Are you suggesting defining the result with `& Record<string, DereferenceResult<T>` ? or something else?

geoffgscott:
Hi @baywet, appreciate the PR. 

In this case using a generic incorrectly typecasts the input without guaranteeing compliance with the the type definition. We can infer the input type but it will then be extended in unpredictable ways based on the bundling and so the output type will still be effectively just `object`. 

If you want to avoid the typecast a validator should be run after bundling. Alternatively if you feel safe in the actual output type of your JSON bundle then a cast may be appropriate. 

Will close this for now.

baywet:
@geoffgscott closing the PR like this is a bit abrupt :/
I do think the API surface here is NOT providing a great developer experience. Essentially it all accepts strings or random objects, and returns random objects.
It's then up to the consumer of the API to figure out what's being returned, what to do with it, etc. And the public documentation is full of gaps at the moment, which does not help.
Maybe the generic type parameter is not the right solution to this issue, I understand that "returning a type without being able to guarantee the value matches that type" might lead to further confusion. But the API should be able to do better than "here is an object, now you figure out what to do with that". By doing what it's currently doing, the consumer loses intellisense, auto-completion, transpile time validation, and much more! (I'm pretty sure the implementing code also suffers from that, and contains bugs because of ti, C.F. my earlier comment)

hanspagel:
I think we should just show how to validate the result using the typebox schemas and get the correct type, so I agree with @geoffgscott.

PR #7121, opened 2025-10-17
feat(openapi-parser): make `dereference` sync

marcalexiei:
**Problem**

[Another change related to the `useAwait` rule](https://github.com/scalar/scalar/pull/7091#discussion_r2429906517):

The `dereference` function was incorrectly marked as `async`, even though it was entirely synchronous.

**Solution**

I renamed `dereference` to `dereferenceAsync` for consistency with other paired methods such as `measureSync` and `measureAsync`.
The async version has been commented out for now, in case asynchronous logic is needed again in the future.

All existing usages of `dereference` have been updated to use `dereferenceSync`.

These changes are limited in scope, as this method was only used within functions that already involve other asynchronous logic.

> [!WARNING]
> Since all affected packages are still in the `0.x` version range, this update has been released as a **minor** change.

**Checklist**

I've gone through the following:

- [x] I've added an explanation _why_ this change is needed.
- [x] I've added a changeset (`pnpm changeset`).
- [x] I've added tests for the regression or new feature.
- [x] I've updated the documentation (not needed).

---

> [!NOTE]
> Convert `dereference` to a synchronous API and propagate changes across parser, mock-server, oas-utils, api-reference, and openapi-to-markdown.
> 
> - **Parser (`@scalar/openapi-parser`)**:
>   - Make `dereference` synchronous; remove async usage and `dereferenceSync` export; update barrel exports.
>   - Update tests to non-async style; adjust `workThroughQueue` to use sync `dereference` and tidy types.
> - **Mock Server (`@scalar/mock-server`)**:
>   - `createMockServer` is now synchronous; use sync `dereference`; update tests to drop `await`.
> - **OAS Utils (`@scalar/oas-utils`)**:
>   - Use sync `dereference` in `transforms/import-spec.ts`.
> - **API Reference (`@scalar/api-reference`)**:
>   - Replace `dereferenceSync` usage with `dereference`; adjust reactive type.
> - **OpenAPI to Markdown (`@scalar/openapi-to-markdown`)**:
>   - Use sync `dereference` in HTML/Markdown generation.
> - **Changeset**:
>   - Minor version bumps for `@scalar/openapi-to-markdown`, `@scalar/openapi-parser`, `@scalar/mock-server`, `@scalar/oas-utils`.
> 
[...]

PR #7174, opened 2025-10-27
feat(mock-server): convert `createMockServer` and all dependent references to synchronous implementation

marcalexiei:
**Problem**

[Another change related to the `useAwait` rule](https://github.com/scalar/scalar/pull/7091#discussion_r2429906517):

`createMockServer` has no asynchronous logic but most of its usage have an `await`.

**Solution**

Add an explicit return type to `createMockServer` and update all its usage and references

**Checklist**

---

> [!NOTE]
> Make `createMockServer` return synchronously (typed `Hono`) and update docs, playground, and tests to remove unnecessary `await`.
> 
> - **Core**
>   - `createMockServer` now returns synchronously with explicit `Hono` type (`packages/mock-server/src/createMockServer.ts`).
> - **Playground**
>   - Update `galaxy-scalar-com.ts` and tests to use synchronous `createMockServer` and `Hono` typings; minor handler/callback type cleanups.
> - **Tests**
>   - Remove `await` from all `createMockServer` calls across authentication/OpenID/OpenAPI tests and `galaxy` tests.
>   - Use `vi.stubEnv`/`vi.unstubAllEnvs`; refine console mocks and optional callback invocation.
> - **Docs**
>   - README examples updated to drop `await` when creating the mock server.
> - **Release**
>   - Add changeset for `@scalar/mock-server` minor version bump.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit d8c9a9575f2c492ddfc45b2fc679a8af743e7b3d. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/mock-server/src/createMockServer.ts
+++ b/packages/mock-server/src/createMockServer.ts
@@ -17,7 +17,7 @@ import { respondWithOpenApiDocument } from './routes/respondWithOpenApiDocument'
 /**
  * Create a mock server instance
  */
-export function createMockServer(options: MockServerOptions) {
+export function createMockServer(options: MockServerOptions): Hono {
   const app = new Hono()
 
   /** Dereferenced OpenAPI document */
--- a/.changeset/little-dolls-clap.md
+++ b/.changeset/little-dolls-clap.md
@@ -1,0 +1,5 @@
+---
+'@scalar/mock-server': minor
+---
+
+feat(mock-server): convert `createMockServer` and all dependent references to synchronous implementation
--- a/packages/mock-server/README.md
+++ b/packages/mock-server/README.md
@@ -69,7 +69,7 @@ const specification = {
 }
 
 // Create the mocked routes
-const app = await createMockServer({
+const app = createMockServer({
   specification,
   // Custom logging
   onRequest({ context, operation }) {
@@ -157,7 +157,7 @@ const specification = {
 }
 
 // Create the mocked routes
-const app = await createMockServer({
+const app = createMockServer({
   specification,
   // Custom logging
   onRequest({ context, operation }) {
--- a/packages/mock-server/playground/src/galaxy-scalar-com.test.ts
+++ b/packages/mock-server/playground/src/galaxy-scalar-com.test.ts
@@ -1,9 +1,11 @@
 import { readFile } from 'node:fs/promises'
+
 import { serve } from '@hono/node-server'
 import { Scalar } from '@scalar/hono-api-reference'
 import { createMockServer } from '@scalar/mock-server'
-import type { Context } from 'hono'
-import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
+import type { Context, Hono } from 'hono'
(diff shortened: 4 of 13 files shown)


geoffgscott:
While this no longer has a strict need to be asynchronous I would prefer to leave it as an async init. It's not unlikely that in the future some async startup logic would be required and it's disruptive to move between the two functions.

Is there a strong reason to move to synchronous?

packages/mock-server/src/createMockServer.ts
```diff
@@ -17,7 +17,7 @@ import { respondWithOpenApiDocument } from './routes/respondWithOpenApiDocument'
 /**
  * Create a mock server instance
  */
-export function createMockServer(options: MockServerOptions) {
```
marcalexiei, line 20: Hello @geoffgscott,

> Is there a strong reason to move to synchronous?

There isn’t any particular need at the moment.  
However, the function isn’t currently marked as asynchronous, and its inferred return type is `Hono<...>`.

I’ve already updated a few of its usages in https://github.com/scalar/scalar/pull/7121.

> While this no longer has a strict need to be asynchronous I would prefer to leave it as an async init. It's not unlikely that in the future some async startup logic would be required and it's disruptive to move between the two functions.

If you’d prefer to keep the function `async`, I can explicitly set `Promise<Hono>` as return type and return `Promise.resolve(app)` at the end.  
That way, it remains compatible with potential asynchronous behavior in the future.

WDYT?

marcalexiei: I just realized that with the changes introduced in #7121 (already merged into `main`), 
the `createMockServer` declaration will become synchronous in the next release.

<details><summary>0.5.40 published code</summary>
<p>

https://www.npmjs.com/package/@scalar/mock-server?activeTab=code

Screenshot 2025-10-27 at 18 28 23

</p>
</details>

Given the current state, this PR effectively follows up on #7121:

- it adds a changeset to clearly document the update to the `createMockServer` signature 
- removes unnecessary `await` keywords across the codebase.

---

If you’d prefer to have `createMockServer` asynchronous, let me know.

I can open a new PR with the approach I mentioned earlier (explicit return type + `Promise.resolve(app)` at the end) and restoring the await across the entire codebase.

geoffgscott: Yeah let's just make the function async and return `Promise<Hono>`. Explicit return type is far preferred and I don't think we need to make an API change. 

I am not even sure you need the Promise.resolve. 

Thanks!

marcalexiei: If the function is marked as `async` and have no `await` it will trigger [`useAwait`  rule](https://github.com/scalar/scalar/pull/7091#discussion_r2429906517).

I'm working on enable this rule in the past week.

Using `Promise.resolve` alongside `Promise<Hono>` as return type,
will allow to keep the function asynchronous without encouraging static analysis issues like `useAwait` rule reports.

I will close this PR and open a new one later today.

PR #7187, opened 2025-10-28
chore: remove unused dependencies and refine CONTRIBUTING

marcalexiei:
- Removed unused dependencies from the root `package.json`:
  - `start-server-and-test`
  - `syncpack`
- Updated `CONTRIBUTING.md` to include instructions for running test servers using `pnpm script run test-servers`

I've gone through the following:

- [x] I've added an explanation _why_ this change is needed.
- [x] I've added a changeset (`pnpm changeset`) (Not needed).
- [x] I've added tests for the regression or new feature (Not needed).
- [x] I've updated the documentation (Not needed).

---

> [!NOTE]
> Removes unused devDependencies and updates CONTRIBUTING to use a single test server script and a text-fenced semantic PR example.
> 
> - **Documentation (`CONTRIBUTING.md`)**:
>   - Replace separate server start commands with `pnpm script run test-servers`.
>   - Use `text` fence for the semantic PR title example.
> - **Tooling**:
>   - Remove unused devDependencies `start-server-and-test` and `syncpack` from root `package.json`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f7f566c0059971a916d713dd9d61957841ec8716. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #7188, opened 2025-10-28
ci: adds “Required CI OK” to simplify branch rule protection match

marcalexiei:
**Problem**

While working on #7187, I noticed that the required job `test-packages` was no longer reporting its status:

image

---

After a quick investigation, I found that this job was removed in #7179.

**Solution**

This PR introduces a new job called "Required CI OK", which depends on all currently required CI jobs.  
If any of those dependent jobs end with a non-`success` result, this job will exit with code `1`.

With this approach, you only need to mark `Required CI OK` as a required check in your repository settings.  
This makes it possible to update required CI jobs directly in the `CI.yml` file,
without needing to modify branch protection rules or rulesets each time.

**Checklist**

- [x] I've added a changeset (`pnpm changeset`) (Not needed).

---

> [!NOTE]
> Adds a `required-ci-ok` job that aggregates required CI jobs and fails if any are not successful, enabling a single required check.
> 
> - **CI**:
>   - Add `required-ci-ok` job in `.github/workflows/ci.yml`.
>     - Depends on `build`, `format`, `test-packages-shard-1`, `test-packages-shard-2`, `test-integrations`, `test-proxy-server`, `types`.
>     - Always runs and exits with error if any dependency is `failure`, `skipped`, or `cancelled`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 022afaa85777dea03d5d8d094f8712b209d65cd4. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1480,3 +1480,22 @@ jobs:
         run: cargo publish
         env:
           CARGO_REGISTRY_TOKEN: ${{ secrets.CARGO_REGISTRY_TOKEN }}
+
+  required-ci-ok:
+    name: Required CI OK
+    runs-on: ubuntu-latest
+    timeout-minutes: 1
+    permissions: {}
+    if: always()
+    needs:
+      - build
+      - format
+      - test-packages-shard-1
+      - test-packages-shard-2
+      - test-integrations
+      - test-proxy-server
+      - types
+    steps:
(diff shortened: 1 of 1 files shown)


geoffgscott (approved):
This is great. Thanks for looking at this.

PR #7196, opened 2025-10-28
fix(openapi-parser): remove internal unneeded async logic without changing the public API

marcalexiei:
**Problem**

The `Validator` class defines `getAjvValidator` as `async`, even though the function contains no asynchronous logic.

**Solution**

This PR follows the same approach as #7147 but avoids introducing any breaking changes 
([see similar related discussion](https://github.com/scalar/scalar/pull/7174)).

The change is straightforward and unblocks my work on enabling the [`useAwait` Biome rule](https://github.com/scalar/scalar/pull/7091#discussion_r2429906517), 
while allowing you to decide later whether to make the public API synchronous.

**Checklist**

---

> [!NOTE]
> Remove unnecessary async from validator internals, keep `validate` returning a Promise, and update types/tests accordingly.
> 
> - **openapi-parser**:
>   - Convert `Validator.validate` and `getAjvValidator` to synchronous methods; cache validators via `Record<string, ValidateFunction>`.
>   - Narrow `version` to `OpenApiVersion`; make `jsonSchemaVersions` internal; refine Ajv class selection typing.
>   - Use `validateSchema.errors[0].message`; stop merging `schemaResult.errors` into returned `errors`.
> - **utils/validate**:
>   - Keep public `validate(...)` returning a Promise via `Promise.resolve/reject`, while using the now-synchronous `Validator` internally.
> - **Tests**:
>   - Update specs to non-async `describe`; adjust thrown-error expectations to match new sync internals while still awaiting the Promise.
> - **Changeset**: add patch note for `@scalar/openapi-parser`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 3c912569fc6a3869abe2875a04e5e36785a17948. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/openapi-parser/src/lib/Validator/Validator.ts
+++ b/packages/openapi-parser/src/lib/Validator/Validator.ts
@@ -1,4 +1,4 @@
-import type { Ajv, ValidateFunction } from 'ajv'
+import type { ValidateFunction } from 'ajv'
 import Ajv2020 from 'ajv/dist/2020.js'
 import Ajv04 from 'ajv-draft-04'
 import addFormats from 'ajv-formats'
@@ -12,7 +12,7 @@ import { transformErrors } from '@/utils/transform-errors'
 /**
  * Configure available JSON Schema versions
  */
-export const jsonSchemaVersions = {
+const jsonSchemaVersions = {
   'http://json-schema.org/draft-04/schema#': Ajv04,
   'https://json-schema.org/draft/2020-12/schema': Ajv2020,
 }
@@ -133,7 +133,7 @@ export class Validator {
     const schema = OpenApiSpecifications[version]
 
     // Load JSON Schema
-    const AjvClass = jsonSchemaVersions[schema.$schema] as typeof Ajv
+    const AjvClass = jsonSchemaVersions[schema.$schema as keyof typeof jsonSchemaVersions]
 
     // Get the correct Ajv validator
     const ajv = new AjvClass({


amritk (commented):
Looks good to me! But will wait for @hanspagel

packages/openapi-parser/src/lib/Validator/Validator.ts
```diff
@@ -144,7 +140,7 @@ export class Validator {
       // Ajv is a bit too strict in its strict validation of OpenAPI schemas.
       // Switch strict mode off.
       strict: false,
-    })
+    }) as Ajv
```
geoffgscott: Why the typecast here?

marcalexiei: image

`jsonSchemaVersions[schema.$schema]` returns `any` type. 
After reading your comment I understand that is better to cast that expression instead of the `new AjvClass(...)`.

I don't understand why the that expression returns `any`, if you wish I can take a look at it separately.

marcalexiei:
```ts
const AjvClass = jsonSchemaVersions[schema.$schema as keyof typeof jsonSchemaVersions]
```

This change removes the need for casting to `Ajv`.
Also, the `jsonSchemaVersions` export was removed since it's only used internally in this module.

PR #7205, opened 2025-10-29
feat(proxy-scalar-com): do not forward cookie header

bgrcs:
**Problem**

We do not need to forward all client cookies on proxy requests. Any cookies that should be intentionally forwarded to the destination, must be specified using the 'X-Scalar-Cookie' header, which is already supported in client.

**Solution**

Always override the cookie header with X-Scalar-Cookie (or empty string if unset)

**Checklist**

---

> [!NOTE]
> Always override outbound `Cookie` with `X-Scalar-Cookie` (empty if unset) and add tests to ensure no cookies are forwarded by default.
> 
> - **Proxy behavior (`projects/proxy-scalar-com/main.go`)**:
>   - Always set outbound `Cookie` from `X-Scalar-Cookie` and delete `X-Scalar-Cookie`.
>   - Stop forwarding client `Cookie` header by default (cookies only pass if explicitly set via `X-Scalar-Cookie`).
> - **Tests (`projects/proxy-scalar-com/main_test.go`)**:
>   - Add test to assert no cookies are forwarded by default.
>   - Verify `X-Scalar-Cookie` is forwarded as `Cookie` and removed from outbound headers.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit ad8867606aa50b687a51e555af94e7c6921fa040. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


hanspagel (approved):
👍

PR #7233, opened 2025-11-02
chore: add `knip`

marcalexiei:
**Problem**

N/A

**Solution**

This PR is a small proof of concept to try out [Knip](https://knip.dev) on this repository.

Currently, it includes analysis for the following packages and files:
- Root `package.json`
- `scripts`
- `integrations/docusaurus`
- `integrations/fastify`
  - removed `customTheme` from `fastifyApiReference.ts`. 
    It was an empty string and wasn't exported from the package entry point so it wasn't part of the public API.
- `packages/mock-server`

Although this PR introduces only a few changes, 
it should provide a good initial sense of how Knip can help improve and maintain the project over time.
E.g., 
- It can detect automatically if a dependency or a dev dependency is used and avoid issues like https://github.com/scalar/scalar/pull/6934
- Finds out if a file is not used

My plan is to proceed gradually with additional PRs to expand coverage.

Let me know what you think about this. Thanks!

**Checklist**

I've gone through the following:

- [x] I've added an explanation _why_ this change is needed.
- [x] I've added a changeset (`pnpm changeset`).
- [x] I've added tests for the regression or new feature.
- [x] I've updated the documentation (Not needed).

---

> [!NOTE]
> Adds Knip config and CI job, removes unused code/deps, and patches Fastify integration by dropping unused customTheme with tests.
> 
> - **CI/Tooling**:
>   - Add Knip (`knip.jsonc`), root `devDependency`, `lint:knip` script, and `knip` job in `.github/workflows/ci.yml`; gate in `required-ci-ok`.
>   - Remove `stdout-update` from workspace/catalogs and scripts; delete `.syncpackrc`.
>   - Point `turbo.json` `$schema` to local path.
> - **Fastify Integration (`integrations/fastify`)**:
>   - Remove empty `customTheme` and update `getHtmlDocument` call in `fastifyApiReference.ts`.
>   - Tests: add `vi` spy to assert warning when no spec source; minor import/order tweaks.
>   - Types: reorder imports only.
> - **Scripts (`scripts/**`)**:
>   - Delete `src/commands/output.ts`; drop related deps.
>   - Internalize helpers (`pnpmWorkspaceSchema`, `formatJson`, `waitForService`) and minor refactors.
> - **Integrations/Packages**:
>   - Docusaurus: fix `dev:doesnt-work` script and add `@scalar/build-tooling` dev dep.
>   - Mock server: trim unused dev deps in `packages/mock-server` and its `playground`.
> - **Release**:
[...]

PR #7235, opened 2025-11-02
feat: `node:path` polyfill refinements

marcalexiei:
**Problem**

https://github.com/scalar/scalar/discussions/7232#discussioncomment-14847756

> Right now the following file pairs have the same content:
>
> - [`packages/json-magic/src/polyfills/index.ts`](https://github.com/scalar/scalar/blob/2318f6c671eb80581e75c3e88eb076cf5f8705a0/packages/json-magic/src/polyfills/index.ts) &  [`packages/json-magic/src/polyfills/path.ts`](https://github.com/scalar/scalar/blob/2318f6c671eb80581e75c3e88eb076cf5f8705a0/packages/json-magic/src/polyfills/path.ts)
> - [`packages/openapi-parser/src/polyfills/index.ts`](https://github.com/scalar/scalar/blob/2318f6c671eb80581e75c3e88eb076cf5f8705a0/packages/openapi-parser/src/polyfills/index.ts) & [`packages/openapi-parser/src/polyfills/path.ts`](https://github.com/scalar/scalar/blob/2318f6c671eb80581e75c3e88eb076cf5f8705a0/packages/openapi-parser/src/polyfills/path.ts)
> 

**Solution**

I added a new export on helpers package: `@scalar/helpers/node/path`.
This exports expose the polyfill that was duplicated in 2 packages.

This polifyll is now used from `@scalar/json-magic` package.
Regarding `@scalar/openapi-parser` usage I noticed that was used alongside `node:fs` so I don't think that the polifyll is necessary.  
I'll add a review comment so we can discuss this easily.

**Checklist**

---

> [!NOTE]
> Introduce a shared `@scalar/helpers/node/path` polyfill, update `json-magic` to use it, and switch `openapi-parser` to native `node:path`, removing duplicated polyfills and adding tests.
> 
> - **Helpers (`@scalar/helpers`)**
>   - **New polyfill**: Add `@scalar/helpers/node/path` with typed POSIX path utilities in `src/node/path.ts` and export via `package.json` (`"./node/*"`).
>   - **Tests**: Add comprehensive unit tests for `path` functions (`packages/helpers/src/node/path.test.ts`).
> - **JSON Magic (`@scalar/json-magic`)**
>   - **Refactor**: Replace internal path polyfill with `@scalar/helpers/node/path` in `src/bundle/bundle.ts`.
>   - **Cleanup**: Remove duplicated polyfill re-exports (`src/polyfills/index.ts`).
> - **OpenAPI Parser (`@scalar/openapi-parser`)**
>   - **Fix**: Use native `node:path` (`dirname`, `join`) in `plugins/read-files/read-files.ts`.
>   - **Cleanup**: Remove duplicated polyfill files (`src/polyfills/*`).
>   - **Tests**: Minor typo fix in `read-files.test.ts`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 2860d2ec207669fa62479c6d3aa78a412065545d. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/helpers/src/node/path.ts
+++ b/packages/helpers/src/node/path.ts
@@ -217,7 +217,7 @@ export function extname(path: string): string {
   return splitPath(path)![3] as string
 }
 
-export default {
+export const path = {
   extname,
   basename,
   dirname,
--- a/packages/json-magic/src/bundle/bundle.ts
+++ b/packages/json-magic/src/bundle/bundle.ts
@@ -1,4 +1,4 @@
-import path from '@scalar/helpers/node/path'
+import { path } from '@scalar/helpers/node/path'
 
 import { convertToLocalRef } from '@/helpers/convert-to-local-ref'
 import { getId, getSchemas } from '@/helpers/get-schemas'


packages/helpers/src/node/path.ts
marcalexiei, line 1: I tried to keep the original implementation and resolve type issues using only type annotations.
However the file contains two helpers that deal with `Array.prototype.filters` and `String.prototype.substr`.

I would consider removing those helpers and using the native version directly (also substitute `substr` with `substring`).
What do you think?

packages/openapi-parser/src/plugins/read-files/read-files.ts
```diff
@@ -1,7 +1,7 @@
 import fs from 'node:fs'
+import { dirname, join } from 'node:path'
```
marcalexiei, line 2: The original intent was to using the new helper path polyfill also here,
however, I noticed that the file was importing `fs` from `node:fs` so I thought that using a polyfill wasn't necessary since we already are in a node env.

hanspagel (commented):
yeah, maybe we can just use `node:path`? I don’t remember why we needed a polyfill, but if the build and the tests pass, it’s fine.

marcalexiei:
Hey @hanspagel,

> yeah, maybe we can just use node:path? I don’t remember why we needed a polyfill,

I done a quick search:
`packages/json-magic/src/bundle/bundle.ts` ends up bundled in the standalone file,
so I think that the polifyll was included because some of the code was executed in the browser?
All path functions are used inside `resolveReferencePath` which might get called when in a browser.

---

> I don’t remember why we needed a polyfill, but if the build and the tests pass, it’s fine.

Ok, I'll update this PR later to see if Ci complains about the change.

hanspagel:
ah yeah, if the relevant code actually ends up in a browser, sure.
I thought it would just be used for the read files plugin or something

thanks for looking into this

marcalexiei:
Just pushed 2 minutes after seeing your comment 😅. 
I restored the usage of the polyfill exported by helpers.

If you have the change, could you please provide feedback on the two review comments I posted earlier?
- [Usage of modern API for string and array operations](https://github.com/scalar/scalar/pull/7235#discussion_r2484230212)
- [`openapi-parser` - Use `node:path` instead of polyfill](https://github.com/scalar/scalar/pull/7235#discussion_r2484231184)

marcalexiei:
Adding for completeness:

Using `node:path` in `json-magic` `bundle` causes CI to fail:
https://github.com/scalar/scalar/actions/runs/19031676179/job/54347195263:

```text
ERROR in node:path
Module build failed: UnhandledSchemeError: Reading from "node:path" is not handled by plugins (Unhandled scheme).
Webpack supports "data:" and "file:" URIs by default.
You may need an additional plugin to handle "node:" URIs.
```

packages/helpers/src/node/path.ts
hanspagel: Yeah, sure! 👍

packages/openapi-parser/src/plugins/read-files/read-files.ts
hanspagel: true

hanspagel (approved):

packages/helpers/src/node/path.ts
marcalexiei: Addressed by https://github.com/scalar/scalar/pull/7235/commits/eb3199de1dc2d604d619fd22f27fcf29e9738cc0

packages/helpers/src/node/path.ts
```diff
@@ -193,56 +198,34 @@ export function dirname(path) {
...
+  basename,
+  dirname,
+  sep,
+  delimiter,
+  relative,
+  join,
+  isAbsolute,
+  normalize,
+  resolve,
 }
```
amritk:
```suggestion
export const path = {
  extname,
  basename,
  dirname,
  sep,
  delimiter,
  relative,
  join,
  isAbsolute,
  normalize,
  resolve,
}
```

hey looks good! Can we just avoid the default export?  I would have thought biome would have caught this.

marcalexiei: > Can we just avoid the default export? 

I usually avoid `default` export, however  using it would make the module signature different from `node:path`.
No problem making the change though.

> I would have thought biome would have caught this.

I think that is because [noDefaultExport](https://biomejs.dev/linter/rules/no-default-export/) rule is not explicitly enabled in the repo.
Consider that recommended rules are disabled: 
https://github.com/scalar/scalar/blob/8351cf41b57d23a3ad6caa1d05078af0b14db135/biome.json#L100

---

After finishing the work on [enabling `knip`](https://github.com/scalar/scalar/pull/7233) across the entire repository, I can check which recommended rules can be enabled to improve code consistency and quality of code. 
Since you mentioned this, I’d start with enabling `noDefaultExport` 😅.

amritk: hey you are a legend! haha. Lets make the change, no need to follow nodes bad habbits.

amritk: we probably didn't turn on recommended because there would be too many infractions 😓, were getting closer though!

marcalexiei: > hey you are a legend! haha. Lets make the change, no need to follow nodes bad habbits.

Addressed via https://github.com/scalar/scalar/pull/7235/commits/2860d2ec207669fa62479c6d3aa78a412065545d

geoffgscott:
I would rather us maintain this. Polyfill libraries have been and endless headache in the past. This also encourages a least-usage pattern as you have to roll any polyfills we need.

PR #7243, opened 2025-11-03
fix(api-reference): can not use locales as slug for sources (e.g. “ru”)

### LINKED ISSUE #7228: Locales used as slug values don't trigger schema loading
### What happens?

When dynamically loading schemas based on a URL parameter named api (slug), values that match locale codes like `en`, `ru`, etc. don't trigger the loading of sources. Any other arbitrary values of the api parameter work correctly.

**Steps to reproduce the problem:**

1.

```tsx
import React from 'react';
import {ApiReferenceReact} from '@scalar/api-reference-react';
import '@scalar/api-reference-react/style.css';

const App: React.FunctionComponent = () => {
    return (
        <>
            <Header/>

            <ApiReferenceReact configuration={{
                showToolbar: 'never',
                hideClientButton: true,
                hideTestRequestButton: true,
                withDefaultFonts: false,
                orderSchemaPropertiesBy: 'preserve',
                orderRequiredPropertiesFirst: true,
                hiddenClients: true,
                defaultHttpClient: {
                    targetKey: 'shell',
                    clientKey: 'curl'
                },
                sources: [
                    {
                        metaData: {
                            title: 'API'
                        },
                        slug: 'ru',
                        url: `/api/ru.yaml`,
                        default: true
                    },
                    {
                        metaData: {
                            title: 'API'
                        },
                        slug: 'en',
                        url: '/api/en.yaml'
                    }
                ]
            }}/>
        </>
    );
};

export {App};
```

2.
https://example.com?api=ru // not working
https://example.com?api=123 // working

hanspagel:
Looks like you can't use short slugs for sources, see #7228.

This PR fixes #7228. :)

EDIT: I wasn’t really convinced of the change, and tested it again. I can’t reproduce the issue anymore. Investigating.

---

> [!NOTE]
> Fix slug detection by parsing the first URL segment and not prefixing with activeSlug in multi-document mode, enabling short locale slugs like "ru".
> 
> - **ApiReference.vue**:
>   - **Slug detection in multi-document mode**:
>     - Avoid using `activeSlug` as prefix when calling `getIdFromUrl`.
>     - Extract slug from the first non-empty URL segment and verify it exists in `configList`.
> - **Release**:
>   - Add patch changeset for `@scalar/api-reference`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit b6ee6edb2d1df1a7e18275c9d27f88ad085e56e6. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/.changeset/nasty-oranges-begin.md
+++ b/.changeset/nasty-oranges-begin.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
+fix: can not use locales as slug for sources (e.g. “ru”)


packages/api-reference/src/components/ApiReference.vue
```diff
@@ -190,16 +190,25 @@ if (typeof window !== 'undefined') {
...
 
-  if (documentSlug && configList.value[documentSlug]) {
+  // Extract the document slug from the initialId
+  // Split by '/' and take the first segment, which should be the document slug
+  const documentSlug = initialId.split('/').filter(Boolean)[0]
+
+  // Check if the extracted slug exists in configList
+  // This handles all slugs including short ones like 'ru', 'en', etc.
+  // We need to check if the slug exists as a key in configList
+  if (documentSlug && documentSlug in configList.value) {
```
hanspagel, line 211: safer check

packages/api-reference/src/components/ApiReference.vue
```diff
@@ -190,16 +190,25 @@ if (typeof window !== 'undefined') {
...
       (p) => p && url.pathname.startsWith(p.startsWith('/') ? p : `/${p}`),
     ),
     isMultiDocument.value ? undefined : activeSlug.value,
   )
-  const documentSlug = initialId.split('/')[0]
 
-  if (documentSlug && configList.value[documentSlug]) {
+  // Extract the document slug from the initialId
+  // Split by '/' and take the first segment, which should be the document slug
+  const documentSlug = initialId.split('/').filter(Boolean)[0]
```
hanspagel, line 206: removes empty segments

packages/api-reference/src/components/ApiReference.vue
geoffgscott: I don't think that is it. We need a config for a given slug or we should not be changing the slug. 

This will allow null to be in the configList. 

I don't think it matters anyways because I think the normalization strips out nullish values but this is not safer

PR #7253, opened 2025-11-04
fix(api-reference): upgrade document for client store

hwkr:
Seems like with the refactoring we stopped upgrading the document for the client store (but it expects a `3.1` document).

I'm not sure if this a good way to fix it but it seems to work.

**Checklist**

- [ ] I've added tests for the regression or new feature. ← We should probably do this but I don't know where. 

---

> [!NOTE]
> Upgrade documents to OpenAPI 3.1 before dereferencing, expose the typed dereferenced doc, and add a test verifying Swagger 2.0 → 3.1 upgrade.
> 
> - **ApiReference component (`packages/api-reference/src/components/ApiReference.vue`)**:
>   - Upgrade document via `upgrade(raw).specification` before `dereference`, ensuring OpenAPI 3.1 schema for the client store.
>   - Handle missing `schema` by setting `dereferenced` to `null`.
>   - Type `dereferenced` as `OpenAPIV3_1.Document` and `defineExpose({ dereferenced })`.
>   - Import `upgrade` and relevant OpenAPI 3.1 types.
> - **Tests (`packages/api-reference/src/components/ApiReference.test.ts`)**:
>   - Add test verifying Swagger 2.0 documents are upgraded and dereferenced to OpenAPI `3.1.1`.
> - **Changeset**:
>   - Patch release for `@scalar/api-reference` with fix description.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit b271c524aa4058f43ca56d8eca16edf009dbc58a. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/api-reference/src/components/ApiReference.vue
+++ b/packages/api-reference/src/components/ApiReference.vue
@@ -491,13 +491,12 @@ const changeSelectedDocument = async (
 
   // Map the document to the client store for now
   const raw = JSON.parse(workspaceStore.exportActiveDocument('json') ?? '{}')
-  const { schema } = dereference(raw)
+  const { schema } = dereference(upgrade(raw).specification)
   if (!schema) {
     dereferenced.value = null
     return
   }
-  const { specification } = upgrade(schema)
-  dereferenced.value = specification
+  dereferenced.value = schema as OpenAPIV3_1.Document
 }
 
 /**
--- a/packages/api-reference/src/components/ApiReference.test.ts
+++ b/packages/api-reference/src/components/ApiReference.test.ts
@@ -187,7 +187,7 @@ describe('multiple configurations', () => {
 })
 
 describe('circular documents', () => {
-  it.only('handles circular references in schemas gracefully', async () => {
+  it('handles circular references in schemas gracefully', async () => {
     const wrapper = mount(ApiReference, {
       props: {
         configuration: {


packages/api-reference/src/components/ApiReference.vue
```diff
@@ -490,7 +491,13 @@ const changeSelectedDocument = async (
 
   // Map the document to the client store for now
   const raw = JSON.parse(workspaceStore.exportActiveDocument('json') ?? '{}')
-  dereferenced.value = dereference(raw).schema
+  const { schema } = dereference(raw)
```
hwkr: Schema here can be any version OpenAPI document

packages/api-reference/src/components/ApiReference.vue
```diff
@@ -600,7 +607,7 @@ onBeforeMount(() =>
  * We keep a copy of the workspace store document in dereferenced format
  * to allow mapping to the legacy client store
  */
-const dereferenced = ref<ReturnType<typeof dereference>['schema'] | null>(null)
```
hwkr, line 603: This return type allowed this to be any version OpenAPI doc. I'm not sure why this was allowed to be [passed into `mapConfigToClientStore`](https://github.com/scalar/scalar/blob/main/packages/api-reference/src/components/ApiReference.vue#L622) which expects a 3.1 doc.

amritk: I believe its because the old types will accept anything. Thats why we don't want to use them anymore. Thats also why I requested changes.

BUT the old client isn't long for the world so maybe we'll let it slide. I'll add a test then send it

hwkr: Sounds good ty!!

PR #7264, opened 2025-11-05
refactor: set `noExportsInTest` rule level to `error`

marcalexiei:
## Problem

The file `packages/api-client/src/libs/send-request/create-request-operation.test.ts` was exporting helper functions that were being imported and used in `packages/api-client/src/libs/send-request/create-fetch-auth.test.ts`.

As a result, running the `create-fetch-auth` tests (e.g., via `pnpm test [file-path]`) would also trigger the `create-request-operation` tests unintentionally.

## Solution

This PR relocates the exported helpers from `create-request-operation.test.ts` (along with the Vitest setup file) into a dedicated `test` directory, 
aligning with the test structure used across other packages in the repository.

#### Repository

* Enforce the `noExportsInTest` Biome rule (set to **error**) to prevent similar issues in the future.

#### `api-client` Package

* Disallow imports from `@test/*` within source files using the `noRestrictedImports` rule.
* Exclude test files from build-time type checking, reducing build time (`scalar-types-build-vue` from ~15s → ~10s).
  * Type checking for test files still runs separately via `scalar-types-check-vue`.

**Checklist**

- [x] I've added a changeset (`pnpm changeset`) (Not needed).

---

> [!NOTE]
> Relocates test helpers/setup to `packages/api-client/test`, updates imports/configs accordingly, and tightens linting to disallow exports in tests and importing test helpers from source.
> 
> - **Linting/Repository**:
>   - Set `noExportsInTest` to `error` in `biome.json`.
>   - Add `noRestrictedImports` (error) for source files to block `@test/*` and `**/test/**` imports.
> - **api-client (tests)**:
>   - Create `packages/api-client/test/helpers.ts` and `test/vitest.setup.ts`.
>   - Update tests to import from `@test/*` and remove cross-test exports; `create-fetch-auth.test.ts` now uses `@test/helpers`.
>   - Refactor `create-request-operation.test.ts` to consume helpers/constants from `@test/helpers`.
> - **Build/Config**:
>   - Update `tsconfig.json` paths to include `@test/*` and include `test` dir.
>   - Update `tsconfig.build.json` to include only `src` and exclude `**/*.test.ts`.
>   - Point Vite test `setupFiles` to `./test/vitest.setup.ts`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit a77b8aef9cba9c35b5c0a815b64572d2a01fe3a2. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/api-client/biome.json
```diff
@@ -0,0 +1,26 @@
+{
+  "$schema": "../../node_modules/@biomejs/biome/configuration_schema.json",
```
hanspagel: can we have this somehow in the root biome.json? I think it’s better to have the configuration for it in one central place. Wdyt?

marcalexiei: I see that you already have package-specific overrides in the root `biome.json`, 
so it’s more consistent to move this configuration there.
Thanks for bringing this up!

On second thought, this approach is also better since `@test/*` appears in other packages as well,
so this behaviour can be applied more broadly.

I'll run a few checks later and let you know.

marcalexiei: Moving this setting in the root config and apply it broadly on `packages/*` is feasible!

<details><summary>Log example of when this rule fails in 2 different packages:</summary>
<p>

 `pnpm run lint:check`

image

</p>
</details>

I’ve rebased and force-pushed the branch, but kept a separate commit to simplify the review.
I also removed an unnecessary match for the `tsconfig` files, which is no longer needed.

amritk (approved):
Improving the codebase, one PR at a time!

PR #7265, opened 2025-11-05
fix(api-reference): teleport example picker to avoid occlusion

### LINKED ISSUE #6610: Endpoint request examples select not works for more than 3 options
### What happens?

When more than 3 options are available in the selection, I cannot pick them:

https://github.com/user-attachments/assets/4f7395ad-8ef5-4222-bb93-ace16ae8ee18

This issue does not occur when using the keyboard:

https://github.com/user-attachments/assets/388fdaad-989e-4157-925f-8762b77e3fe0

I expect to be able to select any available option on select using mouse.

The API docs used as example: https://api.syki.com.br/docs/v1#tag/-academic/post/academic/classes

### OpenAPI Document

https://api.syki.com.br/swagger/v1/swagger.json

hwkr:
The example picker was sometimes getting covered from other elements so this teleports it to the nearest teleport root.

Fixes #6610

**Checklist**

### DIFF
--- a/packages/api-client/src/v2/blocks/operation-code-sample/components/ExamplePicker.test.ts
+++ b/packages/api-client/src/v2/blocks/operation-code-sample/components/ExamplePicker.test.ts
@@ -171,14 +171,20 @@ describe('ExamplePicker', () => {
         examples: problematicExamples,
         modelValue: '',
       },
+      attachTo: document.body,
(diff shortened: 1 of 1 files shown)


hwkr:
> why are tests failing?

Oops, looks like they need to be updated

PR #7266, opened 2025-11-05
fix(json-magic): remove usage of crypto.subtle

amritk:
**Problem**

closes #7263

Currently, crypto.subtle does not exist on non-secure, non-localhost apps

**Solution**

With this PR we move the hashing document functions to use a much faster algorithm. It has a dependency so I moved it out of helpers and back into json-magic.
image

We can still keep using crypto.subtle in the client with the added failsafes. I did write a biome plugin to ban this though if we need it
image

**Checklist**

---

> [!NOTE]
> Switch hashing to xxhash-wasm (removing crypto.subtle), add PKCE SHA-256 fallback, and update builds/tests across packages.
> 
> - **json-magic**:
>   - **Hashing**: Add `helpers/generate-hash` using `xxhash-wasm`; update `bundle/value-generator` to use it and return 7-char ids.
>   - **Build/Exports**: Include `src/helpers/generate-hash.ts` in build and export map; add dependency `xxhash-wasm`.
>   - **Tests**: Update expected `$ref` keys and `x-ext`/URL maps to new hashes.
> - **workspace-store**:
>   - **Integration**: Use `@scalar/json-magic/helpers/generate-hash` for `x-scalar-original-document-hash` and refs; adjust tests to new shorter hashes.
> - **api-client**:
>   - **PKCE**: In `oauth.ts`, handle absence of `crypto.subtle.digest` by logging a warning and falling back to plain code challenge.
> - **api-reference**:
>   - **Test reliability**: Use `vi.waitFor` to wait for dereferencing before assertions.
> - **helpers**:
>   - **Removal**: Delete `crypto/generate-hash` and its tests; note deprecation in `string/create-hash.ts` JSDoc.
> - **Meta**:
>   - Add changeset; update `pnpm-lock.yaml` for `xxhash-wasm`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 058e1b449ec41165e20b94635129eb1058aa0685. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/workspace-store/src/client.test.ts
+++ b/packages/workspace-store/src/client.test.ts
@@ -170,7 +170,7 @@ describe('create-workspace-store', () => {
       },
       'x-original-oas-version': '3.0.0',
       'x-ext-urls': {},
-      'x-scalar-original-document-hash': 'cb852d702f435acc04ac19ac9b9fcf96ed0679cd0f5d63444d4f0125c47d165d',
+      'x-scalar-original-document-hash': 'c5d4d738bfadaeb6',
     })
   })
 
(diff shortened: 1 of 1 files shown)


DemonHa (approved):
LGTM 🚀

PR #7274, opened 2025-11-05
fix(object-utils): enable `knip`

marcalexiei:
**Problem**

- Followup of #7233

**Solution**

Enable `knip` on `object-utils` package

**Checklist**

---

> [!NOTE]
> Enables knip for `object-utils`, switches to explicit build entries, removes the root index export and `type-fest`, and prunes unused nested type utilities.
> 
> - **object-utils**:
>   - **Knip**: Enable by adding `workspaces.packages/object-utils.entry` in `knip.jsonc` and stop ignoring the package.
>   - **Build**: Switch `esbuild` from `entries: 'auto'` to explicit entries for `arrays`, `clone`, `merge`, `mutator-record`, `nested`, `parse`, `transforms`.
>   - **Exports**: Remove root `"."` export and `main/module` fields; keep only subpath exports (`./arrays`, `./clone`, `./merge`, `./mutator-record`, `./nested`, `./parse`, `./transforms`).
>   - **Dependencies**: Drop unused `type-fest`; update lockfile.
>   - **Types**: Remove unused nested path/key utility types in `src/nested/common.ts` (e.g., `Path*`, key/path helpers); comment out `IsNever`.
> - **Changesets**: Add patches noting removal of empty index export and `type-fest`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit dc0f2b68c357f123e54fa69ea6f17cf01f3bb194. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/object-utils/src/nested/common.ts
```diff
@@ -24,7 +24,7 @@ export type IsAny<T> = 0 extends 1 & T ? true : false
  * IsAny<string> = false
  * ```
  */
-export type IsNever<T> = [T] extends [never] ? true : false
```
xC0dex, line 27: I usually find it helpful to leave a quick note about _why_ a type is commented out, but I’m not certain it’s necessary in this case.

hanspagel: I’m not sure we even want to remove them. I think the idea was to just have them (when needed).

marcalexiei: I can add these files to ignores if you wish to keep them.
Consider that they are not even exported from the package `/nested`entry point,  so I'm not sure if they are useful.
Alternatively we should export them from `nested/index`.

hanspagel: I think @geoffgscott should do a review here 👍

geoffgscott: These were originally use for dot-path and array-path deep get/set. I am happy to remove; this was always an uphill battle in TS.

packages/object-utils/src/index.ts
geoffgscott, line 1: Previously these files were required to allow the 'auto' entry points logic to work. We just need to verify that the ESBuild config has explicit entry points (which is better practice anyways)

marcalexiei: Addressed by https://github.com/scalar/scalar/pull/7274/commits/f95d7ca224e0cd1e73a3dfcaf6fb2df10c0f317a

packages/object-utils/src/nested/common.ts
marcalexiei: Addressed from https://github.com/scalar/scalar/pull/7274/commits/dc0f2b68c357f123e54fa69ea6f17cf01f3bb194

PR #7306, opened 2025-11-08
feat(components): add ScalarThemeSwatches component

hwkr:
Adds the swatch component from org to `@scalar/components` and uses it in the dev toolbar.

CleanShot 2025-11-07 at 21 18 32@2x
CleanShot 2025-11-07 at 21 18 00@2x

**Checklist**


hanspagel (approved):
cool

tested & works great

PR #7329, opened 2025-11-10
fix(workspace-store): setting nested proxies directly cases problems

DemonHa:
Previously, our mutators were assigning values directly from the payload without checking whether they contained nested proxies. This caused issues when the values originated from reactive objects or the active document, as we could end up storing proxy references instead of pure objects.

**What's Changed:**
- Added logic to ensure all assigned values are pure objects (non-proxy) before mutation.
- Prevents nested proxies from being assigned to properties.
- Allows safe use of structuredClone on inputs without losing internal references when working with magic proxies.
- Avoids the need for deep unwrapping of nested proxies, improving performance by only unpacking at the top-level.
- - We still do try to unwrap to a certain depth, to ensure that we don't end up with runtime errors in production code

**Why**

Ensuring that only raw objects are stored:

- Prevents accidental proxy nesting.
- Improves clone safety and data integrity.
- Reduces performance overhead from deep proxy traversal.

When the user try to set a proxies value directly we display a user friendly error explaining what is happening and how to fix this issue together with more debug information:

image

We also turned the trace mode on dev environment so we can trace the problem

**Checklist**

---

> [!NOTE]
> Unwrap proxied objects (with depth control) and use this in the event bus, enable debug mode in dev, and adjust API client tests/timeouts.
> 
> - **Workspace Store**:
>   - **Proxy Unwrapping**: Add `helpers/unpack-proxy.ts` with depth-limited unwrapping of Vue/magic/override/detect-changes proxies and warning/trace logs.
>   - **Event Bus**: `events/bus.ts` now unwraps emitted payloads via `unpackProxyObject(payload, { depth: 5 })` before notifying listeners.
>   - **Tests**: Add comprehensive `helpers/unpack-proxy.test.ts` covering primitives, proxy types, depth limits, and circular refs.
> - **API Client**:
>   - **App**: Pass `{ debug: import.meta.env.DEV }` to `createWorkspaceEventBus`.
>   - **Tests**: Increase suite timeout and replace short waits with longer delays (10s) for document loading; fix typo in test name.
> - **Changeset**: Patch bumps for `@scalar/workspace-store` and `@scalar/api-client`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 99806641d6d4c30db1d5a3171629d673e8442e55. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/api-client/src/v2/hooks/use-workspace-selector.test.ts
+++ b/packages/api-client/src/v2/hooks/use-workspace-selector.test.ts
@@ -18,7 +18,7 @@ vi.mock('vue-router', () => ({
   })),
 }))
 
-describe('useWorkspaceSelector', { concurrent: false, sequential: false }, () => {
+describe('useWorkspaceSelector', { concurrent: false, sequential: false, timeout: 50_000 }, () => {
   const persistencePromise = createWorkspaceStorePersistence()
 
   beforeEach(() => {
@@ -51,7 +51,7 @@ describe('useWorkspaceSelector', { concurrent: false, sequential: false }, () =>
     await flushPromises()
 
     // We are loading documents from an url so it takes a while to load
-    await new Promise((resolve) => setTimeout(resolve, 1000))
+    await new Promise((resolve) => setTimeout(resolve, 10_000))
 
     expect(store.value).not.toBeNull()
     assert(store.value)
@@ -103,7 +103,8 @@ describe('useWorkspaceSelector', { concurrent: false, sequential: false }, () =>
 
     await nextTick()
     await flushPromises()
-    await new Promise((resolve) => setTimeout(resolve, 1000))
+    // change the timeout when we remove the loading default documents
+    await new Promise((resolve) => setTimeout(resolve, 10_000))
 
     expect(activeWorkspace.value).not.toBeNull()
     assert(activeWorkspace.value)
@@ -115,7 +116,7 @@ describe('useWorkspaceSelector', { concurrent: false, sequential: false }, () =>
 
     await nextTick()
     await flushPromises()
-    await new Promise((resolve) => setTimeout(resolve, 1000))
+    await new Promise((resolve) => setTimeout(resolve, 10_000))
 
     // this will set the active workspace to the new workspace
     expect(push).toHaveBeenCalledWith({ name: 'workspace', params: { workspaceSlug: 'new-workspace' } })


DemonHa:
> ### Bug: Circular References Break Recursive Function.
> The recursive `dfs` function lacks circular reference detection, causing infinite recursion and stack overflow when `depth: null` is used (debug mode in the event bus) with objects containing circular references. The function mutates and revisits the same object indefinitely without tracking visited objects via `WeakSet` or similar mechanism.
> 
> `packages/workspace-store/src/helpers/unpack-proxy.ts#L21-L59`
> https://github.com/scalar/scalar/blob/2d666fd03d1e7ef8394f5dcb6a3dfa7686d78eee/packages/workspace-store/src/helpers/unpack-proxy.ts#L21-L59
> 
> [Fix in Cursor](https://cursor.com/open?data=eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCIsImtpZCI6ImJ1Z2JvdC12MSJ9.eyJ2ZXJzaW9uIjoxLCJ0eXBlIjoiQlVHQk9UX0ZJWF9JTl9DVVJTT1IiLCJkYXRhIjp7InJlZGlzS2V5IjoiYnVnYm90OmY0NWZmMmQxLTBjOTgtNDNlOS05MjZjLTE5MTA5MTZmMzI0NSIsImVuY3J5cHRpb25LZXkiOiJfLXdnZmhObzRobENKV0xoc2ExRGtPVmlkVnI4TmxYUTRvUDRwOGktOWg0IiwiYnJhbmNoIjoicmVkaXMvZml4LW5lc3RlZC1wcm94eS1pc3N1ZXMifSwiaWF0IjoxNzYyODE3MTUzLCJleHAiOjE3NjM0MjE5NTN9.fHOvDkPQrJUgHKnvhjhqjA6OKjAFpibw7F_ZgKBPM38s0i9jnpuHbhn4hqX_eJNbeM0rgJtQP4RDrP67rHhlCFr6Hoyt0KxpzAYg5WUNEFNo0wm5wOg37259FRuOJOiTG17mNO1H-khZORIeUhGsfiRWHXJOn7TFPKUKXPhHDGQtSZ2wt0gBLK7OBff4vmPs1iPfS0j9QvWzifMjZr3MeaNkp7atJNc_jl8esSpVmOChk62SZPkSCFLk5oBbBlh4_TViz9f9hFeoQcXynRMTEc_7GcyfEU4MluNR--LYqdZCxT_hF0lyfQViEOaXUoXbXL-pzFQKIeWgIUlHgFVUPw) [Fix in Web](https://cursor.com/agents?data=eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCIsImtpZCI6ImJ1Z2JvdC12MSJ9.eyJ2ZXJzaW9uIjoxLCJ0eXBlIjoiQlVHQk9UX0ZJWF9JTl9XRUIiLCJkYXRhIjp7InJlZGlzS2V5IjoiYnVnYm90OmY0NWZmMmQxLTBjOTgtNDNlOS05MjZjLTE5MTA5MTZmMzI0NSIsImVuY3J5cHRpb25LZXkiOiJfLXdnZmhObzRobENKV0xoc2ExRGtPVmlkVnI4TmxYUTRvUDRwOGktOWg0IiwiYnJhbmNoIjoicmVkaXMvZml4LW5lc3RlZC1wcm94eS1pc3N1ZXMiLCJyZXBvT3duZXIiOiJzY2FsYXIiLCJyZXBvTmFtZSI6InNjYWxhciIsInByTnVtYmVyIjo3MzI5LCJjb21taXRTaGEiOiIyZDY2NmZkMDNkMWU3ZWY4Mzk0ZjVkY2I2YTNkZmE3Njg2ZDc4ZWVlIn0sImlhdCI6MTc2MjgxNzE1MywiZXhwIjoxNzYzNDIxOTUzfQ.LN47qeDFl1B7trO7gqz2TK_J8fNE1A7fDQCemp7Rzy3dmlM6y5p7flgLT0QobiQV8YybsEom9G0R5DuT-EwiB2gIMxJlvoWXHCqll4Ym1VG7uOtDpWvKVhEm0tmC6ZYrMMiVlLnv-tkEmDB-2KTWM3THzIiTyUmUcd5n8zFgTcq0D5IAol2ZotyulceaYxiFna4OMuN-n0mnqNq-09j0WDR0ZCwGv5vWtbncXbkuJAnS5NsSPS4-Ob8qZG9Uns7kh5KQl2Hd1q0x4EjDrQX7R81Ixt6rf3vmSurC_19KMr1Bw6507gJXZ-m8dWv6h2bl8dPPLCfaUH1UCq_5T84kzQ)

we don't really care about circular references here, we are not supposed to have a circular object because the input is gonna be first unwrapped from the proxy before traversing the child nodes. Thats the only case there can be a circular structure so we are safe

packages/workspace-store/src/helpers/unpack-proxy.ts
```diff
@@ -7,6 +7,54 @@ import { unpackOverridesProxy } from '@/helpers/overrides-proxy'
...
+
+    // Show a warning if a nested value was a proxy (usually undesired, can cause subtle bugs)
+    if (currentDepth !== 0 && raw !== value) {
+      console.warn(
+        '%c⚠ Warning:%c You tried to assign a proxied object.\n' +
+          '%c💡 Tip:%c Pass a plain object instead — wrapping a proxy inside another proxy may cause weird bugs.',
+        'background: #fdd835; color: #000; font-weight: bold; padding: 2px 4px; border-radius: 3px;',
+        'color: inherit;',
+        'color: #00bfa5; font-weight: bold;',
+        'color: inherit;',
```
geoffgscott, line 49: Might be good to attach the object that failed here as well for debugging.

geoffgscott (approved):

packages/workspace-store/src/events/bus.ts
```diff
@@ -159,6 +161,12 @@ export const createWorkspaceEventBus = (options: EventBusOptions = {}): Workspac
 
   const emit = <E extends keyof ApiReferenceEvents>(...args: EmitParameters<E>): void => {
     const [event, payload] = args
+
+    // We unpack the payload here to ensure that, within mutators, we are not assigning proxies directly,
+    // but are always assigning plain objects at any level. In debug mode, we allow unlimited traversal for easier inspection,
+    // otherwise limit the depth for performance.
+    const unpackedPayload = unpackProxyObject(payload, { depth: debug ? null : 5 })
```
amritk:
```suggestion
    const unpackedPayload = unpackProxyObject(payload, { depth: 5 })
```

I don't think we should do unlimited for dev, keep it 5 for both otherwise someone might think something works in dev but then it doesn't in prod. Very low chance though!

DemonHa: Yeah but we want to see the warning in dev, there should not be a case when you have more than 5 level deep nested proxy but it's like a safe thing to do?

amritk: you should get the warning with 5 anyway. It will hide the error though if there is one

PR #7331, opened 2025-11-11
ci: new setup action

xC0dex:
# WIP

**Checklist**

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,6 +53,24 @@ jobs:
           token: ${{ secrets.GITHUB_TOKEN }}
           webpackStatsFile: ./packages/api-reference/dist/browser/webpack-stats.json
 
+  build-dummy:
+    runs-on: blacksmith-4vcpu-ubuntu-2204
+    timeout-minutes: 10
+    strategy:
+      matrix:
+        node-version: [22]
+    permissions:
+      contents: read
+    needs: [build]
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5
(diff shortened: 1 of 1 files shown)


.github/actions/setup-node/action.yml
marcalexiei, line 1: Sorry to chime in, but I’d like to suggest an improvement:

Since there’s already a [`.nvmrc`](https://github.com/scalar/scalar/blob/main/.nvmrc) file in the repository root, you could infer the Node.js version directly from it.

In other words, this action could optionally accept a `node-version` input,
and if that input isn’t provided, it could automatically read the version from the `.nvmrc` file.  
For example:

```yml
    - name: Setup Node.js (via input)
      if: ${{ inputs.node-version }}
      uses: actions/setup-node@v4
      with:
        node-version: ${{ inputs.node-version }}
        cache: 'pnpm'

    - name: Setup Node.js (via .nvmrc)
      if: ${{ !inputs.node-version }}
      uses: actions/setup-node@v4
      with:
        node-version-file: '.nvmrc'
        cache: 'pnpm'
```

By doing this, the change would propagate to `build-blacksmith` and you could remove the `matrix.node-version` configuration (which currently only specifies `22`) from various workflows, simplifying them to something like:

```diff
strategy:
-    matrix:
-    node-version: [22]
    steps:
      - name: Checkout repository
        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5
      - name: Build packages
        uses: ./.github/actions/setup-blacksmith
        with:
-        node-version: ${{ matrix.node-version }}
          packages: './integrations/**'
```

xC0dex: Awesome idea! 🙂

PR #7346, opened 2025-11-12
docs(api-reference-react): update react getting started docs

hwkr:
Because of a change in the Vite React starter template our instructions created a broken app. Specifically the started added some global CSS which breaks our app.

Rather than add complicated instructions I just made it so you overwrite into `main.jsx` (which is where the bad css is being imported) instead of `App.jsx`. I figure if anyone needs it in another component they can modify the project themselves.

**Checklist**


hanspagel (approved):
nice! thanks

PR #7349, opened 2025-11-13
fix: restore automatic fill of package exports from build-tooling

marcalexiei:
**Problem**

Right now `package.json#exports` are not automatically updated with values coming from build entry points.
This has made some pointing to no longer existing files or to the wrong files.

**Solution**

Restore exports automatic fill from `build-tooling` script and apply them in the codebase.
Applied fixes are detailed in changeset files.

**Checklist**

---

> [!NOTE]
> Restores automatic export generation and updates packages/configs to align `package.json#exports` with actual build entry points.
> 
> - **Build tooling**:
>   - Restore automatic `package.json` exports fill; add `findEntryPoints({ baseDir })` and `addPackageFileExports({ packageDir })`.
>   - Replace `glob` with `fast-glob`; add unit tests; exclude tests from build; minor barrel annotations.
>   - Update Vite configs (`api-client`, `core`, `draggable`, `openapi-to-markdown`) to call `findEntryPoints` with `baseDir`.
> - **Packages/export alignment**:
>   - `@scalar/api-client`: expand and align `exports` (add `v2/*` blocks/components/features) with built outputs.
>   - `@scalar/types`: remove invalid `./api-client` export; switch to `./api-reference` in build and `exports`.
>   - `@scalar/json-magic`: fix `types` paths for `helpers/escape-json-pointer` and `helpers/generate-hash` exports.
>   - `@scalar/helpers`: add `node/*` and `crypto/*` build entries and `exports`.
>   - `@scalar/workspace-store`: add `plugins/bundler`, `plugins/client`, `helpers/*` exports; expose `schemas/v3.1/strict/openapi-document`.
> - **Scripts**:
>   - Lint scripts now use `fast-glob` to detect Vue files.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 7b6390de1d0b821dacb4440aa44634306bc90f3f. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/api-client/package.json
+++ b/packages/api-client/package.json
@@ -260,6 +260,11 @@
       "types": "./dist/v2/features/operation/index.d.ts",
       "default": "./dist/v2/features/operation/index.js"
     },
+    "./v2/features/search": {
+      "import": "./dist/v2/features/search/index.js",
+      "types": "./dist/v2/features/search/index.d.ts",
+      "default": "./dist/v2/features/search/index.js"
+    },
     "./v2/features/settings": {
       "import": "./dist/v2/features/settings/index.js",
       "types": "./dist/v2/features/settings/index.d.ts",


amritk (approved):
Love it, I was wondering what happened to that. Started writing my exports manually like a sucker.

packages/build-tooling/src/helpers.ts
```diff
@@ -70,12 +69,21 @@ function formatEntry(filepath: string, namespacePath: string) {
...
-export async function addPackageFileExports({ allowCss, entries }: { allowCss?: boolean; entries: string | string[] }) {
+export async function addPackageFileExports({
+  allowCss,
+  entries,
+}: {
+  allowCss?: boolean
+  entries: string | string[]
+}): Promise<void> {
+  const paths = Array.isArray(entries) ? entries : [entries]
+  if (paths.length === 0) {
```
DemonHa: I think we would still want to reset the exports fields when there are no entries

DemonHa (changes requested):

packages/build-tooling/src/helpers.ts
marcalexiei: This is a bit tricky.

When running `pnpm test`, 
packages that use a `vite.config.ts` file with a `test` property (e.g., `api-client`) 
have their `process.cwd()` set to the repository root.

As a result, we get the following behavior:

* The function searches for `src/**/*/index`, 
   but finds nothing because there’s no `src` folder at the root.
* Since the current working directory is the repo root, 
   the script ends up modifying the root `package.json` instead of the package’s own file, 
   as it reads from `./package.json`.

<details><summary>See attached screenshot for more details</summary>
<p>

image

</p>
</details> 

---

The simplest fix is to optionally pass the package dirname to the function 
(defaulting to `process.cwd()` if not provided), 
so that all findEntryPoints references are absolute and this issue is avoided.

packages/build-tooling/package.json
```diff
@@ -68,7 +69,7 @@
     "ansis": "catalog:*",
     "chokidar": "catalog:*",
     "esbuild": "catalog:*",
-    "glob": "^10.3.10",
+    "fast-glob": "^3.3.2",
```
marcalexiei, line 72: I have seen that `fast-glob` is used in other packages so I replace glob with it.

packages/build-tooling/src/helpers.ts
marcalexiei, line 1: Regarding cursor comments:
- https://github.com/scalar/scalar/pull/7349#pullrequestreview-3462655626
- https://github.com/scalar/scalar/pull/7349#pullrequestreview-3462684235
- https://github.com/scalar/scalar/pull/7349#pullrequestreview-3463003532

`esbuild` script is called with the right working directory, so I haven't updated it.
Let me know if you prefer to have `baseDir` / `packageDir` options set here. 
If yes we might want to turn them into required options rather than having them as optional.

packages/api-client/src/v2/hooks/use-document-watcher.test.ts
marcalexiei, line 1: I encountered the following error:

```text
Error: TypeError: setTimeout is not a function
```

in these 2 runs:
- https://github.com/scalar/scalar/actions/runs/19353930932/job/55371600469
- https://github.com/scalar/scalar/actions/runs/19353619597/job/55370715844

In `api-client` `vitest` environment is set to `jsdom`. 
I think that, sometimes, there might be conflicts between it and node internal modules.
I have changes the test to use `vitest` timers utils.
Hopefully it should avoid this kind of errors in the future 🤞

marcalexiei: I'm going to open a separate PR.
Cursor is commenting a lot about changes related to previous commits so is better not to additional changes 😅

marcalexiei: https://github.com/scalar/scalar/pull/7368

packages/build-tooling/src/helpers.ts
marcalexiei: Addressed by https://github.com/scalar/scalar/pull/7349/commits/eaff1d2bce422363806d3a7df0d64bc165f3c66e

marcalexiei:
`setTimeout is not a function errors` are no longer present!
 PR ready for review.

[I left an open discussion regarding Cursor bot comment above](https://github.com/scalar/scalar/pull/7349#discussion_r2525659777).

marcalexiei:
@hanspagel @geoffgscott let me know if you need additional changes.
I ignored last cursor bot comments since build is running on node 22 which supports `import.meta.dirname`.
Let me now if you prefer a solution compatible with node `>= 20`

packages/build-tooling/README.md
```diff
@@ -96,7 +96,9 @@ export default defineConfig({
     port: 5000,
   },
   build: createViteBuildOptions({
-    entry: await findEntryPoints({ allowCss: true }),
+    entry: await findEntryPoints({ 
+      baseDir: import.meta.dirname,
+      allowCss: true }),
```
hanspagel: nitpick alert: a line break is missing right before the brackets

hanspagel (commented):
glad those are back!

can @geoffgscott do a review here?

packages/build-tooling/README.md
marcalexiei: Addressed by https://github.com/scalar/scalar/pull/7349/commits/eb13881219fc555fcf62a97d09f0b7f9a5e98c3e

geoffgscott:
Closing in favour of https://github.com/scalar/scalar/pull/7387. 

Really appreciate the contribution on this one but this PR has just a few too many unrelated changes and I wanted to limit the patch to just the root cause (in this case an incorrect use of `import.meta.dirname` which caused the function to never run).

We can open another PR to do the glob migration if you would like.

PR #7362, opened 2025-11-13
test(api-reference): update sources to use local Scalar Galaxy

hwkr:
A handful of fixes for the references tests:

- Use the local version of Galaxy for snapshots to prevent the CDN version of Galaxy from breaking the snapshots on main
- UUIDs weren't generating because `pnpm` wasn't install yet
- Adds back `showToolbar` to the default config because we haven't released the version with the `showDeveloperTools` option yet so it was breaking and the CDN comparisons.

**Checklist**

### DIFF
--- a/packages/api-reference/test/utils/serve-example.ts
+++ b/packages/api-reference/test/utils/serve-example.ts
@@ -18,6 +18,9 @@ const DEFAULT_CONFIGURATION: Partial<HtmlRenderingConfiguration> = {
   cdn: '/scalar.js',
   url: 'https://registry.scalar.com/@scalar/apis/galaxy?format=json',
   proxyUrl: 'https://proxy.scalar.com',
+  // TODO: Remove this once the CDN supports the showDeveloperTools attribute
+  // @ts-expect-error - we need this until next release (after 1.39.3)
+  showToolbar: 'never',
   showDeveloperTools: 'never', // Hide the toolbar by default for snapshots
 }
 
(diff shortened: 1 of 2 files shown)


packages/api-reference/test/utils/sources.ts
```diff
@@ -12,19 +13,19 @@ export const sources = [
...
-    title: 'Scalar Galaxy Registry',
-    slug: 'scalar-galaxy-registry',
-    url: 'https://registry.scalar.com/@scalar/apis/galaxy?format=json',
+    content: galaxy,
   },
   {
     title: 'Scalar Galaxy (Classic Layout)',
     slug: 'scalar-galaxy-classic',
-    url: 'https://registry.scalar.com/@scalar/apis/galaxy?format=json',
+    content: galaxy,
```
hwkr, line 21: If we'd prefer to use this over a URL we can have `serveExample` host it instead

packages/api-reference/test/utils/serve-example.ts
```diff
@@ -18,6 +18,9 @@ const DEFAULT_CONFIGURATION: Partial<HtmlRenderingConfiguration> = {
   cdn: '/scalar.js',
   url: 'https://registry.scalar.com/@scalar/apis/galaxy?format=json',
   proxyUrl: 'https://proxy.scalar.com',
+  // TODO: Remove this once the CDN supports the showDeveloperTools attribute
+  // @ts-expect-error - we need this until next release (after 1.39.3)
+  showToolbar: 'never',
```
hwkr, line 23: @hanspagel because we haven't released yet and this hits the CDN version we need this for a bit still

hwkr:
> nvm I guess that was the point eh lol

Yeah and we still have the registry version as a source which gets snapshotted in the CDN tests

(And can be used for testing)

PR #7371, opened 2025-11-15
feat(sidebar): convert draggable logic to a hook

hwkr:
Updates the draggable logic to use a hook instead of a wrapper element. We could also port this back into the draggable package if we want (or we could just get rid of the `<Draggable>` component.

Why:
- 1 less DOM node per sidebar entry
- The extra `div` was messing up some of the sidebar styles (e.g. using `:first-child` selectors)

Needs some 👀  from @DemonHa to make sure I didn't mess anything up.

**Checklist**

---

> [!NOTE]
> Migrates sidebar drag-and-drop to a new `useDraggable` hook, updates sidebar items to bind drag attrs/events directly, and refactors sidebar group/section structure and tests.
> 
> - **Sidebar (drag-and-drop)**:
>   - Introduces `hooks/use-draggable` (`useDraggable`, `DraggingItem`, `HoveredItem`) and exports them via `src/index.ts`.
>   - Replaces `<Draggable>` wrapper with `useDraggable` in `SidebarItem.vue`; binds `draggableAttrs`/`draggableEvents` to `ScalarSidebarItem`, `ScalarSidebarGroup`, and `ScalarSidebarSection`.
>   - `ScalarSidebar.vue`: passes `isDraggable` per layout (`client` true, `reference` false); `isDroppable` typed via hook options; emits `reorder` on drag end.
>   - Removes `@scalar/draggable` dependency and CSS; adds `@scalar/use-hooks`.
> - **Components package (structure/attrs)**:
>   - `ScalarSidebarGroup.vue`: default `is` → `li`; switch to outer `<component>` wrapper with bound classes; emit `click` from group button; children container now `<ul>`.
>   - `ScalarSidebarSection.vue`: default `is` → `li`; wrap with `<component>` and move items into `<ul>`; use `cx` binding.
> - **Tests**:
>   - Update `ScalarSidebar.test.ts` and `SidebarItem.test.ts` for hook-based dragging and `isDraggable` behavior.
>   - Add `hooks/use-draggable.test.ts` to cover hook attrs/events and state.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 8bec6c11c350140cb0645c09457b90048a1b72f8. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/components/src/components/ScalarSidebar/ScalarSidebarGroup.vue
```diff
@@ -58,8 +65,7 @@ const { stylingAttrsCx, otherAttrs } = useBindCx()
         :disabled
         :indent="level"
         :selected
-        v-bind="otherAttrs"
-        @click="!controlled && (model = !model)">
+        @click="($emit('click', $event), !controlled && (model = !model))">
```
DemonHa: can we define the emit and not use vue magic keywords? 🙂

it's kind of confusing for non vue devs

DemonHa: same goes for the $event and any $ keywords

packages/sidebar/src/hooks/use-draggable.ts
```diff
@@ -0,0 +1,253 @@
...
+export type HoveredItem = {
+  id: string
+  parentId: string | null
+  /**
+   * Offset is used when adding back an item, also for the highlight classes
+   * 0 = above      | .dragover-above
+   * 1 = below      | .dragover-below
+   * 2 = as a child | .dragover-asChild
+   */
+  offset: number
```
DemonHa: can we make this an enum

TOP, IN, BOTTOM or something like that?

DemonHa:
```
/**
 * Drag offset constants.
 * These match the draggable component's offset values.
 */
const DRAG_OFFSET = {
  /** Insert before the hovered item */
  BEFORE: 0,
  /** Insert after the hovered item */
  AFTER: 1,
  /** Drop into the hovered item */
  INTO: 2,
} as const
```

this is what I was using to map the values

hwkr: Oh yeah sure, tbh this is mostly copied from the original implement

packages/components/src/components/ScalarSidebar/ScalarSidebarGroup.vue
hwkr: I'm not 100% on this, there's a lot of Vue keywords and at the end of the day I think when we're writing Vue we're going to have to understand how it works.

That being said I did smush a lot of logic in here which should maybe be pulled out 🤔

hanspagel: yep, pull it out 👍 👍

DemonHa: [Reference](https://docs.scalar.com/editor#/engineering/code-styles)

image

PR #7412, opened 2025-11-24
fix(openapi-types): TS auto-complete does not show `$ref` attribute for `OpenAPIV3_1.ReferenceObject`

hanspagel:
**Problem**

The TypeScript auto-complete does not show the `$ref` property for the `OpenAPIV3_1.ReferenceObject`.

This happens because we’re using a `Modify` helper, that seems to be too complex for TS.

**Solution**

This PR removes the `Modify` helper from said type and just adds the `$ref` attribute directly.

**Checklist**

---

> [!NOTE]
> Tightens OpenAPI TypeScript types to only allow `x-` specification extensions and fixes `$ref` visibility on `OpenAPIV3_1.ReferenceObject`, with widespread guards and minor integrations updates across mock server, example generation, markdown rendering, upgrader, and Postman import.
> 
> - **Types (`@scalar/openapi-types`)**:
>   - Enforce specification extensions: replace permissive "any attribute" with `SpecificationExtensions` (`x-*` keys only) across `Document`, `SchemaObject`, `OperationObject`, `ResponseObject`, etc.
>   - Fix `$ref` autocomplete on `OpenAPIV3_1.ReferenceObject`; add type test.
> - **Mock Server**:
>   - Safer schema access with type guards in `create-mock-server`, `mock-any-response`, auth helpers, and OAuth route setup.
>   - Improve header/content handling and example extraction; adjust playground test callback payload.
> - **Example Generation (`oas-utils`)**:
>   - Harden `getExampleFromSchema` with robust checks for `oneOf/anyOf/allOf`, arrays, required detection, omission rules, and XML wrapping.
>   - Ensure parameter example generation casts schemas correctly.
> - **Import/Transform**:
>   - `import-spec`: preserve `callbacks`/`externalDocs` on requests; refine security scheme selection.
>   - Upgrader (2.0→3.0): safer item/parameter transforms and serialization cleanup.
> - **Markdown UI**:
>   - Vue components use `modelValue` and add schema/type guards for safer rendering.
> - **Postman → OpenAPI**:
>   - Extract response headers as `examples` instead of header `schema`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c0e6058f149580c23631af7f0c9872c315ee1017. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/openapi-types/src/openapi-types.ts
+++ b/packages/openapi-types/src/openapi-types.ts
@@ -8,22 +8,17 @@
  */
 
 /** any other attribute, for example x-* extensions */
-type SpecificationExtension = {
-  /** OpenAPI extension */
-  [_specificationExtensionKey: `x-${string}`]: any
+type SpecificationExtensions = {
   /**
-   * Unknown attribute
+   * Specification Extensions
    *
-   * @deprecated We don't actually want to allow any attribute. This will be removed in a future version.
+   * Allows extensions to the OpenAPI Schema. The field name MUST begin with x-, for example, x-internal-id. Field
+   * names beginning x-oai- and x-oas- are reserved for uses defined by the OpenAPI Initiative. The value can be any
+   * valid JSON value (null, a primitive, an array, or an object.)
+   *
+   * @see https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.1.md#specification-extensions
    **/
-  [key: string]: any
-}
-
-// biome-ignore lint/style/noNamespace: We want it to be a module here.
-export namespace Base {
-  export type ReferenceObject = {
-    $ref?: string
-  }
+  [_specificationExtensionsKey: `x-${string}`]: any
 }
 
 // biome-ignore lint/style/noNamespace: We want it to be a module here.
@@ -33,7 +28,7 @@ export namespace OpenAPI {
   // OpenAPI.Document<{
   //   'x-foobar': Foobar
   // }>
-  export type Document<T extends SpecificationExtension = {}> =
+  export type Document<T extends SpecificationExtensions = {}> =
     | OpenAPIV2.Document<T>
     | OpenAPIV3.Document<T>
     | OpenAPIV3_1.Document<T>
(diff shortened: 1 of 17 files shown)


packages/openapi-types/src/openapi-types.ts
```diff
@@ -500,13 +504,11 @@ export namespace OpenAPIV3_1 {
...
   export type XMLObject = OpenAPIV3.XMLObject
 
-  export type ReferenceObject = Modify<
-    OpenAPIV3.ReferenceObject,
-    {
-      summary?: string
-      description?: string
-    }
-  >
+  export type ReferenceObject = {
```
DemonHa: The real problem here is that the `OpenAPIV3.ReferenceObject` contains the `SpecificationExtension` which has the generic string to unknown mapping type.

One way to solve this is to remove that generic and only let the strings starting with `x-` or we can remove the `& SpecificationExtension` on the base type and add it for each of the versions like this:

```
 Modify<
    OpenAPIV3.ReferenceObject,
    {
      summary?: string
      description?: string
    }
  > & SpecificationExtension
```

hanspagel: Yeah, I wanted to remove the generic string (hence the rename), but that broke so many things. :(

hanspagel: > we can remove the & SpecificationExtension on the base type and add it for each of the versions like this

I did this

hanspagel:
Cursor is so right, but for reasons it was optional for the other version already.

We need to make it required for all versions. Unfortunately, we have to do this in another PR I’m afraid.

Or I make this a bigger PR here?

EDIT: I removed the generic string attributes and added a minor changeset. 👀  

I added a few ugly checks to other packages to narrow down the type. In most cases it’s a) code that’s going away soon anyway or b) something where the data ingress is too naive (and not type-safe yet).

hanspagel:
I think I should work on a proper new major release instead. 😂

PR #7414, opened 2025-11-24
feat(api-reference): show tags without operations

### LINKED ISSUE #7218: show tags even when they don’t have endpoints
### Discussed in https://github.com/scalar/scalar/discussions/6978

<sup>Originally posted by **mderazon** September 30, 2025</sup>
Hey !
I have an introduction tag / group that is supposed to have all kind of documents on how to use the api without any specific endpoint.
For example, things like errors, rate limits, authentication scheme, pagination scheme, versioning etc.

These docs are plain markdown but they don't have any endpoint associated with them but I still would like to show them in the side bar separately and not under one "Introduction" section.

I tried having them as tags without any endpoints, but then it seems that Scalar hides them. 

Is there any way to achieve this ?
Thanks!</div>

hanspagel:
**Problem**

> I have an introduction tag / group that is supposed to have all kind of documents on how to use the api without any specific endpoint. For example, things like errors, rate limits, authentication scheme, pagination scheme, versioning etc.
>
> #7218

I think that's a great way to "misuse" tags! Using them to add more content to your reference.

Unfortunately, we hide them when they don't have any operations attached.

**Solution**

This PR has a small, but _**breaking change**_ (kind of) and shows tags, even when they don’t have any operations.

If users wants to hide a tag, they can still use `x-scalar-ignore`.

Fixes #7218

**Preview**

Screenshot 2025-11-24 at 14 26 34

**Checklist**

### DIFF
--- a/.changeset/wild-eels-act.md
+++ b/.changeset/wild-eels-act.md
@@ -1,0 +1,6 @@
+---
+'@scalar/workspace-store': patch
+'@scalar/api-reference': patch
+---
+
+feat: show tags without operations


geoffgscott:
Would doing the inverse be better? Requiring x-scalar-show-empty or similar on a tag? 

What is the intent here? Just to have a DOM selector for the sidebar link?

DemonHa:
> Would doing the inverse be better? Requiring x-scalar-show-empty or similar on a tag?
> 
> What is the intent here? Just to have a DOM selector for the sidebar link?

I need this PR for the new client, users should be able to create tags and watch them on the sidebar even if they are empty initially.

We can also do conditional rendering if we don't want to show them by default on the refrerences

DemonHa:
Can you add another test to see if the x-scalar-ignore would work here?

hanspagel:
@geoffgscott the use case:

> I have an introduction tag / group that is supposed to have all kind of documents on how to use the api without any specific endpoint. For example, things like errors, rate limits, authentication scheme, pagination scheme, versioning etc.

Source: https://github.com/scalar/scalar/discussions/6978

We have `x-scalar-ignore` already, I think that’s fine?

PR #7416, opened 2025-11-25
feat(api-client): allow updating method while maintaining order

amritk:
**Problem**

When you updated the method, you would get a 404 + the sidebar would re-order

**Solution**

With this PR we maintain the sidebar order + redirect you right after its complete. 

**Checklist**

---

> [!NOTE]
> Allow updating an operation’s HTTP method with preserved sidebar order and automatic redirect; add prototype-pollution guard and new navigation helpers.
> 
> - **API Client (v2)**
>   - Include `exampleKey` in `operation:update:method` meta; emit updated events from `OperationBlock`.
>   - On method change, redirect to the updated example via router in `useWorkspaceClientEvents`.
>   - Initialize first server on mount in `Operation.vue` if none selected.
>   - Minor template prop shorthand in `App.vue`.
> - **Workspace Store**
>   - New `updateOperationMethod` behavior: move operation to new method, update `x-scalar-order` across parents, rebuild sidebar, optional success callback.
>   - Add `getDocumentConfiguration()` API and use it for ID generation.
>   - Always refresh `x-scalar-order` in `traverseDocument`.
>   - Server selection toggle now sets empty string when unselecting.
>   - Protect updates with `preventPollution` in workspace `update`/`updateDocument` and when changing paths.
> - **Navigation Helpers**
>   - Add `getOpenapiObject`, `getOperationEntries`, and `getParentEntry` under `@scalar/workspace-store/navigation` and export them; update consumers (drag-handle, sidebar state).
> - **Helpers**
>   - New `preventPollution` utility with tests.
> - **Sidebar**
>   - Stop exporting `getParentEntry` from `@scalar/sidebar`.
> - **Tests**
>   - Extensive updates/additions validating `x-scalar-order`, navigation rebuilds, method/path updates, server selection behavior, and new helpers.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 30c3910db6902d6a46c8f2449c48a98be3571739. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/workspace-store/src/mutators/operation.ts
+++ b/packages/workspace-store/src/mutators/operation.ts
@@ -252,11 +252,14 @@ export const updateOperationMethod = (
 
     const order = parentOpenAPIObject['x-scalar-order']
     const index = order?.indexOf(entry.id)
-    if (!Array.isArray(order) || typeof index !== 'number' || index < 0 || !('name' in parentOpenAPIObject)) {
+    if (!Array.isArray(order) || typeof index !== 'number' || index < 0) {
       return
     }
 
-    const parentTag = entry.parent.type === 'tag' ? { tag: parentOpenAPIObject, id: entry.parent.id } : undefined
+    const parentTag =
+      entry.parent.type === 'tag' && 'name' in parentOpenAPIObject
+        ? { tag: parentOpenAPIObject, id: entry.parent.id }
+        : undefined
 
     // Generate the new ID based on whether this is an operation or webhook
     order[index] = generateId({
--- a/packages/workspace-store/src/mutators/operation.test.ts
+++ b/packages/workspace-store/src/mutators/operation.test.ts
@@ -98,6 +98,7 @@ describe('updateOperationMethod', () => {
     })
     store.buildSidebar('test')
     const document = store.workspace.documents.test!
+    expect(document['x-scalar-order']).toStrictEqual(['test/get/users', 'test/post/users'])
 
     updateOperationMethod(document, store, {
       meta: { method: 'get', path: '/users', exampleKey: 'default' },


packages/api-client/src/v2/blocks/operation-block/OperationBlock.vue
```diff
@@ -117,6 +120,19 @@ const handleUpdateMethod = (payload: { value: HttpMethodType }) =>
     payload: {
       method: payload.value,
     },
+    /** We need to redirect to the new path */
```
amritk: this callback should be in the hook actually, where the event bus is handled

packages/workspace-store/src/navigation/helpers/traverse-document.ts
```diff
@@ -121,11 +121,11 @@ export const traverseDocument = (documentName: string, document: OpenApiDocument
       const safeIndexB = indexB === -1 ? Number.POSITIVE_INFINITY : indexB
       return safeIndexA - safeIndexB
     })
-
-    // Now update the sort order of the entries
-    document['x-scalar-order'] = unpackProxyObject(entries.map((entry) => entry.id))
   }
 
+  // Now update the sort order of the entries
```
amritk, line 126: @DemonHa this was inside the IF before

packages/workspace-store/src/events/definitions/operation.ts
```diff
@@ -44,7 +44,10 @@ export type OperationEvents = {
       method: HttpMethod
     }
     /** Identifies the target operation by original method and path */
-    meta: OperationMeta
+    meta: OperationMeta & {
```
DemonHa: you can use `OperationExampleMeta` which inclides the example

packages/helpers/src/object/prevent-pollution.ts
DemonHa, line 1: nice, this shouldn't really be much of a problem in our context, but there might be cases where you try to load a malicious openapi document

packages/workspace-store/src/mutators/operation.ts
```diff
@@ -183,39 +186,109 @@ export const updateOperationSummary = (
...
+    }
+
+    const order = parentOpenAPIObject['x-scalar-order']
+    const index = order?.indexOf(entry.id)
+    if (!Array.isArray(order) || typeof index !== 'number' || index < 0) {
+      return
+    }
+
+    const parentTag =
+      entry.parent.type === 'tag' ? { tag: parentOpenAPIObject as TagObject, id: entry.parent.id } : undefined
```
DemonHa: can we avoid casting here?

packages/workspace-store/src/navigation/helpers/traverse-document.ts
DemonHa: yeah I think it's better to get this outside

the idea of this was to only set the order when an order is triggered manually but yeah this is fine as well

packages/workspace-store/src/client.ts
DemonHa, line 987: can you use `preventPollution` also here?

DemonHa (approved):
Great work 🔥, left some non-blocking comments.

packages/workspace-store/src/navigation/helpers/traverse-document.ts
amritk: yea but then the order was never set!

packages/helpers/src/object/prevent-pollution.ts
amritk: yea I know, I thought about not adding it, but it really doesn't hurt just extra protection

packages/workspace-store/src/mutators/operation.ts
amritk: we know its a tag because the entry is a tag, but sure we can add another type guard

packages/workspace-store/src/navigation/helpers/traverse-document.ts
DemonHa: it is set when you try to order them

DemonHa: with this change it will set the order immediately when you load the document and generate the sidebar. But that doesn't hurt

packages/workspace-store/src/mutators/operation.ts
DemonHa: I think you can use the `getParentEntry` helper here it will get the closes parent for the specified type for the entry you provide

packages/workspace-store/src/navigation/helpers/traverse-document.ts
amritk: yea but I didn't set the order, so I didn't have access to it haha. Yea doesn't matter I could have just generated it if it didn't exist 🤷🏾

packages/workspace-store/src/mutators/operation.ts
amritk: I don't need the parent, I need from entry -> openapi object. Due to the narrowing above its typed as 
image
but we know its a tag because of the entry: `entry.parent.type === 'tag'`. 

one more type guard wont' really hurt though

amritk: done

DemonHa: I mean yeah but the using the helper might be better cuz it will traverse up the parent and fine the closes type that you need. For now it's not required cuz operation parent can be either tag or document, but yeah might be easy for the future

amritk: oh you mean instead of just hitting .parent? We only need to go one level up so its ok here!

PR #7421, opened 2025-11-26
ci: dedicated sticky disks for branches

xC0dex:
**Problem**

Quote from the blacksmith support:

> SDs also have last writer wins semantics which means if there are 5 concurrent jobs all writing to the same sticky disk key, only the last commit will be visible to future jobs.

This means having one sticky disk for all branches and workflows can cause concurrency problems.

**Solution**

This PR uses the `ref_name` as part of the sticky disk key. So each branch has its own one.
Since the sticky disk for our main branch now only contains artifacts from our main, I guess it's enough to reset it every week. Stale sticky disks (from our branches) are automagically removed after 7 days, so we don't have to care about them.

I also moved the node, pnpm and stickydisk setup to a dedicated composite action `setup-node`. Now we don't have to run our `build` action as part of our `types` job, which reduces the execution time of the job.
I believe we can use this action in even more places, but thats something for another PR.

**Checklist**

---

> [!NOTE]
> Introduces a reusable `setup-node` composite action and switches turbo sticky disk caching to branch- and Node-version–scoped keys; updates CI to use the new action and adjusts cleanup/schedules.
> 
> - **CI/Actions**:
>   - **New composite action**: Adds `./.github/actions/setup-node` to install `pnpm`, resolve Node version, setup `actions/setup-node`, and configure stickydisk with key `repo-ref_name-turbo-cache-nodeVersion`.
>   - **Refactor build flow**:
>     - `./.github/actions/build` now uses `setup-node` and accepts `node-version`.
>     - `types` job switches from `actions/build` to `setup-node` + `pnpm install` before `types:check`.
>   - **Caching changes**:
>     - Stickydisk keys updated to include `github.ref_name` and Node version in both setup and cleanup (`stickydisk-delete`).
>     - Removes `TURBO_CACHE_DIR` env from `ci.yml`.
>   - **Scheduling**:
>     - `blacksmith-cleanup.yml` cron moved to Sundays 20:00 UTC.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 34adc9d3442192c6391019902235bc8ee1800def. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/.github/actions/setup-node/action.yml
+++ b/.github/actions/setup-node/action.yml
@@ -27,5 +27,5 @@ runs:
       name: Setup persistent turbo cache
       uses: useblacksmith/stickydisk@a652394bf1bf95399f406e648482b41fbd25c51f # v1
       with:
-        key: ${{ github.repository }}-turbo-cache-${{ steps.node-version.outputs.version }}
+        key: ${{ github.repository }}-${{ github.ref_name }}-turbo-cache-${{ steps.node-version.outputs.version }}
         path: .turbo
--- a/.github/workflows/blacksmith-cleanup.yml
+++ b/.github/workflows/blacksmith-cleanup.yml
@@ -17,4 +17,4 @@ jobs:
       - name: Attach sticky disk
         uses: useblacksmith/stickydisk-delete@b41313d28b8647d72114c9ba3c96bb04061562b6 # v1
         with:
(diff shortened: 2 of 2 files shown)


hanspagel (approved):
sick work
let’s try this

thanks for dealing with the blacksmith support

PR #7428, opened 2025-11-26
refactor(biome): disable `noBarrelFile` for index files

marcalexiei:
## Problem

`biome` reports warning for `noBarrelFile` rule on `index` files.

## Solution

Since the purpose of these files conflicts with the purpose of `noBarrelFile` rule,
I turned off this rule for all index files. For other files the rule is set to `error`.

## Checklist

- [x] I added a changeset (Not needed).
- [x] I added tests (Not needed).

### DIFF
--- a/biome.json
+++ b/biome.json
@@ -194,7 +194,7 @@
       "performance": {
         "noDelete": "off",
         "noReExportAll": "warn",
-        "noBarrelFile": "warn"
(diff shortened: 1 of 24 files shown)


hanspagel (approved):
had this thought yesterday, and you did it already :D

PR #7433, opened 2025-11-27
feat(aspire): support .NET 10

xC0dex:
## Problem

Currently, we don't use the latest and greatest .NET 10 features in our Aspire integration.

## Solution

Now we do! 🙂

## Checklist

- [x] I added a changeset. 
- [ ] I added tests.


hanspagel (approved):
👍

PR #7451, opened 2025-12-01
ci: skip deploy-api-reference job for main branch

xC0dex:
## Problem

I believe there is no use case why we run `deploy-api-reference` in our main branch.

I thought it was the same as the `deploy-components` job, but `deploy-components` distinguishes between prod (main) and preview (PRs).

## Solution

I updated the condition of the job to only run for PRs.

## Checklist

- [ ] I added a changeset. 
- [ ] I added tests.


hanspagel (approved):
yep

PR #7459, opened 2025-12-02
fix(api-reference): `onDocumentSelect` is not called

### LINKED ISSUE #7452: `onDocumentSelect` hook not firing
### What happens?

i'm using the AspNetCore integration of scalar (version 2.11.0) and when using the following custom javascript configuration
```js
export default {
    onLoaded: () => console.log("AAAA"),
    onDocumentSelect: () => console.log("BBBBB")
}
```
the `onDocumentSelect` hook is never triggered

The `onDocumentSelect` hook is triggered and every time I select a different document "BBBBB" is printed to the console.

### OpenAPI Document

Not relevant

marcalexiei:
## Problem

- Closes #7452

## Solution

Here’s a clearer, more polished version:

- Invoke `onDocumentSelect` inside the `changeSelectedDocument` handler
- `await` both the `onDocumentSelect` and `onLoaded` hooks (their schemas allow them to return a `Promise`).
   https://github.com/scalar/scalar/blob/61c980cebef219685a05f3a4cb1e379984922cd4/packages/types/src/api-reference/api-reference-configuration.ts#L135-L142 
   https://github.com/scalar/scalar/blob/61c980cebef219685a05f3a4cb1e379984922cd4/packages/types/src/api-reference/api-reference-configuration.ts#L144
- Update the documentation for the `onDocumentSelect` and `onLoaded` hooks to indicate that they may return a `Promise`.

## Checklist

- [x] I added a changeset. 
- [x] I added tests.

### DIFF
--- a/packages/api-reference/src/components/ApiReference.vue
+++ b/packages/api-reference/src/components/ApiReference.vue
@@ -446,7 +446,7 @@ const changeSelectedDocument = async (
     ...configurationOverrides.value,
   }
 
-  // Store `onDocumentSelect` await its execution later before calling `onLoaded`
+  // Store `onDocumentSelect` result to await its execution later, before calling `onLoaded`
   const onDocumentSelectPromise = config.onDocumentSelect?.()
 
   // Set the active slug and update any routing


packages/api-reference/src/components/ApiReference.vue
```diff
@@ -470,7 +472,8 @@ const changeSelectedDocument = async (
             config: mapConfiguration(config),
           },
     )
-    config.onLoaded?.(slug)
+
+    await config.onLoaded?.(slug)
```
hanspagel: the await makes those calls block the further execution, right? is that really what we want here? I think they should be executed asynchronously as long was we don’t need the result of the function.

hanspagel (changes requested):

packages/api-reference/src/components/ApiReference.vue
marcalexiei: I refined the logic so both hooks invoke their handlers without blocking. 
I also ensured that `onLoaded` is triggered only after `onDocumentSelect` has finished (added a new test case).

https://github.com/scalar/scalar/pull/7459/commits/d8720297543c9fc826d4d0caff997e07eb5b8c33

PR #7460, opened 2025-12-03
feat(components): update and polish ScalarLoading

hwkr:
Updates `ScalarLoading` to make things clearer (and also work in general). Important changes:

* Completion functions → `validate` and `invalidate` now clear the loader by default (before you had to set clear in the args)
* Added an `isActive` state to track if the loader should be displayed but is actually loading (e.g. animating or showing the checkmark)
* Standardized the loading state prop to be `loader` (`loading` made it sound like it should be a boolean not an object)
* Broke apart the Vue file into tested utility files to match the component library standards
* Added tests
* Added a bunch of docs
* Fixed a bunch of floating promises

## Checklist

- [x] I added a changeset. 
- [x] I added tests.

### DIFF
--- a/packages/api-client/src/v2/features/command-palette/components/CommandPaletteImport.vue
+++ b/packages/api-client/src/v2/features/command-palette/components/CommandPaletteImport.vue
@@ -73,9 +73,6 @@ const MAX_NAME_RETRIES = 100
 /** Default document name when none can be extracted */
 const DEFAULT_DOCUMENT_NAME = 'document'
 
-/** Duration to show error state in milliseconds */
-const ERROR_DISPLAY_DURATION = 2000
-
 const router = useRouter()
 const loader = useLoadingState()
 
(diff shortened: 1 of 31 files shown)


packages/api-client/src/v2/features/command-palette/components/CommandPaletteImport.vue
```diff
@@ -269,7 +266,7 @@ const handleImport = async (): Promise<void> => {
     await loader.clear()
     navigateToDocument(result.name)
   } else {
-    await loader.invalidate(ERROR_DISPLAY_DURATION, true)
```
amritk, line 272: did we lose some data here?

hwkr: This was using an error display duration of 2000ms and I put it back to the default of 1100ms. If you think it should be 2000 I can put it back but I pretty sure I added the 2000 and it probably should have been the standard 1100 to start with.

amritk: nah thats fine long as it was on purpose

PR #7465, opened 2025-12-03
fix(workspace-store): ensure consumers rely only on public exports (and enable `knip`)

marcalexiei:
## Problem

- Followup of #7233
- Superseds #7462 

## Solution

I noticed that workspace-store consumers were using non public exports.
Probably no error was reported since TypeScript `moduleResolution` is set to `bundler`.

This PR ensures that consumers only use defined exports.
Also all `knip` reported issues for this package are now fixed.

## Checklist

- [x] I added a changeset. 
- [x] I added tests.


hanspagel (approved):
cool

PR #7467, opened 2025-12-03
feat(api-client): handle updating path with conflict handling

amritk:
## Problem

Finish up updating the path without changing the sidebar order

We show a nice error message now when we have a conflicting path + method and do not update our store. Also had to do some focus hacking since we route after creating new sidebar entries.

image

## Checklist

- [x] I added a changeset. 
- [ ] I added tests.

---

> [!NOTE]
> Unifies path/method updates into a single event with conflict handling, updates UI to surface conflicts, and routes + preserves sidebar order on success.
> 
> - **API Client (v2)**
>   - **Address Bar**: Emits `operation:update:pathMethod` with callback; shows conflict UI (warning + outline); debounced path updates; supports focusing address bar; uses conflict refs to reflect pending changes; passes through servers. Adds `alwaysEmitChange` usage and improved focus behavior.
>   - **Header/OperationBlock**: Removes legacy `update:method`/`update:path` re-emits; now only handles `execute` and `update:servers`.
>   - **CodeInput**: Adds `alwaysEmitChange` prop and enhanced `focus(cursorAtEnd)` API.
>   - **Storage**: Prevents storing `/` as current path (avoids redirect loop).
> - **Workspace Store**
>   - **Events**: Replaces `operation:update:method`/`operation:update:path` with unified `operation:update:pathMethod` (with `callback` and statuses: `success|no-change|conflict`); loosens `ui:focus:address-bar` payload to be optional.
>   - **Mutators**: Implements `updateOperationPathMethod` to move operations across path/method, detect conflicts, preserve sidebar order IDs, and sync path parameters; extracts operation entries map; updates exports.
>   - **Navigation**: Exposes `getOperationEntries` and its map type.
> - **Routing/Sidebar**
>   - On successful update, router redirects to new path/method and sidebar is rebuilt to reflect order.
> - **Helpers**
>   - `generate-hash` docs clarified to 64‑bit output.
> - **Tests**
>   - Extensive updates/added tests across client and store for event, routing, conflict UI, parameter syncing, and order preservation.
> - **Changeset**
>   - Patch bumps for `@scalar/workspace-store`, `@scalar/api-client`, `@scalar/helpers`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit a4074b4a4b5bd33481a0fc5287dc7e530595e1d3. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/api-client/src/v2/hooks/use-workspace-client-events.ts
+++ b/packages/api-client/src/v2/hooks/use-workspace-client-events.ts
@@ -77,7 +77,6 @@ export const useWorkspaceClientEvents = ({
   sidebarState: UseSidebarStateReturn
 }) => {
   /** Use route[r] for some redirect business */
-  const route = useRoute()
   const router = useRouter()
 
   /** Use route for the path variables */
@@ -122,8 +121,10 @@ export const useWorkspaceClientEvents = ({
    *
    * @param documentName - The name (id) of the document for which to rebuild the sidebar
    */
-  const rebuildSidebar = (documentName: string) => {
-    workspaceStore.value?.buildSidebar(documentName)
+  const rebuildSidebar = (documentName: string | undefined) => {
+    if (documentName) {
+      workspaceStore.value?.buildSidebar(documentName)
+    }
   }
 
   /**
@@ -293,7 +294,7 @@ export const useWorkspaceClientEvents = ({
         })
 
         // Rebuild the sidebar with the updated order
-        workspaceStore.value?.buildSidebar(document.value?.['x-scalar-navigation']?.id ?? '')
+        rebuildSidebar(document.value?.['x-scalar-navigation']?.id)
       }
       payload.callback(status)
     }),


packages/api-client/src/v2/blocks/operation-block/OperationBlock.vue
```diff
@@ -120,6 +126,7 @@ const handleUpdateMethod = (payload: { value: HttpMethodType }) =>
     payload: {
       method: payload.value,
     },
+    operationEntriesMap,
```
DemonHa: do we need to send the map on entries?

packages/api-client/src/v2/hooks/use-workspace-client-events.ts
```diff
@@ -237,7 +237,24 @@ export const useWorkspaceClientEvents = ({
...
+          name: 'example',
+          params: {
+            method: payload.meta.method,
+            pathEncoded: encodeURIComponent(payload.payload.path),
+            exampleName: payload.meta.exampleKey,
+          },
+        })
+
+        // We want to re-focus the addressbar after routing (nextTick doesn't work here)
+        setTimeout(() => eventBus.emit('ui:focus:address-bar'), 0)
```
DemonHa: does this always work?

DemonHa: I think you can just do `await router.replace` to wait for the navigation to happen. Not sure if that works tho

packages/workspace-store/src/events/definitions/operation.ts
```diff
@@ -65,6 +66,8 @@ export type OperationEvents = {
     }
     /** Identifies the target operation by original method and path */
     meta: OperationExampleMeta
+    /** Map of operation entries */
+    operationEntriesMap: OperationEntriesMap
```
DemonHa: Not sure if this is a good idea to pass the map on the event. Can we just rebuild it on the other side?

packages/workspace-store/src/mutators/operation.ts
```diff
@@ -335,13 +357,9 @@ export const updateOperationMethod = (
...
 
-  // Rebuild the sidebar with the updated order (if store is available)
-  if (store) {
-    const success = store.buildSidebar(documentName)
-    callback?.(success)
-  } else {
-    callback?.(true)
-  }
+  // Rebuild the sidebar with the updated order
+  const success = store.buildSidebar(documentName)
```
DemonHa: This is fine, but we need to choose how do we want to handle the sidebar rebuild, here or when we handle the event

packages/workspace-store/src/mutators/operation.ts
```diff
@@ -363,19 +381,18 @@ export const updateOperationMethod = (
  */
 export const updateOperationPath = (
```
DemonHa: I believe it's better if the whole uniqueness validation happens on the mutator instead on the other side

packages/api-client/src/v2/hooks/use-workspace-client-events.ts
amritk: haha yea this was the hack I was telling you, the only other way I can think of is add a query to the route to get it to focus on mount but I think this is a bit faster

packages/workspace-store/src/events/definitions/operation.ts
amritk: no need to build it again if we already have it

DemonHa: yeah the problem with this is that the document is not the source of truth anymore

amritk:
This is getting a re-work

PR #7480, opened 2025-12-05
ci: update Maven deploy command to include all dependencies

xC0dex:
## Problem

Currently, our `java-publish` job fails 🤯 

## Solution

Passing the `-am` argument should solve the issue. It tells Maven to include the parent `pom.xml`. 

## Checklist

- [ ] I added a changeset. 
- [ ] I added tests.


hanspagel (approved):
let’s try

PR #7489, opened 2025-12-06
feat(api-client): auth + params for building requests

amritk:
## Problem

Started the new request building for client v2. A cleaned up version of what we had with more tests + some style support.

These are currently only testable through tests as they aren't in the UI quite yet.

## Checklist

- [x] I added a changeset. 
- [x] I added tests.

---

> [!NOTE]
> Adds client v2 helpers to build request params and security with env/path variable replacement, content decoding, and cookie schema update, all covered by tests.
> 
> - **api-client (v2)**:
>   - **Request building**: Add `build-request-parameters` and `build-request-security` helpers to assemble headers, query params, cookies, and path variables; supports OpenAPI styles, env var substitution, API key/basic/bearer/OAuth2, and cookie coercion; extensive tests.
>   - **Utilities**: Add `get-example`, `get-delimiter`, and `decode-buffer` (text/binary handling) with tests.
> - **helpers**:
>   - **Regex**: Introduce `replaceEnvVariables` and `replacePathVariables` in `@scalar/helpers/regex/replace-variables` with tests; mark legacy template replacement in `string-template` as deprecated and adjust import usage.
> - **workspace-store**:
>   - **Schema**: Extend `x-scalar-cookies` with optional `path`; leveraged by request builders.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 16729809477ff517a995a512ec497afe82abd6d9. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/api-client/src/v2/blocks/operation-block/helpers/build-request-parameters.ts
+++ b/packages/api-client/src/v2/blocks/operation-block/helpers/build-request-parameters.ts
@@ -1,4 +1,4 @@
-import { replaceEnvVariables, replaceVariables } from '@scalar/helpers/regex/replace-variables'
+import { replaceEnvVariables } from '@scalar/helpers/regex/replace-variables'
 import {
   type XScalarCookie,
   xScalarCookieSchema,


packages/helpers/src/regex/replace-variables.ts
```diff
@@ -19,3 +21,11 @@ export function replaceVariables(
...
   return value.replace(doubleCurlyBrackets, callback).replace(singleCurlyBrackets, callback)
 }
+
+/** Replace {path} variables with their values */
+export const replacePathVariables = (path: string, variables: Record<string, string> = {}) =>
+  path.replace(REGEX.PATH, (match, key) => variables[key] || match)
+
+/** Replace {{env}} variables with their values */
+export const replaceEnvVariables = (path: string, variables: Record<string, string> = {}) =>
+  path.replace(REGEX.VARIABLES, (match, key) => variables[key] || match)
```
DemonHa: can we use the nullish operator here? `??`

amritk: you are correct! Nullish will keep the empty string

DemonHa: also the number 0

DemonHa: also for the other function on top of this, the same thing

amritk: yea anything falsey but I think our env vars are only strings

PR #7491, opened 2025-12-08
test: use `vitest` mock utils instead of manual mock

marcalexiei:
## Problem

I noticed that many tests are mocking global properties manually (store value and restore it manually at the end of the test)

## Solution

Use `vitest` mock utils (`stubGlobal`, `unstubAllGlobals`, spies...) to handle global variables mock.

## Checklist

- [x] I added a changeset (Not needed). 
- [x] I added tests.


hanspagel:
lol, you’re working through my infinite “someday” todo list

PR #7493, opened 2025-12-08
fix(openapi-types): `SchemaObject` merges boolean with `AnyOtherAttribute`

### LINKED ISSUE #7490: OpenAPIV3_1.SchemaObject allows for `boolean & AnyOtherAttribute` which isn't possible
### What happens?

OpenAPIV3_1.SchemaObject is defined as<br>[https://github.com/scalar/scalar/blob/9a8ab3ef39ca015d194d6505cae684f9aaddf12f/packages/openapi-types/src/openapi-types.ts#L186-L187](<https://github.com/scalar/scalar/blob/9a8ab3ef39ca015d194d6505cae684f9aaddf12f/packages/openapi-types/src/openapi-types.ts#L186-L187>)<br>That allows for `boolean & AnyOtherAttribute` which is invalid. If something is a `boolean` it cannot also be an `object`. Besides this not being possible in TypeScript it also doesn't work in the spec.

```yaml
components:
  schemas:
    Foo: true    # no room to put AnyOtherAttribute
    Bar: false   # no room to put AnyOtherAttribute
    Baz: {}
```

I think it should be something like this

```ts
export type SchemaObject =
  | boolean
  | ((ArraySchemaObject | NonArraySchemaObject | MixedSchemaObject) &
      AnyOtherAttribute)
```

### OpenAPI Document

[https://sandbox.scalar.com/e/l14l9](<https://sandbox.scalar.com/e/l14l9>)

hanspagel:
## Problem

Currently, we merge the `AnyOtherAttribute` with the `SchemaObject` union, that has boolean. This will always make it an object.

You might ask how schema can be a boolean:

> The Schema Object allows the definition of input and output data types. These types can be objects, but also primitives and arrays. This object is a superset of the [JSON Schema Specification Draft 2020-12](https://tools.ietf.org/html/draft-bhutton-json-schema-00). The empty schema (which allows any instance to validate) MAY be represented by the boolean value true and a schema which allows no instance to validate MAY be represented by the boolean value false.

## Solution

This PR removes the `AnyOtherAttribute` from the `SchemaObject` (we want to get rid of any *any*way).

Fixes #7490

## Checklist

- [x] I added a changeset. 
- [ ] I added tests.

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) is generating a summary for commit c653005723665bf934a0c79f491dc211e8c5c5f1. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/openapi-types/src/openapi-types.ts
+++ b/packages/openapi-types/src/openapi-types.ts
@@ -17,7 +17,7 @@ type AnyOtherAttribute = {
  * OpenAPI Specification Extension
  **/
 type SpecificationExtension = {
-  [customExtension: `x-${string}`]: any
+  [customExtension: `x-${string}`]: unknown
 }
 
 // biome-ignore lint/style/noNamespace: We want it to be a module here.


hanspagel:
too much quirky workarounds, I just need to do a proper upgrade to the types package

PR #7506, opened 2025-12-09
fix: use caret version for `zod` and `yaml`

marcalexiei:
## Problem

The versions of `zod` and `yaml` are currently pinned.  
If a consuming project uses different versions of these dependencies, package deduplication cannot occur.

## Solution

Use caret-based version ranges for `zod` and `yaml` to allow greater compatibility and enable deduplication.

## Checklist

- [x] I added a changeset. 
- [x] I added tests.

### DIFF
--- a/.changeset/khaki-ears-collect.md
+++ b/.changeset/khaki-ears-collect.md
@@ -1,0 +1,11 @@
+---
+'@scalar/fastify-api-reference': patch
+'@scalar/api-client': patch
+'@scalar/import': patch
+'@scalar/json-magic': patch
+'@scalar/oas-utils': patch
+'@scalar/openapi-parser': patch
+'@scalar/workspace-store': patch
+---
+
+fix: use caret version for `yaml`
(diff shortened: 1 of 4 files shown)


amritk (approved):
So these days this could open you up to npm attacks. But I think we have the 1 week old rule in pnpm so we should be safe. Nice work.

marcalexiei:
> But I think we have the 1 week old rule in pnpm so we should be safe.

yes, `minimumReleaseAge` is set to 1 week:
https://github.com/scalar/scalar/blob/5df94bf2ac71ae807631c73bf437b96e144df44c/pnpm-workspace.yaml#L13-L14

---

> So these days this could open you up to npm attacks.

Adding my personal view about this: pinning an exact version in `dependencies` can cause users to end up with two different copies of the same module (one from the pinned dep and one from another consumer), which may lead to subtle runtime issues (for example, `instanceof` checks failing), which is the primary reason why I opened the PR. 

Also, if a pinned version is later found vulnerable, typically you have to cut a new release to roll in an updated dependency, or users must override/force an upgrade locally.

IMHO security issues can be mitigated by settings like `minimumReleaseAge` and lock files rather than version ranges 😅

PR #7509, opened 2025-12-10
feat(api-client): v2 client app modal version

DemonHa:
## Implement Modal Version for API Client v2

**Summary**

This PR introduces the modal-based workflow for API Client v2, with several UI and UX improvements:

#### Key Changes
**📱 Mobile-Responsive Sidebar**

The sidebar now adapts properly to mobile layouts.
Ensures smooth navigation and access to documents on smaller screens.

**📄 Single Document Sidebar**

The sidebar now displays context for only one document at a time, improving clarity and reducing clutter.
Helps users stay focused on the document they’re currently editing or viewing.

**🔍 Single Document Search**

Search behavior is now scoped to the currently selected document.
Faster, more relevant search results with reduced noise.

**🔒 Modal With Readonly Fields**
The modal now supports readonly fields, except for:

Body
Parameters

These remain editable to allow users to quickly test or modify request payloads.

image

image

## Checklist

- [ ] I added a changeset. 
- [ ] I added tests.

---

> [!NOTE]
> Adds a full modal version of the API Client v2 with dedicated sidebar/navigation, route resolution, scroll-lock, and event wiring, plus refactors and tests across app, sidebar, and operation header.
> 
> - **Modal Client (v2)**:
>   - New `Modal.vue` with focus trap, scroll lock, sidebar toggle, and operation rendering (layout=`modal`).
>   - Factory `create-api-client-modal` with in-app routing via `resolve-route-parameters` and `use-modal-sidebar`.
>   - Modal-specific event handlers `use-workspace-client-modal-events` and environment computation via `getActiveEnvironment`.
> - **Sidebar**:
>   - `@scalar/sidebar` integration updates: layout-aware filtering (`filter-items`), draggable control, and tests.
>   - App client adopts `useAppSidebar`; modal uses `useModalSidebar` for single-document navigation.
> - **Header/Operation/Address Bar**:
>   - Header close emits `ui:close:client-modal`; removes sidebar props; passes `documentUrl` to client button.
>   - AddressBar/CodeInput gain `layout` awareness; modal disables edits appropriately.
> - **App Hooks/Refactors**:
>   - Replace legacy hooks with `useAppSidebar`, `useWorkspaceClientAppEvents`, `useTabs` updates; extract `getActiveEnvironment`.
>   - Sync-path, workspace selector, and tabs integrate new sidebar lookup types.
> - **Events**:
>   - Normalize modal close to `ui:close:client-modal` and update UI events typing.
> - **Playground**:
>   - Demo opens modal and supports opening specific operation via button.
> - **Tests**:
>   - Comprehensive tests for modal, route resolution, modal/sidebar hooks, scroll-lock, sidebar filtering, tabs, and document watcher.
> - **Styles/Components**:
>   - Minor layout/style tweaks (modal/container, header paddings) and accessibility (aria labels, Esc handling in search).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 6bd8aa4d5e7c1b6b377a13483a38d375c74b5af7. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/api-client/src/vite-env.d.ts
+++ b/packages/api-client/src/vite-env.d.ts
@@ -1,9 +1,9 @@
 import type { WorkspaceStore } from '@scalar/workspace-store/client'
 
-/** The version number taken from the package.json */
-declare const PACKAGE_VERSION: string
-
 declare global {
+  /** The version number taken from the package.json */
+  const PACKAGE_VERSION: string
+
   interface Window {
     dataDumpWorkspace: () => WorkspaceStore | null
   }


packages/api-client/src/v2/features/app/App.vue
```diff
@@ -106,9 +102,8 @@ const { store, workspaces, activeWorkspace, setWorkspaceId, createWorkspace } =
 useColorMode({ workspaceStore: store })
 
 /** Sidebar state and selection handling. */
-const sidebarState = useSidebarState({
+const sidebarState = useAppSidebar({
```
geoffgscott, line 104: Lets please call this initializeAppSidebar and put it into the helpers folder.

This prevents people from trying to consume this outside of the top level.

packages/api-client/src/v2/features/app/components/AppSidebar.vue
hwkr, line 1: I think if we're going to use the modal version of the search we should make it open when you click the icon and remove the other search button.

Image

packages/api-client/src/v2/features/modal/Modal.vue
```diff
@@ -41,40 +102,18 @@ const { activate: activateFocusTrap, deactivate: deactivateFocusTrap } =
...
 watch(
   () => modalState.open,
   (open) => {
     if (open) {
-      // Add the escape key listener
-      window.addEventListener('keydown', onEscape)
-
       // Disable scrolling
-      document.documentElement.style.overflow = 'hidden'
+      window.document.documentElement.style.overflow = 'hidden'
```
geoffgscott: Might need an SSG guard here. Also this might just not be safe at all.

packages/sidebar/src/components/SidebarItem.vue
```diff
@@ -210,11 +200,11 @@ const { draggableAttrs, draggableEvents } = useDraggable({
     </template>
   </ScalarSidebarGroup>
   <ScalarSidebarItem
-    v-else
     is="button"
+    v-else
     v-bind="draggableAttrs"
```
hwkr, line 212: Looks like draggable is still getting enabled in the modal layout

Image

hwkr: Might need a different flag for the modal? 🤔

packages/api-client/src/v2/features/app/App.vue
geoffgscott: We can rename all the singleton hook in a FF

packages/api-client/src/v2/features/modal/Modal.vue
hwkr: Yeah this is what we're doing right now but I'm not sure it's ideal

amritk: its good practice to be safe, but also modal doesn't exist in ssg

DemonHa: @amritk Yeah this is also unsafe for the other reason that we are modifying the overflow property on the body, and we are just removing it when we close the modal, but if the value of that is initally something else, that will mess up with the app it's been injected into

packages/api-client/src/v2/features/app/components/AppSidebar.vue
DemonHa: @hwkr Yeah, you are right, but if we do that change, it should not be modal specific change, it should be the same on the app as well. I am resolving this so we can do this on another PR since it's out of scope

packages/api-client/playground/v2/modal/main.ts
```diff
@@ -24,7 +24,12 @@ const { open } = createApiClientModal({
 // Open the API client right-away
 open()
 
-document.getElementById('button')?.addEventListener('click', () => open())
+document.getElementById('button')?.addEventListener('click', () =>
+  open({
+    path: '/user/signup',
```
amritk, line 29: testing data?

packages/api-client/src/v2/features/app/hooks/use-app-sidebar.ts
```diff
@@ -247,8 +246,19 @@ export const useSidebarState = ({
...
+      () => toValue(documentSlug),
+      () => toValue(path),
+      () => toValue(method),
+      () => toValue(exampleName),
+    ],
+    () => {
+      const newDocument = toValue(documentSlug)
+      const newPath = toValue(path)
+      const newMethod = toValue(method)
+      const newExample = toValue(exampleName)
```
amritk: so if they are already reactive, just watch them directly
```suggestion
    [
      workspaceStore,
      documentSlug,
      path,
      method,
      exampleName,
    ],
     ([newDocument, newPath, newMethod, newExample]) => {
```

packages/api-client/src/v2/features/modal/hooks/use-modal-sidebar.ts
```diff
@@ -0,0 +1,212 @@
...
+  method,
+  exampleName,
+  route,
+}: {
+  workspaceStore: MaybeRefOrGetter<WorkspaceStore | null>
+  documentSlug: MaybeRefOrGetter<string | undefined>
+  path: MaybeRefOrGetter<string | undefined>
+  method: MaybeRefOrGetter<HttpMethod | undefined>
+  exampleName: MaybeRefOrGetter<string | undefined>
+  route: (payload: RoutePayload) => void
```
amritk:
```suggestion
  workspaceStore: Computed<WorkspaceStore | null>
  documentSlug: Computed<string | undefined>
  path: Computed<string | undefined>
  method: Computed<HttpMethod | undefined>
  exampleName: Computed<string | undefined>
  route: (payload: RoutePayload) => void
```
we are passing in computed an its only used in one place, no need for the ambiguity of maybeRefOrGetter

packages/api-client/src/v2/features/modal/Modal.vue
amritk: yea store it then restore

packages/api-client/playground/v2/modal/main.ts
DemonHa: yeah

amritk (changes requested):
h

PR #7526, opened 2025-12-11
fix(integrations): Add ElysiaJs and Fastify themes to ScalarTheme enum

ame0410:
## Problem

ElysiaJs and Fastify themes are missing in the ScalarTheme enum.

## Solution

Add ElysiaJs and Fastify themes to ScalarTheme enum.

### DIFF
--- a/integrations/dotnet/shared/src/Scalar.Shared/Enums/ScalarTheme.cs
+++ b/integrations/dotnet/shared/src/Scalar.Shared/Enums/ScalarTheme.cs
@@ -90,7 +90,7 @@ public enum ScalarTheme
     /// <summary>
     /// Elysia.js theme.
     /// </summary>
-    [Description("elysiaJs")]
+    [Description("elysiajs")]
     ElysiaJs,
 
     /// <summary>


xC0dex:
Hey @ame0410! Thanks for opening a PR.

These themes are actually framework-specific (as their names suggest). That’s why we intentionally didn’t include the ElysiaJS and Fastify themes in the enums. I’m curious, how did you come across this? 👀

ame0410:
Hi @xC0dex,

I see, thanks for the clarification!

I ran into this while testing the `Scalar.AspNetCore 2.11.5` package in a `.NET 10 Web API` project. I am currently in the process of replacing Swagger UI with Scalar, and so far it looks very promising.

When I launch my Web API at `/scalar` in `Google Chrome`, the Scalar UI loads with the `Developer Tools` option enabled. From there, I can open `Configure` and select a `Theme`, as shown in the screenshot below.

image

Would it make sense to allow passing the theme as a string instead of an enum in ScalarOptions.WithTheme() as a possible solution?

hanspagel:
Thanks for bringing this up! Technically, your PR is correct. I don’t remember why those two (and only those two) framework themes went into the themes package. I need to do a bit of research, so we can find a proper solution.

hanspagel:
closing in favor of #7811

PR #7532, opened 2025-12-12
fix(api-client): get the correct document using the navigation document name

DemonHa:
**Distinguish `id` from `name` on navigation entry of type `document`**

We previously treated id and name as interchangeable, even though they are not guaranteed to be the same. While they often look similar, they represent different concepts and should not be mixed.

Our store uses the document name as the canonical identifier. However, the navigation entries of type document use their own id field, which does not map to the store identifier. The correct mapping is `name`.

This PR updates all places where we incorrectly used the navigation id as the document name, resolving several mismatches and bugs caused by this confusion. All references now consistently use the document’s name when interacting with the store.

## Checklist

- [ ] I added a changeset. 
- [ ] I added tests.

---

> [!NOTE]
> Switches all document lookups and delete flows to use navigation entry `name` (not `id`), and awaits route updates after deletes.
> 
> - **App sidebar and events**
>   - Use `document.name` instead of `document.id`/`item.id` when emitting delete events for `document`, `tag`, `operation`, and `example` in `AppSidebar.vue`.
>   - In `use-workspace-client-app-events.ts`:
>     - Resolve current document via `['x-scalar-navigation'].name` for sidebar refresh and `rebuildSidebar` calls.
>     - Make `operation:delete:operation` and `operation:delete:example` handlers async and `await` `router.replace` after redirects.
> - **Changeset**
>   - Add patch changeset for `@scalar/api-client` describing the fix.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f6044c0132da137e1b5d74e6818d304fec3ec8b4. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/.changeset/fluffy-scissors-study.md
+++ b/.changeset/fluffy-scissors-study.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
+
+fix: get the correct document using the navigation document name


amritk (approved):
never mix name and ID! 🙂

PR #7536, opened 2025-12-12
feat(api-client): send request for client v2

amritk:
## Problem

Build the request for the client v2
Send the request for the client v2

PR has gotten a bit large, I will split it up into 2-3 PRs

## Checklist

- [ ] I added a changeset. 
- [x] I added tests.

---

> [!NOTE]
> Implements full request build/send pipeline with parameters/body/security/cookies, plugin hooks, hotkeys/toasts wiring, and robust response handling (headers, cookies, streaming) for client v2.
> 
> - **Core (client v2 request pipeline)**:
>   - Add `helpers/build-request` to construct fetch `Request` (env/server vars, path/query/headers, body, security, cookies, proxy/Electron handling) with abort controller and error normalization.
>   - Add `helpers/send-request` to execute requests with plugin hooks, normalize headers, extract Set-Cookie keys, support SSE streaming and binary/text bodies; return structured `ResponseInstance` with timing/size.
>   - Implement helpers: `build-request-parameters`, `build-request-body`, `build-request-security`, `build-request-cookie-header`, `get-cookie-header-keys`, and parameter/body example resolvers.
> - **Plugins**:
>   - Introduce `v2/helpers/plugins` with typed hooks (`beforeRequest`, `responseReceived`) and request/response components; update usage across blocks.
> - **UI wiring**:
>   - Update `OperationBlock` to handle execute/cancel, build/send requests, emit loading events, store request/response, and show toast errors.
>   - Wire `RequestBlock`/`ResponseBlock` props (proxyUrl, plugins, cookies); improve headers/cookies empty states and cookie parsing.
>   - Add `ScalarToasts` to App/Modal; cancel in-flight request on modal close.
> - **Auth/OAuth**:
>   - Enhance OAuth2 flow (PKCE, relative URLs) and token exchange via proxy; propagate `proxyUrl`/`server` through auth components.
> - **Hotkeys & Events**:
>   - Rename enter hotkey to `operation:send:request:hotkey`; add cancel event; update handlers.
> - **Misc**:
>   - Add `BUILDING_REQUEST_FAILED` error; tweak variable replacement util; minor UI tweaks; remove dev-only seeded docs.
> - **Tests**:
>   - Comprehensive unit tests for all new helpers, blocks, OAuth, hotkeys, and plugins.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 2064aec701851b56230ae0bc1e17482cdedd12b2. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/api-client/src/v2/blocks/operation-block/OperationBlock.vue
+++ b/packages/api-client/src/v2/blocks/operation-block/OperationBlock.vue
@@ -144,11 +144,11 @@ const handleExecute = async () => {
     return
   }
 
-  // Store the abort controller and request
+  // Store the abort controller for cancellation
   abortController.value = result.controller
-  request.value = result.request
 
-  // TODO: send start event for animation
+  // Start the animation
+  eventBus.emit('hooks:on:request:sent')
 
   /** Execute the request */
   const [sendError, sendResult] = await sendRequest({
@@ -158,7 +158,8 @@ const handleExecute = async () => {
     request: result.request,
   })
 
-  // TODO: send stop event for animation
+  // Stop the animation
+  eventBus.emit('hooks:on:request:complete')
 
   // Toast the execute error
   if (sendError) {
@@ -168,6 +169,7 @@ const handleExecute = async () => {
 
   // Store the response
   response.value = sendResult.response
+  request.value = sendResult.request
 }
 
 onMounted(() => {


packages/api-client/src/v2/blocks/operation-block/helpers/build-request-body.ts
```diff
@@ -0,0 +1,55 @@
...
+  env: Record<string, string> = {},
+): BodyInit | null => {
+  if (!requestBody) {
+    return null
+  }
+
+  const example = getExampleFromBody(requestBody, contentType, exampleKey)
+
+  // We got an example value or generated example from the schema
+  if (example?.value) {
```
DemonHa: I would prefer reverting those ifs, easier to read

packages/api-client/src/v2/blocks/request-block/components/RequestBody.vue
```diff
@@ -78,13 +78,24 @@ const contentTypes = {
...
+  () =>
     requestBody?.['x-scalar-selected-content-type']?.[exampleKey] ??
     Object.keys(requestBody?.content ?? {})[0] ??
-    'other'
-  )
-})
+    'other',
+)
+
+watch(
```
amritk, line 89: we can also add the content type header here

amritk: or in the mutator

packages/api-client/src/v2/blocks/operation-block/helpers/build-request.ts
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    /** Resolve the selected content type from the request body */
+    const contentType =
+      requestBody?.['x-scalar-selected-content-type']?.[exampleKey] ??
+      Object.keys(requestBody?.content ?? {})[0] ??
+      'application/json'
+
+    /** Build out the request parameters */
+    const params = buildRequestParameters(operation.parameters ?? [], env, exampleKey, contentType)
```
DemonHa: You are using the content type of the body here, which might differ from the header value. I think we want to use the header value here

packages/api-client/src/v2/features/app/hooks/use-workspace-selector.ts
```diff
@@ -103,22 +103,22 @@ export const useWorkspaceSelector = (): UseWorkspaceSelectorReturn => {
...
-        name: 'swagger-petstore-3-0',
-        url: 'https://petstore3.swagger.io/api/v3/openapi.json',
-      })
-
-      await draftStore.addDocument({
-        name: 'stripe',
-        url: 'https://raw.githubusercontent.com/stripe/openapi/refs/heads/master/openapi/spec3.json',
-      })
-    }
+    // if (import.meta.env.DEV) {
```
amritk: @DemonHa do we still need these? we got persistence now can just load documents through the UI

packages/api-client/src/v2/blocks/operation-block/helpers/build-request.ts
```diff
@@ -0,0 +1,154 @@
...
+
+    /** Build the js request object */
+    const request = new Request(proxiedUrl, {
+      method,
+      headers,
+      signal: controller.signal,
+      body,
+    })
+
+    return [
```
DemonHa, line 149: can we change the return type, let's do a discriminated union

packages/api-client/src/v2/blocks/operation-block/helpers/get-example.ts
```diff
@@ -3,11 +3,23 @@ import type { ParameterObject } from '@scalar/workspace-store/schemas/v3.1/stric
...
+
+  // Schema based parameters
   if ('examples' in param) {
-    return getResolvedRef(param.examples?.[exampleKey])
+    if (param.examples?.[exampleKey]) {
+      return getResolvedRef(param.examples?.[exampleKey])
+    }
+
+    // Fallback to default
+    const defaultValue = getResolvedRef(param.schema)?.default
```
DemonHa: cool!

packages/api-client/src/v2/blocks/operation-block/OperationBlock.vue
```diff
@@ -100,15 +120,43 @@ const emit = defineEmits<{
...
+    method,
+    operation,
+    path,
+    securitySchemes,
+    selectedSecurity: selectedSecurity?.selectedSchemes ?? [],
+    server,
+    proxyUrl,
+  })
+
+  if (error) {
```
DemonHa, line 142: I think we should move away from toasters 😬

packages/api-client/src/v2/features/operation/Operation.vue
```diff
@@ -34,6 +42,12 @@ const security = computed(() =>
   getSecurityRequirements(document, operation.value),
 )
 
+/** Combine the workspace and document cookies */
+const globalCookies = computed(() => [
```
DemonHa, line 46: oh this is just merging them, I think we might need to handle the overrides, or we don't care?

packages/api-client/src/v2/blocks/operation-block/helpers/build-request.ts
amritk: We actually want the one from the body here because it will dictate the shape of the data, the user can change the content type as well

packages/api-client/src/v2/features/app/hooks/use-workspace-selector.ts
DemonHa: Yeah we do not need them anymore

packages/api-client/src/v2/blocks/operation-block/helpers/build-request.ts
amritk: we can if its better? This one is stolen from GO haha. 

thought process: puts the error first so you are "forced" to handle it. If we just use a discriminated union you can skip handling the error and just check for data. You can do that here too but you know you are doing something wrong

packages/api-client/src/v2/blocks/operation-block/OperationBlock.vue
amritk: LOL

packages/api-client/src/v2/features/operation/Operation.vue
amritk: I don't think we care too much right now, we'll also need to handle disabling globals per example but the current client doesn't support that yet

packages/api-client/src/v2/blocks/operation-block/helpers/build-request.ts
DemonHa: I would prefer the union here. Typescript can infer the types correctly that way

something like: { ok: true, request: Request, controller: Controller} | { ok: false, message: string }

amritk: oh typescript infers the types here too!
```ts
if (error) {
  result = null
  return
}
result <- typed
```

skips having an extra parameter. plus makes you handle the error

DemonHa: yeah it feels weird tho, since go provides syntactic sugar to support that, we are returning an tuple there. Just to be consistent with other places tho.

packages/api-client/src/v2/blocks/operation-block/helpers/build-request-cookie-header.ts
```diff
@@ -0,0 +1,91 @@
...
+  /** The url of the request used to filter global cookies by domain */
+  url: string
+  /**
+   * If we are running in Electron or using the proxy, we need to add a custom header
+   * that's then forwarded as a `Cookie` header.
+   */
+  useCustomCookieHeader: boolean
+}): null | { name: string; value: string } => {
+  /** Filter the global cookies by domain + parse */
+  const filteredGlobalCookies = globalCookies.flatMap((cookie) => {
```
DemonHa, line 65: can we also filter by path

PR #7548, opened 2025-12-15
fix(java): maven publish

xC0dex:
## Problem

Maven and I are not friends. Java publish job still fails.

## Solution

We became friends now 🙂 I reproduced the issue using a custom Maven account and was able to solve it, FINALLY.

## Checklist

- [x] I added a changeset. 
- [ ] I added tests.


hanspagel (approved):
omg, thank you so much

PR #7550, opened 2025-12-15
fix(api-reference): unstable width on mobile layout

MatteoH2O1999:
## Problem

The width of the mobile layout is unstable (see [example](https://api.webinforete.it/))

## Solution

Use explicit percentage instead of auto in `grid-template-columns` for the `references-layout` class.

## Checklist

- [x] I added a changeset. 
- [ ] I added tests.

### DIFF
--- a/.changeset/wise-bobcats-burn.md
+++ b/.changeset/wise-bobcats-burn.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
+Fix content width in mobile layout


hanspagel:
Handing this over to @hwkr

hwkr:
@MatteoH2O1999 Thanks for the fix! Are you able to upload a video of the issue and the fix?

MatteoH2O1999:
Before: 

https://github.com/user-attachments/assets/1435f2f5-7174-4974-a9b0-46e53f77c4f8

After:

https://github.com/user-attachments/assets/b479c8e0-7f22-4fa3-8a2b-5a6509b86ce0

PR #7558, opened 2025-12-15
fix(ts-to-openapi): enable `knip`

marcalexiei:
## Problem

- Followup of #7233

## Solution

- replace star exports with named exports
- esbuild - explicit entrypoint
-  add explicit return type to each function
- remove `@ts-nocheck` directives
- move test files inside `test`  folder, outside of `src` so they are not published
- `.prettierignore` refinements
  - dedupe `dist` matches
  - move md(x) exclusion at the top of the file
  - add `examples/docusaurus/build/**`
  - separate group for `vite`, `examples`, `integrations` and `packages` matches 

## Checklist

- [x] I added a changeset. 
- [x] I added tests.


hanspagel (approved):
niiice

PR #7562, opened 2025-12-16
feat(components): remove headless ui from dropdown component 

hwkr:
Removes Headless UI from the dropdown component and updates the dropdown 

Changes:

- Removed Headless UI dependency from ScalarDropdown
- Implemented custom dropdown that support targeting arbitrary nodes
- Updated API client v2 sidebar to use the new dropdown API
- Expanded test coverage for new functionality

Things that are the same:

- External dropdown component API (e.g. `<ScalarDropdown> → <ScalarDropdownItem>`)
- Full keyboard navigation (Arrow keys, Enter, Space, Escape)
- ARIA attributes for accessibility

API Client V2 Before / After:

https://github.com/user-attachments/assets/ad1ff5f9-a3b3-4548-8b5a-09c28997fa1c

## Checklist

- [x] I added a changeset. 
- [x] I added tests.

### DIFF
--- a/packages/components/src/components/ScalarDropdown/ScalarDropdown.test.ts
+++ b/packages/components/src/components/ScalarDropdown/ScalarDropdown.test.ts
@@ -193,8 +193,8 @@ describe('ScalarDropdown', () => {
       await nextTick()
 
       const trigger = wrapper.get('button#trigger')
-      // Component checks for 'Space' key name, not ' ' character
-      await trigger.trigger('keydown', { key: 'Space' })
+      // Browser sends ' ' (space character) for Space key, not 'Space'
+      await trigger.trigger('keydown', { key: ' ' })
 
       expect(wrapper.find('[role="menu"]').exists()).toBe(true)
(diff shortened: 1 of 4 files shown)


packages/components/src/components/ScalarDropdown/ScalarDropdownMenu.vue
```diff
@@ -18,7 +18,7 @@ export default {}
 </script>
 <script setup lang="ts">
 import { useBindCx } from '@scalar/use-hooks/useBindCx'
-import type { Component } from 'vue'
+import { type Component } from 'vue'
```
hanspagel, line 21: what happened here? is that biome?

hwkr: I guess so 🤔 I didn't change it intentionally

amritk: should be the first one but doesn't really matter

PR #7563, opened 2025-12-16
fix(api-client): propagate client plugins to operation view

DemonHa:
Correctly propagate client plugins to the operation vue to hook into the request/response

## Checklist

- [x] I added a changeset. 
- [ ] I added tests.

---

> [!NOTE]
> Propagates client plugins from app/modal into the operation view and passes them to OperationBlock, updating types and tests accordingly.
> 
> - **Plugins propagation**
>   - Add `plugins` prop to `App.vue` and pass via `routerViewProps` to routes.
>   - Extend `RouteProps` with `plugins` and thread through collection pages and `Operation.vue`.
>   - In `Operation.vue`, forward `plugins` to `OperationBlock`.
> - **App/Modal initialization**
>   - `createApiClientApp` accepts `plugins` and provides them to `App`.
>   - `Modal.vue` accepts optional `plugins` and forwards to `Operation`.
>   - `createApiClientModal` adds `plugins` option and passes it to `Modal`.
> - **Types**
>   - Simplify `ClientPluginComponents` types to generic `Component` for `request` and `response` in `helpers/plugins.ts`.
> - **Tests**
>   - Update affected tests to include `plugins: []` in props.
> - **Changeset**
>   - Patch release for `@scalar/api-client`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 6204de0081636c4e34d55454b48ce8735c8b674a. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/api-client/src/v2/features/app/App.vue
+++ b/packages/api-client/src/v2/features/app/App.vue
@@ -37,9 +37,9 @@ import WebTopNav from './components/WebTopNav.vue'
 import { useTabs } from './hooks/use-tabs'
 import { useWorkspaceClientAppEvents } from './hooks/use-workspace-client-app-events'
 
-const { layout, plugins } = defineProps<{
+const { layout, plugins = [] } = defineProps<{
   layout: Exclude<ClientLayout, 'modal'>
-  plugins: ClientPlugin[]
+  plugins?: ClientPlugin[]
 }>()
 
 /** Expose workspace store to window for debugging purposes. */
--- a/packages/api-client/src/v2/features/app/helpers/create-api-client-app.ts
+++ b/packages/api-client/src/v2/features/app/helpers/create-api-client-app.ts
@@ -36,10 +36,7 @@ export const createAppRouter = (layout: CreateApiClientOptions['layout']) => {
 /**
  * Create the API Client with router and passes in the workspace store as a prop
  */
-export const createApiClientApp = (
-  el: HTMLElement | null,
-  { layout = 'desktop', plugins = [] }: CreateApiClientOptions,
-) => {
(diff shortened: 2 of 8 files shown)


packages/api-client/src/v2/helpers/plugins.ts
```diff
@@ -15,8 +13,8 @@ export type ClientPluginHooks = {
 
 /** A type representing the components that a client plugin can define */
 export type ClientPluginComponents = {
-  request: Component<{ operation: OperationObject; selectedExample?: string }>
```
amritk, line 18: does this remove  the typed props?

DemonHa: yeah it does remove the typesafe, cuz that type is not really working

DemonHa: the only way to fix this is type it as a component and add type assertion which extracts the props

amritk: kk we'll probably want typed props at some point

packages/api-client/src/v2/features/modal/helpers/create-api-client-modal.ts
```diff
@@ -23,6 +24,8 @@ export type CreateApiClientModalOptions = {
   mountOnInitialize?: boolean
   /** The workspace store must be initialized and passed in. */
   workspaceStore: WorkspaceStore
+  /** Api client plugins to include in the modal */
+  plugins?: ClientPlugin[]
```
amritk, line 28:
```suggestion
  plugins: ClientPlugin[]
```
I generally prefer making everything mandatory for development, just because we had few cases where we forgot to pass down an optional prop. Then when we expose to public we can add back the ?

PR #7568, opened 2025-12-17
fix(workspace-store): object-merge array replacement issue

amritk:
## Problem

Currently if you have `hiddenClients: true` set in your config, that will make httpClients an empty array. However when we merge with the default values, the array gets merged and httpClients gets every client from the default
```ts
httpClients = []
// to
httpClients = ["c/libcurl", "clojure/clj_http", "csharp/httpclient", "csharp/restsharp", etc...]
```

## Solution

First I just prevented the array merging behaviour  in merge-obect but it turns out that our store depends on this behaviour when merging the strictDocument with the coerced one. So I changed it to a flag instead.

## Checklist

- [x] I added a changeset. 
- [x] I added tests.

---

> [!NOTE]
> Introduce `replaceArrays` option in `mergeObjects` and apply it when merging document/workspace configs to prevent array index-wise merging.
> 
> - **Core utilities**:
>   - `mergeObjects` now supports a `replaceArrays` flag to replace arrays wholesale instead of merging indices; flag is propagated in recursive merges (`helpers/merge-object.ts`).
> - **Workspace config merging**:
>   - `getDocumentConfiguration` uses `mergeObjects(..., true)` to ensure arrays in configs are replaced, not merged (`src/client.ts`).
> - **Tests**:
>   - Add unit test verifying array replacement behavior when `replaceArrays` is true (`helpers/merge-object.test.ts`).
> - **Changeset**:
>   - Patch release for `@scalar/workspace-store` noting fix for object merge replacing arrays.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit bcce566655ec33cdfd5702537af95afd7e60c1d5. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/workspace-store/src/helpers/merge-object.test.ts
+++ b/packages/workspace-store/src/helpers/merge-object.test.ts
@@ -135,7 +135,7 @@ describe('mergeObjects', () => {
     })
   })
 
-  it('should deeply merge the objects with the correct value for the array', () => {
+  it('should replace the first array with the second when replaceArray is true', () => {
     const a = {
       arr: [1, 2, 3],
     }


packages/workspace-store/src/helpers/merge-object.test.ts
```diff
@@ -134,4 +134,18 @@ describe('mergeObjects', () => {
       b: 2,
     })
   })
+
+  it('should deeply merge the objects with the correct value for the array', () => {
```
amritk:
```suggestion
  it('should replace the first array with the second when replaceArray is true', () => {
```

PR #7570, opened 2025-12-17
feat(mock-server): docker image

xC0dex:
## Problem

Currently, we don't have a Docker image for our awesome mock server. So devs are forced to use Node.

## Solution

This PR adds a new docker image `scalarapi/mock-server` which wraps the mock server into a lightweight Docker image.

## Checklist

- [x] I added a changeset. 
- [x] I added tests.

### DIFF
--- a/documentation/guides/mock-server/docker.md
+++ b/documentation/guides/mock-server/docker.md
@@ -7,6 +7,21 @@ Run the Scalar Mock Server in a Docker container.
 Pull and run the Docker image:
 
 ```bash
+# Using command-line URL argument (Method 1)
+docker run -p 3000:3000 \
+  scalarapi/mock-server --url https://api.example.com/openapi.yaml
+
+# Using environment variable with document content (Method 2)
+docker run -p 3000:3000 \
+  -e OPENAPI_DOCUMENT="$(cat ./openapi.yaml)" \
+  scalarapi/mock-server
+
+# Using URL environment variable (Method 3)
+docker run -p 3000:3000 \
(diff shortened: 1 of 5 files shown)


jspdown:
This is great! 

I met with @hanspagel at APIDays Paris and we got the occasion to discuss about this feature. Here's my use case:

I want to use the mock server for development, testing, and documentation purposes, allowing me and my users to spin up a fake server from a mocked API based on an OpenAPI spec.

The feature I'm most excited about is the ability to expose the spec on an endpoint. This is particularly valuable for our solution where we're discovering APIs on a Kubernetes cluster. Here's an example of how it could be used on our side:

```yaml
apiVersion: hub.traefik.io/v1alpha1
kind: API
metadata:
  name: admin
spec:
  openApiSpec:
    path: /spec.json
    
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: my-app
spec:
  replicas: 1
  selector:
    matchLabels:
      app: my-app
  template:
    metadata:
      labels:
        app: my-app
    spec:
      containers:
        - name: api
          image: scalarapi/mock-server:latest
          args: ["--url", "https://raw.githubusercontent.com/OAI/OpenAPI-Specification/refs/heads/main/_archive_/schemas/v3.0/pass/petstore.yaml"]
          imagePullPolicy: IfNotPresent
          
```

Thanks to this, I would eventually be able to replace a fairly complex multi-Kubernetes-resource setup with a simple Deployment using the mock image. So for ease of use, being able to specify an OpenAPI URL (as shown above) is ideal, it keeps everything self-contained and straightforward.

Thanks for working on this! 🙌

amritk (approved):
LGTM! Nice work

PR #7576, opened 2025-12-17
fix(sidebar): enable `knip`

marcalexiei:
## Problem

- Followup of #7233

## Solution

- remove index `exports` field  pointing to non-existing cjs file
- use `scalar-format-*` bins in `format` scripts
- remove unused `vite-svg-loader` dev dependency
- remove `DRAG_OFFSETS` and inline its values directly into `DragOffset` type, 
   as it was not referenced by any runtime code
- remove redundant `null` in `DragOffset` type usage
- refine `throttle` signature to infer callback parameters in return type function
- use fake timers in `use-draggable` "sets and removes data-dragging attribute during drag lifecycle" test

## Checklist

- [x] I added a changeset. 
- [x] I added tests.

### DIFF
--- a/.changeset/modern-sails-cover.md
+++ b/.changeset/modern-sails-cover.md
@@ -1,0 +1,5 @@
+---
+'@scalar/sidebar': patch
+---
+
+fix: remove cjs file from index export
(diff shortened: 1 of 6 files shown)


hanspagel (approved):
nice one

PR #7578, opened 2025-12-17
fix(api-reference): does not show `readOnly` attributes in the example responses

### LINKED ISSUE #7566: Read-only rendering regression
### What happens?

Read-only fields in objects that are used as both the input and output to an operation will render incorrectly in the response, where read-only fields should be present but are missing.

Version that works: https://cdn.jsdelivr.net/npm/@scalar/api-reference@1.38.0
Version that's broken: https://cdn.jsdelivr.net/npm/@scalar/api-reference@1.38.1 to latest

Correct behavior:

Image

Broken behavior:

Image

Example schema:

```json
{
  "additionalProperties": false,
  "properties": {
    "$schema": {
      "description": "A URL to the JSON Schema for this object.",
      "examples": [
        "https://example.com/schemas/Group.json"
      ],
      "format": "uri",
      "readOnly": true,
      "type": "string"
    },
    "@id": {
      "examples": [
        "123e4567-e89b-12d3-a456-426614174000"
      ],
      "format": "uri-reference",
      "readOnly": true,
      "type": "string"
    },
    "created_at": {
      "description": "Timestamp when the group was created.",
      "format": "date-time",
      "readOnly": true,
      "type": "string"
    },
    "settings": {
      "additionalProperties": false,
      "properties": {
        "age_limit": {
          "description": "The maximum age in seconds for the group before it is automatically deleted. A value of 0 means no age limit.",
          "examples": [
            0
          ],
          "format": "double",
          "type": "number"
        },
        "requires_users": {
          "description": "Indicates if the group requires at least one 'user' role participant to be present for the group to exist.",
          "examples": [
            true
          ],
          "type": "boolean"
        },
        "user_join_deadline": {
          "description": "The duration in seconds to wait for a 'user' role participant to join before deleting the group. Only applicable if 'requires_users' is true.",
          "examples": [
            60
          ],
          "format": "double",
          "type": "number"
        },
        "zero_users_timeout": {
          "description": "The duration in seconds to wait after the last 'user' role participant leaves before deleting the group. Only applicable if 'requires_users' is true.",
          "examples": [
            30
          ],
          "format": "double",
          "type": "number"
        }
      },
      "type": "object"
    }
  },
  "required": [
    "@id",
    "created_at"
  ],
  "type": "object"
}
```

I expect it to show read-only fields in the response and hide them in the request. With this bug the newest versions of the docs are unusable.

### OpenAPI Document

https://sandbox.scalar.com/e/a9GJz

amritk:
## Problem
closes #7566

Our cache in  `getExampleFromSchema` didn't take options into account

## Solution

Added a cachekey to it

Before:
image

After:
image

## Checklist

- [x] I added a changeset. 
- [ ] I added tests.

---

> [!NOTE]
> Make getExampleFromSchema cache results per-options and centralize omit logic, ensuring correct examples across modes; add caching tests.
> 
> - **Helpers (`get-example-from-schema.ts`)**:
>   - **Options-aware caching**: Switch `resultCache` to `WeakMap<object, Map<string, unknown>>` and store results by an options-derived `cacheKey`.
>   - **API/Types**: Introduce `GetExampleFromSchemaOptions` and `createOptionsCacheKey`; thread `cacheKey` through all code paths and `cache()` calls.
>   - **Omit logic**: Centralize checks for `deprecated`, `readOnly`/`writeOnly` (via `mode`) and `omitEmptyAndOptionalProperties` in `shouldOmitProperty`.
>   - Minor refactors for clarity without changing external API.
> - **Tests (`get-example-from-schema.test.ts`)**:
>   - Add caching tests verifying different options yield different results and identical calls return the same cached object reference.
> - **Changeset**: Add patch note for `@scalar/api-client`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 8e5d2c9b62522c0f1005d64ac767f42127e3d33a. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


amritk:
e2e tests are actually working properly now, just gotta update the snapshots

packages/api-client/src/v2/blocks/operation-code-sample/helpers/get-example-from-schema.ts
```diff
@@ -484,10 +507,14 @@ export const getExampleFromSchema = (
   }
   seen.add(targetValue)
 
+  /** Make the cache key unique per options */
+  const cacheKey = JSON.stringify(options ?? {})
```
DemonHa: can we do some better key generation logic here, this will depend on the object key sorting behaviour

amritk:
Alright update snapshots, @hanspagel I'm going to get this one in then release. theres a bit of a regression on main thats fixed in this branch

PR #7589, opened 2025-12-18
chore(components): upgrade to Storybook 10

hwkr:
Upgrades us to storybook 10 which fixes a security vulnerability and is way smaller.

Storybook-10-Image

## Checklist

- [x] I added a changeset. 
- [x] I added tests.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1703,9 +1703,6 @@ importers:
       '@scalar/use-hooks':
         specifier: workspace:*
         version: link:../use-hooks
-      '@scalar/use-toasts':
-        specifier: workspace:*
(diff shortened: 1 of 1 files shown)


packages/components/.storybook/preview.css
hwkr, line 1: Updated the CSS here to make dark mode work and add some syntax highlighting

packages/components/src/components/ScalarFloating/snapshots/base-max-size-bottom.png
hwkr, line 1: The ones on the right are actually what's in the story we were just missing some CSS before.

geoffgscott (approved):

PR #7591, opened 2025-12-19
fix(api-client): prevent empty body from auto-setting content-type

JianJroh:
## Problem

When a POST request has no body content (empty string), browsers automatically add `Content-Type: text/plain;charset=UTF-8` header. This causes unnecessary headers to be sent, which can lead to server rejections or unexpected behavior.

## Solution

Modified createFetchBody to return undefined instead of an empty string when the body is empty after template variable replacement. Using body: undefined prevents the browser from adding the default Content-Type header, ensuring clean requests without unnecessary headers.

## Checklist

- [ ] I added a changeset. 
- [x] I added tests.

### DIFF
--- a/packages/api-client/src/libs/send-request/create-fetch-body.test.ts
+++ b/packages/api-client/src/libs/send-request/create-fetch-body.test.ts
@@ -22,4 +22,23 @@ describe('createFetchBody', () => {
       contentType: 'text',
     })
   })
+
+  it('should return undefined body for empty string', () => {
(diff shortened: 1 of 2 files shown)


amritk:
This PR is actually on the old code, would you mind trying on the latest references?

hanspagel:
Thanks for bringing this up. Looks like we fixed the issue already!

https://sandbox.scalar.com/e/yMkaf

Screenshot 2026-01-14 at 14 05 56

PR #7605, opened 2025-12-20
feat(api-reference): client v2 modal

amritk:
## Problem

Its not quite done yet but we can start playing around + check the code

To do:
- [x] test all the specs 
- [x] remaining config properties to add to the test:
  - [x] auth
    - [x] preferredSecurityScheme + selectedScopes
  - [x] baseServerUrl
  - [x] proxyUrl
  - [x] theme
  - [x] onRequestSent
  - [x] onBeforeRequest
  - [x] onDocumentSelect
  - [x] onShowMore
  - [x] onSidebarClick
  - [x] documentDownloadType
  - [x] hideDownloadButton
  - [x] hideTestRequestButton
  - [x] expandAllModelSections
  - [x] expandAllResponses

punted to fast follow:
- [ ] persist auth + selected client + server
- [ ] onSpecUpdate - hasn't worked for some time

- the e2e test failures are just sub pixel diffs
- CDN tests are green, bolt now shows complex auth as opposed to simple auth before (upgrade)

## Checklist

- [x] I added a changeset. 
- [x] I added tests.

---

> [!NOTE]
> Introduces the Client v2 modal and refactors request/auth flows to support it.
> 
> - **Client v2 modal integration**: New modal factory (`createApiClientModal`) with event bus support, routing helpers, and authentication configuration merge; modal wires into operation view and collection pages.
> - **Auth/security revamp**: Replaces `selectedSecurity` usage with computed `selectedSecuritySchemes`; adds `getSecurityRequirements`, `getSelectedSecurity`, `getSecuritySchemes`, and `mergeAuthConfig` helpers; OAuth2 defaults redirect URI; adds secret coercion (`convertSecuritySchemeSecrets`) and merged schemes typing/exports.
> - **Plugin API change**: `beforeRequest` hook now receives/returns `{ request }`; all callers/tests updated; `sendRequest` and helpers adapted.
> - **Request building improvements**: `buildRequest` now accepts `selectedSecuritySchemes`; `build-request-security` simplified to take resolved schemes; `build-request-body` unwraps proxied File and sets filename; better example/param derivation and form/table handling; small URL/content-type fixes.
> - **Code samples and UI tweaks**: Operation code sample uses event bus, nullable server, stricter props, improved error logging; auth selector label tweaks and optional sections; request/body/table UX copy updates ("Select File").
> - **Routing/operation helpers**: Adds `getSelectedServer` and consolidates operation security/server selection logic; exposes useful helpers/exports across v2 modules.
> - **Tests**: Broad updates/additions across helpers, components, and flows to cover new behavior.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 9fb2b0e888e1c9bbe881ae5281b22c39a2716f29. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/api-client/src/v2/blocks/request-block/RequestBlock.vue
+++ b/packages/api-client/src/v2/blocks/request-block/RequestBlock.vue
@@ -325,7 +325,7 @@ const labelRequestNameId = useId()
         :selectedSecurity
         :selectedSecuritySchemes
         :server
-        title="Authorization" />
+        title="Authentication" />
 
       <!-- Variables (Path Parameters) -->
       <RequestParams


hanspagel:
previously we prefilled the redirect URL

now:
Screenshot 2026-01-05 at 10 06 52

before:
Screenshot 2026-01-05 at 10 07 29

hanspagel:
we showed a select for variables before:

Screenshot 2026-01-05 at 10 10 51

now:

Screenshot 2026-01-05 at 10 11 05

hanspagel:
same as before, but variables in the server URL should use:

* a given value, or:
* the default, or:
* empty string

and *not* use the {variable} string

Screenshot 2026-01-05 at 10 11 31

hanspagel:
the request body for the DELETE request should be "None"

but it’s "other"

and nitpick: it’s "other" in the label and "Other" in the dropdown

Screenshot 2026-01-05 at 10 13 41

hanspagel:
when I "Upload File" and the key is empty, it’s not visible that a file is selected


* it should actually say "*Select* File" here
* we should show the selected file name, or, if that’s not possible, hide the button, even if there is no key
* when I select a file, the field should be enabled automatically
* the padding seems a bit strange

this is how it looks *AFTER* selecting a file (when the key is empty):

Screenshot 2026-01-05 at 10 14 54

while it should look like this (I added a key and removed it later, just for the screenshot):

Screenshot 2026-01-05 at 10 16 40

hanspagel:
files are sent as "[object File]" string

Screenshot 2026-01-05 at 10 19 09

hanspagel:
I tried the Petstore 2.0 example, and it doesn’t show the fields in the request body:

good:

Screenshot 2026-01-05 at 10 21 20

no good:

Screenshot 2026-01-05 at 10 21 58

hanspagel:
I tried Petstore 2.0’s petstore_auth and didn’t pass client credentials.
I closed the window and tried again.
But when I try again, it always starts with the error message, instead of starting from the beginning.

Do we need to request the window url or something?

Screenshot 2026-01-05 at 10 23 38

amritk:
Alright I'm going to fix the tests now, @hanspagel I think I addressed all of your issues except:
- server variable handling, I think it matches what we have right now, we can fix it after!
- the petstore auth issue I get on https://petstore.swagger.io/ as well so its something on their end

packages/api-reference/src/helpers/map-config-plugins.ts
```diff
@@ -0,0 +1,39 @@
...
+          }
+
+          return payload
+        },
+      },
+    })
+  }
+
+  // onResponseReceived
+  if (config.onRequestSent) {
```
DemonHa, line 27: should we add a new hook for `onRequestSent` because its not the same thing as `onResponseReceived`

amritk: yea we'll change it after release

PR #7609, opened 2025-12-23
fix(Scalar.AspNetCore): escape request path

xC0dex:
## Problem

Currently, we don't escape the request path.

## Solution

Now we do!

## Checklist

- [x] I added a changeset. 
- [x] I added tests.


hanspagel (approved):
🙏

PR #7620, opened 2025-12-31
fix(request-body): handle XML content type correctly

DmitriiKhudiakov:
## Problem

When selecting XML examples in the request body, the code attempted to parse XML content as JSON, which caused errors. Additionally, when updating the request body with XML content, `JSON.stringify()` was used instead of treating XML as a string, resulting in incorrect formatting.

This issue prevented users from properly working with XML request bodies in the API client.

## Solution

- Added a check to skip JSON parsing when the content type is XML in the `selectedExample` computed getter
- Use `String()` instead of `JSON.stringify()` when updating the request body for XML content types in the `selectedExample` computed setter

This ensures XML content is handled as a string throughout the example selection and body update flow, while JSON content types continue to work as before.

## Checklist

- [x] I added a changeset. 
- [ ] I added tests.


DemonHa:
Hi @DmitriiKhudiakov, thank you so much for the PR! 🙌

We’re currently in the process of migrating to api-client v2, and this behavior is going to change as part of that work. You can see the relevant implementation here:
[`/packages/api-client/src/v2/blocks/request-block/components/RequestBody.vue`](https://github.com/scalar/scalar/blob/8bd34ed75a9f41eb6ddd0195eed119232c62ee22/packages/api-client/src/v2/blocks/request-block/components/RequestBody.vue#L140)

The current behavior isn’t quite correct because we always attempt to JSON.stringify the request body, even when the content type is XML or another non-JSON format.

In v2, the request body handling is improved: for content types other than `application/octet-stream` and `multipart/form-data`, we’ll either convert the value to a string or use the existing string value directly. Binary and multipart bodies will be stored and displayed differently in the UI, rather than being treated as a simple input field.

DmitriiKhudiakov:
I’d like to ask if this fix can still be merged, as the current behavior breaks working with XML request bodies in the API documentation. Even though this is improved in v2, it significantly impacts users of the current version today. Realted issue: [#7621](https://github.com/scalar/scalar/issues/7621)

packages/api-client/src/views/Request/RequestSection/RequestBody.vue
```diff
@@ -507,7 +510,11 @@ const selectedExample = computed({
         const exampleValue = exampleOption.value as {
           value: Record<string, string>
         }
-        updateRequestBody(JSON.stringify(exampleValue.value, null, 2))
+        updateRequestBody(
```
DemonHa, line 513: We should default to treating the value as a String in all other cases. Only JSON payloads need special handling, where we can use JSON.stringify for formatting.

Could you update this check accordingly and consider extracting this logic?

hanspagel:
I think the opportunity window closed, so sorry. The api reference modal uses v2 already, client.scalar.com too and we're about to release the new standalone platform app, too. 😬

PR #7633, opened 2026-01-05
feat(api-reference): support rendering custom components for paths

KevSlashNull:
## Problem

Right now, we can render custom components on tag sections but custom extensions on paths aren't rendered. If I specify `paths./.get.x-custom-extension`, it won't render a custom component, even though `tags['Foobar'].x-custom-extension` would.

Therefore, this adds support for rendering custom components based on specification extensions on paths. The change affects both the `modern` and `classic` layouts.

Our use case at GitLab (see [our issue](https://gitlab.com/gitlab-org/gitlab/-/work_items/584095)) is that we want to show custom badges (such as "Generally available") on specific endpoints because this status differs across endpoints not tags.

## Screenshots

Example from the Vue playground.

| Layout | Before | After |
| --- | --- | --- |
| Modern | image | image |
| Classic | image | image |

### Use case example (custom component)

image

## Checklist

- [x] I added a changeset. 
- [ ] I added tests.

### DIFF
--- a/.changeset/plenty-forks-stay.md
+++ b/.changeset/plenty-forks-stay.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': minor
+---
+
+feat(api-reference): support rendering custom components for paths


KevSlashNull:
cc @maxehmookau

packages/api-reference/src/features/Operation/layouts/ClassicLayout.vue
```diff
@@ -166,6 +169,9 @@ const { copyToClipboard } = useClipboard()
     </template>
     <div class="endpoint-content">
       <div class="operation-details-card">
+        <div class="operation-details-card-item">
+          <SpecificationExtension :value="operationExtensions" />
+        </div>
```
hanspagel: we may not want the empty div here, let me check

KevSlashNull: we could use `Object.keys(operationExtensions).length` to conditionally render this like we do for `operation?.callbacks` below?

hanspagel: sounds good

KevSlashNull: thanks, pushed! :)

KevSlashNull:
thanks for the fast review @hanspagel as well :pink_heart:

PR #7654, opened 2026-01-07
fix(components): explicitly set line height for buttons and sections

hwkr:
We had a `text-sm` class that was overriding the line height for section labels because it wasn't explicitly set. This sets it explicitly for all button and section labels.

## Checklist

- [x] I added a changeset. 
- [x] I added tests. ← Covered by snapshots


packages/components/src/components/ScalarSidebar/ScalarSidebarItems.vue
```diff
@@ -34,7 +34,7 @@ const { cx } = useBindCx()
     :is="is"
     v-bind="
       cx(
-        'group/items relative flex flex-col text-base p-3 gap-px transition-transform duration-300',
```
hwkr, line 37: Makes more sense to set it on the button

cameronrohani (approved):

amritk (approved):

PR #7655, opened 2026-01-07
fix(build-tooling): import fast glob as commonJS

hwkr:
This was breaking a bunch of the builds in the Org

## Checklist

- [x] I added a changeset. 
- [ ] I added tests.


hanspagel (commented):
I think you can also just import sync (instead of glob) and you’re good too

PR #7675, opened 2026-01-08
feat(api-client): ensure patch is uppercased when building request

### LINKED ISSUE #7674: PATCH method is being sent lowercase
### What happens?

I have a PATCH endoint. Scalar UI is sending it with a lowercase `patch` for some reason (tested in Firefox)

Image

This is breaking PATCH endpoints in my api doc

Should send uppercase PATCH

### OpenAPI Document

https://sandbox.scalar.com/e/XeGYr

amritk:
## Problem

closes #7674

When we moved to openapi based store our methods became lower cased (OpenAPI style), however patch still needs to be manually uppercased as fetch does not do it implicitly. (https://github.com/JakeChampion/fetch/issues/254)

## Solution

Added an uppercase to method when its a patch

## Checklist

- [x] I added a changeset. 
- [x] I added tests.

---

> [!NOTE]
> Ensures correct HTTP method casing for requests, addressing `fetch`'s handling of lowercase `patch`.
> 
> - Uppercases `method` in `buildRequest` before creating the `Request` object
> - Adds test validating uppercase `PATCH` in `build-request.test.ts`
> - Adds changeset for `@scalar/api-client` patch release
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 2d30ef33fc67d5a719251722ce56d7cc180abb67. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/api-client/src/v2/blocks/operation-block/helpers/build-request.ts
+++ b/packages/api-client/src/v2/blocks/operation-block/helpers/build-request.ts
@@ -134,11 +134,11 @@ export const buildRequest = ({
     /** Build the js request object */
     const request = new Request(proxiedUrl, {
       /**
-       * Ensure that patch is uppercase
+       * Ensure that all methods are uppercased (though only needed for patch)
        *
-       * @see https://github.com/JakeChampion/fetch/issues/254
+       * @see https://github.com/whatwg/fetch/issues/50
        */
-      method: method === 'patch' ? method.toUpperCase() : method,
+      method: method.toUpperCase(),
       headers,
       signal: controller.signal,
       body,


packages/api-client/src/v2/blocks/operation-block/helpers/build-request.ts
```diff
@@ -133,7 +133,12 @@ export const buildRequest = ({
 
     /** Build the js request object */
     const request = new Request(proxiedUrl, {
-      method,
+      /**
+       * Ensure that patch is uppercase
+       *
+       * @see https://github.com/JakeChampion/fetch/issues/254
+       */
+      method: method === 'patch' ? method.toUpperCase() : method,
```
PAHJunior: I understand the Fetch spec [intentionally doesn't normalize PATCH](https://github.com/whatwg/fetch/issues/50) to preserve HTTP method case-sensitivity, but since we're already normalizing it, why not uppercase **all** methods for consistency?

```typescript
method: method.toUpperCase(),
```

Is there a specific reason to keep the ternary?

amritk: not really, I did have it that way initially. But ended up just doing it only for patch after reading through the fetch issues. You think we should just uppercase them all?

amritk: Alright updated!

PAHJunior: Yes, I think uppercasing all methods is better!

**Testing Scalar's supported HTTP methods (excl. `trace`) on Node.js v22:**
```bash
$ node -e "['delete', 'get', 'head', 'options', 'patch', 'post', 'put'].forEach(m => console.log(m + ' -> ' + new Request('http://example.com', {method: m}).method))"

delete -> DELETE
get -> GET
head -> HEAD
options -> OPTIONS
patch -> patch       # ❌ + warning
post -> POST
put -> PUT
```

**Result: Only `patch` isn't normalized** 

**Scalar's supported methods:** [`@scalar/helpers/src/http/http-methods.ts`](https://github.com/scalar/scalar/blob/main/packages/helpers/src/http/http-methods.ts)
(note: `trace` is [forbidden by Fetch spec](https://github.com/nodejs/undici/blob/main/lib/web/fetch/constants.js#L88))

amritk: Sounds good

hanspagel (approved):
nice

PR #7686, opened 2026-01-09
fix(api-client): allow handling arrays and objects in codeInput

### LINKED ISSUE #6061: Incorrect rendering of example request with query list parameters
### What happens?

I have a simple `GET` endpoint that accepts via query parameter a list of values.
In FastAPI, it's declared like such:
```
@router.get(
    "",
)
async def bulk_availability(
    domains: Annotated[
        list[str],
        Query(
            ...,
            openapi_examples={
                "list": Example(
                    summary="A list of domains",
                    value=["example.com", "example.org"],
                )
            },
        ),
    ],
) -> list[str]:
    return domains
```

You'd call it like `https://example.com/v1/availability?domains=example.com&domains=example.org` and the `domains` would contain a list of `['example.com','example.org']`.

I've tried generating API docs using Scalar, but the example requests pass the list of values incorrectly as comma-delimited: `curl 'https://sandbox.scalar.com/v1/availability?domains=example.com%2Cexample.org'

Image

For each element of the example array, the parameter should be "duplicated" in the query string like this: `https://example.com/v1/availability?domains=example.com&domains=example.org`.

I've tested it with SwaggerUI and it's being shown correctly for the same OpenAPI file:

Image

Same for Redoc:

Image

So I think the OpenAPI file is correct and the rendering of the examples need to take into account how to properly handle array of parameters in a query string 🙏 

I've tried looking for open issues, but only found https://github.com/scalar/scalar/issues/1971 where support for examples was added. But I'm guessing it's depending on some default rendering/"serializing" that does not take into account the query context.

### OpenAPI Document

https://sandbox.scalar.com/e/Zlz3U

amritk:
## Problem

fixes #6061

We had an issue with some parameters rendering incorrectly, turns out that CodeInput was converting the array to a string

## Solution

Added functionality to codeinput so that it can handle arrays and objects

We should test the inputs of the client to be sure

## Checklist

- [x] I added a changeset. 
- [ ] I added tests.

---

> [!NOTE]
> Adds robust structured-value handling to inputs and updates tests accordingly.
> 
> - Implements generic typing and `serializeValue`/`deserializeValue` in `CodeInput` to preserve arrays, objects, numbers, booleans, and string primitives; updates change/submit flows to emit deserialized types
> - Updates `DataTableInputSelect` to support array multiselect (comma-joined values) and `DataTableInput`/form handlers to pass through typed values
> - Adjusts event typings/usages (e.g., `handleFieldChange`, emits) and fixes component refs/finders in tests
> - Expands tests for array/object serialization, primitive preservation, and select behavior; adds example test for array parameters; minor e2e assertion tweak and small key fallback cleanups in `get-example`
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit acfcd3859265703b0dc416ab9e8a0e69d27644be. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/api-reference/test/features/parameter-linking.e2e.ts
+++ b/packages/api-reference/test/features/parameter-linking.e2e.ts
@@ -104,7 +104,7 @@ test.describe('parameter linking', () => {
     await expect(button).toBeInViewport()
 
     // Should show a success message
-    await expect(page.getByRole('status').filter({ hasText: 'Copied' })).toBeVisible()
+    await expect(await page.getByRole('status').filter({ hasText: 'Copied' })).toBeVisible()
 
     // Scroll to the top of the page
     await page.evaluate(() => window.scrollTo(0, 0))


hanspagel (changes requested):
ah, what cursors says seems relevant

amritk:
Yepp I'll update the select logic

amritk:
@hanspagel I checked out but we use strings throughout so might as well just keep them as strings for the purpose of the select. Down the road we should upgrade it to an array to get better type safety

Re: cursor issues. We don't care about previously saved data since we don't save it right now. And for the other one going to actually parse by schema type! that should be good? haha 😓

PR #7693, opened 2026-01-09
feat(components): markdown line height

hanspagel:
## Problem

we have a nice line height for `<p>`'s, but if there’s content not in a `<p>`, e.g. in a `<div>`:

Screenshot 2026-01-09 at 14 10 10

it looks like this, irks!

Screenshot 2026-01-09 at 14 10 06

## Solution

set the line height for the container, not just for the `<p>`

@hwkr wdyt?

Screenshot 2026-01-09 at 14 10 53

## Checklist

- [ ] I added a changeset. 
- [ ] I added tests.

---

> [!NOTE]
> Improves markdown typography by standardizing line-height and heading styles.
> 
> - Sets `--markdown-line-height` and applies `line-height` on `.markdown` for consistent spacing across non-`p` content
> - Refactors headings (`h1–h6`) to use `--scalar-heading-*` sizes with explicit line-heights
> - Propagates `var(--markdown-line-height)` to paragraphs, lists, tables, summaries, and alert content where applicable
> - Adds changeset declaring patch bumps for `@scalar/api-reference` and `@scalar/components`
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit a4d4b03165580576beec8dd2bb96560c484ea124. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


hwkr (dismissed):
This seems reasonable to me 🤔 Looks like it affects the overall spacing though if we look at the snapshots. Not sure how best to tackle it. @cameronrohani might have ideas

hanspagel:
I think headings or something are missing an explicit line height

hanspagel:
lol, funny not funny:

those headings in markdown were just rubbish

* they just set variables for font-size and line-height, without actually setting a CSS attribute
* some headings didn’t have an explicit line height
* some of the variables are not even in the code base
* mixed CSS units

= absolute slop

I cleaned it up:

* using the actual font size variables
* every heading level has a line-height now

tested it with the long string example, I think it’s good

let’s see what the snapshot tests say now, they probably still fail, though

image

hanspagel:
COOL, let’s get this in

we can change the header sizes to use our variables too, but we should involve cam then, you’re right

PR #7701, opened 2026-01-10
feat(json-magic): configure external document extensions

robert-dean:
## Problem

This allows users of the `bundle` method to configure a custom extension to override the defaults. Now if a user doesn't want external documents to be put under the `x-ext` or `x-ext-url` path they can do the following:

```ts
        const output = await bundle(`${url}/base/openapi.json`, {
          plugins: [fetchUrls()],
          treeShake: false,
          urlMap: true,
          externalDocumentsKey: 'x-docs',
          externalDocumentsMappingsKey: 'x-docs-url',
        })
```

## Solution

Added `externalDocumentsKey` and `externalDocumentsMappingsKey` as optional strings to the `Config` type in bundle.ts. 
```ts
/**
 * Configuration options for the bundler.
 * Controls how external references are resolved and processed during bundling.
 */
type Config = {
   plugins: Plugin[]
   ...
  /**
   * Custom OpenAPI extension key used to store external references.
   * This key will contain all bundled external documents.
   * The key is used to maintain a clean separation between the main
   * OpenAPI document and its bundled external references.
   * @default 'x-ext'
   */
  externalDocumentsKey?: string

  /**
   * Custom OpenAPI extension key used to maintain a mapping between
   * hashed keys and their original URLs in x-ext.
   * This mapping is essential for tracking the source of bundled references
   * @default 'x-ext-urls'
   */
  externalDocumentsMappingsKey?: string

  compress?: (value: string) => Promise<string> | string
  ...
}
```
It will fallback to using the defaults from 
```ts
export const extensions = {
  externalDocuments: 'x-ext',
  externalDocumentsMappings: 'x-ext-urls',
} as const
```

## Checklist

- [x] I added a changeset. 
- [x] I added tests.

### DIFF
--- a/.changeset/clever-cases-grow.md
+++ b/.changeset/clever-cases-grow.md
@@ -1,5 +1,0 @@
----
-'@scalar/components': patch
----
-
-feat(components): simplify nested sidebar items styles
--- a/.changeset/strict-badgers-type.md
+++ b/.changeset/strict-badgers-type.md
@@ -1,5 +1,0 @@
----
-'@scalar/api-client': patch
----
-
-fix: ensure patch is uppercased
--- a/.changeset/young-news-happen.md
+++ b/.changeset/young-news-happen.md
@@ -1,7 +1,0 @@
----
-'@scalarapi/docker-api-reference': patch
-'@scalar/aspire': patch
-'@scalar/mock-server-docker': patch
----
-
-feat: update Docker Hub description
--- a/.github/workflows/update-contributors.yml
+++ b/.github/workflows/update-contributors.yml
@@ -11,6 +11,9 @@ jobs:
     if: ${{ github.repository_owner == 'scalar' }}
     runs-on: blacksmith-2vcpu-ubuntu-2204
     timeout-minutes: 15
+    permissions:
+      contents: write
+      pull-requests: write
 
     steps:
       - name: Update the list of contributors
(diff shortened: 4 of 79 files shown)


hanspagel:
Thanks for taking the time to put this together! What’s the use case for having it in another key? 🤔

robert-dean:
> Thanks for taking the time to put this together! What’s the use case for having it in another key? 🤔

The use case comes down to flexibility more than anything. If someone uses `x-ext` with their own internal meaning then the current implementation would conflict. This just makes it so if a user wants to have those external bundles in a different root path then it's a simple option. 

My own use case: When I bundle openapi specs into a single doc, being able to distinguish what was added by the bundler is useful to my parsing. Right now I'm doing the following:

```ts
    // moves everything under the bundled path to the component schema path
    const externalSchemas = schema['x-ext'];
    if (externalSchemas) {
      schema.components.schemas ??= {};
      for (const [key, value] of Object.entries(externalSchemas)) {
        schema.components.schemas[key] = value;
      }

      schema['x-ext'] = undefined;
    }
```

then I do 
```ts
const reference = schema.$ref.startsWith('#/x-ext/') ? schema.$ref.replace('#/x-ext/', '#/components/schemas/') : schema.$ref
```
in my internal parser. With this change I don't need to do either as I can just specify:
```ts
        const output = await bundle(`${url}/base/openapi.json`, {
          plugins: [fetchUrls()],
          treeShake: false,
          externalDocumentsKey: 'components/schemas',
        })
```

DemonHa (approved):
Thank you for this PR. LGTM 🔥

PR #7719, opened 2026-01-13
fix(openapi-parser): self-referencing causes infinite loop

SamyPesse:
## Problem

`dereference` is not robust enough and can get stuck in an infinite loop.

## Solution

The logic required to also keep track of all the refs resolved in a chain.
I also added an error when we identify an infinite recursion.

## Checklist

- [x] I added a changeset. 
- [x] I added tests.

### DIFF
--- a/packages/openapi-parser/src/utils/dereference.test.ts
+++ b/packages/openapi-parser/src/utils/dereference.test.ts
@@ -32,6 +32,41 @@ describe('dereference', () => {
     expect(result.schema.info.title).toBe('Hello World')
   })
 
+  it('handles circular references', () => {
+    const DOCUMENT = {
+      openapi: '3.1.0',
+      info: {
+        title: 'Circular Reference',
+        version: '1.0.0',
+      },
+      paths: {},
+      components: {
+        schemas: {
+          CircularReference: {
+            type: 'object',
+            properties: {
+              name: { type: 'string' },
+              parent: { $ref: '#/components/schemas/CircularReference' },
+            },
+          },
+        },
+      },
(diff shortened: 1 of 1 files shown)


packages/openapi-parser/src/utils/dereference.test.ts
```diff
@@ -32,6 +32,86 @@ describe('dereference', () => {
     expect(result.schema.info.title).toBe('Hello World')
   })
 
+  it('handles circular references', () => {
```
hanspagel, line 35: circular references work

packages/openapi-parser/src/utils/dereference.test.ts
```diff
@@ -32,6 +32,86 @@ describe('dereference', () => {
...
+    const circularReferenceSchema = result.schema.components.schemas.CircularReference
+
+    expect(circularReferenceSchema.properties.name.type).toBe('string')
+    expect(circularReferenceSchema.properties.parent.properties.name.type).toBe('string')
+
+    // Circular references can't be JSON.stringify'd (easily)
+    expect(() => JSON.stringify(result.schema)).toThrow()
+  })
+
+  it('throws an error for self-referencing schemas', () => {
```
hanspagel, line 70: self-references throw

hanspagel: this test causes an infinite loop without the fix, having the paths was actually necessary

hanspagel (approved):
@SamyPesse thanks for putting this together! 🤘 let’s get it out quickly

PR #7725, opened 2026-01-13
fix(components): support slot for nested items icon

hwkr:
Turns out I was missing a check here. Also added a full test suite for nested items.

## Checklist

- [x] I added a changeset. 
- [x] I added tests.

### DIFF
--- a/.changeset/funky-eagles-marry.md
+++ b/.changeset/funky-eagles-marry.md
@@ -1,0 +1,5 @@
+---
+'@scalar/components': patch
+---
+
+fix(components): support slot for nested items icon


packages/components/src/components/ScalarSidebar/ScalarSidebarNestedItems.vue
```diff
@@ -127,13 +127,15 @@ const handleBack = (event: MouseEvent) => {
...
-              weight="bold" />
-          </slot>
+          <ScalarIconLegacyAdapter
+            v-if="icon"
+            :icon
+            weight="bold" />
+          <slot
+            v-else
+            name="icon" />
         </template>
```
geoffgscott, line 138: I don't think we need all the conditionals unless you expect a slot and an icon
```
<slot name="icon">
  <ScalarIconLegacyAdapter
    v-if="icon"
    :icon
    weight="bold" />
</slot>
```

hwkr: Sure

hanspagel (approved):
2 lines code change
500 lines of tests added

that’s the way

😁

PR #7740, opened 2026-01-14
fix(api-client): filter out readOnly properties from requestBody

### LINKED ISSUE #7734: Read-only properties are being populated when testing requests with a body
### What happens?

When making a `POST` request, for example, and clicking on `Test Request`, it appears to use the full schema with the read-only properties populated in the body, rather than just the properties that can be set. This is currently happening on version 1.43.3

The body should only contain properties that can be sent as part of the request, not read-only ones. Previous versions did not seem to experience this problem

### OpenAPI Document

https://sandbox.scalar.com/e/WgHtY

amritk:
## Problem

closes #7734

We stopped filtering out readOnly properties from the requestBody

## Solution
The tech was there, was just missing passing the option! Added a test as well!

## Checklist

- [x] I added a changeset. 
- [ ] I added tests.

---

> [!NOTE]
> Ensures request body examples exclude `readOnly` fields.
> 
> - Update `get-request-body-example.ts` to call `getExampleFromSchema(schema, { mode: 'write' })`, filtering `readOnly` properties from generated examples
> - Adjust tests to reflect new example values and add a test verifying `readOnly` fields are omitted; remove unused mocks
> - Add patch changeset for `@scalar/api-client`
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 9f981267300440e366dd4c601fde05ce44c18f50. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/.changeset/solid-regions-shop.md
+++ b/.changeset/solid-regions-shop.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
+
+fix: filter out readonly properties from requestBody


hanspagel (approved):
sure thing

PR #7760, opened 2026-01-15
fix(api-reference): fix test on redis PR

amritk:
The solution was to move the api client creation up, it still seems to work ok. Also did some cleanup and addressed the cursor issue and added a test.

---

> [!NOTE]
> - **ApiReference**: Calls `apiClient.route` immediately after syncing slug/URL to initialize routing before document load; keeps rest of document load/persistence flow unchanged.
> - **Auth persistence**: Refactors `load-from-perssistance` with `SECRET_KEY_PREFIX`, streamlined `isSecretKey`, safer object checks, and clearer early returns; restores secrets only when structures exist; validates stored selected schemes against available schemes, clamps `selectedIndex`, and only sets `x-scalar-selected-security` when valid schemes remain.
> - **Tests**: Expands coverage for auth restore logic, including case where all stored schemes are filtered out; ensures default client loading behavior remains correct.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 82d2a564f85bd7f2c9d43d07b3db00726f05b393. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/api-reference/src/helpers/load-from-perssistance.ts
+++ b/packages/api-reference/src/helpers/load-from-perssistance.ts
@@ -113,15 +113,9 @@ const clampSelectedIndex = (selectedIndex: number, schemesLength: number): numbe
  */
 export const loadAuthSchemesFromStorage = (store: WorkspaceStore): void => {
   const slug = store.workspace['x-scalar-active-document']
-
-  if (!slug) {
-    console.warn('No active document found, skipping auth schemes loading')
-    return
-  }
-
   const activeDocument = store.workspace.activeDocument
 
-  if (!activeDocument) {
+  if (!activeDocument || !slug) {
     console.warn('Active document not found in workspace, skipping auth schemes loading')
     return
   }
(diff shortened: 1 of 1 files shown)


packages/api-reference/src/helpers/load-from-perssistance.ts
```diff
@@ -53,6 +62,20 @@ export const mergeSecrets = (current: unknown, stored: unknown): void => {
...
+ */
+const getActiveDocument = (store: WorkspaceStore) => {
+  const slug = store.workspace['x-scalar-active-document']
+
+  if (!slug) {
+    return null
+  }
+
+  return store.workspace.documents[slug] ?? null
+}
```
geoffgscott: The active document already exists on the workspace.  `store.workspace.activeDocument`

packages/api-reference/src/helpers/load-from-perssistance.ts
```diff
@@ -53,6 +62,20 @@ export const mergeSecrets = (current: unknown, stored: unknown): void => {
...
+
```
amritk:
```suggestion
```

packages/api-reference/src/helpers/load-from-perssistance.ts
```diff
@@ -63,82 +86,93 @@ export const mergeSecrets = (current: unknown, stored: unknown): void => {
...
-  const slug = store.workspace['x-scalar-active-document'] ?? ''
+  const slug = store.workspace['x-scalar-active-document']
+
   if (!slug) {
     console.warn('No active document found, skipping auth secrets loading')
     return
   }
 
-  const activeDocument = store.workspace.documents[slug]
+  const activeDocument = getActiveDocument(store)
```
amritk:
```suggestion
  const activeDocument = store.workspace.activeDocument
```

PR #7765, opened 2026-01-16
fix(api-client): enable `knip`

marcalexiei:
## Problem

- Followup of #7233

## Solution

- remove unneeded export from internal types and values
- remove unused files

I limited the changes to the minimum needed to resolve the Knip issues since you’re actively working on this package. 
The Biome errors can be addressed later.

## Checklist

- [x] I added a changeset. 
- [ ] I added tests.


hanspagel:
Thanks

PR #7778, opened 2026-01-16
fix(api-reference): do not persist auth when option if off

DemonHa:
Previously, authentication data was still being persisted even when the user explicitly disabled auth persistence in the configuration. While persistence was turned off, the existing auth state was never reset and continued to be saved unintentionally.

This PR fixes that behavior by ensuring that no authentication information is saved when auth persistence is disabled in the user configuration.

In addition, the playground has been updated to explicitly set persistAuth: true, making it easier to test and validate auth persistence behavior on Scalar Galaxy.

## Checklist

- [x] I added a changeset. 
- [ ] I added tests.

---

> [!NOTE]
> Ensures authentication is only persisted when explicitly enabled.
> 
> - Passes `persistAuth` from `ApiReference.vue` into `persistencePlugin` (defaulting to `false`)
> - Extends `persistencePlugin` to accept `persistAuth` (boolean or function) and skip auth persistence on document changes when disabled
> - Updates playground (`galaxy-scalar-com.ts` and tests) to set `persistAuth: true` for easier verification
> - Adds a changeset for `@scalar/api-reference` patch
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 1f451e4fd7ae06e6c35f310d49d2bef9fdf900d5. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/api-reference/src/plugins/persistance-plugin.ts
+++ b/packages/api-reference/src/plugins/persistance-plugin.ts
@@ -10,7 +10,7 @@ export const persistencePlugin = ({
   debounceDelay = 500,
   maxWait = 10000,
   prefix = '',
-  persistAuth: shouldPersistAuth = true,
+  persistAuth = false,
 }: {
   debounceDelay?: number
   /** Maximum time in milliseconds to wait before forcing execution, even with continuous calls. */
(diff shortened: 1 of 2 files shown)


amritk (approved):
lgtm!

PR #7781, opened 2026-01-17
fix(api-reference): enable `knip`

marcalexiei:
## Problem

- Followup of #7233

## Solution

- remove export from internal module values / types
- ~~create private workspace in `packages/api-reference/playground/vue` to avoid adding specific examples dependencies on `api-reference` package.json~~
- removed unused dependencies
- removed unused dev dependencies

## Checklist

- [x] I added a changeset. 
- [ ] I added tests.


hanspagel (approved):
awesooomee

PR #7805, opened 2026-01-20
chore(java): enable `knip`

marcalexiei:
## Problem

- Followup of #7233

## Solution

- removed unused dev dependencies
  - `@scalar/build-tooling`
  - `@scalar/snippetz` 
    (`generate:java-enums` is located in this package so there is no reference to it here)

## Checklist

- [ ] I added a changeset. 
- [ ] I added tests.

### DIFF
--- a/knip.jsonc
+++ b/knip.jsonc
@@ -310,7 +310,12 @@
     },
     "integrations/java": {
       "ignoreDependencies": [
-        "@scalar/api-reference" // build step copy standalone file from this package
+        "@scalar/api-reference", // build step copy standalone file from this package
+
+        // Included for convenience.
+        // Java enums are dynamically generated from the snippetz package.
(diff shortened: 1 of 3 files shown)


knip.jsonc
```diff
@@ -306,9 +305,19 @@
...
       ]
+    },
+    "integrations/java": {
+      "ignoreDependencies": [
+        "@scalar/api-reference", // build step copy standalone file from this package
+
+        // Included for convenience.
+        // Java enums are dynamically generated from the snippetz package.
+        // Having snippetz as a dependency ensures they are regenerated whenever clients or targets change.
+        "@scalar/snippetz"
```
marcalexiei, line 318: Thanks for the explanation!
I restored `@scalar/snippetz` as dev dependency and added a comment about the motivation in `knip` config.

xC0dex (approved):
Awesome!

PR #7811, opened 2026-01-20
refactor(fastify): move fastify theme to package

### LINKED ISSUE #3566: `@scalar/fastify-api-reference` doesn't seem to work when bundling
### What happens?

Hi, we're currently using self-hosted Scalar with the React integration for OpenAPI docs, and wanted to replace this with serving the docs directly from our Fastify server, as we'll need to document several API versions, and each one being served directly from Fastify would be easier and doesn't require setting up an app.

We've tried using `@scalar/fastify-api-reference`, and this works great locally. But for production, we're bundling our code and running it from a Docker container, and it seems to not work in this state, with the following error:

scalar js issue with bundling

We expected it to work just like locally, but it doesn't. I'm guessing this "standalone" JS file is expected to be served with the Fastify API or available in the filesystem, but this isn't the case when bundling.

Note: This is a duplicate of the [Discord thread](https://discord.com/channels/1135330207960678410/1263945110052274187/threads/1296449225369976903) I opened a few days ago.

### OpenAPI Document

_No response_

hanspagel:
* we once moved the fastify framework theme to the themes package to have it in the client.
* we found out, we don’t necessarily need it in the client and want to have it in the framework package instead (same as hono, docusaurus and the others…)

This PR:
- removes the fastify and elysiajs themes from the themes package
- applies a few of the best practices to the fastify integration (filenames and such)
- finally bundles the CSS and JS, instead of this quirky asset loading from the dist folder
- fixes #3566 (finally)

---

> [!NOTE]
> **Medium Risk**
> Touches CI/release automation and refactors the Fastify integration’s asset bundling and theme injection, which could impact packaging and runtime behavior. Also introduces a data/config migration that changes persisted theme values, so regressions would affect existing users’ settings.
> 
> **Overview**
> Moves the **Fastify UI theme and bundled `scalar.js` asset handling into `integrations/fastify`**, switching the integration to bundle `?raw` CSS/JS imports via `esbuild` and injecting the Fastify theme directly when rendering the HTML.
> 
> Removes the `fastify`/`elysiajs` themes from `@scalar/themes` and the API client settings UI, and adds **backwards-compat migrations**: stored workspace/theme values are migrated to `default` (`oas-utils` data version `2.6.0`) and `@scalar/types` now preprocesses deprecated config themes to `default` with warnings.
> 
> Updates CI to build `@scalar/api-reference` for integration tests, adds a new `scalar-app` e2e job, and introduces conditional `todesktop` and Cloudflare Pages deploy jobs for the API client.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 9794222cbb55d5d0ad1d576412fa386c6b2d2fec. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


amritk (commented):
The data migration is just for the theme stuff eh!? I gotta write the new big migration soon.

packages/themes/src/index.ts
```diff
@@ -181,20 +171,6 @@ export const presets = {
     theme: laserwaveTheme,
     slug: 'laserwave',
   },
-  elysiajs: {
```
hwkr, line 184: The only thing is technically people are now able to use these in docs and removing them could break their themes. Maybe @geoffgscott or @bgrcs can check if anyone is using these right now?

hanspagel: ah, I can write a migration for the configuration

hanspagel: done

marclave (approved):

hanspagel:
repo moved on, nothing in this PR will work anymore :))

PR #7812, opened 2026-01-20
fix(docs): slider flickers on wide screens

liamdscalar:
## Problem

The scroll does not work great on really wide screens.

## Solution

Set the intersection ratio to 1 to prevent it from flickering between two high ratio gallery items.

before:

https://github.com/user-attachments/assets/1e5cce1f-0ce6-4c61-a3e5-4f112a9da462

after:

https://github.com/user-attachments/assets/b8cba109-2eb9-4c05-b51c-16df66591e27

## Checklist

- [ ] I added a changeset. 
- [ ] I added tests.


hanspagel:
send it

PR #7814, opened 2026-01-20
fix(api-reference): additional properties generate invalid example

hanspagel:
## Problem

When a schema has `additionalProperties` (properties we don't know the name of), we generate invalid JSON where property names include a `*`.

Screenshot 2026-01-20 at 15 27 17

## Solution

* Remove the asterisk.
* Don’t render `[Circular reference]`

## Checklist

- [x] I added a changeset. 
- [ ] I added tests.

---

> [!NOTE]
> - **Example generation:** Uses `additionalProperty` (or trimmed `x-additionalPropertiesName`) for `additionalProperties` keys without appending `*`; updates tests in `api-client` and `oas-utils` accordingly.
> - **Circular references:** Skips circular refs entirely (returns `undefined`/omits keys) instead of returning `'[Circular Reference]'`; updates related tests.
> - **UI:** Changes additional-properties badge label to "unknown property name" in `SchemaProperty.vue` and updates its test.
> - **Meta:** Adds changeset entries for `@scalar/api-reference`, `@scalar/api-client`, and `@scalar/oas-utils`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit ffaf1e57ae35fea196f1b0ced90be008d38d343a. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/api-client/src/v2/blocks/operation-code-sample/helpers/get-example-from-schema.test.ts
+++ b/packages/api-client/src/v2/blocks/operation-code-sample/helpers/get-example-from-schema.test.ts
@@ -1146,7 +1146,7 @@ describe('getExampleFromSchema', () => {
           }),
         ),
       ).toMatchObject({
-        'sensorId*': 1,
+        'sensorId': 1,
       })
 
       expect(
@@ -1160,7 +1160,7 @@ describe('getExampleFromSchema', () => {
           }),
         ),
       ).toMatchObject({
-        'isActive*': true,
+        'isActive': true,
       })
     })
 
@@ -1186,7 +1186,7 @@ describe('getExampleFromSchema', () => {
           }),
         ),
       ).toMatchObject({
-        'metadata*': {
+        'metadata': {
           key: 'version',
           value: '1.0.0',
         },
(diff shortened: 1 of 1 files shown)


packages/oas-utils/src/spec-getters/get-example-from-schema.ts
```diff
@@ -256,7 +256,7 @@ const handleObjectSchema = (
       'x-additionalPropertiesName' in additional &&
       typeof additional['x-additionalPropertiesName'] === 'string' &&
       additional['x-additionalPropertiesName'].trim().length > 0
-        ? `${additional['x-additionalPropertiesName'].trim()}*`
```
liamdscalar, line 259: Any insight into why this was here to begin with @hanspagel ?

hanspagel: I think I copied what redocly did, which is never a good idea

It was supposed to indicate it doesn't have to be the exact string, but any string

PR #7819, opened 2026-01-20
feat(api-reference): hide responses without content

### LINKED ISSUE #6574: hide example responses that don’t have any content
> At points, the API theme expects info we don't have (for example, client library examples, examples for various error codes) Arguably we should have this (especially for error codes), but until we do, it's likely to be annoying for users to click through to empty info. Is there a way to hide unused elements?

I think there’s a point here. We can still list the error codes on the left, but we can hide responses on the right if there’s no example or schema for it.

hanspagel:
## Problem

All responses are listed on the left.

No need to show them on the right side when they don't have a schema or example.

## Solution

Let’s filter out the empty responses and not render a tab for those.

Fixes #6574

## Checklist

- [x] I added a changeset. 
- [ ] I added tests.

---

> [!NOTE]
> Filters response tabs to only show entries with displayable content and stabilizes tab selection when the list changes.
> 
> - Adds `hasResponseContent` to detect content (schema/example/examples) with mime-type prioritization
> - Updates `ExampleResponses.vue` to compute `statusCodesWithContent`, use it for tabs/rendering, and `watch` it to clamp `selectedResponseIndex` and reset `selectedExampleKey`
> - Renders card/tabs only when `statusCodesWithContent.length` is non-zero; `currentResponse` now derives from filtered codes
> - Adds unit tests for `hasResponseContent`
> - Adds a patch changeset for `@scalar/api-reference`
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit a675f50bc91edbca08b22701565d45785c27e9ab. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/api-reference/src/features/example-responses/ExampleResponses.vue
+++ b/packages/api-reference/src/features/example-responses/ExampleResponses.vue
@@ -61,12 +61,17 @@ const selectedResponseIndex = ref<number>(0)
  * Clamp the selected index when the filtered list shrinks.
  * Without this, the index can become out of bounds and cause a mismatch
  * between the visible tabs and the displayed content.
+ *
+ * We also reset `selectedExampleKey` to match the behavior of `changeTab`,
+ * since the new response may not have the same example keys.
  */
 watch(statusCodesWithContent, (codes) => {
   if (codes.length === 0) {
     selectedResponseIndex.value = 0
+    selectedExampleKey.value = ''
   } else if (selectedResponseIndex.value >= codes.length) {
     selectedResponseIndex.value = codes.length - 1
+    selectedExampleKey.value = ''
   }
 })
 


amritk (changes requested):
Nice, cursor issue seems legit?

packages/api-reference/src/features/example-responses/ExampleResponses.vue
```diff
@@ -46,13 +46,38 @@ const orderedStatusCodes = computed<string[]>(() =>
...
+
+  const mediaType =
+    normalizedContent?.['application/json'] ??
+    normalizedContent?.['application/xml'] ??
+    normalizedContent?.['text/plain'] ??
+    normalizedContent?.['text/html'] ??
+    normalizedContent?.['*/*'] ??
+    normalizedContent?.[keys[0] ?? '']
+
+  return !!(mediaType?.schema || mediaType?.example || mediaType?.examples)
```
amritk:
```suggestion
  return Boolean(mediaType?.schema || mediaType?.example || mediaType?.examples)
```
nit: the cool people use Boolean LOL

amritk: I would even move this into its own helper + tests

PR #7820, opened 2026-01-20
fix(api-reference): correctly restore the auth from persistance

DemonHa:
Previously, we were only restoring secret keys.

This PR updates the restore logic to handle all keys, with the exception of the top-level type field. The type field is intentionally excluded because it acts as the union discriminator for the security scheme.

The restore behavior now performs a smart merge: keys are merged only if they already exist on the target object; missing keys are skipped to avoid introducing unintended fields.

## Checklist

- [x] I added a changeset. 
- [x] I added tests.

---

> [!NOTE]
> Expands auth restoration to merge full security scheme values while preserving the top-level discriminator.
> 
> - Replace `mergeSecrets` with `mergeSecuritySchemas` to merge all truthy primitive fields (not just `x-scalar-secret-*`), skipping top-level `type` and allowing nested `type` updates
> - Always call `restoreAuthSecretsFromStorage` inside `loadAuthSchemesFromStorage`; only restore `x-scalar-selected-security` if not already set; validate/filter stored schemes and clamp selected index
> - Update tests to reflect new merge behavior (including overwriting document values from storage, nested handling, falsy-value skips, and top-level `type` protection) and rename usages
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 05cc6fd7690314e67b2dad26035c9cf5e7a55d8e. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/api-reference/src/helpers/load-from-perssistance.test.ts
+++ b/packages/api-reference/src/helpers/load-from-perssistance.test.ts
@@ -95,43 +95,7 @@ describe('mergeSecuritySchemas', () => {
     })
   })
 
-  it('replaces nested objects with truthy stored values', () => {
-    const current = {
-      type: 'oauth2',
-      flows: {
-        authorizationCode: {
-          authorizationUrl: 'https://example.com/oauth/authorize',
-          tokenUrl: 'https://example.com/oauth/token',
-          'x-scalar-secret-client-id': '',
-        },
-      },
-    }
-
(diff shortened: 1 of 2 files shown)


amritk (commented):
I feel like we shouldn't change the in or name either but thats subjective

PR #7906, opened 2026-01-26
feat(api-client): remove parsing from CodeInput

### LINKED ISSUE #7863: Sending a request with a parameter that matches a JSON object results in either nothing being submitted or [object Object] being sent
### What happens?

I'm upgrading an app from dotnet 9 w/ scalar 2.5.11 to dotnet 10 with the latest scalar release (2.12.11).

I have a GET request with 2 simple string parameters and a parameter that accepts a small JSON object payload as text.  In the scalar UI when i fill the parameter with text that results in valid JSON, the client-side JS seemingly attempts to submit it  as a JS value.  For example,

If I specify the following:

Image

In the network tab of my browser it sends the following to the endpoint (NOTE the [object Object] for the filter param):

filter=%5Bobject+Object%5D&startDate=2026-01-22&endDate=2026-01-22

If I prefix the "filter" parameter with something that makes it invalid JSON, it just submits it as expected:

Image

Here's the parameters as seen in the browser network tab:

filter=x%7B%22foo%22%3A%22bar%22%7D&startDate=2026-01-22&endDate=2026-01-22

I would expect it to work as it did in the old version where the JSON string is passed as-is, or perhaps validated, but still passed as the stringified text.

### OpenAPI Document

_No response_

amritk:
## Problem

fixes #7863

The previous thought was to have parsing in the code input to support objects and arrays, however that ran into some issues. 

## Solution

~The new solution is to return codeinput to deal with strings only, but we parse in getExample so it affects the code sample as well as the build parameters function. So now we only json parse strings when we have a schema or content type that requires it.~

Did it properly and added the parsing to parameter processing/building

## Checklist

- [x] I added a changeset. 
- [x] I added tests.

---

> [!NOTE]
> Reworks input/serialization flow so UI emits strings and parsing happens where schema/content is known.
> 
> - **`CodeInput`** now always emits string values; removed internal deserialization; updated emits/handlers and tests
> - **Parameter/request building**: added `deSerializeParameter` and integrated it into `build-request-parameters` and HAR `process-parameters` to JSON-parse example strings based on parameter `schema` or `content`; expanded tests for arrays/objects, primitives, and example selection
> - **`getExample`** enhanced to return appropriate example/default/enum values without UI parsing; added extensive tests
> - **Request body**: tests for primitive JSON examples and content-type selection
> - **Types**: exported `ParameterWithContentObject` and `ParameterWithSchemaObject` from workspace store
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 58a92a414f1b1d5e577a89c44366854a214ffc7a. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/api-client/src/v2/blocks/operation-block/helpers/build-request-body.test.ts
+++ b/packages/api-client/src/v2/blocks/operation-block/helpers/build-request-body.test.ts
@@ -36,7 +36,7 @@ describe('buildRequestBody', () => {
     }
     const env = { username: 'john_doe', userId: '12345' }
     const result = buildRequestBody(requestBody, env, 'default')
-    expect(result).toBe('{"name":"john_doe","id":"12345"}')
+    expect(result).toBe('{"name": "john_doe", "id": "12345"}')
   })
 
   it('builds FormData for multipart/form-data content type', () => {


amritk:
This saved code duplication and was quick, but isn't the best place for it. Going to move it into the parameter sererializers

PR #7925, opened 2026-01-27
fix(api-client): cannot collapse a document in sidebar

amritk:
## Problem

When you collapsed a document in the sidebar it was getting re-opened by the router.

## Solution

Moved the isSelected check higher up as it was previously only applied to the operation.

## Checklist

- [x] I added a changeset. 
- [ ] I added tests.

---

> [!NOTE]
> Ensures sidebar items can be collapsed without being re-expanded by navigation.
> 
> - Moves the selected-item toggle check to the top of `handleSelectItem`, applying to all item types; prevents routing when re-clicking the already selected item and enables collapsing documents
> - Removes the now-redundant toggle logic from the `operation` branch
> - Adds a patch changeset for `@scalar/api-client`
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit d5513321fcd78e7fe9d9761b0cc8a6756a62830d. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/api-client/src/v2/features/app/app-state.ts
+++ b/packages/api-client/src/v2/features/app/app-state.ts
@@ -428,7 +428,7 @@ const handleSelectItem = (id: string) => {
     return
   }
 
-  // If we are already in the operation, just toggle expansion
+  // If we are already in the nav item, just toggle expansion
   if (sidebarState.isSelected(id)) {
     sidebarState.setExpanded(id, !sidebarState.isExpanded(id))
     return


packages/api-client/src/v2/features/app/app-state.ts
```diff
@@ -428,6 +428,12 @@ const handleSelectItem = (id: string) => {
     return
   }
 
+  // If we are already in the operation, just toggle expansion
```
DemonHa: can you update comment since this is for any entry

PR #7929, opened 2026-01-27
fix(api-reference): drop changes to the document when closing the client modal

### LINKED ISSUE #7927: Query Parameters Disappear From Main Reference Screen
### What happens?

I'm using @scalar/api-reference-react 0.8.34 to render an OpenAPI 3.0 document via the "content" property. On the main screen it shows the query parameters and the descriptions fine for each operation. If you click "Test Request" and then delete one of the query parameters in the client window it is also deleted on the main screen.

I expect the query parameters to continue to be listed on the main reference screen even if you delete them in the client window. On the demo site you can delete parameters on the client and they still remain on the main documentation screen, but it isn't working when I use the react component.

### OpenAPI Document

_No response_

amritk:
## Problem

fixes #7927

Now that we are using the same store between the client and references, changes in one reflect in the other.

## Solution

We drop the changes when closing the modal so that the references are unaffected.

## Checklist

- [x] I added a changeset. 
- [x] I added tests.

---

> [!NOTE]
> Ensures any edits made in the client modal do not leak into the shared store after the modal is dismissed.
> 
> - In `Modal.vue`, `cleanUp()` now calls `workspaceStore.revertDocumentChanges(document.value?.['x-scalar-navigation']?.name ?? '')` on close and on unmount
> - Adds tests in `Modal.test.ts` verifying document mutations are reverted when the modal closes and when the component unmounts
> - Adds a changeset for `@scalar/api-client` patch release
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit fe51c46932363d05a07d2a6ea12876dcbb575f3b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/api-client/src/v2/features/modal/Modal.test.ts
+++ b/packages/api-client/src/v2/features/modal/Modal.test.ts
@@ -506,7 +506,7 @@ describe('Modal', () => {
   })
 
   it('drops document changes when modal is closed', async () => {
-    const { props, modalState, store, document } = await createModalProps()
+    const { props, modalState, document } = await createModalProps()
 
     // Open the modal
     modalState.open = true


packages/api-client/src/v2/features/modal/Modal.vue
```diff
@@ -126,6 +126,11 @@ const { activate: activateFocusTrap, deactivate: deactivateFocusTrap } =
 const cleanUp = () => {
   deactivateFocusTrap()
   eventBus.emit('operation:cancel:request')
+
+  // We also drop any changes to the document when we close the modal
+  workspaceStore.revertDocumentChanges(
+    document.value?.['x-scalar-navigation']?.id ?? '',
```
DemonHa: we need to use the name not the id here, sometimes id = name but in other cases not

amritk: lol we should fix that

packages/api-client/src/v2/features/modal/Modal.vue
```diff
@@ -126,6 +126,11 @@ const { activate: activateFocusTrap, deactivate: deactivateFocusTrap } =
 const cleanUp = () => {
   deactivateFocusTrap()
   eventBus.emit('operation:cancel:request')
+
+  // We also drop any changes to the document when we close the modal
+  workspaceStore.revertDocumentChanges(
+    document.value?.['x-scalar-navigation']?.name ?? '',
+  )
```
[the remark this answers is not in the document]
hanspagel: I don’t think that’s an issue

PR #7934, opened 2026-01-28
fix(api-client): z-index issues in addressbar

amritk:
## Problem

We got a lot of z-index going on, turns out it was to make the addressbar components pop up on top of the menu overlay.

## Solution

I removed all the z-indexes, but had to do some tomfoolery myself to remove the bottom border radius when the dropdowns are open. Might be a bit extra but maintains the same style as before. If we wanted to detach the dropdown we could remove the shennanigans.

image

Also fixed the z-index issue with an important, open to better ideas @hwkr 

image

## Checklist

- [x] I added a changeset. 
- [ ] I added tests.

---

> [!NOTE]
> **Low Risk**
> Primarily UI layering/styling changes to address-bar dropdowns plus a global z-index override for Radix popper wrappers; low functional risk but could affect overlay stacking across the app.
> 
> **Overview**
> Fixes address bar layering issues by **removing several address-bar-specific z-index classes** and instead tracking server/history dropdown open state to adjust styling (e.g., drop the bottom border radius while a dropdown is open) via new `update:open` events.
> 
> Updates `ServerDropdown` and `AddressBarHistory` to emit their `open` state (using `ValueEmitter`, now with an `unmount` event) and tweaks popover/backdrop styling/placement to avoid needing high z-index stacking. Also adds a **global CSS override** setting `[data-radix-popper-content-wrapper] { z-index: 1 !important; }` to prevent Radix context menu poppers from being hidden.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 402140c447f0498251d3c5f76d9b252ed4cf3655. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/components/src/components/ScalarContextMenu/ScalarContextMenu.vue
+++ b/packages/components/src/components/ScalarContextMenu/ScalarContextMenu.vue
@@ -24,16 +24,13 @@ const props = withDefaults(
 </script>
 
 <template>
-  <ContextMenuRoot
-    class="bg-red"
-    container="#headlessui-portal-root"
-    style="background-color: red">
+  <ContextMenuRoot>
     <ContextMenuTrigger
       :class="triggerClass"
       :disabled="props.disabled">
       <slot name="trigger" />
     </ContextMenuTrigger>
-    <ContextMenuPortal class="bg-blue">
+    <ContextMenuPortal>
       <ContextMenuContent
         :align="props.align"
         :side="props.side"


packages/api-client/src/v2/blocks/scalar-address-bar-block/components/AddressBar.vue
```diff
@@ -181,6 +181,14 @@ const copyUrl = async () => {
   )
 }
 
+const isServerDropdownOpen = ref(false)
+const isHistoryDropdownOpen = ref(false)
+
+/** Whether either dropdown is open */
+const isDropdownOpen = computed(
```
DemonHa, line 188: do you need both of them separately? maybe you can just merge them into one

DemonHa (approved):
LGTM

packages/api-client/src/v2/blocks/scalar-address-bar-block/components/AddressBar.vue
amritk: Haha I did but cursor complained so I did this. Both work but this is a bit safer in case one closes the other

PR #7938, opened 2026-01-28
fix(nestjs): enable `knip`

marcalexiei:
## Problem

- Followup of #7233

## Solution

- remove unused dev dependency 
   - `vite`
   - `@scalar/api-reference`
- use named instead of star exports
- added run between `pnpm` and `dev` to avoid that `knip` detecting `dev` as binary
- sort imports to match biome config (no biome errors reported for this package)
- resolve `knip` configuration hint 
   ```text
   Configuration hints (1)
   src/v2/features/command-palette/components/index.ts  packages/api-client  knip.jsonc  Refine entry pattern (no matches)
   ``` 
   (file no longer exists, removed in #7926)

## Checklist

- [x] I added a changeset. 
- [ ] I added tests.

### DIFF
--- a/.changeset/curvy-phones-design.md
+++ b/.changeset/curvy-phones-design.md
@@ -1,0 +1,5 @@
+---
+'@scalar/nestjs-api-reference': patch
+---
+
+fix: use named instead of star exports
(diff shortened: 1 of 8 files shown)


hanspagel (approved):
nice

PR #7963, opened 2026-01-28
feat(helpers): is-object unification

amritk:
## Problem

We had like 3-4 different isObject helpers.

## Solution

This deletes all those and uses one which actually checks for a record using the prototype (disallows dates etc) which is helpful when trying to ...spread.

## Checklist

- [x] I added a changeset. 
- [x] I added tests.

---

> [!NOTE]
> **Medium Risk**
> Broad behavioral change to a widely-used type guard: code paths that previously accepted non-plain objects (e.g., `Date`, class instances) will now treat them as non-objects, which could alter bundling/parsing/proxy behavior if such values appear.
> 
> **Overview**
> Consolidates `isObject` usage across `helpers`, `json-magic`, `openapi-parser`, and `workspace-store` by switching call sites to `@scalar/helpers/object/is-object` and removing duplicated local helpers/tests.
> 
> Changes `helpers`’ `isObject` semantics to only treat *record/plain objects* as objects (prototype is `Object.prototype` or `null`), with updated tests to assert `false` for `Date`, `Map/Set`, `Promise`, and class instances; `json-magic` also adds a `hasRef` guard to safely detect `$ref` nodes. Adds `@scalar/helpers` as a dependency to `openapi-parser` and includes a changeset bump for affected packages.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 9c75f47128a7c7f7051bc423a79578eb821ed836. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/json-magic/src/bundle/bundle.ts
+++ b/packages/json-magic/src/bundle/bundle.ts
@@ -13,7 +13,7 @@ import { isYaml } from '../helpers/is-yaml'
 import { getHash, uniqueValueGeneratorFactory } from './value-generator'
 
 /** Type guard to check if a value is an object with a $ref property */
-const hasRef = (value: unknown): value is Record<'$ref', string> =>
+const hasRef = (value: unknown): value is UnknownObject & Record<'$ref', string> =>
   isObject(value) && '$ref' in value && typeof value['$ref'] === 'string'
 
 /**


packages/json-magic/src/bundle/bundle.ts
```diff
@@ -789,7 +789,9 @@ export async function bundle(input: UnknownObject | string, config: Config) {
     const id = getId(root)
 
     if (typeof root === 'object' && '$ref' in root && typeof root['$ref'] === 'string') {
-      const ref = root['$ref']
+      /** Type assertion: since we switched isObject helpers and we know root has $ref at this point */
+      const rootWithRef = root as UnknownObject & Record<'$ref', string>
```
DemonHa: can we get rid of this type casting

DemonHa (approved):

PR #7972, opened 2026-01-29
fix(workspace-store): quick fix for sibling props

### LINKED ISSUE #7956: Fix our parsing on ref-value unions
### Discussed in https://github.com/scalar/scalar/discussions/7955

<sup>Originally posted by **yegork78** January 28, 2026</sup>
Hi!
Before reporting as a bug, wanted to make sure. In my swagger I have endpoints with path parameters, like this:

```
{
  "openapi":"3.1.0",
  "paths":{
    "/api/v3/account/edit":{
      "post":{
        "tags":["Account"],
        "summary":"Update account details",
        "description":"Updates account details including first name, last name, and phone",
        "parameters":[
          {
            "name":"first-name",
            "in":"query",
            "description":"First name",
            "required":true,
            "style":"form",
            "explode":false,
            "allowReserved":false,
            "schema":{"type":"string"}
          },
        ]
      }
    }
  }
}
```

Up until version 1.43.13 they were presented correctly in the body of the endpoint description:
image

In version 1.43.14 and on they are all missing:
image

They are still presented correctly in the [Test Request]:
image

Am I doing something wrong or is it a bug?
Example swagger available here: https://dev.air-dr.com/v3/api-docs/gateway (though it's pretty big)</div>

amritk:
## Problem

fixes #7956

Good ol sibling props has come up again. In this case the base param was validated against the schema but the ref-value was not, making our types incorrect and causing the bug.

## Solution

This is a quick fix for sibling params and _should_ be type safe now. However in the future we should ensure that we parse the schema again after merging these properties for the best type safety.

## Checklist

- [x] I added a changeset. 
- [x] I added tests.

---

> [!NOTE]
> **Low Risk**
> Small, localized change to `getResolvedRef` default behavior for object refs; risk is limited to potential edge-case differences in merge precedence for consumers relying on previous `$ref-value`-only output.
> 
> **Overview**
> `getResolvedRef` now merges *sibling properties* on a `$ref` node with the resolved `'$ref-value'` when the ref value is an object, instead of returning only the ref value.
> 
> Adds a unit test covering the sibling-props merge behavior and includes a changeset to publish a patch release for `@scalar/workspace-store`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 0771653174bdfec6a5b8c8a9f8842d6f7d9e3b39. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/workspace-store/src/helpers/get-resolved-ref.ts
+++ b/packages/workspace-store/src/helpers/get-resolved-ref.ts
@@ -1,7 +1,20 @@
+import { isObject } from '@scalar/helpers/object/is-object'
+
 export type RefNode<Node> = Partial<Node> & { $ref: string; '$ref-value': Node }
 export type NodeInput<Node> = Node | RefNode<Node>
 
-const defaultTransform = <Node>(node: RefNode<Node>) => {
+/** Default transform function that overwrites the $ref properties with the $ref-value properties */
+const defaultTransform = <Node>(node: RefNode<Node>): Node => {
+  // For objects we should merge the $ref-value properties with the node properties
+  if (isObject(node['$ref-value'])) {
+    const { $ref: _, '$ref-value': refValue, ...rest } = node
+    return {
+      ...rest,
+      ...refValue,
+    } as Node
+  }
+
+  // Otherwise we just return the ref-value
   return node['$ref-value']
 }
 
--- a/.changeset/ninety-weeks-guess.md
+++ b/.changeset/ninety-weeks-guess.md
@@ -1,0 +1,5 @@
+---
+'@scalar/workspace-store': patch
+---
+
+fix: merge sibling params in getResolvedRef
--- a/packages/workspace-store/src/helpers/get-resolved-ref.test.ts
+++ b/packages/workspace-store/src/helpers/get-resolved-ref.test.ts
@@ -84,6 +84,23 @@ describe('get-resolved-ref', () => {
       expect(result).toEqual({ id: 1, name: 'John' })
     })
 
+    it('should merge sibling properties with $ref-value properties', () => {
+      const refNode = {
+        $ref: '#/components/schemas/User',
+        '$ref-value': { id: 1, name: 'John' },
+        description: 'A user object',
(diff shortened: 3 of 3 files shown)


packages/workspace-store/src/helpers/get-resolved-ref.ts
```diff
@@ -1,7 +1,20 @@
...
 export type RefNode<Node> = Partial<Node> & { $ref: string; '$ref-value': Node }
 export type NodeInput<Node> = Node | RefNode<Node>
 
-const defaultTransform = <Node>(node: RefNode<Node>) => {
+/** Default transform function that overwrites the $ref properties with the $ref-value properties */
+const defaultTransform = <Node>(node: RefNode<Node>): Node => {
+  // For objects we should merge the $ref-value properties with the node properties
+  if (isObject(node['$ref-value'])) {
+    const { $ref: _, '$ref-value': refValue, ...rest } = node
+    return {
```
DemonHa, line 11: This behavior is incorrect. Our intent is for the base properties to take precedence over the $ref properties. However, in this scenario TypeBox’s coercion introduces extra required properties at the top level, which can unintentionally override properties coming from the $ref—something we explicitly want to avoid.

We can possibly add this as a quick fix 🥶? @geoffgscott

DemonHa (changes requested):

packages/workspace-store/src/helpers/get-resolved-ref.ts
geoffgscott: No let's find the correct fix.

amritk: The correct fix would be to not do any validation of refs beforehand and only do it post merging inside of get resolved ref. This would also give a performance boost. But it's going to be a bigger lift since it requires a bunch of schema changes.

The quick fix is better than what we have now with a one liner

amritk:
GG

PR #7979, opened 2026-01-29
fix(api-reference): prevent re-rendering of ready items in lazy-bus

maxehmookau:
## Problem

Fixes a performance regression introduced in #7497 where large API specs experience severe slowdowns due to items being reprocessed on every scroll or interaction. We were seeing our spec unusable (scrolling stopped working) in Firefox & Chrome after a few clicks.

- Restore readyQueue guard in addToPendingQueue
- Add priorityQueue guard in addToPriorityQueue to prevent duplicates
- Skip adding to readyQueue in processQueue if item already ready

This preserves the callback fix from #7497 while preventing re-rendering.

## Checklist

- [x] I added a changeset. 
- [ ] I added tests.

Made with 🦊 at GitLab.

### DIFF
--- a/.changeset/fix-lazy-bus-performance.md
+++ b/.changeset/fix-lazy-bus-performance.md
@@ -1,0 +1,15 @@
+---
+'@scalar/api-reference': patch
+---
+
+fix: prevent re-rendering of already-ready items in lazy-bus queue
+
+Restores the readyQueue guard in addToPendingQueue to prevent items that are
+already rendered from being re-added to the pending queue. This fixes a
+performance regression introduced in #7497 where large API specs would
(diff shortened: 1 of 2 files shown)


maxehmookau:
Sweet, thanks for the speedy turnaround folks. ❤️  We (GitLab) are currently working on creating a canonical OpenAPI 3 spec. It's _large_, and whilst we're working on trimming the size down where we can, it's always going to be large! We'll upstream any performance improvements that we discover during our work. If you're curious our spec, as it stands today, is here: https://gitlab.com/gitlab-org/gitlab/-/blob/master/doc/api/openapi/openapi_v3.yaml

PR #7987, opened 2026-01-29
feat(api-reference): chat with operations

cameronrohani:
ask agent in refs 

https://github.com/user-attachments/assets/521c6919-d9ab-4de8-8f52-82865e2c9a0b


bgrcs:
If we're doing this in the request, we need to pass in the method + path of the endpoint at least to the chat @marclave @cameronrohani , as the user is expecting info about that endpoint specifically

hanspagel (changes requested):
cool! what about classic layout?

marclave:
not worrying about classic layout right now @hanspagel

marclave:
moving to https://github.com/scalar/scalar/pull/8002

PR #7988, opened 2026-01-29
feat(api-client): restore old search

amritk:
## Problem

We re-used the same modal search as references, but we want the old filtering.

## Solution

- I added back the simple old filtering, it only filters on operations (copying what we had). I have another branch that filters the whole whole tree which I can easily switch it back to if needed.
- removes the curl listener in the addressbar/codeinput
- switches back to ctrl + P for the command pallet 

## Checklist

- [x] I added a changeset. 
- [ ] I added tests.

---

> [!NOTE]
> **Medium Risk**
> Changes affect core client navigation/search and keyboard shortcut behavior; risk is mainly UX/behavior regressions (search result set, clearing behavior, hotkey conflicts) rather than data/security.
> 
> **Overview**
> Restores the API client’s *older search experience* by swapping the sidebar’s modal search button for an inline `ScalarSidebarSearchInput` backed by `useSearchIndex`, and showing either filtered results or the normal tree; selecting a result now clears the query and exits search.
> 
> `useSearchIndex` is simplified to return `null` when the query is empty (no placeholder results) and to only surface operation entries in results, matching prior client behavior; tests are updated accordingly.
> 
> Removes the `importCurl`/`curl` event pathway from `CodeInput` and the address bar placeholder text no longer suggests cURL input. Command palette opening now accepts a keyboard-event payload type (`KeyboardEventPayload`) and prevents default when triggered via hotkeys; the default hotkey mapping is adjusted and the “Add Item” button UI now displays the ⌘/Ctrl+K hint.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 13ab44ead5fef7eb991e2b1559aec6dca428ec02. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/api-client/src/v2/features/app/components/AppSidebar.vue
+++ b/packages/api-client/src/v2/features/app/components/AppSidebar.vue
@@ -7,6 +7,7 @@ import {
   useModal,
   type ScalarListboxOption,
 } from '@scalar/components'
+import { isMacOS } from '@scalar/helpers/general/is-mac-os'
 import {
   ScalarIconDotsThree,
   ScalarIconGlobe,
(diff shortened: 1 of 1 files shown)


hanspagel:
Filtering is great!

Why did we remove the curl listener?

PR #7989, opened 2026-01-30
fix(api-client): prevent API client modal from disappearing on reopen in Safari

### LINKED ISSUE #7983: API client modal isn't displayed after closing it (only on Safari)
### What happens?

1. Open https://galaxy.scalar.com/#scalar-galaxy/tag/authentication/POST/user/signup using Safari
2. Click on "Test Request"
3. Close the modal
4. Click on "Test Request" again

The overlay appears but the modal doesn't.

<details><summary>Safari info</summary>
<p>

Version 26.2 (21623.1.14.11.9)
MacOS 26.2 (25C56)

</p>
</details> 

https://github.com/user-attachments/assets/adf1316a-5b88-4662-ae8c-11b9192c8009

### Workaround

- Use Chrome / FireFox
- Every time you close the modal you need to reload the page.

API client modal can be reopened after closing it. 

### OpenAPI Document

https://galaxy.scalar.com/openapi.yaml

marcalexiei:
## Problem

- closes #7983

## Solution

Safari seems to have a lot of problem handling element changing their `display` property.
Basically it doesn't redraw the element.

This PR uses CSS properties (`opacity` + `visibility`) to control the open status of API client modal.

I attached a short video showing the new behaviour on Safari and Chrome:

https://github.com/user-attachments/assets/e3daf02c-ea83-41d1-9ba6-d07b84906f68

## Checklist

- [x] I added a changeset. 
- [x] I added tests.

### DIFF
--- a/packages/api-client/src/layouts/Modal/ApiClientModal.vue
+++ b/packages/api-client/src/layouts/Modal/ApiClientModal.vue
@@ -147,11 +147,10 @@ onBeforeUnmount(() => {
   font-family: sans-serif;
   position: absolute;
   top: 0;
-  right: 0;
+  right: 12px;
   font-size: 30px;
   font-weight: 100;
   line-height: 50px;
-  right: 12px;
   text-align: center;
   color: white;
   opacity: 0.6;
@@ -174,7 +173,6 @@ onBeforeUnmount(() => {
   bottom: 0;
   left: 0;
   top: 0;
-  left: 0;
   width: 100%;
   height: 100%;
   display: flex;
@@ -189,43 +187,5 @@ onBeforeUnmount(() => {
 
 .scalar .scalar-container {
   line-height: normal;
-}
-.scalar .scalar-app-header span {
-  color: var(--scalar-color-3);
-}
-.scalar .scalar-app-header a {
-  color: var(--scalar-color-1);
-}
-.scalar .scalar-app-header a:hover {
-  text-decoration: underline;
-}
-.scalar-activate {
-  width: fit-content;
-  margin: 0px 0.75rem 0.75rem auto;
-  line-height: 24px;
-  font-size: 0.75rem;
-  cursor: pointer;
-  font-size: 0.875rem;
-  font-weight: 600;
-  display: flex;
-  align-items: center;
-  gap: 6px;
-}
-.scalar-activate-button {
-  display: flex;
-  gap: 6px;
-  align-items: center;
-  color: var(--scalar-color-blue);
-  appearance: none;
-  outline: none;
-  border: none;
-  background: transparent;
-}
-.scalar-activate-button {
-  padding: 0 0.5rem;
-}
-.scalar-activate:hover .scalar-activate-button {
-  background: var(--scalar-background-3);
-  border-radius: 3px;
 }
 </style>


hanspagel:
what! thanks for putting this together!
waiting for @hwkr to take a look

amritk:
Yea what a legend!

packages/api-client/src/v2/features/modal/Modal.vue
```diff
@@ -255,101 +256,62 @@ defineExpose({
...
   overflow: hidden;
   border-radius: 8px;
   border: var(--scalar-border-width) solid var(--scalar-border-color);
+  z-index: 10002;
 }
-/**
- * Allow the modal to fill more space on
- * very short (or very zoomed in) screens
- */
-@variant zoomed {
```
hwkr, line 288: Did this get moved somewhere? It's important for accessibility 🤔

hwkr:
Are you sure it's the display property? I wondering if it's the animation not being triggered because the elements are already in the DOM.

hwkr:
Okay so it seems like just toggling the display property on the `div.scalar-app-layout` (which is that part that has an animation on it) is enough to to trigger the animation for Safari. I'd probably prefer to to do that as I'm a little nervous about messing with the z-index's here and possibly breaking other things.

That being said I do think our whole approach here could do for an overhaul but maybe once we have better snapshot coverage of the client features so we'll know if we broke anything?

CleanShot 2026-01-30 at 16 05 20@2x

packages/api-client/src/v2/features/modal/Modal.vue
marcalexiei: I identified several classes targeting non-existent elements; however, during the cleanup I also removed this unintentionally.

marcalexiei:
> That being said I do think our whole approach here could do for an overhaul but maybe once we have better snapshot coverage of the client features so we'll know if we broke anything?

That makes sense. I actually tried a few different things already, but at a certain point styles were borken, and I think I got a bit discouraged and didn’t try the simplest solution 😅.

> Okay so it seems like just toggling the display property on the div.scalar-app-layout (which is that part that has an animation on it) is enough to to trigger the animation for Safari. I'd probably prefer to to do that as I'm a little nervous about messing with the z-index's here and possibly breaking other things.

I’ll update this PR to include your solution.

packages/api-client/src/v2/features/modal/Modal.vue
marcalexiei: I reverted the majority of changes of the PR, which includes `@variant zoomed` removal

marcalexiei:
I applied your fix and retested on both browsers, everything works correctly!
Thanks for the review!

### Additional changes:

- Removed the duplicate `right` property in the `.scalar .scalar-app-exit:before` selector.
- Removed the duplicate `left` property in the `.scalar-container` selector.
- Removed selectors without target elements:
  - `.scalar .scalar-app-header ...`
  - `.scalar-activate`
  - `.scalar-activate-button`

I also noticed that some of these selectors are used in `packages/api-client/src/layouts/Modal/ApiClientModal.vue`.
Would you like me to open a follow-up PR to clean them up there, or should I include it in this PR?

amritk:
@marcalexiei go ahead and do it in this one

packages/api-client/src/layouts/Modal/ApiClientModal.vue
marcalexiei, line 1: I think `ApiClientModal` and `Modal` could share some logic and styling. 
Since that’s not related to the bug fixed in this PR, I’d prefer to address it in a follow-up PR.

marcalexiei:
Hello @hanspagel, 
let me know if you need any other changes before merging!

hanspagel:
let’s get this in!!

PR #8013, opened 2026-01-31
fix(agent): pass state as prop vs inject for better consumption in other apps

marclave:
## Problem

It's hard to consume the agent from agent-web with provide inject, now we can control the state and use it in other applications

## Solution

We pass the state as a prop before mount

## Checklist

- [x] I added a changeset. 
- [ ] I added tests.

---

> [!NOTE]
> **Medium Risk**
> Moderate risk because it changes component/public API and replaces injection with a global singleton state, which could break existing consumers or cause state leakage if multiple chat instances are mounted concurrently.
> 
> **Overview**
> Refactors `@scalar/agent-chat` to use an explicit `store` object passed into `Chat` instead of Vue `provide/inject`: `App.vue` now builds a store via `createState()` and passes it as `:store`, while `Chat.vue` takes the prop and wires it into the module via new `setState()`/`clearState()` calls.
> 
> Updates the public package surface to export `Chat` from `Chat.vue` (and export `State`, `createState`, `useState`), and adapts `AgentScalarChatInterface.vue` to construct the store with `createState()` and render `<Chat :store="store" />`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 6019cef5c02b9252141f327d68433d262d065c28. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/agent-chat/src/App.vue
+++ b/packages/agent-chat/src/App.vue
@@ -1,8 +1,8 @@
 <script setup lang="ts">
-import { provide, type Ref } from 'vue'
+import type { Ref } from 'vue'
 
 import Chat from '@/Chat.vue'
-import { createState, STATE_SYMBOL, type RegistryDocument } from '@/state/state'
+import { createState, type RegistryDocument } from '@/state/state'
 import { type ChatMode } from '@/types'
 
 const {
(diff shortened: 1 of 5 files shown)


amritk (approved):
Hell ya buddy. Just make sure those things are reactive. The global Singleton comment is real but we'll probably only ever have one at a time. If we need to fix that later we can

PR #8025, opened 2026-02-02
fix(api-reference): use valid CSS value for justify-content

wilcoschoneveld:
## Problem

The CSS property `justify-content: start` is not a valid value in the CSS Flexbox specification. The correct value is `flex-start`. This could cause the flexbox layout to not work as intended in the IntroductionCardItem component, potentially leading to unexpected alignment behavior.

We are getting warnings while building our app

> WARNING
  │ (1:46206) from "autoprefixer" plugin: start value has mixed support, consider using flex-start instead
  │
  │ Code:
  │   justify-content:start
  │

Also this happened before:

https://github.com/scalar/scalar/issues/3003
https://github.com/scalar/scalar/pull/3662

## Solution

Changed `justify-content: start` to `justify-content: flex-start` to use the valid CSS keyword. This ensures the flex items are aligned to the start of the flex container as intended.

## Checklist

- [ ] I added a changeset. 
- [ ] I added tests.

### DIFF
--- a/.changeset/fix-autoprefixer-warning.md
+++ b/.changeset/fix-autoprefixer-warning.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
+fix: use flex-start instead of start for better browser support


hanspagel:
cool! thanks again

PR #8033, opened 2026-02-03
fix(8009): add failsafe to incorrect array parameter style

### LINKED ISSUE #8009: Array not properly represented in Code Snippet
### What happens?

I apologize if this is already logged. I tried to search for existing Issues that matched this; but I didn't find any.

We have an API endpoint and one of the query parameters called status is of type arrayOrSingle. You can see it here: https://github.com/F3-Nation/f3-nation/blob/02ad292a4cdebb8432df9106c71fddc581507a3f/packages/api/src/router/event.ts#L34.

If you use the Scalar API Client and select both Active and Inactive as values for the query parameter, it will concatenate them int he Code Snipped with a comma and space:

Image

If you were to copy the curl address right from Code Snippet to your browser or Postman, it would fail with the following error.

```
{
    "defined": false,
    "code": "BAD_REQUEST",
    "status": 400,
    "message": "Input validation failed",
    "data": {
        "issues": [
            {
                "received": "active, inactive",
                "code": "invalid_enum_value",
                "options": [
                    "active",
                    "inactive"
                ],
                "path": [
                    "statuses",
                    0
                ],
                "message": "Invalid enum value. Expected 'active' | 'inactive', received 'active, inactive'"
            }
        ]
    }
}
```

I opened Network tab in my browser Dev tools and found that when I hit, it's sending the URL `https://proxy.scalar.com/?scalar_url=https%3A%2F%2Fapi.f3nation.com%2Fv1%2Forg%3Fstatuses%3Dactive%26statuses%3Dinactive`. If you decode that, the scalar_url parameter is `https://api.f3nation.com/v1/org?statuses=active&statuses=inactive`

So Code Snippet is showing the two array items concatenated by ", ". But it's actually sending one query parameter for each. 

I think Code Snippet should reflect what is actually being sent to proxy.scalar.com.

### OpenAPI Document

https://api.f3nation.com/docs#tag/org/POST/v1/org

amritk:
## Problem

closes #8009 

Found some issues with how we were displaying enums. 

## Solution

Fixed the original issue, also added a failsafe for incorrect styles being applied.

## Checklist

- [x] I added a changeset. 
- [ ] I added tests.

---

> [!NOTE]
> **Low Risk**
> Behavior changes are limited to parameter serialization edge cases and UI enum option population, with targeted tests added; low chance of broad regression.
> 
> **Overview**
> Prevents invalid query serialization when OpenAPI parameters incorrectly specify `style: deepObject` for non-object values by falling back to `form` style (with a new test covering arrays).
> 
> Improves request table enum handling by deriving `CodeInput` enum options from both `schema.enum` and array `schema.items` (including `$ref` resolution) and normalizing enum values to strings; adds comprehensive tests around these cases.
> 
> Adds a changeset bumping `@scalar/api-client` with a patch fix.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 7f67a48c099f839b2243dd22bb265685d2619719. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/api-client/src/v2/blocks/request-block/components/RequestTableRow.test.ts
+++ b/packages/api-client/src/v2/blocks/request-block/components/RequestTableRow.test.ts
@@ -805,6 +805,7 @@ describe('RequestTableRow', () => {
               type: 'array',
               items: {
                 $ref: '#/components/schemas/Category',
+                '$ref-value': { type: 'string' },
               },
             },
           },
@@ -881,6 +882,7 @@ describe('RequestTableRow', () => {
             name: 'mixed',
             value: 'value1',
             schema: {
+              type: 'string',
               enum: ['value1', 2, null, true],
             },
           },
--- a/packages/api-client/src/v2/blocks/request-block/components/RequestTableRow.vue
+++ b/packages/api-client/src/v2/blocks/request-block/components/RequestTableRow.vue
@@ -133,6 +133,7 @@ const enumValue = computed<string[]>(() => {
 
   return []
 })
+
 const minimumValue = computed(() =>
   data.schema && 'minimum' in data.schema ? data.schema.minimum : undefined,
 )


taterhead247:
@amritk , we're not seeing the fix. You can see our attempt at applying it on https://github.com/F3-Nation/f3-nation/issues/82. 

Could you help us understand what the root issue is so that we can fix it instead of relying on a failsafe. We looked at your comment above but aren't sure how to fix the actual issue.

amritk:
Hello, the issue was the document had style: deep object but that only applies to objects so it was skipping them entirely. I just fell back to style: form when it's not an object. 

Could you share your document and let me know what the issue is?

amritk:
@taterhead247 actually I will re-open the ticket so I can verify it!

Issue #8051, opened 2026-02-04
support custom loaders on the dereference function


DemonHa:
We need to support custom loader functions in the dereference API of @json/magic/dereference, allowing users to control how different types of references are resolved.

Reference: [Discord chat](https://discord.com/channels/1135330207960678410/1180602650476154950/1468422325626994770)

PR #8052, opened 2026-02-04
feat(json-magic): allow custom LoaderPlugin plugins in dereference

TomzBench:
## Problem

We use the [Zephyr RTOS](https://www.zephyrproject.org/) build tooling to enable/disable multiple openapi specs via the autoconf build system. We created a custom LoaderPlugin to resolve $ref's declared within the Zephyr workspace environment. For example, an embedded hardware product such as a sensor, can switch on various api's the hardware supports using CMake configuration infrastructure.

Our solution currently has to walk our entire specs for x-ext and dereference the result of our loader manually. Your tooling has a dereference api that does exactly this, and is purpose built for doing exactly what our workaround does. It may be an oversight that there is a LoaderPlugin interface, but the dereference solution does not accept custom plugins?

fixes: https://github.com/scalar/scalar/issues/8051

## Solution

So, this PR adds a Plugin[] option to the dereference API, which will override the defaults when custom loaders are used. It is backward compatible because the default loaders when no plugins are provided, fallback to the original hard coded fetchUrls plugin.

## Checklist

- [x] I added a changeset. 
- [x] I added tests.

### DIFF
--- a/.changeset/custom-plugins-dereference.md
+++ b/.changeset/custom-plugins-dereference.md
@@ -1,0 +1,5 @@
+---
+'@scalar/json-magic': minor
+---
+
+feat: allow custom LoaderPlugin plugins in dereference
--- a/packages/json-magic/src/dereference/dereference.test.ts
+++ b/packages/json-magic/src/dereference/dereference.test.ts
@@ -1,6 +1,8 @@
 import { type FastifyInstance, fastify } from 'fastify'
 import { beforeEach, describe, expect, it } from 'vitest'
 
+import type { Plugin } from '@/bundle'
+import { getHash } from '@/bundle/value-generator'
 import { dereference } from '@/dereference/dereference'
 
 describe('dereference', () => {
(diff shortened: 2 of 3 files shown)


TomzBench:
hmm, for this PR to be effective in its intent, we need a way to resolve `$ref-value`. currently the magic proxy doesn't work for serializing a complete spec.  Thoughts?

TomzBench:
@DemonHa
  
  I think dereference could be simplified further. The sync option significantly changes its behavior — when true, it skips bundle() entirely and just wraps the input in a createMagicProxy.  This code path is only exercised by a single unit test; I didn't find any call sites that use the dereference api with the sync option. It only appears used a single time by a unit test.

If we drop the sync option, and instead include a "lazy" flag, we can then dereference via proxy, which will lazily follow references as needed, -or- if lazy is not provided, then we eagerly resolve the x-ext JSON pointers and return a plain object. This latter case will satisfy my use case greatly!

I have a resolve routine that i tested in my workspace that I can update this PR with. Please let me know before going ahead with this more opinionated change. thank you!

DemonHa:
Hi @TomzBench,

Thank you for this great PR!

> If we drop the sync option, and instead include a "lazy" flag, we can then dereference via proxy, which will lazily follow references as needed, -or- if lazy is not provided, then we eagerly resolve the x-ext JSON pointers and return a plain object. This latter case will satisfy my use case greatly!

Regarding this, we can’t eagerly resolve references. To stay compliant with how sibling properties alongside a $ref must be handled, we rely on the magic proxy approach. The proxy exposes a $ref-value property, allowing references to be resolved lazily while preserving the original structure. This way, the consumer remains in control of how $refs and their sibling properties are ultimately handled, rather than us making that decision eagerly.

DemonHa:
> hmm, for this PR to be effective in its intent, we need a way to resolve `$ref-value`. currently the magic proxy doesn't work for serializing a complete spec. Thoughts?

If you need to just eagerly resolve references you should use the old dereference function from `@openapi-parser`, which does exactly that, replace any reference and make them pointer to the correct spot on the document

See #6486 for more information

PR #8069, opened 2026-02-05
fix(api-client): checkbox sizing on Safari

### LINKED ISSUE #8067: Parameter Checkbox Can't Be Clicked In Center
### What happens?

When you open the client window in the API reference component the check boxes for parameters can only be selected and deselected along the edges of the box. If you click the center nothing happens. The issue occurs in the `react` examples and the `ssg` example but not in the `web` example. I first noticed it myself using api-reference-react@0.8.50.

Clicking the center of the checkbox should toggle its state.

### OpenAPI Document

_No response_

PrinceManfred:
## Problem
fixes #8067
This fixes an issue where the checkboxes for parameters in the API Client datatable didn't take up the full amount of space on Safari. 

## Solution

I added explicit tailwind classes to the `input` to use 100% width and height%

## Checklist

- [x] I added a changeset. 
- [ ] I added tests.

### DIFF
--- a/packages/api-client/src/components/DataTable/DataTableCheckbox.vue
+++ b/packages/api-client/src/components/DataTable/DataTableCheckbox.vue
@@ -32,7 +32,7 @@ const variants = cva({
   <DataTableCell class="group/cell relative flex min-w-8">
     <input
       :checked="modelValue"
-      class="peer absolute inset-0 cursor-pointer opacity-0 disabled:cursor-default"
+      class="peer absolute inset-0 size-full cursor-pointer opacity-0 disabled:cursor-default"
       :disabled="Boolean(disabled)"
       type="checkbox"
       @change="(e: any) => $emit('update:modelValue', e.target.checked)" />


amritk:
hey thanks for the quick PR! will let @hwkr review this one

packages/api-client/src/v2/components/data-table/DataTableCheckbox.vue
```diff
@@ -32,7 +32,7 @@ const variants = cva({
   <DataTableCell class="group/cell relative flex min-w-8">
     <input
       :checked="modelValue"
-      class="peer absolute inset-0 cursor-pointer opacity-0 disabled:cursor-default"
+      class="peer absolute inset-0 h-full w-full cursor-pointer opacity-0 disabled:cursor-default"
```
hwkr: You can use `size-full` here which is a bit more concise :)
```suggestion
      class="peer absolute inset-0 size-full cursor-pointer opacity-0 disabled:cursor-default"
```

hwkr (approved):
Looks good! Left one comment :)

PrinceManfred:
Ah thanks! I missed that class when I looked :). I don't ever use tailwind.

PrinceManfred:
One more commit to add the fix to the v1 checkbox. I think that should actually be it.

PR #8072, opened 2026-02-06
refactor(api-client): unify logic and styles across `Modal` and `ApiClientModal` in `ModalContainer`

marcalexiei:
## Problem

Followup of https://github.com/scalar/scalar/pull/7989#discussion_r2756757238

> I think ApiClientModal and Modal could share some logic and styling.
> Since that’s not related to the bug fixed in this PR, I’d prefer to address it in a follow-up PR.

## Solution

Create a new `ModalClientContainer.vue` component that contains the shared logic between
- `src/layouts/Modal/ApiClientModal.vue`
- `src/v2/features/modal/Modal.vue`

## Checklist

- [x] I added a changeset. 
- [x] I added tests.

### DIFF
--- a/packages/api-client/src/v2/components/modals/ModalClientContainer.spec.ts
+++ b/packages/api-client/src/v2/components/modals/ModalClientContainer.spec.ts
@@ -70,10 +70,8 @@ describe('ModalClient.vue', () => {
 
     expect(wrapper.find('.scalar').isVisible()).toBe(true)
 
-    onTestFinished(async () => {
-      await wrapper.unmount()
-      // Clean the body after the test
-      document.body.innerHTML = ''
+    onTestFinished(() => {
+      wrapper.unmount()
     })
   })
 
@@ -85,7 +83,6 @@ describe('ModalClient.vue', () => {
       slots: { default: '<button id="focus-target">Target</button>' },
       attachTo: document.body,
     })
-    console.log('document.outerHTML', document.body.outerHTML)
 
     modalState = { ...modalState, open: true }
     await wrapper.setProps({ modalState })
@@ -112,9 +109,8 @@ describe('ModalClient.vue', () => {
      */
     expect([focusTarget, fallbackFocus]).toContain(document.activeElement)
 
-    onTestFinished(async () => {
-      await wrapper.unmount()
-      document.body.innerHTML = ''
+    onTestFinished(() => {
+      wrapper.unmount()
     })
   })
 


PrinceManfred:
It looks like we might be working on the same files. I opened #8075 this morning before I saw this new PR. I managed to tweak #7989 to fix the initial transition in on Safari (it doesn't fade in on first open) and eliminate the double `v-show`. How do we feel about moving that new transition code into this wrapper?

packages/api-client/src/v2/components/modals/ModalClientContainer.vue
marcalexiei, line 1: Hello @PrinceManfred,
thanks for continuing the work of #7989!

> How do we feel about moving that new transition code into this wrapper?

I’d suggest the following approach:

1. Mark your PR as draft.
2. Wait for this one to be merged.
3. Re-apply your changes via rebase.

This way, we avoid reviewing both changes at the same time and you only have to apply them once.
What do you think?

hanspagel:
we need @hwkr here

packages/api-client/src/v2/components/modals/ModalClientContainer.spec.ts
```diff
@@ -0,0 +1,136 @@
...
+  deactivateMock,
+  addScalarClassesToHeadlessMock,
+} = vi.hoisted(() => ({
+  activateMock: vi.fn(),
+  deactivateMock: vi.fn(),
+  addScalarClassesToHeadlessMock: vi.fn(),
+}))
+
+vi.mock('@vueuse/integrations/useFocusTrap', () => ({
+  useFocusTrap: () => ({
```
hwkr: It would be nice to not mock `useFocusTrap` and actually test the focus trap. I believe in jsdom you can test focus by attaching the mount to the dom and checking `document.activeElement`.

packages/api-client/src/v2/components/modals/ModalClientContainer.spec.ts
```diff
@@ -0,0 +1,136 @@
...
+    activate: activateMock,
+    deactivate: deactivateMock,
+  }),
+}))
+
+vi.mock('@scalar/components', () => ({
+  addScalarClassesToHeadless: addScalarClassesToHeadlessMock,
+  ScalarTeleportRoot: {
```
hwkr: Would also be nice not to mock this and to validate that the teleport is playing nice.

marcalexiei: Done via https://github.com/scalar/scalar/pull/8072/changes/954f409f5705c3f035557918ddb79158ebc42fb3

packages/api-client/src/v2/components/modals/ModalClientContainer.spec.ts
marcalexiei: Focus management doesn’t work properly in JSDom,
`document.activeElement` always stays on `body`, 
even though the component works correctly in a real browser. 

Since JSDom cannot fully emulate browser behaviour, 
IMHO it seems reasonable to keep the mock in place for testing the focus trap.

hwkr: Does it? We have a some tests in the component library that do focus testing → https://github.com/scalar/scalar/blob/main/packages/components/src/components/ScalarCombobox/ScalarCombobox.test.ts#L80-L99

Here's a stackoverflow that uses the same attachTo technique https://stackoverflow.com/questions/53041818/how-can-you-test-if-an-input-is-focused-with-vue-test-utils

marcalexiei: I'll give it another try

marcalexiei: I've update the test using a similar approach: https://github.com/scalar/scalar/pull/8072/commits/2962bb3a3c5e10bf00ed2b73b54adbe0eea1d2da

it fails anyway 😭 : https://github.com/scalar/scalar/actions/runs/21855695753/job/63072122317?pr=8072

Sorry, but I'm not very familiar with vue so probably I'm missing something

hwkr: All good! Thanks for trying, I'll take a swing at it today and if I can't get it working we can just merge without 😊

marcalexiei: > Thanks for trying

I just finished debugging the code inside `useFocusTrap` so I end up inside `focus-trap` library code.

when inside the `_tryFocus` function, which invokes element's `focus()`, the `document.activeElement` is set correctly: 

Screenshot 2026-02-10 at 18 33 49

However outside the function something change it back to `document.body`.

I think that the cause might be `focus-trap` using some browser API that are not well supported by JSDOM.

---

> I'll take a swing at it today and if I can't get it working we can just merge without 😊

Thanks!
I hope you are more lucky than me!

hwkr (approved):
@marcalexiei okay so there were several things going on 😓 I was able to get the tests working (mostly)

* For the unit tests there was an issue with some of the tests using `attachTo` body but not unmounting at the end of their tests which led to multiple copies of the modal existing and the queries grabbing an old version of the modal.
* I couldn't get it to actually target the button but I was able to test that `useFocusTrap` at least uses the fallback focus, hopefully the unit test will keep us from breaking that part at least for the foreseeable future.
* I lieu of the unit test I added an end to end test in the references to make sure that focus behaves as expected when opening the modal (as an aside, playwright is _so_ much better at testing stuff like this).

Thanks for waiting a bit for me to play around with it! Let me know if you see anything I changed that I shouldn't have :)

packages/api-client/src/v2/components/modals/ModalClientContainer.spec.ts
hwkr: Pushed some updates! Should be good to go now :)

packages/api-client/src/v2/components/modals/ModalClientContainer.spec.ts
```diff
@@ -69,48 +69,56 @@ describe('ModalClient.vue', () => {
...
-
     let modalState = createModalState(false)
 
     const wrapper = mount(ModalClientContainer, {
       props: { modalState },
-      slots: { default: '<button id="test-button">Toggle</button>' },
+      slots: { default: '<button id="focus-target">Target</button>' },
       attachTo: document.body,
     })
+    console.log('document.outerHTML', document.body.outerHTML)
```
marcalexiei: Debug log that can be removed

packages/api-client/src/v2/components/modals/ModalClientContainer.spec.ts
```diff
@@ -69,48 +69,56 @@ describe('ModalClient.vue', () => {
...
-    expect(document.activeElement).toBe(button)
     expect(wrapper.emitted('open')).toBeTruthy()
 
-    onTestFinished(() => {
-      wrapper.unmount()
-      vi.useRealTimers()
+    const focusTarget = wrapper.find('#focus-target').element
+    const fallbackFocus = wrapper.find('[role="dialog"]').element
+
+    expect([focusTarget, fallbackFocus]).toContain(document.activeElement)
```
marcalexiei, line 110: I would consider to add a comment [referring your earlier findings](https://github.com/scalar/scalar/pull/8072#pullrequestreview-3783424790) so it is straight forward why this test is doing a test with `toContain`

packages/api-client/src/v2/components/modals/ModalClientContainer.spec.ts
```diff
@@ -69,48 +69,56 @@ describe('ModalClient.vue', () => {
     await nextTick()
 
     expect(wrapper.find('.scalar').isVisible()).toBe(true)
+
+    onTestFinished(async () => {
+      await wrapper.unmount()
+      // Clean the body after the test
+      document.body.innerHTML = ''
+    })
```
marcalexiei: Not directly related to this PR, but I think it’s worth discussing.

Currently, each test calls `wrapper.unmount()`. 
In most cases, this should ideally happen after every test.

What do you think about using [`enableAutoUnmount`](https://test-utils.vuejs.org/api/#enableAutoUnmount) so we don’t have to call `wrapper.unmount()` explicitly in each test? 
It might also help prevent a failing test from affecting the next one.

I can explore implementing this in a separate PR if you’d like.

packages/api-client/src/v2/components/modals/ModalClientContainer.spec.ts
```diff
@@ -69,48 +69,56 @@ describe('ModalClient.vue', () => {
     await nextTick()
 
     expect(wrapper.find('.scalar').isVisible()).toBe(true)
+
+    onTestFinished(async () => {
+      await wrapper.unmount()
+      // Clean the body after the test
+      document.body.innerHTML = ''
```
marcalexiei: I tried removing this line here, as well as in "activates focus trap and emits open", and the tests still pass. 
Maybe it’s no longer necessary?

marcalexiei:
> I lieu of the unit test I added an end to end test in the references to make sure that focus behaves as expected when opening the modal (as an aside, playwright is so much better at testing stuff like this).

I concur that the best solution is to rely on `playwright`. 
`jsdom` is just an emulator, whereas `playwright` uses a real browser, so this kind of weird behaviour shouldn’t occur there.

---

> Thanks for waiting a bit for me to play around with it!

No problem! In the end, you’re the one who fixed the test by removing the mock 😅

---

> Let me know if you see anything I changed that I shouldn't have :)

I left a few comments. I can apply the changes myself later today (except for the part about `enableAutoUnmount`, which needs your feedback 😁).

packages/api-client/src/v2/components/modals/ModalClientContainer.spec.ts
hanspagel: yes please! (and separate PR sounds good)

hanspagel:
> I can apply the changes myself later today

🤟 that would be rad, I’ll merge it once done

marcalexiei:
@hanspagel I applied all my suggestions and CI is green. 
Let me know if you need additional adjustments before merging!

packages/api-client/src/v2/components/modals/ModalClientContainer.spec.ts
hwkr: Oh that's cool, yeah we should do that

packages/api-client/src/v2/components/modals/ModalClientContainer.spec.ts
hwkr: Oops ty

PR #8075, opened 2026-02-06
fix(api-client): consistent modal transitions across browsers

PrinceManfred:
## Problem

Safari doesn't animate the modal in smoothly on first render.

## Solution

I refactored the styles a little and switched to using Vue `<Transition>` to manage the fade. This fix is related to #7989 that just came in. I think we were both trying to fix the Safari modal not showing bug. Using `<Transition>` allowed me to eliminate the redundant `v-show`. As an added bonus you also get a smooth fade out on close now. There was discussion of centralizing style in #7989. I kept them scoped since that seems to be the conclusion there as well. I changed where the `z-index` is applied, but overall the z-index is the same as before.

## Checklist

- [x] I added a changeset. 
- [ ] I added tests.

### DIFF
--- a/packages/api-client/src/v2/components/modals/ModalClientContainer.vue
+++ b/packages/api-client/src/v2/components/modals/ModalClientContainer.vue
@@ -55,35 +55,32 @@ onBeforeUnmount(() => {
 </script>
 
 <template>
-  <div
-    v-show="modalState.open"
-    class="scalar scalar-app">
-    <div class="scalar-container">
-      <!-- 
-      adding v-show also here to ensure proper rendering in Safari.
-      @see https://github.com/scalar/scalar/issues/7983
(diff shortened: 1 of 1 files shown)


PrinceManfred:
Github's diff viewer makes it look like this is a huge change. Most of it is just indention level.

hanspagel:
thanks! @hwkr can u take a look 👀

hanspagel:
draft until #8072 is in

marcalexiei:
Hey @PrinceManfred,
#8072 has been merged!

Thanks again for your patience!

PrinceManfred:
Good work over there 😊. I’ll port my changes now.

hanspagel:
awesome! will wait for @hwkr’s review

PR #8080, opened 2026-02-07
feat: Make configurable which client libraries are hidden under 'more…

### LINKED ISSUE #6665: Make configurable which client libraries are hidden under 'more' button
We need a way to configure the "featured" HTTP clients!

### Discussed in https://github.com/scalar/scalar/discussions/6664

<sup>Originally posted by **Davidsv** August 26, 2025</sup>
No matter how I configure the EnabledClients/EnabledTargets arrays, certain client libraries are always hidden under the 'more' button. For example, if I set options.EnabledClients = [ScalarClient.NetHttp] it will look like this:

image

Please add some configuration for which buttons appear by default in this area:

image
</div>

matheusandre1:
## Problem
  Closes #6665  

## Solution

- [x] I added tests.


hanspagel:
thanks for attempting this! we need to refactor how you configure those clients before we add this feature, that’s up on us, so I have to close this here for now

also it’s missing documentation and more tests, but don’t bother to add those, we’ll take care of it 🤞

PR #8094, opened 2026-02-09
fix(api-reference): load selected example when opening client modal

### LINKED ISSUE #8093: API client doesn't open with selected example
### What happens?

Assume multiple request body examples are defined for an endpoint:

- Example 1
- Example 2
- Example 3

When I select Example 2 and click on **Test Request**, I would expect the API client to open with Example 2. But instead, it always opens with Example 1.

You can reproduce this with the standard [HTML/JS example](https://scalar.com/products/api-references/integrations/html-js).

1. Go to the **Create a user** endpoint.
2. Switch the language to **Curl**.
3. Select the **Hans** example.
4. Click on **Test Request**.

→ The API client opens with the "Marc" example.

Image

Image

The API client should open with the selected example. In the example above, it should show the "Hans" example.

### OpenAPI Document

_No response_

wilhelmer:
Fixes #8093.

### DIFF
--- a/packages/api-client/src/v2/blocks/operation-code-sample/components/OperationCodeSample.test.ts
+++ b/packages/api-client/src/v2/blocks/operation-code-sample/components/OperationCodeSample.test.ts
@@ -358,7 +358,8 @@ describe('RequestExample', () => {
 
       const emitted = wrapper.emitted('update:selectedExample')
       if (emitted) {
-        expect(emitted[0]).toEqual(['example2'])
+        // Check the last emission (after the initial mount emission)
+        expect(emitted[emitted.length - 1]).toEqual(['example2'])
       } else {
         expect(true).toBe(true)
       }


packages/api-client/src/v2/blocks/operation-code-sample/components/OperationCodeSample.vue
```diff
@@ -160,7 +161,7 @@ const {
   isWebhook,
   generateLabel,
   globalCookies,
-} = defineProps<OperationCodeSampleProps>()
+} = props
```
hanspagel: naah, we don’t want that

hanspagel:
thanks! @amritk can u take a look here?

packages/api-client/src/v2/blocks/operation-code-sample/components/OperationCodeSample.vue
wilhelmer: Yeah, I see it's not necessary, let me change that

amritk:
hey thanks for the PR! Let me see if we can just send the example ID directly then we don't need to look it up after, if not this works great

packages/api-reference/src/features/test-request-button/TestRequestButton.vue
```diff
@@ -5,15 +5,23 @@ import type { WorkspaceEventBus } from '@scalar/workspace-store/events'
...
 
-/** Route via ID */
-const handleClick = () => eventBus.emit('ui:open:client-modal', { id })
+/** Route via ID and optionally with example name */
+const handleClick = () => {
+  // Include exampleName in payload if provided
+  const payload: { id: string; exampleName?: string } = { id }
+  if (exampleName) {
+    payload.exampleName = exampleName
+  }
```
amritk:
```suggestion
  const payload = { 
    id,
    ...(exampleName && { exampleName }) 
  }
```
we usually do this syntax

packages/api-client/src/v2/features/modal/modal-events.ts
```diff
@@ -45,7 +45,24 @@ export function initializeModalEvents({
...
+      let targetId = payload.id
+
+      // If exampleName is provided, try to find the specific example entry
+      if ('exampleName' in payload && payload.exampleName) {
+        const operationEntry = sidebarState.state.getEntryById(payload.id)
+
+        // Try to find the specific example in the operation's children
+        if (operationEntry && 'children' in operationEntry && operationEntry.children) {
+          const exampleEntry = operationEntry.children.find(
+            (child: any) => child.type === 'example' && child.name === payload.exampleName,
```
amritk:
```suggestion
            (child: TraversedEntry) => child.type === 'example' && child.name === payload.exampleName,
```

wilhelmer:
Thanks @amritk, applied the nitpicks, everything still looks good in dev 👍

amritk:
@wilhelmer I made some updates, could you merge my PR?

wilhelmer:
@amritk ~~Which PR~~ 🤔 Oh it's on my repo, nevermind, I'll check

hanspagel:
fast turnaround 👏 
let’s get this in

hanspagel:
oh no, one test is failing

wilhelmer:
Some (hopefully) wise words from Claude 🤖:

>I can see the issue. When the component mounts, the `onBeforeMount` hook sets the default example to the first one ('example1'), which emits update:selectedExample. 
Then when you change the example, it emits again. The test is checking the first emission (emitted[0]) but should check the second emission (emitted[1]) since there are now two emissions.
The fix was to check the last emission (emitted[emitted.length - 1]) instead of the first one (emitted[0]), which correctly captures the value after the user changes the example.

[Commit](https://github.com/scalar/scalar/pull/8094/changes/bf111bc2c9707317469e48947db44322ac4122c7) added.

amritk:
Teamwork!

PR #8107, opened 2026-02-10
feat(api-reference): proper schema titles

### LINKED ISSUE #6693: don't use the title to replace the type of primitives

hanspagel:
## Problem

we overwrite schema types with titles, that doesn’t make any sense at all

* see https://github.com/scalar/scalar/pull/6953#issuecomment-3466913742

## Solution

🚧 WIP

* always show the type
* show the title, if available
* if it’s $ref and we don't have a title, show the schema key

fixes #6693

* changes the TS-like notation from `array string[]` to `string[]` (they were more absurd examples, where it’s nested)

## Before

No type:

Screenshot 2026-02-10 at 18 25 02

## After

Type + title:

Screenshot 2026-02-10 at 18 24 09

## Checklist

- [ ] I added a changeset. 
- [x] I added tests.

### DIFF
--- a/packages/api-reference/src/components/Content/Schema/SchemaComposition.test.ts
+++ b/packages/api-reference/src/components/Content/Schema/SchemaComposition.test.ts
@@ -71,7 +71,7 @@ describe('SchemaComposition', () => {
       })
 
       const tab = wrapper.find('.composition-selector-label')
-      expect(tab.text()).toBe('array string[]')
+      expect(tab.text()).toBe('string[]')
     })
   })
 


hanspagel:
the scope of this PR became a bit fuzzy, I reimplemented a scoped version here:

#8129

PR #8110, opened 2026-02-11
feat(api-client): support file plugin on the import document palette

DemonHa:
The command palette import currently relies on the browser’s default file picker. However, in Electron we use custom logic for file selection.

This PR refactors the import component to make it more extensible, allowing consumers (such as the Electron app) to plug in a custom file picker while preserving the existing default behavior.

## Checklist

- [x] I added a changeset. 
- [x] I added tests.

---

> [!NOTE]
> **Medium Risk**
> Touches the core document-import path (URL/raw/Postman) and changes the `loadDocumentFromSource` API, which could break existing call sites or alter import behavior, though coverage is expanded with new tests.
> 
> **Overview**
> Enables non-browser environments (e.g. Electron) to import API specs via a **file loader plugin** rather than relying on the browser file picker.
> 
> `CommandPaletteImport` now accepts an optional `fileLoader` prop and exposes a `fileUpload` slot so consumers can trigger imports themselves (including `type: 'file'`). The import pipeline is refactored so `loadDocumentFromSource` takes `{ source, type }`, supports `type: 'file'` by adding documents via `path`, and switches raw parsing/conversion to a loader-based approach (Postman/JSON/YAML); tests are updated and extended to cover file-path imports and watch-mode behavior.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit ea1ae7c07988edc4cf19438a9fc6f3cca5c62122. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/api-client/src/v2/features/command-palette/components/CommandPaletteImport.vue
+++ b/packages/api-client/src/v2/features/command-palette/components/CommandPaletteImport.vue
@@ -241,7 +241,8 @@ const { open: openSpecFileDialog } = useFileDialog({
 
     const onLoad = async (event: ProgressEvent<FileReader>): Promise<void> => {
       const text = event.target?.result as string
-      await handleImport(text)
+      // Import file as raw content since we are in a browser environment
+      await handleImport(text, 'raw')
     }
 
     const reader = new FileReader()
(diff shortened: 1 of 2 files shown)


amritk (approved):
LGTM did not test

PR #8115, opened 2026-02-11
feat(asyncapi-types): new package

hanspagel:
## Problem

We have a package for OpenAPI types, but not for AsyncAPI types. For the upcoming AsyncAPI support, we want to have a dedicated package that has really strict and clean TypeScript types and TypeBox schemas.

## Solution

This PR adds `@scalar/asyncapi-types`:

🚧 WIP 🚧

- [x] AsyncAPI 3.0 types
- [x] AsyncAPI 3.0 TypeBox schemas
- [ ] tests the types and the schemas match
- [x] @scalar’s TypeBox fork is an optional peer dependency, users only have to install it if they want to use the TypeBox schemas

## Checklist

- [x] I added a changeset. 
- [x] I added tests.

### DIFF
--- a/.changeset/major-terms-say.md
+++ b/.changeset/major-terms-say.md
@@ -1,0 +1,5 @@
+---
+'@scalar/asyncapi-types': minor
+---
+
+init :)


hanspagel:
looks like we'll switch to json schema for validation 🆗

PR #8123, opened 2026-02-12
test(vue): use `enableAutoUnmount` to avoid manual unmount

marcalexiei:
## Problem

- Followup of https://github.com/scalar/scalar/pull/8072#discussion_r2792198916

> Currently, each test calls wrapper.unmount().
> In most cases, this should ideally happen after every test.
>
> What do you think about using [enableAutoUnmount](https://test-utils.vuejs.org/api/#enableAutoUnmount) so we don’t have to call wrapper.unmount() explicitly in each test?
> It might also help prevent a failing test from affecting the next one.

## Solution

Use `enableAutoUnmount` on vue tests file

## Checklist

- [ ] I added a changeset. 
- [x] I added tests.


marcalexiei:
Thanks for the clean up!

Issue #7423, opened 2025-11-26
Request/response schemas show as “object” instead of referenced DTO schema name


miladmazaheri:
## Description  
When using an OpenAPI spec with named DTO schemas (e.g. `LoginInputDto`, `UserProfileDto`) referenced via `$ref` under `components/schemas`, Scalar’s UI displays request-body and response schemas as generic `object` (or `object[]`) instead of using the referenced DTO name.

This happens even though the OpenAPI JSON correctly defines the schema reference; other tools (e.g. Swagger UI) display the correct DTO name, but in Scalar UI the name is lost and everything is flattened into plain “object”.

## Actual Behavior  
- Request-body schemas are shown as `object`, even though the OpenAPI spec uses `$ref`.  
- Response schemas (including arrays) are shown as `object` or `object[]`, not the referenced DTO types.  
- This makes it difficult to understand which models are expected by endpoints, and defeats the purpose of having named schemas in the documentation.

## Steps to Reproduce  
Generate an OpenAPI spec with a named schema:  
   ```json
   "components": {
     "schemas": {
       "LoginInputDto": {
         "type": "object",
         "properties": { ... }
       }
     }
   }
```
```json
"paths": {
  "/api/users/login/by-username": {
    "post": {
      "requestBody": {
        "content": { "application/json": { "schema": { "$ref": "#/components/schemas/LoginInputDto" } } },
        "required": true
      },
      "responses": {
        "200": {
          "content": { "application/json": { "schema": { "$ref": "#/components/schemas/LoginDto" } } }
        }
      }
    }
  }
}
```

<img width="776" height="444" alt="Image" src="https://github.com/user-attachments/assets/d6f5898d-58f6-4b4e-8d42-053a0008c168" />

## Expected Behavior  
- For request bodies and responses that reference named schemas (via `$ref`), Scalar UI should display the actual schema name — e.g. `LoginInputDto`, `UserProfileDto`, etc., instead of `object`.  
- For response arrays, the UI should show `DTOName[]` (or similar), not `object[]`.

morrica:
Note that this issue affects properties in the "Models" section as well. 

```
...
"components": {
    "schemas": {
      "ApiResponse": {
        "type": "object",
        "properties": {
          "isSuccess": {
            "type": "boolean"
          },
          "value": {
            "type": "object",
            "description": "The successful result of the API call."
          },
          "errors": {
            "$ref": "#/components/schemas/Errors"
          }
        }
      },
      "Errors": {
[...]

PR #8129, opened 2026-02-12
feat(api-reference): always show the schema type

### LINKED ISSUE #6693: don't use the title to replace the type of primitives

hanspagel:
## Problem

we overwrite schema types with titles, that doesn’t make any sense at all

> Titles are meant for display purposes but they are not the same as types (even in the OpenAPI spec). There are three pieces of data:
>
> * field name (or query parameter name)
> * type
> * title
>
> The former two are required to use an API correctly. The last one is a nice to have. It can't be used to replace either of the above two (especially not in a scalar case, such as for query parameters).

source: https://github.com/scalar/scalar/pull/6953#issuecomment-3466913742

## Solution

* always show the type
* show the title, if available

fixes #6693

## Checklist

- [x] I added a changeset. 
- [x] I added tests.

---

> [!NOTE]
> **Medium Risk**
> Touches shared schema display helpers and headings, so type/name output changes could affect multiple API reference views and snapshots, but it’s isolated to UI formatting logic.
> 
> **Overview**
> Schema type rendering has been changed to **always show the structural OpenAPI type** (e.g. `object`, `array object[]`) rather than substituting schema `title`, `xml.name`, or `$ref` names.
> 
> Schema titles/names are now shown *in addition* to the type in `SchemaPropertyHeading` (appended as `· <title>` and supporting array item titles like `Model[]`), and composition selector labels in `SchemaComposition` prefer schema title/ref name over structural type. Tests were updated accordingly, and the legacy `getModelName` helper was removed in favor of `getModelNameFromSchema`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit a6c7ade97c1d3d21d519aee6c913477becc42159. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


marclave (approved):

silkfire:
Does this also fix https://github.com/scalar/scalar/issues/7423?

So we actually see the real type (model) name instead of just `object`?

hanspagel:
@silkfire I think this will need an additional PR

silkfire:
Okay, hope that can be done soon

PR #8133, opened 2026-02-12
fix(workspace-store): remove special-casing for content parameters

bobheadxi:
## Problem

Currently, when using content parameters like:

```yaml
      parameters:
        - name: message
          in: query
          required: true
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/users.v1.GetUserRequest'
```

the "Test request" modal will ignore any value you populate for this parameter:

CleanShot 2026-02-12 at 23 01 48@2x

It's a bit odd, but this query parameter setup is the standard in Connect Protocol: https://connectrpc.com/docs/protocol/#unary-get-request

## Solution

We currently have a similar patch applied removing the early return on schema parameters, which fixes the "Test request" modal. It seems that content parameters are currently being skipped over for examples, but it looks like content parameters already have special handling here: https://sourcegraph.com/github.com/scalar/scalar/-/blob/packages/api-client/src/v2/blocks/operation-block/helpers/get-example.ts?L47-54 so this special case is not needed.

## Checklist

- [x] I added a changeset. 
- [x] I added tests.

### DIFF
--- a/packages/workspace-store/src/mutators/operation/parameters.test.ts
+++ b/packages/workspace-store/src/mutators/operation/parameters.test.ts
@@ -251,9 +251,6 @@ describe('upsertOperationParameter', () => {
     assert('examples' in param && param.examples)
     expect(getResolvedRef(param.examples.default)?.value).toBe('{"id": 1}')
     expect(getResolvedRef(param.examples.default)?.['x-disabled']).toBe(false)
-    // Content property should remain untouched
-    assert('content' in param && param.content)
-    expect(param.content['application/json']).toBeDefined()
   })
 
   it('adds multiple parameters of different types', () => {


marclave (approved):

hanspagel:
AH, good catch! @amritk @DemonHa can you have a look?

hanspagel (changes requested):
yeah, we need 'example' (singular) too

amritk:
hey I don't think we should make those parameter changes here. Those changes were made in 3.2 and these are 3.1 schemas. 

cc: @hanspagel 

Removing the TODO should be good though if it works!

bobheadxi:
@amritk do you mean the example and examples additions?

amritk:
@bobheadxi yes! I believe those were added in 3.2 and we should probably keep the 3.1 schemas to 3.1. Eventually we should just upgrade our entire schema to 3.2 but we gotta do our upgrader first

bobheadxi:
@amritk I pushed a commit to remove example/examples. To have the type check work in `upsertOperationParameter`, however, I had to mess around with assuming `examples` is available, but functionally it should be the same as the patch we are already using to get the API playground working

hanspagel:
@amritk what? I’m confused.

> Parameter Objects MUST include either a content field or a schema field, but not both. 

Fields for `schema`:

Screenshot 2026-02-16 at 15 21 04

Source: 

OpenAPI 3.1.2 
https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.2.md#fixed-fields-for-use-with-schema

amritk:
> @amritk what? I’m confused.
> 
> > Parameter Objects MUST include either a content field or a schema field, but not both. 
> 
> Fields for `schema`:
> 
> Screenshot 2026-02-16 at 15 21 04
> 
> Source: 
> 
> OpenAPI 3.1.2 
> https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.2.md#fixed-fields-for-use-with-schema

Yes in that link example exists only when it's a schema parameter

bobheadxi:
Thank you everyone!

I think the PR is currently blocked on a changes-requested from @hanspagel that I can't mark as outdated - would appreciate if we can get this landed in a release soon 🙏

Issue #8136, opened 2026-02-13
Rust API framework integrations missing references to scalar.js


foxfriends:
### What happens?

When using the Rust `scalar_api_references` crate, the framework integrations attempt to serve the file `scalar.js`, but because this file is not actually present in the package, it does not work.

~~There seems to be a `package.json` file there, with a script `copy:standalone`, but it seems that it is not run at any point in the Rust release.~~ 

EDITED: I see there is a step to download this artifact during the release, so it is unexpected that the file is missing... maybe this step was added more recently than the publish. Fixing the release and actually publishing might be all that is needed.

### What did you expect to happen?

The JS file should be included in the package and served, so that the thing works.

foxfriends:
Hi, thanks for publishing this again, but it would seem it did not work, the script is still missing from the released package:

https://docs.rs/crate/scalar_api_reference/0.1.5/source/ui/

Something more must be wrong with the publish workflow that it does not pick up these files.

PR #8137, opened 2026-02-13
ci: fix version patching script in rust-publish job

foxfriends:
## Problem

`rust-publish` CI is broken because the step to update the version is updating also the dependency versions incorrectly.

## Solution

Instead of updating ALL occurrences of `version = "..."`, update only the first occurrence.

## Checklist

- [ ] I added a changeset. 
- [ ] I added tests.


foxfriends:
I think the issue #8136 is already fixed, but it requires releasing the package successfully, so this might do it.

PR #8149, opened 2026-02-13
fix(api-client): show parameter examples

### LINKED ISSUE #6639: Example not shown as placeholder for request parameter
### Discussed in https://github.com/scalar/scalar/discussions/6588

<sup>Originally posted by **mariiaStorozhyk** August 18, 2025</sup>
Hi 👋

I’m using Scalar v1.28.14 and noticed that query parameter example is not displayed as a placeholder in the request UI.

For example:

```
{
  "name": "dateFrom",
  "in": "query",
  "description": "Start date",
  "required": true,
  "schema": { "type": "string", "format": "date" },
  "example": "2025-08-01"
}
```

What I expect:
The request input field should show 2025-08-01 as a placeholder (like Swagger UI does when example is provided).

What actually happens:
Scalar shows just Value as a placeholder, ignoring the example.
image

also i do not see this example in curl

curl '${server link}/open/api/acc/{acc-id}/transactions?bookingStatus=both&dateFrom=&dateTo=' \

Question:

Is this intended behavior?

If yes, is there any way to configure Scalar so that example is rendered as placeholder instead of "Value"?

Thanks! 🙏</div>

wilhelmer:
Fixes #6639.

### DIFF
--- a/.changeset/few-bags-shine.md
+++ b/.changeset/few-bags-shine.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
+
+fix: prefill parameter inputs from parameter-level examples


amritk:
nice this works, could you please add a test and changeset?

wilhelmer:
@amritk Done, please check

PR #8156, opened 2026-02-13
fix(api-reference): do not trigger send request unless modal is open

### LINKED ISSUE #8090: Scalar fetches endpoint on every Enter keypress
### What happens?

Whenever you press `Enter` on Scalar's main window, Scalar performs a request on the first endpoint listed in the `paths` section of the spec file.

This can be reproduced in any browser and any spec file, for example, the very basic [HTML/JS sample file](https://scalar.com/products/api-references/integrations/html-js).

Image

This is very undesirable, especially if you have other input fields or forms on the page. Each `Enter` keypress will then silently call an API endpoint.

No API calls should be performed on `Enter`, instead, the default browser behavior should apply.

amritk:
## Problem

closes #8090

We were sending request any time you hit enter on the references.

## Solution

- we ensure its meta + enter
- we don't trigger it when the modal is closed

## Checklist

- [x] I added a changeset. 
- [x] I added tests.

---

> [!NOTE]
> **Low Risk**
> Behavior change is limited to UI hotkey routing and request triggering, with added tests and no changes to request construction or transport.
> 
> **Overview**
> Prevents global hotkeys from unintentionally sending API requests when the API client modal is not open.
> 
> Hotkey handling is tightened so configured hotkeys only fire when their required modifiers are pressed (e.g., *default modifier + Enter*), and `OperationBlock` now guards `operation:send:request:hotkey` execution in `modal` layout using a new `isModalOpen` prop that is wired from `Modal` through `Operation`. Tests were added/updated to cover the new hotkey and modal-open behavior, and a patch changeset was included for `@scalar/api-client`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 6e88fbbc8c8735fe7ba8cfdd3143a6a3de2f0a88. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/.changeset/wise-cloths-drop.md
+++ b/.changeset/wise-cloths-drop.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
+
+fix: only trigger send request when client is open


packages/api-client/src/v2/blocks/operation-block/OperationBlock.vue
```diff
@@ -255,12 +259,21 @@ const handleExecute = async () => {
   request.value = sendResult.request
 }
 
+/** Guard hotkey execution so closed modal instances do not react to global hotkeys. */
```
DemonHa, line 262: we should probably do the same for any modal that gets open

PR #8167, opened 2026-02-14
fix(api-client): derive Accept header default from response content types

thijskok:
## Problem

The `Accept` header was hardcoded to `*/*` for all requests, regardless of what the OpenAPI spec defined as valid response content types. Some APIs treat `*/*` as an unexpected or invalid value and respond with errors or unexpected behavior (for example returning an unsupported format or rejecting the request entirely).

## Solution

- Derive the `Accept` header from all content types defined on the first 2xx response in the OpenAPI operation
- When multiple content types are defined, join them as a comma-separated list (e.g. `application/json, application/xml`) so the server can pick the best match via content negotiation
- Fall back to `*/*` when no 2xx response is defined, or when the response has no `content` (e.g. 204 No Content)

## Checklist

- [x] I added a changeset.
- [x] I added tests.

### DIFF
--- a/.changeset/few-beers-kick.md
+++ b/.changeset/few-beers-kick.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
+
+Accept header now defaults to the response content types defined in the OpenAPI spec instead of always using _/_


amritk:
Yea seems like something we could do, thoughts @marclave @hanspagel?

hanspagel:
love it

PR #8174, opened 2026-02-15
feat(api-client): new option: `createAnySecurityScheme`

jensenbox:
## Summary

Adds a new `authentication.createAnySecurityScheme` configuration option that controls whether generic "Add new authentication" options (API Key in Cookies, API Key in Headers, HTTP Basic, HTTP Bearer, OAuth2 flows, etc.) are shown in the auth dropdown.

When `createAnySecurityScheme: true`, these generic options are shown alongside the schemes defined in the OpenAPI spec. By default (`false`), only schemes from the document are displayed.

Additionally, when `createAnySecurityScheme` is `false` and there are no required schemes, the dropdown returns a flat list of available schemes instead of empty groups.

### Usage

```js
{
  authentication: {
    createAnySecurityScheme: true,
    preferredSecurityScheme: 'bearerAuth',
  }
}
```

### Changes

- **`authentication-configuration.ts`** — Add `createAnySecurityScheme?: boolean` to `AuthenticationConfiguration` type
- **`Auth.vue`** — Thread the option from API reference config to `AuthSelector`
- **`AuthSelector.vue`** — Accept new prop, compute total selectable options, hide dropdown when only one scheme exists
- **`security-scheme.ts`** — Accept `canAddNewAuth` parameter in `getSecuritySchemeOptions`, return flat available list when disabled with no required schemes
- **`security-scheme.test.ts`** — Add tests for `canAddNewAuth` behavior
- **`configuration.md`** — Document the new option

## Test plan

- [x] Verify auth dropdown shows all three groups (Required, Available, Add new) when `createAnySecurityScheme: true`
- [x] Verify "Add new authentication" group is hidden when `createAnySecurityScheme` is not set or `false`
- [x] Verify flat available list is returned when no required schemes and `createAnySecurityScheme: false`
- [x] Verify auth dropdown is hidden entirely when only one scheme is selectable
- [x] Verify existing behavior is unchanged for api-client (default `canAddNewAuth = true`)
- [x] Run `security-scheme.test.ts` unit tests
- [x] Added changeset

### DIFF
--- a/packages/api-client/src/v2/blocks/scalar-auth-selector-block/components/AuthSelector.vue
+++ b/packages/api-client/src/v2/blocks/scalar-auth-selector-block/components/AuthSelector.vue
@@ -104,20 +104,6 @@ const availableSchemeOptions = computed(() =>
   ),
 )
 
-/** Total number of selectable options across all groups (for hiding dropdown when only one exists) */
-const totalSelectableOptions = computed(() => {
-  const options = availableSchemeOptions.value
-  if (!Array.isArray(options)) return 0
-  // Check if grouped (SecuritySchemeGroup[]) or flat (SecuritySchemeOption[])
-  if (options.length > 0 && 'options' in options[0]) {
-    return (options as { options: unknown[] }[]).reduce(
-      (sum, group) => sum + group.options.length,
-      0,
-    )
-  }
-  return options.length
-})
-
 /** Currently active auth schemes selected for this operation or collection */
 const activeSchemeOptions = computed<SecuritySchemeOption[]>(() => {
   const schemes = selectedSecurity?.selectedSchemes
@@ -242,9 +228,7 @@ defineExpose({
     </template>
 
     <!-- Auth Dropdown (hidden when only one scheme is available) -->
-    <template
-      v-if="totalSelectableOptions > 1"
-      #actions>
+    <template #actions>
       <ScalarComboboxMultiselect
         class="w-72 text-xs"
         :modelValue="activeSchemeOptions"


amritk:
Hey thanks for the PR, but maybe we should opt in instead of opt out @hanspagel ?

hanspagel (changes requested):
Thanks for putting this together!

A few changes:

- [ ] Yep, let’s move the change behind a flag (defaults to: not allow new auth schemes)
- [ ] I’d call this `authentication.createAnySecurityScheme: true`
- [ ] It needs documentation in `documentation/configuration.md`
- [ ] `$ pnpm changeset`

packages/api-client/src/v2/blocks/scalar-auth-selector-block/helpers/security-scheme.ts
```diff
@@ -81,6 +81,8 @@ export const getSecuritySchemeOptions = (
   securitySchemes: NonNullable<ComponentsObject['securitySchemes']>,
   /** We need to add the selected schemes if they do not already exist in the calculated options */
   selectedSchemes: SecurityRequirementObject[],
+  /** When true, hides the generic "Add new authentication" options */
+  hideAddNewAuthentication?: boolean,
```
amritk:
```suggestion
  /** Allows adding authentication which is not in the document */
  canAddNewAuth = true,
```

packages/api-client/src/v2/blocks/scalar-auth-selector-block/helpers/security-scheme.ts
```diff
@@ -138,17 +140,19 @@ export const getSecuritySchemeOptions = (
...
+      label: 'Add new authentication',
+      options: Object.entries(authOptions).map(([key, value]) => ({
+        id: key,
+        label: value.label,
+        value: { [key]: [] },
+        payload: value.payload,
+        isDeletable: false,
+      })),
+    })
+  }
```
amritk:
```suggestion
  // We don't return the groups if we don't have any required schemes
  if (!canAddNewAuth) {
    return requiredFormatted.length ? options : availableFormatted
  }

  // Add new authentication options (unless explicitly hidden)
  options.push({
    label: 'Add new authentication',
    options: Object.entries(authOptions).map(([key, value]) => ({
      id: key,
      label: value.label,
      value: { [key]: [] },
      payload: value.payload,
      isDeletable: false,
    })),
  })
```
So we used to have this functionality previously, no need to show groups if there's only one group

jensenbox:
@hanspagel Thank you for the clear direction — all four items addressed:

- [x] Moved behind an opt-in flag (defaults to *not* allowing new auth schemes)
- [x] Renamed to `authentication.createAnySecurityScheme: true`
- [x] Added documentation in `documentation/configuration.md`
- [x] Added changeset (`.changeset/warm-fans-glow.md`)

Each is in its own commit for easy review. Happy to adjust anything further!

packages/api-client/src/v2/blocks/scalar-auth-selector-block/helpers/security-scheme.ts
jensenbox: Great suggestion — adopted exactly as written. The JSDoc is much clearer too. Thanks @amritk\!

packages/api-client/src/v2/blocks/scalar-auth-selector-block/helpers/security-scheme.ts
jensenbox: Love this — returning a flat list when there's only one group is much cleaner than showing empty groups. Adopted as suggested. Thank you\!

jensenbox:
@amritk Totally agree on opt-in over opt-out — makes much more sense as the default behavior. Addressed in the latest push along with both of your inline suggestions. Really appreciate the detailed code suggestions!

jensenbox:
@amritk Totally agree — opt-in is the right call here. Updated to `createAnySecurityScheme` (defaults to `false`). Thanks for the nudge!

hanspagel:
looks like the build fails because of a TS error 😭

packages/api-client/src/v2/blocks/scalar-auth-selector-block/components/AuthSelector.vue
```diff
@@ -97,9 +100,24 @@ const availableSchemeOptions = computed(() =>
     securityRequirements ?? [],
     securitySchemes ?? {},
     selectedSecurity?.selectedSchemes ?? [],
+    createAnySecurityScheme,
   ),
 )
 
+/** Total number of selectable options across all groups (for hiding dropdown when only one exists) */
+const totalSelectableOptions = computed(() => {
```
amritk: Also one more thing, I don't think we want to hide it if there's one option because you still want the ability to disable it. I will make a quick PR to ur pr

amritk: You should be able to target this and hide with css for your case if you like!

PR #8200, opened 2026-02-18
feat(api-client): add "Show Schema" toggle to request examples

wilhelmer:
Addresses #8278.

Adds an optional **Show Schema** toggle to request examples for endpoints that have a request body (POST/PUT/PATCH).

SCR-20260218-klqb

## Configuration

By default, the new toggle is **not shown**, keeping the current visual appearance of the API reference.

The PR adds a config option to control the behavior: 

#### showSchemaToggle

**Type:** `'responses' | 'requests' | 'both' | 'none'`

Controls which example panes display the "Show Schema" toggle that lets users switch
between examples and the raw OpenAPI schema.

- `responses`: Shows the toggle on response examples only
- `requests`: Shows the toggle on request examples only (POST/PUT/PATCH endpoints)
- `both`: Shows the toggle on both request and response examples
- `none`: Hides the toggle everywhere

**Default:** `'responses'`

### DIFF
--- a/.changeset/full-ducks-appear.md
+++ b/.changeset/full-ducks-appear.md
@@ -2,4 +2,4 @@
 '@scalar/api-client': minor
 ---
 
-feat: add "Show Schema" button to request body examples
+feat: add "Show Schema" toggle to request examples


hanspagel:
Actually, the PR is great. I’m a bit unsure about the feature.

It feels like we should show the raw schema for both where we render the schema, not where render the example.

And I think it shouldn't actually be a feature toggle, it should just blend it when we render the schemas visually.

So I'm closing this for now, sorry. Again, technically a great PR.

wilhelmer:
> Actually, the PR is great. I’m a bit unsure about the feature.

@hanspagel After giving it some more thought, I was also unsure and wanted to close the PR myself. After discussing this internally, we found a solution that doesn't require modifying the Scalar codebase.

So all good 👌

PR #8204, opened 2026-02-18
fix(api-client): remove DocumentRedirect component

amritk:
## Problem

This DocumentRedirect component was causing some warnings.

## Solution

While we could make it work, I felt better having the redirection in the app state.

## Checklist

- [x] I added a changeset. 
- [ ] I added tests.

---

> [!NOTE]
> **Medium Risk**
> Changes routing/redirect behavior for initial workspace/document navigation, so incorrect params or redirects could send users to the wrong page or loop, but the scope is limited to client-side navigation.
> 
> **Overview**
> **Removes the `DocumentRedirect` component** and changes the `document.redirect` route to a static redirect to `document.overview` instead of rendering a component.
> 
> Workspace navigation is adjusted so `navigateToWorkspace` now pushes directly to the `example` route for the default `drafts` doc (with `pathEncoded`, `method`, and `exampleName` params) and falls back to `/` if `namespace`/`slug` are missing; the “skip when on `document.redirect`” guard in `navigateToCurrentTab` is also dropped. Separately, the catch-all default route target is updated from `/@local/local/...` to `/@local/default/...`.
> 
> Includes a patch changeset for `@scalar/api-client`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 55e80450038a9eaeb02b8eeaa99734720bf1441f. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/api-client/src/v2/features/app/app-state.ts
+++ b/packages/api-client/src/v2/features/app/app-state.ts
@@ -343,7 +343,6 @@ export const createAppState = async ({
         exampleName: 'default',
       },
     })
-    return
   }
 
   /**
--- a/packages/api-client/src/v2/features/app/helpers/routes.ts
+++ b/packages/api-client/src/v2/features/app/helpers/routes.ts
@@ -81,6 +81,14 @@ export const ROUTES = [
             path: '',
             component: DocumentCollection,
             children: [
+              // Redirect to overview
+              {
+                name: 'document.redirect',
+                path: '',
+                redirect: {
+                  name: 'document.overview',
+                },
+              },
               // Document overview
               {
                 name: 'document.overview',


packages/api-client/src/v2/features/app/app-state.ts
```diff
@@ -321,15 +322,50 @@ export const createAppState = async ({
...
-   * Updates the route based on the given namespace and slug.
-   *
-   * @param namespace - The workspace namespace.
-   * @param slug - The unique workspace slug (identifier).
    */
-  const navigateToWorkspace = async (namespace?: string, slug?: string): Promise<void> => {
-    await router.push({
-      name: 'document.redirect',
-      params: { namespace, workspaceSlug: slug, documentSlug: 'drafts' },
+  const navigateToWorkspace = async (
```
amritk: was thinking of booting this out to a helper actually, clean up this file a bit

packages/api-client/src/v2/features/app/app-state.ts
```diff
@@ -354,35 +390,35 @@ export const createAppState = async ({
...
-      name,
+    const created = await createAndPersistWorkspace({ teamUid, namespace, slug: newSlug, name })
+    if (!created) {
+      console.error('Failed to create the workspace, something went wrong, can not load the workspace')
+      return undefined
     }
 
-    // Create a new client store with the workspace ID and add a default document.
-    const createdWorkspace = await createAndPersistWorkspace(newWorkspaceDetails)
+    const loaded = await loadWorkspace(created.namespace, created.slug)
```
DemonHa: it should load the workspace automatically when you navigate to it

packages/api-client/src/v2/features/app/app-state.ts
DemonHa: you can also remove this now

DemonHa (approved):

packages/api-client/src/v2/features/app/app-state.ts
```diff
@@ -321,16 +321,29 @@ export const createAppState = async ({
...
-      params: { namespace, workspaceSlug: slug, documentSlug: 'drafts' },
+      name: 'example',
+      params: {
+        namespace,
+        workspaceSlug: slug,
+        documentSlug: 'drafts',
+        pathEncoded: encodeURIComponent('/'),
+        method: 'get',
+        exampleName: 'default',
+      },
```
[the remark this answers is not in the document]
amritk: we only call this on new workspace

PR #8205, opened 2026-02-18
chore: went too far in the refactor

amritk:
I think the app state file is getting a bit crazy. This fixed up some things I saw like moving refs -> computed etc. I dont think we should merge this, but we should probably do some of these things, especially when we swtich to state based routing. We should also pull out some of this logic into helpers and have it tested extensively. 

## Checklist

- [ ] I added a changeset. 
- [ ] I added tests.


amritk:
sometimes u go tooo far

PR #8226, opened 2026-02-20
fix(api-reference): recursively extract heading text for correct ID generation

kou029w:
## Problem

Headings that include Markdown links (e.g. `## [foo](/link/to/foo)`) produced incorrect IDs because `node.children[0].value` is `undefined` when the first child is a non-text node such as a link.

## Solution

This change introduces a recursive text extraction using `textFromNode`, aligned with the behavior of `hast-util-to-text`, so that the full inner text of a heading is reliably obtained regardless of its child node structure. As a result, ID and slug generation now works correctly for all headings, including those containing links or other nested phrasing content.

Additionally, the custom `findTextInHeading` helper is removed and replaced with a shared `textFromNode` utility, which is exported from the markdown module and reused across components to ensure consistent heading text handling.

## Checklist

- [x] I added a changeset. 
- [x] I added tests.

### DIFF
--- a/packages/code-highlight/package.json
+++ b/packages/code-highlight/package.json
@@ -104,7 +104,6 @@
     "@scalar/themes": "workspace:*",
     "@types/hast": "^3.0.4",
     "@types/mdast": "^4.0.4",
-    "@types/unist": "^3.0.3",
     "vfile": "^6.0.1",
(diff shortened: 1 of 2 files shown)


hanspagel:
@cursoragent can you review?

PR #8239, opened 2026-02-23
fix(express): enable `knip`

marcalexiei:
## Problem

- Followup of #7233

## Solution

- remove unused `customTheme` variable
- use `RequestHandler` as return type in `apiReference`
   Correct types are inferred correctly from the explicit return type

## Checklist

- [x] I added a changeset. 
- [ ] I added tests.

### DIFF
--- a/.changeset/orange-loops-do.md
+++ b/.changeset/orange-loops-do.md
@@ -1,0 +1,5 @@
+---
+'@scalar/express-api-reference': patch
+---
+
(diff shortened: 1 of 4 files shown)


hanspagel (approved):

hanspagel:
thanks!

marcalexiei:
OMG less than a minute before approval😅

PR #8253, opened 2026-02-24
feat(api-reference): support propertyNames enum for additionalProperties

delputnam:
## Summary

- Adds support for displaying JSON Schema `propertyNames` enum values on `additionalProperties` in the API reference UI
- When a schema uses `additionalProperties` with a `propertyNames` constraint containing an `enum`, the allowed key names are now shown as enum chips (similar to value enums)
- Falls back to `propertyNames.title` for the key label when `x-additionalPropertiesName` is not set
- Uses the first `propertyNames` enum value as the example key in code samples

## Changes

- **workspace-store**: Added `propertyNames` to the `SchemaObject` type definition
- **api-reference**: Updated `SchemaObjectProperties.vue` to extract and pass `propertyNames` enum to `SchemaProperty`
- **api-reference**: Updated `SchemaProperty.vue` to accept and render `propertyNames` enum values
- **api-client**: Updated `get-example-from-schema.ts` to use `propertyNames` enum for example key generation
- **oas-utils**: Same changes to the deprecated copy of `get-example-from-schema.ts`

## Context

JSON Schema's `propertyNames` keyword constrains which keys are valid in an object with `additionalProperties`. When it contains an `enum`, those are the only allowed key names. This is commonly emitted by frameworks like Pydantic when a model uses `dict[SomeEnum, ValueType]`.

Before this change, Scalar displayed these properties with a generic "unknown property name" label and no indication of the allowed keys. Now the enum values are displayed, giving API consumers visibility into which keys are valid.

`x-additionalPropertiesName` takes precedence over `propertyNames` when both are present.

## Test plan

- [x] Added 4 new tests for `propertyNames` enum in code sample generation
- [x] All existing tests pass (api-client, api-reference Schema, oas-utils, workspace-store)


hanspagel:
Just had to look up `propertyNames`, for reference:

https://json-schema.org/understanding-json-schema/reference/object#propertyNames

hanspagel:
I played around with that and noticed, that it's the same rendering as we use for values.

So what I changed:

* the enum ascii tree got a label to indicate whether we show enum values or the enum property names
* I used the accent color and a dashed border to indicate the property name is dynamic (not static)

approved by @cameronrohani 

preview:

* foobar: regular string enum
* “propertyName”: a string enum, where we don’t know the exact property name, but it must be “foo” or “bar”, the value must be “1", “2” or “3".

Screenshot 2026-03-10 at 15 11 56

Issue #8176, opened 2026-02-15
Server variables not substituted in OAuth2 authorization URLs


hillsinfoodle:
### What happens?

When the server URL uses variables (e.g. https://{sitecode}.example.com/apiv2) and a sitecode value is set to test, the OAuth2 auth flow navigates to the literal {sitecode}.example.com instead of test.example.com. Variable substitution works for display but not for constructed OAuth2 URLs.

### What did you expect to happen?

variable replacement in URL

PR #8254, opened 2026-02-24
fix(api-client): resolve OAuth2 URLs against interpolated server URL

rcollas:
###  Problem

When a server URL contains template variables (e.g. https://{env}.example.com), OAuth2 flows were using the raw, uninterpolated URL as the base for resolving authorization and token URLs. This caused OAuth to break for anyone using server variables in their spec.

Fixes https://github.com/scalar/scalar/issues/8176

 ### Solution

Interpolate server URL variables before using the URL in OAuth2 flows. Extracted a getInterpolatedServerUrl helper for the v1 flow and reused the existing getServerUrl helper in v2. Also threaded environmentVariables through to authorizeServers so the token URL gets the same treatment as the request-block.

 ### Checklist

  - [x] I added a changeset.
  - [x] I added tests.

### DIFF
--- a/packages/api-client/src/v2/blocks/scalar-auth-selector-block/helpers/oauth.test.ts
+++ b/packages/api-client/src/v2/blocks/scalar-auth-selector-block/helpers/oauth.test.ts
@@ -1201,7 +1201,7 @@ describe('oauth', () => {
           'x-scalar-secret-client-secret': clientSecret,
           'x-scalar-secret-token': '',
         },
-      } satisfies OAuthFlowsObject
+      } satisfies OAuthFlowsObjectSecret
 
       const serverWithVars: ServerObject = {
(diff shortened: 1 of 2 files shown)


hanspagel:
@DemonHa can u review this one? I’m a bit confused we'd use environment variables here, but maybe that's where the values of the server variables are stored, I don't know

DemonHa:
> @DemonHa can u review this one? I’m a bit confused we'd use environment variables here, but maybe that's where the values of the server variables are stored, I don't know

Yeah actually this takes care for both cases. 

- `{variable}` → standard server variables
- `{{variable}}` → environment variables

We try to preserve this from the old client which I think it's something that users might want since we want to be able to use environment variables pretty much everywhere

PR #8272, opened 2026-02-26
feat(openapi-upgrader): migrate Swagger 2.0 response examples

hanspagel:
looks like we had a bug in the upgrader for Swagger 2.0 response examples, they were put in as media type objects (it's in the application/json dropdown)

**Before**

Screenshot 2026-02-26 at 16 04 14

**After**

Screenshot 2026-02-26 at 16 17 06

Screenshot 2026-02-26 at 16 51 35

---

> [!NOTE]
> **Medium Risk**
> Changes how Swagger 2.0 `responses[*].examples` are interpreted and mapped into OpenAPI 3.0 `content`, which can alter generated specs for existing users (especially when example keys aren’t real media types). Added tests reduce regression risk but output shape changes are user-visible.
> 
> **Overview**
> Fixes Swagger 2.0 → OpenAPI 3.0 response example migration so **named examples** are emitted under `content[defaultMediaType].examples` (as `ExampleObject`s) instead of being treated as media-type entries.
> 
> Adds media-type key detection (`isMediaTypeKey`) to distinguish real `type/subtype` keys from named keys that may include slashes (e.g. `Error 404/Not Found`), introduces a shared `DEFAULT_MEDIA_TYPE`, and extends the test suite to cover minimal specs, named examples, multi-media examples, and slash-containing example names. Includes a patch changeset for `@scalar/openapi-upgrader`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 7f33397cf396360eb82ecac43bc4eb8598e472e4. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/openapi-upgrader/src/2.0-to-3.0/upgrade-from-two-to-three.ts
+++ b/packages/openapi-upgrader/src/2.0-to-3.0/upgrade-from-two-to-three.ts
@@ -71,9 +71,17 @@ function wrapAsExampleObject(value: unknown): OpenAPIV3.ExampleObject {
   return { value }
 }
 
-/** True if the key looks like a media type (e.g. application/json). Used to distinguish from named example keys. */
+/**
+ * True if the key looks like a MIME media type (e.g. application/json, text/plain).
+ * Used to distinguish media-type example keys from named example keys when migrating
+ * Swagger 2.0 examples to OpenAPI 3.0 content. Requires exactly one slash and
+ * token-style type/subtype (no spaces or semicolons) to avoid false positives
+ * for named keys that contain a slash (e.g. "Error 404/Not Found").
+ */
+const MEDIA_TYPE_KEY_PATTERN = /^[a-zA-Z0-9*+.-]+\/[a-zA-Z0-9*+.+-]+$/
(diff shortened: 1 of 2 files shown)


packages/openapi-upgrader/src/2.0-to-3.0/upgrade-from-two-to-three.ts
```diff
@@ -69,6 +71,11 @@ function wrapAsExampleObject(value: unknown): OpenAPIV3.ExampleObject {
   return { value }
 }
 
+/** True if the key looks like a media type (e.g. application/json). Used to distinguish from named example keys. */
+function isMediaTypeKey(key: string): boolean {
+  return key.includes('/')
+}
```
hanspagel, line 85: woke up in the middle of the night, remembered adding this as a placeholder, need to improve this before we merge

hanspagel: done

PR #8275, opened 2026-02-26
fix(openapi-types): make $ref property required in ReferenceObject

marikaner:
## Problem

The $ref property in the ReferenceObject should not be optional. Otherwise it is not a ReferenceObject.
This currently makes it very difficult for me to use the types without bigger workarounds.

## Solution

Remove the `?` ;) 

## Checklist

- [x] I added a changeset. 
- [ ] I added tests.

### DIFF
--- a/packages/openapi-types/src/openapi-types.ts
+++ b/packages/openapi-types/src/openapi-types.ts
@@ -428,7 +428,7 @@ export namespace OpenAPIV3_1 {
 
   export type ParameterLocation = OpenAPIV3.ParameterLocation
 
-  export type ParameterObject = OpenAPIV3.ParameterObject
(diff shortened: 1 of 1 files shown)


hanspagel:
ok, CI says nope

I mean, technically the change is great, it's just that it requires more work in the other packages

let's go back to draft for now

hanspagel:
I think I got it, let's run CI again

packages/openapi-types/src/openapi-types.ts
```diff
@@ -775,7 +775,7 @@ export namespace OpenAPIV3 {
   }
 
   export type ReferenceObject = {
-    $ref?: string
+    $ref: string
```
hanspagel, line 778: this is the actual change

Issue #8282, opened 2026-02-27
fix(java): use relative paths for html script.js asset in scalar-webmvc


mveselov-stack-wizard:
## Summary

The Scalar UI fails to load when a Spring Boot application is deployed behind a reverse proxy and a custom server context path. 

Used [`scalar-webmvc`](https://mvnrepository.com/artifact/com.scalar.maven/scalar-webmvc) version `0.5.56` (Feb 24, 2026).

Other dependencies like [`springdoc-openapi`](https://mvnrepository.com/artifact/org.springdoc/springdoc-openapi-starter-webmvc-ui) use relative paths by default to ensure portability, e.g. when serving swagger ui.

## Problem

Suppose configuration `scalar.path: /scalar-ui`.

Currently, `ScalarHtmlRenderer` generates asset paths as absolute paths `/scalar-ui/scalar.js` instead relative paths `./scalar-ui/scalar.js`.

Serving scalar ui at `example.com/myapp/scalar-ui`  tries resolving `example.com/scalar-ui/scalar.js` instead `example.com/myapp/scalar-ui/scalar.js`.

## Changes

Modifying a line of code by adding the `.` prefix for relative paths in `com.scalar.maven.core.ScalarHtmlRenderer` like

```
return "." + path + "/" + ScalarConstants.JS_FILENAME;
```

would resolve the issue in all cases and also align the behaviour of the dependency with other established dependencies like mentioned.

This change should not cause any other effects, but can someone familiar with the scalar mvc code base confirm?

## Context

I work on a spring boot mvc app. These apps are usually deployed with a context path and behind a reverse proxy like nginx.

This means scalar ui does not resolve properly in our app, but something like springdoc swagger ui does resolve as expected.

For example, when the root path is reserved for serving the frontend and context path prefix is used to serve the backend.

The problem is not present when testing locally with our setup (have root and prefix serverlets), but it is present in the standard deployment setup.

Current workaround is to override `configureProperties` in `ScalarWebMvcController` and force relative path prefix `.` in path property.

This workaround confirms the case.

### What did you expect to happen?

The scalar mvc dependency should be aligned with other similar mvc dependencies to serve relative paths.

hanspagel:
cool, can you submit a PR?

PR #8284, opened 2026-02-27
feat(java): relative paths for assets in scalar-webmvc

mveselov-stack-wizard:
Fixes: https://github.com/scalar/scalar/issues/8282

---

## Problem

The Scalar UI HTML fails to load when a Spring Boot application is deployed behind a reverse proxy and a server context path.

Currently, `ScalarHtmlRenderer` generates `/scalar/scalar.js` instead relative path `./scalar/scalar.js` in served HTML.

Other dependencies like Springdoc's Swagger UI HTML use relative paths by default, .e.g. `./swagger-ui-bundle.js`.

image

## Solution

Add the `.` prefix for relative paths in `com.scalar.maven.core.ScalarHtmlRenderer`

```
return "." + path + "/" + ScalarConstants.JS_FILENAME;
```

## Workaround

Override `configureProperties` in `ScalarWebMvcController` and modify the path property there.

Append relative path prefix `.`  or append the actual server context path prefix, e.g. `/myapp`.

## Checklist

- [ ] I added a changeset. 
- [ ] I added tests.

### DIFF
--- a/.changeset/chubby-tools-kneel.md
+++ b/.changeset/chubby-tools-kneel.md
@@ -1,0 +1,5 @@
+---
+'@scalar/java-integration': patch
+---
+
+feat(java): relative paths for assets in scalar-webmvc


hanspagel (commented):
ok cool, just needs a changeset

PR #8292, opened 2026-02-27
feat(aspire): serve static OpenAPI documents via file server

### LINKED ISSUE #8270: feat(aspire): Serve static OpenAPI documents via file server
## Summary

Some REST APIs implement a contract but do not have an OpenAPI endpoint. It would be great if `Scalar.Aspire` could serve static OpenAPI documents so that no additional resources are required.

## Motivation

Not all APIs expose a live OpenAPI endpoint. Allowing Scalar.Aspire to serve static OpenAPI documents (e.g., from a file) would remove the need for additional infrastructure just to display API documentation.

## Proposed Solution

Use [`UseFileServer`](https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.builder.fileserverextensions.usefileserver) to serve static OpenAPI files. This can be:

- Scoped to a specific path
- Guarded/protected similarly to how [YARP does it](https://github.com/dotnet/yarp/blob/main/src/Application/Program.cs#L34-L38)

## References

- Discussion: https://github.com/scalar/scalar/discussions/8180

paulomorgado:
## Problem

 Some REST APIs don't expose a live `/openapi.json` endpoint — for example, legacy services, third-party APIs, or services whose spec is maintained separately. When using `Scalar.Aspire`, these services
couldn't be documented because `WithApiReference` assumed the target resource would serve its own OpenAPI document at runtime.

Closes #8270 

 ## Solution

 Added `WithApiReference(resourceBuilder, FileInfo, ...)` overloads that bind-mount a local OpenAPI file (JSON or YAML) into the Scalar container and serve it via a new static file middleware at
`/openapi/{filename}`.

 Key changes:
 - **`WithApiReference(IResourceBuilder, FileInfo)`** — mounts the file into the container and configures the Scalar UI to load it from the container's own static endpoint. The service resource is still
referenced so the "Try It" server URL is set correctly.
 - **`UseOpenApiFiles()` middleware** — added to `Scalar.Aspire.Service`; serves files from `/openapi` only when the directory exists (i.e., when a file has been bind-mounted). Includes YAML MIME type
mappings.
 - **`BaseDocumentUrl` (`ReferenceExpression?`)** — new property on the Aspire-specific `ScalarOptions` partial. Allows overriding or suppressing the service base URL when resolving the OpenAPI document
URL. Typed as `ReferenceExpression` because endpoint values are only known at `ConfigureScalarResourceAsync` execution time. `ReferenceExpression.Empty` is used internally for the static file case to
signal "no prefix".
 - **`WithBaseDocumentUrl(ReferenceExpression?)`** — fluent extension for advanced users who want to point the Scalar UI at a custom document URL.

 The alternative considered was serving the file from the application host side and exposing it via a separate HTTP endpoint, but mounting directly into the Scalar container keeps the solution
self-contained and avoids requiring users to add an extra resource.

 ## Checklist

 - [ ] I added a changeset. 
 - [x] I added tests.

### DIFF
--- a/documentation/integrations/aspire.md
+++ b/documentation/integrations/aspire.md
@@ -8,7 +8,7 @@ Scalar for Aspire provides:
 
 - **Unified API Documentation**: View documentation for all services in a single, cohesive interface
 - **Simplified Service Discovery**: Automatically discover and configure API endpoints from your Aspire services
-- **Multiple Document Support**: Each service can expose multiple OpenAPI specifications
+- **Multiple Document Support**: Each service can expose multiple OpenAPI documents
 - **CORS Issue Elimination**: Built-in proxy (enabled by default) handles API requests without requiring CORS configuration
 - **HTTPS Support**: Complete support for both HTTP and HTTPS endpoints with automatic handling
 
@@ -18,10 +18,10 @@ The Scalar Aspire integration requires a container solution such as **Docker** o
 
 ### Service Requirements
 
-Each service you want to include in the API Reference must:
+Each service you want to include in the API Reference must implement the `IResourceWithServiceDiscovery` interface and either:
 
-- Expose OpenAPI documents over HTTP or HTTPS endpoints
-- Implement the `IResourceWithServiceDiscovery` interface
+- Expose OpenAPI documents over HTTP or HTTPS endpoints, or
+- Provide a static OpenAPI document (a local file) via the file-based `WithApiReference` overload
 
 ## Quick Start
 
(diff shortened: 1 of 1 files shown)


paulomorgado:
> great PR, thanks! we need to mention it in the docs and add a changeset (`pnpm changeset`)

@hanspagel, what do I need to do?

hanspagel:
I do it :) 1 sec

paulomorgado:
@hanspagel, I found a few issues with this and I'm working on it.

paulomorgado:
@hanspagel, I think I fixed it: #8333

PR #8298, opened 2026-03-01
feat(api-reference): show explicit no-content response tabs

silkfire:
## Summary
- show explicitly defined response keys in Example Responses tabs even when there is no body content
- include `204`/`default`/range keys (for example `2XX`) when present in the operation `responses`
- keep existing behavior for rendering body content and empty-state (`No Body`)
- ignore non-response keys (for example `x-*`) in tab rendering

## Changes
- updated response filter helper to accept an optional response key and validate valid OpenAPI response keys
- wired key-aware filtering in `ExampleResponses.vue`
- added helper tests for explicit no-content response keys and invalid keys
- added component tests for explicit `204` rendering, mixed contentful/no-content tabs, and non-response key exclusion

Refs discussion: https://github.com/scalar/scalar/discussions/8264

### DIFF
--- a/packages/api-reference/src/features/example-responses/ExampleResponses.test.ts
+++ b/packages/api-reference/src/features/example-responses/ExampleResponses.test.ts
@@ -709,4 +709,80 @@ describe('ExampleResponses', () => {
     expect(wrapper.text()).toContain('New Single Example')
     expect(wrapper.text()).not.toContain('Example 2')
   })
+
+  it('renders explicit 204 response tab and shows No Body', () => {
+    const wrapper = mount(ExampleResponses, {
+      props: {
+        responses: {
+          '204': {
+            description: 'No Content',
+          },
+        },
+      },
+    })
+
+    const tabs = wrapper.findAllComponents({ name: 'ExampleResponseTab' })
+    expect(tabs.length).toBe(1)
+    expect(tabs[0]?.text()).toContain('204')
+    expect(wrapper.text()).toContain('No Body')
+  })
+
+  it('renders both contentful and explicit no-content responses', async () => {
+    const wrapper = mount(ExampleResponses, {
+      props: {
+        responses: {
+          '200': {
+            description: 'Success response',
+            content: {
+              'application/json': {
+                examples: {
+                  example1: { value: { status: 'success' } },
+                },
+              },
+            },
+          },
+          '204': {
+            description: 'No Content',
+          },
+        },
+      },
+    })
+
+    const tabs = wrapper.findAllComponents({ name: 'ExampleResponseTab' })
+    expect(tabs.length).toBe(2)
+    expect(tabs[0]?.text()).toContain('200')
+    expect(tabs[1]?.text()).toContain('204')
+
+    await tabs[1]?.trigger('click')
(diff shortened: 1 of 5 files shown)


packages/api-reference/src/features/example-responses/ExampleResponses.test.ts
```diff
@@ -709,4 +709,80 @@ describe('ExampleResponses', () => {
...
+              'application/json': {
+                examples: {
+                  example1: { value: { status: 'success' } },
+                },
+              },
+            },
+          },
+          'x-internal': {
+            description: 'Internal metadata',
+          },
```
hanspagel: we expect x-internal to be a boolean flag for hiding stuff, so this is kind of confusing. can we use a different key here?

packages/api-reference/src/features/example-responses/has-response-content.ts
```diff
@@ -21,11 +21,23 @@ function hasMediaTypeContent(mediaType: MediaTypeObject | undefined): boolean {
   return hasSchema || hasExample || hasExamples
 }
 
+function isResponseKey(responseKey: string): boolean {
+  return responseKey === 'default' || /^[1-5][0-9]{2}$/.test(responseKey) || /^[1-5]XX$/.test(responseKey)
```
hanspagel, line 25: haven't seen this `[1-5]XX` specifically in the wild, do you consider this a convention?

hanspagel (changes requested):
Thanks for submitting this PR! You took another approach than I expected and now I’m wondering:

* Should we only allow 204 to have no content (because that's always the case)?
* Or should we just go back and show all responses, whether they have a body or not?

The decision to just show "204", "5XX", but not "5xx" or responses labeled as "no content" (or whatever string) seems kind of random.

silkfire:
> Thanks for submitting this PR! You took another approach than I expected and now I’m wondering:
> 
> * Should we only allow 204 to have no content (because that's always the case)?
> * Or should we just go back and show all responses, whether they have a body or not?
> 
> The decision to just show "204", "5XX", but not "5xx" or responses labeled as "no content" (or whatever string) seems kind of random.

To be honest I think any response that doesn't have any content should be shown. Do you think this should be an opt-in feature that consumers should be able to configure in options?

packages/api-reference/src/features/example-responses/has-response-content.ts
silkfire: It's the defined range form in OpenAPI, so nothing custom here actually :)

Source:

https://spec.openapis.org/oas/v3.1.1.html#patterned-fields-0

packages/api-reference/src/features/example-responses/ExampleResponses.test.ts
silkfire: Changed it to use a non-response key instead

packages/api-reference/src/features/example-responses/has-response-content.ts
hanspagel: ha! learning something every day

hanspagel (approved):
okay, I’m convinced, thanks for the PR :))

PR #8324, opened 2026-03-03
feat(snippetz): code splitting

hanspagel:
## Problem

The whole `@scalar/snippetz` package has 440 kb, though for our use case we only need a single plugin until the users selects another plugin, and probably no one ever needs all plugins loaded.

## Solution

This PR refactors the library to allow lazy loading plugins:

https://github.com/user-attachments/assets/5c8aacca-04b1-42f1-aa69-41b681dcabff

Unfortunately, the main distribution way for `@scalar/api-reference` is to just bundle everything together, so we won't see an immediate benefit. We'd need to switch to an ESM CDN as the main distribution (I’ll try that again asap).

I made this a major release, and I think I’ll add a bit more to the PR for the v1. So just a draft for now.

## Checklist

- [x] I added a changeset. 
- [x] I added tests.

### DIFF
--- a/.changeset/silly-pears-rush.md
+++ b/.changeset/silly-pears-rush.md
@@ -1,0 +1,5 @@
+---
+'@scalar/snippetz': major
+---
+
+feat!: lazy load plugins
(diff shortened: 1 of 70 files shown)


hanspagel:
nobody got time for this

PR #8333, opened 2026-03-03
feat(aspire): add static OpenAPI file serving and fix BaseDocumentUrl ownership

paulomorgado:
## Problem

 The `WithApiReference(FileInfo)` overloads always mounted the file directly under `/openapi/{filename}`, making it impossible to have two services with identically named OpenAPI files
without collisions.

 `BaseDocumentUrl` was always overwritten with the `resourceBuilder` endpoint URL during configuration — even when `WithApiReference(fileInfo)` had already set it to signal "use the path
 as-is". This caused the URL to be doubled (e.g. `http://my-api/openapi/my-api.json/openapi/my-api/spec.yaml`), and meant `WithBaseDocumentUrl()` user overrides could also be silently
discarded.

 ## Solution

 ### Optional folder path

 Added an optional `folderPath` parameter to both `WithApiReference(FileInfo)` overloads. When omitted, the resource name is used as the default folder, so files are mounted at
`/openapi/{resourceName}/{filename}` out of the box.

 ```csharp
 // default → /openapi/my-service/openapi.yaml
 scalar.WithApiReference(service, new FileInfo("openapi.yaml"));

 // explicit → /openapi/v2/openapi.yaml
 scalar.WithApiReference(service, new FileInfo("openapi.yaml"), folderPath: "v2");
```

### BaseDocumentUrl ownership redesign

ScalarAnnotation is now the sole owner of BaseDocumentUrl. The configurator no longer overwrites it:

 - WithApiReference(resourceBuilder) stores a ResourceBaseUrlExpression on the annotation, configured lazily after the user's callback so DefaultProxy/PreferHttpsEndpoint are final.
 - WithApiReference(fileInfo) sets annotation.BaseDocumentUrl = ReferenceExpression.Empty, signalling the configurator to use OpenApiRoutePattern as-is — no service URL prepended.
 - ScalarOptions.BaseDocumentUrl (via WithBaseDocumentUrl()) is reserved for explicit user overrides and takes priority over the annotation value.
 - ConfigureOpenApiRoutePatternAsync no longer has a resourceUrl fallback; it reads scalarOptions.BaseDocumentUrl ?? annotation.BaseDocumentUrl.

### Tests

17 unit tests covering the full API_REFERENCE_CONFIG generation surface: endpoint URL, proxy mode, HTTPS preference, static file path, folder path (default and explicit),
WithBaseDocumentUrl overrides, absolute URL pass-through, and multi-annotation scenarios.

## Checklist

 - [X]  I explained why the change is needed.
 - [ ]  I added a changeset. 
 - [X]  I added tests.
 - [ ]  I updated the documentation.

### DIFF
--- a/documentation/integrations/aspire.md
+++ b/documentation/integrations/aspire.md
@@ -116,14 +116,26 @@ scalar.WithApiReference(bookService, options =>
 
 ### Static OpenAPI Documents
 
-Use the file-based overload when a service does not expose a live OpenAPI endpoint—for example, when the description document is generated at build time or when documenting an external API from a local file. The file is mounted into the Scalar container and served at `/openapi/{filename}`. The document URL in the API Reference uses that path; the `resourceBuilder` is still used to configure the **Try It** server URL (the API base URL for requests).
+Use the file-based overload when a service does not expose a live OpenAPI endpoint—for example, when the description document is generated at build time or when documenting an external API from a local file. The file is mounted into the Scalar container and served at `/openapi/{folderPath}/{filename}`. When the optional `folderPath` parameter is not provided, the resource name is used as the folder, so the path is `/openapi/{resourceName}/{filename}`. You can pass an explicit `folderPath` to override the default folder or to avoid name collisions when multiple services serve static files. The document URL in the API Reference uses that path; the `resourceBuilder` is still used to configure the **Try It** server URL (the API base URL for requests).
 
 Supported file formats: `.json`, `.yaml`, `.yml`.
+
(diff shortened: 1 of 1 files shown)


paulomorgado:
> thanks again!
> 
> needs a changeset (`$ pnpm changeset`) and an addition to the docs
> 
> you can do, or I’ll do when I’ll find the time (may take a few days this time)

what do I need to do to do that?

hanspagel:
sorry, was super busy
just finished up, thanks for the patience :))

PR #8371, opened 2026-03-08
test(api-reference): verify built css does not contain deep selectors

### LINKED ISSUE #8044: Write test that checks built css for deep selectors
### What happens?

if vue hasn't removed the :deep from the built css, it will crash on bun

See https://github.com/scalar/scalar/issues/8042

Ensure none of our built css includes :deep

### OpenAPI Document

_No response_

AbdelhamidKhald:
## Problem

`@scalar/api-reference` uses Vue `:deep(...)` selectors in source styles, but those selectors should not survive in the built CSS output. If `:deep` leaks into the published stylesheet, Bun can crash as described in #8044.

## Solution

- add a regression test for the built `@scalar/api-reference` stylesheet
- verify the generated CSS does not contain `:deep`

## Validation

- `pnpm --filter @scalar/build-tooling build`
- `pnpm --filter @scalar/themes build`
- `pnpm --filter @scalar/api-reference build:default`
- `pnpm exec biome check --diagnostic-level=error packages/api-reference/src/dist.test.ts`
- `pnpm exec vitest run packages/api-reference/src/dist.test.ts`

## Checklist

- [ ] I added a changeset.
- [x] I added tests.

Fixes #8044

### DIFF
--- a/packages/api-reference/src/dist.test.ts
+++ b/packages/api-reference/src/dist.test.ts
@@ -1,0 +1,9 @@
+import { describe, expect, it } from 'vitest'
+
+import builtCss from '../dist/style.css?raw'
+
+describe('dist', () => {
+  it('does not include Vue deep selectors in built css', () => {
+    expect(builtCss).not.toContain(':deep')
+  })
+})


hanspagel:
thanks for the PR! isn't that something we can do in the linting? just thinking out loud here

PR #8372, opened 2026-03-09
fix(workspace-store): avoid inlining $ref path parameters during path parameter sync

DemonHa:
This PR fixes an issue with how path parameters were being synchronized.

Previously, the synchronization logic caused `$ref` path parameters to be inlined by the bundler, instead of preserving the reference. Additionally, the sync process was replacing all path parameters, including those that did not require any changes, which could lead to unnecessary reordering at the end of the parameter list.

**Changes**

- Preserve $ref path parameters instead of inlining them during the sync process.
- Mutate the original referenced object rather than replacing the reference.
- Avoid unnecessary reordering by only updating path parameters that actually require changes.

These changes ensure that references remain intact and that the parameter list stays stable unless a modification is truly required.

## Checklist

- [x] I added a changeset. 
- [x] I added tests.

---

> [!NOTE]
> **Medium Risk**
> Changes how operation `parameters` are rebuilt during path/operation updates and bundling; incorrect filtering or mutation could drop/rename parameters or alter ordering, but scope is limited and covered by new tests.
> 
> **Overview**
> Fixes path-parameter synchronization so referenced (`$ref`) path parameters stay as references instead of being inlined/resolved during sync, including when path variables are added or renamed.
> 
> Refactors `syncParametersForPathChange` to take a `resolve` callback, mutate renamed path params in-place, and rebuild the returned `parameters` list from existing raw nodes plus newly-created minimal path params (dropping removed path params while leaving non-path params untouched). Updates operation mutators and the bundler `syncPathParameters` plugin accordingly, and adds targeted unit/integration tests to assert `$ref` preservation, de-duplication with path-item parameters, and stable parameter ordering.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 0fa709937af262cb8b0458e4edfcba2d86362742. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/workspace-store/src/mutators/operation/helpers/sync-path-parameters.ts
+++ b/packages/workspace-store/src/mutators/operation/helpers/sync-path-parameters.ts
@@ -70,7 +70,7 @@ export const syncParametersForPathChange = <T extends MinimalParameterObject>(
   const newPositions = getParameterPositions(newPath, newPathParams)
 
   // Keep a map of path parameters by name for quick lookup (references unchanged objects)
-  const pathParameters: Record<string, ReferenceType<T>> = {}
+  const pathParameters: Record<string, ReferenceType<T>> = Object.create(null)
 
   // Populate the map with the existing path parameters
   for (const param of existingParameters) {


marclave:
just gotta fix up the types

PR #8387, opened 2026-03-10
fix(api-client): restore the example picker in the request body editor

### LINKED ISSUE #8384: example picker in request body missing
there used to be an example picker in the request body editor, but it's missing now

Image

VedantMadane:
Fixes #8384

## Summary
The example picker was missing from the request body editor. This PR restores it.

## Changes
- Added a ScalarListbox (example picker) that appears when the OpenAPI spec defines examples for the request body content type
- The picker shows next to the content type selector in the request body header
- When users select a different example, the request body value is updated with that example's payload
- Supports both single and multiple examples (per PR #1410, shows even with one example)
- Uses \summary\ from the example object when available, falls back to the example key

## Testing
- The example picker appears when \
equestBody.content[contentType].examples\ has entries
- Selecting an example loads its value into the editor
- The selected example is determined by matching the current body value to the examples

### DIFF
--- a/.changeset/restore-example-picker.md
+++ b/.changeset/restore-example-picker.md
@@ -1,0 +1,7 @@
+---
+'@scalar/api-client': patch
+---
+
+fix: restore the example picker in the request body editor
+
+Fixes #8384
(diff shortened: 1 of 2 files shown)


hanspagel:
thanks for putting this together

I’m not a big fan of the huge computed

but I leave it up to @DemonHa to review it 👍

DemonHa:
Hi @VedantMadane, thank you for your contribution but for api-client the example is already selected from the sidebar, so there is no need for an example picker on the request body component.

This is the example selector for the client: 
Screenshot 2026-03-10 at 2 21 36 PM

cc @hanspagel

PR #8397, opened 2026-03-10
fix(api-reference): render direct download as source anchor link

### LINKED ISSUE #8378: `documentDownloadType: 'direct'` still renders as a `<button>` instead of an `<a href>` link
## Description

Setting `documentDownloadType: 'direct'` in `createApiReference` config should, per the [documentation](https://scalar.com/products/api-references/configuration#documentdownloadtype):

> "When 'direct' is passed, it just outputs a regular link to the passed URL."

However, in testing, the element still renders as a `<button>` that only triggers a JSON download — not a navigable `<a href>` anchor.

## Configuration

```js
  sources: [
    { title: "Routines", url: "/routinemanager-examples.json" },
    // ... more sources
  ],
  darkMode: true,
  hideClientButton: true,
  documentDownloadType: 'direct',
  searchHotKey: 'k',
})
```

## Observed behavior

The download element still renders as a `<button>`:

```html
<button data-v-59b5011b="" class="download-button" type="button">
  <span data-v-59b5011b=""> Download OpenAPI Document </span>
  <div data-v-3dedb7e4="" data-v-59b5011b="" class="badge extension hidden group-hover:flex">json</div>
</button>
```

## Expected behavior

A plain `<a href="...">` anchor linking directly to the spec URL (so the spec can be opened in a new tab, copied as a link, etc.).

## Version

`@scalar/api-reference` **1.48.0**

## Additional context

PR #6854 introduced this feature, but the [current implementation](https://github.com/scalar/scalar/blob/main/packages/api-reference/src/blocks/scalar-info-block/components/DownloadLink.vue) no longer appears to do this.

### OpenAPI Document

_No response_

SergioChan:
## Problem

`documentDownloadType: 'direct'` currently renders the same button flow as JSON download, which triggers an event that fetches and downloads the spec. This does not match the documented behavior for `direct`, which says it should render a regular link to the configured source URL.

Fixes #8378.

## Solution

- Pass the document source URL (`x-scalar-original-source-url`) into the info block download component.
- Render an `<a href="..." target="_blank">` for `documentDownloadType: 'direct'` instead of a `<button>`.
- Keep existing button behavior for `json`, `yaml`, and `both`.
- Add tests covering direct-link rendering, no event emission for direct mode, and hidden state when no direct URL is available.
- Add a changeset for `@scalar/api-reference`.

## Checklist

- [x] I added a changeset.
- [x] I added tests.

## Validation

- `pnpm exec eslint --no-warn-ignored packages/api-reference/src/blocks/scalar-info-block/components/DownloadLink.vue packages/api-reference/src/blocks/scalar-info-block/components/InfoBlock.vue packages/api-reference/src/components/Content/Content.vue`
- Attempted: `pnpm vitest run packages/api-reference/src/blocks/scalar-info-block/components/DownloadLink.test.ts packages/api-reference/src/blocks/scalar-info-block/components/InfoBlock.test.ts` (blocked in this environment by unresolved `@scalar/build-tooling` project config resolution during Vitest multi-project startup).

### DIFF
--- a/.changeset/funny-coats-share.md
+++ b/.changeset/funny-coats-share.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
+fix: render `documentDownloadType: "direct"` as a direct anchor link to the source URL
--- a/packages/api-reference/src/blocks/scalar-info-block/components/DownloadLink.test.ts
+++ b/packages/api-reference/src/blocks/scalar-info-block/components/DownloadLink.test.ts
@@ -40,6 +40,7 @@ describe('DownloadLink', () => {
       props: {
         eventBus: mockEventBus,
         documentDownloadType: config.value.documentDownloadType,
+        documentDownloadUrl: 'https://example.com/openapi.json',
         ...props,
       },
     })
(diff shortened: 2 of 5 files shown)


hanspagel:
Thanks, mate! I got you already: #8380

PR #8399, opened 2026-03-11
fix(api-client): restore the example picker in the request body editor

### LINKED ISSUE #8384: example picker in request body missing

VedantMadane:
Restores the example picker in the request body editor that was previously missing.

Fixes #8384

The example picker allows users to select from predefined examples in the OpenAPI spec when editing request bodies. It appears as a dropdown next to the content type selector when the request body has multiple examples defined.

### DIFF
--- a/.changeset/restore-example-picker.md
+++ b/.changeset/restore-example-picker.md
@@ -1,0 +1,7 @@
+---
+'@scalar/api-client': patch
+---
+
+fix: restore the example picker in the request body editor
+
+Fixes #8384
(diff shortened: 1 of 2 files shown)


hanspagel:
wait, needs discussion in the issue first

DemonHa:
If you wanna share your ideas feel free to do so here: https://github.com/scalar/scalar/discussions/8413

PR #8403, opened 2026-03-11
feat(api-client): persist auth url and token url in auth secrets

liamdscalar:
## Problem

Adds auth url and token url to auth secrets store so that they are persisted and also accessible when exporting the store.

## Checklist

- [ ] I added a changeset. 
- [ ] I added tests.


hanspagel (approved):
LGTM

I don't know much about how auth works rn, so let's see what CI and big brain cursor say

PR #8404, opened 2026-03-11
fix(snippetz): preserve duplicate query params in code snippets

### LINKED ISSUE #8009: Array not properly represented in Code Snippet

AbdelhamidKhald:
Fixes #8009.

## Summary

Code snippets could collapse repeated query parameters to the last value for generators that convert query params into object-style `params` / `query` options. The request layer already sent repeated keys correctly, but the snippet output still misrepresented them.

## Changes

- preserve repeated query params as arrays in the shared query-object reducer
- use that reducer in the shared httpsnippet-lite conversion path and the `js/node ofetch` and Python requests-style generators
- update the `r/httr` target to emit repeated query entries when a query value is an array
- add focused regression coverage for duplicate query parameter output
- add a patch changeset for `@scalar/snippetz`

## Testing

- `pnpm exec biome check packages/snippetz/src/libs/http.ts packages/snippetz/src/libs/http.test.ts packages/snippetz/src/utils/convertWithHttpSnippetLite.ts packages/snippetz/src/utils/convertWithHttpSnippetLite.test.ts packages/snippetz/src/plugins/js/ofetch/ofetch.ts packages/snippetz/src/plugins/js/ofetch/ofetch.test.ts packages/snippetz/src/plugins/node/ofetch/ofetch.ts packages/snippetz/src/plugins/node/ofetch/ofetch.test.ts packages/snippetz/src/plugins/python/requestsLike.ts packages/snippetz/src/plugins/python/requests/requests.test.ts packages/snippetz/src/httpsnippet-lite/targets/r/httr/client.ts`
- `pnpm exec vitest run src/libs/http.test.ts src/utils/convertWithHttpSnippetLite.test.ts src/plugins/js/ofetch/ofetch.test.ts src/plugins/node/ofetch/ofetch.test.ts src/plugins/python/requests/requests.test.ts`

### DIFF
--- a/.changeset/green-plants-tell.md
+++ b/.changeset/green-plants-tell.md
@@ -2,4 +2,4 @@
 '@scalar/snippetz': patch
 ---
 
-Preserve duplicate query parameters in generated code snippets.
+Preserve duplicate query parameters in generated code snippets. php/guzzle: preserve duplicate query parameters as arrays.
--- a/packages/snippetz/src/plugins/php/guzzle/guzzle.test.ts
+++ b/packages/snippetz/src/plugins/php/guzzle/guzzle.test.ts
@@ -110,6 +110,33 @@ $response = $client->request('GET', 'https://example.com', [
 ]);`)
   })
 
+  it('preserves repeated query parameters as arrays', () => {
+    const result = phpGuzzle.generate({
+      url: 'https://example.com',
+      queryString: [
+        {
+          name: 'statuses',
+          value: 'active',
+        },
+        {
+          name: 'statuses',
+          value: 'inactive',
+        },
+      ],
+    })
+
+    expect(result).toBe(`$client = new GuzzleHttp\\Client();
+
+$response = $client->request('GET', 'https://example.com', [
+  'query' => [
+    'statuses' => [
+      'active',
+      'inactive'
+    ]
+  ]
+]);`)
+  })
+
   it('has cookies', () => {
     const result = phpGuzzle.generate({
       url: 'https://example.com',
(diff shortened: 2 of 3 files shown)


hanspagel:
we just needed it in one more :)

AbdelhamidKhald:
Thanks, that makes sense. Appreciate you catching the extra path and pushing the follow-up.

PR #8406, opened 2026-03-11
feat(api-reference): x-order

### LINKED ISSUE #7600: x-order support for properties
```yaml


Varun Shoor11min ago
Re: Cannot publish docs
to:Hans at Scalar, from:Varun Shoor
One more issue :/

Seems like you do follow `x-order` extension BUT, your sorting for `x-order` is as a string and not number. So `object` which has an order of `1` and `objects` which has an order of `10` are clubbed together.



  EPP:
    description: EPP server information aggregating greeting data, client configuration,
      and pool statistics.
    properties:
      client_state:
        description: Current client connection state
        example: connected
        type: string
        x-order: "6"
      dcp:
        allOf:
        - $ref: '#/definitions/DCP'
        description: Data collection policy
        x-order: "12"
```

hanspagel:
## Problem

* By default, we sort properties alphabetically.
* There's a configuration option to preserve their order when rendering them.
* But if you want to explicitly control the order, you're lost.

Or as a user wrote:

> The problem is Swaggo, the most popular automated OpenAPI annotation lib for Golang uses maps for annotations. Maps in Golang are unordered, so all the order in annotations is lost when the spec is generated. It makes reading resources really difficult.

## Solution

This PR adds support for the `x-order` specification extension to put properties in a certain order (this is a thing that others have already).

Fixes #7600

## Checklist

- [x] I added a changeset. 
- [x] I added tests.

---

> [!NOTE]
> **Low Risk**
> Low risk UI-only change that adjusts schema property display ordering; main risk is unintended ordering changes in existing specs that already include `x-order` or rely on previous sort behavior.
> 
> **Overview**
> Adds support for the `x-order` OpenAPI specification extension to control the display order of object schema properties in `@scalar/api-reference`.
> 
> Updates `sortPropertyNames` so properties with numeric `x-order` sort first (ascending), while still keeping discriminators first and falling back to existing required/alphabetical behavior for properties without `x-order`; adds unit tests covering precedence and fallback cases, updates docs, and includes a patch changeset.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c35de61fc344f8c228ed3d8933390a6395c3fc2c. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/.changeset/beige-meteors-speak.md
+++ b/.changeset/beige-meteors-speak.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
+feat: x-order


hanspagel:
sorry, I’m just too eager to get this in, too
and I think it's low risk

PR #8414, opened 2026-03-12
fix(api-reference): limit height of example response descriptions

hwkr:
Fixes some issues with long response descriptions escaping the response view. Limits the height on the right to 3 lines before scrolling.

Before / After

 
CleanShot 2026-03-11 at 22 40 31@2x
CleanShot 2026-03-11 at 22 45 51@2x

## Checklist

- [x] I added a changeset. 
- [x] I added tests.


hanspagel (approved):
thank you picking this one up 👏

PR #8419, opened 2026-03-12
feat(aspire): add optional endpointName to WithApiReferenceFeat/endpoint name

paulomorgado:
## Problem

When a resource has more that one endpoint, it's not guaranteed that the first one is the one exposing the API.

## Solution

 Adds an optional `endpointName` parameter to all `WithApiReference`
 overloads. When provided, only the endpoint with that name is registered
 for Aspire service discovery, injecting a single
 `services__{resourceName}__{scheme}__{index}` environment variable
 instead of one per endpoint.

 Uses `resourceBuilder.GetEndpoint(endpointName)` internally rather than
 `WithReference(resource, alias)`, which in Aspire 13.1.0 would rename
 the resource in discovery keys rather than filter endpoints.

 Backward-compatible `Func<ScalarOptions, CancellationToken, Task>`
 overloads (without `endpointName`) are preserved as delegates to the new
 core overloads.

 Adds three tests to `ScalarResourceConfiguratorTests` covering:
 - no `endpointName` → all endpoints exposed
 - `endpointName: "https"` → only the https endpoint exposed
 - file overload with `endpointName` → same filtering behavior

## Checklist

- [X] I added a changeset. 
- [X] I added tests.
- [X] I updated the documentation.

### DIFF
--- a/.changeset/icy-ibis-talk.md
+++ b/.changeset/icy-ibis-talk.md
@@ -1,0 +1,5 @@
+---
+'@scalar/aspire': patch
+---
+
+feat(aspire): add optional endpointName to WithApiReference
--- a/documentation/integrations/aspire.md
+++ b/documentation/integrations/aspire.md
@@ -152,6 +152,27 @@ scalar.WithApiReference(
     });
 ```
 
+### Scoping Service Discovery Endpoints
+
+By default, `WithApiReference` exposes **all** endpoints of a service to Aspire service discovery. This injects one `services__{resourceName}__{scheme}__{index}` environment variable per endpoint into the Scalar container (e.g., `services__book-service__http__0` and `services__book-service__https__0`).
+
+Use the optional `endpointName` parameter to restrict discovery to a single named endpoint:
+
+```csharp
+// Only expose the "https" endpoint for service discovery
+scalar.WithApiReference(bookService, endpointName: "https");
+```
+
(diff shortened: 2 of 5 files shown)


hanspagel (approved):
nice

devtekve:
FYI, this broke my existing code: 

```cs
scalar.WithApiReference(gatewayService, options =>
{
    options.WithOpenApiRoutePattern("/identity/openapi/{documentName}.json");
    
    options
        .WithTheme(ScalarTheme.Mars)
        .WithDefaultHttpClient(ScalarTarget.CSharp, ScalarClient.HttpClient);
});
```
image
image

## Solution
Forcing me to change it to 
```cs
scalar.WithApiReference(gatewayService, configureOptions: options =>
{
    options.WithOpenApiRoutePattern("/identity/openapi/{documentName}.json");
    
    options
        .WithTheme(ScalarTheme.Mars)
        .WithDefaultHttpClient(ScalarTarget.CSharp, ScalarClient.HttpClient);
});
```

Explicitly defining options as named parameter `configureOptions` otherwise it picks the wrong overload. 

***
It's not a big deal but I am mentioning in case somebody else encounters this when updating

paulomorgado:
@devtekve, that's common to many Aspire APIs, because they have a lot of optional arguments that can change from version to version.

PR #8421, opened 2026-03-12
test(api-reference): use masks instead of element snapshots

hwkr:
We've have some flakiness issues with our element snapshots especially in scroll containers. This updates Playwright it to take full viewport snapshots and using masks to only capture the part of the page we want to be testing.

## Checklist

- [x] I added tests.


hanspagel:
woahh, that's smart

PR #8442, opened 2026-03-13
fix(api-reference): show object schema descriptions

### LINKED ISSUE #7017: Description is not displayed when schema type is object
### What happens?

Scalar is not able to render the description of a query parameter if the schema is of type object. Here is the bug reproduced in a sandbox: https://sandbox.scalar.com/e/zgCeR

Ideally in this case, the description, and the whole parameter schema is rendered correctly. At the moment, the description is not rendered, and there is a UNKNOWN. 

### OpenAPI Document

https://sandbox.scalar.com/e/zgCeR

AbdelhamidKhald:
Fixes #7017.

## Summary

Object schema descriptions could disappear in parameter rendering because the schema property description helper dropped descriptions whenever the schema had `properties` or `additionalProperties`. That left object parameters showing the type without the authored description.

## Changes

- allow object schema descriptions to render in `SchemaProperty`
- keep composition-based descriptions suppressed to avoid duplicate composition output
- strip the nested object schema description when the parent property already renders it, so expanded object properties do not duplicate the same text
- add focused regression coverage for the helper logic and the expanded object-schema rendering path

## Testing

- `pnpm exec biome check packages/api-reference/src/components/Content/Schema/helpers/should-display-description.ts packages/api-reference/src/components/Content/Schema/helpers/should-display-description.test.ts packages/api-reference/src/components/Content/Schema/SchemaProperty.vue packages/api-reference/src/components/Content/Schema/SchemaProperty.test.ts`
- `pnpm exec vitest run packages/api-reference/src/components/Content/Schema/helpers/should-display-description.test.ts packages/api-reference/src/components/Content/Schema/SchemaProperty.test.ts`


hanspagel:
need to update the snapshots

Issue #4542, opened 2025-01-22
Scalar API Client populates DELETE request body by default, even when OpenAPI JSON schema explicitly forbids it


Greenheart:
### Discussed in https://github.com/scalar/scalar/discussions/4541

<div type='discussions-op-text'>

<sup>Originally posted by **Greenheart** January 22, 2025</sup>
Hi - first of all, thanks for this project!

## Question
How can I make the DELETE request body empty by default, to respect the OpenAPI JSON schema? Right now, it's always set to an empty JSON object by default for DELETE requests, despite the OpenAPI JSON schema not allowing any request body.

I've explored this further, and here are my notes:

## Potential bug
The underlying issue seems to be caused by a potential bug in the Scalar API Client which defaults to `{ "Content-Type": "application/json" }` for all request bodies, even for DELETE requests.

This is the PR that introduced JSON bodies for DELETE requests, however it does not take into account if the OpenAPI schema actually allows any request body. So technically it is true that DELETE requests *can* have a body, but it seems to be more common that they don't have a body. I think a more sensible default behaviour would be to only include the body if the schema eexplicitly allows/expects it.
https://github.com/scalar/scalar/pull/1877/files

The current behaviour makes it confusing to send DELETE requests from Scalar because they automatically gets the `Content-Type: application/json` header and an empty `{}` or `` JSON request body (depending on the schema). And this happens even when the OpenAPI JSON Schema disallow DELETE requests with a body.

## Expected behaviour

The fix would be to set the default `Body` to `None` unless the OpenAPI docs expect a valid JSON body. Also, given that DELETE requests should not include a body (but could), this might be worth.

### Current invalid default state:
**NOTE:** that this also shows up with `{}` as the JSON body in depending on how I change the OpenAPI schema
![image](https://github.com/user-attachments/assets/635532b3-ab45-4239-a9d7-7ccfa1a12390)

### Expected default state:
![image](https://github.com/user-attachments/assets/6aadb7d0-4354-42ec-a8e0-9d866f6eefc0)

## Potential solutions:
Add better tests here:
https://github.com/scalar/scalar/blob/a0f953fa979502fadd29315ba814b9039825beba/packages/api-client/src/libs/send-request/create-fetch-body.test.ts

Here's what determines which requests *can* have a body. This is not necessarily the same as *should* have a body though:
[...]

Issue #3142, opened 2024-09-19
Example request adds "content-type: application/json" header in GET request


subzero10:
### What happens?

I have two endpoints which are almost identical in the openapi.json file (GET /api/client and GET /api/account).
GET /api/account works as expected, but GET /api/client is causing my backend to throw an exception because the request is sent with "Content-Type: application/json" and my backend is trying to deserialize the non-existent body the request.

I can't share the full openapi.json file, but here's an excerpt of the two endpoints:
```
"/api/client": {
  "get": {
    "tags": [
      "Client"
    ],
  "summary": "Get clients",
  "operationId": "ApiEndpointsClientIndex",
  "parameters": [...],
  "responses": {...}
  },
},
"/api/account": {
  "get": {
    "tags": [
      "Account"
    ],
    "summary": "Get accounts",
    "operationId": "ApiEndpointsAccountIndex",
    "parameters": [...],
    "responses": {...}
  },
}
```

When testing an example request for GET /api/account, it does not add the header and I get a valid response.
![CleanShot 2024-09-19 at 10 30 36@2x](https://github.com/user-attachments/assets/151faae7-7de7-4120-bb7b-2a9046d016f1)

When testing an example request for GET /api/client, the Content-Type: application/json header is added which which causes the exception on my backend.
![CleanShot 2024-09-19 at 10 31 14@2x](https://github.com/user-attachments/assets/793c4e19-1138-43fe-a380-24e96c4813e9)

### What did you expect to happen?

I would expect to either both requests to have the included Content-Type header by default, or none of them to have it, since they have almost identical definitions in the json file.

amritk:
Hello! If you update the client, you will find that we no longer set any content type and leave it up to the user!

subzero10:
Hey @amritk, unfortunately I can't use the latest version of the client - it's crashing the UI. I've reported the issue here #3165.

subzero10:
@amritk, I just tested this with 1.25.17 and it seems that the issue is resolved.
Thanks!

jonasfroeller:
This still happens to me using `v1.25.58`.

wegylexy:
How to eliminate the `Content-Type` header by default? The `Test Request` dialog says `This is a readonly property and you can not modify it! If you want to change it you have to override it or disable it using the checkbox` but I don't want to click the checkbox every time. I am at `@scalar/fastify-api-reference:1.48.0`.

Issue #2486, opened 2024-07-15
Content-Type header with empty body


inyourtime:
### What happens?

My API in Swagger does not have a body, as shown in the example below:
```
{
  "openapi": "3.0.3",
  "info": { "title": "My Fastify App", "version": "1.0.0" },
  "components": {
    "securitySchemes": {
      "apiKey": { "type": "apiKey", "name": "apiKey", "in": "header" }
    },
    "schemas": {}
  },
  "paths": {
    "/example-route/{id}": {
      "put": {
        "summary": "qwerty",
        "tags": ["user", "code"],
        "description": "post some data",
        "parameters": [
          {
            "schema": { "type": "string" },
            "in": "path",
            "name": "id",
            "required": true,
            "description": "user id"
          }
        ],
        "responses": {
          "201": {
            "description": "Successful response",
            "content": {
              "application/json": {
                "schema": {
                  "description": "Successful response",
                  "type": "object",
                  "properties": { "hello": { "type": "string" } }
                }
              }
            }
          }
        }
      }
    }
  }
}
```

So, the body of the example request is set to none. However, when I change it to JSON, the request header is not set to application/json.

<img width="662" alt="Screenshot 2567-07-15 at 11 39 33" src="https://github.com/user-attachments/assets/018ea9c4-d291-4155-9ec2-57887f0b8c35">
<img width="562" alt="Screenshot 2567-07-15 at 11 38 57" src="https://github.com/user-attachments/assets/b4d2abf2-0a56-4615-9a1d-3a7a22ce623e">

### What did you expect to happen?

It should set the header correctly.

### How can we reproduce the issue?

1. Add a Swagger file with an example request that does not have a body.
2. Change the body to JSON.
3. Send the request.

marclave:
fixed in https://github.com/scalar/scalar/pull/2526 : O

PR #8459, opened 2026-03-16
fix(api-client): do not add Content-Type header when request has no body

### LINKED ISSUE #8458: Content-Type is set to `application/json` for requests that does not have request body
### What happens?

Hi, we are using `"@scalar/fastify-api-reference": "^1.48.7"` and we are running into this issue.

When a request does not have a request body, Scalar adds `Content-Type: application/json` header to the request. With Fastify this results in `FST_ERR_CTP_EMPTY_JSON_BODY` error, requires unticking the header manually for a successful request. The example is DELETE method, but this also happens with other requests

Example OpenAPI JSON:

```json
{
  "openapi": "3.0.3",
  "info": {
    "title": "API Reference",
    "version": "0.0.0"
  },
  "components": {"schemas": {}},
  "paths": {
    "/": {
      "delete": {
        "operationId": "deleteTest",
        "summary": "Test",
        "tags": [
          "Test"
        ],
        "description": "Test endpoint",
        "security": [],
        "responses": {
          "200": {
            "description": "Delete Success",
            "content": {
              "application/json": {
                "schema": {
                  "type": "object",
                  "properties": {
                    "status": {
                      "type": "string",
                      "description": "Status"
                    }
                  },
                  "required": [
                    "status"
                  ],
                  "additionalProperties": false,
                  "description": "Delete Success"
                }
              }
            }
          }
        }
      }
    }
  }
}
```
Current UI:

Image

I think 'Content-Type' header should not be added by default if there is no body in the request, respecting the schema.

Related issues: #3142 #2486 #4542

### OpenAPI Document

_No response_

brick-pixel:
## What

Added a `requestBody` existence check before adding the `Content-Type` header to outgoing requests.

## Why

When an operation does not define a `requestBody` (e.g. a simple `DELETE` endpoint), the API client still adds `Content-Type: application/json` to the request. This causes servers like Fastify to reject the request with `FST_ERR_CTP_EMPTY_JSON_BODY`.

The existing check only verified whether the HTTP method *can* have a body (`canMethodHaveBody`), but not whether the operation actually *defines* one.

Reported in #8458. Previously reported in #3142, #2486, #4542.

## Changes

**File:** `packages/api-client/src/v2/blocks/request-block/helpers/get-default-headers.ts`

```diff
-  if (canMethodHaveBody(method)) {
+  if (canMethodHaveBody(method) && requestBody) {
```

`requestBody` is already resolved from the operation on the line above, so this is a minimal, safe addition.

## Behavior

| Method | requestBody defined | Content-Type added |
|--------|--------------------|--------------------|
| POST | Yes | ✅ Yes (unchanged) |
| PUT | Yes | ✅ Yes (unchanged) |
| DELETE | Yes | ✅ Yes (unchanged) |
| DELETE | **No** | ✅ **No (fixed)** |
| GET | N/A | ✅ No (unchanged) |

Fixes #8458

### DIFF
--- a/packages/api-client/src/v2/blocks/request-block/helpers/get-default-headers.ts
+++ b/packages/api-client/src/v2/blocks/request-block/helpers/get-default-headers.ts
@@ -82,7 +82,8 @@ export const getDefaultHeaders = ({
   const requestBody = getResolvedRef(operation.requestBody)
 
   // Add Content-Type header only for methods that support a request body
-  if (canMethodHaveBody(method)) {
+  // and only when the operation actually defines a request body
+  if (canMethodHaveBody(method) && requestBody) {
     const contentType =
       requestBody?.['x-scalar-selected-content-type']?.[exampleKey] ??
       Object.keys(requestBody?.content ?? {})[0] ??


hanspagel:
Thanks! Closing in favor of #8470

Issue #6572, opened 2025-08-15
can’t search in large response body


sabvente-gti:
### What happens?

In the test request dialog, after a successful request, i get like a 3000 row json response.
In the response "code editor" only 27 rows visible at a time on my screen.

The browser search function is broken, it only searches in the 27 currently visible lines. There is not other search function either.

### What did you expect to happen?

Search should work in Scalar. When I'm testing an API with different parameters, I'd like to search in the result.

PR #8462, opened 2026-03-16
chore(api-client): replace codemirror with monaco editor for response body

DemonHa:
Fixes: #6572 

We are currently using CodeMirror to render the response body. However, we also attempt to implement our own virtualization for large responses, which is not working reliably at the moment.

This change replaces CodeMirror with Monaco Editor, which has built-in virtualization and handles large documents more efficiently. By relying on Monaco’s internal mechanisms, we can simplify our implementation while significantly improving performance for large response bodies.

## Checklist

- [x] I added a changeset. 
- [ ] I added tests.


hanspagel:
I don’t think we can ship Monaco with the api-reference

* it’s huge
* web workers are tough to set up (our main distribution way is a single JS umd build)

so we need to make this a plugin or something, so we can stick to codemirror for the modal

hanspagel:
me don't like stale PRs, I close it until we figure out how to properly ship this to users or wrap this in a plugin

Issue #6757, opened 2025-09-05
`anyOf` (union array) is not rendered correctly in the example section


matzeeable:
### What happens?

I try to migrate from the Swagger UI to Scalar. I really love how Scalar displays and handles OpenAPI documents. Thanks for this! 

I have the following schema:

```yml
openapi: 3.1.0
info:
  title: Scalar Galaxy
  version: 0.5.7
externalDocs:
  description: Documentation
  url: https://github.com/scalar/scalar
servers:
  - url: https://galaxy.scalar.com
paths:
  /planets:
    get:
      tags:
        - Planets
      summary: Get all planets
      responses:
        "200":
          description: OK
          content:
            application/json:
              schema:
                type: array
                items:
                  anyOf:
                    - type: object
                      properties:
                        code:
                          type: string
                          enum:
                            - UserSessionMagicLinkTokenNotFound
                        message:
                          type: string
                        description:
                          type: string
                      required:
                        - code
                        - message
                      description: tests
                    - type: object
                      properties:
                        code:
                          type: string
                          enum:
                            - UserSessionMagicLinkTokenExpired
                        message:
                          type: string
                        description:
                          type: string
                        expiredAt:
                          type: string
                          format: date
                      required:
                        - code
                        - message
                        - expiredAt
                    - type: object
                      properties:
                        code:
                          type: string
                        message:
                          type: string
                        description:
                          type: string
                      required:
                        - code
                        - message
```

We are using `anyOf` for an array of unions.

When rendering the OpenAPI schema in the Swagger UI Editor (https://editor.swagger.io/), we are getting a preview / example of all available error codes:

<img width="1624" height="1076" alt="Image" src="https://github.com/user-attachments/assets/da6a68f2-ff8a-4bfa-9802-7f84888aed37" />

### What did you expect to happen?

When using the same OpenAPI schema in Scalar we do only see one possible union type in the example section:

<img width="1616" height="1618" alt="Image" src="https://github.com/user-attachments/assets/ca68835b-7894-4835-8314-c4ebb9deba04" />

### OpenAPI Document

https://sandbox.scalar.com/e/p6msQ

amritk:
hello @matzeeable. All of the options are there, we just have a dropdown to select them
<img width="466" height="282" alt="Image" src="https://github.com/user-attachments/assets/b761b04b-7307-4985-88c3-e22ed1063db0" />

They are all called object because they don't have a title. 

cc: @cameronrohani for any design feedback.

[...]

Issue #5965, opened 2025-06-15
cURL preview does not update when switching between anyOf request bodies


alonfaraj:
### What happens?

When defining an OpenAPI requestBody schema using anyOf, the Scalar UI shows a dropdown to toggle between schema types. However, the cURL preview panel does not update when switching between types — it remains stuck on the first one.

```
openapi: 3.1.0
info:
  title: Scalar Curl Preview Bug
  version: 1.0.0
paths:
  /example:
    post:
      summary: Example endpoint
      requestBody:
        required: true
        content:
          application/json:
            schema:
              anyOf:
                - $ref: '#/components/schemas/TypeA'
                - $ref: '#/components/schemas/TypeB'
      responses:
        '200':
          description: OK
components:
  schemas:
    TypeA:
      type: object
      properties:
        type:
          type: string
          const: A
        message:
          type: string
    TypeB:
      type: object
      properties:
        type:
          type: string
          const: B
        value:
          type: integer
```

TypeA shown correctly
![Image](https://github.com/user-attachments/assets/a0c009aa-66a8-486f-9979-fbc1254e2308)

TypeB doesn't shown correctly
![Image](https://github.com/user-attachments/assets/5faed597-c3fb-434b-bd44-3e557feff0ea)

### What did you expect to happen?

The cURL preview should update to reflect the currently selected schema.

OnDroo:
Hello, 

simillar/same issue here. When I want to use examples it does not update the preview nor it fills in the body

Error while updating the curl:
![Image](https://github.com/user-attachments/assets/95142174-3a1c-48ac-a076-ad027839e89e)

Errors while loading the body:
![Image](https://github.com/user-attachments/assets/cc6b88bd-8a03-4c96-a109-612b129418aa)

![Image](https://github.com/user-attachments/assets/52481ee7-c28a-401c-b66c-a111baa41573)

openapi.json:
```json
{
  "openapi": "3.0.3",
  "info": {
    "title": "My Authentication API",
    "version": "1.0.0"
  },
  "paths": {
    "/connect/token": {
      "post": {
        "tags": [
          "Authentication"
        ],
        "summary": "Request an Access Token",
        "description": "Exchanges user credentials, a refresh token, or other grants for an access token. The request body format is `application/x-www-form-urlencoded`.",
        "requestBody": {
          "description": "The grant type and corresponding credentials for the token request.",
          "required": true,
          "content": {
            "application/x-www-form-urlencoded": {
              "schema": {
[...]

PR #8473, opened 2026-03-17
fix(api-client): anyOf code example sync

alonfaraj:
## Problem

Changing the request body schema dropdown for `anyOf` or `oneOf` variants in the API reference did not update the generated request snippet.

Fix https://github.com/scalar/scalar/issues/5965
Fix https://github.com/scalar/scalar/issues/6757

## Solution

In `@scalar/api-client`, I added optional request-body composition index support so example generation can use the selected `anyOf` or `oneOf` variant instead of always defaulting to the first one.

In `@scalar/api-reference`, I wired the request body schema dropdown selection to the code sample so both stay in sync. I also replaced the string-based provide/inject key with a shared typed symbol and a defaulted `inject` to avoid missing-injection warnings outside the operation layouts.

I added coverage in two places:
- schema example generation tests for choosing a non-default composition branch
- a layout-level integration test that mounts a real operation layout, changes the schema dropdown, and verifies the rendered code sample updates

## Checklist

- [x] I added a changeset. 
- [x] I added tests.

## Example File
[openapi.json](https://github.com/user-attachments/files/26083319/openapi.json)

### DIFF
--- a/packages/api-client/src/v2/blocks/operation-code-sample/helpers/operation-to-har/process-body.ts
+++ b/packages/api-client/src/v2/blocks/operation-code-sample/helpers/operation-to-har/process-body.ts
@@ -1,7 +1,11 @@
 import { json2xml } from '@scalar/helpers/file/json2xml'
 import { getResolvedRef } from '@scalar/workspace-store/helpers/get-resolved-ref'
 import { unpackProxyObject } from '@scalar/workspace-store/helpers/unpack-proxy'
-import type { MediaTypeObject, RequestBodyObject, SchemaObject } from '@scalar/workspace-store/schemas/v3.1/strict/openapi-document'
+import type {
+  MediaTypeObject,
+  RequestBodyObject,
+  SchemaObject,
+} from '@scalar/workspace-store/schemas/v3.1/strict/openapi-document'
 import { getExampleFromSchema } from '@v2/blocks/operation-code-sample/helpers/get-example-from-schema'
 import { getResolvedRefDeep } from '@v2/blocks/operation-code-sample/helpers/get-resolved-ref-deep'
 import type { Param, PostData } from 'har-format'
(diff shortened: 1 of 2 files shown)


hanspagel:
pls review @DemonHa

packages/api-client/src/v2/blocks/operation-code-sample/helpers/get-example-from-schema.ts
```diff
@@ -627,10 +628,20 @@ export const getExampleFromSchema = (
   // Handle composition schemas (oneOf, anyOf) at root level
   const discriminate = _schema.oneOf || _schema.anyOf
   if (Array.isArray(discriminate) && discriminate.length > 0) {
-    // Find the first non-null type without allocating intermediate arrays
```
DemonHa, line 630: I think we need to handle nested compositions as well

packages/api-client/src/v2/blocks/operation-code-sample/helpers/operation-to-har/process-body.ts
```diff
@@ -117,10 +122,14 @@ export const processBody = ({ requestBody, contentType, example }: ProcessBodyPr
   // Try to extract examples from the schema
   const contentSchema = getResolvedRef(requestBody.content[_contentType]?.schema)
   if (typeof contentSchema !== 'undefined') {
-    const extractedExample = getExampleFromSchema(contentSchema, {
-      mode: 'write',
-      xml: isXml,
-    })
+    const extractedExample = getExampleFromSchema(
```
DemonHa, line 148: we can probably add the compositionIndex as part of the `getExampleFromSchema` options

alonfaraj:
Hey @DemonHa, 
I addressed the points you mentioned, also updated the PR with `openapi.json` example with both simple and nested. 
Let me know what you think

alonfaraj:
@hanspagel @DemonHa 
Seems like the `Test Request` Modal didn't reflect the above change, fixed now, including nested compositions.

alonfaraj:
@hanspagel rebased

PR #8483, opened 2026-03-17
feat(python): new package

hanspagel:
## Problem

the python packages are constantly outdated, lacking configuration options that we introduce

## Solution

* one new python package
* that's used by fastapi and django-ninja (and more soon)
* the configuration options are auto-generated
* restructure the docs for all python integrations under one sidebar entry (like we did for rust)

See #8460

## Checklist

- [x] I added a changeset. 
- [x] I added tests.


hanspagel:
good, but needs more time, I don't have time for it yet, will pick it up again

Issue #8492, opened 2026-03-17
Query parameters with special characters not handled correctly


AX-LLaidley:
### What happens?

When I use Test Request to test a request with a sort parameter, like `?sort=name:asc` I get a 400 error: 

```
{
  "status": 400,
  "code": "bad-request",
  "title": "Validation Error",
  "message": "One or more validation errors were encountered while processing this request.",
  "stackTrace": [],
  "errors": [
    {
      "field": "sort",
      "errors": [
        "'sort=name%3Aasc' is not a valid sort criteria. Valid criteria example: sort=name:desc."
      ]
    }
  ]
}
```

Our backend is expecting to receive the query param as `?sort=name:asc` and not the URL encoded character in place of the `:` in the string.

This is an issue both in the latest CDN version of Scalar, as well as 1.0.0 for macOS.

### What did you expect to happen?

I expected to receive a 200 response with the appropriate response body.

### OpenAPI Document

https://sandbox.scalar.com/e/wzoIl

hanspagel:
I think this is what `allowReserved: true` is for: 

https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.1.md#fixed-fields-for-use-with-schema

https://github.com/scalar/scalar/pull/8501 added support for it

PR #8501, opened 2026-03-18
fix(api-client): respect explicit parameter allowReserved false

hanspagel:
## Problem

Fixes part of #8492.

`isAllowReserved` treated `allowReserved: false` on a parameter the same as if `allowReserved` were absent. That allowed the non-standard schema-level fallback (`schema.allowReserved`) to override an explicit parameter-level `false`.

## Solution

- Updated `isAllowReserved` so parameter-level `allowReserved` takes precedence whenever it is explicitly defined (`true` or `false`).
- Kept schema-level `allowReserved` as a fallback only when parameter-level `allowReserved` is absent/undefined.
- Added a regression test for `allowReserved: false` on parameter with `schema.allowReserved: true` to verify the parameter-level `false` wins.
- Fixed CI typecheck by casting the two non-standard `schema.allowReserved` test fixtures to `ParameterObject` in the test.
- Removed stale `@scalar/oas-utils` bump from the changeset because this PR no longer changes that package.

## Checklist

- [x] I added a changeset. 
- [x] I added tests.

<div><a href="https://cursor.com/agents/bc-604857c7-7b0d-4869-8568-caa66e750fb3"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-604857c7-7b0d-4869-8568-caa66e750fb3"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/.changeset/fresh-hounds-encode.md
+++ b/.changeset/fresh-hounds-encode.md
@@ -1,6 +1,5 @@
 ---
 '@scalar/api-client': patch
-'@scalar/oas-utils': patch
 '@scalar/workspace-store': patch
 ---
 
--- a/packages/api-client/src/v2/blocks/operation-block/helpers/build-request-parameters.test.ts
+++ b/packages/api-client/src/v2/blocks/operation-block/helpers/build-request-parameters.test.ts
@@ -311,11 +311,12 @@ describe('buildRequestParameters', () => {
             name: 'sort',
             in: 'query',
             required: true,
+            // Legacy/non-standard fallback: schema-level allowReserved.
             schema: { type: 'string', allowReserved: true },
             examples: {
               default: { value: 'run_time:desc', 'x-disabled': false },
             },
-          },
+          } as unknown as ParameterObject,
         ]
 
         const result = buildRequestParameters(params)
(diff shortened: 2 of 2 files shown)


packages/api-client/src/v2/blocks/operation-block/helpers/apply-allow-reserved-to-url.test.ts
DemonHa, line 1: We don't need to do any changes on the legacy client anymore since we are only using the new client now

hanspagel: cursor did, human me cleaned it up 👍

PR #8524, opened 2026-03-20
refactor(workspace-store, api-client): move request building and context logic into workspace store and refactor for consistency

DemonHa:
Previously, request construction was handled inside the `api-client`, which introduced friction when attempting to send requests outside of it.

This PR extracts all request-building logic along with the request example context into the `workspace-store`, enabling requests to be constructed from an operation example independently of the client.

Additionally, this PR introduces a request factory responsible for generating browser-ready requests, while allowing consumers to modify the request before sending without unnecessary constraints.

> Note: I change the example extracting logic, it looks like before we were fallback to the depricated example field or generating a random example from the schema if the example was not found which should be wrong (cc @amritk)

## Checklist

- [ ] I added a changeset. 
- [x] I added tests.

---

> [!NOTE]
> **Medium Risk**
> Medium risk because it rewires core request construction/execution paths (URL resolution, auth placement, cookies, hooks) across multiple packages, which could subtly change outgoing requests and plugin behavior.
> 
> **Overview**
> **Request construction is migrated to `@scalar/workspace-store/request-example`.** Consumers now obtain request context via `getRequestExampleContext`, build a request builder with `requestFactory`, and finalize it with `buildRequest` (including env var expansion), replacing `api-client`’s internal build/security/url helpers.
> 
> **`api-client`’s `OperationBlock` execution flow is refactored accordingly.** It now accepts explicit `selectedSecurity`/`selectedSecuritySchemes`, `securityRequirements`, `defaultHeaders`, and separate `workspaceCookies`/`documentCookies`; request execution runs plugin hooks via `executeHook` (`beforeRequest`, `responseReceived`) and tightens error handling/cleanup.
> 
> **`agent-chat` and playground code are updated to the new APIs.** Auth/server/environment helpers are imported from `workspace-store`, request security is applied by translating security options into headers/query/cookies (including Base64 for basic auth), and package/knip exports are adjusted (including adding `js-base64` and removing `api-client`’s `./v2/helpers` export).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c971c2d7560ab339826970700d4a912d653d2dac. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/api-client/src/v2/blocks/operation-block/OperationBlock.vue
+++ b/packages/api-client/src/v2/blocks/operation-block/OperationBlock.vue
@@ -209,15 +209,6 @@ const handleExecute = async () => {
     response.value.reader.cancel()
   }
 
-  // Execute the hooks
-  eventBus.emit('hooks:on:request:sent', {
-    meta: {
-      method,
-      path,
-      exampleKey,
-    },
-  })
-
   // Build the actual request we will send
   const requestResult = (() => {
     try {
@@ -243,6 +234,15 @@ const handleExecute = async () => {
 
   // Store the abort controller for cancellation
   abortController.value = requestResult.result.controller
+
+  // Execute the hooks
+  eventBus.emit('hooks:on:request:sent', {
+    meta: {
+      method,
+      path,
+      exampleKey,
+    },
+  })
 
   // Execute the beforeRequest hook
   const { request: finalRequest } = await executeHook(


geoffgscott:
> Note: I change the example extracting logic, it looks like before we were fallback to the depricated example field or generating a random example from the schema if the example was not found which should be wrong (cc @amritk)

Maybe we want to add a separate function to generate random data for request bodies.

packages/api-client/src/v2/blocks/operation-block/helpers/send-request.ts
```diff
@@ -73,12 +71,12 @@ export const sendRequest = async ({
   }>
 > => {
   try {
-    // Apply any beforeRequest hooks from the plugins
-    const { request: modifiedRequest } = await executeHook({ request }, 'beforeRequest', plugins)
+    // // Apply any beforeRequest hooks from the plugins
+    // const { request: modifiedRequest } = await executeHook({ request }, 'beforeRequest', plugins)
 
     // Execute the request and measure duration
     const startTime = Date.now()
```
geoffgscott: we probably want to use performance.now() here.

packages/api-client/src/v2/blocks/operation-block/OperationBlock.vue
```diff
@@ -196,27 +197,25 @@ const handleExecute = async () => {
     return
   }
 
-  const [error, result] = buildRequest({
+  const requestBulder = requestFactory({
```
geoffgscott:
```suggestion
  const requestBuilder = requestFactory({
```

packages/api-client/src/v2/features/operation/Operation.vue
```diff
@@ -47,117 +43,77 @@ const {
...
+  return [
+    ...(ctx.cookies.workspace.map((it) => ({
+      ...it,
+      location: 'workspace',
+    })) satisfies ExtendedScalarCookie[]),
+    ...(ctx.cookies.document.map((it) => ({
+      ...it,
+      location: 'document',
+    })) satisfies ExtendedScalarCookie[]),
+  ]
```
geoffgscott: I think We could clean this up. The complex destructuring is hard to read. Probably better to just declare a couple local vars and then destructure it all together at the end.

packages/workspace-store/src/request-example/builder/body/get-request-body-example.ts
```diff
@@ -0,0 +1,15 @@
+import { getResolvedRef } from '@scalar/workspace-store/helpers/get-resolved-ref'
+import type { RequestBodyObject } from '@scalar/workspace-store/schemas/v3.1/strict/openapi-document'
+
+/**
+ * Basically getExample + we generate an example from the schema if no example is found
```
geoffgscott, line 13: We should just expand this comment to clarify why we need a function for this at all (extracting from content type and nullish coercion)

packages/workspace-store/src/request-example/builder/header/serialize-parameter.ts
geoffgscott, line 1: Love this file.

packages/workspace-store/src/request-example/builder/helpers/get-environment-variables.ts
```diff
@@ -0,0 +1,17 @@
...
+ * @returns An object mapping variable names to their resolved values.
+ */
+export const getEnvironmentVariables = (environment: XScalarEnvironment) => {
+  return environment.variables.reduce(
+    (acc, curr) => {
+      acc[curr.name] = typeof curr.value === 'string' ? curr.value : curr.value.default
+      return acc
+    },
+    {} as Record<string, string>,
+  )
```
geoffgscott:
```suggestion
export const getEnvironmentVariables = (environment: XScalarEnvironment) => {
  return environment.variables.reduce<Record<string, string>>(
    (acc, curr) => {
      acc[curr.name] = typeof curr.value === 'string' ? curr.value : curr.value.default
      return acc
    },
    {},
  )
```

geoffgscott: Nitpick but this avoids a type cast

packages/api-client/src/v2/blocks/operation-block/OperationBlock.vue
```diff
@@ -236,19 +218,46 @@ const handleExecute = async () => {
     },
   })
 
+  // Execute the beforeRequest hook
+  await executeHook(
+    // @ts-expect-error - TODO: fix this update to use the new request factory
```
amritk: @DemonHa is this going to work?

DemonHa: not really, I will change it to keep it compatible with what we have, and we should probably do a breaking change after to pass only the factory request

DemonHa: for now let's keep the same behaviour

packages/api-client/src/v2/blocks/operation-block/OperationBlock.vue
```diff
@@ -236,19 +218,45 @@ const handleExecute = async () => {
     },
   })
 
+  // Build the actual request we will send
+  const requestResult = buildRequest(requestBuilder, {
+    envVariables: getEnvironmentVariables(environment),
+  })
```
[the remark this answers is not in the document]
marclave: @DemonHa this seems like a good catch

packages/workspace-store/src/request-example/builder/build-request.ts
```diff
@@ -0,0 +1,183 @@
...
+        replaceEnvVariables(value, options.envVariables),
+      )
+    }
+
+    // Replace the query params with the environment variables
+    for (const [key, value] of request.query.params.entries()) {
+      url.searchParams.set(
+        replaceEnvVariables(key, options.envVariables),
+        replaceEnvVariables(value, options.envVariables),
+      )
```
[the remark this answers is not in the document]
DemonHa: this is fine

PR #8526, opened 2026-03-20
feat(api-reference): improve filtering of client selector

wilhelmer:
Addresses #8525.

### DIFF
--- a/packages/api-client/src/v2/blocks/operation-code-sample/helpers/get-clients.test.ts
+++ b/packages/api-client/src/v2/blocks/operation-code-sample/helpers/get-clients.test.ts
@@ -51,6 +51,7 @@ describe('getClients', () => {
     // First group should be the custom code examples
     expect(result[0]).toEqual({
       label: 'Code Examples',
+      key: 'custom',
       options: [
         {
           id: 'custom/python',


hanspagel:
OMG this is bugging me since forever

I can fix the snapshots

wilhelmer:
Yes, thanks for taking care of the snapshots, I'm on macOS and according to the README, I would have to switch to OrbStack or whatever to use the docker image.

hanspagel:
I tried to update the snapshots, let's see. I leave this for @hwkr to review

hwkr (changes requested):
Thanks so much for the PR! Looks great, unfortunately since we use the combobox in lots of places we probably don't want to have this be the default behavior (if you have huge groups and multiple group labels match then you effectively would have no filtering).

To options I could see for doing this not as the default:
1. We add a prop to enable it like `filterWithGroupLabels` → Not the worst but would add a big logic fork to the inside of the component
2. Add a prop like `filterFn` that let's you pass in a custom filter function and do whatever you want → Might be a bit more complicated but would open things up to be even more flexible in the future and would make the filter functions nice and unit testable

I'm going to take a swing at #&#8203;2 right now and see if I can get it working, let me know if you have any thoughts / preferences :)

wilhelmer:
Thanks for your efforts, both approaches sound reasonable to me! I probably would've chosen 1) since 2) sounds more complex, but if you can pull it off, happy to see the results ✌️

hwkr:
Okay updated! Should be good now, ended up doing a bit more to the client selector to align the types and add the filtering. Most exciting → typing `js` now shows the Javascript options first! We can also extend filterClientsByQuery in the future to be even smart / with different data if we want.

Before / Afters

CleanShot 2026-03-20 at 14 53 06@2x
CleanShot 2026-03-20 at 14 52 52@2x

CleanShot 2026-03-20 at 14 52 38@2x
CleanShot 2026-03-20 at 14 52 20@2x

hwkr:
Also @wilhelmer I made quite a few changes, happy to move them to a different PR you'd prefer just let me know :)

wilhelmer:
Yeah, that has changed quite a bit, but only for the better, thank you 🚀 

Maybe you could elaborate a bit more on the changes here, for visibility?

As far as I can see, this adds a new `:filterFn` property to combo boxes. If it is not set, nothing changes and the search behavior is the same as before. If it is set, which is the case for `OperationCodeSample` and `ClientDropdown`, we can pass in a custom filter function, which can be anything, but here it is including the group labels, and also the `key` property. So users can search for both `JavaScript` (the group label) and `js` (the key).

Does that accurately summarize the changes?

hwkr:
Yeah @wilhelmer exactly!

PR #8537, opened 2026-03-20
feat: package cleanup and reduce bundle sizes

geoffgscott:
## Problem

Removes a few dependencies from workspace store to hopefully reduce docs bundle size. Moves to oxc minifier for standalone which is significantly faster. Cleans up the icon gen a bit. 

## Checklist

- [ ] I added a changeset. 
- [ ] I added tests.

---

> [!NOTE]
> **Medium Risk**
> Medium risk because it changes build/minification and packaging/exports, which can affect bundle output, tree-shaking, and downstream consumers; functional logic changes are limited but could surface as build or import regressions.
> 
> **Overview**
> Switches the `@scalar/api-reference` standalone build to `minify: 'oxc'` (removing custom terser settings) and updates published package contents to exclude generated bundle analysis HTML files.
> 
> Extends `@scalar/helpers` with a new `markdown/*` export and a dependency-free `getMarkdownHeadings` utility (with tests/benchmarks), marking the package as `sideEffects: false` and excluding `*.bench.ts` from build output.
> 
> Simplifies `@scalar/icons` by making the generated index use direct re-exports (including aliases) instead of assembling import/export blocks, and normalizes package metadata (`files`, `sideEffects`).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit bff599af3e1bafdeb199da7971e47261f3ce929c. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/helpers/tsconfig.build.json
+++ b/packages/helpers/tsconfig.build.json
@@ -1,7 +1,7 @@
 {
   "extends": "./tsconfig.json",
   "include": ["src"],
-  "exclude": ["**/*.test.ts"],
+  "exclude": ["**/*.test.ts", "**/*.bench.ts"],
   "compilerOptions": {
     "declaration": true,
     "declarationMap": true,


packages/workspace-store/src/navigation/helpers/traverse-tags.ts
```diff
@@ -240,3 +238,28 @@ export const traverseTags = ({
 
   return tags
 }
+
+/**
+ * Immutably sorts an array by a custom order specified in a separate list.
+ *
+ * Copied from @scalar/object-utils/arrays/sort.ts to reduce dependencies.
```
DemonHa, line 245: Seams like you still didn't remove object-utils as a dependency

Is this gonna reduce the bundle size? It should already tree shake since we are importing it directly

packages/icons/scripts/generate-icons.ts
```diff
@@ -140,27 +140,6 @@ const { bind, weight } = useScalarIcon(props)
 }
 
 function generateExports(mappings: Record<string, Record<string, string>>) {
-  const imports = Object.entries(mappings).map(([name]) => {
```
hwkr, line 143: Were we not using this?

packages/api-reference/vite.standalone.config.ts
```diff
@@ -59,18 +59,7 @@ export default defineConfig({
...
-    // * Copying their options fixes the build.
-    // * `max_line_len: 80` is the one setting that makes the difference.
-    //
-    // Source: https://github.com/highlightjs/highlight.js/blob/b9ae5fea90514b864f2c9b2889d7d3302d6156dc/tools/build_config.js#L58-L73
-    terserOptions: {
-      format: {
-        max_line_len: 80,
-      },
-    },
+    minify: 'oxc',
```
hanspagel, line 62: dangerous, because the issue was only happening in the CDN build in jsdelivr
definitely need to monitor this right after release

packages/icons/scripts/generate-icons.ts
hanspagel: I think we still need this, yeah

hanspagel: ah, I think geoff wants to get rid of the barrel file

hanspagel:
I try to split this up in smaller PRs

hanspagel:
Closing in favor of the smaller PRs

PR #8543, opened 2026-03-23
chore(tooling): Tsdown migration POC

liamdscalar:
## Problem

This PR shows how we could migrate from vite to tsdown for libraries. There are a number of reasons we'd want to considering doing this for libraries including:

- automatic dts generation
- automatic package.json exports generation
- automatic dependency exclusion
- First class Vue dts support

This PR also demonstrates a 3 file tsconfig setup

## Checklist

- [ ] I added a changeset. 
- [ ] I added tests.


geoffgscott:
Closing as this has unrelated tsconfig changes (3 tier). We should explore tsdown soon but it seems like it will be a race to see if Vite8 resolved the preserve modules or tsdown resolves its tailwind issues first.

PR #8563, opened 2026-03-23
fix: type declarations don't use the full path for imports

geoffgscott:
## Problem

In nested folders some path extensions would not be correct in the .d.ts files because we were not preserving modules in the rolldown builds

## Checklist

- [ ] I added a changeset. 
- [ ] I added tests.

---

> [!NOTE]
> **Medium Risk**
> Touches shared library build output configuration across multiple packages; changes how Rolldown emits module files and names Vue SFC virtual chunks, which could affect published bundle structure and downstream imports.
> 
> **Overview**
> Fixes incorrect/overly-qualified imports in generated `.d.ts` files by switching affected package builds to **preserve module structure** via a new shared `createPreserveModulesOutput()` Rolldown output helper.
> 
> Adds chunk-name sanitization to avoid `.vue` query-string virtual module filename collisions, and introduces targeted tests for the new output naming logic. Updates Vitest workspace config to include `tooling/*`, and adds a changeset bumping several packages as patch releases.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 007b84331a68097e02247fcecf7bbbab2271cd9c. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/tooling/scripts/vite-lib-config.ts
+++ b/tooling/scripts/vite-lib-config.ts
@@ -14,24 +14,35 @@ import { resolve } from 'node:path'
  * type, index, and lang suffixes to avoid collisions:
  * `Foo.vue?vue&type=style&index=1&lang.css` → `Foo.vue.style.1.css`.
  *
+ * `type` and `index` use `URLSearchParams`. Vue encodes language as
+ * `lang.<ext>` in the query string (not `lang=<ext>`), so we match that
+ * token with a small regex instead of `get('lang')`.
+ *
  * @see https://github.com/rolldown/rolldown/pull/8817
  */
 const sanitizeChunkName = (name: string): string => {
-  const queryIndex = name.indexOf('?')
+  // Look for a query string in the chunk name (e.g., 'Foo.vue?vue&type=...')
+  const queryStart = name.indexOf('?')
 
-  if (queryIndex === -1) {
+  // If there is no query string, this is a facade module; return as-is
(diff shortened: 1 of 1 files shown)


tooling/scripts/vite-lib-config.ts
```diff
@@ -1,6 +1,51 @@
...
+  }
+
+  const base = name.slice(0, queryIndex).replace(/\.vue$/, '')
+  const query = name.slice(queryIndex + 1)
+
+  const typeMatch = query.match(/(?:^|&)type=([^&]+)/)
+  const suffix = typeMatch ? `_vue-${typeMatch[1]}` : '_vue-virtual'
+
+  return `${base}${suffix}`
+}
```
hanspagel, line 49: I was coming to this point, too. but it seems sooo hacky and I actually found even more special cases to handle I think.

There must be a better solution with Vite/rolldown!?

hanspagel: this seems relevant:
https://github.com/rolldown/rolldown/issues/8761

hanspagel: I think we want this fix
https://github.com/rolldown/rolldown/pull/8817

hanspagel:
before we upgraded to Vite 8

image

after

image

this PR seems to fix it, let's try it!

PR #8564, opened 2026-03-23
refactor(components, api-client): move context menu to api client

### LINKED ISSUE #8560: CSS globally overrides z-index on [data-radix-popper-content-wrapper], breaking host app Radix poppers
## Description

When embedding `@scalar/api-reference-react` in a React app that also uses Radix UI, Scalar's CSS applies a global rule that breaks all Radix poppers (dropdowns, tooltips, popovers, selects) in the host application — not just Scalar's own internal ones.

## Root Cause

Scalar's stylesheet includes a rule like:

```css
[data-radix-popper-content-wrapper] {
  z-index: 1 !important;
}
```

This selector is unscoped — it matches **every** `[data-radix-popper-content-wrapper]` in the DOM, including those belonging to the host app. Since it uses `!important`, the host app cannot override it without also using `!important` with higher specificity.

## Impact

Any Radix-based popper in the host app (e.g. `DropdownMenu`, `Select`, `Tooltip`, `Popover`) gets its z-index forced to `1`, rendering it behind other content.

## Expected Behavior

Scalar's CSS should only affect elements inside its own container (e.g. `.scalar-api-reference`). A scoped selector would fix this:

```css
.scalar-api-reference [data-radix-popper-content-wrapper] {
  z-index: 1 !important;
}
```

## Workaround

Host apps can counter-override globally:

```css
:root [data-radix-popper-content-wrapper] {
  z-index: 100000 !important;
}
```

But this is fragile and affects stacking context app-wide.

## Version

`@scalar/api-reference-react@0.9.13`

hwkr:
Fixes #8560

Moves `ScalarContextMenu` out of `@scalar/components` and into `@scalar/api-client` where it's actually used. The old component had a global CSS rule on `[data-radix-popper-content-wrapper]` that was overriding `z-index` for *all* Radix poppers in the DOM — including ones in host apps embedding Scalar. 

* Moves `ScalarContextMenu.vue` → `@scalar/api-client/v2/components/context-menu/ContextMenu.vue`
* Scopes the `z-index` override to only target Scalar's own context menu using `:has(.scalar-api-client-context-menu)` instead of a blanket global selector
* Removes the global `[data-radix-popper-content-wrapper] { z-index: 1 !important; }` rule from `utilities.css`

Also added proper tests for the new `ContextMenu` component.

## Checklist
- [x] I added a changeset.
- [x] I added tests.

### DIFF
--- a/.changeset/blue-cats-stare.md
+++ b/.changeset/blue-cats-stare.md
@@ -1,0 +1,5 @@
+---
+'@scalar/components': patch
+---
+
+fix: remove global radix override


amritk (approved):
Code looks good, did not test

PR #8565, opened 2026-03-24
fix(components): align sidebar to top after opening

hwkr:
Fixes two issues with the sidebar (well one of them is only kinda an issue with the sidebar),

1. Scrolls the container to the top after opening a nested group
2. Ensures the offset is set during animation with an important

Before / after for the jump after opening

https://github.com/user-attachments/assets/4fe7e951-95b6-4e1b-8234-ea5d87da240a

https://github.com/user-attachments/assets/b966d9db-294a-4810-8d25-e26bc7d0cf8a

Todo in the future:
- Handle opening to a specific selected item better (skipping the animation)
- Handle opening and scrolling to a selected item

## Checklist

- [x] I added a changeset. 
- [ ] I added tests.

### DIFF
--- a/.changeset/chatty-papayas-lick.md
+++ b/.changeset/chatty-papayas-lick.md
@@ -1,0 +1,5 @@
+---
+'@scalar/components': patch
+---
+
+fix(components): align sidebar to top after opening


packages/components/src/components/ScalarSidebar/ScalarSidebarNestedItems.vue
```diff
@@ -142,8 +146,9 @@ const handleBack = (event: MouseEvent) => {
     <!-- Make sure the div is around for the entire transition -->
     <Transition
       :duration="300"
-      enterActiveClass="top-(--nested-items-offset)"
-      leaveActiveClass="top-(--nested-items-offset)"
+      enterActiveClass="top-(--nested-items-offset)!"
+      leaveActiveClass="top-(--nested-items-offset)!"
```
hwkr, line 150: This fixes an issue in `ssg-docs` where `top-(--nested-items-offset)` wasn't being applied due to `inset-0` have hight css precedence.

PR #8573, opened 2026-03-24
fix(api-reference): infer discriminator variants from mapping

### LINKED ISSUE #8286: No variant selector for schemas with discriminator.mapping but without explicit oneOf
### What happens?

When a schema uses `discriminator` with `mapping` but does **not** include an explicit `oneOf`, Scalar does not show a variant selector dropdown. The `discriminator.mapping` clearly defines which schemas are the possible variants, but Scalar ignores this information unless `oneOf` is also present.

This is a very common pattern produced by code-first and model-first API tools (e.g. TypeSpec's `@typespec/openapi3` emitter, .NET generators) where:
- The **parent** schema has `discriminator.propertyName` + `discriminator.mapping`
- The **child** schemas use `allOf` to reference the parent

When a schema has `discriminator.mapping`, Scalar should infer the variant list from the mapping entries and display a variant selector dropdown — the same way it does when `oneOf` is explicitly present.

The `discriminator.mapping` already provides all the information needed:

```yaml
discriminator:
  propertyName: shapeType
  mapping:
    circle: "#/components/schemas/Circle"
    rectangle: "#/components/schemas/Rectangle"
    triangle: "#/components/schemas/Triangle"
```

This is semantically equivalent to having:

```yaml
oneOf:
  - $ref: "#/components/schemas/Circle"
  - $ref: "#/components/schemas/Rectangle"
  - $ref: "#/components/schemas/Triangle"
```

Other rendering tools (e.g. Redocly/Redoc) already infer `oneOf` from `discriminator.mapping` and display the variant selector.

### OpenAPI Document

https://sandbox.scalar.com/e/XGW-N

AbdelhamidKhald:
Fixes #8286.

## Summary

Schemas that rely on `discriminator.mapping` without an explicit `oneOf` did not render a variant selector, even though the mapping already defines the variant set.

## Changes

- infer a synthetic `oneOf` from `discriminator.mapping` when no explicit `oneOf` or `anyOf` exists
- normalize plain schema-name mapping entries into local component-schema refs
- reuse the existing schema composition rendering path instead of introducing a new discriminator-only UI branch
- add focused helper regression coverage for inferred discriminator variants

## Testing

- `pnpm exec biome check packages/api-reference/src/components/Content/Schema/helpers/get-compositions-to-render.ts packages/api-reference/src/components/Content/Schema/helpers/get-compositions-to-render.test.ts .changeset/fair-seas-smile.md`
- `pnpm --dir packages/api-reference exec vitest run src/components/Content/Schema/helpers/get-compositions-to-render.test.ts`


DemonHa (dismissed):
Hi @AbdelhamidKhald, thank you so much for the PR — really appreciate the effort here.

That said, I don’t think this is the right place to implement this change. We should avoid modifying the user’s document directly. Instead, it would be better to handle this on the frontend purely for display purposes, transforming the data into the format expected by our current rendering logic.

Thanks again for your contribution!

(cc @hanspagel)

AbdelhamidKhald:
Updated this to the frontend-only path you asked for. The current branch no longer modifies the workspace-store document; instead it injects the active document into the schema render path and builds a display-only inferred composition from discriminator.mapping using the existing component schemas at render time.

hanspagel (changes requested):
looks like the type check is failing, would you mind to fix this?

> @scalar/api-reference@1.49.6 types:check /home/runner/_work/scalar/scalar/packages/api-reference
> vue-tsc --noEmit
>
> Error: src/components/Content/Schema/helpers/get-compositions-to-render.test.ts(61,24): error TS2352: Conversion of type '{ components: { schemas: { Circle: { title: string; type: "object"; properties: { radius: { type: "number"; }; }; }; Rectangle: { title: string; type: "object"; properties: { width: { type: "number"; }; }; }; }; }; }' to type 'OpenApiDocument' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
  Type '{ components: { schemas: { Circle: { title: string; type: "object"; properties: { radius: { type: "number"; }; }; }; Rectangle: { title: string; type: "object"; properties: { width: { type: "number"; }; }; }; }; }; }' is missing the following properties from type '{ openapi: string; info: InfoObject; jsonSchemaDialect?: string | undefined; servers?: ServerObject[] | undefined; paths?: PathsObject | undefined; ... 4 more ...; externalDocs?: ExternalDocumentationObject | undefined; }': openapi, info
 ELIFECYCLE  Command failed with exit code 2.
Error:  command finished with error: command (/home/runner/_work/scalar/scalar/packages/api-reference) /home/runner/setup-pnpm/node_modules/.bin/pnpm run types:check exited (2)

hanspagel (approved):
thanks!

packages/api-reference/src/components/ApiReference.vue
```diff
@@ -307,6 +308,10 @@ const clientStore = createWorkspaceStore({
     }),
   ],
 })
+provide(
+  ACTIVE_DOCUMENT_SYMBOL,
+  computed(() => workspaceStore.workspace.activeDocument ?? null),
+)
```
hanspagel, line 314: @AbdelhamidKhald do you think there would be a way to avoid the provide/inject? we tend to use props and avoid the provide/inject pattern

hanspagel (changes requested):

hanspagel:
I can work on that at some point

hanspagel:
done, ready to re-review @amritk @DemonHa

hanspagel (approved):
let's get this in

PR #8606, opened 2026-03-27
fix(helpers): preserve special characters in query parameters

### LINKED ISSUE #8492: Query parameters with special characters not handled correctly

cyphercodes:
## Problem

Query parameters with special characters like <code>:</code> were being incorrectly URL-encoded. For example, <code>sort=name:asc</code> was being sent as <code>sort=name%3Aasc</code>, which caused backend validation errors.

Fixes #8492

## Solution

Introduced a custom <code>searchParamsToString()</code> function that:
- Preserves commonly used special characters (<code>:</code>, <code>@</code>, <code>#</code>, etc.) that are valid in query values per RFC 3986
- Only encodes characters that are truly problematic in query strings (<code>&amp;</code>, <code>=</code>, <code>%</code>, brackets, and spaces)
- Prevents double-encoding by handling <code>%</code> first

## Changes

- Added <code>encodeQueryParamValue()</code> helper with selective encoding
- Added <code>searchParamsToString()</code> export for converting URLSearchParams without over-encoding
- Updated <code>mergeUrls()</code> to use the new function
- Added comprehensive tests for the new functionality

## Testing

- Added tests for <code>searchParamsToString()</code> covering colons, @ symbols, special characters, encoding of delimiters, and spaces
- Added tests in <code>mergeUrls()</code> for query parameters with colons and special characters
- All existing tests continue to pass

### DIFF
--- a/packages/helpers/src/url/merge-urls.test.ts
+++ b/packages/helpers/src/url/merge-urls.test.ts
@@ -3,7 +3,47 @@
  */
 import { describe, expect, it } from 'vitest'
 
-import { combineUrlAndPath, mergeSearchParams, mergeUrls } from './merge-urls'
+import { combineUrlAndPath, mergeSearchParams, mergeUrls, searchParamsToString } from './merge-urls'
+
+describe('searchParamsToString', () => {
+  it('preserves colons in values', () => {
+    const params = new URLSearchParams('sort=name:asc')
+    expect(searchParamsToString(params)).toBe('sort=name:asc')
+  })
+
+  it('preserves special characters like @ and #', () => {
+    const params = new URLSearchParams('email=user@example.com&hash=test#123')
+    expect(searchParamsToString(params)).toBe('email=user@example.com&hash=test#123')
+  })
+
+  it('encodes ampersands and equals signs', () => {
+    const params = new URLSearchParams('filter=a=b&test=c&d')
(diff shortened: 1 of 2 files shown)


hanspagel:
I think this is what `allowReserved: true` is for: 

https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.1.md#fixed-fields-for-use-with-schema

https://github.com/scalar/scalar/pull/8501 added support for it

PR #8610, opened 2026-03-27
fix(api-client): improve editor background color handling

hwkr:
So the `alternate` theme was breaking things because it has a `.light-mode .dark-mode` selector. This makes it so it has to be be an exact selector e.g. `.light-mode`. I then added some tests and discovered there were a lot of other values that didn't resolve properly either due to them being `rgb` instead of hex or them referencing a variable so I updated the script to account for that.

Also fixed up the corner rounding so the background doesn't clip it.

Alternate Theme Before / After:

CleanShot 2026-03-27 at 14 44 59@2x
CleanShot 2026-03-27 at 14 44 09@2x

Default Theme Before / After:

CleanShot 2026-03-27 at 14 45 18@2x
CleanShot 2026-03-27 at 14 45 28@2x

## Checklist

- [x] I added a changeset. 
- [x] I added tests.

### DIFF
--- a/packages/api-client/src/v2/features/editor/helpers/theme/load-css-variables.ts
+++ b/packages/api-client/src/v2/features/editor/helpers/theme/load-css-variables.ts
@@ -1,10 +1,128 @@
 import { z } from 'zod'
 
+/** Which theme bucket a CSS rule applies to after parsing. */
+type ColorMode = 'light' | 'dark'
+
+/** Accumulated custom properties per mode while walking stylesheet rules. */
+type Variables = Record<ColorMode, Record<string, string>>
+
+/**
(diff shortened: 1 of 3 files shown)


packages/api-client/src/v2/features/editor/helpers/theme/load-css-variables.ts
```diff
@@ -40,13 +40,13 @@ export const loadCssVariables = async (css: string) => {
 
         return style
       }, {})
-
+      console.log(cur.selectorText)
```
DemonHa: forgot console.log

DemonHa (approved):

PR #8620, opened 2026-03-30
chore(oas-utils): remove @scalar/openapi-parser dependency

hanspagel:
with the new rolldown, tree-shaking doesn't work as well and the bundle size increased

so to help here, we can remove @scalar/openapi-parser (heavy dependencies) from @scalar/oas-utils
the methods that imported it are not used in the repository anymore


hanspagel:
my mistake, it's still used in the api-client v1 code that we are about to delete

PR #8690, opened 2026-04-01
build: fix turbo/eslint filter broken by MSYS path conversion

YagUber:
## Problem

The `build:packages`, `build:integrations`, and `lint:vue` scripts use single quoted glob patterns (e.g. `'./packages/**'`). On Windows with MSYS, the path conversion layer mangles these patterns before they reach Turbo/ESLint, causing `build:packages` and `build:integrations` to run 0 packages.

## Solution

Replace single quoted glob patterns with escaped double quotes (e.g. `"./packages/**"`), consistent with how the `clean` script already handles its arguments. This prevents MSYS path conversion from interfering while keeping identical behaviour on Linux and macOS.

No alternative solutions were considered that wouldn't risk changing semantics, such as switching to name based filters like `@scalar/*` as it would exclude non `@scalar` scoped packages such as `scalar.aspnetcore`.

## Checklist

- [ ] I added a changeset.
- [ ] I added tests.


hanspagel:
cool thanks

PR #8711, opened 2026-04-02
fix(api-client): surface OAuth2 token endpoint error responses

lawrence3699:
## What

Noticed while debugging a Client Credentials setup with a wrong `client_secret` — the toast just said *"Failed to authorize"* with no hint about what went wrong. Traced it to `authorizeServers` in both `oauth2.ts` and the v2 `oauth.ts`: the fetch call never checks `resp.ok`, so when the token endpoint returns e.g. HTTP 401 with `{"error":"invalid_client","error_description":"Client authentication failed"}`, the body is parsed fine but `responseData.access_token` is `undefined`. The function returns `[null, undefined]` and the caller's fallback message kicks in.

Meanwhile `fetchOpenIDConnectDiscovery` in the same package already does the right thing:

```ts
if (!response.ok) {
  return [new Error(`Failed to fetch …: ${response.status} ${response.statusText}`), null]
}
```

## Fix

Added a `!resp.ok` check after `resp.json()` in both `authorizeServers` implementations. When the token server responds with an error status, the `error_description` (or `error` code, per RFC 6749 §5.2) is extracted from the parsed body and returned as the Error message. This way the user sees e.g. *"Token request failed: Client authentication failed"* instead of the generic fallback.

## Before / After

| | Before | After |
|---|---|---|
| Wrong `client_secret` | Toast: *"Failed to authorize"* | Toast: *"Token request failed: Client authentication failed"* |
| Expired auth code | Toast: *"Failed to authorize"* | Toast: *"Token request failed: invalid_grant"* |
| Network error | Toast: *"Failed to get an access token…"* | *(unchanged)* |

## Tests

- Updated all existing fetch mocks to include `ok: true` (they were missing it, which would have made them fail silently under the new check)
- Added 2 test cases per implementation (v1 + v2 = 4 total):
  - token endpoint returns 401 with `error_description` → error message includes the description
  - token endpoint returns 400 with only `error` code → error message falls back to the code
- All 31 v1 tests and 37 v2 tests pass

### DIFF
--- a/.changeset/swift-tokens-surface.md
+++ b/.changeset/swift-tokens-surface.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
+
+fix: surface OAuth2 token endpoint error responses with RFC 6749 error_description
--- a/packages/api-client/src/v2/blocks/scalar-auth-selector-block/helpers/oauth.test.ts
+++ b/packages/api-client/src/v2/blocks/scalar-auth-selector-block/helpers/oauth.test.ts
@@ -734,6 +734,19 @@ describe('oauth', () => {
       expect(error!.message).toBe('Token request failed: invalid_grant')
     })
 
+    it('should fall back to HTTP status when the error response body is not JSON', async () => {
+      global.fetch = vi.fn().mockResolvedValueOnce({
+        ok: false,
+        status: 503,
+        json: () => Promise.reject(new SyntaxError('Unexpected token')),
+      })
+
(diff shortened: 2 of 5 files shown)


hanspagel (changes requested):
cool
pls add a changeset with "pnpm changeset"

lawrence3699:
Added the changeset: `.changeset/swift-tokens-surface.md` for `@scalar/api-client` patch.

PR #8733, opened 2026-04-03
feat(openapi-types): generate openapi types from schemas

### LINKED ISSUE #6382: Lack of type-safety in @scalar/openapi-types
I would like to use `@scalar/openapi-parser` but currently that would require a lot of type casting:

```ts
import { dereference } from "@scalar/openapi-parser";

const { schema, errors } = await dereference({});

if (schema?.openapi) {
  const pathItems = Object.values(schema!.paths!)
  // pathItems has type any[] :(
}
```

Apparently this is because of the `[key: string]: any` in `AnyOtherAttribute`

https://github.com/scalar/scalar/blob/c85d313923f83be32ab5d3a08721d1eabae0decd/packages/openapi-types/src/openapi-types.ts#L11-L16

when I remove it the type is inferred correctly.

amritk:
Generated the openapi types, they are now more strict while still allowing x-extensions in most places.

All so namespaces are not [really recommended](https://biomejs.dev/linter/rules/no-namespace/) anymore so we have direct module exports. Do we still want to keep the Object suffix? 

## Checklist

- [ ] I added a changeset. 
- [ ] I added tests.

---

> [!NOTE]
> **Medium Risk**
> Large type-surface change plus stricter OpenAPI object guards in `@scalar/mock-server` may affect downstream TypeScript builds and edge-case auth/response generation behavior.
> 
> **Overview**
> **Generates and publishes new strict OpenAPI type declarations** for 2.0/3.0/3.1 (module exports + new `*.d.ts` files) and fixes `@scalar/openapi-types` packaging to include the root `index.d.ts`.
> 
> **Hardens OpenAPI handling across the stack** by adding shared OpenAPI guard helpers (e.g., `$ref` vs concrete objects) and using them in `@scalar/mock-server` to narrow security-scheme filtering/logging and to avoid treating referenced responses/headers as concrete during mock response generation.
> 
> **Type-safety cleanups in clients/parsers**: adjusts API client/server variable types, Postman import casting, request watch-mode payload building, safer example/parameter extraction in `@scalar/oas-utils`, and minor parser join/validation/test tweaks.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 090fa17589f0daaec9192bdeb808629b79ccbbe7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

Fixes #6382


DemonHa (commented):
Didn't check in details but yeah they looks really nice with all the comments

packages/openapi-types/2.0/api-key-security.d.ts
```diff
@@ -0,0 +1,17 @@
+/**
+ * Security Scheme object
+ *
+ * Allows the definition of a security scheme that can be used by the operations. Supported schemes are basic authentication, an API key (either as a header or as a query parameter) and OAuth2's common flows (implicit, password, application and access code).
+ *
+ * @see {@link https://swagger.io/specification/v2/#security-scheme-object}
```
hanspagel: this is a smartbear site
the real source of truth for everything is github:
https://github.com/OAI/OpenAPI-Specification

packages/openapi-types/3.0/api-key-security-scheme.d.ts
hanspagel, line 1: I think we only want 2.0, 3.0, 3.1 and 3.2, not the patch versions e.g. 3.0.4
Those patch versions don't introduce features, they just clarify, so the comments might update, that's fine.
I don't think anyone wants to explicitly have types for let's say 3.0.4, but for 3.0.

packages/openapi-types/3.0/api-key-security-scheme.d.ts
```diff
@@ -0,0 +1,17 @@
+/**
+ * Security Scheme object
+ *
+ * Defines a security scheme that can be used by the operations.  Supported schemes are HTTP authentication, an API key (either as a header, a cookie parameter, or as a query parameter), OAuth2's common flows (implicit, password, client credentials, and authorization code) as defined in [RFC6749](https://tools.ietf.org/html/rfc6749), and [[OpenID-Connect-Core]]. Please note that as of 2020, the implicit flow is about to be deprecated by [OAuth 2.0 Security Best Current Practice](https://tools.ietf.org/html/draft-ietf-oauth-security-topics). Recommended for most use cases is Authorization Code Grant flow with PKCE.
+ *
+ * @see {@link https://spec.openapis.org/oas/v3.0.4#security-scheme-object}
```
hanspagel: I’d prefer GitHub here too

https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.4.md#security-scheme-object

hanspagel (changes requested):
awesome work!

I think this is a rare use case where namespaces make sense. And I can't read anything about a recommendation to move away from them in the official docs: https://www.typescriptlang.org/docs/handbook/namespaces.html

```ts
// namespace
import type { OpenAPI31 } from '@scalar/openapi-types/3.1'
import type { OpenAPI32 } from '@scalar/openapi-types/3.2'
```

vs.

```ts
// name conflicts
import type { Document as OpenAPI31Document } from '@scalar/openapi-types/3.1'
import type { Document as OpenAPI32Document } from '@scalar/openapi-types/3.2'
```

feels super odd to be forced to rename the imports, wdyt?

* if it's just "feels outdated" or "someone says it’s not recommended, no matter what use case", I’d keep the namespaces
* if there's some user value in moving away, we can say good bye to the namespaces

(keeping the namespaces makes migration easier too)

amritk:
Namespaces are a relic from the time before ESM. Here's a link to the Anders speaking on it

Anders Heilsberg's statement: [https://youtu.be/tXK50czRbdA?t=990](https://youtu.be/tXK50czRbdA?t=990)

If you want to keep them you should just do a separate export wrapped in namesspaces I think, no need to make it the main export. I know these are just types so it doesn't matter as much but if we did add the parser functions here, namespaces aren't as easily tree shakable as esm.

packages/openapi-types/3.0/api-key-security-scheme.d.ts
amritk: yea it grabs it from the jsdoc but we can overwrite

packages/openapi-types/3.0/api-key-security-scheme.d.ts
amritk: but we still want the latest versions right

amritk:
Alright @hanspagel added back in "namespace" exports

```ts
import type OpenAPIV3_2  from '@scalar/openapi-types'
```

hanspagel (changes requested):
nice!

- [ ] add a major changeset for openapi-types
- [ ] add a section to the README to describe how to migrate to the new version (let a LLM write it, I’ll check the result)
- [ ] make the build pass

amritk:
I'm going to change this to a much less jarring change, putting on ice for now as switching gears

amritk:
closing in favor of https://github.com/scalar/scalar/pull/8942

PR #8759, opened 2026-04-07
fix(api-reference): hide response tabs without body content

### LINKED ISSUE #4302: Example responses contains all status code that provided in OpenApi file even those that doesn't have any content
### What happens?

Code Snippet contains all status code that provided in OpenApi file even those that doesn't have any content

image

https://developers.thomsonreuters.com/pages/api-reference/d6f0ac89-b7e3-49af-811d-2d453193b7be#tag/e-file-reference/GET/v1/e-file-jurisdictions

If we have a response with no content, we shouldn't be able to select that empty response. In this example only 200 response should be displayed as it is the only one that has data:

image

### OpenAPI Document

_No response_

DemonHa:
Fixes: #4302

## Summary

- Hide response tabs (like 204 No Content) that have only a description but no actual body content (schema, example, or examples)
- Simplify `hasResponseContent` to check all media types for content instead of special-casing certain status codes
- Export `hasMediaTypeContent` for direct use and add comprehensive tests for it

| Before      | After |
| ----------- | ----------- |
| image       | image      | 

## Checklist

- [x] I added a changeset. 
- [x] I added tests.

### DIFF
--- a/.changeset/happy-buttons-dance.md
+++ b/.changeset/happy-buttons-dance.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': minor
+---
+
+feat: hide response tabs without body content
--- a/packages/api-reference/src/features/example-responses/ExampleResponses.test.ts
+++ b/packages/api-reference/src/features/example-responses/ExampleResponses.test.ts
@@ -710,7 +710,7 @@ describe('ExampleResponses', () => {
     expect(wrapper.text()).not.toContain('Example 2')
   })
 
-  it('renders explicit 204 response tab and shows No Body', () => {
+  it('does not render explicit 204 response', () => {
     const wrapper = mount(ExampleResponses, {
       props: {
         responses: {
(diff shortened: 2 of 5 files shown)


hanspagel (changes requested):
we had this, and I had to revert this
no time to gather the context right now, but pls don't merge

DemonHa:
> we had this, and I had to revert this no time to gather the context right now, but pls don't merge

Can you elaborate more on what the problem with this is?

marclave:
the problem is some people wanted it and some didnt, so we made this design decision but maybe its configurable?

hanspagel:
This one is the other PR with the discussion:
https://github.com/scalar/scalar/pull/8298

And the issue is old (2024) and had no reactions and nothing, so I don't think we need to add a config for it. Let's see if it comes up again.

Sorry @DemonHa, thanks for looking into it

PR #8779, opened 2026-04-10
feat(component): don't add word breaks for the start of text

hwkr:
Previously we were adding a `<wbr>` at the start of ScalarWrappingText which could cause weird behavior if no other word break opportunities were found.

Before / After
CleanShot 2026-04-10 at 13 43 03@2x
CleanShot 2026-04-10 at 13 46 29@2x

Also set the sidebar to treat example labels as properties since that's what folks seem to sometimes have there.

## Checklist

- [x] I added a changeset. 
- [x] I added tests.

### DIFF
--- a/.changeset/floppy-canyons-heal.md
+++ b/.changeset/floppy-canyons-heal.md
@@ -1,0 +1,5 @@
+---
+'@scalar/sidebar': patch
+---
+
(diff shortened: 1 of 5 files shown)


marclave (approved):
AWESOME

PR #8780, opened 2026-04-10
fix(api-reference): run intersect for the top of the document

DemonHa:
This PR fixes an issue introduced by the `Intersection Observer` changes, where `documentStartRef` was no longer triggering because the observed intersection area was positioned in the middle of the viewport.

With this update, the event now correctly fires when the user scrolls back to the top of the page, ensuring the hash is removed as expected.

## Checklist

- [x] I added a changeset. 
- [ ] I added tests.

---

> [!NOTE]
> **Medium Risk**
> Touches scroll/IntersectionObserver behavior used for in-page navigation, which can be browser- and layout-sensitive and may subtly affect active section selection.
> 
> **Overview**
> Fixes top-of-page nav/hash syncing by making the document-start sentinel intersection fire immediately when it enters the viewport, rather than only in the viewport-center strip.
> 
> Extends `useIntersection` to support `immediate` root margins and an optional `onExit` callback; `ApiReference.vue` uses `onExit` to re-emit the currently centered `SECTION` id when leaving the top sentinel to keep the active nav item correct.
> 
> Includes a patch changeset for `@scalar/api-reference`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3e0fabd7e56f8c4b805d7e97d57e26d1adefbdf7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


amritk (approved):
tested and works for me, I would like to switch to something like a sorted map for a more robust solution but we can do that later

marclave (approved):

PR #8790, opened 2026-04-13
feat(analytics): Add posthog to api-reference

liamdscalar:
## Problem

Adds posthog analytics to api-reference. Disabled by default.

## Checklist

- [ ] I added a changeset. 
- [ ] I added tests.


hanspagel:
adds 175.33KiB for everyone

new PR where the code is moved to plugins is coming in a second

Issue #8789, opened 2026-04-12
Test Request does not use exploded form data in query params


jonasehrlich:
### What happens?

I have an API endpoint that expects a `bbox` array in the query parameters. Note the schema does not explicitly set the `explode` attribute to `true`. But this should be the default anyways.

``` json
 "/v1/flights/positions": {
      "get": {
        "operationId": "get_flight_positions",
        "parameters": [
          {
            "in": "query",
            "name": "bbox",
            "required": true,
            "schema": {
              "type": "array",
              "examples": [
                [
                  13.0,
                  48.0,
                  18.0,
                  52.0
                ]
              ],
              "items": {
                "type": "number",
                "format": "double",
                "maximum": 180.0,
                "minimum": -180.0
              },
              "maxItems": 4,
              "minItems": 4
            },
            "style": "form"
          },
         
        ],
    ...
}
```

This is rendered correctly in the documentation:

<img width="773" height="140" alt="Image" src="https://github.com/user-attachments/assets/6e21352d-883a-4890-a860-097ec5ae9a2f" />

And also seemingly correct in the test client (in the API Reference, click on `Test Request`):

<img width="1277" height="1113" alt="Image" src="https://github.com/user-attachments/assets/51e7c905-ee47-4e9f-8157-0e417ab92257" />

But when this client is used to send a request, only the last value in the array persists:

<img width="548" height="585" alt="Image" src="https://github.com/user-attachments/assets/f82edb2e-2a3e-47e7-a899-d6ff685d4f33" />

Which will then obviously create an API error. I also tried using the Scalar App, and using **the same OpenAPI spec**, the parameters are exploded correctly and the server responds:

<img width="1392" height="832" alt="Image" src="https://github.com/user-attachments/assets/82e1de89-2e30-4576-86fc-0ad45794ea17" />

### `@scalar/api-reference`

```yaml
'@scalar/api-reference@1.52.1(axios@1.14.0)(tailwindcss@4.2.2)(typescript@6.0.2)':
    dependencies:
      '@headlessui/vue': 1.7.23(vue@3.5.32(typescript@6.0.2))
      '@scalar/agent-chat': 0.10.3(axios@1.14.0)(tailwindcss@4.2.2)(typescript@6.0.2)
      '@scalar/api-client': 2.43.0(axios@1.14.0)(tailwindcss@4.2.2)(typescript@6.0.2)
      '@scalar/code-highlight': 0.3.2
[...]

PR #8791, opened 2026-04-13
fix(api-reference): preserve exploded array query params in test requests

alonfaraj:
## Problem

API Reference `Test Request` was collapsing exploded query array params to the last value. In the issue repro, `bbox` should be sent as repeated query params, but the web request ended up with only the final entry.

## Solution

I fixed the request URL builder in `@scalar/workspace-store` so repeated query keys are preserved instead of overwritten during URL resolution.

I also added regression tests for the repeated `bbox` query-param case.

## Checklist

- [x] I added a changeset.
- [x] I added tests.

## Fixes

https://github.com/scalar/scalar/issues/8789

### DIFF
--- a/.changeset/cuddly-tools-hug.md
+++ b/.changeset/cuddly-tools-hug.md
@@ -1,0 +1,6 @@
+---
+'@scalar/workspace-store': patch
+'@scalar/api-reference': patch
+---
+
+fix test requests so exploded array query parameters keep all values instead of collapsing to the last one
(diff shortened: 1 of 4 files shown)


hanspagel:
@DemonHa can u review pls

amritk (approved):
Looks like it matches previous behaviour

DemonHa (approved):
LGTM ✨

Thank you for this PR 🙏

PR #8826, opened 2026-04-13
feat(Scalar.AspNetCore): Deprecated Attribute

Thiago-Mausbach:
## Problem
Scalar does not provide a built-in way to mark OpenAPI endpoints as deprecated using a C# attribute, similar to SwaggerGen's [Obsolete] support.

When migrating from SwaggerGen, there was no equivalent attribute available for marking endpoints as deprecated in OpenAPI output.

## Solution
This PR introduces support for a custom [Deprecated] attribute.

The attribute is detected during OpenAPI generation and maps to the OpenAPI "deprecated: true" field, allowing endpoints to be visually marked as deprecated in Scalar UI.

## Checklist

- [x] I added a changeset. 
- [x] I added tests.


marclave (approved):
awesome thanks for this work @Thiago-Mausbach !

hanspagel (approved):
that's cool!

hanspagel:
I take over here :v:

bartvanandel:
I'm curious: why not use the default [`System.ObsoleteAttribute`](https://learn.microsoft.com/en-us/dotnet/api/system.obsoleteattribute?view=net-10.0)? We've been using that ourselves now for a while, with a custom transformer to set `deprecated: true`, not just for endpoints but also for schemas and schema properties. I can send our transformer if you like to play around with it.

Side note: your solution concatenates the deprecation reasons straight onto the description, without adding any separator like a space or newline. So you may end up with a description like "Description of old endpointuse /new/path".

PR #8828, opened 2026-04-14
fix(api-client): Add support for text/json media type

da411d:
## Problem

Some weirdos (me) use `content-type: text/json` instead of `application/json`.  
Because of that, in the response it doesn't show the json tree, instead is just says "Binary file".

image

## Solution

Of course I can just update headers to `application/json`. And I did it. But why not make life easier for future users?

## Checklist

- [X] I added a changeset. 
- [ ] I added tests.

---

To be honest, I didn't manage to run any build scripts in my IDE. Maybe because all scripts are not made for Windows. So this is blind pull request, but I'm 100% sure that will work!

Thank you

### DIFF
--- a/.changeset/deep-jeans-like.md
+++ b/.changeset/deep-jeans-like.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
+
+Add support for `text/json` media type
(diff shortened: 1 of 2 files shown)


DemonHa:
@amritk I wonder if we should allow people to customize preview of different content types using plugins

marclave (approved):
not weird at all :) let's get this in! 

appreciate the contribution!

on an aside we probs need a better fallback tho cc @amritk @DemonHa @hanspagel

DemonHa (approved):
Thank you for your contribution 🚀

PR #8838, opened 2026-04-14
build(agent-chat, code-highlight): replace cp with shx cp in package build scripts

YagUber:
## Problem

The `agent-chat` and `code-highlight` package build scripts use the Unix `cp` command to copy build artifacts into `dist`.

This works in Unix-like environments, but can fail on Windows where `cp` is not available by default.

## Solution

Replace `cp` with `shx cp` in the affected package `build` scripts:
- `packages/agent-chat/package.json`
- `packages/code-highlight/package.json`

Many other scripts already use `shx`, so this appears the be the intended solution.

## Checklist

- [ ] I added a changeset.
- [ ] I added tests.

### DIFF
--- a/packages/agent-chat/package.json
+++ b/packages/agent-chat/package.json
@@ -15,7 +15,7 @@
     "node": ">=22"
   },
   "scripts": {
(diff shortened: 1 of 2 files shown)


hanspagel (approved):
🙏 🙏 🙏 🙏 🙏

PR #8839, opened 2026-04-14
fix(api-client): replace v-show transition with visibility/opacity to fix modal stutter

YagUber:
## Problem

The modal fade-in animation stuttered on every open because `v-show` hides the container with `display: none`, which forces the browser to recalculate layout for the entire modal subtree (Operation, Sidebar, RequestBlock, etc.) on every open, blocking the main thread during the first animation frames.

Attempts to fix this by reordering `modalState.show()` and `handleSelectItem` only shifted when the JS work happened, but the forced layout recalculation when `display: none` is removed is unavoidable regardless of JS ordering.

## Solution

Replaced `<Transition name="scalar-client-fade">` + `v-show` on the outer div with a CSS visibility/opacity approach on `.scalar-container` directly:
- The container is always in the layout tree (`visibility: hidden` instead of `display: none`), eliminating the forced layout recalculation on open entirely.
- `will-change: opacity` keeps a GPU compositor layer allocated so the opacity animation runs on the compositor thread independently of any Vue patch work on the main thread.
- Open: visibility snaps to visible immediately, opacity fades in over 0.35s.
- Close: opacity fades out over 0.35s, then visibility snaps to hidden via `transition-delay`.

Also fixed an issue I noticed:
`fallbackFocus` broken selector `useFocusTrap` was passed `#:r0:` (Vue's `useId()` output), which is an invalid CSS selector. `querySelector` silently returned null so the fallback never worked. Replaced with the ref directly.

## Demo

<details><summary><strong>Before</strong></summary>
<p>

https://github.com/user-attachments/assets/96a8f494-c9f7-406b-af9f-49fdd88e4f0f

</p>
</details> 

<details><summary><strong>After</strong></summary>
<p>

https://github.com/user-attachments/assets/22f51369-92a1-4683-912f-e625b2121e78

</p>
</details> 

## Checklist

- [x] I added a changeset.
- [ ] I added tests.

### DIFF
--- a/.changeset/cozy-sites-bake.md
+++ b/.changeset/cozy-sites-bake.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
+
+Fix modal fade-in animation stutter by replacing v-show with CSS visibility/opacity to avoid forced layout recalculation on open
--- a/packages/api-client/src/v2/components/modals/ModalClientContainer.spec.ts
+++ b/packages/api-client/src/v2/components/modals/ModalClientContainer.spec.ts
@@ -53,7 +53,8 @@ describe('ModalClient.vue', () => {
       },
     })
 
-    expect(wrapper.find('.scalar').isVisible()).toBe(false)
+    // Visibility is controlled via CSS class (not v-show/display:none) to avoid layout reflow on open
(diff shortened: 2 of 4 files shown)


amritk (approved):
I wonder if this will solve the react issues!

amritk:
I need this one so I'm merging @hwkr

PR #8841, opened 2026-04-14
fix(dev): use cross-env for Windows compatible dev commands

YagUber:
## Problem

Some development commands currently rely on Unix-style inline environment variable assignment, for example:

- `CI=1 pnpm --filter @scalar/void-server dev`
- `CI=1 pnpm --filter proxy-scalar-com dev`
- `ENV=dev PORT=5051 go run main.go`

As a result, these dev commands will fail when running the local development workflow on Windows.

## Solution

Replace inline environment variable assignment with `cross-env` in the affected scripts so the commands behave consistently across Windows, Linux, and macOS.

Many other scripts already use `cross-env`, so this appears the be the intended solution.

## Checklist

- [ ] I added a changeset.
- [ ] I added tests.

### DIFF
--- a/projects/proxy-scalar-com/package.json
+++ b/projects/proxy-scalar-com/package.json
@@ -20,7 +20,7 @@
   },
   "scripts": {
     "build": "go build main.go",
-    "dev": "ENV=dev PORT=5051 go run main.go",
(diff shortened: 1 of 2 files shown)


hanspagel (approved):
🙏

PR #8842, opened 2026-04-15
perf(api-reference): lazy-load CodeMirror and api-client modal chunks

YagUber:
## Problem

For #8829 / codemirror.

Every page that uses `@scalar/api-reference` downloads all `@codemirror/*` packages on first load, even when the spec has no `x-scalar-environments` and the user never opens the request panel.

CodeMirror is heavy (~300 kB minified), and it blocks TTI for the majority of read-only documentation pages.

## Solution

Two independent code-split boundaries were introduced:

#### `@scalar/use-codemirror`

All `@codemirror/*` imports were moved from `useCodeMirror.ts` into a new `codemirror-setup.ts` file that is never imported statically. `useCodeMirror` now calls `import('./codemirror-setup')` on first mount, caching the result in a module-level singleton so the chunk is fetched at most once regardless of how many editors are on the page.

`preloadCodeMirror()` is exported for call sites that want to start fetching the chunk early without blocking.

#### `ApiReference.vue`

`createApiClientModal` is no longer statically imported. The modal chunk is fetched lazily on the first `ui:open:client-modal` event.

Additionally, if the parsed spec contains `x-scalar-environments`, the modal chunk begins prefetching immediately after spec load so it is already in cache before the user clicks "Test Request".

**Tests**

The `vitest.setup.ts` for `api-reference` calls `await preloadCodeMirror()` in a `beforeAll` so the module is fully resolved before any test mounts a component. Without this, the dynamic import fires during mount but resolves after the test environment tears down (`EnvironmentTeardownError`).

**ESM Playground Results**
- Before `index.js`
  - `1,191.81 kB │ gzip: 376.94 kB`
- After `index.js`
  - `922.07 kB │ gzip: 276.64 kB`

## Checklist

- [x] I added a changeset.
- [ ] I added tests.

---

Notes:

This is more of a PoC than anything else. I am confident in the solution, but like I mentioned in that discussion, I am not sure if this is semantically correct for this codebase. Scalar is very large and somewhat overwhelming. These somewhat "free" gains are nice though.

This might also require changes to the documentation.

### DIFF
--- a/packages/api-reference/src/components/ApiReference.vue
+++ b/packages/api-reference/src/components/ApiReference.vue
@@ -1,10 +1,8 @@
 <script setup lang="ts">
 import { provideUseId } from '@headlessui/vue'
 import { OpenApiClientButton } from '@scalar/api-client/v2/blocks/operation-block'
-import {
-  createApiClientModal,
-  type ApiClientModal,
-} from '@scalar/api-client/v2/features/modal'
+import type * as ApiClientModalModule from '@scalar/api-client/v2/features/modal'
+import type { ApiClientModal } from '@scalar/api-client/v2/features/modal'
 import {
   addScalarClassesToHeadless,
   ScalarColorModeToggleButton,
@@ -43,7 +41,6 @@ import {
   computed,
   onBeforeMount,
   onBeforeUnmount,
-  onMounted,
   onServerPrefetch,
   provide,
   ref,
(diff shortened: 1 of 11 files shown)


hanspagel (commented):
thanks, sounds great!

@amritk wdyt

amritk:
Hello there! Thanks for doing this! We were thinking to get our ESM CDN build working before doing this but it definitely helps out the non-cdn folks. Do you think you can make a generic lazy load helper in @scalar/helpers? Then we can use it in the other packages as well. For reference this is what we do on the react client

https://github.com/scalar/scalar/blob/main/packages/api-client-react/src/lazy-load.ts

packages/api-reference/src/components/ApiReference.vue
```diff
@@ -723,12 +722,52 @@ provide(AGENT_CONTEXT_SYMBOL, agent)
...
+ */
+let _modalModulePromise: Promise<typeof ApiClientModalModule> | null = null
+
+/** Start fetching the api-client chunk without blocking the caller. */
+const loadModalModule = () => {
+  if (!_modalModulePromise) {
+    _modalModulePromise = import('@scalar/api-client/v2/features/modal')
+  }
+  return _modalModulePromise
+}
```
DemonHa, line 754: can we possibly refactor this to use the same helper `makeLazySingleton` from `packages/api-client-react/src/lazy-load.ts`

packages/api-reference/src/components/ApiReference.vue
```diff
@@ -738,8 +777,56 @@ onMounted(() => {
...
+ * loading delay.
+ *
+ * When no environments are configured the chunk is deferred until the user
+ * actually opens the panel for the first time.
+ */
+watch(
+  () => clientStore.workspace.activeDocument?.['x-scalar-environments'],
+  (envs) => {
+    if (envs && Object.keys(envs).length > 0) {
+      loadModalModule()
```
DemonHa, line 800: @amritk should we just always do this instead, so we try to load it on the background

packages/api-reference/src/components/ApiReference.vue
```diff
@@ -738,8 +777,56 @@ onMounted(() => {
...
+  { immediate: true },
+)
+
+/**
+ * Intercept the first `ui:open:client-modal` event so the modal can be
+ * created lazily. Once the modal exists its own `initializeModalEvents`
+ * handler takes over for all subsequent events.
+ */
+let _unsubscribeInterceptor: (() => void) | null = null
+_unsubscribeInterceptor = eventBus.on('ui:open:client-modal', (payload) => {
```
DemonHa, line 812: I think this is a good usecase for `.once` which I think we had before not sure what happend (cc @amritk)

DemonHa (changes requested):
Great work! 🚀

Left some comments

packages/api-reference/src/components/ApiReference.vue
amritk: here you go

https://github.com/scalar/scalar/pull/9028

amritk:
hey @YagUber feel free to make those changes, if not when I get some time I can give it a once over

hanspagel:
Thanks for bringing this up! I’ll close this PR in favor of #9310 and others :)

PR #8849, opened 2026-04-15
fix(api-client): improve sidebar menu and drag regions

hwkr:
We were setting a drag region in the sidebar below the items which was interfering with the sidebar menu. This PR:

- Removes the sidebar drag region (it was a little confusing anyway)
- Makes it so you can drag the top tab if you only have one tab open
- Fixes ScalarMenu so the buttons are semantically correct

Before:

https://github.com/user-attachments/assets/85f24c27-84dd-4e25-9cb3-1dfecc92770c

After:

https://github.com/user-attachments/assets/4e599311-4f6e-4a79-86c0-b29b0a40edab

## Checklist

- [x] I added a changeset. 
- [x] I added tests.

### DIFF
--- a/.changeset/silly-rockets-wink.md
+++ b/.changeset/silly-rockets-wink.md
@@ -1,0 +1,5 @@
+---
+'@scalar/components': patch
+---
+
+fix(components): apply ScalarMenuLink is prop correctly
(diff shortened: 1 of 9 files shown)


hanspagel (approved):
OMG so much better, thanks
looks like I broke it then

PR #8852, opened 2026-04-15
fix(workspace-store): check for `'null'` origin on base url

### LINKED ISSUE #8851: Framed scalar document fails to send request
### What happens?

When I test a request from my openapi service, I get failed url

Image

I can send a request the same way as an unframed instance of scalar

### OpenAPI Document

_No response_

bmanth60:
## Problem

Fixes https://github.com/scalar/scalar/issues/8851

Allows sending requests when scalar is framed in srcdoc

## Solution

Added a string 'null' check and if true, use localhost as the base for the url

## Checklist

- [x] I added a changeset. 
- [x] I added tests.

### DIFF
--- a/.changeset/plain-monkeys-talk.md
+++ b/.changeset/plain-monkeys-talk.md
@@ -1,0 +1,5 @@
+---
+'@scalar/workspace-store': patch
+---
+
+fix: Allows sending requests when scalar is framed in srcdoc
--- a/packages/workspace-store/src/request-example/builder/resolve-request-factory-url.test.ts
+++ b/packages/workspace-store/src/request-example/builder/resolve-request-factory-url.test.ts
@@ -1,4 +1,4 @@
-import { describe, expect, it } from 'vitest'
+import { describe, expect, it, vi, afterEach } from 'vitest'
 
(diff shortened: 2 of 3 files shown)


amritk (approved):
hey thanks for the quick fix, this seems reasonable @hanspagel ?

hanspagel (changes requested):
TIL srcdoc, what's your use case for it?

great find!

I can take over here and bring it over the finish line

- [x] I think we can clean up the code a bit and add a comment
- [x] add a changeset

hanspagel (approved):

DemonHa (approved):

bmanth60:
Thanks for the quick review and merge, I appreciate it!

My use case is that I have a set of backend apis that is using [hono openapi](https://hono.dev/examples/hono-openapi) to generate the openapi json. I want all of these to be parsed by scalar, and specifically I want them to be consistent in theming, and they share the same security schemes, so this is provided by a single docs service. These docs are displayed in our app and is only displayed for privileged users, so the srcdoc frame allows me to simply forward credentials from our app into scalar to populate the requests.

PR #8865, opened 2026-04-16
feat(api-client): add plugin support for custom response body content types

### LINKED ISSUE #6846: Support application/msgpack or plugin support for custom encodings
### Discussed in https://github.com/scalar/scalar/discussions/6759

<sup>Originally posted by **jackadrianglass** September  5, 2025</sup>
I just integrated Scalar into our server at work. We use [msgpack](https://msgpack.org/) as our serialization format for messages since we tend to send large amounts of data over the wire. I was hoping to have Scalar be able to encode and decode msgpack directly in the UI.

My current workflow is
* Serialize the body of messages using an external tool
* Send the message to the server
* Download the binary payload
* Decode that file with an external tool

This flow works for me but it would be much easier for the rest of my team if they could do it directly in the UI.

I'd be happy to help implement this one if someone points me in the right direction for this. My thoughts are either
* Have Scalar directly support the `application/msgpack` type. It's a self describing format with support from many different languages so seems like a good fit for this project
* Add support in the plugin system to have custom encoding/decoding of messages so that other types can be implemented</div>

DemonHa:
Fixes: #6846

Currently we only have hardcoded response previews and decoders for different content types...

This PR:
- Extends `ClientPlugin` with a `responseBody` field that lets plugins register custom decoders and renderers for arbitrary content types (e.g. MessagePack, CBOR, Protobuf)
- Adds `ResponseBodyHandler` type with mutually exclusive `rawComponent`/`language` options and support for `decode`, `previewComponent`, and wildcard MIME matching

## Checklist

- [x] I added a changeset. 
- [x] I added tests.

---

> [!NOTE]
> **Medium Risk**
> Changes response decoding/rendering paths to allow plugins to override handling based on MIME type, which could affect how bodies are parsed and displayed across many responses (especially when `Content-Type` is missing). Surface area is moderate but covered by new unit tests and docs.
> 
> **Overview**
> Adds **API client plugin support for custom response body content types** by extending `ClientPlugin` with `responseBody` handlers (`ResponseBodyHandler`) that can decode raw bytes and/or supply custom raw/preview Vue components with wildcard MIME matching.
> 
> Wires plugin handlers through request execution and UI: `sendRequest` now accepts `plugins`, resolves an effective fallback content type when headers are missing, selects a matching handler, and lets `decodeBuffer` run plugin decoders; `ResponseBody` uses the same handler to control raw/preview rendering and language selection. Includes new tests for handler resolution and fallback behavior, plus updated API client and plugin documentation and a changeset for minor bumps to `@scalar/api-client` and `@scalar/oas-utils`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f3365f705a0e39c45101919fbaee18e88ba983d4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


DemonHa:
@amritk @hanspagel Should we ship ready-made plugins for popular content types (msgpack, protobuf, CBOR, etc.) so users can just import and register them instead of writing their own decode logic?

DemonHa:
@amritk we would need a way to propagate the plugins to the client from api-reference config.. should we create a new config entry?

documentation/guides/app/plugins.md
hanspagel, line 1: sooo I knew this will come, and here it is:

this whole docs section is about the app, not the api-client package
we can't put this here
and as close to no one is actually using the api-client package on its own, I’d rather put in the api-reference plugin section (documentation/plugins.md).

hanspagel (changes requested):
nice work, I can fix the documentation (or you do, whoever gets to it first)

documentation/guides/app/plugins.md
hanspagel: done

hanspagel (approved):
good to go

packages/api-client/src/v2/blocks/operation-block/helpers/send-request.ts
```diff
@@ -244,7 +253,9 @@ const buildStandardResponse = async ({
   const clonedResponse = response.clone()
   const arrayBuffer = await clonedResponse.arrayBuffer()
   const responseType = contentType ?? 'text/plain;charset=UTF-8'
-  const responseData = decodeBuffer(arrayBuffer, responseType)
+  const mimeEssence = parseMimeType(responseType).essence
+  const pluginHandler = resolveResponseBodyHandler(mimeEssence, plugins)
+  const responseData = await decodeBuffer(arrayBuffer, responseType, pluginHandler)
```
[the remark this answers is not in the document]
hanspagel: okay mister big brain

amritk (approved):
Nice! Seems pretty good, did not test but we can adapt this api as people use it. Ensure its in beta so it may change

PR #8866, opened 2026-04-16
feat(components): update and align header buttons

hwkr:
Updates the header buttons to better align with our specs. See snapshots for before and after.

## Checklist

- [x] I added a changeset. 
- [x] I added tests.


hanspagel (approved):
🙏

PR #8867, opened 2026-04-16
feat(api-client): support header for web layout

hwkr:
Updates the client to support a header in the web layout.

- Hides app menu for the web layout
- Makes the search always visible on web
- Fixes some layout issues

CleanShot 2026-04-15 at 23 10 19@2x

## Checklist

- [x] I added a changeset. 
- [ ] I added tests.

### DIFF
--- a/packages/api-client/src/v2/components/sidebar/Sidebar.vue
+++ b/packages/api-client/src/v2/components/sidebar/Sidebar.vue
@@ -122,7 +122,9 @@ const handleSelectItem = (id: string) => {
           <div
             class="bg-sidebar-b-1 z-1 flex flex-col gap-1.5 px-3 pb-1.5"
             :class="{ 'max-md:pt-12': layout !== 'modal' }">
-            <div class="flex items-center justify-between">
+            <div
+              v-if="layout !== 'web'"
+              class="flex items-center justify-between">
               <!-- Desktop gets the workspace menu here  -->
               <SidebarMenu
                 v-if="layout !== 'modal'"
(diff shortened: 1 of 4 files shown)


packages/api-client/src/v2/components/sidebar/Sidebar.vue
```diff
@@ -122,7 +122,9 @@ const handleSelectItem = (id: string) => {
           <div
             class="bg-sidebar-b-1 z-1 flex flex-col gap-1.5 px-3 pb-1.5"
             :class="{ 'max-md:pt-12': layout !== 'modal' }">
-            <div class="flex items-center justify-between">
+            <div
+              v-if="layout !== 'web'"
```
hwkr, line 126: This hides the ScalarMenu on web

packages/api-client/src/v2/components/sidebar/Sidebar.vue
```diff
@@ -122,7 +122,9 @@ const handleSelectItem = (id: string) => {
...
+              class="flex items-center justify-between">
               <!-- Desktop gets the workspace menu here  -->
               <SidebarMenu
```
hwkr, line 129: @DemonHa @amritk  I tried to remove this so the app could just inject the whole menu but there was a lot of logic tied to it. Would probably still make sense to let scalar-app inject the whole menu though

amritk: yea it should

PR #8875, opened 2026-04-16
fix(server-side-rendering): serialize functions for hydration

hanspagel:
## Summary
- preserve top-level function config values in SSR hydration output to match client-side rendering
- throw when nested functions cannot be serialized safely instead of silently dropping them
- refactor SSR hydration serialization helpers for clearer formatting and indentation handling

## Testing
- pnpm --filter @scalar/server-side-rendering test

---

> [!NOTE]
> **Medium Risk**
> Touches inline hydration script generation and function serialization, which is XSS-sensitive and could impact SSR hydration if escaping or formatting is incorrect. Scope is contained and covered by new tests, but the output format/behavior changes may affect consumers relying on previous function-stripping behavior.
> 
> **Overview**
> SSR hydration config serialization now emits a JS object literal that **preserves top-level function properties** and **top-level arrays containing functions**, instead of stripping them, to better match client behavior.
> 
> The serializer was refactored into helpers that (1) escape function source to prevent `<script>` breakouts, (2) escape/quote dynamic property keys, and (3) **throw** when nested functions are encountered rather than silently dropping them. Tests were expanded to cover these new behaviors and injection edge cases, and changesets bump `@scalar/server-side-rendering` with patch releases.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d03aaef2c5127402953e1de27b285c3a88b52416. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/fair-lobsters-turn.md
+++ b/.changeset/fair-lobsters-turn.md
@@ -1,0 +1,5 @@
+---
+'@scalar/server-side-rendering': patch
+---
+
+Preserve top-level function config values in SSR hydration output and throw when nested functions cannot be serialized safely.
--- a/.changeset/soft-pears-protect.md
+++ b/.changeset/soft-pears-protect.md
@@ -1,0 +1,5 @@
+---
+'@scalar/server-side-rendering': patch
+---
+
+fix(server-side-rendering): escape dynamic hydration property keys
--- a/packages/server-side-rendering/src/ssr.test.ts
+++ b/packages/server-side-rendering/src/ssr.test.ts
@@ -1,7 +1,13 @@
 // @vitest-environment node
 import { describe, expect, it, vi } from 'vitest'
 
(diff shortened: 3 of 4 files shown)


marclave (approved):

amritk (commented):
Are you sure you want to do it yourself? Before we just used unhead which already exists in the html api

```js
// On render, inject server data into the client
    useHead({
      script: [
        {
          children: `window.__SCALAR__ = ${JSON.stringify(ssrContext?.payload?.data ?? {})}`,
        },
      ],
    })
```

but I guess before we never had the workspace store, ssr context was serializable so your mileage may vary!

hanspagel:
@amritk TIL, I created an issue for it

amritk:
Yea I had to hunt way back for it LOL, back when we used to support SSR

PR #8880, opened 2026-04-16
feat(api-client): add resizable test request modal panes

pollotherunner:
## Summary
- make the Test Request modal container resizable
- add a draggable splitter to resize Request and Response panes in modal layout
- keep a visible divider line while preserving hover/focus affordance
- add unit coverage for modal splitter presence by layout

## Testing
- pnpm --filter @scalar/api-client build
- pnpm --filter @scalar/api-client exec vitest --run src/v2/components/modals/ModalClientContainer.spec.ts src/v2/blocks/operation-block/OperationBlock.test.ts

## Ticket
N/A

### DIFF
--- a/.changeset/odd-apples-clean.md
+++ b/.changeset/odd-apples-clean.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
+
+add resize support for the test request modal container and its internal request/response split view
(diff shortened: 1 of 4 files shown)


hanspagel:
Ahh, great idea! I think we actually had this a while ago. It's a bit out of scope for now, though, and as it seems to require a decent amount of work to make it really good, I’d tend to close it for now.

PR #8827, opened 2026-04-13
chore(api-client): code deletion part 2

amritk:
Removes *MOST* of the old code from client v1, we still use a few things that we can fix in a later PR

Next up is oas-utils

## Checklist

- [x] I explained why the change is needed.
- [x] I added a changeset. 
- [ ] I added tests.
- [ ] I updated the documentation.

---

> [!NOTE]
> **Medium Risk**
> Removes a large set of legacy `@scalar/api-client` entrypoints, components, and playgrounds; any downstream imports still referencing v1 paths will break at build time. Runtime risk is otherwise low since this is primarily deletion and type tightening.
> 
> **Overview**
> **Removes most remaining API Client v1 codepaths.** This deletes legacy client components (e.g. address bar, command palette, import collection flow, search, data-table utilities), v1 playgrounds, and trims `@scalar/api-client` package `exports`/dependencies to focus on the v2 feature modules.
> 
> Also updates tooling to match the new public surface (knip workspace entry list), makes a small border class tweak in `@scalar/agent-chat`’s `ServerSelector`, and tightens `HttpMethod` typing to use `@scalar/helpers` method types/`objectEntries`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit db739638cafc86ff3495be7021c2c5a87efe2b67. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #8884, opened 2026-04-16
fix(api-reference): restore position: fixed on .section-flare

### LINKED ISSUE #8883: Regression: `.section-flare` reserves 100vh at top of docs in 1.52.2 (breaks `kepler`, `bluePlanet`, `mars` themes)
## Summary

`@scalar/api-reference@1.52.2` (published 2026-04-16) ships without the `.section-flare { position: fixed; top: 0; right: -50dvw }` rule that was present in the compiled bundle of `@1.52.1`. As a result, the decorative flare — which multiple first-party themes size at `100vw × 100vh` — falls into normal document flow and pushes all content down by a full viewport height.

## Affected themes (first-party presets)

Reproduced on 1.52.2 with no custom CSS, using the same OpenAPI document across themes:

| Theme        | intro.y (px) | `.section-flare` computed position | Visibly broken? |
|--------------|-------------:|-----------------------------------|:---------------:|
| `kepler`     |          467 | `static`                          | yes — 100vh empty space |
| `bluePlanet` |          467 | `static`                          | yes — 100vh empty space |
| `mars`       |          467 | `static`                          | yes — 100vh empty space |
| `deepSpace`  |           50 | `absolute` (theme self-declares)  | no              |
| `default`, `alternate`, `moon`, `purple`, `solarized`, `saturn`, `laserwave`, `none` | 50 | `static` | no (these themes don't give the flare a height) |

On `@1.52.1` the same pages render with `.section-flare` computed as `position: fixed` and `intro.y = 50` across every theme.

## Root cause

`kepler.css`, `bluePlanet.css`, and `mars.css` set `height: 100vh` (or tall fixed heights) on `.section-flare` without declaring `position`. The component's scoped CSS (`SectionFlare.vue`) sets `top: 0; right: 0; pointer-events: none` but also omits `position`. In 1.52.1 the compiled bundle included a fallback `.section-flare { position: fixed; top: 0; right: -50dvw }` that satisfied those themes' positioning contract. That rule is missing from the 1.52.2 bundle.

Bundle diff (`dist/browser/standalone.js`):

```diff
  .section-flare[data-v-…]{pointer-events:none;top:0;right:0}
- .section-flare{position:fixed;top:0;right:-50dvw}
```

No source files were modified between releases in a way that would intentionally change this:
- `packages/api-reference/src/components/SectionFlare/SectionFlare.vue` — last touched 2025-07-21
- `packages/themes/src/presets/kepler.css` — last touched 2025-08-27

The drop appears to be an unintended side effect of the build/bundling changes between 1.52.1 and 1.52.2.

## Reproduction

```html
<!doctype html>
<html><body style="margin:0"><div id="app"></div>
<script src="https://cdn.jsdelivr.net/npm/@scalar/api-reference@1.52.2"></script>
<script>
  Scalar.createApiReference("#app", {
    url: "https://petstore3.swagger.io/api/v3/openapi.json",
    theme: "kepler"
  });
</script></body></html>
```

The page renders with a full viewport of blank space before the API title.

Swap `@1.52.2` → `@1.52.1` and the empty space disappears.

## Suggested fix

The component is the right place to establish the positioning invariant, since the existing scoped `top: 0; right: 0; pointer-events: none` already implies a positioned element. One-line change in `packages/api-reference/src/components/SectionFlare/SectionFlare.vue`:

```diff
 <style scoped>
 .section-flare {
+  position: fixed;
   top: 0;
   right: 0;
   pointer-events: none;
 }
 </style>
```

This restores the 1.52.1 rendering exactly (scoped `right: 0` already won on specificity in 1.52.1 as well, so the effective styles are unchanged). Themes that self-declare `position` (`deepSpace`, `elysiajs`) are unaffected.

PR coming up.

zgcalebp:
## Summary

Fixes #8883 — `@scalar/api-reference@1.52.2` drops the `.section-flare { position: fixed; top: 0; right: -50dvw }` rule from the compiled bundle, so the decorative flare falls into normal flow and pushes all content down by 100vh on any theme that sizes the flare at `100vh` (`kepler`, `bluePlanet`, `mars`).

## Fix

One-line addition to the component's scoped CSS. The existing `top: 0; right: 0; pointer-events: none` declarations already imply a positioned element — this restores `position: fixed` so the layout contract that themes rely on is preserved inside the component itself, independent of whatever gets emitted into the final compiled bundle.

```diff
 <style scoped>
 .section-flare {
+  position: fixed;
   top: 0;
   right: 0;
   pointer-events: none;
 }
 </style>
```

Effective styles in 1.52.1 had scoped `right: 0` winning on specificity over the non-scoped `right: -50dvw`, so this produces the same rendering as 1.52.1 without needing to reintroduce the non-scoped rule.

## Verification

Rendered an OpenAPI doc with vanilla `@scalar/api-reference` (no custom CSS) on 1.52.2 before and after the patch, across all first-party themes:

| Theme        | `intro.y` on 1.52.2 (before) | `intro.y` with patch | `.section-flare` position (after) |
|--------------|:----------------------------:|:--------------------:|:---------------------------------:|
| `kepler`     | 467                          | 50                   | `fixed`                           |
| `bluePlanet` | 467                          | 50                   | `fixed`                           |
| `mars`       | 467                          | 50                   | `fixed`                           |
| `deepSpace`  | 50 (already OK)              | 50                   | `absolute` (unchanged — theme self-declares) |
| other themes | 50 (already OK)              | 50                   | `fixed` (was `static` but had no height, so no visible change) |

`deepSpace` and `elysiajs` self-declare `position: absolute` on `.section-flare`; with `fixed` in the scoped CSS, the scoped rule now wins on specificity. I confirmed visually that `deepSpace` still renders identically — its flare item children are absolutely positioned relative to the flare, which is fine under either `position`.

## Changeset

Added `.changeset/fix-section-flare-position.md` as a patch for `@scalar/api-reference`.

## Test plan

- [ ] Local render of `kepler`, `bluePlanet`, `mars`, `deepSpace`, `default` themes looks correct
- [ ] Existing api-reference unit / E2E tests pass


hwkr:
Thanks for the PR @zgcalebp! Going to dig in quick to why it was removed in the first place and see what's up.

hwkr:
Okay so I tracked it down to some unscoped CSS that was deleted in [PR 8827](https://github.com/scalar/scalar/pull/8827/changes#diff-cbfe52a13af47832eccedcde76c078ef29dfc494a0c59cd2dce95ba058ab216aL397-L401) (cc @amritk) and was (as a side affect) setting this to fixed. I think this PR should work but I want to do some testing of it quick. I'm also going to see about adding snapshots so we don't break stuff like this again.

hwkr (approved):
Added and then updated the tests, fix looks good!

hanspagel:
thanks to both of you

PR #8888, opened 2026-04-17
fix(workspace-store): resolve selected security schemes using selectedIndex

### LINKED ISSUE #5629: Only the first security scheme is provided when multiple are enabled
### What happens?

I have an app where both `Basic` and `Bearer` auth are used on various endpoints. When I enable both in the Scalar UI only one is actually sent to my application. This makes it slightly confusing for users where they get `401` responses back even though they have valid credentials entered.

When both `httpBasic` and `httpBearer` security schemes are provided both should be set in the request headers.

### OpenAPI Document

https://sandbox.scalar.com/e/AGRo3

DemonHa:
Fixes: #5629

## Summary

- Fix `getSecuritySchemes` to accept a single `SecurityRequirementObject` instead of the full array, and update `getRequestExampleContext` to pass only the actively selected requirement using `selectedIndex` — previously, all security requirements were flattened into `selectedSchemes`, ignoring which one the user actually selected.
- Change `buildRequest` to use `headers.append` / `securityQueryParams.append` instead of `.set` so that multiple security schemes with the same header or query parameter name are preserved rather than silently overwritten.

## Checklist

- [x] I added a changeset. 
- [x] I added tests.

---

> [!NOTE]
> **Medium Risk**
> Touches auth selection and request construction paths; behavior changes could alter which credentials are applied and how repeated auth values are sent, though changes are targeted and covered by new tests.
> 
> **Overview**
> Fixes security scheme resolution to only hydrate the *actively selected* security requirement (via `selectedIndex`) instead of flattening all `selectedSchemes`, updating `getSecuritySchemes` to accept a single `SecurityRequirementObject` and adjusting call sites (`getRequestExampleContext`, API reference operation filtering, and agent-chat helpers).
> 
> Updates request example building to *append* security headers/query params (using `Headers.append` and `URLSearchParams.append`) so multiple auth schemes sharing the same name don’t overwrite each other, with added tests covering selected-index behavior and repeated header/query values. Includes a patch changeset for `@scalar/workspace-store`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d2c1f84c8c921713ca6a68c08a139803897b8e43. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/strict-owls-rhyme.md
+++ b/.changeset/strict-owls-rhyme.md
@@ -1,0 +1,5 @@
+---
+'@scalar/workspace-store': patch
+---
+
+fix: resolve selected security schemes using selectedIndex
--- a/packages/agent-chat/src/helpers.ts
+++ b/packages/agent-chat/src/helpers.ts
@@ -39,7 +39,7 @@ function getSecurityFromDocument(
     securityRequirements,
   )
 
-  return getSecuritySchemes(mergedSecurity, selectedSecurity.selectedSchemes)
+  return getSecuritySchemes(mergedSecurity, selectedSecurity.selectedSchemes[selectedSecurity.selectedIndex] ?? {})
 }
 
 /** Generate document settings from workspace store */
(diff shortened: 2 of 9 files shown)


hanspagel (approved):
👍

PR #8892, opened 2026-04-17
fix(api-client): allow body on get 

amritk:
## Problem

We disallow body on get as it doesn't work in the browser. We also create a new Request a few places which doesn't allow this.

## Solution

Instead we create a safeRequest with a special header when using get with a body. This will then be read on the electron side and piped through undici instead.

TBH im not a fan of this, maybe we can just use harRequest everywhere and just convert it back to a fetch payload right before sending.

## Checklist

- [x] I added a changeset. 
- [x] I added tests.

---

> [!NOTE]
> **Medium Risk**
> Changes how requests are constructed and how the displayed HTTP method is derived, introducing a sentinel header + method rewrite path that could affect request semantics and intermediaries if the header isn’t stripped as expected.
> 
> **Overview**
> Enables replaying/sending *GET-with-body* requests in Electron without violating the Fetch spec by introducing `createSafeRequest` (rewrites GET-with-body to POST and stores the intended method in `x-scalar-original-method`).
> 
> Request construction paths (`buildRequest`, `harToFetchRequest`) now use `createSafeRequest`, and response metadata in `sendRequest` uses `getOriginalMethod` so the UI reports the user-intended method even when the underlying `Request` was rewritten. Adds/updates tests around Electron-specific body support and the rewrite/sentinel behavior, plus a patch changeset for `@scalar/helpers`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 311687ce23a05043ec655d17bf2fcd7a81538068. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/cute-pillows-punch.md
+++ b/.changeset/cute-pillows-punch.md
@@ -1,0 +1,5 @@
+---
+'@scalar/helpers': patch
+---
+
+feat: allow body on get
--- a/packages/api-client/src/v2/blocks/operation-block/helpers/har-to-fetch-request.test.ts
+++ b/packages/api-client/src/v2/blocks/operation-block/helpers/har-to-fetch-request.test.ts
@@ -1,3 +1,4 @@
+import { X_SCALAR_ORIGINAL_METHOD } from '@scalar/helpers/http/safe-request'
 import type { HarRequest } from '@scalar/snippetz'
 import { describe, expect, it } from 'vitest'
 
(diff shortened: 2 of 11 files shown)


amritk:
closing in favor of 

https://github.com/scalar/scalar/pull/8911

PR #8906, opened 2026-04-17
fix(api-reference): preserve composed schema descriptions inline

### LINKED ISSUE #6748: Components that represent an interface do not show the description
### What happens?

When you have a component which represents an interface (i.e. uses discriminator and oneOf), the description of the interface itself is not shown in the documentation.

Image

I would expect the description of the interface to be displayed in the documentation. 

### OpenAPI Document

https://sandbox.scalar.com/e/ww-JR

alonfaraj:
## Problem

Inline response schema docs could hide parent descriptions for composed `oneOf`/`anyOf` schemas, even though the same descriptions appeared on the model page.

## Solution

Preserved composed-schema descriptions in the inline schema rendering path by:
- keeping parent descriptions during schema optimization
- preserving referenced schema descriptions when `$ref` properties also have sibling descriptions
- adding regression tests for both cases

## Checklist

- [x] I added a changeset. 
- [x] I added tests.

Fixes https://github.com/scalar/scalar/issues/6748


hanspagel (approved):
nice

amritk:
Need to rebase then re-generate screenshots

alonfaraj:
@amritk Done

marclave (approved):

PR #8907, opened 2026-04-17
fix(api-reference): update code sample when switching request body content type

### LINKED ISSUE #6990: cURL example doesn't update when switching from JSON to form-urlencoded content type
### What happens?

When switching the content type from `application/json` to `application/x-www-form-urlencoded` in the request body, the cURL example remains in JSON format instead of updating to form-encoded format.

When the `application/json` content type is removed from the OpenAPI specification and only `application/x-www-form-urlencoded` remains, the cURL example correctly displays the form-encoded data.

application/json
Image

application/json -> application/x-www-form-urlencoded

Image

sample
```
openapi: 3.1.1
paths:
  /hoge:
    post:
      requestBody:
        required: true
        content:
          application/json:
            schema:
              type: object
              properties:
                user_id:
                  type: string
            examples:
              json-sample:
                value:
                  user_id: aaaa
            schema:
              type: object
              properties:
                user_id:
                  type: string
            examples:
              form-sample:
                value:
                  user_id: bbbb
      responses:
        '200':
          description: Success
          content:
            application/json:
              schema:
                type: object
                properties:
                  message:
                    type: string
```

When switching from `application/json` to `application/x-www-form-urlencoded` content type, the cURL example should:

1. **Update the Content-Type header** from `Content-Type: application/json` to `Content-Type: application/x-www-form-urlencoded`
2. **Change the request body format** from JSON format (`{"user_id": "aaaa"}`) to form-encoded format (`user_id=bbbb`)
3. **Reflect the selected content type** in the generated cURL command, making it ready to copy and use for testing

### OpenAPI Document

_No response_

alonfaraj:
## Problem

Switching the request body content type, the code example does not update.

## Solution

Thread the selected content type from `RequestBody` up through `OperationParameters` to the layout components (`ClassicLayout` / `ModernLayout`) using `v-model`, then pass it down to `OperationCodeSample` as a prop. Inside the code sample component, a watcher resets the selected example key when the content type changes and the current example no longer exists in the new content type.

- Converted `selectedContentType` in `RequestBody.vue` from a local `ref` to a `defineModel` so it can be two-way bound
- Added `v-model:selectedContentType` plumbing through `OperationParameters.vue` → layout components
- Passed the selected content type to `OperationCodeSample` in both `ClassicLayout` and `ModernLayout`
- Added a `watch` in `OperationCodeSample.vue` to reset the example key when the content type changes
- Added unit tests for the `v-model` behavior and the layout-level sync

## Checklist

- [x] I added a changeset.
- [x] I added tests.

Fixes https://github.com/scalar/scalar/issues/6990

### DIFF
--- a/.changeset/social-rockets-buy.md
+++ b/.changeset/social-rockets-buy.md
@@ -1,0 +1,6 @@
+---
+'@scalar/api-reference': patch
+'@scalar/api-client': patch
+---
+
+Fix code sample missing body content type update
--- a/packages/api-client/src/v2/blocks/operation-code-sample/components/OperationCodeSample.vue
+++ b/packages/api-client/src/v2/blocks/operation-code-sample/components/OperationCodeSample.vue
@@ -201,6 +201,18 @@ onBeforeMount(() => {
   selectedExampleKey.value ||= Object.keys(requestBodyExamples.value)[0] ?? ''
 })
 
+/** Reset the selected example key if the content type changes and the new content type doesn't have the previously selected example */
+watch(
+  () => selectedContentType,
+  () => {
+    if (
+      !Object.keys(requestBodyExamples.value).includes(selectedExampleKey.value)
+    ) {
+      selectedExampleKey.value = Object.keys(requestBodyExamples.value)[0] ?? ''
+    }
+  },
+)
+
 /** Grab any custom code samples from the operation */
 const customCodeSamples = computed(() => getCustomCodeSamples(operation))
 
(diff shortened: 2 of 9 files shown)


DemonHa (approved):
Nice work 🚀
Didn't test it but looks good!

-----
Thank you so much for this PR 🙏

hanspagel:
tested, works great, at least for the api-reference, not so much in the api-client yet
anyway, let's get this in

PR #8701, opened 2026-04-02
feat(api-reference): show required auth and scopes in operation headers

amritk:
## Problem

Issue #5062 requests clear endpoint-level visibility for authentication requirements in references.

## Solution

- Compute operation-level required auth in `Operation.vue` using existing OpenAPI-aware security helpers.
- Extract required OAuth scopes from effective security requirements (operation overrides document).
- Render an auth-required lock icon in operation headers only (both classic and modern layouts).
- Show required scopes in the icon tooltip when scopes are present.
- Keep sidebar and operations list unchanged.
- Add focused tests to verify:
  - required vs optional auth behavior,
  - required scopes propagation to classic/modern layouts,
  - updated classic layout test fixture typing for new required props.
- Add changeset for `@scalar/api-reference`.

## Checklist

- [x] I explained why the change is needed.
- [x] I added a changeset.
- [x] I added tests.
- [ ] I updated the documentation.

## Ticket

Closes #5062

<div><a href="https://cursor.com/agents/bc-8c380f73-c38e-4cd2-a11b-ab7b5a07c221"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-8c380f73-c38e-4cd2-a11b-ab7b5a07c221"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

Issue #5062, opened 2025-03-11
Show which auth is required on references


amritk:
### What happens?

At the very minimum we need an icon to show that auth is required on an endpoint in references.

### What did you expect to happen?

For bonus points we can also show which auth + scopes are required on hover?

bislerium:
is this still in progress @amritk ?
For now the document level security scheme declaration allows to select auth at endpoints that don't need authorization.

amritk:
Yea just keep bumping it. We're just crushing a big refactor at the moment but will get back to regularly scheduled programming after

dominic-mallo:
hey @amritk - I think it would be nice to see the following displayed in the documentation (perhaps near the title/description).

<img width="1681" height="1291" alt="Image" src="https://github.com/user-attachments/assets/0398c656-4523-4200-8e70-4de3202c4d2a" />

devin-brenton:
My team could really use this feature as scopes are pivotal to how we manage access to our apis. There are workarounds for now but would love to see this officially supported.

adnweedon:
It would be super helpful to have scopes visible on each endpoint - maybe even being able to select scopes in the test client at the top, and have some visual thing to show that certain endpoints wouldn't be covered as you browse the docs?!

DemonHa:
looks like it's completed

<img width="1400" height="595" alt="Image" src="https://github.com/user-attachments/assets/e9fd8782-3b58-4edb-9db5-0097f4517942" />

PR #8908, opened 2026-04-17
feat(api-reference): show auth-required/optional badge next to operation path

zgcalebp:
Addresses #5062. Supersedes the stale draft #8701.

## Summary

Adds a small text chip beside each operation's path reading **`auth required`** (lock icon) or **`auth optional`** (open-lock icon). Hovering the chip surfaces a tooltip with the scheme names, types, and any required scopes — e.g. `Requires oauth2 (read:items, write:items) · apiKey`.

## Detection logic

Per maintainer guidance on #8701 — simple fallback, no borrowing from `getSecurityRequirements` / `isAuthOptional`:

```ts
const requirements = operation.security ?? document.security ?? []
```

Three states:

| State      | Condition                                                                       |
|------------|---------------------------------------------------------------------------------|
| `required` | At least one non-empty requirement, and none are `{}`                           |
| `optional` | At least one non-empty requirement AND at least one `{}` (explicit unauthenticated alternative) |
| `none`     | No requirements, or the only requirement is `{}`                                |

Operation-level `security` is a full override — `security: []` explicitly opts out of document-level auth and yields `none`.

Scheme objects are resolved through `getResolvedRef` (same helper used by 9 sibling files in `features/Operation/`). Scopes referenced by the same scheme across multiple requirements are unioned.

## Known limitation: AND-of-ANDs collapse in the tooltip

OpenAPI's `security` array is an OR-of-ANDs — each requirement object ANDs its schemes internally, and the array ORs the alternatives. This PR flattens by scheme name, so

- `[{ apiKey: [] }, { oauth2: [] }]` — "apiKey OR oauth2"
- `[{ apiKey: [], oauth2: [] }]` — "apiKey AND oauth2"

both render as `Requires apiKey · oauth2`. The chip itself and the `required / optional / none` state are correct in both cases — just the AND/OR distinction inside the tooltip is lost. Happy to replace the tooltip with a structured list if you'd rather preserve it.

## Rendering

The chip lives next to `OperationPath` in both layouts:
- **ModernLayout:** inside the `#header` slot of `OperationCodeSample`, in a small flex row with the path.
- **ClassicLayout:** inline in the endpoint-label row between `OperationPath` and the operation title.

Built on existing `@scalar/components` primitives (`ScalarTooltip` + `Badge`) and `@scalar/icons` (`ScalarIconLockSimple` / `ScalarIconLockSimpleOpen`). No new shared components beyond the operation-feature-scoped `SecurityRequirementBadge.vue`.

## Tests

- `get-required-security.test.ts` — 12 unit tests covering fallback precedence (`operation.security` overrides `document.security`, `[]` opt-out), null/undefined operation, the three state transitions, scheme resolution from `components.securitySchemes`, unresolved scheme names, scope unioning across requirements, and empty-scope-string filtering.
- `SecurityRequirementBadge.test.ts` — 3 render tests covering `required` with schemes + scopes, `optional`, and `none` (no render).

## Test plan

- [x] Unit tests (`pnpm vitest` — 15/15 passing)
- [x] Rendered locally against a FastAPI service using `HTTPBearer` — all `POST`/`PATCH`/`DELETE` operations show `auth required`, read operations show no chip, tooltip displays scheme type
- [ ] CI — build, lint, types, tests, e2e, snapshots
- [ ] Visual review from maintainers on both Modern and Classic layouts

## Note on #5062's broader scope

This PR covers the issue's core ask ("icon to show auth is required on an endpoint" + "show auth + scopes on hover"). The follow-up asks from later commenters (scope selection in the test client, per-endpoint coverage indicator against a user-declared scope set) are API-client-side features and are left as follow-ups. I've used "Addresses" rather than "Closes" so you can decide whether to keep #5062 open for those follow-ups or close it here.

## Changeset

`.changeset/security-requirement-badge.md` — `@scalar/api-reference` **minor**.


amritk:
Wow I will look into this properly later but looks like exactly what I was looking for! Many thanks!!

amritk:
Sorry I kept getting sidetracked, I will check this right now!

amritk:
alright @zgcalebp I made some updates:
- updated the optional logic
- moved the badge beside webhooks
- simplified the requirements logic

Its my understanding that as long as there is a {} present it should say optional as the spec states:
> To make security explicitly optional, an empty security requirement ({}) can be **included** in the array.

 Let me know what you think!

hwkr (approved):
Looks good! Made some updates to how we render the badge and switched it to a popover since it can have a lot of content.

Image
Image

.changeset/security-requirement-badge.md
```diff
@@ -0,0 +1,5 @@
+---
+'@scalar/api-reference': minor
+---
+
+feat(api-reference): show an "auth required" / "auth optional" badge next to each operation's path. Requirements are resolved via `operation.security ?? document.security`; hover reveals the scheme names, types, and any required scopes.
```
DemonHa, line 5: not: we don't really put the scope on the changeset

packages/api-reference/src/features/Operation/helpers/get-required-security.ts
```diff
@@ -0,0 +1,74 @@
...
+
+/**
+ * Determine whether an operation requires authentication, using `operation.security ?? document.security`
+ * as the source of truth. Operation-level `security` fully overrides document-level — including `security: []`,
+ * which explicitly opts the operation out of auth.
+ *
+ * OpenAPI encodes "auth is optional" by including an empty requirement object `{}`. Whenever `{}`
+ * appears — whether alongside real requirements or as the only entry — auth is treated as optional (state: 'optional').
+ */
+export const getRequiredSecurity = (
```
DemonHa, line 36: This function does not correctly handle multiple security pointing to the same scheme + complex auth

When you have:

```
security: [{ bearerAuth: ["a"] }, { bearerAuth: ["b"], basicAuth: [] }]
```

the output would be `[{ schemes: { bearerAuth: ["a", "b"] } }]`

which I don't this is what we want??

cc @amritk

DemonHa (changes requested):

packages/api-reference/src/features/Operation/helpers/get-required-security.ts
amritk: you are correct, I can go over once more for complex auth

zgcalebp:
Oh I like your changes! This looks great!

hanspagel:
I’m updating the snapshots

packages/api-reference/src/components/Badge/Badge.vue
```diff
@@ -16,24 +17,34 @@ const badgeStyle = computed(() =>
...
 <style scoped>
 .badge {
   color: var(--badge-text-color, var(--scalar-color-2));
   font-size: var(--scalar-mini);
+  font-weight: var(--scalar-font-medium);
   background: var(--badge-background-color, var(--scalar-background-2));
   border: var(--scalar-border-width) solid
     var(--badge-border-color, var(--scalar-border-color));
-  padding: 2px 6px;
+  padding: 3px 8px;
```
hanspagel: this changed all badges and I don't like it 😬

packages/api-reference/src/components/Badge/Badge.vue
```diff
@@ -16,24 +17,34 @@ const badgeStyle = computed(() =>
...
     <slot />
-  </div>
+  </component>
 </template>
 
 <style scoped>
 .badge {
   color: var(--badge-text-color, var(--scalar-color-2));
   font-size: var(--scalar-mini);
+  font-weight: var(--scalar-font-medium);
```
hanspagel: changes all badges

hanspagel (changes requested):
it's green! I just have two design questions to cam, will approve once we talked

hanspagel:
I’d prefer to have it on the right and without the badge style actually, thanks @cameronrohani for putting this together

I’m not sure who can work on this. I can do, but probably not today/tomorrow

image

amritk:
@hanspagel we can ask our trusty agent and see how they do

amritk:
closing this one in favor of the next incarnation!

https://github.com/scalar/scalar/pull/8970

PR #8911, opened 2026-04-17
feat(api-client): allow Body with GET requests

amritk:
## Problem

We disallow body on get as it doesn't work in the browser. We also create a new Request a few places which doesn't allow this.

## Solution

Now instead of creating a JS request, we just create a RequestPayload ([url, RequestInit]) which is the payload to Request without actually creating the request. Now we can safely pass the payload around and not run into the browser limitations of Request when we want to break the law.

## Checklist

- [x] I added a changeset. 
- [x] I added tests.

---

> [!NOTE]
> **Medium Risk**
> Replaces `Request` objects with a `[url, RequestInit]` `RequestPayload` across request building/execution, response rendering, and history serialization, which can subtly change header/body handling and abort behavior. Also alters fetch behavior in Electron vs browser via `buildSafeBodyRequest`/`window.proxiedFetch`, so regressions may appear in edge-case requests and plugins.
> 
> **Overview**
> Enables *request bodies with GET* by switching the request representation throughout the API client stack from a constructed `Request` to a `RequestPayload` tuple (`[url, RequestInit]`) that can be passed around without browser `Request` limitations.
> 
> Request execution (`sendRequest`) now consumes `RequestPayload`, uses `buildSafeBodyRequest` to strip bodies from methods that browsers reject (while allowing broader method+body support in Electron via `window.proxiedFetch`), and all related UI/state (`OperationBlock`, `ResponseBlock`, response caching) and hooks/history events now store/emit `requestPayload` instead of `request`.
> 
> HAR conversion helpers were updated accordingly: `harToFetchRequest` now returns `RequestPayload`, and `fetchRequestToHar` now serializes from `RequestPayload`/`RequestInit.body` types (FormData/URLSearchParams/text/blob/buffers) rather than cloning/reading a `Request`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 56abe6fdf4e79cd81d94e5244c59136c0ce70ccf. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/api-client/src/v2/blocks/response-block/ResponseBlock.vue
```diff
@@ -97,15 +99,30 @@ const shouldVirtualize = computed(() => {
...
-    ? [...request.headers].map((header) => ({
-        name: header[0],
-        value: header[1],
+const requestHeaders = computed(() => {
+  const headers = requestPayload?.[1]?.headers
+  if (!headers) {
+    return []
+  }
+  // headers in RequestInit can be Headers, string[][], or Record<string, string>
+  if (typeof (headers as Headers)[Symbol.iterator] === 'function') {
```
DemonHa: maybe we can use the build in header object here

`[...new Headers(headers)]`

packages/helpers/src/http/can-method-have-body.ts
```diff
@@ -1,9 +1,23 @@
...
 /** HTTP Methods which can have a body */
-const BODY_METHODS = ['post', 'put', 'patch', 'delete'] as const satisfies HttpMethod[]
-type BodyMethod = (typeof BODY_METHODS)[number]
+const BODY_METHODS = new Set(['post', 'put', 'patch', 'delete'])
 
-/** Makes a check to see if this method CAN have a body */
-export const canMethodHaveBody = (method: HttpMethod): method is BodyMethod =>
-  BODY_METHODS.includes(method.toLowerCase() as BodyMethod)
+/** HTTP Methods which can have a body when running in Electron, where the fetch implementation allows bodies on GET. */
+const ELECTRON_BODY_METHODS = new Set([...BODY_METHODS, 'get'])
```
DemonHa: for electron allow all of them to have body...

packages/api-client/src/v2/blocks/operation-block/helpers/send-request.ts
```diff
@@ -64,25 +65,31 @@ const NO_BODY_STATUS_CODES = [204, 205, 304]
...
   try {
     // Execute the request and measure duration
     const startTime = performance.now()
-    const response = await fetch(request.clone())
+
+    // We may use a custom fetch function on electron
+    const response =
+      'proxiedFetch' in window && window.proxiedFetch
+        ? await window.proxiedFetch(...requestPayload)
+        : await fetch(...requestPayload)
```
DemonHa: You want to sanitize the body here for GET requests

packages/api-reference/src/helpers/map-config-plugins.ts
```diff
@@ -39,12 +39,16 @@ export const mapConfigPlugins = (
...
-            await onBeforeRequest({
-              // We need to build the request to get the fetch `Request`
-              request: buildRequest(payload.requestBuilder, { envVariables }).request,
-              requestBuilder: payload.requestBuilder,
-              envVariables,
-            })
+            try {
+              await onBeforeRequest({
+                // We need to build the request to get the fetch `Request`
+                request: new Request(...buildRequest(payload.requestBuilder, { envVariables }).requestPayload),
```
DemonHa: Just sanitize the body, you don't want this to fail if we have a body, we still want to run the hooks

DemonHa (changes requested):

packages/api-client/src/v2/blocks/operation-block/helpers/send-request.ts
amritk: we don't need to actually, since it will only happen in proxiedFetch

packages/api-reference/src/helpers/map-config-plugins.ts
amritk: I don't think so, they should be using the request builder this is the old stuff right?

packages/api-client/src/v2/blocks/operation-block/helpers/send-request.ts
amritk: I could just do the isElectron check here instead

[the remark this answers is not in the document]
amritk: we are not making requests in non window contexts

packages/api-client/src/v2/blocks/response-block/ResponseBlock.vue
amritk: ya way better

amritk: done

packages/api-client/src/v2/blocks/operation-block/helpers/send-request.ts
DemonHa: we do cuz if you pass the body to fetch it will throw

packages/api-reference/src/helpers/map-config-plugins.ts
DemonHa: yeah but if this fails the entire thing will fail, so you want to buidl the "safe" request

amritk: ok

packages/api-client/src/v2/blocks/operation-block/helpers/send-request.ts
amritk: yea but body with get will only be passed to proxiedFetch, k leme change to electron so that it matches completely

packages/api-reference/src/helpers/map-config-plugins.ts
amritk: k done lol not sure why I didn't see that, thought requestBuilder used a different hook but it was right underneath

packages/api-client/src/v2/blocks/operation-block/helpers/send-request.ts
```diff
@@ -64,25 +67,31 @@ const NO_BODY_STATUS_CODES = [204, 205, 304]
...
   try {
     // Execute the request and measure duration
     const startTime = performance.now()
-    const response = await fetch(request.clone())
+
+    // We may use a custom fetch function on electron
+    const response =
+      isElectron() && window.proxiedFetch
+        ? await window.proxiedFetch?.(...requestPayload)
+        : await fetch(buildSafeBodyRequest(...requestPayload))
```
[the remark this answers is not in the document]
amritk: yep what we wanted

marclave (approved):

PR #8609, opened 2026-03-27
feat(api-reference): display tag groups for classic layout

DemonHa:
Fixes: #3766

Screenshot 2026-03-27 at 5 04 30 PM

## Checklist

- [x] I explained why the change is needed.
- [x] I added a changeset. 
- [x] I added tests.
- [ ] I updated the documentation.

---

> [!NOTE]
> **Low Risk**
> Low risk UI-only change that adjusts navigation rendering for `x-tagGroups` when using the classic layout and adds Playwright snapshot coverage to catch regressions.
> 
> **Overview**
> Adds support for rendering `x-tagGroups` as **visible sections in the classic layout** (instead of always flattening them), by treating tag-group entries like tags when `options.layout === 'classic'`.
> 
> Updates the classic tag section styling to visually distinguish tag groups, and adds a new Playwright visual regression test plus an OpenAPI fixture in `sources.ts` to cover tag-group sidebar and classic-page rendering.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f7a2b4f93b7fe40aefdea66720e574feb9dc9d5d. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

Issue #3766, opened 2024-11-05
bring tag groups to the classic layout


hanspagel:
Hello Scalar Team! 

At the moment Scalar supports "x-tagGroups" by grouping operations in the Sidebar.
For the the classic layout as of now tag groups are not being rendered. 

Imo - for more complex APIs the format of 

It would be nice if tag groups would put a spoiler around the tags which are grouped inside of them in the classic view 

Similar to but better designed than: 
![image](https://github.com/user-attachments/assets/8d172208-3fde-46e1-9054-f73a9db1f2cc)

It would honestly be also nice if there was a way to do nested tags. E.g. making it possible to split tag names by some configureable delimiter e.g. "scope.my.tag.better" or using something more flexible than the tagGroups approach.

_Originally posted by @Aderinom in https://github.com/scalar/scalar/discussions/3765_

philsturgeon:
OpenAPI v3.2 is brnging `tags.kind` which can help split tags by purpose, and also `tags.parent` to allow structure for tags without relying on x-tagGroups: https://github.com/OAI/OpenAPI-Specification/blob/v3.2-dev/src/oas.md#tag-object

PR #8914, opened 2026-04-19
feat(api-reference): collapsible x-tagGroups in modern layout

TWulfZ:
## Summary

Adds collapsible `x-tagGroups` support to the modern sidebar layout. Each tag group now renders as a section with an expand/collapse toggle. Groups default to expanded to preserve the previous flat rendering for existing users.

Classic layout behavior is unchanged (shipped in #8609). This PR is the modern-layout counterpart.

## Before / After

| Before | After |
| --- | --- |
| antes_cambios | final |

## What changed

- **`ScalarSidebarSection`** (`@scalar/components`): new opt-in `collapsible` prop with `open` v-model and `toggle` event. When collapsible, the header becomes a real button with a chevron (`ScalarSidebarGroupToggle`) and children nest one level deeper to reflect hierarchy. Non-collapsible sections keep the original flat layout — no regression for existing consumers.
- **`SidebarItem`** (`@scalar/sidebar`): tag groups (`item.isGroup === true`) are now wired to the collapsible Section. Clicking the header emits `toggleGroup` and syncs with the existing sidebar state.
- **`ApiReference.vue`** (`@scalar/api-reference`): on mount, walks the sidebar tree once and expands every `x-tagGroups` entry so users who never interacted with the sidebar still see every tag by default.

Search integration already works out of the box: `create-search-index.ts` indexes tag groups, and `sidebarState.setSelected` recursively expands parents, so searching for a tag inside a collapsed group auto-expands it.

## Non-goals

- Nested tag hierarchy via `parent` (tracked in #6866 — separate spec).
- Cross-session persistence of collapse state.
- Custom animations — relies on `v-show` + existing sidebar styles.

## Test plan

- [x] Unit tests for `ScalarSidebarSection` collapsible mode (default header stays non-interactive; collapsible emits `toggle`, swaps element to `button`, adds `aria-expanded`, and hides items when closed).
- [x] Unit tests for `SidebarItem` tag-group wiring (`collapsible` + `open` props, `toggleGroup` emission).
- [x] `@scalar/components` full suite passes (528/528).
- [x] `@scalar/sidebar` full suite passes (200/200).
- [x] Manual visual check against the existing `Tag Groups` fixture in `test/utils/sources.ts`.
- [ ] Playwright snapshot for `modern-collapse` — intentionally deferred; maintainer CI can regenerate a cross-platform baseline consistent with the existing `tag-groups.e2e.ts` fixtures.

## Accessibility

- Header renders as `<button>` with `aria-expanded` synced to open state.
- `ScalarSidebarGroupToggle` provides an `sr-only` label (`Open/Close Group`).
- Keyboard `Enter` / `Space` toggle the group as expected on a native button.

## Related

- Precedent PR (classic layout): #8609
- Tracking issue: #3766
- Out-of-scope discussion (nested groups): #6866

### DIFF
--- a/.changeset/collapsible-tag-groups-modern.md
+++ b/.changeset/collapsible-tag-groups-modern.md
@@ -1,0 +1,13 @@
+---
+'@scalar/api-reference': minor
+'@scalar/components': minor
+'@scalar/sidebar': patch
+---
+
+feat(api-reference): collapsible `x-tagGroups` in modern layout
+
+The modern sidebar now renders each `x-tagGroups` entry as a collapsible
+section with an expand/collapse toggle. Groups default to expanded so the
+existing flat rendering is preserved for users who did not interact with the
+sidebar. `ScalarSidebarSection` gained an opt-in `collapsible` prop plus an
+`open` model and `toggle` event.
--- a/packages/api-reference/src/components/ApiReference.vue
+++ b/packages/api-reference/src/components/ApiReference.vue
@@ -355,6 +355,25 @@ const setChildrenOpen = (items: TraversedEntry[]): void => {
   })
 }
 
+/**
+ * Open all `x-tagGroups` by default so the modern sidebar preserves the
+ * previous flat rendering when collapsible sections are introduced.
+ */
+const openTagGroupsByDefault = (items: TraversedEntry[]): void => {
+  items.forEach((item) => {
+    if (
+      item.type === 'tag' &&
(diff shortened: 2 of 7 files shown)


hanspagel:
Thanks for the PR, I am not convinced we really want this. Real-world OpenAPI documents often have really long titles and the few extra pixels are really valuable for those. :(

TWulfZ:
Finally we got this 😑
image

PR #8921, opened 2026-04-20
fix(workspace-store): handle nullable numeric empty-string defaults

### LINKED ISSUE #8721: .net 10 webapi, int? with DefaultValue(null) shows as empty string in scalar version 2.13.18
### What happens?

<PackageReference Include="Scalar.AspNetCore" Version="2.13.18" />

    public class LoginRqst
    {
        [DefaultValue(null)]
        public string? Username { get; set; } = null!;
        [DefaultValue(null)]
        public string? Password { get; set; } = null!;
        [DefaultValue(null)]
        public int? Passcode { get; set; }
    }

showed as 
{
  "username": null,
  "password": null,
  "passcode": ""
}

Image

expecting

{
  "username": null,
  "password": null,
  "passcode": null
}

### OpenAPI Document

_No response_

hanspagel:
## Problem

`[DefaultValue(null)]` on nullable numeric fields in .NET can surface in the document as an empty string default for a nullable numeric schema. Scalar then renders request-body examples with `""` instead of `null` for those fields.

## Solution

- Keep the runtime fix scoped to `getExampleFromSchema`.
- Normalize empty-string defaults to `null` when the schema allows `null` and does not allow `string`.
- Keep existing empty-string behavior for nullable strings.
- Add regression tests in `get-example-from-schema.test.ts` only.
- Follow-up CI fix: tighten the helper argument type to the schema primitive union so `workspace-store` TypeScript build succeeds.

## Checklist

- [x] I added a changeset. 
- [x] I added tests.

## Ticket

Fixes #8721

<div><a href="https://cursor.com/agents/bc-85a7170e-9ccc-4586-bb28-5037ee2befa3"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-85a7170e-9ccc-4586-bb28-5037ee2befa3"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/packages/workspace-store/src/request-example/builder/helpers/get-example-from-schema.ts
+++ b/packages/workspace-store/src/request-example/builder/helpers/get-example-from-schema.ts
@@ -217,6 +217,59 @@ const mergeExamples = (baseValue: unknown, newValue: unknown): unknown => {
 
 type CompositionKeyword = 'anyOf' | 'oneOf'
 
+const schemaIncludesType = (
+  schema: SchemaObject,
+  targetType: string,
+  seen: WeakSet<object> = new WeakSet(),
+): boolean => {
+  const rawSchema = getSchemaCacheTarget(schema)
+
+  if (seen.has(rawSchema)) {
+    return false
+  }
+  seen.add(rawSchema)
+
+  if ('type' in schema) {
+    if (Array.isArray(schema.type)) {
+      return schema.type.includes(targetType)
+    }
+
+    if (schema.type === targetType) {
+      return true
+    }
+  }
+
+  const compositions = [schema.oneOf, schema.anyOf, schema.allOf]
+  for (const composition of compositions) {
+    if (!Array.isArray(composition)) {
+      continue
+    }
+
+    for (const item of composition) {
+      const resolved = resolve.schema(item)
+      if (resolved && schemaIncludesType(resolved, targetType, seen)) {
+        return true
+      }
+    }
+  }
+
+  return false
+}
+
+const normalizeSchemaDefault = (schema: SchemaObject): unknown => {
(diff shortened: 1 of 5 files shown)


marclave (approved):
gunna wait for @amritk & @DemonHa to take a look

amritk:
Forgive me if I misunderstand this, but is this something we want to do on ingress instead? maybe fire me a message hans

hanspagel:
actually, we might only want to do this when we generate the examples 🤔 let me think

EDIT: yes, that seems cleaner

packages/workspace-store/src/request-example/builder/helpers/get-example-from-schema.ts
```diff
@@ -216,6 +216,60 @@ const mergeExamples = (baseValue: unknown, newValue: unknown): unknown => {
...
+        return true
+      }
+    }
+  }
+
+  return false
+}
+
+const normalizeSchemaDefault = (schema: SchemaObject): unknown => {
+  if (schema.default !== '') {
```
DemonHa: this is way too specific.. I think we want to handle any type mismatch here not just empty strings

DemonHa (changes requested):

PR #8935, opened 2026-04-20
feat(openapi-parser): pass resolved object to `onDereference`

gary5:
## Problem

To track the source of dereferenced object. It is useful help us to optimize for sharing components between endpoints.

## Solution

Pass the resolved source to onDereference callback with result schema and ref.

## Checklist

- [ v ] I explained why the change is needed.
- [ v ] I added a changeset. 
- [ v ] I added tests.
- [ v ] I updated the documentation.

### DIFF
--- a/.changeset/twelve-geckos-make.md
+++ b/.changeset/twelve-geckos-make.md
@@ -1,0 +1,5 @@
+---
+'@scalar/openapi-parser': minor
+---
+
+feat: pass resolved object to onDereference callback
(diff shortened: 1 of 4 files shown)


hanspagel:
thanks! can you describe your use case a bit more?

gary5:
hello hans,

we use this tool to parse openapi doc to build a graph.
it is helpful to review whole api design especially in a large bundled api doc.
a general purpose is to identify duplicated components and change them to reference to a single one.

hence, we need any approach to collect reused components, and the relations between dereferenced object and its source. after investigated, the onDereference callback is a simplest way to get dereferenced object's reference source.

hanspagel:
can you just run "pnpm format" to make CI pass?

hanspagel:
ah wait, I’ll do it

hanspagel (approved):

PR #8939, opened 2026-04-20
test(api-reference): add end-to-end theme snapshot tests

hwkr:
Adds snapshot tests for the themes to avoid issues like #8883.

## Checklist

- [ ] I added a changeset.  ← Only adds tests
- [x] I added tests.


packages/api-reference/test/snapshots/themes.e2e.ts.snapshots/bluePlanet-classic-dark.png
hwkr, line 1: This should be fixed by #8884

amritk (approved):
Sweet, when do they run? At some point we may need to shard our e2e tests like the unit ones

hwkr:
I think they always run right now (because refs has so many dependencies that can break the UI). They should already be configured to run concurrently on CI (each test runs it's own example server and tests that) so sharding shouldn't be too hard?

PR #8943, opened 2026-04-21
feat(api-client): allow passing in a custom fetch function

amritk:
## Problem

We need to be able to override fetch on the electron app.

## Solution

Instead of some hacky overrides, lets actually just allow passing in a customFetch function as the store already supports this, we just need to add support to send-request. 

Also cleaned up some other options while I was there.

## Checklist

- [x] I added a changeset. 
- [x] I added tests.

---

> [!NOTE]
> **Medium Risk**
> Changes the core request dispatch path by allowing a caller-provided `customFetch`, including Electron-specific argument handling; mistakes here could break all outgoing requests across app/modal/runner flows. Risk is mitigated by added unit tests covering default/custom fetch behavior and Electron vs browser dispatch.
> 
> **Overview**
> Adds a new `customFetch` option that can be passed through API client *app*, *modal*, *runner*, and the React wrapper, allowing callers (notably Electron) to override how network requests are executed.
> 
> Refactors option typing by introducing a shared `ApiClientOptions` type (replacing `ApiClientModalOptions`) and threads `options` through app state/route props, while updating `sendRequest` to accept `customFetch` and choose between `customFetch(buildSafeBodyRequest(...))` (browser) vs `customFetch(...requestPayload)` (Electron). Comprehensive tests were added/updated to validate fetch selection, error propagation, and Electron vs browser request argument/body behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4aa5641eecba67ec852926f640a26a6cdf8b65a7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/api-client/src/v2/blocks/operation-block/helpers/send-request.ts
```diff
@@ -86,11 +92,10 @@ export const sendRequest = async ({
...
 
-    // We may use a custom fetch function on electron
-    const response =
-      isElectron() && window.proxiedFetch
-        ? await window.proxiedFetch?.(...requestPayload)
-        : await fetch(buildSafeBodyRequest(...requestPayload))
+    // In electron we allow GET requests to have a body
+    const response = isElectron()
+      ? await customFetch(...requestPayload)
+      : await customFetch(buildSafeBodyRequest(...requestPayload))
```
DemonHa, line 98: cool 😎

packages/api-client/src/v2/features/modal/Modal.vue
```diff
@@ -21,19 +25,7 @@ export type ModalProps = {
...
-        | 'authentication'
-        | 'baseServerURL'
-        | 'hideClientButton'
-        | 'hiddenClients'
-        | 'oauth2RedirectUri'
-        | 'servers'
-      >
-    >
-  >
+  options: MaybeRefOrGetter<ApiClientOptions>
```
DemonHa, line 28: This makes wayy more sense

packages/api-client/src/v2/blocks/operation-block/helpers/send-request.ts
```diff
@@ -52,6 +52,9 @@ export type ResponseInstance = Omit<Response, 'headers'> & {
 /** HTTP status codes that should not include a response body */
 const NO_BODY_STATUS_CODES = [204, 205, 304]
 
+/** Custom fetch function signature compatible with the global fetch API */
+export type CustomFetch = (input: string | URL | globalThis.Request, init?: RequestInit) => Promise<Response>
```
DemonHa: should you just do typeof fetch...

looks like in v6 typeof fetch is: `(input: RequestInfo | URL, init?: RequestInit) => Promise<Response>`

DemonHa (approved):
Nicee 🔥

packages/api-client/src/v2/blocks/operation-block/helpers/send-request.ts
amritk: ya ur right should work

PR #8944, opened 2026-04-21
feat(api-reference): add anchors and copy link buttons to allOf body parameters

wilhelmer:
## Problem

Request body parameters composed via `allOf` are rendered as a merged group inside a bordered box. 

These parameters have no anchor IDs and no copy link buttons, even though top-level body parameters do.

### Regular body parameters

SCR-20260421-kjdh

### Body parameters composed via allOf

SCR-20260421-kioz

## Root cause

When a schema uses `allOf`, SchemaComposition.vue merges the sub-schemas and passes the result to a nested `<Schema>`, which adds one extra level of nesting. This pushed those properties to `level 2`, while `shouldHaveLink` only enabled anchors for `level <= 1`. So `WithBreadcrumb` never received a breadcrumb for those properties, meaning no `id` anchor and no copy button were rendered.

## Solution

- SchemaProperty.vue — raise the `shouldHaveLink` threshold from `level <= 1` to `level <= 2`, so properties inside `allOf` groups receive their breadcrumb and the resulting anchor + copy-link button.
- SchemaProperty.vue — add a scoped CSS rule for `.property--level-2` that shifts the copy-link button further left (`left: -2rem`). Without this, the button's default position (`-left-4.5`) lands on top of the left border of the `allOf` group box. The extra offset moves it outside the box, consistent with how the button appears for level-1 properties outside any box.

### DIFF
--- a/.changeset/tricky-parrots-move.md
+++ b/.changeset/tricky-parrots-move.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
+feat(api-reference): add anchors and copy link buttons to allOf body parameters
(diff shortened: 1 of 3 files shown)


hanspagel:
niiice

hanspagel (approved):

PR #8916, opened 2026-04-20
docs(pricing): simplify pricing page layout

marclave:
## Problem

The `/pricing` page used a large custom HTML and CSS implementation that was hard to scan and harder to maintain. We want a simpler Scalar markdown page with a clearer structure for pricing cards, product sections, and FAQ content.

## Solution

Replaced the bespoke layout in `documentation/guides/pricing.md` with a markdown-first page using built-in docs patterns:
- added a new H1 and description
- simplified the pricing overview into card-based plan summaries
- added a trusted-by logo wall
- reorganized content into `Developer Docs`, `Agent Scalar`, and `Team & Access`
- rewrote the FAQ section to match the requested heading and description

## Checklist

- [x] I explained why the change is needed.
- [ ] I added a changeset. 
- [ ] I added tests.
- [x] I updated the documentation.

<div><a href="https://cursor.com/agents/bc-a8effa33-bb26-4034-92a6-71be5f3eeae3"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-a8effa33-bb26-4034-92a6-71be5f3eeae3"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

PR #8973, opened 2026-04-23
docs(pricing): simplify pricing page layout

cameronrohani:
follow up to #8916 

image

### DIFF
--- a/documentation/guides/pricing.md
+++ b/documentation/guides/pricing.md
@@ -943,7 +943,7 @@ h4.t-editor__heading {
   height: 70px;
 }
 .pricing-table-group_header._sticky-2 {
-  height: 58px;
+  height: 57px;
   top: 118px;
 }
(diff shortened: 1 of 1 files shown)


hanspagel (commented):
visually great, but copy is unclear

I’m not sure whether I can use scalar docs or not, because it says self-hosting, but I have 1 editor seat for scalar docs

I think "Free" shouldn't even be in the table, but a separate "self-hosting"/"open-source" card (or whatever) above the table

cameronrohani:
> visually great, but copy is unclear
> 
> I’m not sure whether I can use scalar docs or not, because it says self-hosting, but I have 1 editor seat for scalar docs
> 
> I think "Free" shouldn't even be in the table, but a separate "self-hosting"/"open-source" card (or whatever) above the table

yup! @marclave is going to wrap this up now :)

hanspagel:
going back to draft for now

PR #8984, opened 2026-04-24
fix(workspace-store): unable to select scopes when using preferredSecurityScheme

### LINKED ISSUE #8983: Unable to select scopes when using `preferredSecurityScheme`
### What happens?

I am using an authorization code OAuth2 flow, and when I set the `preferredSecurtyScheme` under the `authentication` section to this scheme and then attempt to click on the dropdown and select a scope, the box checks, but the `Scopes Selected` count does not increase, and the scope is not included in the authorization call. Unchecking the preferred scheme and re-selecting it causes it to work again.

This issue does not occur in the Sandbox because it requires these configuration changes, but I've supplied a test document I used to test.

Based on my testing, the issue started occurring when using `api-reference@1.51.0` and the browser cache is also hiding the issue. Something is cached if you went to the page when it was using an older version, as it continues to work into newer versions, but a fresh page load will exhibit the issue.

Here is an example of the HTML page with configurations

```html
<!doctype html>
<html>
<head>
	<title>API Reference</title>
	<meta charset="utf-8" />
	<meta
	name="viewport"
</head>

<body>
	<div id="app"></div>

	<!-- Load the Script -->
	<script src="https://cdn.jsdelivr.net/npm/@scalar/api-reference@1.51.0"></script>

	<script>
		url: 'openapi.json',

		// Configure authentication
		authentication: {
			preferredSecurityScheme: 'oauth2'
		}
	})
	</script>
</body>
</html>
```

I expect to be able to select the required scopes and successfully authorize.

### OpenAPI Document

https://sandbox.scalar.com/e/DJ6vB

amritk:
## Summary

Fixes OAuth2 scope selection not working when using `preferredSecurityScheme` configuration option.

## Problem

When `preferredSecurityScheme` is configured in the authentication settings, the security selection is computed on-the-fly by `getSelectedSecurity()` but **not persisted** to the auth store. This caused `updateSelectedScopes()` to fail silently when users tried to select OAuth2 scopes - the checkbox would appear to check, but the "Scopes Selected" count would not increase and the scopes were not actually selected.

## Root Cause

In commit 537db72ec (#8712), the authentication logic was refactored to use separate stores. The `watch` that initialized the document's security selection was removed, and the logic was moved to `getSelectedSecurity()`. However, this function only _computes_ the selection without persisting it.

When `updateSelectedScopes()` runs, it calls `getAuthSelectedSchemas()` which returns `undefined` because nothing was ever stored. The function then exits early without updating the scopes.

## Solution

Modified `updateSelectedScopes()` to initialize the selection in the auth store when it doesn't exist. The initialization now uses `getSelectedSecurity()` to ensure consistency with how selections are computed elsewhere, properly handling default scopes via `x-default-scopes`.

This ensures that:

1. First scope selection initializes the security requirement with proper default scopes
2. Subsequent scope updates work as expected
3. The selection persists across page interactions
4. Logic remains consistent with `getSelectedSecurity()` behavior

The initialization only happens when:
- A scope update is attempted (user clicks a scope checkbox)
- No selection exists in the auth store
- The security scheme requires scopes (OAuth2/OpenID Connect)

## Testing

- ✅ Added unit test covering the scenario: "updates scopes when using preferredSecurityScheme without stored selection"
- ✅ All existing workspace-store tests pass (2247 passed)
- ✅ Manual testing confirms scope selection now works correctly with `preferredSecurityScheme`

## Visual

Initial OAuth2 state showing Scopes Selected 0 / 3

After selecting first scope - count shows 1 / 3

After selecting second scope - count shows 2 / 3

Expanded scopes list with checkboxes and selection state

## Ticket

Fixes #8983

[Slack Thread](https://apidocumentation.slack.com/archives/C05TFBEPV6Z/p1777045677563569?thread_ts=1777045677.563569&cid=C05TFBEPV6Z)

<div><a href="https://cursor.com/agents/bc-620205d3-9cfe-5fd0-917d-eeddf2a51938"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-620205d3-9cfe-5fd0-917d-eeddf2a51938"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/packages/workspace-store/src/mutators/auth.ts
+++ b/packages/workspace-store/src/mutators/auth.ts
@@ -376,63 +376,27 @@ export const updateSelectedScopes = (
     return store?.auth.getAuthSelectedSchemas({ type: 'operation', documentName, path: meta.path, method: meta.method })
   }
 
-  let target = getTarget()
+  const targetFromStore = getTarget()
+  const securityRequirements =
+    meta.type === 'document'
+      ? (document?.security ?? [])
+      : ((getResolvedRef(document?.paths?.[meta.path]?.[meta.method])?.security ?? []) as SecurityRequirementObject[])
+  const securitySchemes = document?.components?.securitySchemes ?? {}
+  const preferredScheme = id.length === 1 ? id[0] : id
 
-  // If no selection exists (e.g., when using preferredSecurityScheme), initialize it
-  // using the same logic as getSelectedSecurity to ensure consistency
-  if (!target) {
-    const securityRequirements = meta.type === 'document' ? (document?.security ?? []) : undefined
-    const securitySchemes = document?.components?.securitySchemes ?? {}
-
-    // Use the scheme being updated (from id) as the preferred scheme
-    // This ensures we initialize with the correct scheme that matches what the user is trying to select
-    const preferredScheme = id.length === 1 ? id[0] : id
-
-    // Compute the default selection using getSelectedSecurity logic
-    // Type assertion is safe here as we're passing the same structure getSelectedSecurity expects
-    const defaultSelection = getSelectedSecurity(
-      undefined,
-      undefined,
-      securityRequirements,
-      securitySchemes as Record<string, { type?: string; 'x-default-scopes'?: string[] } | undefined>,
-      preferredScheme,
-    )
-
-    // Deep clone the selection to avoid mutating the original document.security object
-    // This prevents scope updates from polluting the OpenAPI document structure
-    const clonedSelection = {
-      selectedIndex: defaultSelection.selectedIndex,
-      selectedSchemes: defaultSelection.selectedSchemes.map((scheme) => ({ ...scheme })),
-    }
-
-    // If we computed a selection, use it; otherwise create a minimal one with the scheme from id
-    const initialSelection =
-      clonedSelection.selectedSchemes.length > 0
-        ? clonedSelection
-        : {
-            selectedIndex: 0,
-            selectedSchemes: [
-              // Fallback: build a requirement from the id array with empty scopes
-              Object.fromEntries(id.map((schemeId) => [schemeId, []])) as SecurityRequirementObject,
-            ],
-          }
-
-    // Initialize the selection
-    if (meta.type === 'document') {
-      store?.auth.setAuthSelectedSchemas({ type: 'document', documentName }, initialSelection)
-    } else {
-      store?.auth.setAuthSelectedSchemas(
-        { type: 'operation', documentName, path: meta.path, method: meta.method },
-        initialSelection,
-      )
-    }
-
-    // Re-fetch the target after initialization
-    target = getTarget()
-    if (!target) {
-      return
-    }
-  }
+  const fallbackTarget = getSelectedSecurity(
+    undefined,
+    undefined,
+    securityRequirements,
+    securitySchemes as Record<string, { type?: string; 'x-default-scopes'?: string[] } | undefined>,
+    preferredScheme,
+  )
+  const target = targetFromStore
+    ? targetFromStore
+    : {
+        selectedIndex: fallbackTarget.selectedIndex,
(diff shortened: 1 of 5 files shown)


packages/workspace-store/src/mutators/auth.ts
```diff
@@ -335,9 +335,34 @@ export const updateSelectedScopes = (
     return store?.auth.getAuthSelectedSchemas({ type: 'operation', documentName, path: meta.path, method: meta.method })
   }
 
-  const target = getTarget()
+  let target = getTarget()
+
+  // If no selection exists (e.g., when using preferredSecurityScheme), initialize it
   if (!target) {
-    return
+    // Build a security requirement from the id array
```
DemonHa: can we use `getDefaultSecurity` function here so we can keep the same logic?

DemonHa (changes requested):

packages/workspace-store/src/mutators/auth.ts
amritk: @cursoragent implement this plz

amritk:
@cursoragent address the latest cursor-bot issues

amritk:
Okay this one was done purely by cursor, did we go too far?

packages/workspace-store/src/mutators/auth.ts
```diff
@@ -335,9 +336,62 @@ export const updateSelectedScopes = (
...
   }
 
-  const target = getTarget()
+  let target = getTarget()
+
+  // If no selection exists (e.g., when using preferredSecurityScheme), initialize it
+  // using the same logic as getSelectedSecurity to ensure consistency
   if (!target) {
-    return
+    const securityRequirements = meta.type === 'document' ? (document?.security ?? []) : undefined
```
DemonHa: you don't really need to do any of this tho. can we simplify this to just be `getTarget() ?? getSelectedSecurity(...props)` no need to initialize i think

DemonHa (changes requested):

packages/workspace-store/src/mutators/auth.ts
```diff
@@ -335,10 +336,27 @@ export const updateSelectedScopes = (
...
+    securityRequirements,
+    securitySchemes as Record<string, { type?: string; 'x-default-scopes'?: string[] } | undefined>,
+    preferredScheme,
+  )
+  const target = targetFromStore
+    ? targetFromStore
+    : {
+        selectedIndex: fallbackTarget.selectedIndex,
+        selectedSchemes: fallbackTarget.selectedSchemes.map((scheme) => ({ ...scheme })),
+      }
```
DemonHa: can we do just `targetFromStore ?? fallbackTarget`

packages/workspace-store/src/mutators/auth.ts
```diff
@@ -349,6 +367,22 @@ export const updateSelectedScopes = (
...
+    if (meta.type === 'document') {
+      store?.auth.setAuthSelectedSchemas({ type: 'document', documentName }, target)
+      return
+    }
+
+    store?.auth.setAuthSelectedSchemas(
+      { type: 'operation', documentName, path: meta.path, method: meta.method },
+      target,
+    )
+  }
```
DemonHa: We don't need to persist here

DemonHa (changes requested):

marclave (approved):

PR #8987, opened 2026-04-25
chore(api-reference): move version logs to component

amritk:
## Problem

Currently we only log the verison in the standalone app.

## Solution

We should be logging the version anywhere the ApiReference component exists so I moved them in there instead.

## Checklist

- [x] I added a changeset. 
- [ ] I added tests.

---

> [!NOTE]
> **Low Risk**
> Low risk: only changes build-time defines and moves a `console.info` version log from the standalone entrypoint into the main `ApiReference` component.
> 
> **Overview**
> Moves the `@scalar/api-reference` version log from `src/standalone.ts` into `ApiReference.vue`, so the version is printed anywhere the component is used (browser-only).
> 
> Replaces the old `process.env.SCALAR_API_REFERENCE_VERSION` define with a build-time `PACKAGE_VERSION` constant across Vite configs (including the ESM playground), and adds `vite-env.d.ts` typing for the new global. Adds a patch changeset for `@scalar/api-reference`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit daae7e6390164f112abf2ec241dbedb9b03f7c3d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/api-reference/src/standalone.ts
+++ b/packages/api-reference/src/standalone.ts
@@ -1,11 +1,6 @@
 import { findDataAttributes, getConfigurationFromDataAttributes } from '@/standalone/lib/html-api'
 import { registerGlobals } from '@/standalone/lib/register-globals'
 
-// Log the package version
-if (process.env.SCALAR_API_REFERENCE_VERSION) {
-  console.info(`@scalar/api-reference@${process.env.SCALAR_API_REFERENCE_VERSION}`)
-}
-
 registerGlobals()
 
 // Look for data attributes in the HTML (legacy)
--- a/.changeset/small-zoos-rush.md
+++ b/.changeset/small-zoos-rush.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
+Fix the ESM playground config by defining `process.env.SCALAR_API_REFERENCE_VERSION` so module loading works in browsers.
--- a/.changeset/small-zoos-rush.md
+++ b/.changeset/small-zoos-rush.md
@@ -1,5 +1,0 @@
----
-'@scalar/api-reference': patch
----
-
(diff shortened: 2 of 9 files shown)


marclave (approved):

packages/api-reference/src/standalone.ts
```diff
@@ -1,11 +1,6 @@
 import { findDataAttributes, getConfigurationFromDataAttributes } from '@/standalone/lib/html-api'
 import { registerGlobals } from '@/standalone/lib/register-globals'
 
-// Log the package version
```
hanspagel, line 4: if we don't use SCALAR_API_REFERENCE_VERSION anymore, we can remove it from the vite.config.ts and vite.standalone.config.ts

hanspagel (changes requested):

packages/api-reference/src/standalone.ts
hanspagel: ah, we can't easily import from the package.json, that's why we need to process.env thing

amritk: you sure? I thought package.json was safer than process.env?

amritk: but also import.meta should get replaced at build time right. Let me check

amritk: oh its not process.env lol

    'process.env.SCALAR_API_REFERENCE_VERSION': `"${version}"`,

😓 I'll just change the name so its not confusing

amritk: alright I matched it to what we do in the client now!

Issue #7398, opened 2025-11-20
[Circular Reference] in polymorphic extension type resolution


Oja95:
### What happens?

Example for an endpoint produces scalar specific '[Circular Reference]' within the example. Doesn't show up in Swagger.

### What did you expect to happen?

Would expect not to see such fields in the sample request.

### OpenAPI Document

https://sandbox.scalar.com/e/Gq6Ni

kaarelj:
@hanspagel, thank you for assigning the correct labels! Are there any updates regarding this issue?

hanspagel:
This is fixed :))

Issue #6704, opened 2025-08-31
recursive schema for request body crashes the scalar open api client


azxza:
### What happens?

when i use the body schema:
```json
{
  "required": true,
  "content": {
    "application/json": {
      "schema": {
        "type": "object",
        "properties": {
          "id": {
            "type": "string"
          },
          "type": {
            "const": "section",
            "type": "string"
          },
          "linkedDiagram": {
            "type": "string"
          },
          "name": {
            "type": "string"
          },
          "projectID": {
            "type": "string"
          },
          "content": {
            "type": "string"
          },
          "children": {
            "type": "array",
            "items": {
              "anyOf": [
                {
                  "$id": "T0",
                  "type": "object",
                  "properties": {
                    "id": {
                      "type": "string"
                    },
                    "content": {
                      "type": "string"
                    },
                    "type": {
                      "const": "section",
                      "type": "string"
                    },
                    "children": {
                      "type": "array",
                      "items": {
                        "anyOf": [
                          {
                            "$ref": "T0"
                          },
                          {
                            "type": "object",
                            "properties": {
                              "id": {
                                "type": "string"
                              },
                              "type": {
                                "const": "content",
                                "type": "string"
                              },
                              "content": {
                                "type": "string"
                              }
                            },
                            "required": [
                              "id",
                              "type",
                              "content"
                            ]
                          }
                        ]
                      }
                    }
                  },
                  "required": [
                    "id",
                    "content",
                    "type",
                    "children"
                  ]
                },
                {
                  "type": "object",
                  "properties": {
                    "id": {
                      "type": "string"
                    },
                    "type": {
                      "const": "content",
                      "type": "string"
                    },
                    "content": {
                      "type": "string"
                    }
                  },
                  "required": [
                    "id",
                    "type",
                    "content"
                  ]
                }
              ]
            }
          }
        },
        "required": [
          "id",
          "type",
          "name",
          "projectID",
          "content",
          "children"
        ]
      }
    },
    "multipart/form-data": {
      "schema": {
        "type": "object",
        "properties": {
          "id": {
            "type": "string"
          },
          "type": {
            "const": "section",
            "type": "string"
          },
          "linkedDiagram": {
            "type": "string"
          },
          "name": {
            "type": "string"
          },
          "projectID": {
            "type": "string"
          },
          "content": {
            "type": "string"
          },
          "children": {
            "type": "array",
            "items": {
              "anyOf": [
                {
                  "$id": "T0",
                  "type": "object",
[...]

Issue #6660, opened 2025-08-26
"Maximum call stack size exceeded" still occurs with v0.7.13 despite PR #6650 fix


Darren567144:
### What happens?

When navigating to the /api/ route with Scalar v0.7.13 (latest version that includes PR #6650
  fix), the browser immediately throws:

  Uncaught (in promise) RangeError: Maximum call stack size exceeded
      at WM (standalone.js:12088)
      at standalone.js:12096
      at WM (standalone.js:12095)
      [pattern repeats hundreds of times]

  Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'deprecated')
      at vye (standalone.js:25354)

  The API documentation fails to render completely. This occurs even after:
  - Updating to v0.7.13 (with PR #6650)
  - Fresh npm install

  Environment:
  - @scalar/docusaurus: 0.7.13
  - @docusaurus/core: 3.8.1
  - Node: v24.4.0
  - OpenAPI 3.1.0 spec with ~97 $ref references to components/schemas and components/parameters

### What did you expect to happen?

The API documentation should render successfully without any circular reference errors, as PR
  #6650 was supposed to fix the "Maximum call stack size exceeded" issue by updating to TypeBox
  with the circular clone fix.

  Based on issue #6641 and the merged PR #6650, v0.7.13 should handle OpenAPI specs with nested
  $ref references without hitting infinite recursion in the WM function.

  How to reproduce

  1. Install Scalar in a Docusaurus project:
  npm install @scalar/docusaurus@0.7.13

  2. Configure in docusaurus.config.ts:
  plugins: [
    [
      "@scalar/docusaurus",
      {
        id: "api",
        label: "API Reference",
        route: "/api/",
        configuration: {
          spec: {
            url: "/index.yaml",
          },
        },
      },
    ],
  ]

  3. Use an OpenAPI spec with multiple $ref references to #/components/schemas/* and
  #/components/parameters/* (around 90-100 references)
  4. Navigate to /api/ route

  The issue appears to be triggered by OpenAPI specs with a certain complexity of component
  references, even though they don't have obvious circular dependencies. Happy to provide my
  OpenAPI spec privately if that would help debug this edge case.

amritk:
Hello @Darren567144 could you please send me your document? I added a complex one with many refs as a test case but I guess it didn't cover your case. Would love to see what's causing it!

Darren567144:
Please let me know if there's anything else, want to get this resolved
asap.

[...]

Issue #6641, opened 2025-08-25
Maximum call stack size exceeded with complex OpenAPI specifications


hanspagel:
### Discussed in https://github.com/scalar/scalar/discussions/6636

<div type='discussions-op-text'>

<sup>Originally posted by **Darren567144** August 25, 2025</sup>
All Scalar versions cause `RangeError: Maximum call stack size exceeded` when processing complex OpenAPI specifications with nested `$ref` references.

  ## Impact
  - **Severity**: Critical - Complete failure to render content
  - **Scope**: ALL published versions affected (due to mass
  republish on Aug 22, 2024)
  - **Users affected**: Anyone with OpenAPI specs containing nested references

  ## Root Cause ( I believe )
  **File**: `packages/api-reference/src/features/example-responses/helpers/get-resolved-ref-deep.ts`
  **Lines**: 56, 62, 75, 79

The `getResolvedRefDeep` function calls `visited.delete(raw)` after processing each node, removing it from circular reference protection. This allows exponential recursion growth with shared references.

  ## Steps to Reproduce
  1. Use any complex OpenAPI spec with nested `$ref` references
  2. Load with any Scalar version + Docusaurus
  3. Navigate to API reference page
  4. Browser console shows: `Maximum call stack size exceeded`

  ## Environment
  - **Browser**: All browsers
  - **Scalar versions**: ALL (mass republished Aug 22)
  - **Integration**: @scalar/docusaurus

  ## Proposed Fix
  Remove all `visited.delete(raw)` calls from lines 56, 62, 75, 79 in `get-resolved-ref-deep.ts`</div>

amritk:
I was trying to replicate this but was unable. I think this is actually the issue from addDocument and not getResolvedRefDeep. I will await a replication from the user otherwise will close with another test

PR #9019, opened 2026-04-28
fix(api-reference): cap recursion depth in Schema to avoid freezing on circular $refs

RapidPoseidon:
## Problem

Schemas that recursively reference themselves — either directly (`Category` whose `children: array of Category`), or via a polymorphic discriminator that closes on itself (`oneOf` whose subtypes contain a `$ref` back to the discriminator base) — are well-formed OpenAPI 3.x. The `api-reference` `Schema` component renders them by inlining each subtype recursively, never terminating. The tab freezes and eventually OOMs.

A real-world Firefox profiler trace from a user clicking an endpoint backed by such a schema shows ~52% of CPU in `libc malloc`, ~41% in Vue's `runIfDirty`, with `Schema → SchemaObjectProperties → SchemaProperty → Schema` cycles 300+ frames deep. Every level spreads the schema object (`{...obj}`, firing Proxy traps via `getOwnPropertyDescriptor` and `Reflect.ownKeys`) before re-rendering.

This same class of bug is tracked by a long line of issues — #6704, #7398, #6641, #6660, #5213, #4421, #2736 — most marked closed but the underlying recursion in `api-reference/Schema` was never capped. The fix that landed in #8757 (which I'm directly mirroring here) only patched `packages/openapi-to-markdown/src/components/Schema.vue`.

## Solution

Mirror #8757 in the `api-reference` Schema component:

- Add `MAX_DEPTH = 10` to `Schema.vue`.
- Add a `depth?: number` prop (default `0`) to `Schema`, `SchemaObjectProperties`, `SchemaProperty`, and `SchemaComposition`.
- Thread the prop through the four components, incrementing **only** at the recursive boundary — i.e. wherever one `Schema` renders another `Schema` (in `SchemaProperty`'s object/array-items branches and in `SchemaComposition`'s `allOf` and selected-composition branches). Intermediate components (`SchemaObjectProperties`, the non-recursive parts of `SchemaProperty`) just forward `depth` unchanged.
- When `depth >= MAX_DEPTH`, render a `[Circular Reference]` placeholder instead of recursing. CSS-wise it's a small muted block consistent with the existing schema-card description style.

I considered reusing the existing `level` prop, but `level` increments in different places along the chain (it's a CSS/layout counter, not a recursion counter) and overloading it would couple the depth limit to visual nesting. Keeping `depth` separate matches what #8757 did for `openapi-to-markdown` exactly.

`MAX_DEPTH = 10` is the same constant as #8757; in practice it's far past anything a real schema would naturally nest, and the placeholder is recoverable.

### Tests

`Schema.test.ts` adds three cases under a new `circular references` describe block:

1. `depth: 99` mounts → `[Circular Reference]` rendered, no `schema-card--level-` classes.
2. `depth: 0` with a normal nested schema → no placeholder, `schema-card--level-` present.
3. A truly self-referencing in-memory schema object (`children.items === parent`) → `mount()` does not throw a "Maximum call stack size exceeded".

All 26 tests in `Schema.test.ts` pass locally; all 455 tests across the api-reference Schema-related suite still pass.

## Checklist

- [x] I added a changeset.
- [x] I added tests.
- [ ] I updated the documentation. _(no doc changes needed — this is a defensive guard)_

### DIFF
--- a/.changeset/sleepy-poseidons-cap.md
+++ b/.changeset/sleepy-poseidons-cap.md
@@ -1,0 +1,9 @@
+---
+'@scalar/api-reference': patch
+---
+
+fix(api-reference): cap recursion depth in Schema component to prevent freezing on circular `$ref` chains
+
+Adds a `MAX_DEPTH = 10` guard to the `Schema` component, mirroring the fix applied to `@scalar/openapi-to-markdown` in #8757. Schemas that recursively reference themselves — directly, or via a polymorphic discriminator that closes on itself — are well-formed OpenAPI but were causing the api-reference renderer to recurse indefinitely, freezing the browser tab. Past the depth limit the component now renders a `[Circular Reference]` placeholder.
+
+The new `depth?: number` prop is threaded through `Schema`, `SchemaObjectProperties`, `SchemaProperty`, and `SchemaComposition`, and increments only at the recursive boundary (`Schema → Schema`).
--- a/packages/api-reference/src/components/Content/Schema/Schema.test.ts
+++ b/packages/api-reference/src/components/Content/Schema/Schema.test.ts
@@ -1,4 +1,5 @@
 import { coerceValue } from '@scalar/workspace-store/schemas/typebox-coerce'
+import type { SchemaObject } from '@scalar/workspace-store/schemas/v3.1/strict/openapi-document'
 import { SchemaObjectSchema } from '@scalar/workspace-store/schemas/v3.1/strict/openapi-document'
 import { mount } from '@vue/test-utils'
 import { describe, expect, it } from 'vitest'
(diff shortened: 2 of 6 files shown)


LucStr:
Hey @hanspagel @marclave or anyone else it might concern. @RapidPoseidon is our coding agent. I am hereby responsible for this pr and happy to elaborate if necessary.

RapidPoseidon:
Force-pushed a fix for the `types` job: `Schema.test.ts` had a TS2322 from a wrong cast around the self-referencing test fixture (`Parameters<typeof coerceValue<...>>[1]` resolved to `unknown`). Now uses a direct `SchemaObject` cast and a typed local for the recursive `children.items` assignment.

Verified locally on the new commit:
- `pnpm vue-tsc --noEmit` — clean for the touched files
- `pnpm --filter @scalar/api-reference test Schema.test` — 26/26 pass
- `pnpm format:check` — clean
- `pnpm lint:check` — clean

CI for the new SHA shows `action_required` for the heavier workflows — would appreciate an approval-and-run when you have a moment.

amritk:
hello! Do you think you could replicate this in the [sandbox](https://sandbox.scalar.com/)? 

In the references we should only render it when you hit show children so unless the user opens it that many times it shouldn't be an issue? Unless I'm missing the use case here.

image

hanspagel:
idk about that, in the UI I think it shouldn't be a problem since we lazily expand the schemas

moving to draft until we have a reproduction

LucStr:
Hey @hanspagel @amritk 
I did some more digging here and turns out that I will be able to fix our specific issue by changing our OpenAPI document generator. It seems like the issue is that our generator created and allOf within another allOf, which seems to break scalar. It can be reproduced like this https://sandbox.scalar.com/p/UQU_M  

I didn't figure out how this can be mitigated easily, so feel free to close this PR.

hanspagel:
thanks for the follow up 🙏

PR #9023, opened 2026-04-28
feat(helpers): added our own slugify

amritk:
## Problem

We used a few different sluggers and I needed one in the client app.

## Solution

Made one in helpers so we can use it across. 

@DemonHa there were some cases that we are using just the slug instead of slugify. So just double check those, they do not auto increment the number. But it seems like we should do that in `packages/workspace-store/src/navigation/get-navigation-options.ts`

## Checklist

- [x] I added a changeset. 
- [x] I added tests.

---

> [!NOTE]
> **Medium Risk**
> Changes slug/id generation across `api-reference`, `workspace-store`, and migration/import paths, which can affect persisted document keys, navigation IDs, and heading anchors. Behavior shifts (e.g., stripping punctuation/emoji and truncating to 255 chars) could cause subtle link/ID mismatches or collisions in existing data.
> 
> **Overview**
> Introduces a new shared `@scalar/helpers` string `slugify` (Unicode-aware, punctuation/emoji stripping, 255-char limit, optional case preservation and allowed special chars) plus a stateful `slugger` that appends numeric suffixes to avoid collisions.
> 
> Replaces `github-slugger` usage in `api-reference` (heading/config slugs), `oas-utils` legacy-to-IndexedDB migration (unique document names), and `workspace-store` navigation ID generation (document/tag/model/example/webhook IDs) to use the new helpers, and updates related tests/expectations for the new slug output.
> 
> Removes the api-client’s local `slugify` helper in favor of `@scalar/helpers/string/slugify`, adjusts unique document slug generation to allow `.` characters, drops `github-slugger` dependencies, and adds `test:watch` scripts to several packages.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d76ee6c51e98e02b8e5e8399b8f7ab6592286b51. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/workspace-store/src/navigation/get-navigation-options.ts
```diff
@@ -27,7 +27,7 @@ export type NavigationOptions =
  */
 export const getNavigationOptions = (documentName: string, options?: NavigationOptions): TraverseSpecOptions => {
   const generateId: IdGenerator = (props) => {
-    const documentId = `${slug(documentName)}`
+    const documentId = `${slugify(documentName)}`
```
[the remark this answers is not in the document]
amritk: should be ok as they will just get re-generated, we can update it to allow this but I think this is fine

marclave (approved):

packages/workspace-store/src/navigation/get-navigation-options.ts
```diff
@@ -27,7 +27,7 @@ export type NavigationOptions =
  */
 export const getNavigationOptions = (documentName: string, options?: NavigationOptions): TraverseSpecOptions => {
   const generateId: IdGenerator = (props) => {
-    const documentId = `${slug(documentName)}`
```
DemonHa, line 30: Yeah as you say in the description, this was supposed to be just the slug of whatever the document name is.. but I guess we can end up with conflicts here if we have two document with same name but different casing.

Like `hello` with `Hello` which would both produce `hello` as the output..

We could try to use the slugify instance so we generate unique id's, the problem with that is on the references we load documents on demand and we build the sidebar too.. so depending on the order you load documents the id so we have to make sure we don't really have any conflicts there

DemonHa (changes requested):

packages/workspace-store/src/navigation/helpers/traverse-document.test.ts
```diff
@@ -197,7 +197,7 @@ describe('traverseDocument', () => {
     })
     expect((result.children[1] as TraversedTag).children).toHaveLength(1)
     expect((result.children[1] as TraversedTag).children?.[0]).toEqual({
-      'id': 'doc-1/model/TestModel',
+      'id': 'doc-1/model/testmodel',
```
hanspagel: model ids are case sensitive BTW

when we change the test here, does that mean we introduce a regression?

https://sandbox.scalar.com/e/K6qDr

hanspagel (changes requested):
For such a PR, it would be great if we wouldn't have to modify existing tests.
I think for preserving the case, we want to have the same behaviour as before.

packages/workspace-store/src/navigation/helpers/traverse-document.test.ts
amritk: going to do another pass on this one

packages/workspace-store/src/navigation/get-navigation-options.ts
amritk: references doesn't matter here as you only have one document

packages/workspace-store/src/navigation/helpers/traverse-document.test.ts
amritk: ok will preserveCase by default and if you want lowerCase you can pass it in

amritk:
so @hanspagel @DemonHa I went to preserveCase by default BUT that didn't work either!

image

Turns out before we lowerCase some parts and preserveCase some parts, shouldn't we do the same thing consistently?

hanspagel:
we all love consistency! so let's preserve the cases in all places?

amritk:
Alright to reduce the surface for bugs we just went with maintaining the previous casing behavior, we can come back to it if we every want to change!

DemonHa (approved):
LGTM, better not to do braking changes on this PR. We can do that separately

PR #9035, opened 2026-04-29
feat(proxy): forward selected forbidden headers via `X-Scalar-*` when proxying

### LINKED ISSUE #9022: Scalar test request does not supply date header in request
### What happens?

The test request does not send along the date header.

Example:
https://scalar.com/tools/api/tag/registry/get/v1/apis/namespace/slug/version/semver
Perform the test request by adding Date header with "Sun, 06 Nov 1994 08:49:37 GMT" and while the test request overview will say it should be included the 

Request example with Scalar informing Date is included:
Image

Actual invocation:
Image

Date header to be included and filled with supplied value

### OpenAPI Document

_No response_

DemonHa:
Fixes: #9022

## Checklist

- [X] I added a changeset. 
- [x] I added tests.

---

> [!NOTE]
> **Medium Risk**
> Touches request/proxy header handling across browser, Electron, and Go proxy; mistakes could silently drop or spoof upstream headers. Scope is limited to a small explicit allowlist (`Date`, `DNT`, `Referer`, `User-Agent`) with added tests reducing risk.
> 
> **Overview**
> **Enables forwarding of selected browser-forbidden headers when proxying.** `buildRequest` now mirrors `Date`, `DNT`, `Referer`, and `User-Agent` into `X-Scalar-*` and removes the originals when using the proxy or Electron, so these values survive browser stripping.
> 
> **Proxy/Electron now understand the new headers.** The Go proxy forwards `X-Scalar-*` values back to their corresponding upstream headers (and deletes the extension headers), and Electron’s `applyRequestHeaderTransforms` is generalized to promote the same allowlist.
> 
> **Tests updated/added** to cover the new rewrite/promotion behavior end-to-end, and the earlier Electron-only `User-Agent` rewrite in `request-factory` is removed in favor of the centralized logic.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 70f258cc56995c8a648607b67e4eb66ab3868d68. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


amritk (approved):
Should work but need to test in the app

packages/workspace-store/src/request-example/builder/build-request.ts
```diff
@@ -166,6 +167,31 @@ export const buildRequest = (
...
+   * Browsers do not let us override the `User-Agent` header on outgoing requests.
+   * In Electron we mirror the value to `X-Scalar-User-Agent` so the main process can pick it up
+   * and apply it as the actual `User-Agent` header on the outgoing request.
+   */
+  if (request.options?.isElectron) {
+    const userAgentHeader = headers.get('user-agent')
+    if (userAgentHeader) {
+      headers.set(X_SCALAR_USER_AGENT, userAgentHeader)
+    }
+  }
```
hanspagel: let us use an array of those headers, and not add an if/else for every single one

projects/proxy-scalar-com/main.go
```diff
@@ -351,6 +351,13 @@ func (ps *ProxyServer) executeProxyRequest(w http.ResponseWriter, r *http.Reques
 	// Remove the X-Scalar-Cookie header
 	outreq.Header.Del("X-Scalar-Cookie")
 
+	// Browsers strip the `Date` header from outgoing requests because it's a forbidden header.
```
hanspagel: same, let's use an array for those headers

hanspagel (changes requested):
Thanks for picking this up so quickly

I think we can keep arrays of those headers and iterate over them

And while we’re at it, can we add more of the "forbidden" headers?

We might want to think about the implications for every single one, there might be a reason they are forbidden and we might not want to support all.

hanspagel:
I’ve refactored it a bit to support more headers, and to have the same logic for all headers (including user-agent), for some reason we only supported the X-Scalar-User-Agent logic in Electron?

I think in Electron we can directly set those headers (with a tiny workaround), for the web client, we can rewrite the X-Scalar-* prefixed headers to then be forwarded as the original headers (e.g. `User-Agent`)

amritk:
> I’ve refactored it a bit to support more headers, and to have the same logic for all headers (including user-agent), for some reason we only supported the X-Scalar-User-Agent logic in Electron?
> 
> I think in Electron we can directly set those headers (with a tiny workaround), for the web client, we can rewrite the X-Scalar-* prefixed headers to then be forwarded as the original headers (e.g. `User-Agent`)

@hanspagel yea electron can read/write any of the headers

marclave (approved):

PR #9055, opened 2026-04-30
fix(api-client): expand query params in object

### LINKED ISSUE #8615: Scalar doesn't expand "Query Parameters" when the parameter type is object
### What happens?

- Scalar does not expand object-type query parameters in the "Query Parameters" section.
- Nested properties are not shown. Instead, the object parameter is displayed as a single item, which makes it difficult to provide the expected query input.

Image

Image

It would be helpful if parameters with an object type could be expanded in the "Query Parameters" section, similar to Postman

Image

### OpenAPI Document

https://sandbox.scalar.com/e/KuijI

alonfaraj:
## Problem

When a query parameter has `type: object` with nested properties, the api-client displayed it as a single row instead of expanding the individual properties.

## Solution
Object-type query parameters are now expanded into one row per property, matching how tools like Postman handle this. The expansion respects OpenAPI serialization styles (`form` with `explode` and `deepObject`).

- **Added `createParameterRows` helper** - takes a parameter and returns one or more `TableRow` entries. For object-type query parameters with `explode: true` (default for `form` style) or `deepObject` style, it expands each property into its own row. Non-object parameters pass through as a single row like before.

- **`RequestBlock.vue`** - replaced the inline parameter-to-row mapping with `createParameterRows`.

- **Updated `createParameterHandlers`** - when upserting an expanded row, all sibling rows are reassembled into a single object value and emitted under the parent parameter name.

- **Added `sourceParameterValuePath` to `TableRow`** - tracks which property path each expanded row maps to inside the parent parameter, so the handler knows how to reconstruct the object.

- **Changed `operation:upsert:parameter` value type** - from `string` to `string | Record<string, unknown>` to support object payloads.

## Checklist

- [x] I added a changeset. 
- [x] I added tests.

Fixes #8615

**After fix (example from the original issue):**
image

### DIFF
--- a/packages/api-client/src/v2/blocks/request-block/helpers/create-parameter-handlers.ts
+++ b/packages/api-client/src/v2/blocks/request-block/helpers/create-parameter-handlers.ts
@@ -1,28 +1,9 @@
+import { setValueAtPath } from '@scalar/helpers/object/set-value-at-path'
 import type { OperationExampleMeta, WorkspaceEventBus } from '@scalar/workspace-store/events'
 
 import type { TableRow } from '@/v2/blocks/request-block/components/RequestTableRow.vue'
 
 type ParameterType = 'path' | 'cookie' | 'header' | 'query'
-
-const setValueAtPath = (target: Record<string, unknown>, path: readonly string[], value: unknown): void => {
-  const [key, ...rest] = path
-  if (!key) {
-    return
-  }
-
-  if (!rest.length) {
-    target[key] = value
-    return
-  }
-
-  const next =
-    typeof target[key] === 'object' && target[key] !== null && !Array.isArray(target[key])
-      ? (target[key] as Record<string, unknown>)
-      : {}
-
-  target[key] = next
-  setValueAtPath(next, rest, value)
-}
 
 const isEmptyValue = (value: unknown): boolean => value === undefined || value === null || value === ''
 
--- a/packages/api-client/src/v2/blocks/request-block/helpers/create-parameter-handlers.ts
+++ b/packages/api-client/src/v2/blocks/request-block/helpers/create-parameter-handlers.ts
@@ -4,6 +4,55 @@ import type { TableRow } from '@/v2/blocks/request-block/components/RequestTable
 
 type ParameterType = 'path' | 'cookie' | 'header' | 'query'
 
+const setValueAtPath = (target: Record<string, unknown>, path: readonly string[], value: unknown): void => {
+  const [key, ...rest] = path
+  if (!key) {
+    return
+  }
+
+  if (!rest.length) {
+    target[key] = value
+    return
+  }
+
+  const next =
+    typeof target[key] === 'object' && target[key] !== null && !Array.isArray(target[key])
+      ? (target[key] as Record<string, unknown>)
+      : {}
+
+  target[key] = next
+  setValueAtPath(next, rest, value)
+}
+
+const isEmptyValue = (value: unknown): boolean => value === undefined || value === null || value === ''
+
+const getExpandedObjectPayload = (
+  row: TableRow,
+  context: TableRow[],
+  payload: { name: string; value: string; isDisabled: boolean },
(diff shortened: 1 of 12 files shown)


packages/api-client/src/v2/blocks/request-block/helpers/create-parameter-rows.ts
```diff
@@ -0,0 +1,209 @@
...
+  typeof value === 'object' && value !== null && !Array.isArray(value)
+
+const getParameterStyleAndExplode = (parameter: ParameterObject): { style: string; explode: boolean } => {
+  const style = 'style' in parameter && parameter.style ? parameter.style : 'form'
+  const explode = 'explode' in parameter && parameter.explode !== undefined ? parameter.explode : style === 'form'
+
+  return { style, explode }
+}
+
+const getValueAtPath = (source: unknown, path: readonly string[]): unknown => {
```
DemonHa: Can you use `getValueAtPath` from `@scalar/helpers/object/get-value-at-path`?

packages/api-client/src/v2/blocks/request-block/helpers/create-parameter-handlers.ts
```diff
@@ -4,6 +4,55 @@ import type { TableRow } from '@/v2/blocks/request-block/components/RequestTable
 
 type ParameterType = 'path' | 'cookie' | 'header' | 'query'
 
+const setValueAtPath = (target: Record<string, unknown>, path: readonly string[], value: unknown): void => {
```
DemonHa: can you use `import { setValueAtPath } from '@/helpersset-value-at-path'`

packages/api-client/src/v2/blocks/request-block/helpers/create-parameter-rows.ts
```diff
@@ -0,0 +1,209 @@
...
+const isParameterWithSchema = (parameter: ParameterObject): parameter is ParameterWithRequiredSchema =>
+  'schema' in parameter && parameter.schema !== undefined
+
+const resolveSchema = (schema: unknown): SchemaObject | undefined => {
+  const resolvedSchema = getResolvedRef(schema as SchemaObject | { '$ref': string; '$ref-value': SchemaObject })
+
+  return resolvedSchema as SchemaObject | undefined
+}
+
+const isRecord = (value: unknown): value is Record<string, unknown> =>
```
DemonHa: you can use `isObject` from '@scalar/helpers/object/is-object'?

packages/api-client/src/v2/blocks/request-block/helpers/create-parameter-rows.ts
DemonHa, line 1: can we try to add some more comments and maybe clean up the flow a little bit? it's kindda hard to follow

DemonHa (changes requested):

packages/api-client/src/v2/blocks/request-block/helpers/create-parameter-rows.ts
alonfaraj: Hope it looks better and easier now. Let me know what you think

hanspagel:
### Code review

Found 1 issue:

1. Deleting an expanded child row deletes the entire parent parameter. When an object query parameter is expanded into multiple sibling rows, every row is created with the same `originalParameter` reference (only `sourceParameterValuePath` differs). The `delete` handler emits `operation:delete:parameter` with that shared reference, so clicking delete on one child row removes the parent parameter and every sibling row along with it. The UI also does not hide the delete button for child rows.

https://github.com/scalar/scalar/blob/4809608d78e2c854e41257e8c7ce44e86fd1c5d2/packages/api-client/src/v2/blocks/request-block/helpers/create-parameter-handlers.ts#L54-L71

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

alonfaraj:
@hanspagel should be fixed now. Thanks

PR #9064, opened 2026-05-01
feat: code split tailwind CSS

hwkr:
Splits the bundled Tailwind styles out so CSS gets generated as a final build step instead of being shipped inside each package.

* `components`, `sidebar`, `api-client`, `agent-chat` → split bundled tailwind styles
* `api-reference` → split bundled tailwind styles + fixes some CSS conflicts

Before

CleanShot 2026-05-01 at 13 56 10@2x

After

CleanShot 2026-05-01 at 13 59 01@2x

## Checklist

- [x] I added a changeset.
- [x] I updated tests.


packages/api-client/src/index.ts
```diff
@@ -1 +1 @@
-import './style.css'
+// Nothing
```
hwkr, line 1: @amritk I'm not sure if we need this anymore, there's an export in the package.json but I don't think it does anything?

amritk: yolo

amritk (approved):
new contender for PR of the week

PR #9068, opened 2026-05-01
fix(api-client): constrain response body scrolling

### LINKED ISSUE #8217: Inconsistent scroll behavior in response body panel between local and production
### What happens?

I’m using @scalar/express-api-reference in an Express backend to serve my OpenAPI docs on /docs.

Locally and in production I’m using the same code, same OpenAPI spec, same browser and same zoom level (110%), but the response body panel behaves differently:

Local:

The response body panel shows an internal vertical scrollbar when the JSON response is long (100+ lines).
I can scroll inside the body panel, which is very convenient to inspect long responses.
Production (deployed Docker image):

The internal scrollbar is not visible in the response body panel.
The only way to see all the lines is to scroll the entire right column / page, or start selecting text and dragging down.
This makes it harder to inspect long responses, because it’s not obvious that the content overflows and that I need to scroll the whole column instead of just the body panel.
This happens, for example, on the endpoint:

GET /users/clients (tag: Users)
with a paginated response that returns ~100+ lines of JSON.
Even when I open both URLs in the same browser side by side:

Local: http://localhost:8080/docs#tag/users/GET/users/clients
Prod: https://api.nexocr.site/docs#tag/users/GET/users/clients

the local instance shows an internal scrollbar in the response body, and the production instance does not.
I would expect the layout and scroll behavior to be identical.

My integration code (simplified):

import express from 'express';
import { apiReference } from '@scalar/express-api-reference';
import { loadOpenAPISpec } from './src/config/openapi.config.js';

const app = express();

app.use(
  '/docs',
  apiReference({
    content: loadOpenAPISpec(),
    documentId: 'zonafit-api-v1',
  }),
);

There is no custom CSS applied on top of Scalar.

The response body panel should have consistent scroll behavior between local and production for the same OpenAPI document and same browser/zoom.
When the JSON response is taller than the visible area, I expect:
A clear scrollbar on the response body panel (or always-visible scrollbars in that area), and
The ability to scroll inside that panel without needing to scroll the entire page/column.
In short: same layout and scroll behavior in local and production so that long responses are easy to inspect in both environments.

Environment

Package: @scalar/express-api-reference 0.8.39 (version pinned, no ^)
Express: 5.2.1
Local Node: 20.x
Production Node (inside container): 18.20.5
(I see EBADENGINE warnings because Scalar requires node >= 20, but the docs UI still renders and works otherwise.)
Browser: Brave (Chromium-based), also reproduced by a teammate in another Chromium-based browser
Zoom: 110% in both local and production

### OpenAPI Document

Local: http://localhost:8080/docs Production: https://api.nexocr.site/docs

amritk:
## Problem

Long API client response bodies could expand the response Body disclosure instead of keeping scrolling inside the body panel. In embedded API reference/modal layouts this made the outer response column or page become the scroll target, which matches #8217.

CI was failing in `test-packages (3, 3)` because `ResponseBodyStreaming.test.ts` still expected the streaming body content element to expose an internal `.overflow-auto` scroll container, but the PR branch had moved the streaming body to flex sizing without that overflow class.

## Solution

Constrain the response Body disclosure and raw/streaming renderers with flex and `min-h-0` sizing so the body content keeps an internal scroll container. The raw renderer now normalizes broad response data before pretty-printing, and the regression test exercises the regular long JSON response path.

Restore `overflow-auto` on the streaming response content container so streaming responses retain the same internal scrolling behavior as raw responses and the existing streaming tests assert the real behavior again.

## Checklist

- [x] I added a changeset.
- [x] I added tests.

## Ticket

Fixes #8217

<div><a href="https://cursor.com/agents/bc-39760f50-f2d5-4891-ab80-a2605194b7d1"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-39760f50-f2d5-4891-ab80-a2605194b7d1"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/.changeset/quiet-lemons-scroll.md
+++ b/.changeset/quiet-lemons-scroll.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
+
+fix: keep long response bodies scrollable within the body panel
--- a/packages/api-client/src/v2/blocks/response-block/ResponseBlock.test.ts
+++ b/packages/api-client/src/v2/blocks/response-block/ResponseBlock.test.ts
@@ -159,6 +159,33 @@ describe('ResponseBlock', () => {
       const metaInfo = wrapper.findComponent(ResponseMetaInformation)
       expect(metaInfo.props('response')).toEqual(mockResponse)
     })
+
+    it('constrains the body panel so long responses scroll internally', () => {
+      const wrapper = mount(ResponseBlock, {
+        props: {
+          ...defaultProps,
+          response: getDefaultResponse({
+            data: {
+              users: Array.from({ length: 100 }, (_, index) => ({
+                id: index,
+                name: `User ${index}`,
+              })),
+            },
+            headers: {
+              'content-type': 'application/json',
+            },
+          }),
+        },
+      })
+
+      const body = wrapper.find('.response-section-content-body')
+
+      expect(body.classes()).toEqual(
+        expect.arrayContaining(['flex', 'min-h-0', 'overflow-hidden']),
+      )
+      expect(body.find('.body-raw').classes()).toContain('min-h-0')
+      expect(body.find('.body-raw-scroller').classes()).toContain('overflow-auto')
+    })
   })
 
   describe('title display', () => {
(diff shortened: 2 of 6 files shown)


amritk:
another time

PR #9087, opened 2026-05-04
test(api-client): fix flaky Modal test timeout by hoisting store setup to module level

amritk:
## Problem

`Modal.test.ts` had a flaky `passes hideClientButton option to Operation component` test that would occasionally time out in CI with the default 5 000 ms vitest timeout.

### Root cause

The `createModalProps()` helper was `async` and called inside every single `it` block (15 tests × 1 call each). Each call ran the full OpenAPI bundling pipeline via `store.addDocument()`:

- `bundle()` with tree-shaking, ref normalisation, external value resolution
- TypeBox schema validation
- Navigation structure building

On a loaded CI runner this pipeline takes **3–4 s per call**. The last test (`passes hideClientButton`) called it once and then mounted the component **three times**, each time calling `waitForUpdates()`. The cumulative cost regularly hit or exceeded 5 000 ms.

A secondary issue: `waitForUpdates()` included an unconditional `setTimeout(resolve, 100)` sleep. With three mounts in one test that was an extra 300 ms of dead time serving no purpose — `nextTick() + flushPromises()` is sufficient to drain Vue's async queue.

### Fix

**Hoist `createWorkspaceStore()` and `store.addDocument()` to module level** using a top-level `await`, so the expensive bundler pipeline runs **once per file** instead of once per test.

Replace the async `createModalProps()` with a synchronous `createProps()` that only creates cheap reactive wrappers (`computed`, `ref`, `useModal()`, `useModalSidebar()`) against the shared store.

Add an `afterEach` that resets any store state mutated by individual tests (sidebar width, workspace/document environments).

Remove the 100 ms `setTimeout` sleep from `waitForUpdates()`.

### Before / after timing (local, same machine)

| | Before | After |
|---|---|---|
| `passes hideClientButton` | ~4 500 ms (flaky on CI) | **~2 200 ms** |
| Full file | ~96 s (CI) | **~8 s** |

The previously flaky test now completes in well under half its old time, with significant headroom below the 5 000 ms limit even on loaded CI runners.

---

> [!NOTE]
> **Low Risk**
> Test-only refactor that changes shared setup/teardown and async timing; low product risk, but could introduce subtle test state leakage if resets are incomplete.
> 
> **Overview**
> Speeds up and de-flakes `Modal.test.ts` by hoisting `createWorkspaceStore()` + `store.addDocument()` to module scope (top-level `await`) so the expensive OpenAPI bundling runs once per file, and replacing the per-test async `createModalProps()` with a cheap synchronous `createProps()`.
> 
> Removes an unconditional `setTimeout` sleep from `waitForUpdates()` and adds `afterEach` cleanup to reset shared store mutations (sidebar width and environment state) between tests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d9208ff1cd959f8f754b7bc9e62753f1252c992d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/api-client/src/v2/features/modal/Modal.test.ts
+++ b/packages/api-client/src/v2/features/modal/Modal.test.ts
@@ -39,35 +39,38 @@ const getDocument = (overrides: Partial<OpenApiDocument> = {}): OpenApiDocument 
 const createModalOptions = (options: ApiClientOptions = {}): ApiClientOptionsRef => ref(options)
 
 /**
- * Creates a complete modal props setup for testing.
- * This mimics the setup done by createApiClientModal.
+ * Shared workspace store - created once for the entire test file.
+ * Avoids re-running the expensive OpenAPI bundling pipeline on every test.
  */
-const createModalProps = async (documentOverrides: Partial<OpenApiDocument> = {}) => {
-  const store = createWorkspaceStore()
+const store = createWorkspaceStore()
 
-  await store.addDocument({
-    name: 'test-doc',
-    document: getDocument({
-      paths: {
-        '/users': {
-          get: { summary: 'Get users', operationId: 'getUsers' },
-          post: { summary: 'Create user', operationId: 'createUser' },
-        },
-        '/pets': {
-          get: { summary: 'Get pets', operationId: 'getPets' },
-        },
+await store.addDocument({
+  name: 'test-doc',
+  document: getDocument({
+    paths: {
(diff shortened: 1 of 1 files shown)


DemonHa (approved):

hwkr (approved):
🔥

PR #9089, opened 2026-05-04
feat(api-client): let the command palette pick a registry document version

DemonHa:
Builds on the recent registry grouping work in the command palette so the "select document" dropdown is followed by an optional version picker whenever the chosen document is a registry group with more than one loaded version. 

The user can now create a request, tag, or example against a specific version without leaving the palette — previously the flow was listing all the versions as a standalone documents.

## Checklist

- [x] I added a changeset. 
- [x] I added tests.

---

> [!NOTE]
> **Medium Risk**
> Changes command-palette document targeting logic for create flows (request/tag/example), so regressions could create entities in the wrong document/version. Mostly UI/state refactors with good test coverage, but impacts a core interaction path.
> 
> **Overview**
> Command palette create flows can now target **specific loaded registry versions** instead of treating each version as a separate document, via a new `CommandPaletteDocumentSelect` combobox that groups multi-version registry docs and lets users pick a version.
> 
> `TheCommandPalette` now receives `app` and `registryDocuments`, derives grouped document options from sidebar logic, and pre-fills document-aware commands with the currently active document; the request/tag/example forms were refactored to use a shared `useCommandPaletteDocumentSelection` hook and accept injected grouped `documents` + `activeDocumentName`.
> 
> Adds comprehensive unit tests for the new selector, grouping hook (`useCommandPaletteDocuments`), and selection rules; includes a changeset for a minor `@scalar/api-client` bump and small package metadata/script tweaks.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 781244c7e870b222a239012c377bdc506158bb41. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/api-client/src/v2/features/command-palette/components/CommandPaletteDocumentSelect.vue
```diff
@@ -0,0 +1,196 @@
...
+  optionGroups,
+) => {
+  if (!query) {
+    return options
+  }
+  const q = query.toLowerCase()
+  const matchedGroupOptionIds = new Set(
+    optionGroups
+      .filter((g) => g.label.toLowerCase().includes(q))
+      .flatMap((g) => g.options.map((o) => o.id)),
```
amritk: even though its a small list, its a good habbit to save a loop, especially if typing

```suggestion
.flatMap(g =>
  g.label.toLowerCase().includes(q)
    ? g.options.map(o => o.id)
    : []
)
```
or something like that

flatMap is a natural filter so you never need both of those, you can reduce as well if you like

amritk: This whole component is actually pretty ripe for performance improvmenets, I got ai to do a quick pass but you can implement something better


```ts
/**
 * 1. Single-pass group generation
 * Reduces array iterations and pre-calculates lowercase labels for faster searching.
 */
const processedData = computed(() => {
  const flat: DocumentOption[] = []
  const versioned: DocumentOptionGroup[] = []
  const allOptionsMap = new Map<string, DocumentOption>()

  for (let i = 0; i < props.documents.length; i++) {
    const doc = props.documents[i]
    
    if (doc.versions?.length) {
      const options = doc.versions.map((v) => {
        const opt = { id: v.id, label: v.label, documentLabel: doc.label }
        allOptionsMap.set(v.id, opt)
        return opt
      })

      versioned.push({
        label: doc.label,
        options,
      })
    } else {
      const opt = { id: doc.id, label: doc.label, documentLabel: doc.label }
      flat.push(opt)
      allOptionsMap.set(doc.id, opt)
    }
  }

  return {
    groups: [{ label: '', options: flat }, ...versioned] as DocumentOptionGroup[],
    lookup: allOptionsMap
  }
})

/** * 2. O(1) Selection Lookup
 * Replaces .find() on a flatMap with a Map lookup.
 */
const selectedOption = computed(() => 
  props.modelValue ? processedData.value.lookup.get(props.modelValue) : undefined
)

/**
 * 3. High-Performance Filter
 * Optimized to avoid redundant .toLowerCase() calls and nested flatMaps during search.
 */
const filterFn: ScalarComboboxFilterFunction = (query, options, optionGroups) => {
  if (!query) return options
  
  const q = query.toLowerCase()
  
  // Pre-calculate which groups match to avoid repeated work in the loop
  const matchedGroupLabels = new Set(
    optionGroups
      .filter((g) => g.label.toLowerCase().includes(q))
      .map((g) => g.label)
  )

  return options.filter((o) => {
    const docOpt = o as DocumentOption
    return (
      matchedGroupLabels.has(docOpt.documentLabel) ||
      o.label.toLowerCase().includes(q) ||
      (docOpt.documentLabel !== o.label && docOpt.documentLabel.toLowerCase().includes(q))
    )
  })
}
````

amritk (changes requested):
Good for the most part, but whenever I see any nested loops I think we can do better. I know they are probably small lists but doesn't hurt to do it properly

Get cursor to do a performance pass and you can pick up a lot of easy wins

marclave (approved):

amritk (approved):

amritk (approved):

PR #9095, opened 2026-05-05
fix(api-reference): use overflow-wrap break-word for tag labels

### LINKED ISSUE #9092: Generated API Reference word wraps on mobile
### What happens?

This page breaks the summary description of this endpoint summary: "Get a single synced crypto account" in the middle of the word on my mobile view.

Url: https://lm-v2-api-next-a7fabcab8e9a.herokuapp.com/v2/docs

Image: https://discord.com/channels/1135330207960678410/1135330208627564597/threads/1500982236856582325

Words would break on spaces between words

### OpenAPI Document

https://lm-v2-api-next-a7fabcab8e9a.herokuapp.com/v2/openapi

hwkr:
We were using [`word-break: break-word;`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/word-break#break-word) (or trying to) where we should have been using [`overflow-wrap: break-word;`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/overflow-wrap#:~:text=Note%3A%20In%20contrast%20to%20word%2Dbreak%2C%20overflow%2Dwrap%20will%20only%20create%20a%20break%20if%20an%20entire%20word%20cannot%20be%20placed%20on%20its%20own%20line%20without%20overflowing.).

Fixes #9092

## Checklist

- [x] I added a changeset. 
- [ ] I added tests.

### DIFF
--- a/packages/api-reference/src/components/Anchor/Anchor.vue
+++ b/packages/api-reference/src/components/Anchor/Anchor.vue
@@ -14,7 +14,7 @@ const labelId = useId()
 const { cx } = useBindCx()
 </script>
 <template>
-  <span v-bind="cx('group/heading word-break-all relative')">
+  <span v-bind="cx('group/heading wrap-break-word relative')">
     <span
       :id="labelId"
       class="contents">
--- a/.changeset/small-ravens-help.md
+++ b/.changeset/small-ravens-help.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
+fix(api-reference): use overflow-wrap break-word for tag labels


packages/api-reference/src/components/Anchor/Anchor.vue
```diff
@@ -14,7 +14,7 @@ const labelId = useId()
 const { cx } = useBindCx()
 </script>
 <template>
-  <span v-bind="cx('group/heading word-break-all relative')">
```
hwkr, line 17: I think this was supposed to apply `word-break: break-word;` or `word-break: all;` but this isn't the right Tailwind class so it wasn't doing anything (if it had applied either of those it would have broken wrong anyway).

packages/api-reference/src/components/Anchor/Anchor.vue
```diff
@@ -14,7 +14,7 @@ const labelId = useId()
...
+  <span v-bind="cx('group/heading wrap-break-word relative')">
```
hwkr, line 17: This applies `overflow-wrap: break-word;` which:

> In contrast to [word-break](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/word-break), overflow-wrap will only create a break if an entire word cannot be placed on its own line without overflowing.

[MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/overflow-wrap#:~:text=Note%3A%20In%20contrast%20to%20word%2Dbreak%2C%20overflow%2Dwrap%20will%20only%20create%20a%20break%20if%20an%20entire%20word%20cannot%20be%20placed%20on%20its%20own%20line%20without%20overflowing.)

marclave (approved):

PR #9125, opened 2026-05-06
chore(scalar-app): add better analytics to scalar app

amritk:
## Problem

closes #DOC-5245

Previously we were tracking a few things.

## Solution

Now we track all event bus events. 

Also added a glob pattern to the event bus and payload sanitization.

## Checklist

- [x] I added a changeset. 
- [x] I added tests.

---

> [!NOTE]
> **Medium Risk**
> Adds a new `onAny`/`offAny` wildcard subscription path to the workspace event bus and rewires plugin event handling to use it, which could change event delivery order/volume. Also expands analytics capture surface (with an allowlist/sanitization layer), so mistakes could still impact privacy or noise levels.
> 
> **Overview**
> Adds `WorkspaceEventBus.onAny/offAny` with tagged-union typing and deterministic ordering (exact listeners run before wildcard), plus tests for unsubscribe, debouncing, and error isolation.
> 
> Refactors client plugin event wiring to a single wildcard `ClientPlugin.on` listener via new `subscribePluginEvents`, and updates `scalar-app` and modal setup to use it.
> 
> Expands PostHog analytics: replaces per-event subscriptions with a wildcard handler, adds `sanitizeEventPayload` allowlist/property extraction to prevent leaking payload fields, and handles `log:user-login` via `identify` and `log:user-logout` via `reset` (with new/updated tests).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 56b33cdad9938d1ae64cadf922943c15979f6274. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/api-client/src/plugins/posthog/index.ts
```diff
@@ -27,17 +30,19 @@ export const PostHogClientPlugin = (config: PostHogConfig): ClientPlugin => {
...
-      'document:create:empty-document': () => posthog?.capture('document:create:empty-document'),
-      'document:delete:document': () => posthog?.capture('document:delete:document'),
-      'tag:create:tag': () => posthog?.capture('tag:create:tag'),
-      'server:add:server': () => posthog?.capture('server:add:server'),
-      'auth:update:selected-security-schemes': () => posthog?.capture('auth:update:selected-security-schemes'),
-      'environment:upsert:environment': () => posthog?.capture('environment:upsert:environment'),
-      'ui:open:client-modal': () => posthog?.capture('ui:open:client-modal'),
-      'ui:download:document': () => posthog?.capture('ui:download:document'),
+      'log:user-login': ({ uid, email, teamUid }) => {
+        posthog?.identify(uid, { email, teamUid })
```
DemonHa: Can we move it to the global matcher?

packages/api-client/src/plugins/posthog/sanitize-payload.ts
```diff
@@ -0,0 +1,56 @@
+/**
+ * Extracts a safe, flat set of analytics properties from a raw event payload.
+ *
+ * Only a known whitelist of fields is forwarded to PostHog — no PII, no
+ * request bodies, no auth secrets. Nested values are accessed defensively
+ * so any event shape that does not carry a given field simply omits it.
+ */
+export const sanitizePayload = (payload: unknown): Record<string, unknown> => {
```
DemonHa: Do we need to pass in the payload to posthog?

packages/workspace-store/src/events/bus.ts
```diff
@@ -6,6 +6,71 @@ import type { ApiReferenceEvents } from './definitions'
...
+ *
+ * - `'*'` matches every event.
+ * - A pattern ending with `':*'` (e.g. `'operation:*'`) matches every event
+ *   whose name starts with that prefix (e.g. `'operation:send:request:hotkey'`).
+ *
+ * @example
+ * bus.onAny('*', (event, payload) => console.log(event, payload))
+ * bus.onAny('operation:*', (event, payload) => console.log(event, payload))
+ */
+export type EventGlob = '*' | `${string}:*`
```
DemonHa: This can be any string actually it doesn't have to contain the * at the end

DemonHa (changes requested):

packages/api-client/src/plugins/posthog/sanitize-payload.ts
amritk: we want some things

packages/api-client/src/plugins/posthog/index.ts
amritk: this gives you types haha I had it the other way before. Sure can move it back

packages/workspace-store/src/events/bus.ts
amritk: we get much better perf with the simplified listener, lets only expand it if we need to

packages/oas-utils/src/helpers/client-plugins.ts
```diff
@@ -121,12 +121,87 @@ export type ClientPlugin = {
...
+  const unsubscribes: (() => void)[] = []
+
+  // Exact-key handlers — a single glob listener dispatches to the correct
+  // handler by looking up the event name, avoiding N individual subscriptions
+  unsubscribes.push(
+    eventBus.onGlob('*', (event: keyof ApiReferenceEvents, payload: ApiReferenceEvents[keyof ApiReferenceEvents]) => {
+      const handler = on[event] as ((payload: ApiReferenceEvents[keyof ApiReferenceEvents]) => void) | undefined
+      handler?.(payload)
+    }),
+  )
```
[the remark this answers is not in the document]
amritk: yep should fix this

packages/api-client/src/plugins/posthog/sanitize-payload.ts
DemonHa: can we specifically choose which payloads we want to pass through based on the event names?? That way we don't slip in any data we don't want?

DemonHa (changes requested):

packages/api-client/src/plugins/posthog/sanitize-payload.ts
amritk: sure we can do that, but I think I used common terms here across events. Like we use format and type a lot the same way

packages/oas-utils/src/helpers/client-plugins.ts
```diff
@@ -121,12 +121,95 @@ export type ClientPlugin = {
...
+    return () => {
+      // no-op
+    }
+  }
+
+  const on = plugin.on
+  const unsubscribes: (() => void)[] = []
+
+  const keys = Object.keys(on)
+  const hasExactKeyHandlers = keys.some((key) => !key.includes('*'))
```
DemonHa: instead of doing only .on and supporting glob in there maybe we can create a separate interface something like `.onAll()` which will capture `.onGlob("*")` or expose `.onGlob` so we can skip this part

packages/workspace-store/src/events/bus.ts
DemonHa: how do we get a better perf here? we do have a separate handler for glob events so I think it should be fine

packages/workspace-store/src/events/bus.ts
```diff
@@ -6,6 +6,71 @@ import type { ApiReferenceEvents } from './definitions'
...
+/**
+ * A glob pattern for subscribing to multiple events at once.
+ *
+ * - `'*'` matches every event.
+ * - A pattern ending with `':*'` (e.g. `'operation:*'`) matches every event
+ *   whose name starts with that prefix (e.g. `'operation:send:request:hotkey'`).
+ *
+ * @example
+ * bus.onAny('*', (event, payload) => console.log(event, payload))
+ * bus.onAny('operation:*', (event, payload) => console.log(event, payload))
```
DemonHa: is it called `.onAny` or `.onGlob`

DemonHa (changes requested):

packages/workspace-store/src/events/bus.ts
amritk: Haha I thought you meant handle proper globs but just all events is much better then the consumer can handle 😁

DemonHa: oh we are calling it `onGlob` tho...

If we don't want to support glob pattern on regex let's not call it like that then

amritk: Yea it should just be all events and the consumer handles it. Much better ty

packages/api-client/src/plugins/posthog/sanitize-event-payload.test.ts
```diff
@@ -0,0 +1,223 @@
+import type { ApiReferenceEvents } from '@scalar/workspace-store/events'
+import { describe, expect, it } from 'vitest'
+
+import { TRACKED_EVENTS, sanitizeEventPayload } from './sanitize-event-payload'
+
+/**
+ * Bypass the generic payload constraint for tests that deliberately pass
+ * malformed data to verify runtime defensive behaviour.
+ */
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
```
DemonHa: why do you have to define it like this?

packages/api-client/src/plugins/posthog/sanitize-event-payload.ts
```diff
@@ -0,0 +1,212 @@
...
+  return typeof value === 'string' ? value : undefined
+}
+
+const extractNestedString = (obj: unknown, outerKey: string, innerKey: string): string | undefined => {
+  if (!obj || typeof obj !== 'object') {
+    return undefined
+  }
+  const nested = (obj as Record<string, unknown>)[outerKey]
+  return extractString(nested, innerKey)
+}
```
DemonHa: can we just use the helpers we have for the bundler..

we have something called `getValueByPath`

DemonHa (changes requested):
Can we switch to the shared helpers?

Other than that looks good 😊

packages/api-client/src/plugins/posthog/sanitize-event-payload.ts
amritk: we should move them to helpers

packages/api-client/src/plugins/posthog/sanitize-event-payload.test.ts
amritk: for type safety

packages/api-client/src/plugins/posthog/sanitize-event-payload.ts
amritk: lol it is in helpers

DemonHa (approved):
LGTM

PR #9126, opened 2026-05-07
fix(use-codemirror): codemirror performance issues

amritk:
## Problem

Codemirror is eating up a lot of cpu cycles which is exacerbated on react.

## Solution

I saw something in the console and decided to try and memoize some of the watch triggers. This is the same issue that bites us time and time again of creating new objects and feeding into a watcher. 

Also isReadOnly wasn't even used as its handled in a div

Lets give this a shot

## Checklist

- [x] I added a changeset. 
- [x] I added tests.

---

> [!NOTE]
> **Medium Risk**
> Touches CodeMirror extension reconfiguration and plugin update behavior; a mistake could cause stale decorations or missed reconfigures, but changes are narrowly scoped and covered by new regression/perf tests.
> 
> **Overview**
> Reduces CodeMirror CPU churn by **avoiding redundant `StateEffect.reconfigure` dispatches** during mount and on unrelated reactive updates. `useCodeMirror` now watches a per-field list of primitive sources (plus extensions array identity) instead of a computed config object, and no longer runs the reconfigure watch immediately.
> 
> In `CodeInput`, the CodeMirror extensions list is made a stable constant and the environment “pill” ViewPlugin is instantiated once using an `environment` getter, preventing plugin replacement on prop identity changes. To keep pills fresh when the environment changes without editor interaction, `CodeInput` dispatches a no-op transaction to trigger the plugin update cycle.
> 
> Adds a changeset plus targeted performance/regression tests asserting stable extension references and ensuring environment/content updates do not trigger reconfiguration (while true config changes still do).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ef83d2a9f78db5cf095d4545e4f3cfda34c62d1b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/sparkly-worms-doubt.md
+++ b/.changeset/sparkly-worms-doubt.md
@@ -1,0 +1,6 @@
+---
+'@scalar/use-codemirror': patch
+'@scalar/api-client': patch
+---
+
+fix: codemirror performance issues
--- a/packages/api-client/src/v2/components/code-input/CodeInput.test.ts
+++ b/packages/api-client/src/v2/components/code-input/CodeInput.test.ts
@@ -1,6 +1,8 @@
+import { EditorView, StateEffect } from '@scalar/use-codemirror'
 import type { XScalarEnvironment } from '@scalar/workspace-store/schemas/extensions/document/x-scalar-environments'
 import { mount } from '@vue/test-utils'
-import { beforeEach, describe, expect, it, vi } from 'vitest'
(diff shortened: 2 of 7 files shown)


amritk:
This probably should work but let me see if we can do something drastic instead

PR #9133, opened 2026-05-07
feat(helpers): add load-css-variables to scalar helpers

hwkr:
Adds the `load-css-variables` helper to `@scalar/helpers` so it can be used for Monaco elsewhere. I did some quick testing with the theme but it seemed like the colors aren't showing up on prod right now in the editor? cc @DemonHa 

## Checklist

- [x] I added a changeset. 
- [x] I added tests.


packages/helpers/src/theme/load-css-variables.ts
```diff
@@ -1,5 +1,3 @@
-import { z } from 'zod'
-
```
amritk, line 2: 🔥

amritk (approved):

PR #9137, opened 2026-05-08
feat(api-reference): add ESM standalone bundle alongside UMD

### LINKED ISSUE #5743: ESM CDN
There is an example for an ESM CDN, but it’s transferring more than 5 MB.

Go to `packages/api-reference/playground/esm/index.html` and comment out:

```
import { createApiReference } from 'https://esm.sh/@scalar/api-reference'
```

Then run the playground:

```
cd packages/api-reference
pnpm playground:esm
```

If we can get this down to roughly the same size of the UMD build, we can start promoting the future.

It loads ajv (huge), so it might be worth trying to remove the wildcard exports from the parser. We don’t actually need ajv.

See #5734

amritk:
## Summary

closes #5743

Adds an ESM standalone build (`dist/browser/standalone.esm.js`) alongside the existing UMD bundle, then layers code-splitting and tree-shaking improvements on top of it. The new bundle works as a side-effect script (registers `window.Scalar.createApiReference`, reads `data-*` configuration) **and** exports `createApiReference` for direct ESM consumers, so the same file covers both legacy `<script type="module">` usage and `import { createApiReference } from '...'`.

UMD is unchanged in behaviour — UMD can't code-split, so its dynamic imports get inlined back into the single file.

## Bundle layout (ESM standalone)

```
Sync (loaded with the entry, in parallel):
  standalone.esm.js                              577 KB
  chunks/map-hidden-clients-config-*.js        1,941 KB
  chunks/local-storage-*.js                      120 KB
  chunks/browser-*.js                            104 KB
  chunks/runtime-core.esm-bundler-*.js            71 KB
  chunks/interface-content-folder.svg-*.js         1 KB   ← default sidebar icon, eagerly bundled
  chunks/rolldown-runtime-*.js                     1 KB
  ────────────────────────────────────────────────────
  TOTAL SYNC                                  ~2,816 KB

Lazy (fetched on demand):
  chunks/modal-*.js                              265 KB   ← API client modal + CodeMirror
  chunks/AgentScalarChatInterface-*.js           200 KB   ← AI agent UI
  83 × per-icon chunks                            ~92 KB total (~1.1 KB each)
  ────────────────────────────────────────────────────
  TOTAL LAZY (worst case)                       ~558 KB
```

| | UMD (today) | New ESM | Δ |
|---|---|---|---|
| Initial sync | ~3,479 KB | **~2,816 KB** | **−663 KB** |
| Total worst case | ~3,479 KB | ~3,373 KB | −106 KB |

For a typical document the user only fetches the modal chunk (when they open Test Request) plus a handful of icons (whichever the sidebar actually displays), so real-world page weight is meaningfully lower.

## What's lazy

**API client modal (~265 KB)** — `ApiReference.vue` was statically importing `createApiClientModal` from `@scalar/api-client/modal`; that module pulls in CodeMirror, the request editor, the response viewer, and the auth/header/body editors. Replaced with `await import('@scalar/api-client/modal')` inside `onMounted` so the modal chunk loads in the background after first paint. Existing optional-chaining on `apiClient.value?.route(...)` already handles "modal not ready yet" gracefully.

To keep this static import out of the entry graph, `Content.vue` now imports `mapHiddenClientsConfig` from a new `@scalar/api-client/modal/map-hidden-clients-config` deep export rather than the modal barrel, and `@scalar/api-client` declares `"sideEffects": ["**/*.css"]` so downstream bundlers can tree-shake the barrel themselves.

**AgentScalar chat interface (~200 KB)** — already wrapped in `defineAsyncComponent`, but with `codeSplitting: false` it was being inlined. Now it gets its own chunk that only loads when the agent is enabled (default: only on local URLs).

## Icons

`@scalar/icons/library` uses `import.meta.glob('./icons/*.svg', { eager: false })` so every SVG becomes a lazy chunk. The standalone bundle takes advantage of that:

- Each of the 84 library SVGs is its own ~1 KB chunk, fetched on demand as the sidebar renders. A typical OpenAPI doc references only a handful, so most chunks never load.
- The default fallback `interface-content-folder` (used by every sidebar item without a custom `x-scalar-icon`) is eagerly bundled via a second `import.meta.glob(..., { eager: true })` and the icon cache is pre-populated so `getLibraryIcon('interface-content-folder')` resolves synchronously. No HTTP delay or "no icon yet" frame on first sidebar render.

I considered grouping the 84 SVGs into a single ~125 KB chunk — that's smaller in worst-case total (−39 KB) but always loads everything; per-icon nets a smaller real-world page weight.

## Tree-shaking

- `@scalar/api-reference` declares `"sideEffects": ["**/*.css", "./dist/browser/**"]` — only CSS and the pre-bundled standalone files are side-effectful; the rest of the library is pure. Verified with esbuild that importing a single helper produces a ~491 byte bundle (vs ~3.67 MB when importing `createApiReference`), proving unused exports are dead-code-eliminated.
- `@scalar/api-client` declares `"sideEffects": ["**/*.css"]` for the same reason.
- Both standalone configs (UMD and ESM) set `treeshake.moduleSideEffects: id => id.includes('.css')` to match the default lib config.
- ESM standalone uses Rolldown's native `output.minify: true` to bypass Vite's library-mode default of `minifyWhitespace: false`, so the ESM file is genuinely minified instead of pretty-printed-but-mangled.

## Test plan

- [ ] `pnpm --filter @scalar/api-reference build` succeeds and emits `dist/browser/standalone.esm.js` plus the chunks/ directory.
- [ ] `pnpm --filter @scalar/api-reference test` (standalone unit tests) passes.
- [ ] `pnpm --filter @scalar/api-reference types:check` is clean.
- [ ] `pnpm --filter @scalar/icons test` passes.
- [ ] Manually load the standalone ESM bundle in a browser, confirm the API reference renders and the modal opens on "Test Request" (modal chunk fetches lazily in the network panel).
- [ ] Verify that custom `x-scalar-icon` values still resolve (lazy per-icon fetch) and that items without a custom icon use `interface-content-folder` immediately.

https://claude.ai/code/session_01J6EEZ5mBRX4rH4H1NRCN4r

### DIFF
--- a/.changeset/esm-standalone-bundle.md
+++ b/.changeset/esm-standalone-bundle.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
+Add an ESM standalone build (`dist/browser/standalone.esm.js`) alongside the existing UMD bundle. The new bundle works as a side-effect script (registers `window.Scalar.createApiReference` and reads `data-*` configuration) and exports `createApiReference` for direct ESM consumers. It is fully minified through Rolldown's native minifier so the total size is on par with — slightly smaller than — the UMD bundle.
--- a/.changeset/esm-standalone-bundle.md
+++ b/.changeset/esm-standalone-bundle.md
@@ -3,3 +3,5 @@
 ---
 
 Add an ESM standalone build (`dist/browser/standalone.esm.js`) alongside the existing UMD bundle. The new bundle works as a side-effect script (registers `window.Scalar.createApiReference` and reads `data-*` configuration) and exports `createApiReference` for direct ESM consumers. It is fully minified through Rolldown's native minifier so the total size is on par with — slightly smaller than — the UMD bundle.
+
+Also declares `sideEffects` in `package.json` so downstream bundlers can tree-shake unused exports from the regular ESM library build. Only CSS files and the standalone CDN bundles in `dist/browser/` are marked as having side effects; everything else is pure.
--- a/.changeset/esm-standalone-bundle.md
+++ b/.changeset/esm-standalone-bundle.md
@@ -1,7 +1,14 @@
 ---
 '@scalar/api-reference': patch
+'@scalar/api-client': patch
 ---
 
-Add an ESM standalone build (`dist/browser/standalone.esm.js`) alongside the existing UMD bundle. The new bundle works as a side-effect script (registers `window.Scalar.createApiReference` and reads `data-*` configuration) and exports `createApiReference` for direct ESM consumers. It is fully minified through Rolldown's native minifier so the total size is on par with — slightly smaller than — the UMD bundle.
+Add an ESM standalone build (`dist/browser/standalone.esm.js`) alongside the existing UMD bundle. The new bundle works as a side-effect script (registers `window.Scalar.createApiReference` and reads `data-*` configuration) and exports `createApiReference` for direct ESM consumers. It is fully minified through Rolldown's native minifier and uses code splitting so heavy features load asynchronously after first paint:
 
-Also declares `sideEffects` in `package.json` so downstream bundlers can tree-shake unused exports from the regular ESM library build. Only CSS files and the standalone CDN bundles in `dist/browser/` are marked as having side effects; everything else is pure.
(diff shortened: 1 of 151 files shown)


amritk:
Haven't tested it yet but it might be pretty close

amritk:
Going to split this one up

hanspagel:
cool cool cool, thanks for splitting it up, I think we can close this here

PR #9144, opened 2026-05-08
fix(api-reference): better search results for parameter, body, and schemas

### LINKED ISSUE #8937: searching param names doesn't always surface the right endpoint
"Search could be smarter searching param names doesn't always surface the right endpoint."

https://x.com/aminnnn_09/status/2046256128724541521

hanspagel:
Search couldn't find endpoints by parameter name reliably. A few things piled up:

- Parameter and body field names were buried inside a noisy string (`"limit optional query integer Maximum number of results"`) at the lowest weight, so any unrelated description would outrank an exact name match.
- The walker that builds the index didn't follow `oneOf` / `anyOf` / `allOf`, so polymorphic bodies (like Galaxy's `CelestialBody`) had zero properties indexed.
- Model entries didn't index their property names at all — searching `diameter` couldn't find the `Satellite` model.

This PR splits clean names into their own high-weight fields, drops the noise tokens, walks composition keywords (with `$ref` resolution and a cycle guard), and applies the same treatment to model schemas.

**Before**
Screenshot 2026-05-08 at 14 48 31

**After**
Screenshot 2026-05-08 at 14 43 13

**Before**
Screenshot 2026-05-08 at 14 36 16

**After**
Screenshot 2026-05-08 at 14 36 11

## Ticket

Closes #8937

### DIFF
--- a/.changeset/clever-keys-rank.md
+++ b/.changeset/clever-keys-rank.md
@@ -2,4 +2,4 @@
 '@scalar/api-reference': patch
 ---
 
-fix(api-reference): improve search ranking for parameter and request-body field names
+fix(api-reference): improve search ranking for parameter and request-body field names, including polymorphic (oneOf/anyOf/allOf) bodies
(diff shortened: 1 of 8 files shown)


hanspagel:
this is just better, let's roll it out

PR #9145, opened 2026-05-08
fix(api-client): send nested object multipart parts as JSON

hanspagel:
When a `multipart/form-data` request body had a property of type `object` (like a `props` field with sub-fields), we were flattening it into multiple parts with dotted keys (`props.name`, `props.description`, …). Servers don't expect that — the OpenAPI spec says complex properties in multipart should be sent as a single JSON-encoded part.

Now we send one part per object, encoded as JSON. So instead of three garbled `props.*` form fields, you get one clean `props={"name":"…","description":"…"}` part with `Content-Type: application/json`.

**Before**

```bash
curl /widget/v1/widgets \
  --request POST \
  --header 'Content-Type: multipart/form-data' \
  --form 'file=@filename' \
  --form 'props.name=' \
  --form 'props.description='
```

**After**

```bash
curl https://void.scalar.com/widget/v1/widgets \
  --request POST \
  --header 'Content-Type: multipart/form-data' \
  --form 'file=@filename' \
  --form 'props={
  "name": "",
  "description": "",
  "created_at": null
};type=application/json'
```

## More examples

While fixing the original report we noticed `process-body.ts` ignored the Encoding Object's `style` / `explode` / `allowReserved` and `contentType` fields almost entirely. The PR now implements them per OAS 3.1.x §Encoding Object for both `multipart/form-data` and `application/x-www-form-urlencoded`. A few representative cases:

### 1. `encoding.style: form` + `explode: true` on a nested object

Many real-world specs use `style: form, explode: true` to declare that an object property should be split into one form field per inner key (no parent prefix), mirroring how query parameters expand.

```yaml
content:
  multipart/form-data:
    schema:
      type: object
      properties:
        props:
          type: object
          properties:
            name: { type: string, example: widget }
            count: { type: integer, example: 3 }
    encoding:
      props:
        style: form
        explode: true
```

Generates:

```bash
curl https://example.com/widgets \
  --request POST \
  --header 'Content-Type: multipart/form-data' \
  --form 'name=widget' \
  --form 'count=3'
```

Inner keys become part names; the outer `props` name disappears, matching the Style Examples table (`?R=100&G=200&B=150`).

### 2. `encoding.style: deepObject` for nested settings

A common pattern for "settings" or "filter" objects where the API wants `parent[child]` bracket notation:

```yaml
content:
  multipart/form-data:
    schema:
      type: object
      properties:
        props:
          type: object
          properties:
            name: { type: string, example: widget }
            meta:
              type: object
              properties:
                region: { type: string, example: eu }
    encoding:
      props:
        style: deepObject
        explode: true
```

Generates:

```bash
curl https://example.com/widgets \
  --request POST \
  --header 'Content-Type: multipart/form-data' \
  --form 'props[name]=widget' \
  --form 'props[meta][region]=eu'
```

Arbitrary nesting via brackets — the spec only defines one level, but we recurse so deeper structures still reach the wire (everything beyond level one is implementation-defined per OAS §Appendix C).

### 3. `encoding.contentType` with a `+json` vendor media type

JSON:API, HAL, and other API styles use `application/vnd.*+json` (RFC 6839). When an author overrides the part's contentType to such a type, the request goes out with that header **and** the cURL snippet now pretty-prints the body.

```yaml
content:
  multipart/form-data:
    schema:
      type: object
      properties:
        settings:
          type: object
          properties:
            enabled: { type: boolean, example: true }
    encoding:
      settings:
        contentType: application/vnd.api+json
```

Generates:

```bash
curl https://example.com \
  --request POST \
  --header 'Content-Type: multipart/form-data' \
  --form 'settings={
  "enabled": true
};type=application/vnd.api+json'
```

Previously the `--form` value would have stayed on one line because the cURL plugin only matched the exact string `application/json`. The detection now follows RFC 6839 `+json` and parameterized variants like `application/json;charset=utf-8`.

### 4. `style` on `application/x-www-form-urlencoded`

Per spec, `encoding.style` applies to urlencoded bodies just as it does to multipart. Previously this was a silent no-op; we now route through the same RFC6570 serializers used for multipart and query parameters.

```yaml
content:
    schema:
      type: object
      properties:
        props:
          type: object
          properties:
            name: { type: string, example: widget }
            count: { type: integer, example: 3 }
    encoding:
      props:
        style: form
        explode: true
```

Generates:

```bash
curl https://example.com \
  --request POST \
  --header 'Content-Type: application/x-www-form-urlencoded' \
  --data-urlencode 'name=widget' \
  --data-urlencode 'count=3'
```

Without an explicit `encoding` block, urlencoded bodies still use the legacy dotted-key flattening (`props.name=widget`) for backwards compatibility — opt into spec-correct behavior by declaring `style`.

### 5. `style: form, explode: false` on an array

The "single comma-joined part" form, which some Rails / PHP backends parse natively:

```yaml
content:
  multipart/form-data:
    schema:
      type: object
      properties:
        tags:
          type: array
          items: { type: string }
          example: [a, b, c]
    encoding:
      tags:
        style: form
        explode: false
```

Generates:

```bash
curl https://example.com \
  --request POST \
  --header 'Content-Type: multipart/form-data' \
  --form 'tags=a,b,c'
```

## Ticket

See #4834

---

> [!NOTE]
> **Medium Risk**
> Changes request-body and query parameter serialization for `multipart/form-data` and `application/x-www-form-urlencoded`, which can alter the wire shape sent to servers and generated snippets. Risk is mitigated by extensive new test coverage, but regressions are possible for consumers relying on the previous dotted-key flattening behavior.
> 
> **Overview**
> **Aligns form body serialization with OpenAPI 3.1 Encoding Object rules.** `process-body` now (1) emits top-level multipart object/complex array values as a single JSON part (defaulting to `application/json`), (2) supports `encoding.style`/`explode`/`allowReserved` for both multipart and urlencoded bodies by routing through RFC6570 serializers, and (3) ignores `encoding.contentType` on `application/x-www-form-urlencoded` when it’s the only encoding hint.
> 
> **Improves generated cURL snippets for JSON bodies/parts.** The cURL plugin now detects JSON via `application/*+json` and parameterized JSON media types and pretty-prints JSON for both `--data` bodies and multipart `--form` parts when the value parses as JSON.
> 
> **Schema/types updated.** The workspace-store `EncodingObject` schema/type now exposes `style`, `explode`, and `allowReserved`, with new/updated tests covering the new serialization behavior and deepObject fallbacks.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dafb4b0a4e027a1f2f1dec2089b14a428399fd1e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/api-client/src/v2/blocks/operation-code-sample/helpers/operation-to-har/process-body.ts
+++ b/packages/api-client/src/v2/blocks/operation-code-sample/helpers/operation-to-har/process-body.ts
@@ -26,6 +26,18 @@ type ProcessBodyProps = Pick<OperationToHarProps, 'contentType' | 'example' | 'r
 type MultipartEncodingMap = MediaTypeObject['encoding']
 
 /**
+ * Stringify a value emitted by `serializeFormStyle` into a HAR param value.
+ *
+ * Form-style serialization only addresses one level of nesting per RFC6570;
+ * deeper structures (an object or array still sitting in `entry.value`) are
+ * spec-undefined. JSON-stringify them so authors get readable output instead
+ * of `String(value)` returning `"[object Object]"`. Primitives pass through
+ * `String()` to preserve the existing wire shape (e.g. `true` → `"true"`).
+ */
+const stringifyEntryValue = (value: unknown): string =>
+  value !== null && typeof value === 'object' ? JSON.stringify(value) : String(value)
+
+/**
  * Converts a form-data body into HAR `Param[]` entries for `multipart/form-data`
  * and `application/x-www-form-urlencoded` requests.
  *
@@ -109,7 +121,7 @@ const objectToFormParams = (
           const serialized = serializeFormStyle(unpacked, true)
           if (Array.isArray(serialized)) {
             for (const entry of serialized) {
-              params.push({ name: entry.key || key, value: String(entry.value) })
+              params.push({ name: entry.key || key, value: stringifyEntryValue(entry.value) })
             }
           } else {
             params.push({ name: key, value: String(serialized) })
@@ -131,7 +143,11 @@ const objectToFormParams = (
           for (const entry of serialized) {
             // Arrays: entry.key === '' → fall back to the outer name.
             // Objects: entry.key is the inner property name (spec strips the outer name).
-            params.push({ name: entry.key || key, value: String(entry.value) })
+            // Nested objects/arrays inside entry.value are spec-undefined (RFC6570 form-style
+            // only addresses one level of nesting); JSON-stringify them instead of letting
+            // String() emit "[object Object]". The escape hatch for cleaner output is
+            // style: deepObject + explode: true.
+            params.push({ name: entry.key || key, value: stringifyEntryValue(entry.value) })
           }
         } else {
           params.push({ name: key, value: String(serialized) })
--- a/packages/api-client/src/v2/blocks/operation-code-sample/helpers/operation-to-har/process-body.ts
+++ b/packages/api-client/src/v2/blocks/operation-code-sample/helpers/operation-to-har/process-body.ts
@@ -27,6 +27,7 @@ const objectToFormParams = (
   obj: object | { name: string; value: unknown; isDisabled: boolean }[],
   encoding?: MultipartEncodingMap,
   parentKey?: string,
+  isMultipart = false,
 ): Param[] => {
   const params: Param[] = []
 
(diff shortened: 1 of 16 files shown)


amritk (approved):
Should this be determined by style?

hanspagel:
absolutely! good catch

packages/workspace-store/src/schemas/v3.1/strict/encoding.ts
```diff
@@ -16,6 +16,12 @@ export const EncodingObjectSchemaDefinition = Type.Object({
   contentType: Type.Optional(Type.String()),
   /** A map allowing additional information to be provided as headers. Content-Type is described separately and SHALL be ignored in this section. This field SHALL be ignored if the request body media type is not a multipart. */
   headers: Type.Optional(Type.Record(Type.String(), Type.Union([HeaderObjectRef, reference(HeaderObjectRef)]))),
+  /** Describes how a specific property value will be serialized depending on its type. See the Parameter Object for details on the style field. The behavior follows the same values as query parameters, including default values. Valid values are "form", "spaceDelimited", "pipeDelimited", and "deepObject". This field SHALL be ignored if the request body media type is not application/x-www-form-urlencoded or multipart/form-data. If a value is explicitly defined, then the value of contentType (implicit or explicit) SHALL be ignored. */
+  style: Type.Optional(Type.String()),
```
amritk:
```suggestion
style: Type.Optional(Type.Union([
  Type.Literal('form'),
  Type.Literal('spaceDelimited'),
  Type.Literal('pipeDelimited'),
  Type.Literal('deepObject')
])),
```
why is this a string? Lets put in the actual values here

packages/api-client/src/v2/blocks/operation-code-sample/helpers/operation-to-har/process-body.ts
```diff
@@ -40,19 +76,117 @@ const objectToFormParams = (
       continue
     }
 
-    const partContentType = parentKey ? undefined : encoding?.[key]?.contentType
+    const partEncoding = parentKey ? undefined : encoding?.[key]
+    // Per OpenAPI 3.1.1: when style, explode, or allowReserved is explicitly set on the
+    // encoding entry, contentType (implicit or explicit) is ignored and the value is
+    // serialized as if it were a query-style parameter.
```
amritk: just use comment blocks for multi line comments so you don't have all these extra slashes!

```
/**
 * Hello
 */
```

amritk: also I find the slashes are a bit harder to read when stacked like this since all the slashes are blurring together 👓

amritk (approved):
Nice nailed it I think, from what I could follow + comments

packages/api-client/src/v2/blocks/operation-code-sample/helpers/operation-to-har/process-body.ts
```diff
@@ -40,19 +76,117 @@ const objectToFormParams = (
...
+    const hasFormStyle =
+      !!partEncoding &&
+      (partEncoding.style !== undefined ||
+        partEncoding.explode !== undefined ||
+        partEncoding.allowReserved !== undefined)
```
amritk, line 89: haha I hate this big ifs, this looks slightly better imo

```suggestion
const hasFormStyle = partEncoding && (
  partEncoding.style !== undefined ||
  partEncoding.explode !== undefined ||
  partEncoding.allowReserved !== undefined
)
```

PR #9155, opened 2026-05-11
feat(api-client): edit nested multipart object properties as individual form rows

### LINKED ISSUE #4834: Scalar misinterprets multipart/form-data with part that contains structured JSON

hanspagel:
Nested object properties in `multipart/form-data` schemas used to collapse into a single JSON-stringified field. Now each leaf gets its own editable row (`props.name`, `props.description`, `props.created_at`), and dotted rows are regrouped into one `props` multipart part on send.

cURL preview also pretty-prints JSON inside `--form` parts, matching what we already do for `--data`.

**Before**
Screenshot 2026-05-11 at 11 26 02

**After**
Screenshot 2026-05-11 at 12 35 08

Fixes #4834

## OpenAPI Document

```yaml
openapi: 3.1.0
info:
  title: Multipart Nested Rows Test
  version: 1.0.0
paths:
  /widgets:
    post:
      summary: Create widget (multipart)
      requestBody:
        required: true
        content:
          multipart/form-data:
            schema:
              type: object
              required: [file, props]
              properties:
                file:
                  type: string
                  format: binary
                  description: File to upload
                props:
                  type: object
                  required: [name, description]
                  properties:
                    name: { type: string }
                    description: { type: string }
                    created_at: { type: string, format: date-time }
      responses:
        '200': { description: OK }

  /widgets-url:
    post:
      summary: Create widget (urlencoded) — same schema as multipart
      requestBody:
        required: true
        content:
            schema:
              type: object
              required: [token, props]
              properties:
                token: { type: string}
                props:
                  type: object
                  properties:
                    name: { type: string}
                    description: { type: string}
      responses:
        '200': { description: OK}
```

---

> [!NOTE]
> **Medium Risk**
> Moderate risk because it changes how multipart form rows are generated and serialized, including schema-driven regrouping of dotted keys into JSON parts; mistakes could alter request payload shape or field ordering.
> 
> **Overview**
> Enables **nested `multipart/form-data` object properties** to be edited as individual dotted-name rows (e.g. `props.name`) by walking the request-body schema + example and emitting one row per leaf, while preserving per-leaf schema metadata and *cascading* `required` semantics.
> 
> Updates request-body building for multipart so **edited dotted rows are regrouped back into a single top-level JSON text part** (one part per top-level object property), preserving row order and avoiding accidental folding of literal dotted names (e.g. `user.email`) unless the schema declares a nested object. Adds comprehensive tests for nesting depth, null/file handling, undeclared example keys, urlencoded non-expansion, and dotted-name edge cases; also tightens HAR param building by normalizing row-array names to strings.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 352faf2678c76db3caf7d55cddbfdc62f3c1a9ff. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/api-client/src/v2/blocks/operation-code-sample/helpers/operation-to-har/process-body.ts
+++ b/packages/api-client/src/v2/blocks/operation-code-sample/helpers/operation-to-har/process-body.ts
@@ -68,7 +68,7 @@ const objectToFormParams = (
 
   /** Ensure we do not include disabled items */
   const entries: [string, unknown][] = Array.isArray(obj)
-    ? obj.filter((item) => !item.isDisabled).map((item) => [String(item.name), item.value] as [string, unknown])
+    ? obj.filter((item) => !item.isDisabled).map((item) => [String(item.name), item.value])
     : Object.entries(obj)
 
   for (const [key, value] of entries) {
--- a/packages/api-client/src/v2/blocks/operation-code-sample/helpers/operation-to-har/process-body.ts
+++ b/packages/api-client/src/v2/blocks/operation-code-sample/helpers/operation-to-har/process-body.ts
@@ -67,8 +67,8 @@ const objectToFormParams = (
   const params: Param[] = []
 
   /** Ensure we do not include disabled items */
-  const entries = Array.isArray(obj)
-    ? obj.filter((item) => !item.isDisabled).map((item) => [item.name, item.value])
+  const entries: [string, unknown][] = Array.isArray(obj)
+    ? obj.filter((item) => !item.isDisabled).map((item) => [String(item.name), item.value] as [string, unknown])
     : Object.entries(obj)
 
   for (const [key, value] of entries) {
--- a/packages/api-client/src/v2/blocks/request-block/helpers/get-form-body-rows.ts
+++ b/packages/api-client/src/v2/blocks/request-block/helpers/get-form-body-rows.ts
@@ -117,8 +117,10 @@ export const getFormBodyRows = (
 
   // Schema-driven path: when the form body schema describes nested objects, emit one row
   // per leaf so users can edit individual fields. The dotted name (`props.name`) is folded
-  // back into a single JSON multipart part by `process-body.ts` before the request is sent.
-  if (leafByDottedName.size > 0 && typeof example.value === 'object') {
+  // back into a single JSON multipart part by `build-request-body.ts` before the request is
+  // sent. Gated to `multipart/form-data` because urlencoded has no regrouping step on send,
+  // so dotted leaves would otherwise reach the wire as separate `props.name`-style fields.
+  if (contentType === 'multipart/form-data' && leafByDottedName.size > 0 && typeof example.value === 'object') {
     return Array.from(leafByDottedName.values()).map(({ path }) => {
       const rawValue = getValueAtPath(example.value, path)
       // Missing values and explicit `null` (e.g. for a nullable schema) render as empty
--- a/packages/api-client/src/v2/blocks/request-block/helpers/get-form-body-rows.ts
+++ b/packages/api-client/src/v2/blocks/request-block/helpers/get-form-body-rows.ts
@@ -32,21 +32,33 @@ type LeafRow = {
  * (`file` + `props.{name,description,created_at}`) yields four leaves.
  */
 const collectLeafProperties = (schema: SchemaObject, parentPath: string[] = [], parentRequired = true): LeafRow[] => {
+  // Nothing to walk if this is not an object schema or has no declared properties.
   if (!isObjectSchema(schema) || !schema.properties) {
     return []
   }
+
+  // `required` is a list of property names on *this* schema, so build it once per level.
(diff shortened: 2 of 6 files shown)


packages/api-client/src/v2/blocks/operation-code-sample/helpers/operation-to-har/process-body.ts
```diff
@@ -31,8 +26,8 @@ const objectToFormParams = (
   const params: Param[] = []
 
   /** Ensure we do not include disabled items */
-  const entries = Array.isArray(obj)
-    ? obj.filter((item) => !item.isDisabled).map((item) => [item.name, item.value])
+  const entries: [string, unknown][] = Array.isArray(obj)
+    ? obj.filter((item) => !item.isDisabled).map((item) => [String(item.name), item.value] as [string, unknown])
```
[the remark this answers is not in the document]
hanspagel: covered in #9145

packages/api-client/src/v2/blocks/request-block/helpers/get-form-body-rows.ts
```diff
@@ -16,6 +17,39 @@ const stringifyValue = (value: unknown) => {
...
+  const requiredSet = new Set(schema.required ?? [])
+  const leaves: LeafRow[] = []
+  for (const [key, rawChildSchema] of objectEntries(schema.properties)) {
+    const childSchema = resolve.schema(rawChildSchema)
+    const path = [...parentPath, String(key)]
+    const isRequired = parentRequired && requiredSet.has(String(key))
+    if (childSchema && isObjectSchema(childSchema) && childSchema.properties) {
+      leaves.push(...collectLeafProperties(childSchema, path, isRequired))
+    } else {
+      leaves.push({ path, schema: childSchema, isRequired })
```
amritk, line 57: would be nice to space this out with some comments

packages/api-client/src/v2/blocks/operation-code-sample/helpers/operation-to-har/process-body.ts
amritk:
```suggestion
    ? obj.filter((item) => !item.isDisabled).map((item) => [String(item.name), item.value])
```

no need

amritk (changes requested):
If there's one thing I hate, its nested loops! It should be ok depending on the example size but lets flatten it out if we can, cursor should be able to do it. 
```
    const dottedRows = exampleValue.filter(({ name, value }) => isDottedNestedRow(name, value))
```

amritk (approved):
cool I think its good

marlon-claus:
This doesn't seem to work correctly anymore. It's split into multiple rows correctly but on send the properties don't seem to be merged into a json object on our end. This causes an error because our api needs one multipart row that contains a json string and not multiple ones.
Could this be a problem on our end? :D

PR #9169, opened 2026-05-11
fix(api-reference): forward fetch to API client

Fjandin:
## Problem

The configuration.fetch option was only used when fetching the OpenAPI document — the API client (which executes "Try Request") was reading a separate customFetch option that nothing in the api-reference layer set.

## Solution

Bridge the two so a user-supplied fetch is also used for outgoing API requests, enabling things like credentials: 'include'.

## Checklist

- [x] I added a changeset. 
- [x] I added tests.

### DIFF
--- a/.changeset/forward-custom-fetch-to-client.md
+++ b/.changeset/forward-custom-fetch-to-client.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
+fix: forward `configuration.fetch` to the API client so "Test Request" calls use the custom fetch (enabling things like `credentials: 'include'`), not only the OpenAPI document load
--- a/.changeset/forward-custom-fetch-to-client.md
+++ b/.changeset/forward-custom-fetch-to-client.md
@@ -1,5 +1,7 @@
 ---
 '@scalar/api-reference': patch
+'@scalar/schemas': patch
+'@scalar/types': patch
 ---
 
-fix: forward `configuration.fetch` to the API client so "Test Request" calls use the custom fetch (enabling things like `credentials: 'include'`), not only the OpenAPI document load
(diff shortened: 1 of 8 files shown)


amritk (changes requested):
Ah this is my mistake, maybe instead we should just call them both customFetch? or fetch but I like customFetch better. What you think @hanspagel @DemonHa

DemonHa:
> Ah this is my mistake, maybe instead we should just call them both customFetch? or fetch but I like customFetch better. What you think @hanspagel @DemonHa

yeah I would go with the `customFetch`

DemonHa (approved):
LGTM 🚀

Thank you for your contribution 🙏

Fjandin:
@DemonHa thanks. Any changes i could do from here on?

hanspagel:
yeah, let's call both customFetch
@Fjandin can you rename the configuration option for the api-reference?

Fjandin:
> yeah, let's call both customFetch
> 
> @Fjandin can you rename the configuration option for the api-reference?

Sure. Can do 👌🏻 Just to be clear, this would mean a breaking change

DemonHa:
@Fjandin we can probably just deprecate the `fetch` field for now and add a new once called `customFetch`. We can remove it on the next major release I think.

(cc @amritk)

DemonHa:
> @Fjandin we have a deprecation section here
> 
> https://github.com/scalar/scalar/blob/main/packages/types/src/api-reference/api-reference-configuration.ts#L440
> 
> so if you see .fetch used you can log a warning and transform it to customFetch

Actually that's the old schemas... we need to remove those @amritk 

[This](https://github.com/scalar/scalar/blob/7a308aa68873421ebb29bf37269fd80007bccec0/packages/types/src/api-reference/api-reference-configuration.ts#L440) is the new schemas we migrate to..

Fjandin:
Pushed a rename to customFetch while keeping the old fetch working through a deprecation path

Fjandin:
FYI @DemonHa

hanspagel (approved):
nice, thanks!

Issue #6257, opened 2025-07-15
Arbitrary additionalProperties not correctly displayed anymore


joe-akeem:
### What happens?

After Updating from Version 1.28.31 to 1.32.7, additional properties are not displayed as they used to be. As a user I don't understand what this means:

<img width="348" height="99" alt="Image" src="https://github.com/user-attachments/assets/57ab7937-0152-49dd-844a-f9815368b816" />

The generated example is also not correct:

<img width="250" height="102" alt="Image" src="https://github.com/user-attachments/assets/f3256e61-328a-400b-9ccf-9e49dfda4963" />

### What did you expect to happen?

Display of additional properties similar to what it was in version 1.28.31:

<img width="362" height="105" alt="Image" src="https://github.com/user-attachments/assets/007e365d-7450-49af-8a78-9030196c1752" />

An example that matches the specification such as
```json
{
  "phone_number": "string",
  "name1": null,
  "name2": "value2"
}
```

### OpenAPI Document

https://sandbox.scalar.com/e/MxGvU

hanspagel:
Thanks for the feedback, that’s super helpful! I need to think about that, we need to make those additional properties make look great for a lot of use cases.

dshchupak:
I am experiencing the same issue here using the latest C# library.

This is what appears to happen when the field is a dictionary/key-value-pair:

`"metadata": {
	"type": "object",
	"additionalProperties": {
		"type": "string"
	},
	"description": "Maximum 40 characters per key and 250 characters per value.",
	"nullable": true
}`

<img width="748" height="148" alt="Image" src="https://github.com/user-attachments/assets/bb6a0c89-dad0-41cc-960b-4c39fccf0fd3" />

In the example payload, it shows like so:

<img width="164" height="32" alt="Image" src="https://github.com/user-attachments/assets/0e24f546-051b-4ca3-903d-e4e83b8d774a" />

This previously showed like so:

<img width="745" height="136" alt="Image" src="https://github.com/user-attachments/assets/214791ab-ca8a-4c68-b71d-d5fbd324bf85" />

I would ask for you to consider having the example maybe show like swagger UI has it, which can be seen below:

<img width="246" height="84" alt="Image" src="https://github.com/user-attachments/assets/aad9e8b4-ea15-4562-817a-3d37d840cddc" />

bmacalindong:
is there any update on this?

SebastianBienert:
@hanspagel are there any plans for this adjustment?

[...]

PR #9192, opened 2026-05-13
fix(api-reference): resolve $refs in additional properties

wilhelmer:
Addresses #6257.

## Problem

When `additionalProperties` contains a `$ref` (e.g. pointing to an external schema file), the referenced schema's contents were not rendered — only "propertyName" was shown with no type information.

The root cause was in `getAdditionalPropertiesValue`: it checked for a `type` key on the raw `$ref` object, which never has one, causing it to always return `{ type: 'anything' }` instead of the actual referenced schema.

The fix resolves the schema via `resolve.schema()` before the type check, consistent with how `getAdditionalPropertiesName` already handles the same input.

## Changes
- `packages/api-reference/src/components/Content/Schema/SchemaObjectProperties.vue` — resolve `$ref` before checking for `type` in `getAdditionalPropertiesValue`
- `packages/api-reference/src/components/Content/Schema/SchemaObjectProperties.test.ts` — regression test for `additionalProperties` with a `$ref` value

### DIFF
--- a/.changeset/crisp-queens-glow.md
+++ b/.changeset/crisp-queens-glow.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
+fix: resolve $ref in additionalProperties before rendering schema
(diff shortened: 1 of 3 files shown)


hanspagel (approved):
LGTM

PR #9194, opened 2026-05-13
fix(api-reference): break cycles in mergeAllOfSchemas

jan-krueger:
## Problem

Rendering a schema whose array items $ref back to the schema itself (e.g. ValidationIssue with `children: { items: { $ref: '...ValidationIssue' } }`) crashed the docs preview with "too much recursion" (Firefox) / "Maximum call stack size exceeded" (Chrome/Node). mergeItems and mergePropertiesIntoResult mutually recursed without any cycle detection.

## Solution

Track in-flight $ref strings in mergeItems; when the incoming schema's $ref is already being processed higher up the call stack, return the partially merged existing schema. $ref is a stable identity even across resolve.schema() calls (which coerce into fresh objects) because mergeSiblingReferences preserves the original $ref on the resolved value.

## Checklist

- [x] I added a changeset. 
- [x] I added tests.

### DIFF
--- a/.changeset/merge-cyclic-recursive-schemas.md
+++ b/.changeset/merge-cyclic-recursive-schemas.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
+fix(api-reference): break cycles in `mergeAllOfSchemas` for self-referencing schemas (e.g. a schema whose array items `$ref` back to itself), which previously crashed the docs preview with "too much recursion"
--- a/.changeset/merge-cyclic-recursive-schemas.md
+++ b/.changeset/merge-cyclic-recursive-schemas.md
@@ -2,4 +2,4 @@
 '@scalar/api-reference': patch
 ---
 
-fix(api-reference): break cycles in `mergeAllOfSchemas` for self-referencing schemas (e.g. a schema whose array items `$ref` back to itself), which previously crashed the docs preview with "too much recursion"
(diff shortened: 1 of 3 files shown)


hanspagel:
thanks for the PR!
@DemonHa wdyt?

amritk (dismissed):
having the inFlightRefs at the module level is an issue since you can load multiple documents and you may get back incorrect or cross document contaminated results. 

You should pass in your cache something like this generated code but fix it for the case:

```ts
// Define a clean entry interface if needed, or pass an internal parameter
export const mergeAllOfSchemas = (
  schema: SchemaObject, 
  seenRefs: Set<string> = new Set()
): SchemaObject => {
  // ... core logic
}

// Update your internal helpers to forward the contextual state:
const mergeItems = (
  existing: SchemaObject, 
  incoming: SchemaObject, 
  seenRefs: Set<string>
): SchemaObject => {
  const incomingRef = (incoming as { $ref?: string }).$ref
  
  if (typeof incomingRef === 'string') {
    if (seenRefs.has(incomingRef)) {
      // Safely break cycle
      return { ...existing } 
    }
    // Create a local branch clone of the set to prevent cross-branch contamination
    const nextSeenRefs = new Set(seenRefs).add(incomingRef)
    return mergeItemsInner(existing, incoming, nextSeenRefs)
  }

  return mergeItemsInner(existing, incoming, seenRefs)
}
```

jan-krueger:
@amritk Thanks for the feedback. I've addressed it as suggested.

hanspagel (approved):
let's get this one finally in

amelnikoff:
Still same error https://github.com/scalar/scalar/issues/9442 after release with his merged

PR #9219, opened 2026-05-14
fix(ci): don't setup test servers since we don't need them

DemonHa:
Looks like we don't really need to run the test servers (proxy and void server) for the tests anymore

## Checklist

- [ ] I added a changeset. 
- [ ] I added tests.

---

> [!NOTE]
> **Medium Risk**
> Changes CI job topology and removes the explicit test-server startup, which could cause previously server-dependent tests to silently fail or hang if assumptions are wrong. Also adjusts scalar-app Vitest setup/globals, which may affect test behavior and masking of environment issues.
> 
> **Overview**
> CI no longer sets up Go or starts the local test servers as part of `test-packages`, and instead introduces a new `test-projects` job to run `pnpm` workspace tests under `projects/*` (excluding the Go proxy) and makes it required for PRs and `main`.
> 
> `scalar-app` test scripts are updated to use `vitest --run` (with a new `test:watch`), a test is hardened by clearing `mockEventBus.emit` between cases, and the Vitest setup now polyfills `CSS.escape` and `matchMedia` to prevent Monaco/jsdom-related unhandled rejections.
> 
> The Cloudflare Pages preview cleanup script is made macOS-compatible by replacing `mapfile` with a bash-3.2-safe read loop.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bafc6680cbe06a3e9ca8bd079b012d2431b3bce6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


tooling/scripts/delete-cloudflare-pages-preview.sh
```diff
@@ -103,7 +103,13 @@ while true; do
   page=$((page + 1))
 done
 
-mapfile -t deployment_ids < <(sort -u "$deployment_ids_file")
+# Avoid `mapfile` (bash 4+); macOS still ships bash 3.2 as /bin/bash.
```
amritk, line 106: this should only run in CI anyway?

projects/scalar-app/package.json
```diff
@@ -26,7 +26,8 @@
     "lint:fix": "biome check --write --unsafe && pnpm eslint \"**/*.vue\" --fix",
     "preview:app": "electron-vite preview",
     "preview:web": "vite preview",
-    "test": "vitest",
+    "test": "vitest --run",
+    "test:watch": "vitest --watch",
```
amritk, line 30:
```suggestion
    "test:watch": "vitest",
```
watch is default 😄

amritk (approved):

tooling/scripts/delete-cloudflare-pages-preview.sh
DemonHa: it fails when I run it locally

DemonHa: I mean when I run the tests

projects/scalar-app/package.json
DemonHa: it was not when I run it locally

tooling/scripts/delete-cloudflare-pages-preview.sh
amritk: right

marclave (approved):

PR #9222, opened 2026-05-14
chore(ci): stop relying on string checks

amritk:
## Problem

We were previously relying on string checks of "Releasing" to gate our prod releases. This broke a couple different ways and we couldn't release for a while.

## Solution

The plan is to eventually make this work off of releases but for now we switched to using an output. This should work better than the string (if it works)

## Checklist

- [ ] I added a changeset. 
- [ ] I added tests.

---

> [!NOTE]
> **Medium Risk**
> Changes the CI/CD release gating logic for npm, Docker, NuGet, PyPI, Maven, and production deploys; misconfiguration could block releases or deploy the wrong environment.
> 
> **Overview**
> Stops using commit-message string checks (e.g. `RELEASING:`) to trigger production publishing/deployments, and instead gates downstream release jobs on `changesets/action`’s `published` output from the `npm-publish` job.
> 
> Makes `npm-publish` always run on a GitHub-hosted runner (required for `npm publish --provenance`), exposes `published`/`publishedPackages` as job outputs, serializes workspace publishing to reduce npm registry 409s, and updates `deploy-api-client` to choose *production vs staging* based on whether a publish actually occurred (with explicit `needs.*.result` safety checks).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f88f0c599dc2a1761535d13c4cd0942056a63846. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1545,7 +1545,8 @@ jobs:
     concurrency:
       group: django-ninja-publish
       cancel-in-progress: false
-    needs: [required-jobs-main, django-ninja-build-test, check-changes, npm-publish]
+    needs:
+      [required-jobs-main, django-ninja-build-test, check-changes, npm-publish]
     permissions:
       contents: read
     steps:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -887,12 +887,21 @@ jobs:
   deploy-api-client:
     # Run for PR previews and every main push. On main, we wait for npm-publish
     # so we know whether this commit actually released and should deploy to
-    # production (otherwise staging). `!cancelled()` lets this job proceed when
-    # npm-publish is skipped (PRs) or succeeded-without-publish (main pushes
-    # that only opened/updated the release PR). A failed npm-publish blocks
-    # deploys so a broken publish never silently lands on staging.
+    # production (otherwise staging).
+    #
+    # `!cancelled()` lets this job proceed when npm-publish is skipped (PRs) or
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -617,8 +617,11 @@ jobs:
 
   npm-publish:
     if: github.ref == 'refs/heads/main'
-    # Use a GitHub-hosted runner for npm publish to ensure signed releases
-    runs-on: ${{ startsWith(github.event.head_commit.message, 'RELEASING:') && 'ubuntu-22.04' || 'blacksmith-4vcpu-ubuntu-2204' }}
+    # `npm publish --provenance` requires a GitHub-hosted runner. Self-hosted
+    # runners (including Blacksmith) are rejected by the npm registry with HTTP
+    # 422 because the OIDC issuer is not in npm's trusted list. Always use a
+    # github-hosted runner here, regardless of whether this run will publish.
+    runs-on: ubuntu-22.04
```
DemonHa, line 624: do we always want to run on GitHub-hosted runner?

DemonHa:
do we also want to use the cached build here if we are not going to publish?

.github/workflows/ci.yml
amritk: for npm publish yes, for the other one probably not but we can try this way first and if needed go back. I have many more plans for the CI anyway this might go by then

so worst case it gets a little slower before it gets much faster

DemonHa (approved):

PR #9223, opened 2026-05-15
fix(components): prevent app header overflow on mobile

DemonHa:
## What

The API client app header overflowed horizontally on narrow / mobile viewports.

`ScalarHeader` pinned itself to `min-w-min` and forced every section to equal
width (`*:flex *:flex-1`), so its content (menu + breadcrumb) could never shrink
and pushed the header past the viewport. This reproduces in both Chromium and
WebKit — at a 375px viewport the header measured ~551px wide.

## Fix

- `ScalarHeader`: drop `min-w-min` and the equal-width columns. When there is no
  center slot the start section grows and can shrink/ellipsize; when there is
  one the center grows and start/end hug their content.
- `ScalarMenu`: wrap the trigger in an `inline-flex … shrink-0` box so the logo
  button stays icon-sized instead of squishing in the now-flexible header.
- `AppHeader`: drop the now-redundant `*:first:flex-none` override.

Scoped to layout only — no portaling of the menu, so `ScalarMenu` behaviour
(and the api-client sidebar menu) is otherwise unchanged.

## Note

`ScalarHeader` and `scalar-app` page-tour visual snapshots change with this
layout and need refreshing in CI (`--update-snapshots`).

## Ticket

Fixes DOC-5424


hwkr:
Ended up rewriting how ScalarHeader works to better reflect how we've been using it.

hwkr:
Fixed in https://github.com/scalar/scalar/pull/10025

PR #9234, opened 2026-05-15
fix(api-reference): prevent security badge from shrinking

hwkr:
Stops the security badge from shrinking and wrapping when the operation header overflows.

Before:
CleanShot 2026-05-15 at 15 53 26

After:
CleanShot 2026-05-15 at 15 53 07

## Checklist

- [x] I added a changeset. 
- [ ] I added tests.


amritk (approved):
needs changeset

marclave (approved):

PR #9240, opened 2026-05-16
feat(aspnetcore): cryptographic nonce for script tags

fredrik-aschehoug:
It's not uncommon for APIs to apply a CSP policy globally to all responses. This change will allow conforming to such a CSP policy by applying a nonce to the script tags. 

An alternative to this change is to add exceptions for Scalar to the CSP policy, which also works OK. But I would like to avoid exceptions if possible.

### DIFF
--- a/integrations/dotnet/aspnetcore/src/Scalar.AspNetCore/Extensions/ScalarEndpointRouteBuilderExtensions.cs
+++ b/integrations/dotnet/aspnetcore/src/Scalar.AspNetCore/Extensions/ScalarEndpointRouteBuilderExtensions.cs
@@ -152,6 +152,12 @@ public static class ScalarEndpointRouteBuilderExtensions
                 ? string.Empty
                 : $" nonce=\"{HtmlEncoder.Default.Encode(options.Nonce)}\"";
 
+            if (!string.IsNullOrWhiteSpace(options.Nonce))
+            {
+                // Prevent intermediaries and browsers from replaying a one-time nonce to another client.
+                httpContext.Response.Headers.CacheControl = "no-store";
+            }
+
             return Results.Content(
                 $$"""
                   <!doctype html>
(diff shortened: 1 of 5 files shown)


hanspagel:
thanks for the PR! claude and I tried to take it a bit further

hanspagel:
@cursoragent can you review this PR?

integrations/dotnet/aspnetcore/src/Scalar.AspNetCore/Extensions/ScalarEndpointRouteBuilderExtensions.cs
```diff
@@ -147,6 +149,22 @@ public static IEndpointConventionBuilder MapScalarApiReference(this IEndpointRou
 
             var escapedRequestPath = Uri.EscapeDataString(httpContext.Request.Path);
 
+            // Auto-generated values win when both are set — a per-request nonce is the secure path.
+            var effectiveNonce = options.DynamicNonce
+                ? GenerateNonce()
+                : options.Nonce;
```
amritk, line 155: where does options.nonce come from? Is it generated per request?

hanspagel: this is if the user provides a static nonce

hanspagel (approved):

integrations/dotnet/aspnetcore/src/Scalar.AspNetCore/Extensions/ScalarEndpointRouteBuilderExtensions.cs
amritk: Yea my point was a static nonce is no nonce at all, only per request is useful

fredrik-aschehoug: In my case, the nonce is generated per request in a custom middleware  and made available in the HttpContext which can be referenced when specifying the options.

amritk: Good good!

PR #9244, opened 2026-05-18
fix(openapi-types): support OpenAPI 3.2 schema forms

### LINKED ISSUE #9190: MediaTypeObject does not support ReferenceObject
### What happens?

See example using v0.8.0 of `@scalar/openapi-types`.

```typescript
// ✅  No type error (using old types)
import type { OpenAPIV3_2 } from "./node_modules/@scalar/openapi-types/dist/openapi-types.d.ts";
const success: OpenAPIV3_2.MediaTypeObject = {
  schema: { 
    $ref: "test" 
  }
};

// ❌ Type error: Object literal may only specify known properties, and '$ref' does not exist in type 'SchemaObject'
import type { MediaTypeObject } from "@scalar/openapi-types/3.2";
const fail: MediaTypeObject = {
  schema: { 
    $ref: "test" 
  }
};

```

From what I can tell, the issue is that the stricter types (#8942) do not allow `schema` to be a `ReferenceObject`:

https://github.com/scalar/scalar/blob/2d36d4b0fa5576566d1c242648921259b6815d8c/packages/openapi-types/3.2/media-type.d.ts#L12-L18

Compared to the old types that allow `ReferenceObject`:

https://github.com/scalar/scalar/blob/2d36d4b0fa5576566d1c242648921259b6815d8c/packages/openapi-types/src/openapi-types.ts#L229-L240

I expect the `schema` property to allow `ReferenceObject`. 

```typescript
// ✅  No type error
import type { MediaTypeObject } from "@scalar/openapi-types/3.2";
const success: MediaTypeObject = {
  schema: { 
    $ref: "test" 
  }
};
```

I believe the same thing applies to `itemSchema`.

### OpenAPI Document

_No response_

### LINKED ISSUE #9191: SchemaObject type does not support boolean (not spec compliant)
### What happens?

I'm trying to get `@scalar/openapi-types` to work with my `zod` schemas but am running into an issue because the `SchemaObject` type is not compatible between these packages (it's defined as `_JSONSchema` in `zod`).
 
From what I can tell, `zod` is correct in allowing `SchemaObject` to be a boolean according to the OpenAPI specification:

> The empty schema (which allows any instance to validate) MAY be represented by the boolean value true and a schema which allows no instance to validate MAY be represented by the boolean value false.
> https://spec.openapis.org/oas/v3.2.0.html#schema-object

See example using `@scalar/openapi-types` v0.8.0:

```typescript
// ❌ Type 'boolean' is not assignable to type 'SchemaReferenceType<SchemaObject>'.

import type { SchemaObject } from "@scalar/openapi-types/3.2";

const fail: SchemaObject = {
  type: "object",
  properties: {
    anything: true
  }
};
```

I'd expect `boolean` to be allowed as a `SchemaObject`

```typescript
// ✅ No type error

import type { SchemaObject } from "@scalar/openapi-types/3.2";

const fail: SchemaObject = {
  type: "object",
  properties: {
    anything: true
  }
};
```

### OpenAPI Document

_No response_

cyphercodes:
## Problem

`@scalar/openapi-types/3.2` rejects two valid OpenAPI 3.2 schema patterns reported in #9191 and #9190:

- boolean schemas (`true` / `false`) where a Schema Object is accepted
- media type `schema` and `itemSchema` values that are references

Fixes #9191.
Fixes #9190.

## Solution

- Include `boolean` in the OpenAPI 3.2 `SchemaObject` union.
- Allow `ReferenceObject` for OpenAPI 3.2 `MediaTypeObject.schema` and `MediaTypeObject.itemSchema`.
- Add compile-time regression coverage for the version-specific `@scalar/openapi-types/3.2` exports.
- Add a patch changeset for `@scalar/openapi-types`.
- Format `projects/scalar-app/RELEASE_NOTES.schema.json` so the repo-wide format check passes on the current base.

## Verification

- `pnpm format:check`
- `pnpm --filter @scalar/openapi-types types:check`
- `pnpm --filter @scalar/openapi-types test`
- `pnpm --filter @scalar/openapi-types build`
- `pnpm changeset status --since upstream/main`
- `git diff --check`

## Checklist

- [x] I added a changeset.
- [x] I added tests.
- [x] I updated the documentation. (Not applicable: type declaration fix.)

### DIFF
--- a/.changeset/openapi-32-schema-types.md
+++ b/.changeset/openapi-32-schema-types.md
@@ -1,0 +1,5 @@
+---
+'@scalar/openapi-types': patch
+---
+
+Allow OpenAPI 3.2 boolean schemas and media type schema references.
--- a/packages/openapi-types/3.2/media-type.d.ts
+++ b/packages/openapi-types/3.2/media-type.d.ts
@@ -16,9 +16,9 @@ export type MediaTypeObject = {
   examples?: Record<string, ExampleObject | ReferenceObject>
   description?: string
   /** A schema describing the complete content of the request, response, parameter, or header. */
-  schema?: SchemaObject
+  schema?: SchemaObject | ReferenceObject
   /** A schema describing each item within a [sequential media type](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.2.0.md#sequential-media-types). */
-  itemSchema?: SchemaObject
+  itemSchema?: SchemaObject | ReferenceObject
   /** A map between a property name and its encoding information, as defined under [Encoding By Name](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.2.0.md#encoding-by-name).  The `encoding` field SHALL only apply when the media type is `multipart` or `application/x-www-form-urlencoded`. If no Encoding Object is provided for a property, the behavior is determined by the default values documented for the Encoding Object. This field MUST NOT be present if `prefixEncoding` or `itemEncoding` are present. */
   encoding?: Record<string, EncodingObject>
   /** An array of positional encoding information, as defined under [Encoding By Position](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.2.0.md#encoding-by-position).  The `prefixEncoding` field SHALL only apply when the media type is `multipart`. If no Encoding Object is provided for a property, the behavior is determined by the default values documented for the Encoding Object. This field MUST NOT be present if `encoding` is present. */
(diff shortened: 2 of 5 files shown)


hanspagel:
those are auto-generated from the official JSON schemas, so we might want to patch the schemas and re-generate? @amritk

amritk:
> those are auto-generated from the official JSON schemas, so we might want to patch the schemas and re-generate? @amritk 

Actually we're detached now so it's fine we can merge this ☺️

hanspagel:
okay, cool
thanks @cyphercodes
🙏 🙏

hanspagel (approved):

PR #9271, opened 2026-05-20
fix(scalar-app): apply layout-aware CORS proxy defaults for document loading

DemonHa:
- Pass `layout` into `createAppState` from the web and desktop entrypoints so workspace stores can pick the correct default CORS proxy.
- Set `x-scalar-active-proxy` on new workspace stores via `getActiveProxyUrl(undefined, layout)` — matching collection settings — so web uses `https://proxy.scalar.com` and desktop skips the proxy unless the user configures one.
- Apply the same proxy defaults to import draft stores (`ImportListener`, `ImportModal`) so URL imports work in the browser without manual proxy setup.

## Checklist

- [x] I added a changeset. 
- [x] I added tests.

---

> [!NOTE]
> **Medium Risk**
> Changes default network-fetch behavior by seeding `x-scalar-active-proxy` based on client layout, which can affect how documents are fetched in web vs desktop and may surface regressions in proxy/URL loading flows.
> 
> **Overview**
> Applies **layout-aware CORS proxy defaults** by requiring `layout` when creating app state and computing a `defaultProxyUrl` used when a workspace has no explicit `x-scalar-active-proxy`.
> 
> New workspace stores (including draft stores created during workspace creation and import flows) are now initialized with proxy metadata so the **web client defaults to `https://proxy.scalar.com`** while **desktop defaults to no proxy** unless configured. Tests were updated to pass `layout` and to assert proxying behavior for global `fetch` vs `customFetch`, and import components now receive/forward `defaultProxyUrl`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2b8df8d9e31774a4d6212f671540b0f9663bf28d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


hanspagel (commented):
instead of passing the layout and doing an if/else based on the layout, we can consider to do the if/else outside and explicitly pass a setting to control the proxy

amritk (approved):

PR #9272, opened 2026-05-20
fix(aspnetcore): escape request path interpolated into initialize() JS string

hanspagel:
## Problem

The aspnetcore integration interpolates the request path into the inline `initialize('...')` call wrapped in single quotes, but only runs it through `Uri.EscapeDataString` — a URL encoder, not a JS-string encoder. A request like `/scalar/robots.txt'-alert(1)-'` could therefore land in the rendered `<script>` as a quote-breakout payload. The same defect was one line below, on the (developer-controlled) `JavaScriptConfiguration` option.

## Solution

Run both values through `JavaScriptEncoder.Default.Encode` — the same encoder we already use for the nonce on line 166 — and switch the surrounding quotes to double quotes so the style stays consistent. Added a small theory test pinning the two reported PoC paths.

## Checklist

- [x] I added a changeset.
- [x] I added tests.

---

> [!NOTE]
> **Medium Risk**
> Touches HTML generation for the ASP.NET Core endpoint and changes how values are encoded inside an inline script; small but security-sensitive because it mitigates a potential XSS vector and could affect initialization if encoding is wrong.
> 
> **Overview**
> Hardens the ASP.NET Core `MapScalarApiReference` HTML response against inline-script injection by JavaScript-encoding the request path and the `JavaScriptConfiguration` option before interpolating them into the `initialize(...)` call (and switching the string literals to double quotes).
> 
> Adds a regression test covering quote-breakout request paths and includes a patch changeset for `@scalar/aspnetcore`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 153f1447e843b2cf212beb0d5b1bfdd06fe4ca9e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/aspnetcore-initialize-xss-fix.md
+++ b/.changeset/aspnetcore-initialize-xss-fix.md
@@ -1,0 +1,5 @@
+---
+'@scalar/aspnetcore': patch
+---
+
+fix(aspnetcore): JSON-encode the request path and `JavaScriptConfiguration` option before interpolating them into the inline `initialize(...)` script, so a crafted request path cannot break out of the JS string literal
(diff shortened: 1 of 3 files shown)


DemonHa (changes requested):
This is not really an issue @hanspagel.

we are using Uri.EscapeDataString which should strip all the special characters so it will never be able to escape the string 😊

Image

geoffgscott (approved):

hanspagel:
agree, my fault, not an actual issue! the issue was just a lack of sleep it seems

PR #9289, opened 2026-05-20
fix(api-reference): response content type sharing across components

### LINKED ISSUE #9250: CSV content type not selectable
### What happens?

In my responses, CSV is not a selectable content type. I've done a quick test in ReDoc, which shows fine, but in Scalar it just shows "JSON" when I hover over the example. Apologies if I've missed something.

Image
Image

[test.json](https://github.com/user-attachments/files/27967681/test.json)

Screenshots and test JSON attached.

Thanks! 

CSV should be selectable as a response type.

### OpenAPI Document

_No response_

alonfaraj:
## Problem

When an OpenAPI spec defines multiple response content types (e.g. `application/json` and `text/csv`), the example response panel on the right always showed the first content type regardless of what was selected in the left-side response dropdown. There was no connection between the two.

## Solution

Added a `provide/inject` channel (`RESPONSE_CONTENT_TYPE_SYMBOL`) scoped to each operation layout. `ParameterListItem` writes the selected content type into a shared `Record<statusCode, mimeType>` map whenever the response dropdown changes. `ExampleResponses` injects that map and reads from it to display the matching example.

## Checklist

- [x] I added a changeset.
- [ ] I added tests.

## Ticket

Fixes #9250


hanspagel:
we recently discussed that, and weren't sure we want this from a product perspective. and I have to say that we learned to avoid provide/inject. so I leave this open as a draft until we made a decision here, but if decide to do the change it'll be without provide/inject. no need to update the PR yet, we can take it from here.

hanspagel (approved):
thanks, let's get this in (FINALLY) :)

PR #9291, opened 2026-05-20
refactor(components): import from component subpaths instead of the barrel

amritk:
## Summary

First slice extracted from #9137, now also completing the barrel-import cleanup repo-wide and enforcing it.

1. **Migrate to subpath imports** — replace barrel `@scalar/components` imports with component subpaths (e.g. `@scalar/components/scalar-button`) everywhere: `api-reference`, `api-client`, `sidebar`, `agent-chat`, plus the remaining consumers (`scalar-app`, `pre-post-request-scripts`, `openapi-to-markdown`, `examples/web`).
2. **Add the subpath exports** — `@scalar/components` gains a `./helpers` export and individual component subpath exports; the last internal barrel import is removed.
3. **Enforce it** — a Biome `noRestrictedImports` rule (TypeScript) and an ESLint `no-restricted-imports` rule (Vue) make importing from the `@scalar/components` barrel an error going forward.

This unblocks the ESM standalone bundle / tree-shaking work, which follows in a separate PR.

### Notes

- `cva` / `cx` / `useBindCx` / `compose` / `tw` were only re-exported by the barrel; they now import from `@scalar/use-hooks/useBindCx` directly. `pre-post-request-scripts` gains `@scalar/use-hooks` as a dependency (previously transitive).
- Test mocks that targeted the barrel (`vi.mock('@scalar/components', ...)`) were repointed to the relevant subpath.

### Verification

- `pnpm build:packages` and `pnpm build:integrations` — pass
- `types:check` for every affected package/project (api-reference, api-client, components, sidebar, agent-chat, pre-post-request-scripts, openapi-to-markdown, scalar-app, examples/web) — pass
- Tests for the touched packages/projects — pass
- `pnpm lint:check` (Biome, repo-wide) — 0 errors; `lint:vue` — no new errors (the one pre-existing `vue/valid-v-for` error in `SpecificationExtension.vue` is unrelated)
- `pnpm knip` and `pnpm changeset status` — pass

## Ticket

Part of #9137

> Note: the headline feature (ESM standalone bundle, closes #5743) lands in the follow-up PR, not this one.

---

> [!NOTE]
> **Medium Risk**
> Medium risk due to large-scale import-path churn across many packages plus new lint restrictions; mistakes could surface as build/runtime module resolution issues, but behavior changes are minimal.
> 
> **Overview**
[...]

PR #9310, opened 2026-05-21
feat(api-reference): add ESM standalone bundle with code-splitting

### LINKED ISSUE #5743: ESM CDN

amritk:
## Summary

Second slice of #9137, now that the subpath-import work (#9291) has landed. Adds an ESM standalone build (`dist/browser/standalone.esm.js`) alongside the existing UMD bundle, then layers code-splitting and tree-shaking on top of it.

The new bundle works both as a side-effect script (registers `window.Scalar.createApiReference`, reads `data-*` config) **and** as a real ESM module that exports `createApiReference`, so one file covers legacy `<script>` usage and modern ESM consumers.

### Changes
- **ESM standalone bundle** — new `src/standalone.esm.ts` entry + `vite.standalone.esm.config.ts`, exposed via `api-reference/package.json`.
- **`sideEffects` flag** on `@scalar/api-reference` so downstream bundlers can tree-shake.
- **Lazy-load the API client modal** — `ApiReference.vue` loads it via `await import('@scalar/api-client/modal')` in `onMounted`; `Content.vue` imports `mapHiddenClientsConfig` from a deep export; `@scalar/api-client` adds the `./modal/map-hidden-clients-config` export + `sideEffects`.
- **Per-icon lazy chunks** for the icon library SVGs.
- **Eagerly bundle the default sidebar fallback icon** so it is not a lazy round-trip.
- **Per-format webpack-stats** files, keeping the UMD stats at `webpack-stats.json` for relative-ci.

### Build output (verified)
`standalone.esm.js` with the heavy pieces split into lazy chunks — modal (~278 KB), AgentScalar chat (~200 KB), and per-icon SVG chunks — plus `webpack-stats.esm.json` / `webpack-stats.json`.

### Verification
- `pnpm build:packages` (42/42) and the dedicated `build:standalone:esm` — pass
- `types:check` for api-reference, api-client, icons — pass
- `pnpm lint:check`, `pnpm knip`, `pnpm changeset status` — pass
- `api-reference` component tests pass, except `ApiReference.ssr.test.ts`, which fetches external URLs and fails only due to no network egress in the build sandbox (unrelated to this change)

## Ticket

Closes #5743
Part of #9137

---
_Generated by [Claude Code](https://claude.ai/code/session_018iCtWgb2AQG8CsYr7BCkLd)_

---

> [!NOTE]
> **Medium Risk**
> Touches packaging/build outputs and introduces new async chunk loading paths (dynamic imports, code splitting) that could affect runtime loading behavior and CDN/asset resolution if misconfigured.
> 
> **Overview**
> Adds a new **ESM standalone** browser build for `@scalar/api-reference` (`dist/browser/standalone.esm.js`) via `vite.standalone.esm.config.ts`, including Rolldown minification, code-splitting, and separate `webpack-stats.esm.json`; `package.json` now runs this build and excludes both stats files from published `dist`.
> 
> Refactors API reference runtime to reduce initial load by **lazy-loading the API client modal** (`await import('@scalar/api-client/modal')` on mount) and switching `mapHiddenClientsConfig` usage to a new deep export (`@scalar/api-client/modal/map-hidden-clients-config`) added to `@scalar/api-client` exports.
> 
> Updates `@scalar/icons/library` to **eagerly bundle and cache** the default fallback icon (`interface-content-folder`) while keeping other SVG icons lazy-loaded.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 73839a3912e86170bdc230da987dbe9479eaf965. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/esm-standalone-bundle.md
+++ b/.changeset/esm-standalone-bundle.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
+Add an ESM standalone build (`dist/browser/standalone.esm.js`) alongside the existing UMD bundle. The new bundle works as a side-effect script (registers `window.Scalar.createApiReference` and reads `data-*` configuration) and exports `createApiReference` for direct ESM consumers. It is fully minified through Rolldown's native minifier so the total size is on par with — slightly smaller than — the UMD bundle.
--- a/.changeset/esm-standalone-bundle.md
+++ b/.changeset/esm-standalone-bundle.md
@@ -3,3 +3,5 @@
 ---
 
 Add an ESM standalone build (`dist/browser/standalone.esm.js`) alongside the existing UMD bundle. The new bundle works as a side-effect script (registers `window.Scalar.createApiReference` and reads `data-*` configuration) and exports `createApiReference` for direct ESM consumers. It is fully minified through Rolldown's native minifier so the total size is on par with — slightly smaller than — the UMD bundle.
+
+Also declares `sideEffects` in `package.json` so downstream bundlers can tree-shake unused exports from the regular ESM library build. Only CSS files and the standalone CDN bundles in `dist/browser/` are marked as having side effects; everything else is pure.
--- a/.changeset/esm-standalone-bundle.md
+++ b/.changeset/esm-standalone-bundle.md
@@ -11,4 +11,4 @@ Add an ESM standalone build (`dist/browser/standalone.esm.js`) alongside the exi
 
 Net effect: initial sync load drops from ~3.32 MB (UMD) to ~2.73 MB (ESM) — a ~570 KB improvement — while total bundle size shrinks by ~140 KB.
 
-Also declares `sideEffects` in both packages' `package.json` so downstream bundlers can tree-shake unused exports, and adds an `@scalar/api-client/modal/map-hidden-clients-config` deep export so consumers that only need the lightweight client-list helper don't pull the full modal barrel into their static graph.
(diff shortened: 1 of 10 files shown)


hanspagel (approved):
exciting! do we want to add this to the package.json as an export?

marclave (approved):

amritk:
> exciting! do we want to add this to the package.json as an export?

@hanspagel we don't export the standalone it seems? Is this for the CDN style export? Otherwise the html api is already the ESM version of this?

hanspagel:
as a user, I somehow need to include `src/standalone.esm.ts‎` in my HTML, right?
for jsdelivr/umd this line does the trick:

https://github.com/scalar/scalar/blob/main/packages/api-reference/package.json#L101

but for now we can just use the direct path to the dist/ file, and I can ff with a PR to make it nicer for our users

PR #9324, opened 2026-05-22
fix(workspace-store): missing migration

hanspagel:
The v2 IndexedDB migration was rewritten in place between `@scalar/workspace-store@0.48.x` and `0.50.0` without bumping the schema version (the version is derived from `migrations.length`, which stayed at 2). Two different schemas shipped as "version 2":

- **Superseded v2** (0.48.x–0.49.x): `workspace` store keyed by `[teamSlug, slug]`, no indexes.
- **Current v2** (0.50.0+): `workspace` store keyed by `workspaceUid` with `teamSlug_slug` + `teamUid` indexes.

Installs created by the superseded v2 are stranded — `onupgradeneeded` never fires again because the requested version already matches. The runtime then asks for the `teamSlug_slug` index that does not exist and `getItemBySlug` crashes on startup with `NotFoundError: The specified index was not found`.

This adds a **v3 migration** that recreates the workspace and chunk stores in the current UID-based shape. It is a no-op for installs already on that shape, detected via the presence of the `teamSlug_slug` index — so fresh installs and v1 upgrades are untouched.

Tests cover the repair path, chunk re-keying, the restored `teamUid` index, and the no-op path for already-migrated databases.

---

> [!NOTE]
> **Medium Risk**
> Adds an IndexedDB repair migration that drops/recreates object stores and re-keys persisted data; mistakes could lead to data loss or broken workspace lookups on upgrade, though it’s guarded to no-op on already-correct schemas.
> 
> **Overview**
> Prevents a startup crash (`NotFoundError` on missing `teamSlug_slug` index) for installs created on the superseded “v2” schema by appending a new **v3 IndexedDB migration**.
> 
> The v3 migration detects the stranded schema (missing `teamSlug_slug`), then rebuilds the `workspace` store to the UID-based layout (new `workspaceUid` PK + `teamSlug_slug`/`teamUid` indexes) and re-keys all chunk tables (`meta`, `documents`, etc.) from legacy `workspaceId` to the new `workspaceUid`, dropping orphaned chunks. It explicitly no-ops when the UID-based indexes already exist, and adds comprehensive `fake-indexeddb` tests covering repair, chunk re-keying, index restoration, and no-op behavior.
> 
> Includes a patch changeset documenting the fix.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f14d3aae21d429f39df9ea6ab9054354b1e8d34c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/fix-indexdb-v3-repair-uid-schema.md
+++ b/.changeset/fix-indexdb-v3-repair-uid-schema.md
@@ -1,0 +1,5 @@
+---
+'@scalar/workspace-store': patch
+---
+
+Fixed a `NotFoundError: The specified index was not found` crash on startup for installs created by `@scalar/workspace-store@0.48.x`–`0.49.x`. The v2 IndexedDB migration was rewritten in place without bumping the schema version, so those installs were stranded on the superseded schema (no `teamSlug_slug` index) and never re-ran migrations. A new v3 migration brings them up to the current UID-based shape and is a no-op for installs already on it.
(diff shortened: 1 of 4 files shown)


DemonHa:
Closing this as we never shipped that old v2 schemas in prod. That was the reason we didn't bump the version

PR #9328, opened 2026-05-22
fix(fastify): inline the API Reference JavaScript so bundled builds work

### LINKED ISSUE #3566: `@scalar/fastify-api-reference` doesn't seem to work when bundling

hanspagel:
not good yet, don't look at it

The Fastify integration read the bundled API Reference JavaScript from disk at runtime (`fs.readFileSync` of a path next to the module). That file disappears once an app is bundled into a single file, so production builds (Docker, esbuild, bun, …) crashed with `JavaScript file not found`.

It is now inlined as a string at build time (a small generated module), so it travels with the package and survives downstream bundling. The build stays on `tsc` — no bundler added.

Verified with a single-file bundle run in an isolated folder (no `node_modules`, no `dist`): the JavaScript is served correctly.

## Ticket

Fixes #3566


hanspagel:
mmhhh, still too quirky

PR #9342, opened 2026-05-24
fix(workspace-store): resolve operations when OpenAPI path items use `$ref`

DemonHa:
Currently we didn't display any operations if the `path item` object was a reference object.

Now, path entries and webhooks can reference `components.pathItems` instead of inlining operations. Navigation, mutators, search, and markdown export now resolve path-item references before reading HTTP methods and path-level parameters.

## Checklist

- [x] I added a changeset. 
- [x] I added tests.

---

> [!NOTE]
> **Medium Risk**
> Broad refactor of how operations are read and edited across the monorepo, but behavior is constrained to path-item `$ref` resolution with heavy test coverage; regressions would mainly affect OpenAPI navigation, editing, and search rather than runtime auth or network calls.
> 
> **Overview**
> Fixes a gap where **path and webhook entries that only held a `$ref` to `components.pathItems`** showed no operations and broke anything that read HTTP methods via `paths[path][method]`.
> 
> **Schema and typing:** `paths`, `webhooks`, and `components.pathItems` now allow **path-item reference objects** (with `$ref` / `$ref-value`), matching OpenAPI 3.1; generated types and coercion tests were updated.
> 
> **Central helpers** in `@scalar/workspace-store` (`getResolvedPathItem`, `getPathItemOperation`, `forEachPathItemOperation`, plus get/set/delete and `pathItemIsEmpty`) **merge sibling fields** (e.g. path-level `parameters`) with the referenced path item, then expose or mutate operations on `$ref` wrappers correctly—including removing sibling overrides on delete.
> 
> **Call sites** across workspace-store (navigation, mutators, request examples, server externalization), api-reference/api-client search and routing, openapi-to-markdown, oas-utils migration, and scalar-app now use these helpers instead of assuming inline path items. Path cleanup after deleting the last operation **keeps path-level metadata** instead of dropping the whole path entry when only operations are gone.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ad5168ae8cca35f9f495c1f2d5061138abcc75c1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


DemonHa:
Add e2e tests make sure typebox actually correctly coerces the $refs on the path item object

marclave (approved):

hanspagel:
Rebased and addressed the review feedback.

amritk (approved):
didn't go through the whole pr but what I saw looked decent

packages/workspace-store/src/schemas/v3.1/strict/components.ts
```diff
@@ -79,5 +79,5 @@ export type ComponentsObject = {
   /** An object to hold reusable Callback Objects. */
   callbacks?: Record<string, ReferenceType<CallbackObject>>
   /** An object to hold reusable Path Item Objects. */
-  pathItems?: Record<string, PathItemObject>
+  pathItems?: Record<string, ReferenceType<PathItemObject>>
```
DemonHa, line 82: we would also need to update the new schemas on the @scalar/schemas package, which are currently being used for the validation

PR #9344, opened 2026-05-25
fix(api-reference): show request example below operation details on narrow viewports

cameronrohani:
## Summary

On narrow viewports (≤900px), the request block (HTTP method + URL / curl) should appear below the operation title and description, not above path parameters and related schema content.

Implemented by setting flex-direction: column-reverse on .section-columns in SectionColumns.vue.

image

### DIFF
--- a/.changeset/section-columns-column-reverse.md
+++ b/.changeset/section-columns-column-reverse.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
+fix: reverse section column stack order on narrow viewports (≤900px)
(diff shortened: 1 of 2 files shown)


hwkr (approved):
Seems good! Not sure if that pic is before or after...

hanspagel (changes requested):
I don't think that's better

* I personally like everything better in the "before" screenshot
* the description is too far away from the operation title

before vs. after

Image

hwkr:
Closing in favor of https://github.com/scalar/scalar/pull/9370

Issue #7936, opened 2026-01-28
[Bug]: objects in `multipart/form-data` must be converted to bracket notation


gelsonmatavela:
### What happens?

<img width="951" height="280" alt="Image" src="https://github.com/user-attachments/assets/19ef4394-8dfe-4f88-a29a-a316b3ee12e4" />

- The object in the value field (address) is showing in as json format 

### What did you expect to happen?

<img width="962" height="397" alt="Image" src="https://github.com/user-attachments/assets/0a3ba331-c133-420d-9ad0-7f2aa131e1e9" />

- Those should be showed like this instead

Uanela:
I did this on my framework which uses scalar, but still waiting for this feat will much better to come from scalar.

Btw the repo is [The Express And Prisma RESTful Framework](https://github.com/uanela/arkos)

giovannipiller:
Just wanted to note that this might be considered a breaking change, because not all servers automatically convert bracket notation to an object (e.g. Fastify).

I'd love for this to be implemented, but as an opt-in feature.

Uanela:
> Just wanted to note that this might be considered a breaking change, because not all servers automatically convert bracket notation to an object (e.g. Fastify).
> 
> I'd love for this to be implemented, but as an opt-in feature.

I did know about this, even express does not convert but I think is real helpful thing (even as an opt-in feat as you said).

But for query parameter I guess all uses bracket notation.

Could you consider checking the repo I just shared? @giovannipiller

gelsonmatavela:
That will really help us.
As @Uanela said on ArkosJs, another open-source project I’m contributing to… 
And that problem was actually solved there.
Even Express doesn’t support/convert it.

If you guys need any help, I’m open to it!

amritk:
> I'd love for this to be implemented, but as an opt-in feature.

Yea we'll probably follow the style examples on this one, so if you have a schema.explode = true we can break it out

https://spec.openapis.org/oas/latest.html#style-examples

I'll take a look next week

PR #9351, opened 2026-05-27
fix(workspace-store): preserve large integer parameter values

### LINKED ISSUE #9343: API Client 'BigInt' support not working as intended
### What happens?

I attempt to pass in values that are greater than 56 bits, in my case a Twitter-style snowflake, and find that the values are losing precision and undergoing rounding, causing the id lookups within my API to fail.

I attached a minimal reproducible example in scalar/scalar#2332. Additionally I have provided a OpenAPI sandbox link showing the issue. You can change the URL to whatever is needed and you can see in the raw preview on Chrome and Firefox that the values change. If you prefer, you can also add `app.MapGet("/{id:long}", (long id) => id.ToString());` to the playground project to check the echo result showing the issue.

When sending in a request via the client to a localhost where one of the parameters, defined by an OpenAPI schema, is a 64bit value I expect it to make it to the destination endpoint without any precision loss.

Following scalar/scalar#2332 I would expect that passing in values numerical values greater than 56bits would work as intended, but it appears if there is any kind of type constraint some conversion or checking will be done that prevents it working.

I spent a great many hours looking through the code but I couldn't easily found where the JSON parsing happens. I do not think it is related to the download code that works with the schema, but this isn't my project so could be off the mark by some margin.

### OpenAPI Document

[https://sandbox.scalar.com/e/T5Jax](<https://sandbox.scalar.com/e/T5Jax>)

### Extra

If you need anything else from me please let me know. I have taken a look around the codebase quite a bit but cannot rightly say I understand the dataflow all that well. I originally said with confidence in that original issue that it worked on the web client. That was a misunderstanding on my part. The ID I used for testing was being interpreted as a string, and thus didn't reveal the bug and [ASP.NET](<http://ASP.NET>) was converting the string implicitly, so I never caught the error in my echo test.

Here are some numbers for easy testing. What you send vs what you get atm:

* 1507323546366377984 -> 1507323546366378000
* 862817670527975424 -> 862817670527975400
* 2057268305600708608 -> 2057268305600708600

DemonHa:
Fixes: #9343 

Stop JSON-parsing primitive parameter values when building requests. Integer and number path, query, and header fields keep the exact string from the editor so values larger than Number.MAX_SAFE_INTEGER are not rounded. Array and object parameters are still parsed for OpenAPI style serialization.

## Checklist

- [x] I added a changeset. 
- [x] I added tests.

---

> [!NOTE]
> **Low Risk**
> Targeted request-building coercion change with tests; serializers already handle string primitives for HTTP output.
> 
> **Overview**
> **`deSerializeParameter`** in `@scalar/workspace-store` no longer runs **`JSON.parse`** on primitive schema types (`integer`, `number`, `boolean`, `null`). Path, query, and header values from the request editor stay as the exact strings the user typed, so integers beyond **`Number.MAX_SAFE_INTEGER`** are not rounded when requests are built.
> 
> **Array** and **object** parameters are still parsed (including comma-split fallback for arrays) so OpenAPI style serialization keeps working. JSON **content**-based parameters are unchanged.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 35c46de08d4ff3a4e58fc065cc3caf3a21fd2a0c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/workspace-store/src/request-example/builder/header/de-serialize-parameter.test.ts
+++ b/packages/workspace-store/src/request-example/builder/header/de-serialize-parameter.test.ts
@@ -88,7 +88,7 @@ describe('de-serialize-parameter', () => {
       expect(result).toEqual({ data: 'value' })
     })
 
-    it('parses boolean string when schema type is boolean', () => {
+    it('keeps boolean strings unchanged when schema type is boolean', () => {
       const param: ParameterWithSchemaObject = {
         name: 'test',
         in: 'query',
@@ -100,10 +100,10 @@ describe('de-serialize-parameter', () => {
 
       const result = deSerializeParameter(example, param)
 
-      expect(result).toBe(true)
+      expect(result).toBe('true')
     })
 
-    it('parses number string when schema type is number', () => {
+    it('keeps number strings unchanged when schema type is number', () => {
       const param: ParameterWithSchemaObject = {
         name: 'test',
         in: 'query',
@@ -115,10 +115,10 @@ describe('de-serialize-parameter', () => {
 
       const result = deSerializeParameter(example, param)
 
-      expect(result).toBe(42.5)
+      expect(result).toBe('42.5')
     })
 
-    it('parses integer string when schema type is integer', () => {
+    it('keeps integer strings unchanged when schema type is integer', () => {
       const param: ParameterWithSchemaObject = {
         name: 'test',
         in: 'query',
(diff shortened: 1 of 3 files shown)


DemonHa:
Right now we are doing de-serialization for the parameters based on the schema but I think this behaviour is about to change when we correctly handle the arrays and object into their own entry

see #7936

cc @amritk

packages/workspace-store/src/request-example/builder/header/de-serialize-parameter.test.ts
```diff
@@ -100,10 +100,10 @@ describe('de-serialize-parameter', () => {
 
       const result = deSerializeParameter(example, param)
 
-      expect(result).toBe(true)
+      expect(result).toBe('true')
```
hanspagel, line 103: I’m a bit cautious when we have to change unrelated test cases

Does this mean we'll send "null" instead of nothing for `type: ['null', 'string']`?

DemonHa: The input component always returns a string value, so keeping it as a string is safe. When the request is sent, it will still be serialized correctly as a string.

We previously tried parsing the value as JSON because, in cases where the value is an object, we want to apply the OpenAPI `style` and `explode` serialization rules correctly for objects.

marclave (approved):

hanspagel (changes requested):
would love to get redis thoughts on my question before merging

hanspagel (approved):
ok

PR #9370, opened 2026-05-30
fix(api-reference): show request example below description on narrow screens

hwkr:
Another go at #9344 — that one set `flex-direction: column-reverse` on the shared `SectionColumns`, which pushed the description too far from the title and leaked into every layout that uses it.

Instead the modern operation layout is now its own CSS grid, so `grid-template-areas` can reorder the pieces on narrow screens without touching the shared components. The request example now sits right under the description instead of all the way at the bottom:

* `badge → title → description → request example → params/responses` once the container is ≤900px
* Drops the `SectionHeader` half-width grid — the title and auth badge are just their own grid areas now
* Auth badge is an eyebrow above the title on narrow screens, and top-right (above the code) on wide cc @cameronrohani 
* Shared `SectionColumns` / `SectionColumn` / `SectionHeader` are left alone, so `TagSection` and the intro layout are unaffected
* Sticky code panel still sticks

Supersedes #9344.

## Checklist
- [x] I added a changeset.
- [ ] I added tests.


hanspagel:
before vs. after

Screenshot 2026-06-01 at 10 37 12

hanspagel (approved):
Awesome, I’m okay with this one

cameronrohani (approved):

PR #9371, opened 2026-05-31
chore(themes): change default font stack to match tailwind's

vixalien:
- **chore(themes): change default font stack to match tailwind's**
- **docs(changeset): Use the correct system fonts when withDefaultFonts is set to false**

## Problem

When withDefaultFonts is set to false, scalar would display a random font instead of my configured system font

## Solution

Changed the default font stack to match tailwind's" https://tailwindcss.com/docs/font-family

## Checklist

- [x] I added a changeset. 
- [ ] I added tests.

### DIFF
--- a/.changeset/four-moons-judge.md
+++ b/.changeset/four-moons-judge.md
@@ -1,0 +1,5 @@
+---
+'@scalar/themes': patch
+---
+
+Use the correct system fonts when withDefaultFonts is set to false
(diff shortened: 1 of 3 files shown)


hanspagel:
leaving this for @hwkr and @cameronrohani

hwkr (approved):
Looks good! Thanks for the fix :)

PR #9380, opened 2026-06-01
feat(api-reference): render AsyncAPI operations & messages

hanspagel:
Renders AsyncAPI messages in the API reference.

Each operation now gets its own section under its channel, and its messages are listed as model-style accordions — title (falling back to name, then key), summary, and the payload rendered with the existing `Schema` component.


hanspagel:
nah, we'll take a different approach. we'll make the sidebar nesting different (no channels on the top-level)

PR #9382, opened 2026-06-01
fix(docs): pricing page design

cameronrohani:
removed colors from pricing page, added more padding between each element

image

Fixed button colors


marclave (approved):
🫡

hwkr (approved):
Code looks fine I think, we can move it to MDX soon 🫡

PR #9388, opened 2026-06-02
fix(api-reference): add copy button to virtualized code blocks

AshishKumarPathak:
## Problem

In the API Reference right-side example panel, when JSON responses or schema content exceed ~500 lines (20,000 characters), the rendering switches from `ScalarCodeBlock` to `ScalarVirtualText` for performance reasons (virtualized scrolling). However, the virtualized fallback loses the copy button entirely, making it impossible for users to copy large JSON responses or schemas to clipboard. The content also appears as unstyled plain text, looking inconsistent with smaller responses that render inside a proper code block container.

This affects both:
- **Example Response tab** — Large JSON response examples
- **Schema tab** — Large schema definitions shown via "Show Schema"

## Solution

Wrapped the `ScalarVirtualText` component in a container `<div>` that replicates the `ScalarCodeBlock` wrapper structure, and added `ScalarCodeBlockCopy` alongside it:

- Added `scalar-code-block` and `group/code-block` classes to the wrapper (required for the copy button's hover visibility CSS to work)
- Positioned `ScalarCodeBlockCopy` absolutely in the top-right corner, matching the exact UX of non-virtualized code blocks
- Applied proper `bg-b-2` background and layout classes for visual consistency

**Alternative considered:** Adding copy functionality directly inside `ScalarVirtualText` component — rejected because it would change a shared generic component's API for a specific use case.

**Trade-off:** Syntax highlighting is still not available for virtualized content (highlight.js cannot efficiently handle 500+ lines), but users now have a functional copy button and consistent styling.

**Files changed:**
- `packages/api-reference/src/features/example-responses/ExampleResponse.vue`
- `packages/api-reference/src/features/example-responses/ExampleSchema.vue`

## How to Test

1. Open any API Reference with a large schema or response (500+ lines of JSON)
2. Navigate to the right-side example panel
3. Verify the copy button appears on hover in the top-right corner
4. Click the copy button and confirm the full JSON content is copied to clipboard
5. Switch to the "Schema" tab and repeat steps 3-4

## Checklist

- [x] I added a changeset.
- [x] I added tests.


packages/api-reference/src/features/example-responses/ExampleResponse.vue
hanspagel, line 80: would probably make sense to create a new "ScalarVirtualCodeBlock" component with snapshot tests

I can take over here

hanspagel (changes requested):

packages/api-reference/src/features/example-responses/ExampleResponse.vue
AshishKumarPathak: Done! I've extracted a ScalarVirtualCodeBlock component in packages/components/src/components/ScalarVirtualCodeBlock/ that encapsulates the ScalarVirtualText + ScalarCodeBlockCopy pattern.

Changes:

- Created ScalarVirtualCodeBlock.vue with props: content, lang, copy, lineHeight
- Added Storybook stories (Base, CopyAlways, NoCopy) and an e2e snapshot test file
- Exported via @scalar/components/virtual-code-block
- Refactored both ExampleResponse.vue and ExampleSchema.vue to use it

Snapshot images haven't been generated yet — my local Docker isn't available. The .e2e.ts and .stories.ts files are ready, so snapshots will generate on first CI run or when someone runs pnpm test:e2e:update.

Let me know if you'd like any adjustments to the component API!

hanspagel:
yeah, great, thanks. it just doesn't have a snapshot in the PR. :)

I add it

hanspagel:
done, leaving this for @hwkr

hwkr (approved):
Code looks good! Thanks for the contrib :)

hwkr:
Ooops looks like it just needs a changeset (mostly because we're adding a component to the components package).

AshishKumarPathak:
> Ooops looks like it just needs a changeset (mostly because we're adding a component to the components package).

Done! Added a changeset — @scalar/components as minor (new component export) and @scalar/api-reference as patch.

hanspagel (approved):
nice

PR #9395, opened 2026-06-02
refactor(api-reference): tiny routing helpers

hanspagel:
Adds a few routing helpers so we don't repeat the routing logic in ApiReference.vue again and again.

not sure we want this, but if we want it:
merge #9396 first, and then rebase this one

### DIFF
--- a/packages/api-reference/src/components/ApiReference.vue
+++ b/packages/api-reference/src/components/ApiReference.vue
@@ -78,10 +78,10 @@ import SearchButton from '@/features/Search/components/SearchButton.vue'
 import { getSystemModePreference } from '@/helpers/color-mode'
 import { downloadDocument } from '@/helpers/download'
 import {
+  createRouting,
+  getDocumentSlugFromUrl,
   getIdFromUrl,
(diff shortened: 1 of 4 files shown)


packages/api-reference/src/components/ApiReference.vue
```diff
@@ -205,12 +205,10 @@ if (typeof window !== 'undefined') {
...
-  const initialId = getIdFromUrl(
-    url,
-    basePaths.find((p) => (p ? matchesBasePath(url, p) : false)),
-    isMultiDocument.value ? undefined : activeSlug.value,
-  )
-  const documentSlug = initialId.split('/')[0]
+  const documentSlug = getDocumentSlugFromUrl(url, basePaths, {
+    isMultiDocument: isMultiDocument.value,
+    activeSlug: activeSlug.value,
+  })
```
hanspagel, line 211: isn't that nicer than having the `p ? matchesBasePath(url, p) : false)` and `split('/')[0]` here?

hanspagel:
nope, nobody wants that

PR #9396, opened 2026-06-02
refactor(api-reference): generic redirects

hanspagel:
Untangles the URL redirect logic in `id-routing.ts` and makes it list-driven, then adds a second redirect on top of it.

## How it works

A navigation id like `default/model/User` can sit in three different URL shapes depending on routing mode (bare `#hash`, `#base/hash`, or `/base/` path). The old code baked that routing chrome into every regex, so it needed a separate pattern per mode. The refactor splits *where the id lives* from *what it should become*:

- **`locateIdCarrier(url, basePath)`** — reduces any URL to its bare id plus a `write` callback that puts a new id back in the same shape. After this, nothing downstream cares about routing.
- **`IdRedirect` + `applyIdRedirects`** — a redirect is just `{ match, replace }` (the args to `String.replace`). Rules run in order against the bare id; first match wins. Because they only see the id, every rule works in all three routing modes for free.
- **`buildRedirects(context)`** — the list itself. The target slug is dynamic (`models`/`schemas`/custom), so the list is built per call from config values. Adding a redirect is a one-line list entry.

## Redirects in the list

1. `model/<name>` → current slug — old singular prefix that always 404'd against the plural section.
2. `models/<name>` → current slug — the section used to live at the literal `models/`; once a custom label moves it (e.g. `schemas/`), old bookmarks should follow. Registered **only when the slug differs from the default `models`**, so default setups are untouched.

The entry point `redirectUrl` (renamed from `redirectLegacyModelUrl`) keeps the same signature. 170 `id-routing` tests pass (8 new for the second redirect).

---

> [!NOTE]
> **Low Risk**
> Navigation URL canonicalization on load only; behavior is heavily tested and scoped to legacy model/models bookmarks without changing auth or data paths.
> 
> **Overview**
> Replaces **`redirectLegacyModelUrl`** with **`redirectUrl`**, backed by a routing-agnostic redirect engine in `id-routing.ts`. Navigation ids are extracted via **`locateIdCarriers`** (hash, hash-base-path, pathname, plus legacy bare-hash fallbacks), rewritten with ordered **`{ match, replace }`** rules in id-space, then written back through each carrier so pathname and stale hash can both be fixed in one load.
> 
> **`buildRedirects`** centralizes rules: legacy singular **`model/`** → current models section slug, and when the section label is customized, **`models/`** → that slug (skipped when the slug is still `models`). **`ApiReference.vue`** calls **`redirectUrl`** on init with the same arguments as before.
> 
> Shared path encoding moves into **`encodeBasePath`** (used by **`getIdFromPath`** and **`matchesBasePath`**). Tests rename the suite and add coverage for hash-base-path, path+hash dual rewrites, and customized section slugs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 70625f0b2046effb405cd83e103c3a87fb43cbe0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/api-reference/src/helpers/id-routing.ts
+++ b/packages/api-reference/src/helpers/id-routing.ts
@@ -170,22 +170,30 @@ type IdCarrier = {
   write: (next: string) => void
 }
 
+/** Carrier for an id that lives in the bare hash fragment (`#<id>`). */
+const bareHashCarrier = (url: URL): IdCarrier => ({
+  value: decodeURIComponent(url.hash.slice(1)),
+  write: (next) => {
+    url.hash = next
+  },
+})
+
 /**
  * Locates the navigation id inside a URL regardless of the routing mode.
  *
  * The id can live in three places — the bare hash, a hash base path (`#<base>/<id>`), or the
- * pathname after a path base path. Returning the decoded id alongside a `write` callback lets
+ * pathname after a path base path. Each carrier pairs the decoded id with a `write` callback, so
  * callers edit ids in id-space without re-deriving the routing chrome for each mode.
+ *
+ * Multiple carriers are returned in priority order: the canonical location for the active routing
+ * mode comes first, followed by any legacy fallback. Path routing falls back to the bare hash so a
+ * stale `#default/model/User` bookmark (left over from before path routing was enabled) is still
+ * canonicalized — the previous implementation rewrote both the path and the hash on every load.
  */
-const locateIdCarrier = (url: URL, basePath: string | undefined): IdCarrier => {
+const locateIdCarriers = (url: URL, basePath: string | undefined): IdCarrier[] => {
   // Hash routing: the id is the bare fragment.
   if (typeof basePath !== 'string') {
-    return {
-      value: decodeURIComponent(url.hash.slice(1)),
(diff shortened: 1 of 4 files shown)


packages/api-reference/src/helpers/id-routing.ts
```diff
@@ -159,19 +161,155 @@ export const makeUrlFromId = (_id: string, basePath: string | undefined, isMulti
...
+    // though the section itself was plural, so old bookmarks like `default/model/User` would 404.
+    renameSectionSegment('model'),
+  ]
+
+  // The models section used to live at the literal `models/`. Once a custom label moves it
+  // elsewhere (e.g. `schemas/`), older `models/<name>` bookmarks must follow it. When the slug is
+  // still the default `models` this would be a no-op, so only register it when it actually differs.
+  if (modelsSectionSlug !== 'models') {
+    redirects.push(renameSectionSegment('models'))
+  }
```
hanspagel, line 319: list of redirects

amritk (approved):

PR #9407, opened 2026-06-03
feat(api-reference): support `url` in x-scalar-sdk-installation

hanspagel:
Extends `x-scalar-sdk-installation` with an optional `url` per SDK entry that links out to the package or repository (GitHub, npm, PyPI, …). The Client Libraries card renders it as a "View on …" link below the install command, with a friendly label for well-known hosts and a fallback to the raw URL. Entries that only have a `url` now render too.

Built on top of #9399.

## Visual

| Full entry | URL-only entry |
|---|---|
| "View on npm" link under the install command | "View on GitHub" link, no install command |


hanspagel:
Folded these `url` changes directly into #9399 (cherry-picked onto `claude/sdk-installation-instructions`), so this stacked PR is no longer needed. Closing.

PR #9413, opened 2026-06-03
feat(dotnet): add AsyncAPI document support with DocumentType discriminator

bielu:
Introduces AsyncAPI document methods (AddAsyncApiDocument, AddAsyncApiDocuments, WithAsyncApiRoutePattern) and a DocumentType enum so OpenAPI and AsyncAPI documents share the same Documents collection but resolve their default route patterns lazily at configuration time. AddAsyncApiDocument now also accepts an agentKey parameter.

## Problem

The .NET integration only supported OpenAPI documents. There was no way to register AsyncAPI documents in the Scalar API Reference from a .NET application. Additionally, OpenAPI and AsyncAPI documents needed to coexist in the same document list (union type) so the Scalar frontend can render them together in a single reference.

## Solution

- Added `AsyncApiRoutePattern` property to `ScalarOptions` (default: `/asyncapi/{documentName}.json`).
- Added `AddAsyncApiDocument`, `AddAsyncApiDocuments`, and `WithAsyncApiRoutePattern` extension methods mirroring the existing OpenAPI document API.
- Introduced a `DocumentType` enum (`OpenApi` / `AsyncApi`) on `ScalarDocument` so the mapper knows which default route pattern to resolve when no explicit pattern is provided. This makes both document types behave consistently: the pattern is resolved lazily at configuration time rather than eagerly at add time.
- Added `agentKey` parameter to `AddAsyncApiDocument` to match the `AddDocument` signature.
- Both document types are stored in the same `Documents` collection — the Scalar frontend already treats sources as a union type.

Alternative considered: eagerly resolving and storing the AsyncAPI route pattern at call time (no `DocumentType` needed). Rejected because it creates an inconsistency — changing `AsyncApiRoutePattern` after adding documents would silently have no effect, unlike the OpenAPI equivalent.

## Checklist

- [ ] I added a changeset. 
- [x] I added tests.

### DIFF
--- a/.changeset/async-api-documents-dotnet.md
+++ b/.changeset/async-api-documents-dotnet.md
@@ -1,0 +1,8 @@
+---
+'@scalar/aspnetcore': minor
+'@scalar/aspire': minor
+---
+
+feat(dotnet): add AsyncAPI document support
+
+Adds `AddAsyncApiDocument`, `AddAsyncApiDocuments`, and `WithAsyncApiRoutePattern` so AsyncAPI documents can be registered alongside OpenAPI documents in the same Scalar API Reference. AsyncAPI documents use a separate default route pattern (`/asyncapi/{documentName}.json`) that is resolved lazily at configuration time.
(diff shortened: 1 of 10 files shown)


hanspagel (approved):
niiiice, thank you so much :)

PR #9415, opened 2026-06-04
feat(themes): add design-system skill to @scalar/themes

hwkr:
Ships a `scalar-design-system` agent skill so AI tools — especially Paper — actually know our design system instead of guessing. It covers our tokens, theming, and `@scalar/components`, and bridges both directions of Paper work (code-to-design and design-to-code).

## What's in it

- A `scalar-design-system` skill (`SKILL.md` + `references/` for tokens, theming, components, and a Paper bridge) living in `packages/themes/skills/`.
- Shipped in the `@scalar/themes` npm package (added to `files`) so it reaches consumers.
- Symlinked into `.agents/skills/` for internal discovery — single source of truth, no drift.

## Notes

- To use it in another repo, link or copy `node_modules/@scalar/themes/skills/scalar-design-system` into that project's `.claude/skills` (agents don't auto-discover skills in `node_modules`).
- Source-of-truth pointers use package names (`@scalar/themes`, `@scalar/components`, `@scalar/icons`) so they resolve for installed consumers, not just the monorepo.
- Complements the existing Paper `code-to-design` / `design-to-code` plugin skills.
- No auto-sync — the values are a hand-maintained snapshot plus references.


packages/themes/scripts/cli.mjs
```diff
@@ -0,0 +1,96 @@
+#!/usr/bin/env node
```
hanspagel: we have this: https://github.com/scalar/scalar/tree/main/tooling/scripts/src/commands

should we put it there?

hanspagel: of, if we want to keep it, let's make it a .ts file and add some tests, even if simple

packages/themes/package.json
```diff
@@ -27,6 +27,9 @@
   "main": "./dist/index.js",
   "module": "./dist/index.js",
   "types": "./dist/index.d.ts",
+  "bin": {
+    "scalar-themes": "./scripts/cli.mjs"
+  },
```
hanspagel: I don't think we want to publish this CLI, do we?

hanspagel (dismissed):

packages/themes/package.json
hwkr: This would be required for org to install it. I think I might just symlink them manually though instead of having this script. Seems like a lot

packages/themes/scripts/cli.mjs
hwkr: Needs to be run by a consumer I think (e.g. org) probs why mjs makes sense for portability. I agree ts might be nice though 🤔

amritk (approved):
smart! didn't test but looks cool

cameronrohani (approved):

PR #9417, opened 2026-06-04
fix(api-client): preserve nested multipart field types

### LINKED ISSUE #9416: Test request - Multipart requests sends strings instead of json after interacting with the Request body form
### What happens?

I'm upgrading `@scalar/fastify-api-reference` from `1.49` to `1.58` and encountered an issue with multipart form handling.

I have a multipart request with the following schema:

```jsonc
{
  // omitted for brevity
  "multipart/form-data": {
    "schema": {
      "type": "object",
      "properties": {
        "jsonField": {
          "default": {
            "isSomething": false,
            "mood": []
          },
          "description": "Options",
          "type": "object",
          "properties": {
            "isSomething": {
              "type": "boolean"
            },
            "mood": {
              "type": "array",
              "items": {
                "type": "string"
              }
            }
          },
          "required": [
            "isSomething",
            "mood"
          ],
          "additionalProperties": false
        },
        "stringField": {
          "type": "string",
          "maxLength": 102400000,
          "description": "string field"
        }
      },
      "required": [
        "jsonField",
        "stringField"
      ]
    }
  }
}      
```

If I submit the request without interacting with the form, the generated payload is correct:

```
------boundary
Content-Disposition: form-data; name="jsonField"

{"isSomething":false,"mood":[]}
------boundary
Content-Disposition: form-data; name="stringField"


------boundary
```

However, after modifying any field in the form, the generated payload changes to:

```
------boundary
Content-Disposition: form-data; name="jsonField"

{"isSomething":"false","mood":"[]"}
------boundary
Content-Disposition: form-data; name="stringField"

ciao2
------boundary
```

As shown above, the values inside `jsonField` are converted from their original types (`boolean` and `array`) to strings.

---

> [!NOTE]
> The UI correctly infers the property types: 
> Image

Json field properties are not converted to a string after interacting with the request body form.

### OpenAPI Document

https://sandbox.scalar.com/e/hHmH-#POST/testing-multi-part

puneetdixit200:
## Problem

Fixes #9416.

Nested multipart object fields are displayed as editable dotted rows in the request body form. After editing, those row values are strings, and the request builder folded them directly back into the grouped JSON multipart part. That made boolean and array leaves send as `"false"` and `"[]"` instead of `false` and `[]`.

## Solution

Resolve the multipart schema leaf for each dotted nested row before regrouping it, then coerce string row values only when the schema type does not allow `string`. The existing schema-driven dotted-row predicate still keeps literal dotted field names flat.

Added a regression test covering a nested multipart object with a boolean leaf and array leaf after form edits.

## Checklist

- [x] I added a changeset.
- [x] I added tests.
- [x] I updated the documentation. Not applicable: this is a request-building bug fix with regression coverage and no docs surface change.

### DIFF
--- a/.changeset/nested-multipart-types.md
+++ b/.changeset/nested-multipart-types.md
@@ -1,0 +1,5 @@
+---
+'@scalar/workspace-store': patch
+---
+
+Preserve nested multipart object field types after request body form edits.
--- a/packages/workspace-store/src/request-example/builder/body/build-request-body.test.ts
+++ b/packages/workspace-store/src/request-example/builder/body/build-request-body.test.ts
@@ -879,6 +879,53 @@ describe('buildRequestBody', () => {
     ])
   })
 
+  it('coerces regrouped dotted multipart rows using their leaf schema types', () => {
+    const requestBody = coerceValue(RequestBodyObjectSchema, {
+      content: {
+        'multipart/form-data': {
+          schema: {
+            type: 'object',
+            properties: {
+              jsonField: {
+                type: 'object',
+                properties: {
+                  isSomething: { type: 'boolean' },
+                  mood: {
+                    type: 'array',
+                    items: { type: 'string' },
+                  },
+                },
+              },
+              stringField: { type: 'string' },
+            },
+          },
+          examples: {
+            default: {
+              value: [
+                { name: 'jsonField.isSomething', value: 'false' },
+                { name: 'jsonField.mood', value: '[]' },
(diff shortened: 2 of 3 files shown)


hanspagel:
ah, sorry. I just assigned the issue to me and worked on it. I might end up using my PR.
next time, pls comment in the issue so I know you're working on it.

hanspagel:
I might be biased, but I think I’ll go with this PR #9418 :)
thanks for taking the time though

PR #9439, opened 2026-06-05
feat(json-magic): add $dynamicRef/$dynamicAnchor resolution for JSON Schema 2020-12

### LINKED ISSUE #9414: OpenAPI 3.1 schemas with $dynamicRef render as empty or unresolved shapes
## Summary

OpenAPI 3.1+ schemas that use JSON Schema 2020-12 `$dynamicRef` / `$dynamicAnchor` currently load without an error in Scalar, but the referenced schema shapes are not resolved. In practice this makes valid schemas render as empty/unknown shapes, and it affects generated examples and mock data as well.

This is easy to miss because the failure mode is silent: the document imports, but the schema shown to users is not the schema described by the OpenAPI document.

## Reproduction

The smallest fixture that demonstrates the issue is a generic pagination wrapper using `$dynamicRef` for the item type:

https://github.com/aqeelat/openapi-dynamicref-adoption-tracker/blob/main/fixtures/generic-schema-binding.yaml

The same pattern also appears in these fixtures:

- Inline response binding: https://github.com/aqeelat/openapi-dynamicref-adoption-tracker/blob/main/fixtures/paginated-response.yaml
- Double-wrapped API envelope: https://github.com/aqeelat/openapi-dynamicref-adoption-tracker/blob/main/fixtures/api-envelope.yaml
- Recursive category tree: https://github.com/aqeelat/openapi-dynamicref-adoption-tracker/blob/main/fixtures/recursive-category-tree.yaml
- Combined showcase spec: https://github.com/aqeelat/openapi-dynamicref-adoption-tracker/blob/main/petstore-dynamicref-showcase.yaml

These fixtures are valid OpenAPI 3.1 / JSON Schema 2020-12 documents. They pass spec-level validation with Redocly CLI, Spectral, openapi-spec-validator, and swagger-cli. The dynamic-reference behavior is also validated against Hyperjump JSON Schema.

## Expected Behavior

For the pagination fixture, Scalar should resolve the dynamic item binding through JSON Schema dynamic scope:

- `PaginatedUserResponse` should render `items` as `User[]`
- `PaginatedGroupResponse` should render `items` as `Group[]`
- Generated examples should include concrete `User` / `Group` objects, not empty placeholders
- Mock responses should conform to the dynamically bound concrete item type

More generally, schemas containing `$dynamicRef` should resolve against the active `$dynamicAnchor` in the dynamic evaluation path, not fall back to an empty schema or display an unresolved placeholder.

## Actual Behavior

The document loads, but `$dynamicRef` is treated as an unrecognized schema property rather than a resolvable reference keyword.

The observed failure mode is:

- Schema rendering sees an object like `{ "$dynamicRef": "#itemType" }` with no `type`, `properties`, or `$ref`, so the slot renders as empty/unknown
- Example generation has no concrete schema to walk, so dynamic slots become empty objects/null-like placeholders
- Mock data generation follows the same unresolved shape and produces empty values for dynamically bound fields

## Relevant Code Path

The issue appears to be in Scalar's shared reference-resolution layer rather than in the UI components themselves.

`@scalar/json-magic` currently discovers references via `hasRef()` in `packages/json-magic/src/bundle/bundle.ts`, which only checks for `$ref`:

```ts
const hasRef = (value: unknown): value is UnknownObject & Record<'$ref', string> =>
  isObject(value) && '$ref' in value && typeof value['$ref'] === 'string'
```

The downstream resolver path also appears to be `$ref`-only:

- `bundle()` in `@scalar/json-magic` walks and resolves `$ref`
- `dereference()` wraps the bundled document
- `resolve.schema()` in `@scalar/workspace-store` calls `getResolvedRef()`
- `getResolvedRef()` checks for `$ref`

I could not find dynamic-scope tracking for `$dynamicAnchor`, or a resolver path that treats `$dynamicRef` as a JSON Schema 2020-12 reference keyword.

The OpenAPI 3.1 `SchemaObject` types also appear to omit JSON Schema core reference fields such as `$id`, `$anchor`, `$dynamicAnchor`, and `$dynamicRef`.

## Why This Matters

OpenAPI 3.1 adopts JSON Schema 2020-12 semantics for Schema Objects, including `$dynamicRef` and `$dynamicAnchor`. OpenAPI 3.2 also explicitly calls out dynamic references as the intended mechanism for generic/template schema patterns.

The common real-world patterns affected are:

- Generic pagination wrappers like `PaginatedResponse<T>`
- Generic API envelopes like `ApiResponse<T>`
- Recursive/extendable tree structures where recursive children should resolve to the active extended type

When these schemas render as empty or unknown, users get misleading API documentation and incorrect generated examples/mock responses even though the OpenAPI document is valid.

## Notes

This seems separate from Scalar's AJV metaschema workaround in #8359. That PR replaces `$dynamicRef` with `$ref` inside Scalar's bundled OpenAPI metaschemas so AJV can validate OpenAPI documents. This issue is about user-authored schemas inside OpenAPI documents after they have been loaded.

Related prior work that may be useful context:

- #8359 handled internal OpenAPI metaschema `$dynamicRef` compatibility with AJV
- #6617 covered OpenAPI 3.1 Schema Object type alignment with JSON Schema reference semantics
- #5582 fixed `$ref` sibling handling via `mergeSiblingReferences()`

Compatibility tracker and fixtures:

https://github.com/aqeelat/openapi-dynamicref-adoption-tracker

aqeelat:
## Summary

Adds scope-aware `$dynamicRef` / `$dynamicAnchor` resolution to Scalar's reference-resolution layer (`@scalar/json-magic` and `@scalar/workspace-store`), so OpenAPI 3.1 schemas using JSON Schema 2020-12 dynamic references render correct shapes instead of empty/unknown.

Fixes #9414

### Core resolution engine (`@scalar/json-magic`)
- **`get-schemas.ts`**: Extended `getSchemas()` to collect `$dynamicAnchor` entries with resource-scoping metadata, added `SchemaMap` type with `dynamicAnchors` and `objectPaths` maps, fixed JSON Pointer escaping (`~` → `~0`, `/` → `~1`) in `getPath()`.
- **`proxy.ts`**: Added `parseDynamicRef()` (handles all URI-reference forms: plain anchors, JSON Pointer fragments, absolute URIs), `resolveDynamicAnchor()` (scope-aware lookup walking the dynamic scope from innermost to outermost), and anonymous resource path-based matching.

### Type layer (`@scalar/workspace-store`)
- **`schemas/v3.1/strict/schema.ts`**: Added `$id`, `$anchor`, `$dynamicRef`, `$dynamicAnchor`, `$defs` to the OpenAPI 3.1 `SchemaObject` type.
- **`get-resolved-ref.ts`**: Added `DynamicRefNode` type, `mergeDynamicSiblingReferences()` transform, and extended `getResolvedRef()` and `Dereference` to handle `$dynamicRef`.
- **`plugins/bundler/index.ts`**: Extended `normalizeRefs()` keep-properties set to preserve `$id`, `$anchor`, `$dynamicRef`, `$dynamicAnchor`, `$defs` during ref normalization.

### Tests & fixtures
- `bundle.test.ts` — bundler tests with dynamic anchors
- `proxy.test.ts` — unit tests for `parseDynamicRef`, `resolveDynamicAnchor`, proxy resolution, edge cases
- `dynamicref-integration.test.ts` — end-to-end integration tests using workspace-store
- `dynamicref-fthm.test.ts` — tests using Fathom-style fixture specs
- `petstore-dynamicref-showcase.yaml` — comprehensive fixture covering pagination wrappers, API envelopes, recursive trees, and combined patterns

## How to test

1. `pnpm vitest packages/json-magic --run`
2. `pnpm vitest packages/workspace-store --run`
3. Open `packages/api-reference` dev playground and load the `petstore-dynamicref-showcase.yaml` fixture — `PaginatedUserResponse` should show `User` properties, `PaginatedGroupResponse` should show `Group` properties, etc.

## Follow-up work

Some changes in this PR are independent of dynamicRef resolution and can be extracted into separate PRs:

- **JSON Pointer escaping fix** in `getPath()` — standalone bugfix affecting all `$anchor` resolution
- **JSON Schema 2020-12 type alignment** (`$id`, `$anchor`, `$defs` in schema.ts + bundler) — stops Scalar from dropping valid keywords
- **Example/mock data generation** for dynamic ref slots — not addressed here, schemas resolve correctly but generated examples may still need updates

## Ticket

Fixes #9414


hanspagel:
Thanks for the PR! This is a bigger change and I personally prefer smaller PRs, hence there is #9419. Also it's so core, that we might want to work on this ourselves.

PR #9262, opened 2026-05-19
feat(validation): support recursive schemas in validate and coerce

DemonHa:
- Add cyclic-graph handling to `validate` and `coerce` by tracking visited `(object, schema)` pairs, so self-referential nodes, mutual `lazy` graphs, and cyclic records terminate instead of overflowing the stack.
- Validate and coerce nested `intersection` members recursively; improve union branch scoring so property-less object schemas win over primitives for `{}`.
- Improve `Static` inference for circular schemas via `LazyStatic` and tuple-folding union statics; relax `union` / `intersection` call-site constraints with `UnionMember` and `IntersectionMember` so `lazy(() => self)` type-checks without depth-limit errors.
- Return `SafeStatic<S>` from `coerce` (precise `Static<S>` for concrete schemas, `any` when `S` is the full `Schema` union) and export `UnionMember` / `IntersectionMember` from the package entry point.

## Checklist

- [x] I explained why the change is needed.
- [x] I added a changeset. 
- [x] I added tests.
- [ ] I updated the documentation.

---

> [!NOTE]
> **Medium Risk**
> Touches core `@scalar/validation` runtime behavior and TypeScript surface types (cycle detection, union scoring, schema typing), which could subtly change validation/coercion results or inferred types in downstream packages.
> 
> **Overview**
> Adds **cycle-safe** `validate` and `coerce` by tracking in-progress `(value, schema)` pairs so self-referential objects/arrays/records and recursive `lazy` schemas terminate without stack overflows, while ensuring the cache is scoped to the live call stack (no cross-union-branch short-circuiting).
> 
> Improves coercion behavior by memoizing lazy schema resolution per call, pre-allocating array/object/record results to preserve cycles, scoring `union` branches with cycle detection, preferring empty `object({})` over primitives for `{}`, and recursively handling nested `intersection` members (including score accumulation).
> 
> Updates type-level API: enhances `Static` inference for circular/lazy schemas (`LazyStatic`, tuple-folded union statics), relaxes `union`/`intersection` call-site constraints via exported `UnionMember`/`IntersectionMember`, and changes `coerce` to return `SafeStatic<S>` (falls back to `any` when `S` is the full `Schema` union), with downstream call sites updated via casts and new runtime/type tests added.
> 
[...]

PR #9444, opened 2026-06-08
test(workspace-store): cover recursive $ref schema example generation

### LINKED ISSUE #9442: Maximum call stack size exceeded on openapi schema with recursive reference
### What happens?

Scalar standalone.js 1.58.0 fails to show example of response when openapi file has schema containing self reference to same schema, e.g.:

```yaml
    CommentDto:
      title: CommentDto
      allOf:
        - $ref: '#/components/schemas/Dto'
        - type: object
          properties:
            text:
              type: string
            responses:
              type: array
              items:
                $ref: '#/components/schemas/CommentDto'
```

In browsers console:

scalar.standalone.js:275 [ERROR] RangeError: Maximum call stack size exceeded
    at Object.ownKeys (scalar.standalone.js:24:340855)
    at Reflect.ownKeys (<anonymous>)
    at Qn.ownKeys (scalar.standalone.js:16:55411)
    at qf (scalar.standalone.js:24:337890)
    at K (scalar.standalone.js:24:337960)
    at Object.schema (scalar.standalone.js:28:141880)
    at f9 (scalar.standalone.js:2376:75064)
    at d9 (scalar.standalone.js:2376:74135)
    at u9 (scalar.standalone.js:2376:73701)
    at f9 (scalar.standalone.js:2376:75104)

Correct work without failures.

### OpenAPI Document

https://sandbox.scalar.com/p/z1i5l#GET/comments

hanspagel:
Verified #9442 (stack overflow on a self-referencing `CommentDto` via `allOf` + $ref) against current `main`: the response example now renders fine as `{ id, text, responses: [{}] }` with no console errors. The crash was fixed after 1.58.0 was published (recursive coerce #9262, allOf cycle-break #9194).

This adds a regression test so it stays fixed — the existing circular-reference tests only covered hand-built identity cycles, not the $ref/magic-proxy path that actually broke.

## Ticket

Closes #9442

### DIFF
--- a/packages/workspace-store/src/request-example/builder/helpers/get-example-from-schema.test.ts
+++ b/packages/workspace-store/src/request-example/builder/helpers/get-example-from-schema.test.ts
@@ -1,5 +1,6 @@
 import { describe, expect, it } from 'vitest'
 
+import { createWorkspaceStore } from '@/client'
 import { coerceValue } from '@/schemas/typebox-coerce'
 import { type SchemaObject, SchemaObjectSchema } from '@/schemas/v3.1/strict/openapi-document'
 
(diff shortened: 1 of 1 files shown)


hanspagel:
we don't need this anymore, it's fixed already

Issue #9440, opened 2026-06-06
Not work in nuxt 4 with pnpm


nont65:
### What happens?

I use @scalar/nuxt 0.6.42 with nuxt 4.4.6 in pnpm (I have already set `shamefully-hoist=true` at .npmrc) and I try with default config and fix by [this](https://github.com/scalar/scalar/issues/6367#issuecomment-3238729592). It's still not work

```
export default defineNuxtConfig({
  compatibilityDate: "2025-07-15",
  devtools: { enabled: true },
  routeRules: {
    "/docs/**": { ssr: false },
  },
  vite: {
    optimizeDeps: {
      include: [
        "@scalar/nuxt > @scalar/openapi-parser",
        "@scalar/nuxt > debug",
        "@scalar/nuxt > extend",
        "@scalar/nuxt > highlightjs-curl",
        "@scalar/nuxt > highlight.js/lib/core",
      ],
    },
  },
  nitro: {
    experimental: {
      openAPI: true,
    },
  },
  scalar: {
    hideModels: false,
    metaData: {
      title: "API Documentation by Scalar",
    },
    proxyUrl: "https://proxy.scalar.com",
    showSidebar: true,
    pathRouting: {
      basePath: "/docs",
    },
    url: "https://registry.scalar.com/@scalar/apis/galaxy?format=yaml",
  },
  modules: ["@scalar/nuxt"],
});
```

When I run `http://localhost:3000/docs` it show only `exports is not defined`

<img width="901" height="557" alt="Image" src="https://github.com/user-attachments/assets/803cea04-8e80-44b0-9002-4e280d35b9f7" />

<img width="677" height="395" alt="Image" src="https://github.com/user-attachments/assets/4986871f-6b6d-4f43-acea-7c691bd430aa" />

### What did you expect to happen?

It should display normally.

hanspagel:
Merged #9447, #9449, #9451 and should be fixed :)

PR #9447, opened 2026-06-08
fix(nuxt): pre-bundle @vercel/oidc so the docs page renders under pnpm

hanspagel:
Now that the CommonJS offenders are fixed at their source packages, the Nuxt module no longer needs to shim them:

- `cookie` → dropped from `@scalar/api-client` (#9449)
- `extend` / `debug` / `highlight.js` → bundled into `@scalar/code-highlight` (#9451)

So this drops all the CJS shims. The only thing left to handle here is `@vercel/oidc` (pulled in by the AI assistant), which Vite pre-bundles when it's hoisted into the project.

## Ticket

Part of #9440

---

> [!NOTE]
> **Low Risk**
> Integration-only change to Vite dependency optimization and removed local shims; scoped to docs rendering under pnpm with a filesystem check to avoid resolve warnings.
> 
> **Overview**
> **Removes the Nuxt module’s custom CJS workarounds** (scoped `debug`/`extend` shims and a `highlight.js` transform) now that upstream Scalar packages handle those dependencies, and **drops the build step that copied shim files** (`shx`, `shx` devDependency).
> 
> **Adds a targeted Vite fix for pnpm:** when `@vercel/oidc` is hoisted into the app’s top-level `node_modules`, the module pushes it into `optimizeDeps.include` so esbuild pre-bundles it instead of serving broken raw CommonJS from the AI assistant chain.
> 
> **E2E:** the Scalar docs smoke test now fails on typical CJS interop errors (`exports`/`require`/`export named` messages) in addition to checking the page renders.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1cefc2019aae80b125c1fecc89c5326a3580771d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


amritk (approved):
Here it is

PR #9451, opened 2026-06-08
build(code-highlight): bundle dependencies into the output

hanspagel:
`@scalar/code-highlight` now bundles its runtime dependencies into self-contained ESM at build time, so CommonJS-only packages (`extend`, `debug`) no longer leak to consumers and break Vite dev under pnpm.

## Ticket

Part of #9440

---

> [!NOTE]
> **Medium Risk**
> Build and dependency graph changes affect every consumer of the package; bundling the full markdown/highlight stack increases bundle size and could surface subtle runtime differences, though scope is limited to this library’s dist output.
> 
> **Overview**
> **`@scalar/code-highlight`** now builds with **Vite/Rolldown** instead of shipping transpiled source that still externalizes the unified/remark/rehype stack. The build **inlines all runtime dependencies** into self-contained ESM so CommonJS-only transitives (`extend`, `debug`, etc.) and `highlight.js` `require()` usage no longer reach consumers and trigger **`exports is not defined`** under pnpm + Vite dev.
> 
> The **`build` script** runs `vite build` first, then **`tsc --emitDeclarationOnly`** (plus `tsc-alias` and CSS copy) so `dist` layout still matches the existing **`exports`** map. Lib config uses shared **`findEntryPoints` / `createLibEntry`** tooling, ESM-only output, sourcemaps, and **`chunks/`** for shared chunks.
> 
> A **pre-resolve Vite plugin** forces the universal **`decode-named-character-reference`** entry instead of the browser build that touches **`document` at load**, avoiding SSR/Vitest failures after bundling.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 391b54fd5f82423e7f5234ce3fcd070bc2f870fe. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/code-highlight-bundle-deps.md
+++ b/.changeset/code-highlight-bundle-deps.md
@@ -1,0 +1,5 @@
+---
+'@scalar/code-highlight': patch
+---
+
+Bundle runtime dependencies into the build so CommonJS-only packages (like `extend` and `debug`) no longer leak to consumers and break Vite dev under pnpm
--- a/packages/code-highlight/package.json
+++ b/packages/code-highlight/package.json
@@ -21,7 +21,7 @@
     "node": ">=22"
   },
   "scripts": {
-    "build": "tsc -p tsconfig.build.json && tsc-alias -p tsconfig.build.json && shx cp -r src/css dist/css",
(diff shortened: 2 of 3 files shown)


amritk (changes requested):
oh should we delete the shims then?

amritk (approved):
nvm found it on the other PR

PR #9456, opened 2026-06-08
fix(ci): detect releases reliably so production deploys run

hanspagel:
`deploy-galaxy-production` (and every other release-gated job) has been silently skipping on release commits.

### Why

All release gates key off `needs.release.outputs.published`, which we wired to changesets/action's `published` output. That action only marks a package as published when the publish command prints changesets' own `New tag:` lines — and only `changeset publish` emits those. Our publish command is `pnpm -r publish`, so the action never sees them: `published` is always `false`, even though packages do reach npm.

Introduced in #9222, which swapped the old `RELEASING:` commit-message gate for `published == 'true'`.

So on every release the following skipped or fell back to staging:

- `deploy-galaxy-production`
- `create-github-release`
- `push-to-scalar-registry`
- the integration publishes (Docker, NuGet, PyPI, crates, Maven)
- the CDN purge
- `deploy-galaxy` / `deploy-api-client` deployed to **staging** instead of production

Confirmed on the latest release run (`chore: release` #9453, `0efa0f327`): npm-publish succeeded and published 30 packages, but `deploy-galaxy-production`, `create-github-release`, and `push-to-scalar-registry` all show **skipped**.

### Fix

Derive `published` from a version diff of public `package.json` files against the previous commit, mirroring the existing scalar-app detection step right next to it — instead of trusting the changesets output. Verified against `0efa0f327` (→ `true`, 30 packages) and ordinary commits (→ `false`).

CI-only change, so no changeset.

---

> [!NOTE]
> **Medium Risk**
> Changes how production and registry deploys are gated; incorrect detection could skip production or run it on non-release commits, though guards mirror the existing scalar-app logic.
> 
> **Overview**
> Release-gated CI (production deploys, GitHub releases, integration publishes, CDN purge) was wired to **changesets/action**’s `published` output, which stays **`false`** when publish runs via **`pnpm -r publish`** instead of **`changeset publish`**. That caused those jobs to skip on real release commits even though npm publish succeeded.
> 
> This PR sets **`published`** from a new **Detect package release** step: it scans changed **`package.json`** files, skips private packages, and marks a release when a public package’s **version** differs from **`HEAD~1`** (same pattern as the existing scalar-app bump check). The npm-publish job output and CDN purge **`if`** now use **`release-check`** instead of **`changesets`**, and workflow input descriptions are updated to match.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 054575f7b1e7f770cc8bd0921713b62fee23fd5d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/publish-integrations.yml
+++ b/.github/workflows/publish-integrations.yml
@@ -10,7 +10,7 @@ on:
   workflow_call:
     inputs:
       published:
-        description: 'changesets/action outputs.published — gate for every publish'
+        description: "release.yml's package-release detection — gate for every publish"
         type: string
         required: true
       docker_package_changed:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -103,18 +103,25 @@ jobs:
       repository-projects: write
       # Required for npm provenance statements
       id-token: write
-    # Expose the changesets/action outputs so downstream jobs can gate on the
-    # actual publish result instead of parsing the commit message. `published`
-    # is 'true' only when changesets/action actually pushed at least one
-    # package to npm; `publishedPackages` is a JSON array of {name, version}.
+    # 'true' when this commit released at least one public package — every
+    # release-gated job downstream (production galaxy deploy, GitHub release,
+    # integration publishes, CDN purge, production web deploy) keys off this.
+    #
+    # We deliberately do NOT use changesets/action's own `published` output:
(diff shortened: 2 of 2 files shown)


amritk (approved):
yea this was the exact issue I was talking about but missed again! 😓

PR #9462, opened 2026-06-09
fix(api-reference): align examples panel and disable sticky when narrow

hwkr:
Fixes the operation `ModernLayout` so the examples panel is sticky at the top of the grid instead of stretching, and stops it from being sticky on narrow viewports where it was breaking the layout.

* Add `grid-template-rows: auto auto 1fr` to `.operation-layout` and `align-self: start` on `.examples` so it stays sticky to the top
* Add `position: static` for `.examples` in the narrow media query so it's no longer sticky when there's no room for it

Before

https://github.com/user-attachments/assets/a5d4ef28-f1e6-4a45-aa5c-a20ea7e3c36f

After

https://github.com/user-attachments/assets/a1053534-5dd7-40e5-a858-03a01d175b36

Before

CleanShot 2026-06-08 at 17 52 15@2x

After

CleanShot 2026-06-08 at 18 12 21@2x

## Checklist
- [x] I added a changeset.
- [ ] I added tests.

## Ticket

Fixes DOC-5579
Part of DOC-5576


hanspagel (approved):
niice

PR #9464, opened 2026-06-09
feat(mock-server): `Prefer` header and named examples

hanspagel:
The mock server only ever returned the singular `example` and a fixed status. Now you can steer it with the `Prefer` header (same as Prism):

- `Prefer: code=404` → return that response/status
- `Prefer: example=bob` → return that named example from the `examples` map

They compose, and anything unknown just falls back to the old behavior. Also reads the OpenAPI `examples` map now, not just `example`.

---

> [!NOTE]
> **Low Risk**
> Additive mock-server behavior with fallbacks for invalid Prefer values; limited to dev/testing tooling, covered by new tests.
> 
> **Overview**
> Adds **RFC 7240 `Prefer` header** support so clients can steer mock responses (Prism-style): **`code=<status>`** selects a defined operation response when present, otherwise behavior is unchanged; **`example=<name>`** picks from the media type **`examples`** map. Directives can be combined (status first, then example). Unknown codes or names **fall back** without error.
> 
> Mock response building now uses shared **`selectResponseExample`** logic: named example → singular **`example`** → first **`examples`** entry → schema generation ( **`externalValue`-only** examples are skipped). **`mockAnyResponse`** applies both directives; **`mockHandlerResponse`** honors **`example=`** when falling back to spec examples for null/undefined handler results.
> 
> Fixes body serialization so **`null`** and other non-object primitives are not passed to **`json2xml`**. Docs and a changeset describe usage; unit and integration tests cover parsing, selection precedence, and end-to-end behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2a9313680a8c49472f0de8a37941129615383b9f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/mock-server-prefer-header.md
+++ b/.changeset/mock-server-prefer-header.md
@@ -1,0 +1,5 @@
+---
+'@scalar/mock-server': patch
+---
+
+Add `Prefer` header support to control mock responses: use `code=<status>` to request a specific response status and `example=<name>` to pick a named example from the `examples` map. Also adds support for the OpenAPI `examples` map (previously only the singular `example` was used).
--- a/documentation/guides/mock-server/getting-started.md
+++ b/documentation/guides/mock-server/getting-started.md
@@ -187,6 +187,67 @@ The given OpenAPI document is automatically exposed:
 
 - `/openapi.json` and `/openapi.yaml`
 
+### Selecting responses
+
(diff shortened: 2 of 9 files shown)


amritk (approved):
gg

bgrcs (approved):

PR #9466, opened 2026-06-09
feat(mock-server): validate requests against the OpenAPI contract

hanspagel:
The mock server now checks incoming requests against your OpenAPI document and returns a `422` (`application/problem+json`) when they break the contract, instead of always handing back a mock response.

It validates path/query parameters (with string→type coercion, so `?limit=10` is a real integer) and `application/json` request bodies. Validators are compiled once per operation, and a bad schema fails open so the server never crashes.

This is on by default. Opt out with `validateRequest: false`.

---

> [!NOTE]
> **Medium Risk**
> Default-on validation changes behavior for existing clients that sent invalid requests and still got mocks; middleware ordering and body stream handling must stay correct for x-handler routes.
> 
> **Overview**
> **OpenAPI contract validation is now on by default** for `@scalar/mock-server`. Each matched operation runs new middleware (after auth, before mock/`x-handler`) that checks path/query params and `application/json` bodies with Ajv; failures return **422** and **`application/problem+json`** listing all violations instead of a mock response. Opt out with **`validateRequest: false`**.
> 
> Route setup compiles validators once per operation (string params coerced for integer/boolean). **`onRequest`** moved to middleware so it still runs for rejected requests; mock handlers no longer take `MockServerOptions`. **`build-handler-context`** aligns JSON body parsing with validation (case-insensitive `Content-Type`, treat missing type as JSON). Body reads use a cloned stream so downstream handlers still get JSON/form bodies. Uncompilable schemas fail open per validator slice.
> 
> Adds **`ajv`** / **`ajv-formats`**, docs, changeset, and broad integration tests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2def411b816c788aa972e89f860ffcf9c318616d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/mock-server-validate-request.md
+++ b/.changeset/mock-server-validate-request.md
@@ -1,0 +1,5 @@
+---
+'@scalar/mock-server': minor
+---
+
+Validate incoming requests against the matched operation by default. Path/query parameters and the `application/json` request body are checked against their schema, and contract violations return a `422` with an `application/problem+json` body listing every violation. Set `validateRequest: false` to opt out and always return a mock response.
--- a/documentation/guides/mock-server/getting-started.md
+++ b/documentation/guides/mock-server/getting-started.md
@@ -16,6 +16,7 @@ A powerful Node.js mock server that automatically generates realistic API respon
 - Supports Swagger 2.0 and OpenAPI 3.x documents
 - Write custom JavaScript handlers for dynamic responses
 - Automatically seed initial data on server startup
+- Validates incoming requests against your OpenAPI contract
 
 ## Quickstart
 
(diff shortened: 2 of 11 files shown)


suzunn:
I noticed the middleware reads the request body before handing off to custom `x-handler` code. The test coverage already proves validation behavior, but I would add one fixture where a valid JSON request reaches an `x-handler` that also calls `c.req.json()` or `c.req.text()`. That would lock down whether Hono's body cache keeps this safe, or expose the need to validate from `c.req.raw.clone()` so custom handlers do not see an already-consumed stream.

amritk (approved):

liamdscalar (approved):

bgrcs (approved):

PR #9469, opened 2026-06-09
fix(api-reference): avoid lh inside calc() to silence postcss-calc warnings

hanspagel:
Consumer production builds that embed `@scalar/api-reference` (e.g. the Nuxt example) print PostCSS parse warnings during the build:

```
[warn] [vite:css][postcss] Parse error on line 1:
0.5lh + 4px
-^
Expecting end of input, "ADD", "SUB", "MUL", "DIV", got unexpected "UNKNOWN_DIMENSION"
```

The source is `SchemaEnumPropertyItem.vue`, which styled the enum decorator with `calc(0.5lh)` and `calc(0.5lh + 4px)`. Older `postcss-calc` versions (pulled in transitively via cssnano in those builds) cannot parse the `lh` unit inside `calc()`, so they warn and leave the value untouched.

The fix:
- Drop the redundant `calc()` from `top: calc(0.5lh)` → `top: 0.5lh`.
- Keep the half-line in a custom property (`--enum-decorator-height: 0.5lh`) so the reducer skips the expression instead of trying to parse the `lh` unit.

Rendered output is unchanged — both modern `postcss-calc` and the browser produce identical CSS.

Verified against the exact warning-producing version (`postcss-calc@9.0.1`): both `calc(0.5lh)` and `calc(0.5lh + 4px)` warn, while `top: 0.5lh` and `calc(var(--enum-decorator-height) + 4px)` produce no warnings.

### DIFF
--- a/.changeset/enum-lh-calc-warning.md
+++ b/.changeset/enum-lh-calc-warning.md
@@ -1,0 +1,7 @@
+---
+'@scalar/api-reference': patch
+---
+
+fix(api-reference): stop `lh`-in-`calc()` PostCSS warnings during consumer builds
+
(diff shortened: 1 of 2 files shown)


hanspagel:
nah, I don't like that

PR #9470, opened 2026-06-09
fix(api-reference): avoid Nuxt @unhead/vue import warning

hanspagel:
Using `@scalar/api-reference` inside a Nuxt app logs this on dev:

```
[warn] [nuxt] You are importing from `@unhead/vue` in `.../packages/api-reference/dist/helpers/map-config-to-workspace-store.js`. Please import from `#imports` instead for full type safety.
```

Nuxt's unhead module rewrites direct `useSeoMeta`/`useHead`/… imports from `@unhead/vue` to its own `#imports` composable, and warns whenever it finds such an import in application code (anything resolved outside of `node_modules`). When this package is linked into a Nuxt app as a workspace dependency, Vite resolves its built files to their real path under `packages/`, so Nuxt treats them as app code and warns.

Instead of importing the `useSeoMeta` composable directly, the SEO meta is now applied through unhead's public `headSymbol` injection — which is the same head instance both Nuxt and our own standalone setup provide. `headSymbol` is not one of the composables Nuxt rewrites, so the warning is gone while every other integration keeps working unchanged.

A test renders the head via SSR and asserts the output is identical to `@unhead/vue`'s own `useSeoMeta`, so the inline stays faithful if unhead's internals change.

### DIFF
--- a/.changeset/api-reference-unhead-imports-warning.md
+++ b/.changeset/api-reference-unhead-imports-warning.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
+Stop triggering Nuxt's `Please import from #imports instead` warning by applying SEO meta through unhead's public head injection instead of importing the `useSeoMeta` composable directly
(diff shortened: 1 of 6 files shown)


hanspagel:
what

PR #9481, opened 2026-06-10
fix(api-client): replace selected auth scheme instead of appending

hanspagel:
The Authentication dropdown for an operation with multiple OR security alternatives was additive: picking another method only appended it as a hidden second tab while the active tab stayed on the first scheme, so the fields panel looked frozen. Now selecting an OR alternative replaces the active one and the fields switch to it immediately. AND-combo requirements (e.g. `apiKeyHeader & apiKeyQuery`) remain a single option that activates both schemes at once, and additive composition is preserved for the collection builder.

### DIFF
--- a/.changeset/auth-selector-replace-or-alternative.md
+++ b/.changeset/auth-selector-replace-or-alternative.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
+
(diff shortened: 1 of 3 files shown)


hanspagel:
fixing the wrong bug in a bad way, that's not it

PR #9482, opened 2026-06-10
fix(api-reference): auth method dropdown doesn't open

hanspagel:
Opening the auth scheme dropdown in the request modal did nothing. The chevron flipped but no menu appeared, so you were stuck on the first scheme (e.g. `basicAuth`) and could not pick another.

The cause: the API reference renders two client app instances on the page (a hidden one plus the visible modal), and both set `app.config.idPrefix = 'scalar-client'`. Vue derives `useId()` - and the teleport target ids built from it - from that prefix, so both apps emitted the same id (`scalar-teleport-scalar-client-0`). The dropdown teleported its popover to that id, `querySelector` returned the first match (the hidden app), and the menu rendered off in an invisible subtree.

The fix suffixes the prefix per instance (`scalar-client-0`, `scalar-client-1`, …) so the ids stay unique and the popover lands in its own visible app.

**Before**

Screenshot 2026-06-10 at 11 45 26

**After**

Screenshot 2026-06-10 at 12 41 13

---

> [!NOTE]
> **Low Risk**
> Small, localized change to modal app initialization and a new unit test; no auth or API behavior changes beyond fixing DOM id collisions.
> 
> **Overview**
> Fixes the **auth scheme dropdown** in the request modal appearing to toggle but showing no menu when **more than one API client Vue app** is on the same page (typical API reference embed: a hidden client plus the visible modal).
> 
> Each modal app previously used the same `app.config.idPrefix` (`scalar-client`), so **`useId()` / teleport target ids collided**. The dropdown popover teleported to the first DOM match, often inside the hidden app, so it was invisible.
> 
> **`createApiClientModal`** now assigns a **monotonic per-instance prefix** (`scalar-client-0`, `scalar-client-1`, …) via a module-level counter, with a regression test asserting distinct prefixes across two modal instances.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0be6c8988ef3bb1fbcf3b671f02a79639c721401. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/auth-dropdown-teleport-collision.md
+++ b/.changeset/auth-dropdown-teleport-collision.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
+
+Fix the auth scheme dropdown not opening when multiple client apps are on the page (for example, the API reference modal). Each modal app now gets a unique id prefix, so teleport targets no longer collide and the popover renders in the visible app instead of a hidden one
--- a/packages/api-client/src/v2/features/modal/helpers/create-api-client-modal.test.ts
+++ b/packages/api-client/src/v2/features/modal/helpers/create-api-client-modal.test.ts
@@ -682,4 +682,29 @@ describe('createApiClientModal', () => {
       'x-scalar-is-dirty': true,
     })
   })
+
+  it('assigns a unique id prefix to each modal app instance', async () => {
+    const workspaceStore = await setupWorkspaceStore()
+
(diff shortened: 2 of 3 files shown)


amritk (approved):
Why are we rendering two instances? Should only be one?

hanspagel:
agent scalar renders a second modal it seems

PR #9488, opened 2026-06-10
fix(api-reference): preserve schema property order by default 🌶️

### LINKED ISSUE #9015: Attributes are shown in alphabetical order
### What happens?

I am working on some documentation, and I noticed this behaviour. The response example is correctly shown in the right order but the attributes on the left are generated alphabetically. It's not breaking but I just do not understand why this happens. 

Image

Image

I expected the documentation to be shown in the way i generated it, like the example response did

### OpenAPI Document

_No response_

hanspagel:
Schema attributes were sorted alphabetically by default, while the example response kept the document order. That mismatch was confusing (#9015).

This flips the `orderSchemaPropertiesBy` default from `'alpha'` to `'preserve'`, so the attribute list now follows the order in your OpenAPI document. Required properties still float to the top (`orderRequiredPropertiesFirst` is unchanged). If you preferred the old behavior, set `orderSchemaPropertiesBy: 'alpha'`.

## Ticket

Closes #9015

### DIFF
--- a/.changeset/preserve-schema-property-order.md
+++ b/.changeset/preserve-schema-property-order.md
@@ -1,0 +1,7 @@
+---
+'@scalar/api-reference': patch
+'@scalar/types': patch
+'@scalar/schemas': patch
+---
+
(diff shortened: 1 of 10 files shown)


hanspagel:
I'd prefer this, but it would be a breaking change for most people.

PR #9489, opened 2026-06-10
fix(api-reference): hide empty info link list

cameronrohani:
LinkList.vue renders on the DOM even when it's empty. this creates an unfortunate state where on mobile if it's empty we push down the "download openapi" button
 
 now fixed ✨ 

before:
image

after:
image

### DIFF
--- a/.changeset/fix-empty-info-links.md
+++ b/.changeset/fix-empty-info-links.md
@@ -2,4 +2,4 @@
 '@scalar/api-reference': patch
 ---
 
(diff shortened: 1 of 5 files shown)


marclave (approved):

hanspagel (changes requested):
I think we should use Vue logic and not CSS here 🤔 
Let me look into that

hanspagel (approved):

PR #9490, opened 2026-06-10
fix(api-reference): align SDK install instructions

cameronrohani:
align client library panel w/ sdk panel

before:
image

after:
image

### DIFF
--- a/.changeset/sdk-install-code-panel.md
+++ b/.changeset/sdk-install-code-panel.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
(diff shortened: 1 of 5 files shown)


marclave (approved):

cameronrohani:
@marclave asked me to just remove hljs 
image

hanspagel (changes requested):
ehhh, installation instructions can be text + multiple code blocks, e.g. for java you might want to show a gradle and maven example

and those are a block of xml

sooo.. if you want plain text, just pass plain text. we can modify that (in the sdk-generator). but we should have the option to render more than that.

hanspagel:
closing this one here! if we want to pass plain text, let's do it, but in the other repository ✌️

PR #9494, opened 2026-06-11
fix(scalar-app): watch mode for documents imported from local file paths

### LINKED ISSUE #9491: Desktop app: Watch Mode never updates documents imported from a local file path (path is sent to the HTTP fetcher)
### What happens?

When a document is imported into the **desktop app** from a **local file** and Watch Mode is enabled, the document never updates. The main process logs this error on every poll (with exponential backoff):

```
Error occurred in handler for 'customFetch': InvalidArgumentError: Invalid URL protocol: the URL must start with `http:` or `https:`.
    at Object.parseURL (...\app.asar\node_modules\undici\lib\core\util.js:185:13)
    at handleCustomFetch (file:///...app.asar/dist/main/index.js:382:47)
```

The file is never read — `handleReadFile` (`[actions]: Reading …`) is never logged during watch polling, only `customFetch` failures.

**Root cause**

The watch poller always passes the source as `url`, never as `path`:

https://github.com/scalar/scalar/blob/main/projects/scalar-app/src/features/app/hooks/use-document-watcher.ts#L164-L172

```ts
const sourceUrl = document.value?.['x-scalar-original-source-url']
// ...
const result = await storeValue.rebaseDocument({
  name: documentName,
  url: sourceUrl,   // <-- always `url`, even when the source is a local file path
})
```

`loadDocument` in `@scalar/workspace-store` dispatches on the property name, so a local path like `C:\Users\me\Documents\openapi.json` goes down the HTTP branch and is handed to undici, which (correctly) rejects it:

```ts
// packages/workspace-store — loadDocument
if ('url' in input) return fetchUrls({ fetch }).exec(input.url)   // <-- local path lands here
if ('path' in input) return fileLoader.exec(input.path)           // <-- never reached by watch mode
```

The `readFiles` loader and the desktop `fileLoader` (IPC `readFile`) work fine for the initial import — only the watch/rebase polling is affected.

**Steps to reproduce**

1. Scalar desktop app on Windows (also applies to macOS/Linux by code inspection)
2. Import an OpenAPI document from a local `.json` file
3. Enable Watch Mode in document settings
4. Change the file on disk
5. Document never updates; main process logs `Invalid URL protocol` every poll

**Suggested fix**

In `use-document-watcher.ts`, dispatch based on the shape of the source, e.g.:

```ts
const isRemote = /^https?:\/\//.test(sourceUrl)
await storeValue.rebaseDocument(
  isRemote
    ? { name: documentName, url: sourceUrl }
    : { name: documentName, path: sourceUrl },
)
```

(assuming `rebaseDocument` forwards `path` + `fileLoader` to `loadDocument` — the loader chain already supports it)

**Environment**

- scalar-app 1.1.2 (Windows 11, installed via todesktop build)
- Reproduced and root-caused against the bundled `dist/renderer` / `dist/main` of 1.1.2; code references above point at current `main`

Watch Mode should poll local file sources via the file loader (like the initial import does) and update the document when the file changes.

### OpenAPI Document

_Not document-specific — reproduces with any OpenAPI document imported from a local file path._

Gulum:
Fixes #9491

## What

In the desktop app, Watch Mode never worked for documents imported from a local file path. The watch poller always passed the document source as `url` to `rebaseDocument`, so local paths were handed to the HTTP fetcher (undici), which rejects anything that isn't http(s):

```
Error occurred in handler for 'customFetch': InvalidArgumentError: Invalid URL protocol: the URL must start with `http:` or `https:`.
```

The file loader branch in `loadDocument` was never reached by watch polling — only by the initial import.

## How

`useDocumentWatcher` now dispatches on the shape of the source using the existing `isHttpUrl` helper from `@scalar/json-magic`:

- `http(s)://…` sources → `{ url }` (unchanged behaviour)
- anything else (local file paths) → `{ path }`, which routes through the workspace store's `fileLoader` (the desktop app's IPC `readFile`)

`rebaseDocument` already accepted the `FileDoc` (`path`) input variant and forwards `workspaceProps.fileLoader` to `loadDocument`, so no store changes were needed.

## Tests

Added a test that imports a document from a real temp file via a `fileLoader`, enables watch mode, modifies the file on disk, and asserts the document gets rebased. (It uses an inline file loader because `readFiles()` from json-magic refuses to run in the jsdom test environment.)

No changeset added since `scalar-app` is private — happy to add one if you want it anyway.

---

**Update — two more fixes uncovered while testing this end-to-end on Windows:**

1. **The watcher died permanently when the source file was missing** (e.g. a build wiped a generated spec): the `readFile` IPC handler rejects on ENOENT, but the file loader assumed it resolves to `undefined` on failure. The rejection escaped through the unguarded poll and stopped polling forever after one failed attempt. Read errors are now caught in the loader (loader plugins must not throw), the poll is guarded with backoff as a second line of defense, and rebase failures are logged instead of swallowed. Regression test included (file disappears → watcher survives → file reappears → content syncs).

2. **Every rebase wiped document-level UI settings** — `x-scalar-watch-mode` (turning watch mode off after its first successful pull!), the selected server, environments, sidebar order and user-configured servers. They're now carried across rebases the same way `x-scalar-registry-meta` already was. User servers are only preserved when the upstream document defines none, so specs that declare servers stay authoritative. Regression test + changeset for `@scalar/workspace-store` included.

### DIFF
--- a/.changeset/watch-mode-preserve-document-settings.md
+++ b/.changeset/watch-mode-preserve-document-settings.md
@@ -1,0 +1,5 @@
+---
+'@scalar/workspace-store': patch
+---
+
+Preserve document-level UI settings (watch mode, selected server, environments, sidebar order) and user-configured servers across `rebaseDocument` — previously every rebase (e.g. a watch mode pull) silently wiped them
--- a/packages/workspace-store/src/client.test.ts
+++ b/packages/workspace-store/src/client.test.ts
@@ -3867,6 +3867,13 @@ describe('create-workspace-store', () => {
       // Servers configured by the user — the upstream document defines none
       active.servers = [{ url: 'http://localhost:1234' }]
 
+      // Bake the settings into the saved baseline. This is what makes the test
+      // meaningful: as unsaved local edits they would survive the rebase merge
+      // on their own, so the regression would not be caught. Once they are part
+      // of the baseline, the incoming upstream diff carries a deletion for each
+      // (upstream has none), and only the explicit preservation keeps them.
+      await store.saveDocument(documentName)
+
       const upstream = {
         ...getDocument(),
         info: { title: 'Updated upstream', version: '1.0.0' },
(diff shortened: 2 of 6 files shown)


hanspagel (approved):
nice

PR #9506, opened 2026-06-11
fix(code-highlight): add custom scroll to markdown code blocks

cameronrohani:
when we put markdown in HLJS we need to apply "custom-scroll" on it. 

working now! :) 

image


marclave (approved):

hanspagel (commented):
Approach is a bit crazy, but if it works ok, I can maybe clean it up next week

PR #9507, opened 2026-06-11
docs: dedupe themes documentation pages

hwkr:
The Docs configuration themes page duplicated the built-in theme list and the `none` example that already live on the canonical API References themes page. Remove the duplicated content and link to the shared Themes reference instead, keeping only the scalar.config.json usage that is specific to the Docs product.

Part of DOC-5575

@marclave @cameronrohani Let me know if you want them consolidated into one page somewhere outside docs or refs instead

## Checklist

- [ ] I added a changeset. 
- [ ] I added tests.


hanspagel (approved):
let's get this in, we can further tweak it anytime

PR #9508, opened 2026-06-12
fix(code-highlight): highlight python function and method call sites

hwkr:
## Problem

Python code samples looked mostly grey and unstyled next to other languages. Turns out highlight.js's Python grammar only colors `def`/`class` definitions and a fixed list of built-ins — it never colors call sites. Since our API snippets are basically all method calls (`conn.request(...)`, `response.json()`), almost everything fell back to the default grey text.

## Solution

* New custom Python grammar (`languages/python.ts`) that extends the bundled one to color function and method call sites as function titles — same trick highlight.js's own JavaScript grammar uses (word-boundary anchor + negative lookahead so keywords and built-ins keep their scopes)
* Also removed a dead duplicate `.hljs-built_in` rule in `code.css` — leftover from the Prism→highlight.js migration, it sat outside `.scalar-app` so it was already being overridden anyway

Before / After

CleanShot 2026-06-11 at 20 16 34@2x

CleanShot 2026-06-11 at 20 16 43@2x

## Checklist

- [x] I added a changeset.
- [x] I added tests.

## Ticket

Fixes DOC-5618

---

> [!NOTE]
> **Low Risk**
> Presentation-only syntax highlighting in `@scalar/code-highlight`; no auth, data, or runtime API changes, with regression tests for grammar edge cases.
> 
> **Overview**
> Python API snippets were mostly default grey because the bundled highlight.js grammar never scopes **call sites**—only `def`/`class` names and built-ins. This PR **replaces** the stock Python language with a wrapper in `languages/python.ts` that adds one rule: identifiers immediately before `(` get `title.function` (orange), with keywords/built-ins/literals/types excluded via negative lookahead and word boundaries (including a performance-conscious lookahead design for untrusted input).
> 
> **`basic.ts`** and **`standard.ts`** now import `./python` instead of `highlight.js/lib/languages/python`. **`code.css`** drops a redundant global `.hljs-built_in` block that duplicated styling already defined under `.scalar-app`. Vitest coverage in **`python.test.ts`** locks in call-site highlighting, keyword/`if (` edge cases, built-ins, and defs/classes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 231e21102dd93a40b6e05a8083d1af9ee39fa37f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/code-highlight/src/css/code.css
+++ b/packages/code-highlight/src/css/code.css
@@ -188,7 +188,3 @@
 .hljs.language-objectivec .hljs-built_in {
   color: var(--scalar-color-orange);
 }
-
-.hljs-built_in {
-  color: var(--scalar-color-orange);
-}
--- a/.changeset/python-call-site-highlighting.md
+++ b/.changeset/python-call-site-highlighting.md
@@ -1,0 +1,5 @@
+---
+'@scalar/code-highlight': patch
+---
+
+Fix Python code samples appearing mostly unstyled by highlighting function and method call sites
--- a/packages/code-highlight/src/languages/basic.ts
+++ b/packages/code-highlight/src/languages/basic.ts
@@ -6,12 +6,13 @@ import json from 'highlight.js/lib/languages/json'
 import less from 'highlight.js/lib/languages/less'
 import markdown from 'highlight.js/lib/languages/markdown'
 import plaintext from 'highlight.js/lib/languages/plaintext'
-import python from 'highlight.js/lib/languages/python'
 import scss from 'highlight.js/lib/languages/scss'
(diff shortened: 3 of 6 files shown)


packages/code-highlight/src/css/code.css
```diff
@@ -188,7 +188,3 @@
 .hljs.language-objectivec .hljs-built_in {
   color: var(--scalar-color-orange);
 }
-
-.hljs-built_in {
```
hwkr, line 192: Duplicate rule

hanspagel (approved):
niice

PR #9511, opened 2026-06-12
feat(code-highlight): improve Python syntax highlighting

amritk:
## What

Python code samples in the API reference were barely highlighted. The built-in highlight.js Python grammar only tags keywords, strings, numbers and the names in `def`/`class` _definitions_ — so a typical API snippet (mostly calls, attribute chains and keyword arguments) rendered almost entirely gray.

This extends the Python grammar and adds two theme colors so Python reads as richly as TypeScript.

## Changes

**Grammar — `packages/code-highlight/src/languages/python.ts`** (new)

A thin wrapper around the upstream highlight.js Python grammar that appends four modes, each using a string `scope` with a lookahead/lookbehind and `relevance: 0` (so auto-detection is unaffected):

| Token | Example | Scope |
|-------|---------|-------|
| Constructor / class call | `Profound(` | `title.class_` |
| Function / method call | `client.reports.get(` | `title.function_` |
| Attribute / property access | `client.reports` | `property` |
| Keyword argument / assignment | `api_key_header=`, `client =` | `attr` |

Ordering ensures `.method(` reads as a call rather than a plain attribute. The modes are appended after the upstream ones, so strings, comments and `def`/`class` titles keep priority and the new modes never fire inside a string or comment. Reserved words are excluded (so `if (...)` stays a keyword and `importlib` is not mistaken for `import`), and comparisons (`==`, `!=`, `>=`, `:=`) are not treated as assignments.

**Theme — `packages/code-highlight/src/css/code.css`**

- `.hljs-title.class_` now uses `--scalar-color-green` (was plain foreground) so constructors and types stand out.
- New `.hljs-property` rule uses `--scalar-color-yellow` (previously unused for syntax) so attribute chains get their own hue.

> Note: both theme rules are global, so **TypeScript and JavaScript** property chains and class names pick up the same colors. Worth an eyeball in `api-reference` during review.

**Tests** — `packages/code-highlight/src/languages/python.test.ts` (new), 10 cases covering class vs function vs property, comparisons, strings/comments and reserved-word edges. Includes a changeset.

## Visual

Before → after of the original snippet — constructor green, calls orange, property chain yellow, keyword args brightened:

| | Token | Color |
|---|---|---|
| Constructor / class | `Profound` | green |
| Function / method call | `get`, `query_…`, `print` | orange |
| Property / attribute access | `environ`, `reports`, `visibility` | yellow |
| Keyword arg / assignment / object key | `date_interval=` | bright foreground |
| Strings | `"day"` | blue |
| Keywords | `from`, `import` | purple |

_Before/after screenshots were shared with the maintainer in the working session._

### DIFF
--- a/.changeset/python-syntax-highlighting.md
+++ b/.changeset/python-syntax-highlighting.md
@@ -1,0 +1,5 @@
+---
+'@scalar/code-highlight': patch
+---
+
+Improve Python syntax highlighting: function and method calls, class instantiation, and keyword arguments are now colored. The built-in highlight.js Python grammar only tagged keywords, strings and definitions, which left most of a typical API reference snippet (mostly calls and keyword arguments) unstyled.
--- a/.changeset/python-syntax-highlighting.md
+++ b/.changeset/python-syntax-highlighting.md
@@ -2,4 +2,4 @@
 '@scalar/code-highlight': patch
 ---
 
-Improve Python syntax highlighting: constructor calls, function and method calls, attribute access, and keyword arguments are now colored. The built-in highlight.js Python grammar only tagged keywords, strings and definitions, which left most of a typical API reference snippet (mostly calls, attribute chains and keyword arguments) unstyled. Class titles are now colored green and property access gets a color in the theme, so method chains light up across all languages.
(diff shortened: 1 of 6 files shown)


amritk:
Not that great Claude, bad work

Issue #9512, opened 2026-06-12
APIReference - `onBeforeRequest` - multipart body hash does not match the sent request


marcalexiei:
### What happens?

Computing a body hash inside `onBeforeRequest` no longer works for `multipart/form-data` requests.

The `request` object passed to the hook is built from the `requestBuilder` inside the hook wrapper. 
After the hook returns, the actual fetch `Request` is rebuilt from `requestBuilder` again. 
Because each `new Request(url, { body: formData })` call generates a fresh random multipart boundary, the bytes the user hashed and the bytes that are actually sent are different.

I'm using `@scalar/fastify-api-reference@1.51.0` where this was previously achievable: .

## Steps to reproduce

```js
onBeforeRequest: async ({ request }) => {
  const clone = request.clone()
  const bytes = await clone.arrayBuffer()
  const digest = await crypto.subtle.digest('SHA-256', bytes)
  const hash = btoa(String.fromCharCode(...new Uint8Array(digest)))

  // `hash` does not match the body received by the server for multipart requests
}
```

For `multipart/form-data` requests the multipart boundary in `request` differs from the
boundary in the request that is actually sent, so the hash is always wrong.

### What did you expect to happen?

The `request` passed to `onBeforeRequest` represents the exact bytes that will be sent,
so a hash computed from it can be used for request signing.

hanspagel:
Fixed with #9515 I think :)

marcalexiei:
Me too, I'll check if the issue is fixed in the real app once https://github.com/scalar/scalar/pull/9560 is merged!

marcalexiei:
By using the new `onRequestBuilt` hook I was able to restore automatic request body signing in the documentation. 

Thanks!

PR #9515, opened 2026-06-12
feat(api-reference): add `requestBuilt` hook and `onRequestBuilt` callback with the exact outgoing `Request`

marcalexiei:
## Problem

- See #9512

`onBeforeRequest` used to make request signing possible: 
you could hash the request body and attach signature headers before the request was sent. 
This is no longer achievable.

The `request` passed to `onBeforeRequest` is built from the request builder *inside the hook wrapper*, and after the callback returns the actual fetch `Request` is rebuilt from the builder. 
For `multipart/form-data` bodies every rebuild generates a fresh random multipart boundary, so the bytes you hashed in the callback never match the bytes the server receives — signature verification always fails. 
Header mutations on `request` are silently discarded for the same reason.

Mutating `requestBuilder` (the documented path) does not help here: 
there is no way to know the final multipart body bytes before the fetch `Request` is built.

## Solution

Build the fetch `Request` once and send that exact instance, and expose it through a new hook that runs after the build:

- **`@scalar/oas-utils`**: new `requestReady` client plugin hook. It fires after the fetch `Request` is built, right before it is sent. The `request` in its payload is the exact object handed to fetch, so header mutations apply and body hashes match what goes over the wire.
- **`@scalar/api-client`**: `OperationBlock` builds the `Request` once after the `beforeRequest` hooks run, executes `requestReady` on it, and `sendRequest` sends that same instance instead of rebuilding it.
- **`@scalar/api-reference` / `@scalar/types` / `@scalar/schemas`**: new `onRequestReady` configuration callback (sibling of `onBeforeRequest`, same payload shape) mapped to the `requestReady` hook. It works through the HTML/CDN integrations (Fastify, Express, ...) the same way `onBeforeRequest` does.

`onBeforeRequest` is untouched, so the change is purely additive: builder mutations (method, path, query, body, `disableSecurity`, ...) keep working exactly as documented. 
The docs now clarify that the `request` passed to `onBeforeRequest` is not the wire request and point to `onRequestReady` for exact-bytes use cases.

Tests include a regression test that computes a SHA-256 hash of a multipart body inside the hook and asserts it equals the hash of the request handed to fetch — and that a request rebuilt from the same `FormData` hashes differently, so reintroducing a rebuild fails the test.

## Checklist

- [x] I added a changeset.
- [x] I added tests.

### DIFF
--- a/.changeset/spicy-falcons-shake.md
+++ b/.changeset/spicy-falcons-shake.md
@@ -1,0 +1,11 @@
+---
+'@scalar/api-reference': minor
+'@scalar/oas-utils': minor
+'@scalar/api-client': minor
+'@scalar/schemas': minor
+'@scalar/types': minor
+---
+
+feat: add `requestReady` client plugin hook and `onRequestReady` configuration callback that receive the exact fetch `Request` that is sent over the wire
+
+The hook runs after the request has been built, right before it is sent. Header mutations apply to the outgoing request and the body bytes match what the server receives, which makes request signing possible: hashing the body of a rebuilt `multipart/form-data` request would produce a different multipart boundary than the request that is actually sent.
--- a/.changeset/spicy-falcons-shake.md
+++ b/.changeset/spicy-falcons-shake.md
@@ -6,6 +6,6 @@
 '@scalar/types': minor
 ---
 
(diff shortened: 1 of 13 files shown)


hanspagel (approved):
nice work
I just renamed them, but otherwise: LGTM

PR #9517, opened 2026-06-12
fix(mock-server): prefer defined 2xx responses

DemonHa:
Prefer defined 2xx responses before error responses when mocking operations with multiple response codes.

## Checklist

- [x] I added a changeset. 
- [x] I added tests.

---

> [!NOTE]
> **Low Risk**
> Behavior change is limited to mock-server default response selection when no Prefer header is set; explicit Prefer codes and error handling paths are unchanged, with broad test coverage.
> 
> **Overview**
> **Default mocked responses** no longer follow a fixed key list that could pick `default` ahead of `200`. `findPreferredResponseKey` now ranks OpenAPI response keys in order: lowest **2xx**, then lowest non-1xx code, then **`default`**, then 1xx only if nothing else exists. Explicit codes beat range keys like **`2XX`** at the same tier; ranges sort by their lowest member (e.g. `2XX` → 200).
> 
> **HTTP status on the wire** in `mockAnyResponse` maps selected keys the same way: `default` still becomes **200**, and range keys get **`XX` → `00`** so a chosen `2XX` response returns status **200** while using that response’s body.
> 
> Unit and integration tests cover success-vs-error, default fallback, and range behavior. **`Prefer: code=`** selection is unchanged.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 81c078d9c126bc0e59209db5384f822fccb64884. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/mock-server/src/utils/find-preferred-response-key.ts
+++ b/packages/mock-server/src/utils/find-preferred-response-key.ts
@@ -1,12 +1,23 @@
+const sortStatusCodes = (responses: string[]): string[] =>
+  responses
+    .filter((response) => /^\d{3}$/.test(response))
+    .sort((left, right) => Number.parseInt(left, 10) - Number.parseInt(right, 10))
+
 /**
- * Find the preferred response key: default, 200, 201 …
+ * Find the preferred response key: default, a 2xx success, then the lowest status code.
  */
 export function findPreferredResponseKey(responses?: string[]) {
+  if (!responses?.length) {
+    return undefined
+  }
+
+  if (responses.includes('default')) {
+    return 'default'
+  }
+
+  const statusCodes = sortStatusCodes(responses)
+
   return (
-    // Regular status codes
-    ['default', '200', '201', '204', '404', '500'].find((key) => responses?.includes(key) ?? false) ??
-    // Lowest status code
-    responses?.sort()[0] ??
-    undefined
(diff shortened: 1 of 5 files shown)


marclave (approved):

packages/mock-server/src/utils/find-preferred-response-key.ts
```diff
@@ -1,12 +1,23 @@
...
+
+  const statusCodes = sortStatusCodes(responses)
+
   return (
-    // Regular status codes
-    ['default', '200', '201', '204', '404', '500'].find((key) => responses?.includes(key) ?? false) ??
-    // Lowest status code
-    responses?.sort()[0] ??
-    undefined
+    statusCodes.find((response) => response.startsWith('2')) ?? statusCodes[0] ?? [...responses].sort()[0] ?? undefined
```
hanspagel: we should make sure
* 'default' comes first
* 100/1xx doesn't come first

related:
https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.1.md#patterned-fields-1

hanspagel (changes requested):

hanspagel (approved):
good to go now

PR #9520, opened 2026-06-13
feat(api-reference): render AsyncAPI operations and messages inside channels

bielu:
## Problem

Scalar already lists AsyncAPI operations and messages in the sidebar, but the content area stops at the channel — `Channel.vue` only renders the channel heading, description, and address parameters. The navigation traversal builds the full `channel → operations → messages` tree and gives every node an id, so clicking an operation or message in the sidebar scrolls to a channel that shows nothing about it.

This makes Scalar unusable for documenting my AsyncAPI: the operations and their message payloads never appear. I want Scalar to reach parity with the AsyncAPI UI (which does render operations and message payloads under each channel), since that tool doesn't fit my use case and I'd rather standardize on Scalar.

## Solution

Render the operations — and their messages with payload/header schemas — nested inside each channel, mirroring the existing channel → operation → message sidebar nesting.

- **`Operation.vue` (new):** renders an operation with a `send`/`receive` action badge, title, summary/description, and its nested messages. Resolves the operation from `document.operations` and merges traits via the existing `resolveOperationWithTraits` helper. Wires the operation id for sidebar anchor/scroll-spy.
- **`Message.vue` (new):** renders a message title/description plus its **Headers** and **Payload** schemas using the shared `Schema` component (same approach as `RequestBody.vue`).
- **`get-async-api-message-payload-schema.ts` (new):** `unwrapAsyncApiSchema` resolves `$ref`s and unwraps Multi-Format Schema wrappers into the `SchemaObject` the `Schema` component expects, plus payload/header accessors. The existing `get-async-api-model-schema.ts` was refactored to reuse this instead of duplicating the unwrap logic.
- **`Channel.vue` (modified):** renders its operation children inside the content area in both the classic (accordion body) and modern (`SectionContent`) layouts, after the existing parameters block.

Rendering goes directly through dedicated components (Channel → Operation → Message) rather than routing operations back through the recursive `AsyncApiTraversedEntry`, to avoid an import cycle and keep each level's layout self-contained.

### Verification

- New tests: `Operation.test.ts`, `Message.test.ts`, `get-async-api-message-payload-schema.test.ts`; extended `Channel.test.ts` to cover nested operation rendering in both layouts.
- `vitest --run` for the touched files: all green. `vue-tsc --noEmit`: 0 errors. `eslint`: 0 errors.
- Manually verified against the Galaxy AsyncAPI doc (the **Scalar Galaxy Events (AsyncAPI)** sources in the dev playground) in both modern and classic layouts.

## Checklist

- [x] I added a changeset. 
- [x] I added tests.

### DIFF
--- a/.changeset/asyncapi-channel-operations.md
+++ b/.changeset/asyncapi-channel-operations.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': minor
+---
+
+Render AsyncAPI operations and their messages (with payload and header schemas) nested inside each channel, in both the modern and classic layouts
--- a/documentation/asyncapi.md
+++ b/documentation/asyncapi.md
@@ -2,7 +2,7 @@
 
 We're in the process of adding [AsyncAPI](https://www.asyncapi.com/) support.
 
-You can already load AsyncAPI documents. The rendering is very limited, though:
+You can load AsyncAPI documents the same way you load OpenAPI documents:
 
 ```javascript
 Scalar.createApiReference('#app', {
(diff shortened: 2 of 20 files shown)


hanspagel (approved):
insane work!

I think we want to tweak the rendering, but it's definitely better than not rendering them 😅

so let's get this in

PR #9527, opened 2026-06-15
test(api-reference): cover missing configuration options

hanspagel:
Adds e2e tests for the configuration options that had no coverage — slug generators, sorters, event callbacks, auth/oauth, proxy, plugins, and the various toggles.

34 new files under `test/configuration/`, all passing against the standalone build.

A few options turned out to be quirky and are worth separate follow-ups (handled in their own PRs): `onSpecUpdate` and `isLoading` are unwired, and `generateHeadingSlug` isn't honored by sidebar navigation.

---

> [!NOTE]
> **Low Risk**
> Test-only additions with no production code changes; main risk is CI runtime and flaky e2e if the standalone harness or UI selectors shift.
> 
> **Overview**
> Adds **34 Playwright e2e suites** under `test/configuration/` that exercise previously untested `@scalar/api-reference` config via the existing **`serveExample`** standalone harness.
> 
> Coverage spans **UI toggles** (tag/model/response expansion, `showOperationId`, favicon, `metaData`, search hotkey), **slug generators** (tag, operation, model, webhook, heading), **sorters and schema ordering** (`tagsSorter`, `operationsSorter`, `orderSchemaPropertiesBy`, `orderRequiredPropertiesFirst`), **auth and networking** (API key prefills, OAuth redirect URI, `customFetch`, `proxyUrl`), **lifecycle hooks** (`onLoaded`, `onBeforeRequest`, `onRequestSent`, sidebar/server/document/show-more callbacks), **plugins**, and **agent** visibility.
> 
> A few tests are intentionally **smoke-level** where behavior is hard to observe in the OSS standalone build (`isLoading`, `telemetry`) or document known quirks called out for follow-up (`generateHeadingSlug` vs sidebar navigation).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d2b012c421001f62218848337f8aa5e0b83c8428. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #9530, opened 2026-06-15
fix(api-reference): use `generateHeadingSlug` in sidebar navigation

hanspagel:
Sidebar navigation ignored `generateHeadingSlug`: clicking a description heading went to the plain anchor instead of the custom slug, unlike the other four `generate*Slug` options.

The cause is that the custom heading id was returned without the `${documentId}/description/` prefix the other generators get, so the URL round-trip (which strips the leading document segment) ate the custom part. This prefixes it consistently on both the navigation and content sides.

**Contract change:** `generateHeadingSlug` should now return a bare slug segment (e.g. `heading.slug`) rather than a full `#description/...` hash. Docs updated to match.

Verified in the browser: clicking a heading now navigates to `#description/custom-…`.

Note: the heading e2e test in #9527 asserts the old behavior and should be updated to assert the URL once this lands.

Found while adding configuration test coverage (#9527).

---

> [!NOTE]
> **Medium Risk**
> Minor semver bump reflects a breaking change for integrators who still return full `#description/...` hashes from `generateHeadingSlug`; navigation and deep-link behavior for description headings changes site-wide.
> 
> **Overview**
> Sidebar clicks on description headings now use **`generateHeadingSlug`** the same way as the other `generate*Slug` hooks, instead of falling back to the default anchor.
> 
> **`generateHeadingSlug` contract:** callers return only the trailing slug segment (e.g. `heading.slug` or `custom-${heading.slug}`). The framework prepends `${documentSlug}/description/` in **`getNavigationOptions`** (sidebar IDs) and in **`ApiReference.vue`** (content heading IDs), matching `model/`, `tag/`, etc.
> 
> Documentation and a Playwright e2e cover default and custom URL hashes (`#description/overview`, `#description/custom-overview`). Unit expectations in **`get-navigation-options.test.ts`** were updated for the prefixed IDs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8fd9daf466ea7ec3153a34e72643b4ab7dfa9f7b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/honor-heading-slug.md
+++ b/.changeset/honor-heading-slug.md
@@ -1,0 +1,8 @@
+---
+'@scalar/workspace-store': minor
+'@scalar/api-reference': minor
+---
+
+Make sidebar navigation honor `generateHeadingSlug`, consistent with the other `generate*Slug` options. The custom value is now prefixed with `description/` (like `model/`, `tag/`, etc. are for the other generators), so clicking a description heading navigates to the custom slug instead of the plain anchor.
+
+Note: `generateHeadingSlug` should now return a bare slug segment (e.g. `heading.slug`) rather than a full hash like `#description/...`.
--- a/documentation/configuration.md
+++ b/documentation/configuration.md
@@ -1155,19 +1155,19 @@ Custom [fetch function](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_A
 
 **Type:** `(heading: Heading) => string`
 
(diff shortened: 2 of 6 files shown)


hanspagel:
actually, I think we should keep it as-is
because streamlining it, would be a breaking change
and people might just want to modify the prefix, too

PR #9533, opened 2026-06-15
feat(scalar-app): make watch mode source URL editable

### LINKED ISSUE #9493: Desktop app: Watch Mode source URL cannot be changed in the UI
### What happens?

In the desktop app, the Watch Mode **Source** shown in the document settings (`x-scalar-original-source-url`) is read-only — it's rendered as a plain link with no way to edit it:

https://github.com/scalar/scalar/blob/main/projects/scalar-app/src/features/settings/DocumentSettings.vue#L150-L161

Once a document has been imported, there is no way in the UI to re-point it at a different source. The only workarounds are:

- delete the document and re-import it from the new source (losing settings such as the selected server, environments, auth, …), or
- close the app and hand-edit the record in the IndexedDB LevelDB store under `%APPDATA%\Scalar` (which is how I ended up here).

This comes up quite naturally:

- the spec file was moved/renamed on disk (e.g. exported download → a build-generated path)
- switching a document from a downloaded file to the live `…/openapi/v1.json` endpoint of a dev server (also the practical workaround for #9491, since file sources currently can't be watched at all)
- a dev server changed ports

The Source field in document settings should be editable (or have an "Edit" affordance), updating `x-scalar-original-source-url` on the document. The plumbing already exists — e.g. the generic `document:update:extension` event merges extension fields into the document root, and the watch poller already reacts to source-url changes:

https://github.com/scalar/scalar/blob/main/projects/scalar-app/src/features/app/hooks/use-document-watcher.ts#L193

so an edited source would immediately be picked up by Watch Mode without a restart.

### OpenAPI Document

_Not document-specific._

hanspagel:
The Watch Mode **Source** in document settings was a read-only link, so once a document was imported there was no way to re-point it at a different source short of deleting and re-importing it (losing servers, auth, environments…) or hand-editing the IndexedDB store.

This makes the source editable inline. **Edit** an existing source, or **Add** one when none is configured. Saving emits the generic `document:update:extension` event to update `x-scalar-original-source-url`, which the watch poller already reacts to, so the new source is picked up without a restart.

## Visual

Screenshot 2026-06-15 at 17 48 18

## Ticket

Closes #9493

---

> [!NOTE]
> **Low Risk**
> Scoped UI and document extension updates reusing an existing event path; no auth or new polling logic.
> 
> **Overview**
> **Watch Mode source URL** in document settings is no longer read-only: users can **Edit** an existing URL or **Add** one when missing, then save inline without deleting and re-importing the collection.
> 
> Saving emits `update:documentUrl`, which `Settings` maps to `document:update:extension` with `x-scalar-original-source-url` so the existing watch poller picks up the new source without a restart. The UI handles trim, no-op saves, cancel, empty-source guards, and draft sync when the prop updates while editing.
> 
> Tests cover the new `DocumentSettings` behavior and the `Settings` event wiring; a patch changeset documents the user-facing change.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bc296957171f133898cb51bb2ec5a876efd8169a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


marclave (approved):
looks good just needs cursor fixes :)

PR #9540, opened 2026-06-16
fix(use-hooks): SSR hydration mismatch in color-mode toggle

hanspagel:
The dark-mode toggle read the system preference (`matchMedia`) during render, so the server sent light and a dark-preferring client hydrated dark — a mismatch. Now we default to the server value and resolve the real preference in `onMounted`.

Part of #4458

---

> [!NOTE]
> **Low Risk**
> Theming/SSR-only behavior change in `@scalar/use-hooks`, covered by new unit tests; users on system-dark may see a brief light state after hydration before the preference applies.
> 
> **Overview**
> Fixes an SSR hydration mismatch when color mode is **system**: the hook no longer reads `matchMedia` during the first render, so the client’s initial HTML matches the server (which effectively renders as light).
> 
> **`useColorMode`** now keeps a shared module-level **`systemPreference`** ref defaulting to `'light'`. **`darkLightMode`**, body **`light-mode`/`dark-mode`** classes, and the color-mode toggle all use that ref instead of calling **`getSystemModePreference()`** synchronously. The real OS preference is set in **`onMounted`**, and OS theme changes update the same ref (with **`watch`** on **`colorMode`** + **`systemPreference`** driving **`applyColorMode`**).
> 
> Tests assert pre-mount vs post-mount behavior, body class alignment, and that multiple hook instances share one resolved preference. Changeset: patch for **`@scalar/use-hooks`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c3c0ced70b6728e5bdc8c05582b58209c3043d8d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/fix-color-mode-hydration.md
+++ b/.changeset/fix-color-mode-hydration.md
@@ -1,0 +1,5 @@
+---
+'@scalar/use-hooks': patch
+---
+
+Fix an SSR hydration mismatch in the color-mode toggle: the system preference is now resolved after mount so the first client render matches the server.
--- a/packages/use-hooks/src/useColorMode/useColorMode.test.ts
+++ b/packages/use-hooks/src/useColorMode/useColorMode.test.ts
@@ -1,5 +1,5 @@
 import { beforeEach, describe, expect, it, vi } from 'vitest'
-import { nextTick } from 'vue'
+import { nextTick, onMounted } from 'vue'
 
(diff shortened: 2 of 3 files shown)


hwkr (approved):
Looks good! Thank you!

PR #9544, opened 2026-06-16
test(api-reference): SSR hydration regression test

hanspagel:
A generic guard for the SSR hydration mismatches in #4458: it server-renders the API reference with `renderToString`, hydrates the same markup, and fails if Vue reports any hydration mismatch (`Hydration completed but contains mismatches.`, plus per-node `[Vue warn]`s in dev builds).

**This is expected to be red right now.** It stays failing until every mismatch in #4458 is fixed, then goes green.

## What's left to make it pass

Right now the server renders the full document, but the client hydrates a loading skeleton instead — so the two don't match. To fix it:

- **Give the client the document up front.** The server preloads it (`onServerPrefetch`), but the client loads it in an async `onBeforeMount` that Vue doesn't wait for, so the first render is empty. We need the document available synchronously at hydration (pass the SSR state through to `ApiReference` instead of re-fetching on the client).
- **Make `Lazy` render the same on both sides.** On the server it renders every operation; on the client it shows placeholders until each one is queued. It should render a placeholder on the server too when the chunk isn't ready, so both sides agree.

Both are part of the chunked-SSR work in #4458. This test only goes green once both are done.

Part of #4458


hanspagel:
I think we can close this in favor of #9982

PR #9550, opened 2026-06-16
feat(scalar): add `@scalar/scalar` ESM CDN package

hanspagel:
The old UMD build is here, the URL is nice and short:

🚨 `https://cdn.jsdelivr.net/npm/@scalar/api-reference`

But we have this ESM CDN now. Unfortunately, to load it you need a pretty long URL:

🫨 `https://cdn.jsdelivr.net/npm/@scalar/api-reference/dist/browser/standalone.esm.js`

We could shorten the URL to be:

😔 `https://cdn.jsdelivr.net/npm/@scalar/api-reference/esm.js`

Too bad jsdelivr doesn't support entrypoints, then we could drop the extension at least:

❌ `https://cdn.jsdelivr.net/npm/@scalar/api-reference/esm`

we could use a new CDN provider, but I personally think jsdelivr is the most reliable provider

sooo … here's an alternative: a new package.

✅ `https://cdn.jsdelivr.net/npm/@scalar/scalar`

## Usage

```ts
import { createApiReference } from 'https://cdn.jsdelivr.net/npm/@scalar/scalar'

createApiReference('#app', {
  url: 'https://registry.scalar.com/@scalar/apis/galaxy?format=json',
  // Avoid CORS issues
})
```

Once that is released/published, we can update the docs, README, `@scalar/client-side-rendering` … and so on.


amritk (approved):
Yea lets get the ESM build mainstream for all those sweet lazy savings

geoffgscott:
IMO this is not a great idea. The package name should be representative of its scope. @scalar/scalar has no sematic association. 

This `https://cdn.jsdelivr.net/npm/@scalar/api-reference/esm.js` looks very reasonable and inline with best practices.

hanspagel:
I’m not really happy with either solution, that's why it's still in draft. 👍 I try to come up with something better.

hanspagel:
let's do `https://cdn.jsdelivr.net/npm/@scalar/api-reference/esm.js` instead

PR #9557, opened 2026-06-17
fix(api-reference): preserve `oneOf`/`anyOf` nested inside `allOf`

### LINKED ISSUE #5577: Multiple Nested Schema References not followed
### What happens?

I have an API that has nested schema references. Scalar only renders the first element and none of the other elements. 

It ignores the next `oneOf` which has more nested schema. 
(related to #4434)

## Excerpt of request schema
```yaml
ConversionCreationRequest:
      title: ConversionCreationRequest
      allOf:
      - type: object
        properties:
          customerComment:
            $ref: '#/components/schemas/Comment_Customer'
      - oneOf:
        - allOf:
          - title: With Quote Id
            type: object
            properties:
              quoteId:
                $ref: '#/components/schemas/QuoteId'
          - $ref: '#/components/schemas/EitherSourceOrDestinationAmount'
        - allOf:
          - title: With Currency Pair
            type: object
            properties:
              sourceCurrencyCode:
                $ref: '#/components/schemas/SCC'
              destinationCurrencyCode:
                $ref: '#/components/schemas/DCC'
          - $ref: '#/components/schemas/EitherSourceOrDestinationAmount'
```
Only the first element `customerComment` is rendered. There are several references nested within `oneOf` -> `allOf` etc
Scalar is only following the first element and dropping the rest. 

You can see this running on my site below

- reference: [Create Conversion](https://docs.nium.com/api#tag/conversions/POST/api/v1/client/{clientHashId}/customer/{customerHashId}/wallet/{walletHashId}/conversions)
- schema file: https://docs.nium.com/oas/nium.yaml
  - GitHub Friendly Version: https://github.com/nium-global/nium-openapi/blob/main/nium.yaml#L31692

## Swagger.io
I checked the schema in swagger.io to verify

Image

Render all nested schema, follow the schema references (`allOf` `oneOf`)

### OpenAPI Document

https://docs.nium.com/oas/nium.yaml

hanspagel:
When a `oneOf`/`anyOf` sits next to other members inside an `allOf`, we were flattening it away, so only the first member rendered and the variants disappeared (worst case: branches that are themselves an `allOf` left no trace at all).

Now the composition is kept, so the variant selector renders alongside the merged base properties.

## Ticket

Fixes #5577

---

> [!NOTE]
> **Low Risk**
> Scoped display/merge change in `@scalar/api-reference` schema helpers with tests; no auth or runtime API behavior.
> 
> **Overview**
> **`mergeAllOfSchemas` no longer flattens `oneOf`/`anyOf` into merged `properties` when those compositions appear as siblings inside an `allOf`.** It now copies the composition arrays onto the merged schema (when not already set), so base fields from other `allOf` members still merge while variant selectors can render.
> 
> This fixes cases like **#5577** where a `oneOf` next to a base object in `allOf` disappeared from the docs—especially branches that only define an inner `allOf` without top-level `properties`. Unit expectations and a **`SchemaComposition`** mount test were updated to assert preserved `oneOf`/`anyOf` alongside merged properties.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0d4b384469d797768cdf72860c360a973cbb1d86. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/merge-allof-oneof-preserve.md
+++ b/.changeset/merge-allof-oneof-preserve.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
+Fix `oneOf`/`anyOf` being dropped when nested inside an `allOf`. The composition is now preserved so its variants keep rendering alongside the merged base properties, instead of only the first `allOf` member showing up.
--- a/packages/api-reference/src/components/Content/Schema/SchemaComposition.test.ts
+++ b/packages/api-reference/src/components/Content/Schema/SchemaComposition.test.ts
@@ -416,4 +416,41 @@ describe('SchemaComposition', () => {
       },
     ])
   })
+
+  it('keeps a oneOf composition nested inside allOf (issue #5577)', () => {
+    const wrapper = mount(SchemaComposition, {
+      props: {
+        eventBus: null,
+        composition: 'allOf',
+        schema: coerceValue(SchemaObjectSchema, {
+          title: 'ConversionCreationRequest',
+          allOf: [
+            {
+              type: 'object',
+              properties: {
+                customerComment: { type: 'string' },
+              },
+            },
+            {
+              oneOf: [
+                { title: 'With Quote Id', type: 'object', properties: { quoteId: { type: 'string' } } },
+                { title: 'With Currency Pair', type: 'object', properties: { sourceCurrencyCode: { type: 'string' } } },
+              ],
+            },
+          ],
+        }),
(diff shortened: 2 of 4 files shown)


amritk (approved):
Once again I got no idea what we do with this stuff but I presume you do

marclave (approved):

PR #9558, opened 2026-06-17
fix(api-client): auth token with multi-document sources dropped

### LINKED ISSUE #9554: Bub: Typing a token in the runner sends Authorization: Bearer <token> regardless of whether the document was loaded via content or sources.
## Description

  When an API reference is configured with the multi-document `sources: [...]` API, a Bearer token **typed into the request runner's Authentication panel** is not applied to the outgoing request — the `Authorization` header is sent as a bare `Bearer` with no token. Switching the same spec to the single-document `content` API fixes it.

  ## Versions

  - `@scalar/server-side-rendering` 0.1.25 → `@scalar/api-reference` 1.59.3 (also reproduces on 1.58.0)
  - `@scalar/workspace-store` 0.53.0, `@scalar/api-client` 3.9.0

  ## Security scheme (standard HTTP bearer)

  ```yaml
  components:
    securitySchemes:
      userAuth: { type: http, scheme: bearer }
  security:
    - userAuth: []
  ```

  ## Reproduction

  Render the same document two ways and, in each, open an operation's request runner, type any token into the Bearer Token field, and Send:

  | Config passed to `renderApiReference` | Outgoing `Authorization` |
  | --- | --- |
  | `{ sources: [{ slug, content }] }` (even a **single**-element array) | `Bearer` ❌ (token dropped) |
  | `{ content }` (single document) | `Bearer <token>` ✅ |

  The token value is visibly present in the input field in both cases; only the `sources` path fails to forward it. Confirmed via request interception.

  Config-level `authentication.securitySchemes.userAuth.token` **is** applied correctly under `sources`, so only the *manual-entry → store* binding is broken in the
  workspace-store path.

Typing a token in the runner sends Authorization: Bearer <token> regardless of whether the document was loaded via content or sources.

### OpenAPI Document

_No response_

hanspagel:
Typing a Bearer token in the request runner sent an empty `Authorization: Bearer` when the document was loaded via `sources` (single-document `content` worked fine).

The runner stores auth secrets under the document name but the modal handed the Operation the slugified navigation `id` instead. When the name was not slug-safe, reads and writes used different keys and the token was dropped. The modal now passes the document name.

## Ticket

Fixes #9554

---

> [!NOTE]
> **Low Risk**
> Single prop wiring change in the modal plus a targeted test; aligns with existing auth store keying by document name.
> 
> **Overview**
> Fixes the request runner sending **`Authorization: Bearer`** with no token when a document name is not slug-safe (typical for **`sources`** loads), because auth secrets are stored under the **document name** while the modal passed the slugified navigation **`id`** to **`Operation`**.
> 
> **`Modal.vue`** now passes **`x-scalar-navigation.name`** as **`documentSlug`** so reads and writes use the same key. A regression test covers a document named **`My API`** where navigation **`id`** is **`my-api`** but **`documentSlug`** must stay **`My API`**. Patch notes are in the changeset for **`@scalar/api-client`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit eb36d7e4b47da92760b3ff90f46cc8b043382788. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/api-client/src/v2/features/modal/Modal.vue
+++ b/packages/api-client/src/v2/features/modal/Modal.vue
@@ -180,7 +180,7 @@ defineExpose({
         :activeWorkspace="activeWorkspace"
         class="flex-1"
         :document="document.value"
-        :documentSlug="document.value['x-scalar-navigation']?.id ?? ''"
+        :documentSlug="document.value['x-scalar-navigation']?.name ?? ''"
         :environment
         :eventBus
         :exampleName="exampleName?.value"
--- a/.changeset/fix-runner-auth-token-sources.md
+++ b/.changeset/fix-runner-auth-token-sources.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
+
+Fix the request runner dropping a manually typed auth token when the document name is not slug-safe (e.g. documents loaded via `sources`). The modal now reads auth secrets under the same document key they are written to.
--- a/packages/api-client/src/v2/features/modal/Modal.test.ts
+++ b/packages/api-client/src/v2/features/modal/Modal.test.ts
@@ -371,6 +371,64 @@ describe('Modal', () => {
     expect(operation.props('layout')).toBe('modal')
   })
 
+  it('passes the document name (not the slugified navigation id) as documentSlug', async () => {
+    // The auth store keys secrets by the document name, but the navigation `id` is a
+    // slugified version of it. When the name is not slug-safe (common for `sources`
+    // documents with an explicit slug), passing the `id` here would make the runner
+    // read auth secrets under a different key than they were written, dropping the
+    // typed token. See https://github.com/scalar/scalar/issues/9554.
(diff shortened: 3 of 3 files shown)


packages/api-client/src/v2/features/modal/Modal.vue
```diff
@@ -180,7 +180,7 @@ defineExpose({
         :activeWorkspace="activeWorkspace"
         class="flex-1"
         :document="document.value"
-        :documentSlug="document.value['x-scalar-navigation']?.id ?? ''"
+        :documentSlug="document.value['x-scalar-navigation']?.name ?? ''"
```
hanspagel, line 183: lol, one line fix

amritk (approved):
one liners are the best liners

marclave (approved):

PR #9559, opened 2026-06-17
fix(api-reference): render factored-out properties alongside compositions

### LINKED ISSUE #8593: Factoring Schema's not rendering correctly
### What happens?

When using a json schema which is factoring out the common parts of the subschemas. 
Then rendering is not showing any of the properties in the API reference.

Related to:
- https://json-schema.org/understanding-json-schema/reference/combining#factoringschemas

The models and their properties to be rendered correctly.

### OpenAPI Document

https://sandbox.scalar.com/e/7cUWO

hanspagel:
Schemas that factor common `properties` out to the top level next to an `anyOf`/`oneOf`/`allOf`/`not` showed the composition but dropped the properties. Now the object's own properties render too.

## Ticket

Fixes #8593

---

> [!NOTE]
> **Low Risk**
> Display-only change in schema rendering with targeted tests; no auth, data, or API behavior changes.
> 
> **Overview**
> Fixes schema docs where **top-level `properties` sit next to composition keywords** (`anyOf`, `oneOf`, `not`, or nested cases like `allOf` + `anyOf`). The UI used to show the composition but **hide those shared properties** because `isTypeObject` treats composed schemas as non-objects.
> 
> **`SchemaProperty.vue`** now renders a separate object-properties block when `properties` / `additionalProperties` exist, except when **`allOf` is present** (merged rendering already includes sibling properties, so a second block would duplicate them). For other compositions it still shows factored-out fields unless the schema is an **explicit non-object** type.
> 
> Nested property rendering uses **`objectSchemaForChildren`** with composition keywords stripped so compositions stay in their own section and the nested `Schema` does not recurse.
> 
> Adds **`factoring schemas`** tests in `Schema.test.ts` (issue #8593) and a patch changeset for `@scalar/api-reference`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3a13dab9607564719551919702d49687588238db. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/factoring-schema-properties.md
+++ b/.changeset/factoring-schema-properties.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
+Render an object's own properties when they are factored out alongside a composition keyword (`anyOf`/`oneOf`/`allOf`/`not`)
--- a/packages/api-reference/src/components/Content/Schema/Schema.test.ts
+++ b/packages/api-reference/src/components/Content/Schema/Schema.test.ts
@@ -1055,5 +1055,25 @@ describe('Schema', () => {
       expect(text).toContain('one')
       expect(text).toContain('another')
     })
+
+    it('does not duplicate properties factored out alongside allOf', () => {
+      // `allOf` already merges the factored-out sibling `properties` into its
+      // rendered result, so they must not also render in a separate object
+      // block. Two members keep the `allOf` intact (a single member is flattened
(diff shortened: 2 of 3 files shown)


amritk (approved):
Do I fully understand? no but it loooks cool

marclave (approved):

PR #9565, opened 2026-06-19
fix(components,themes): match textarea placeholder styling to inputs

hwkr:
Textarea placeholders weren't picking up the themed placeholder styling that inputs get, so they looked a little out of place. This wires them into the same reset and lets the placeholder font inherit, so textareas and inputs now match.

- **`@scalar/themes`** — extend the `::placeholder` color/font reset to textareas
- **`@scalar/components`** — let `ScalarTextArea`'s placeholder inherit the font

See snapshots for before/after


amritk (approved):

hanspagel:
nice

PR #9567, opened 2026-06-20
feat(components): render Mermaid diagrams in markdown

pingmukul:
## Problem

API descriptions often need to convey *behavior* — request flows, state
transitions, and especially event sequences — which prose handles poorly.
Mermaid is the de-facto way developers express this in markdown (GitHub,
GitLab, and most docs tooling render fenced `mermaid` blocks), so people
already put Mermaid diagrams in OpenAPI descriptions expecting them to render.

Today Scalar shows them as a plain code block. This is especially relevant
now that Scalar renders AsyncAPI — event-driven APIs are all about sequence
and choreography, which sequence/flow diagrams capture much better than text.

## Solution

Render fenced `mermaid` code blocks (in `info.description`, tag, operation,
and other descriptions — anywhere `ScalarMarkdown` is used) as diagrams.

- **Pipeline** (`@scalar/code-highlight`): a new `rehypeMermaid` plugin marks
  `mermaid` blocks for client rendering, keeping the source inside
  `<code class="no-highlight">` so its newlines survive `rehype-format`
  (a bare `<pre>` gets its whitespace collapsed, which breaks Mermaid parsing).
- **Rendering** (`@scalar/components`): `ScalarMarkdown` finds those
  placeholders after render, **lazy-loads** Mermaid (only when a diagram is
  present, so it stays out of the bundle otherwise), renders to SVG, and
  re-renders on light/dark color-mode change.
- **Security**: Mermaid runs with `securityLevel: 'strict'`, so untrusted
  diagram sources cannot inject HTML or click handlers.

Mermaid is heavy, hence the lazy dynamic `import()` and the lockfile growth.

## Visual

`mermaid` code blocks in an OpenAPI `info.description` now render as diagrams
(flowchart + sequence), adapting to the active light/dark color mode.

**Light mode**

light mode

**Dark mode**

dark mode

## Checklist

- [x] I added a changeset.
- [x] I added tests.

### DIFF
--- a/.changeset/mermaid-diagrams.md
+++ b/.changeset/mermaid-diagrams.md
@@ -1,0 +1,8 @@
+---
+'@scalar/code-highlight': minor
+'@scalar/components': minor
+---
+
+feat: render Mermaid diagrams embedded in markdown
+
+Mermaid fenced code blocks (` ```mermaid `) in OpenAPI descriptions (and anywhere `ScalarMarkdown` is used) are now rendered as diagrams instead of plain code blocks. Mermaid is loaded lazily, so it is only downloaded when a document actually contains a diagram, and diagrams re-render to match the active light/dark color mode.
--- a/documentation/guides/docs/components/markdown-support.mdx
+++ b/documentation/guides/docs/components/markdown-support.mdx
@@ -123,6 +123,32 @@ function example() {
 ```
 ````
 
+## Diagrams
+
+Render [Mermaid](https://mermaid.js.org/) diagrams with a fenced code block using the `mermaid` language. Diagrams follow the active light or dark color mode.
+
+<Preview>
+
+```mermaid
+sequenceDiagram
+  Client->>API: GET /planets
+  API->>Database: SELECT * FROM planets
+  Database-->>API: rows
+  API-->>Client: 200 OK
+```
+
+</Preview>
+
+````md
+```mermaid
+sequenceDiagram
+  Client->>API: GET /planets
+  API->>Database: SELECT * FROM planets
+  Database-->>API: rows
+  API-->>Client: 200 OK
+```
+````
+
 ## Blockquotes
 
 Quote text with `>`, spanning multiple lines as needed.
--- a/documentation/markdown.md
+++ b/documentation/markdown.md
@@ -11,6 +11,7 @@ What's working here, is probably also working in the API reference:
 - tables
 - images
 - alerts
+- diagrams (with Mermaid)
 - …
 
 ## Alerts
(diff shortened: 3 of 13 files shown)


hanspagel (changes requested):
ahh, thanks for the PR. probably not gonna happen, because it increases the bundle size too much
let me check

bielu:
@hanspagel if Mermaid  are too heavy, might be worth looking into draw.io renderings? the bundle of it is around 70kb based on someones react component https://www.npmjs.com/package/react-drawio?activeTab=code

hanspagel:
yeah, this PR adds +3,157 kB to the bundle. that's insane, we can't ship this. :))

(BTW we support mermaid on scalar.com, where we render the docs of our users on the server-side)

pingmukul:
@hanspagel 

Totally fair, thanks for checking. +3,157 kB in the bundle is a blocker, no argument there.

A bit of context on why I pushed for this. We're actually doing the same server-side workaround you mentioned, converting mermaid to SVG on the server and embedding it into our API docs. It works, but it's becoming a lot of glue code to maintain and doesn't feel like the right long-term approach, which is what led me here.

Why it matters to us right now: our APIs are increasingly consumed by AI agents, and mermaid has become one of the most agent- and human-friendly ways to express the less-obvious behaviors an API can exhibit (state transitions, sequencing, edge cases). Having that notation render directly in the docs is genuinely valuable for that.

Since the blocker is bundle size and not the feature itself, is there an approach you'd be open to? For example, making mermaid an optional peerDependency / lazy chunk so it never touches the core bundle and is only pulled in when a consumer opts into diagrams, or, if you'd rather keep this server-side, any pointers on the direction you'd accept. Happy to rework the PR along whatever lines you'd recommend.

Poliklot:
Jumping in with a possible smaller compromise, @hanspagel.

The bundle-size concern makes sense. I would not ship Mermaid directly inside `@scalar/components` either.

Maybe the core change could be much smaller: do not render Mermaid in Scalar by default at all. Instead, expose a small opt-in hook/extension point for fenced code blocks. By default, ` ```mermaid ` would stay a normal code block, so the core bundle and standalone build would stay unchanged.

Consumers who need diagrams could then plug in their own renderer, using their own `mermaid` dependency, a CDN/global `window.mermaid`, or maybe a separate optional package later if that ever makes sense.

So the feature would be less “Scalar supports Mermaid out of the box” and more “Scalar lets users render Mermaid blocks themselves without forcing everyone else to pay +3 MB”.

This feels like a pretty common pattern in docs/markdown tooling:

- Docusaurus Mermaid support is an extra [`@docusaurus/theme-mermaid`](https://docusaurus.io/docs/next/markdown-features/diagrams) + `markdown.mermaid: true`, not always-on core behavior.
- VitePress keeps heavier markdown features opt-in too, for example math via `markdown-it-mathjax3`, and exposes [`markdown.config`](https://vitepress.dev/guide/markdown#advanced-configuration) for custom markdown-it plugins.
- `react-markdown` exposes [`components`](https://github.com/remarkjs/react-markdown) so consumers can replace rendering for elements/code blocks themselves.

Very rough shape, not saying this should be the exact API:

```ts
markdownRenderers: {
  mermaid: async (code, element) => {
    // consumer renders this however they want
  },
}
```

If this direction sounds acceptable, I could try a smaller PR just for the extension point + docs example, instead of shipping Mermaid rendering itself.

PR #9569, opened 2026-06-21
feat(api-reference): add server and protocol labels to AsyncAPI chann…

bielu:
## Problem

An AsyncAPI document can expose the same-looking channels, operations, and messages on different servers and protocols (e.g. the same event published over `wss` in production and `ws` in development, or a channel that also carries a `kafka` binding). When we render channels and their messages, none of that context was visible — readers couldn't tell which server(s) a channel lived on or which protocol(s) a given message was carried over without digging into the raw spec. That makes near-identical entries impossible to tell apart.

## Solution

Surface server and protocol labels inline:

- **Channels** show, in the header, the **servers** they're available on and those servers' **protocols** — in both the `modern` and `classic` layouts.
- **Messages** show **protocol** labels in their accordion header.

Implementation:

- New reusable `AsyncApiLabels.vue` renders a wrapping row of pills (neutral pills for server names, uppercase pills for protocols) with `sr-only` "Servers:" / "Protocols:" prefixes, and renders nothing when empty.
- New helper `getChannelServerLabels(document, channel)` resolves servers from `document.servers`, restricted to `channel.servers` when declared, and returns deduplicated `{ servers, protocols }` (declaration order preserved, all protocols — not just WebSocket).
- `Channel.vue` renders server + protocol labels; `Message.vue` renders protocol labels.

A message is carried over whatever protocols its **channel's servers** speak — `message.bindings` only declares per-protocol config and usually covers a single protocol (or none). So a message's protocol labels are the **union of its channel's server protocols and its own binding protocols**, deduplicated; otherwise a multi-protocol message would surface only the one protocol it happens to declare a binding for.

## Checklist

- [x] I added a changeset. 
- [x] I added tests.

### DIFF
--- a/.changeset/asyncapi-server-protocol-labels.md
+++ b/.changeset/asyncapi-server-protocol-labels.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': minor
+---
+
+Show server and protocol labels when rendering AsyncAPI channels and messages: each channel header lists the servers it's available on and their protocols, and each message surfaces every protocol it's carried over (its channel's server protocols unioned with its own binding protocols)
--- a/documentation/asyncapi.md
+++ b/documentation/asyncapi.md
@@ -15,9 +15,10 @@ Scalar.createApiReference('#app', {
 The reference renders the AsyncAPI document grouped by channel. For each channel you'll see:
 
 - The channel title (or address) and description.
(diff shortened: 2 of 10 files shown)


hanspagel (approved):
this is great, thank you :))

bielu:
No worries, I use scalar everyday, so do adjustments to make it easier to use, and now i am contributing in asyncapi to make it easier to use as well :)

PR #9571, opened 2026-06-22
feat(aspire): add `AddScalarMockServer` to run a mock server as an Aspi…

bielu:
## Problem

Sometimes I need to expose an API to another team before I've implemented it. With a contract-first workflow, the OpenAPI document exists well before the service does — the consuming team wants to start integrating immediately, but there's no running backend to point them at.

Scalar already ships a mock server (`@scalar/mock-server`, published as the `scalarapi/mock-server` container) that turns an OpenAPI document into a working, realistic API. But Aspire users had no first-class way to use it: to stand up that mock alongside the rest of their app, they had to hand-roll a container resource and wire up env vars and endpoints themselves.

The API Reference already has a clean `AddScalarApiReference` Aspire resource, so it felt natural for the mock server to have the same one-liner story: declare it from an OpenAPI document and let the consuming team (and other resources) depend on it like any real service — before a single handler is written.

## Solution

Added `AddScalarMockServer`, a thin Aspire hosting wrapper around the published `scalarapi/mock-server` image — cloned from the existing `AddScalarApiReference` resource pattern. Because the mock server already ships as a container, this needs no engine port and keeps full fidelity (`x-handler`/`x-seed`, validation, auth, Faker-based responses all run in the canonical Node implementation).

- `ScalarMockServerResource` is a `ContainerResource` implementing `IResourceWithServiceDiscovery`, so other resources can `.WithReference(mock)` it as if it were the real service.
- Document sources (mutually exclusive): `WithDocument` (inline), `WithDocumentFile` (read at AppHost build time), `WithDocumentUrl` (fetch on startup), and `WithDocumentFrom` (derive from another resource's endpoint, resolved via service discovery).
- Wires the image/tag, container port (3000), a `/openapi.json` health check, and removes OTLP exporter annotations to keep the container tunnel bootstrap non-blocking — mirroring the API Reference resource.

Alternatives considered: a native in-process ASP.NET port of the mock engine (`MapScalarMockServer`) and a sidecar-and-reverse-proxy approach. Both were set aside for now — the container model is the idiomatic Aspire composition unit and the lowest-risk, highest-fidelity path. The native/in-process story can follow later.

## Checklist

- [x] I added a changeset. 
- [x] I added tests.

### DIFF
--- a/.changeset/aspire-mock-server.md
+++ b/.changeset/aspire-mock-server.md
@@ -1,0 +1,7 @@
+---
+'@scalar/aspire': minor
+---
+
+feat(dotnet): add `AddScalarMockServer` to run a Scalar Mock Server as an Aspire resource
+
+Adds `AddScalarMockServer`, which registers the `scalarapi/mock-server` container as an Aspire resource that generates realistic mock API responses from an OpenAPI/Swagger document — so other resources can depend on it as if it were the real service. The document can be supplied inline (`WithDocument`), from a local file (`WithDocumentFile`), from a URL (`WithDocumentUrl`), or derived from another resource's endpoint (`WithDocumentFrom`).
--- a/documentation/integrations/aspire.md
+++ b/documentation/integrations/aspire.md
@@ -388,6 +388,76 @@ var scalar = builder.AddScalarApiReference(options =>
 });
 ```
 
+## Mock Server
+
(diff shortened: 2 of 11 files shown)


hanspagel (approved):
niiiice

PR #9572, opened 2026-06-22
feat(mock-server): mock AsyncAPI 3.1 documents over WebSocket and SSE

bielu:
# feat(mock-server): mock AsyncAPI 3.1 documents over WebSocket and SSE

## Problem

`@scalar/mock-server` could only mock **request/response (REST)** APIs from OpenAPI/Swagger documents — it walks `paths` → operations → methods and registers HTTP routes. It had no way to mock **event-driven** APIs.

Scalar already supports AsyncAPI elsewhere (the API reference renders AsyncAPI channels, the API client connects to channels over WebSocket), but there was no server to develop and test those event-driven clients against. Frontend and integration work that depends on a WebSocket/SSE backend had nothing to point at until the real service existed.

We want the same "point it at a document and get a working mock" experience for AsyncAPI that we already have for OpenAPI.

## Solution

Add `createAsyncApiMockServer`, the event-driven counterpart of `createMockServer`. It resolves an AsyncAPI 3.1 document into normalized channels and serves each one through a **transport**:

- **WebSocket** (`ws`/`wss`): `receive` operations push a generated message when a client connects; `send` operations echo a generated reply to each inbound frame. Built on `@hono/node-ws`.
- **SSE** (`sse`, or `http`/`https` server-push channels): a `GET` on the channel route opens an event stream. Pure Hono `streamSSE`, no extra runtime dependency.

Messages are generated from each message's `payload` schema using the **same `getExampleFromSchema` generator the REST mocker uses**, preferring defined `examples` — so mock data is consistent across REST and AsyncAPI.

Because WebSocket support needs the Node `http.Server` returned by `serve()`, the function returns `{ app, injectWebSocket }`; the caller attaches WebSocket handling after `serve()`.

### Extensibility

Transports are pluggable via the `transports` option. A transport declares the channels it owns via `supports()` and wires them up via `register()`, which receives both the Hono `app` (for HTTP handshakes) and Hono's `upgradeWebSocket` helper (for socket upgrades). This lets protocols that negotiate over HTTP before upgrading — e.g. **SignalR** — be added without changing the core. Brokered protocols (Kafka/MQTT/AMQP) are intentionally out of scope since they need external broker infrastructure.

### Docker

The Docker mock server now auto-detects AsyncAPI documents (top-level `asyncapi` field) and starts the AsyncAPI engine, injecting WebSocket support. The volume scan recognizes AsyncAPI files too.

### Alternatives considered

- **Overloading `createMockServer` with auto-detection** — rejected: WebSocket needs the `http.Server` handle that `createMockServer` does not own, and auto-detection muddied the return type. A separate entry point keeps types clean.
- **Mocking brokered protocols (Kafka/MQTT/AMQP)** — out of scope; requires running real brokers, not feasible for an in-process Node mock. The transport extension point leaves the door open later.
- **AsyncAPI 2.x support** — deferred; 3.1-only for now, with an upgrader path to follow (mirroring the OpenAPI upgrader).

### Scope notes

- The **Aspire** integration (`AddAsyncApiDocument`) lives on a separate branch and is a follow-up PR; it will mirror the existing `WithApiReference(...)` document-attach pattern.
- Inbound message validation is a planned follow-up (matches the REST mocker's `validateRequest`).

## Checklist

- [x] I added a changeset.
- [x] I added tests.

### DIFF
--- a/.changeset/mock-server-asyncapi.md
+++ b/.changeset/mock-server-asyncapi.md
@@ -1,0 +1,5 @@
+---
+'@scalar/mock-server': minor
+---
+
+Add `createAsyncApiMockServer` to mock event-driven APIs from an AsyncAPI 3.1 document. Channels are served over WebSocket and SSE, with messages generated from each message's payload schema (the same generator the REST mocker uses). Additional protocols (e.g. SignalR) can be added through the `transports` extension point. The Docker mock server now auto-detects AsyncAPI documents.
--- a/documentation/guides/mock-server/asyncapi.md
+++ b/documentation/guides/mock-server/asyncapi.md
@@ -81,6 +81,8 @@ For each message the mock server, in order of preference:
 
 Channel addresses with path parameters (for example `rooms/{roomId}`) become route parameters (`/rooms/:roomId`).
 
+> The SSE transport emits one message per `receive` operation and then closes the stream, rather than streaming continuously. Reconnect (or let your SSE client reconnect) to get another batch.
+
 ## Document detection
 
 The Docker mock server and CLI automatically detect AsyncAPI documents (by their top-level `asyncapi` field) and start the AsyncAPI mock instead of the REST mock — no extra flag is required.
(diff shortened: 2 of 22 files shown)


hanspagel:
craazy

bielu:
Yeah @hanspagel there will be few more, I want to include signalr as well,  but need figure out how to make it best, so it is separated enough so it is not included in main package, but easy enough we can use in .net integration :)

hanspagel (approved):
niceee, I just refactored it to use our workspace-store more

Issue #9492, opened 2026-06-11
Array parameter is not expanded correctly


cyrilchapon:
### What happens?

Hi.

Array parameters are not expanded correctly, see video :
https://supercut.ai/share/corma/W2uV3Howh9M83JcAA12OyM

I have a parameter defined like so :
```
// Displayed by Scalar UI (straight from OpenAI definition)
filters[applicationInstanceId][in] Type:array integer[]
```

The trouble is it's wrongly intepreted as
```
filters[applicationInstanceId][in]
```

instead of

```
filters[applicationInstanceId][in][]
```

Please also note on the video that editing the "name" of an auto-suggested parameter doesn't work.

------

There are 2 troubles here to be clear :

1. The default array (+ `explode` ?) configuration to be without trailing `[]` (I know this is all about conventions, but this is so common...). And I couldn't find anything to configure that on scalar.
2. The fact that editing the left part (name) of an "auto" (suggested) parameter doesn't actually affect the underlying query

The combination of those 2 troubles make the page unusable for our customers... there is no way to make the filter work, appart just copy/pasting the parameter on a random one.

### What did you expect to happen?

I'd expect :

1. The default array (+ `explode` ?) configuration to be with trailing `[]`. Or it to be configurable somewhere.
2. Editing the left part (name) of an "auto" (suggested) parameter actually changes to query exactly as described visually

Related : https://github.com/scalar/scalar/issues/4975

### OpenAPI Document

https://sandbox.scalar.com/e/BGTen

hanspagel:
I’ve submitted #9497 to make the renaming work
For the brackets to appear, you just need to set `allowReserved: true` for the parameter I think

cyrilchapon:
Actually, my spec already had `allowReserved`

```
          {
            "name": "filters",
            "in": "query",
            "schema": {
              "type": "object",
              "properties": {
                "userId": {
                  "type": "object",
                  "properties": {
                    "in": {
                      "anyOf": [
                        {
                          "type": "string"
                        },
                        {
                          "type": "array",
                          "items": {
                            "type": "string"
                          }
                        }
                      ]
                    },
                    "notIn": {
                      "anyOf": [
                        {
                          "type": "string"
                        },
                        {
                          "type": "array",
                          "items": {
                            "type": "string"
                          }
                        }
                      ]
                    }
                  }
                },
                "applicationInstanceId": {
                  "type": "object",
                  "properties": {
                    "in": {
                      "anyOf": [
                        {
[...]

PR #9583, opened 2026-06-23
fix(workspace-store): serialize `deepObject` array params with trailing brackets

hanspagel:
Array values in a `deepObject` query param were collapsed into one comma-joined value (`filters[id][in]=1,2`). Now each item is its own entry, using the common trailing-bracket convention:

```
filters[id][in][]=1&filters[id][in][]=2
```

This is the wire-format half. Table display + edit round-trip is #9584.

## Ticket

Part of #9492

### DIFF
--- a/.changeset/deep-object-array-brackets.md
+++ b/.changeset/deep-object-array-brackets.md
@@ -1,0 +1,5 @@
+---
+'@scalar/workspace-store': patch
+---
+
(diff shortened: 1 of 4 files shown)


amritk (approved):
We should ensure we follow the document for this kind of stuff, as all servers dont accept all bracket syntaxes

PR #9587, opened 2026-06-23
fix(api-reference): prevent stack overflow on `allOf` self-references

### LINKED ISSUE #9570: Model rendering crashes with `RangeError: Maximum call stack size exceeded` when a self-reference passes through `allOf`
### Describe the bug

In the **Models** section (and any response/request schema view), a schema that references itself **through an `allOf` composition** crashes the schema renderer with:

```
Oops, something went wrong here.
RangeError: Maximum call stack size exceeded
```

A plain recursive `$ref` (a property pointing directly back to its own schema) renders fine — the renderer clearly has a cycle guard. But as soon as the self-reference sits inside an `allOf` branch, the guard no longer fires and the renderer recurses until the stack overflows.

Interestingly, the **example/sample generator handles the cycle correctly** (it emits `"self": null`); only the schema/model renderer blows up. So the cycle handling appears to be inconsistent between the two renderers.

### Minimal reproduction

A single self-contained schema is enough — no external `$ref`, no large spec:

```json
{
  "openapi": "3.0.0",
  "info": { "title": "Node", "version": "1" },
  "paths": {
    "/x": {
      "get": {
        "operationId": "getX",
        "responses": {
          "200": {
            "description": "ok",
            "content": {
              "application/json": {
                "schema": { "$ref": "#/components/schemas/Node" }
              }
            }
          }
        }
      }
    }
  },
  "components": {
    "schemas": {
      "Node": {
        "allOf": [
          { "type": "object", "properties": { "name": { "type": "string" } } },
          { "type": "object", "properties": { "self": { "$ref": "#/components/schemas/Node" } } }
        ]
      }
    }
  }
}
```

Standalone HTML used to reproduce:

```html
<!doctype html><html><head><meta charset="utf-8"></head><body>
<script id="api-reference" type="application/json"> ...the spec above... </script>
</body></html>
```

### Steps to reproduce

1. Load the spec above.
2. Open **Models → `Node`** (or open the `GET /x` `200` response schema).
3. The model body shows `Oops, something went wrong here. / RangeError: Maximum call stack size exceeded`.

### What I narrowed down

| Variant | Self-reference shape | Result |
|---|---|---|
| Plain `$ref` self-reference (`Node.properties.self → Node`) | no `allOf` | ✅ renders |
| Self-reference inside an `allOf` branch's property | `allOf: [..., { properties: { self: { $ref: Node } } }]` | ❌ crash |
| Self-reference wrapped in `allOf` + `nullable` | `self: { allOf: [{ $ref: Node }], nullable: true }` | ❌ crash |

So the trigger is specifically **a self/circular `$ref` reached via `allOf` merging**. This is common output from generators that model TypeScript inheritance / `Omit`/`Pick` utility types as `allOf` (e.g. tsoa), where a recursive type (a chat message that quotes another message, a recursive condition tree, etc.) ends up self-referencing through `allOf`.

### Expected behavior

The model should render to a bounded depth (showing the recursive `self` property as a collapsed/leaf reference), the same way a plain recursive `$ref` already does — and the same way the example generator already handles it (`"self": null`).

### Environment

- `@scalar/api-reference` **1.60.0** (latest), loaded as the standalone bundle from `https://cdn.jsdelivr.net/npm/@scalar/api-reference`
- OpenAPI 3.0.0 (also reproduces with 3.1.0)
- Chrome (CDP)

### Related (but distinct)

- #5213 — circular reference crash (closed)
- #6660 — `Maximum call stack` on a large, deeply-`$ref`'d spec without obvious cycles (closed, fix in #6673). The case here is different: a **single self-contained schema** whose self-reference is reached **through `allOf`**, reproducing on the latest `1.60.0`.
- #2736 — max call stack with circular refs in request body

hanspagel:
A schema whose self-reference is reached through an `allOf` branch crashed the model renderer with `RangeError: Maximum call stack size exceeded`. A plain recursive `$ref` already rendered fine — only the `allOf` path blew up.

The cause is in `mergeAllOfSchemas`: `mergePropertiesIntoResult` guards its *existing property* branch against self-referential `$ref`s, but not its *new property* branch. The repro's `self` property only appears once, so it always took the unguarded branch and recursed forever.

The fix applies the same `seenRefs` guard to that branch and keeps the raw `$ref` when a cycle is detected, so the renderer's existing cycle detection bounds it. Verified the crash reproduces in a unit test (no rendering) and is gone after the fix; existing merge tests still pass.

## Ticket

Fixes #9570

---

> [!NOTE]
> **Low Risk**
> Localized fix in schema merge logic with a targeted regression test; behavior for non-cyclic schemas should match prior merge paths.
> 
> **Overview**
> Fixes a **Maximum call stack size exceeded** crash when the API reference merges and renders schemas whose self-`$ref` appears only on a **new** property inside an `allOf` branch (typical of TypeScript inheritance → OpenAPI).
> 
> `mergePropertiesIntoResult` already stopped cycles when merging an **existing** property or array items; the **first-time** property path could still call `mergeAllOfSchemas` on the resolved `$ref-value` without checking `seenRefs`, so a node like `self: { $ref: Node }` on an `allOf` member recursed forever.
> 
> The change adds the same guard: if the new property’s `$ref` is already in `seenRefs`, keep the raw property; otherwise pass an updated `seenRefs` into nested `mergeAllOfSchemas`. A unit test reproduces issue #9570; the changeset records a patch for `@scalar/api-reference`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 47306f8906bb8e93e544c734d3eebffa6a91c71e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/witty-pugs-grin.md
+++ b/.changeset/witty-pugs-grin.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
+Fix `Maximum call stack size exceeded` crash when rendering a schema whose self-reference is reached through an `allOf` branch
--- a/packages/api-reference/src/components/Content/Schema/helpers/merge-all-of-schemas.test.ts
+++ b/packages/api-reference/src/components/Content/Schema/helpers/merge-all-of-schemas.test.ts
@@ -1236,4 +1236,32 @@ describe('mergeAllOfSchemas', () => {
 
     expect(() => mergeAllOfSchemas({ allOf: [node, node] } as any)).not.toThrow()
   })
+
+  it('does not infinitely recurse when a self-reference is reached through allOf', () => {
+    // Reproduces https://github.com/scalar/scalar/issues/9570: a schema whose
+    // self-reference sits inside an `allOf` branch (common output from tools
+    // that model TypeScript inheritance as `allOf`).
+    //
+    //   Node:
+    //     allOf:
+    //       - { type: object, properties: { name: { type: string } } }
+    //       - { type: object, properties: { self: { $ref: '#/components/schemas/Node' } } }
+    //
+    // `self` only ever appears as a brand-new property, so it took the
+    // unguarded new-property branch and recursed forever.
+    const node: any = {
+      allOf: [
+        { type: 'object', properties: { name: { type: 'string' } } },
+        {
+          type: 'object',
+          properties: {
+            self: { $ref: '#/components/schemas/Node', '$ref-value': null },
+          },
+        },
+      ],
+    }
+    node.allOf[1].properties.self['$ref-value'] = node
+
+    expect(() => mergeAllOfSchemas(node)).not.toThrow()
+  })
 })
--- a/packages/api-reference/src/components/Content/Schema/helpers/merge-all-of-schemas.ts
+++ b/packages/api-reference/src/components/Content/Schema/helpers/merge-all-of-schemas.ts
@@ -230,9 +230,23 @@ const mergePropertiesIntoResult = (
     }
 
     if (!result[key]) {
+      // Break self-referential cycles reached through `allOf`. When a brand-new
+      // property's `$ref` is already being merged higher in the call stack,
+      // keep the raw reference instead of recursing into its `allOf` forever.
+      // This mirrors the guard in the "merge existing property" branch below and
+      // covers schemas that point back at themselves through an `allOf` member
+      // (e.g. a node whose `self` property $refs the node, which is an allOf).
+      const rawProperty = properties[key]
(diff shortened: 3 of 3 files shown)


bgrcs (approved):

amelnikoff:
In my case fix didn't work again well.

Now It shows another error:

> Models
> CommentDto
> Oops, something went wrong here.
> InternalError: too much recursion
 

https://sandbox.scalar.com/p/z1i5l#GET/comments

Issue #7400, opened 2025-11-20
Security schemes not disabled despite explicitly defining per-endpoint security attribute


jemgunay:
### What happens?

I have two auth schemes defined under `components > securitySchemes`. Some endpoints support neither, some endpoints support both and some support just one of the two. However, both auth schemes are always selectable for every single endpoint which is misleading/confusing despite explicitly setting the per-endpoint `security` attribute.

### What did you expect to happen?

The [OpenAPI spec](https://swagger.io/docs/specification/v3_0/authentication/#describing-security) states that defining `security: []` on an endpoint indcates "No security" is supported, yet both auth schemes are always selectable. I would expect them to be omitted from the auth selection drop down as it is confusing which scheme is supported per-endpoint.

My global `security` is set to `[]`, and each endpoint references only the security scheme(s) that it supports, or defines `security: []` on the endpoint if no auth scheme is supported for that endpoint as the specification recommends. However, both schemes are always selectable in the scalar docs UI.

A specific example is the `Login` endpoint in the linked doc, where both schemes are selectable despite both schemes being unsupported. Another is the `Search Malware Samples` endpoint where one is rendered in the auth drop down as `Required` and another as `Available` - however the "Available" scheme is not supported at all for this endpoint, so I'd expect it to be omitted/disabled from the list - otherwise users may select it expecting it to work, causing confusion.

<img width="291" height="137" alt="Image" src="https://github.com/user-attachments/assets/08959243-11f9-4234-aaab-82583f97cd29" />
<img width="291" height="165" alt="Image" src="https://github.com/user-attachments/assets/c5c00af9-8c55-4849-a7b9-d9eed0ac792d" />

### OpenAPI Document

https://sandbox.scalar.com/e/UucEM

ikeohachidi:
I'd like to take this. Just to clarify from the attached [docs](https://sandbox.scalar.com/e/UucEM) the `Login` endpoint should actually not have any available authentication and the `Search Malware Samples` Should have the BearerJWT in available authentication instead of required authentication?

jemgunay:
Hey, thanks for looking into it. So I was expecting:

- `Login` would have no available/required auth schemes to select from (as `security: []` is set for this endpoint).
[...]

PR #9593, opened 2026-06-23
fix(api-client): present multiple security requirements as alternatives

hanspagel:
Follow-up to #9592. When an operation lists more than one security requirement, the auth dropdown labeled them all "Required authentication" — implying you need every one. They are actually a choice.

The OpenAPI spec:

> When the `security` field is defined on the OpenAPI Object or Operation Object and contains multiple Security Requirement Objects, only one of the entries in the list needs to be satisfied to authorize the request.

So multiple requirements are OR-alternatives. They now show as available choices instead of "Required". A single requirement — including a combined `{ A: [], B: [] }` AND requirement — stays marked as required. This matches the read-only auth badge, which already says "one of:" for alternatives.

## Ticket

Part of #7400

---

> [!NOTE]
> **Low Risk**
> UI labeling logic in the auth selector only; no request signing or credential handling changes, with behavior covered by updated unit tests.
> 
> **Overview**
> Fixes the auth dropdown so it matches OpenAPI semantics when an operation declares **more than one** security requirement object.
> 
> **`getSecuritySchemeOptions`** now treats multiple top-level requirements as **OR alternatives**: the "Required authentication" group is empty and those schemes appear under **Available authentication** instead of all being labeled required. A **single** requirement—including one object with multiple schemes (AND, e.g. `apiKey & httpBasic`)—still shows as required.
> 
> Tests and a patch changeset for `@scalar/api-client` document the behavior (including scope variants as alternatives).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 61303d1c386b43b0d41b5c230fbe4d6a72718baa. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/great-laws-attack.md
+++ b/.changeset/great-laws-attack.md
@@ -1,0 +1,7 @@
+---
+'@scalar/api-client': patch
+---
+
+fix: label multiple security requirements as alternatives instead of required
+
+When an operation lists more than one security requirement, those are OR-alternatives — only one needs to be satisfied. The auth dropdown no longer labels them all as "Required authentication"; it presents them as available choices. A single requirement (including a combined AND requirement) stays marked as required.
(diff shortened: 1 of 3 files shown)


hanspagel:
not sure i like this

Issue #9589, opened 2026-06-23
OAuth2 scope deselection does not update counter or token request


i1zolotariov:
### What happens?

With `Scalar.AspNetCore` `2.16.5`, OAuth2 scope deselection does not reliably update Scalar's selected scope state.
Setup:
- ASP.NET Core OpenAPI
- OAuth2 client credentials flow
- `AddPreferredSecuritySchemes("oauth2")`
- `EnablePersistentAuthentication()`
- `flow.SelectedScopes = []`
Minimal Scalar setup:
```csharp
app.MapOpenApi();
app.MapScalarApiReference(options => options
    .AddPreferredSecuritySchemes("oauth2")
    .AddDocument("v1", "API", isDefault: true)
    .AddClientCredentialsFlow("oauth2", flow =>
    {
        flow.ClientId = "example-client";
        flow.SelectedScopes = [];
    })
    .EnablePersistentAuthentication());
```

Observed behavior:

Selecting a scope can update the checkbox but not reliably update the Scopes Selected X / Y counter.
Deselecting a scope may not decrement the Scopes Selected X / Y counter.
Deselect All may not reset the selected scope state.
A deselected scope may still be sent in the /connect/token request body.
The checkbox state, counter, and outgoing token request can become inconsistent.
This appears related to #8983 / #8984, but still reproduces with Scalar.AspNetCore 2.16.5.
Persistent authentication is enabled in our setup. However it does not appear to be the root cause of the deselection/counter mismatch.

Current workaround:

We are temporarily tracking scope checkbox changes outside Scalar and rewriting the outgoing token request body so the scope parameter matches the user's latest checkbox selection. The same workaround also manually keeps the Scopes Selected X / Y label aligned with the tracked state.

### What did you expect to happen?

The OAuth2 scope checkbox state, `Scopes Selected X / Y` counter, and `/connect/token` request body should always stay in sync.
Expected behavior:
- Selecting one scope increments the counter.
- Deselecting one scope decrements the counter.
- `Deselect All` resets the counter to `0 / Y`.
- Deselected scopes are not included in the token request body.
- The `scope` parameter sent to `/connect/token` exactly matches the scopes currently selected in the UI.

kemsky:
Scalar.AspNetCore 2.16.20 - Select All does nothing. It was working in version 2.14.3.

hanspagel:
True! New PR: https://github.com/scalar/scalar/pull/9938

PR #9595, opened 2026-06-23
fix(api-client): prevent oauth2 scope checkbox from toggling twice

hanspagel:
Clicking directly on an OAuth2 scope checkbox toggled the scope twice:
* once from the checkbox and 
* once from the surrounding row's click handler
 
which made the click land on the wrong state. 

The checkbox now stops the click from bubbling to the row.

## Ticket

Part of #9589

---

> [!NOTE]
> **Low Risk**
> Small UI event-handling change in OAuth scope selection with no auth, data, or API behavior changes.
> 
> **Overview**
> Fixes OAuth2 scope selection in the auth selector so **clicking the row checkbox** updates selection once instead of appearing to do nothing or flip the wrong way.
> 
> `DataTableRow` already toggles a scope on row click while `DataTableCheckbox` updates via `@update:modelValue`. A direct checkbox click used to **bubble to the row** and run both handlers, toggling twice. **`@click.stop`** on the checkbox stops that propagation; a short comment documents why.
> 
> Includes a **patch changeset** for `@scalar/api-client`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a082dc3484789d932db84c879353a15d54a40ea6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #9594, opened 2026-06-23
fix(api-client): keep oauth2 scope selection in sync on quick clicks

hanspagel:
Selecting/deselecting OAuth2 scopes quickly could drop changes: the "Scopes Selected X / Y" counter, the checkboxes, and the scopes sent to the token endpoint would disagree. Clicking two scopes in quick succession could silently lose one.

The component was emitting a freshly computed scope list built from a prop that had not updated yet, so a second click overwrote the first. Now it emits a single-scope toggle and the store applies it against the currently stored scopes, so successive clicks compose correctly.

Tested at the store level (`updateSelectedScopes`), including the rapid-toggle case and the `preferredSecurityScheme` fallback.

## Ticket

Part of #9589

---

> [!NOTE]
> **Medium Risk**
> Changes how OAuth2 scope selection is persisted and affects token authorization scope lists, but behavior is narrowly scoped with new store tests and safer handling of invalid payloads.
> 
> **Overview**
> Fixes a race where **rapid OAuth2 scope checkbox clicks** could drop selections: the UI counter, checkboxes, and scopes sent to the token endpoint could disagree (#9589).
> 
> **`OAuthScopesInput`** no longer builds and emits a full `scopes` array from the current prop on each click. Single-scope changes now emit **`{ scope, selected }`**; **Select All / Deselect All** still send an absolute **`scopes`** list.
> 
> **`updateSelectedScopes`** in the workspace store applies single-scope toggles against **scopes already in auth state** (not a component snapshot), so successive clicks compose. Payloads with neither `scopes` nor a valid `scope` + `selected` pair are **ignored** so malformed events cannot wipe the selection.
> 
> Component and store tests cover toggles, rapid successive updates, and the preferred-scheme fallback.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dc51aad38093fd09d1aa558188db0e4c41c96f17. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #9596, opened 2026-06-23
fix(api-client): keep DataTableCheckbox in sync with its model value

hanspagel:
A native checkbox flips itself the instant you click it. If the parent then leaves `modelValue` unchanged, the `:checked` binding has nothing to patch and the box stays visually wrong. This re-asserts `modelValue` after each change so the checkbox always reflects the value it represents.

Defensive hardening — `DataTableCheckbox` is a shared base component, so this affects every checkbox in the client. It is a no-op wherever the model already tracks the value.

## Ticket

Part of #9589

---

> [!NOTE]
> **Low Risk**
> Localized UI fix in a shared checkbox component with tests; no auth, data, or API behavior changes.
> 
> **Overview**
> Fixes **data table checkboxes** that could **flicker or look wrong** after a click when the parent updates `modelValue` slowly (e.g. scope toggles on the async workspace event bus).
> 
> `DataTableCheckbox` now **watches `modelValue`** and writes it to the native input’s `checked` state when the value settles, so the UI stays aligned with the source of truth without snapping back while an update is in flight. Change handling is routed through a small `onChange` helper with an input `ref` instead of an inline template emit.
> 
> Adds **unit tests** for initial state, emit behavior, prop updates, and the async “optimistic toggle” case.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8cd0096da2071ba09f1a823057624737ea4690b0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/fix-datatable-checkbox-controlled.md
+++ b/.changeset/fix-datatable-checkbox-controlled.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
+
+Keep `DataTableCheckbox` visually in sync with its value. When a click does not end up changing `modelValue`, the checkbox now reverts to the model state instead of showing the browser's optimistic toggle.
--- a/.changeset/fix-datatable-checkbox-controlled.md
+++ b/.changeset/fix-datatable-checkbox-controlled.md
@@ -2,4 +2,4 @@
 '@scalar/api-client': patch
 ---
 
-Keep `DataTableCheckbox` visually in sync with its value. When a click does not end up changing `modelValue`, the checkbox now reverts to the model state instead of showing the browser's optimistic toggle.
+Keep `DataTableCheckbox` visually in sync with its value. The checkbox now mirrors `modelValue` onto the native input whenever the value settles, including when the parent updates it asynchronously, so a successful toggle no longer flickers back before the update lands.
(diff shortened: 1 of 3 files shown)


hanspagel:
naahh, I don't think that's necessary

marclave (approved):

hanspagel:
Addressed the flicker Bugbot flagged: the `nextTick` reset re-asserted `modelValue` before async parent updates landed (scope toggles go through the async `withHook` bus in `workspace-events.ts`), briefly reverting a successful toggle.

Replaced it with a `watch` on `modelValue` that mirrors the value onto the native input once it settles — sync or async — so the optimistic toggle stays put until the store confirms it, and no fixed-tick revert races the update. Added a `DataTableCheckbox.test.ts` covering the in-flight async case.

hanspagel:
Closing as unnecessary. Digging into the two Bugbot reviews together, the plain `:checked="modelValue"` binding already re-patches the input whenever `modelValue` changes (sync or async), so it covers every real case here — both consumers (`OAuthScopesInput.setScope`, `RequestTableRow.handleUpdateRow`) always emit an accepted update, and the quick-click/double-toggle bugs were fixed at the source in #9594 and #9595.

The only scenario needing extra imperative code is a *rejected* toggle (parent leaves `modelValue` unchanged), which does not occur. Every attempted fix was either a flicker (the `nextTick` reset, Bugbot Medium) or a no-op over the existing binding (the `watch`, Bugbot Low), so there is nothing correct left to add.

PR #9309, opened 2026-05-21
feat(api-reference): configure 'Models' section label

DemonHa:
Fixes: #9238

Currently we have the hardcoded "models" text to display the openapi schemas... This PR makes that behavior configurable

### What stays the same (backward compatibility)?
- Config keys: hideModels, expandAllModelSections, generateModelSlug
- Navigation type: models, IDs like default/models, hash #models (cc @amritk  is this what we want??)
- Internal file/component names

------

image

## Checklist

- [x] I explained why the change is needed.
- [x] I added a changeset. 
- [x] I added tests.
- [ ] I updated the documentation.

---

> [!NOTE]
> **Medium Risk**
> Changes public deep-link shapes (`model/` → `models/` or custom slugs) and URL rewrite behavior on load; mitigated by redirects and tests but still affects bookmarks and integrations that hardcode old paths.
> 
> **Overview**
> Adds **`modelsSectionLabel`** (`'Models' | 'Schemas' | string`, default **`Models`**) so the `components.schemas` section can use OpenAPI-style **Schemas** (or any custom label) in the **sidebar**, **main content**, **search**, and **developer toolbar** copy.
> 
> Navigation and deep links now derive the section slug from that label (e.g. `#…/schemas/User` when set to **Schemas**). Per-schema IDs move from the legacy singular **`…/model/…`** segment to **`…/models/…`** (or the slugified custom label). **`redirectLegacyModelUrl`** rewrites old **`/model/<name>`** hash/path URLs on load so bookmarks keep working, with guards so operation paths like **`POST …/model/train`** are not rewritten.
> 
> Wiring runs through **`@scalar/types`**, **`@scalar/schemas`**, **`workspace-store`** navigation traversal, and **`api-reference`** UI/search; docs add **`modelsSectionLabel`**; Playwright and unit tests cover labels, redirects, and snapshots.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ce177faa888fffd2b173b3dc2281286e340cd681. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

Issue #9238, opened 2026-05-16
Rename "Models" to "Schemas"


marclave:
### Discussed in https://github.com/scalar/scalar/discussions/9226

<div type='discussions-op-text'>

<sup>Originally posted by **piljac1** May 15, 2026</sup>
Is there any config that allows us to rename "Models" to "Schemas" to match OpenAPI/Swagger's naming? Of course, we can CSS hack the displayed label if we really want to, but not the `data-sidebar-id` and other attributes that would refer to models.</div>

hanspagel:
As someone living in the OpenAPI-world, I’d love to rename them for everyone. But you'd probably prefer to make schemas an opt-in feature?

marclave:
maybe theres a better way to do general translations?

DemonHa:
Yeah naming them models is confusing coming from an openapi world

PR #9597, opened 2026-06-23
feat(api-reference): add i18n and RTL support

Poliklot:
**Problem**

The API Reference still has a lot of UI copy hard-coded in English. That makes framework integrations difficult to use for teams that need localized navigation, search, schema labels, action buttons, and mobile/reference chrome. RTL users hit an extra issue: forcing `dir="rtl"` with CSS flips the surrounding layout, but search and keyboard hints still need direction-aware behavior.

This came up in the API Reference i18n/RTL discussions linked below, and it also builds on the recent `Models`/`Schemas` label configurability work where a broader translation layer was suggested.

**Solution**

This PR adds API Reference UI i18n support without changing how OpenAPI documents are loaded or routed:

- Add an `i18n` configuration object with `locale`, `direction`, and `translations` overrides.
- Add built-in UI translations for English, Russian, Spanish, French, German, Simplified Chinese, and Arabic.
- Localize API Reference chrome across search, navigation, sidebar, schema labels, operation sections, response examples, developer tools, and action buttons.
- Apply RTL direction for Arabic by default and allow explicit direction overrides.
- Keep search usable in RTL: the placeholder follows the input direction, while keyboard shortcuts stay LTR and do not mirror.
- Cover configuration parsing and runtime translation behavior with tests.

Out of scope: this does not translate OpenAPI document content such as summaries, descriptions, tags, or examples. Users should still provide localized OpenAPI documents/sources for that content.

## Visual

Arabic RTL API Reference overview:

Arabic RTL API Reference overview

Arabic RTL search modal with LTR keyboard shortcut:

Arabic RTL search modal with LTR keyboard shortcut

## Checks

- `pnpm biome check --write --diagnostic-level=error --no-errors-on-unmatched --files-ignore-unknown=true <changed files>`
- `pnpm prettier --write <changed files>`
- `pnpm install`
- `pnpm build:packages`
- `pnpm --filter @scalar/api-reference test`
- `pnpm --filter @scalar/components test`
- `pnpm --filter @scalar/sidebar test`
- `pnpm --filter @scalar/types test`
- `pnpm --filter @scalar/schemas test`
- `pnpm --filter @scalar/api-reference --filter @scalar/components --filter @scalar/sidebar --filter @scalar/types --filter @scalar/schemas types:check`
- `pnpm changeset status`
- `pnpm build:integrations`
- `pnpm knip`

## Related

- Addresses the API Reference UI-label side of https://github.com/scalar/scalar/discussions/6577
- Addresses the API Reference UI-label side of https://github.com/scalar/scalar/discussions/6487
- Addresses the API Reference UI-label and RTL side of https://github.com/scalar/scalar/discussions/5613
- Addresses the API Reference RTL/UI-label side of https://github.com/scalar/scalar/discussions/7383
- Addresses the native RTL layout request in https://github.com/scalar/scalar/discussions/9375
- Related demand: https://github.com/scalar/scalar/discussions/2458
- Builds on #9309 and the discussion in #9238 about moving from one-off label configuration toward general translations.
- Related but intentionally out of scope: https://github.com/scalar/scalar/discussions/6052 and https://github.com/scalar/scalar/discussions/7034, which focus on translating OpenAPI document content rather than API Reference UI copy.


Poliklot:
I noticed I had missed a few remaining UI strings, so I moved this back to draft while I fixed them. The follow-up is pushed now.

packages/components/src/components/ScalarSearchInput/ScalarSearchInput.vue
```diff
@@ -57,8 +59,8 @@ const { stylingAttrsCx, otherAttrs } = useBindCx()
       autocapitalize="off"
       autocomplete="off"
       autocorrect="off"
-      class="flex-1 appearance-none rounded border-none bg-transparent outline-none"
-      placeholder="Search..."
+      class="flex-1 appearance-none rounded border-none bg-transparent text-start outline-none"
```
hwkr: We could probably make text-start the default in our reset right? Instead of text-left?

hwkr: Or maybe the user agent stylesheets are the thing including the text-left

packages/schemas/src/api-reference/api-reference-configuration.test.ts
hwkr, line 1: Not an issue in your PR but it looks like we have a typo in this filename @hanspagel  😅

hwkr (approved):
The components changes look good to me! I'll leave it to @hanspagel and others to review the rest. Thank you for the PR!

hwkr (commented):
Also it would be great to add a snapshot test set with one of the languages to capture the rendering

Poliklot:
Thanks for the review, @hwkr!

I moved the logical text alignment into the shared theme reset and removed the local `text-start` override from the search input. I also added Arabic snapshots for the API Reference covering the main reference view, the search modal, and filtered search results.

The snapshots were regenerated and verified with the Docker Playwright runner.

hanspagel (changes requested):
nice

* looks like some snapshots still need an update
* I think I’d rename a few of the things
* we have to put the actual translations in separate files, so we don't increase the bundle size too much

I’m taking this one over here

Poliklot:
Thanks @hanspagel!

I’m happy to keep working on this PR. Since you mentioned you’re taking this one over, should I wait for your follow-up before changing the architecture/names, or would you prefer that I start by updating the affected API Reference snapshots and splitting the non-English locale packs out of the main bundle?

I can also adjust the names once you point out the ones you’d like changed.

hanspagel:
Okay, they gzip well. Those 7 translations only add ~13-15 KB gzipped (~1.4% of bundle). I think that's fine, even if we add another 20.

@Poliklot Let me do the updates. ✌️

hanspagel (approved):
I think this is good to go :))

PR #9602, opened 2026-06-25
feat(api-reference): show document type postfix in document selector

bielu:
## Problem

In multi-document workspaces, the document selector only showed each document's title. When a workspace mixes OpenAPI and AsyncAPI documents (which render very differently), there was no way to tell from the selector which kind of document each entry is — you had to open it to find out.

## Solution

Append the document type (`OpenAPI` / `AsyncAPI`) as a postfix to each option label in the document selector, e.g. `Scalar Galaxy (OpenAPI)`, `Scalar Galaxy Events (AsyncAPI)`.

Documents are loaded lazily (only the active one is fetched up front), so the type is resolved through a fallback chain:

1. The parsed document in the workspace store (most accurate — covers loaded URL sources).
2. The inline `source.content` from the configuration (available immediately for inline documents).
3. A best-effort guess from the source URL for documents not yet loaded — substring match on `asyncapi` / `openapi` / `swagger`, so paths like `/openapi/v1.json` or `/asyncapi/v1.json` are recognized without a network request.

The postfix is baked into the option `label` rather than passed as a separate field, because the shared `ScalarListboxItem` renders only `option.label` (no slot for a suffix) — this keeps it visible in both the selected-button text and the dropdown list with no changes to the shared listbox components.

**Alternatives considered:** eagerly fetching every URL document on load to detect its type. Rejected — it would defeat lazy loading and add N network requests for a cosmetic label. URL documents with no type hint in the path simply stay un-postfixed until they're opened.

## Checklist

- [x] I added a changeset. 
- [ ] I added tests.

### DIFF
--- a/.changeset/document-selector-type-postfix.md
+++ b/.changeset/document-selector-type-postfix.md
@@ -2,4 +2,4 @@
 '@scalar/api-reference': patch
 ---
 
-Append the document type (`OpenAPI`/`AsyncAPI`) as a postfix in the document selector only when multiple documents include a mix of OpenAPI and AsyncAPI documents.
+Show the document type (`OpenAPI`/`AsyncAPI`) as a right-aligned badge in the document selector, but only when multiple documents include a mix of OpenAPI and AsyncAPI documents.
--- a/.changeset/document-selector-type-postfix.md
+++ b/.changeset/document-selector-type-postfix.md
@@ -2,4 +2,4 @@
 '@scalar/api-reference': patch
 ---
 
-Append the document type (`OpenAPI`/`AsyncAPI`) as a postfix to each option in the document selector
+Append the document type (`OpenAPI`/`AsyncAPI`) as a postfix in the document selector only when multiple documents include a mix of OpenAPI and AsyncAPI documents.
--- a/.changeset/document-selector-type-postfix.md
+++ b/.changeset/document-selector-type-postfix.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
+Append the document type (`OpenAPI`/`AsyncAPI`) as a postfix to each option in the document selector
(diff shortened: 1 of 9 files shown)


hanspagel (commented):
I was thinking about this and came to the conclusion, that

a) 99% of our users have OpenAPI documents only and they don't care to see it's all OpenAPI
b) even if you have two documents, it's not that the enduser cares which standard is used to describe an API, do they?

Wdyt?

bielu:
@hanspagel when working with microservices i have bunch of documents (open/async api)
image
and i think it is actually make difference :)

bielu:
just had though, maybe we should show only if there is +2 documents with different types?

bielu:
@hanspagel added conditional logic now, will need do little more testing but as soon I am happy will tag as ready for review again, let me know if you want any other changes here :)

bielu:
works fine in my testing, switching to ready for review, let me know if you hit any issue!

bielu:
@hanspagel now it is small change any chance to include it with this release 🙏

hanspagel:
mmhh .. guessing it from the URL is smart, but I don't like that it's only appearing when I load the document.
and I think it should be more visual, and not just text.

sorry, this one needs some more work, too. 😬  I need to think about a good solution.

bielu:
@hanspagel yeah it is one of reason for https://github.com/scalar/scalar/pull/9662#issuecomment-4891011810 to exist, as now until we load document we do not detect type for local files 😓

hanspagel:
ik ik
maybe we want to do this: #9676 
also I tweaked the design a bit

bielu:
I think both things make sense, but your proposed change makes it quicker to see benefit, so yeah cool with it.

bielu:
@hanspagel any blocker here?

hanspagel:
I don't like the design TBH, I think I might close it actually. I messaged our design team, though, maybe they can come up with something.

hanspagel:
Thanks for the contribution! We’ve decided against this feature.

PR #9611, opened 2026-06-26
fix(validation): memoize recursive union scores across cyclic rings

hanspagel:
Recursive union alternatives repeatedly scored the same nested objects, causing exponential work even for compact chains and cyclic rings. Memoize scores within each coercion call while preserving the existing weights, selected branches, values, aliases, and circular output references.

Cycle-dependent scores record the active and inactive ancestor pairs needed for safe reuse. Acyclic scores use a direct cache; each cyclic pair keeps one contextual result. Evaluate callbacks must return stable results without mutating inputs or schemas, as documented on the public builder and in the README.

Dependency masks are capped at 1,024 bits per call. When another cycle-dependent pair needs a bit, stop reading and writing cached scores for the remainder of the call, including pending recursive frames, and fall back to the original active-pair cycle guard. Many independent cycles therefore cannot cause unbounded mask widths.

This cap bounds dependency-mask width, not total runtime, memory, or recursion depth. Recursive unions and large rings encountered after the cap can still incur the original exponential rescoring; fresh evaluate allocations still add bookkeeping. This is a bounded memoization optimization, not a complete denial-of-service defense.

### DIFF
--- a/packages/validation/src/coerce-dos.test.ts
+++ b/packages/validation/src/coerce-dos.test.ts
@@ -1,0 +1,57 @@
+import { describe, expect, it } from 'vitest'
+
+import { coerce } from '@/coerce'
+import { type Schema, lazy, literal, object, optional, union } from '@/schema'
+
+/**
+ * Security regression: `coerce` must stay roughly linear in the nesting depth of
+ * its input, even for recursive `union` schemas.
+ *
+ * `coerce` picks a union branch with `scoreUnion`, which scores *every* branch
+ * and takes the max. The recursion guard in `coerce.ts` is a cycle guard only
+ * (it marks `(value, schema)` pairs currently on the call stack and clears them
+ * in `finally`); it does not memoize completed sub-scores. So for a recursive
+ * schema whose child is reachable through more than one branch, every nesting
+ * level of the input is re-scored once per branch — O(2^depth) CPU for a
+ * linear-size value.
+ *
+ * This is reachable from untrusted input: `@scalar/workspace-store` calls
+ * `coerce(openapiSchema, document)` on user-supplied OpenAPI/AsyncAPI documents,
+ * and the OpenAPI Schema Object is exactly this shape (`items` is recursive and
+ * appears in two union branches). A sub-1KB document hangs the event loop for
+ * seconds-to-minutes.
+ *
+ * The schema below distills that structure: a tagged node whose `next` child is
+ * declared in both union branches. The fix is to memoize completed, cycle-free
+ * `(value, schema)` scores in `scoreUnion`.
+ */
+const node: Schema = lazy(() =>
+  union([object({ kind: literal('a'), next: optional(node) }), object({ kind: literal('b'), next: optional(node) })]),
+)
+
+/** Builds a linear-size chain of `depth` nested nodes — a few hundred bytes of JSON. */
+const buildChain = (depth: number): Record<string, unknown> => {
+  let value: Record<string, unknown> = { kind: 'a' }
+  for (let i = 0; i < depth; i++) {
+    value = { kind: 'a', next: value }
+  }
+  return value
+}
+
+describe('coerce', () => {
+  it('coerces a recursive union in roughly linear time (no exponential branch scoring)', () => {
+    // At depth 19 the input is ~600 bytes. Without memoization, scoring is
+    // ~2^19 work and takes several seconds; with memoization it is a few ms.
+    const value = buildChain(19)
+
+    const start = performance.now()
+    const result = coerce(node, value)
+    const elapsed = performance.now() - start
+
+    // Coercion still works: the value matches a branch and is returned intact.
+    expect(result).toMatchObject({ kind: 'a' })
+
+    // The actual regression: this must not blow up exponentially.
+    expect(elapsed).toBeLessThan(500)
+  }, 30_000)
+})


amritk (commented):
Good fix, and the split into `scoreUnion` (cycle/memo bookkeeping) and `scoreUnionInner` (the unchanged scoring heuristics) makes it easy to see that the weights themselves are untouched. Threading one `ScoringContext` through the whole `coerce` call so completed scores are shared across successive union selections is the right shape, and the `readsAtDepth` test measuring property-getter reads at depths 8/16/32 is a much better regression guard than a wall-clock assertion.

**One gap I think is worth closing before this merges: `encounteredCycle` is traversal-global, so a chain that ends in a cycle falls back to the exponential behaviour this PR fixes.**

```ts
if (trackable && !scoring.encounteredCycle) { /* memoize */ }
…
scoring.depth--
if (scoring.depth === 0) { scoring.encounteredCycle = false }
```

Once any cycle is hit, the flag stays set until the top-level `scoreUnion` call unwinds, and memoization is suppressed for *everything* on the way back up — including subtrees that had no cycle at all. Scores are only written on the unwind, so a cycle discovered during the descent poisons the entire call.

For the two cases in the tests that is fine: a purely linear chain never sets the flag (each level is a distinct value object, so `active` never hits), and a small self-referential value terminates quickly via the cycle guard. But the composition of the two does not:

> a chain of N nodes whose last `next` points back at the first

The cycle guard fires at depth N, `encounteredCycle` goes true, and none of the N levels above it get memoized — so scoring is back to ~2^N. Given this branch is named `claude/coerce-dos-poc` and the framing is denial of service, that is the shape an attacker-supplied document would take, and it is only a few characters different from the PoC that motivated the PR.

Suppressing memoization *conservatively* is the right instinct — a score computed while a cycle is open genuinely depends on which ancestors are active, and the "does not reuse scores computed while an enclosing cycle is open" test shows what goes wrong otherwise. The fix is to scope the suppression rather than widen it: track whether *this* subtree's score depended on an active ancestor (for example, have `scoreUnionInner` propagate a cycle-dependency flag upward, and only skip memoizing frames on that path) rather than a single flag for the whole traversal. Then a chain-then-cycle memoizes the acyclic prefix and only the frames inside the cycle stay uncached.

If you would rather not take that on here, the next best thing is to add the chain-then-cycle case as a test with the current behaviour documented, so the limitation is recorded rather than discovered later.

Two smaller notes:

- `expect(elapsed).toBeLessThan(500)` is a wall-clock assertion on shared CI runners. The 500 ms bound against a several-second failure is generous and the 30 s timeout helps, but this is the test most likely to flake under load. The `readsAtDepth` test already proves the same property deterministically — consider whether the timing one earns its place.
- `completed` is keyed on `(value, schema)` and persists for the whole `coerce` call. That is safe as long as nothing mutates a value mid-coercion, but `evaluate` schemas run `schema.expression(value)` during scoring. Probably fine in practice; worth a comment noting the assumption, since `coerce` also accepts caller-supplied `cache`/`lazyCache` and someone may later be tempted to hoist the scoring context out too.

Not blocking on the small notes — the chain-then-cycle case is the one I would want addressed.

---

amritk (commented):
The exponential rescoring is a real problem and the dependency-mask approach is a clever fix — recording which ancestor activations a cyclic score depended on, and only reusing it when those still hold, is the right shape for this. Not approving yet because `scoreUnion` decides which union branch every coerced value takes, so a subtle error here is a silent wrong-branch selection across the whole workspace store rather than a crash.

**The tests are about the attack, not the invariant.** The only new file is `coerce-dos.test.ts` — it proves the blow-up is gone, which is the headline, but nothing new pins "memoized scoring picks the same branch as unmemoized scoring". The property that matters is exactly that. Could you add a differential test: generate a spread of cyclic/recursive schema-and-value pairs, score each with the memoized path and with a deliberately non-memoizing reference implementation, and assert the selected branches match? That converts "preserving existing branch selection" from a claim in the changeset into something CI enforces.

**The new implicit contract on `evaluate`.** The changeset says "Evaluate callbacks must return stable results without mutating inputs while scoring." That is a new requirement on every schema author, enforced nowhere and invisible at the call site — previously an unstable `expression` was merely wasteful, now it silently poisons `pair.completed`. Two things would help: state it in the doc comment on `evaluate` itself where authors will see it, and confirm the existing `evaluate` users in the repo actually satisfy it. Related: `schema.expression(value)` returns a fresh object per call for most implementations, so each call mints a new `pairs` entry — is there a case where that grows the WeakMap materially within one `coerce`?

**Reuse across the whole call.** `pair.completed` is cached for the lifetime of one `coerce`, keyed on object identity. That is sound as long as nothing mutates the input mid-coercion — worth a line in the comment saying so explicitly, since `coerceInner` builds new objects rather than mutating and that is what makes it safe.

**Minor:** `nextBit <<= 1n` grows unbounded across a long coercion. Fine with BigInt, but a document with many distinct cyclic pairs ends up doing wide BigInt masking on every cache probe. Have you measured the memoized path against the old one on an *acyclic* large document, to confirm the bookkeeping does not cost more than it saves in the common case?

---

cameronrohani (approved):

amritk (commented):
Not approving, and I want to be specific about why, because the measurements in the description argue against the change in one important case.

**The fix introduces a worse regression than the problem it solves, for a shape an attacker controls.** From the PR's own table:

| Fixture | Time | Max RSS |
| --- | --- | --- |
| 2,048 independent self-cycles | 9.4 → **38.9 ms** | 57 → **143 MiB** |

That is 4× slower and 2.5× more memory. The description explains exactly why: "The BigInt width grows with distinct cycle-dependent pairs throughout a call, which explains why many independent cycles can regress" — 6,144 bits of mask width for 1,024 independent self-cycles.

This branch is named `claude/coerce-dos-poc`, so the premise is hardening `coerce` against adversarial input. But an attacker who can supply a document with many *independent* self-cycles is cheaper to serve today than after this change. Trading exponential blowup on deeply *shared* recursion for quadratic-ish mask growth on *many* cycles is a real improvement only if the first shape is the one that actually reaches you. For a tool that ingests arbitrary user-supplied OpenAPI documents, both shapes are equally reachable, so the mitigation has a hole in the same threat model that motivates it.

I do not think that sinks the PR — the Stripe number (385 → 263 ms) is a genuine win on realistic input — but it needs one of:

- a cap on mask width, falling back to the unmemoized path beyond it, so the bad case degrades to today's behavior rather than worse than it;
- or an explicit statement that many-independent-cycles is out of scope, with the threat model written down.

Right now the README documents the limits, which is honest, but documenting a regression is not the same as bounding it. "Memoization has no hard resource or recursion-depth budget" is the sentence I would most want to stop being true before this lands in a DoS-hardening PR.

**What is genuinely impressive here:** committing a reference scorer frozen from `main` and diffing 4,800 graph/schema cases against it — comparing selected branches, values, cycles, aliases *and* input non-mutation — is a much stronger equivalence argument than a test suite, and exactly right for a change that must preserve scoring behavior exactly. Auditing all four production `evaluate` callbacks for the new stability contract, and writing that contract into the JSDoc and README, is the other half of the job done properly.

`mergeable_state` currently reads `unknown`, and this has been open since June — worth a rebase before it is picked up either way.

---

packages/validation/test/non-memoized-coerce.ts
DemonHa: maybe you want to delete this one, looks like leftover test file

hanspagel:
closing in favor of #10315

PR #9615, opened 2026-06-27
feat(api-reference): render AsyncAPI bindings and previously-hidden fields

bielu:
## Problem

Scalar's AsyncAPI schema layer (`packages/schemas/src/asyncapi/3.1`) parses almost the entire 3.1 spec, but the docs renderer only displayed a small subset of it. A lot of meaningful, fully-resolved information never reached the reader, which made the AsyncAPI reference noticeably thinner than the OpenAPI one:

- **Protocol bindings were invisible.** Only the protocol *name* was shown as a pill. The actual binding details — Kafka topic, AMQP exchange, MQTT QoS, the WebSocket handshake, etc. — were parsed and then dropped on the floor. For broker-centric APIs (especially Kafka) this is where a lot of the important configuration lives.
- **Several fields were parsed but never rendered:** operation `reply` (request/reply flows), operation `security`, `tags` + `externalDocs` on channels/operations/messages, and message `examples` + `correlationId`.

## Solution

Additive rendering on top of the objects already resolved via `resolveAsyncApiChannel/Operation/Message` — no schema changes.

- **Bindings:** a generic `AsyncApiBindings.vue` renders one labeled group per protocol as a key/value grid, so all 20 protocols display at once. Primitive values render inline; object/array values (e.g. Kafka `groupId`, WS handshake schemas) render as JSON via `ScalarCodeBlock`. Backed by a small, tested `format-async-api-bindings.ts` helper. Wired into Channel, Operation, and Message.
- **Reply:** `AsyncApiOperationReply.vue` shows the reply channel address, reply address `location`/`description`, and reply message names.
- **Security:** `get-async-api-required-security.ts` adapts AsyncAPI's flat OR-list of schemes into a `RequiredSecurity`, reusing the existing `SecurityRequirementBadge` rather than building a parallel UI.
- **Tags + externalDocs:** `AsyncApiTags.vue` renders tags as `Badge` pills and reuses the existing `ExternalDocs.vue`.
- **Examples + correlationId:** rendered in `Message.vue` (examples with name/summary + payload/headers JSON; correlationId location + description).
- **Types & i18n:** re-exported three already-defined AsyncAPI types from the public barrel, and added 8 translation keys to the `ApiReferenceTranslations` type and all 7 locale files.
- **Docs:** updated `documentation/asyncapi.md` — expanded "What renders" to cover the new fields and added a "Protocol bindings" section.

Considered (and deferred as out of scope): typed per-protocol binding cards, multi-format (Avro/Protobuf) payload rendering, a server variables/security section, message trait merging, and an interactive message client. The generic key/value binding view was chosen so every protocol ships now instead of a curated few.

## Checklist

- [x] I added a changeset. 
- [x] I added tests. 


bielu:
@hanspagel i think this might be worth review, it is not perfect way of displaying, but yeah i think we should for now focus on displaying everything and later on clean up, what you think?

hanspagel:
function: awesome!
design: that's a little bit wild

let me check whether we want to get this in as-is and take it from there, or iterate on the design in the PR already

going back to draft for now

bielu:
Cool, as mentioned already i am not really design guy 😅, but yeah he didn't look great when i was testing it :D

bielu:
@hanspagel did you had chance to catch up with team, if maybe we can get some designs? I am okay with making changes, however yeah Design is not my strong side ;P

bielu:
Okay had some spare time on my hand, so lets compare options:
- **Current ui:**
    -  Channel: image
    - Operation+message:imageimage
- **Option 1**:
    -  Channel: image
    -  Operation+message:imageimage
- **Option 2**:
    - Channel: image
    -  Operation+message:
imageimage
- **Option 3**:
    - Channel:image
    -  Operation+message::imageimage
    - 
Personally i think option 3 looks best

bielu:
hey @hanspagel  sorry for being pain in the a... but I would like to help on that when i have time to do work :), so if you could come back to me which of options above make sense would be great!

hanspagel:
we're busy bringing asyncapi support to our platform, once that's landed, I’ll start working on asyncapi support here more again 👍

bielu:
@hanspagel all good! Ping me when you you will have time to look into that, then if I have time I am more than happy to support again on it :)

hanspagel:
Thanks for working on this! I’ll work through this in smaller PRs.

PR #9616, opened 2026-06-27
fix(workspace-store): upgrade AsyncAPI documents on ingestion

bielu:
## Problem

AsyncAPI documents below 3.x render as a blank page. A user reported an AsyncAPI **2.4.0** (MQTT) spec that showed no channels and no operations — only the Models section.

The cause is that AsyncAPI ingestion in `workspace-store` deliberately skips the upgrade pipeline (to avoid the OpenAPI `coerce` step injecting `openapi: ''` and breaking the type discriminator), but never runs an AsyncAPI-specific upgrade in its place. So a 1.x/2.x document reaches the 3.x traversal unchanged.

In 3.x, operations live in a top-level `operations` map; in 2.x they're nested under each channel as `publish`/`subscribe`. With `operations` empty, every channel bucket ends up with zero operations, `createChannelEntry` drops the channel, and nothing renders.

`@scalar/asyncapi-upgrader` already does the full structural 1.x → 2.x → 3.x conversion — it just was never wired into the store.

## Solution

Wire `@scalar/asyncapi-upgrader` into the AsyncAPI ingestion branch in `client.ts`, mirroring how the OpenAPI path calls its own upgrader. The original document version is captured **before** the upgrade (which bumps `asyncapi` to the latest) and preserved on `x-original-aas-version`.

- `packages/workspace-store/src/client.ts` — run the AsyncAPI upgrader during ingestion; keep the pre-upgrade version on `x-original-aas-version`.
- `packages/workspace-store/package.json` + `pnpm-lock.yaml` — add the `@scalar/asyncapi-upgrader` workspace dependency.
- `packages/workspace-store/src/client.test.ts` — flip the old "does not upgrade" test, update two `3.0.0`→`3.1.0` expectations, and add a regression test for the reported 2.4.0 MQTT shape (asserts `subscribe` lifts to a top-level operation, `server.url`→`host`, and the channel appears in navigation).
- `documentation/asyncapi.md` — add a "Supported versions" section noting that 1.x/2.x documents are upgraded to 3.x automatically on load.
- `packages/asyncapi-upgrader/src/upgrade.ts` — correct the stale doc comment that claimed no structural transforms are applied.

Verified manually in the api-reference dev playground with the reported spec: channel, parameter, operation, and message now render correctly.

**Alternatives considered:** adding a 2.x adapter directly in the traversal layer — rejected because the dedicated upgrader already exists and the OpenAPI path establishes the "upgrade on ingestion" pattern this should match.

## Checklist

- [x] I added a changeset. 
- [x] I added tests.

---

### Follow-ups (not blocking)

- The upgrader keeps the 2.x parameter `schema` field on channel parameters (not a 3.x Parameter field). `adaptAsyncApiParameters` ignores it, so it's harmless — optional cleanup in the upgrader's 2.6→3.0 step.
- The doc comment in `asyncapi-upgrader/src/upgrade.ts` ("only bumps the version string… no structural transformations") is stale; the 2.6→3.0 step does the full structural transform.

### Note

One pre-existing test failure in `client.test.ts` (`'resolves external references in file-loaded documents'`) is unrelated to this change — a Windows path-separator issue, confirmed failing identically on `main`.

### DIFF
--- a/.changeset/asyncapi-upgrade-on-ingestion.md
+++ b/.changeset/asyncapi-upgrade-on-ingestion.md
@@ -1,0 +1,5 @@
+---
+'@scalar/workspace-store': patch
+---
+
+Run the AsyncAPI upgrader when ingesting AsyncAPI documents so 1.x/2.x specs are converted to the 3.x shape the renderer expects. Previously a 2.x document (operations nested under channels as `publish`/`subscribe`) was passed through unchanged, leaving `operations` empty and dropping every channel from the navigation — the document rendered blank. The original version is preserved on `x-original-aas-version`.
--- a/documentation/asyncapi.md
+++ b/documentation/asyncapi.md
@@ -10,6 +10,12 @@ Scalar.createApiReference('#app', {
 })
 ```
 
+## Supported versions
+
+The reference renders against the AsyncAPI **3.x** shape. Older documents are upgraded automatically on load, so you can pass a **1.x** or **2.x** document and it renders the same way — there's nothing extra to configure.
+
+Behind the scenes the document is converted to the latest 3.x version (for example `subscribe`/`publish` operations nested under a channel are lifted into the top-level `operations` map, and a server `url` is split into `host` and `pathname`). The original version is preserved on the document as `x-original-aas-version` for reference.
+
 ## What renders
 
 The reference renders the AsyncAPI document grouped by channel. For each channel you'll see:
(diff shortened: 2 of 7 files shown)


hanspagel (approved):
ah nice! I actually thought I did this already 😅

bielu:
@hanspagel you did halve way :), the upgrader was done but not wired up lol

PR #9618, opened 2026-06-28
feat(api-reference): render document-wide auth for AsyncAPI

bielu:
## Problem

AsyncAPI documents already had their security schemes read and normalized at the store layer (`getAsyncApiSecurityRequirements`, the channel connection context), but none of it ever reached the rendered reference. The introduction's Authentication selector was fed OpenAPI-only schemes (`Content.vue` computed `securitySchemes` strictly from `openApiClientDocument.components.securitySchemes`), and `Auth.vue` gated everything on `isOpenApiDocument`. So an AsyncAPI document that defined `components.securitySchemes` showed **no authentication at all** — inconsistent with OpenAPI, where the same schemes render an interactive auth selector.

## Solution

Wire AsyncAPI into the existing document-level Auth selector instead of building a parallel renderer:

- **New helper `getAsyncApiDocumentSecurityRequirements`** — AsyncAPI has no root-level `security`, so document-wide requirements are derived from the deduped union of every server's `security` (a server applies to the whole connection), mapped to OpenAPI-style requirement objects. Operation/channel-level security is intentionally excluded here.
- **`Content.vue`** — the document-level `securitySchemes` computed now branches: for AsyncAPI docs it resolves `components.securitySchemes` (handling a `$ref` wrapper) and merges them through the existing `mergeSecurity`.
- **`Auth.vue`** — generalized beyond OpenAPI: computes requirements via the new helper for AsyncAPI docs and resolves the auth-store document name from `x-scalar-navigation.name` for both document types.
- **Decoupled the auth gate** — AsyncAPI auth no longer rides on `hideTestRequestButton` (an OpenAPI/test-client concept); it shows whenever the document is present.

Scheme types shared with OpenAPI (`http`, `oauth2`, `openIdConnect`, `apiKey`) get full input UI. Broker-specific AsyncAPI types (`userPassword`, `httpApiKey`, `X509`, `symmetricEncryption`, `asymmetricEncryption`, `plain`, `scramSha256`, `scramSha512`, `gssapi`) still appear in the selector but degrade to the existing fallback panel — `mergeSecurity` preserves the original `type`, so nothing crashes.

**Alternatives considered:** a bespoke read-only AsyncAPI auth component (rejected — duplicates UI and loses interactivity for the shared types) and a full AsyncAPI→OpenAPI scheme-type adapter for the broker types (deferred — those need dedicated input widgets and belong with the operation/channel-level auth follow-up).

## Checklist

- [x] I added a changeset. 
- [x] I added tests.


hanspagel:
depends on #9626, going back to draft + on hold

bielu:
that's okay, let me know if you need me to do additional work on it.

bielu:
@hanspagel found another interesting case, where Api key in async api != openapi api key, but http api key = openapi api keys, so had to adjust it, 
For reference 

image
so done adjustments

bielu:
oh last commit explains issue which i had, that just shows i still need learn a lot about internals of scalar 😓

hanspagel (approved):
finally! :)

PR #9620, opened 2026-06-28
feat(azure-functions): add Scalar.Azure.Functions integration

bielu:
## Problem

Scalar had no first-class way to render the API reference from an **Azure Functions** app. The existing `Scalar.AspNetCore` integration is built entirely around `IEndpointRouteBuilder`, `HttpContext`, minimal-API `Results`, and `IOptionsSnapshot` — none of which exist in the Azure Functions isolated worker, where routes come from `[Function]` + `[HttpTrigger]` and the built-in HTTP model uses `HttpRequestData`/`HttpResponseData` rather than `HttpContext`. Functions users had to hand-roll the HTML shell, embed/serve the bundle themselves, and reinvent ETag/gzip/nonce handling.

## Solution

Add a new **`Scalar.Azure.Functions`** package (isolated worker) and, to avoid duplicating the reference-rendering logic, first extract a hosting-agnostic core out of `Scalar.AspNetCore` so every integration shares one implementation.

**Hosting-agnostic core (in `Scalar.Shared`)**
- `ScalarHtmlBuilder` (the HTML shell + config serialization), `ScalarNonce`, and `ScalarStaticAssets` (embedded-resource resolution with content-based ETag + gzip/decompress) — all framework-free and reading embedded resources via `Assembly.GetManifestResourceStream`, so they work in any host.
- The HTML-shell options (`Title`, `BundleUrl`, `HeadContent`, nonce, …) moved into the shared `ScalarOptions` under `#if !SCALAR_ASPIRE`, and a `SCALAR_AZURE_FUNCTIONS` namespace branch was added across the shared sources.
- `Scalar.AspNetCore` was refactored onto this core — **behavior-preserving**; its existing test suite passes unchanged (39 Debug / 42 Release).

**The new package**
- One `IScalarApiReference` service (registered via `services.AddScalarApiReference(...)`) with a shared `ScalarRequestProcessor` that dispatches index / static asset / trailing-slash redirect / 304 / 404, plus two thin adapters so **both** HTTP models are supported:
  - ASP.NET Core integration: `HandleAsync(HttpContext)`
  - Built-in: `HandleAsync(HttpRequestData) → HttpResponseData`
- The consumer declares a small catch-all HTTP trigger (`Route = "scalar/{*path}"`, parameter named `path`) and forwards it to the service. A ready-made auto-discovered `[Function]` is intentionally deferred (the Functions SDK generator doesn't reliably discover `[Function]`s in a referenced assembly) and documented as a future improvement.
- The in-process model is not supported (end of support Nov 2026).

**Supporting changes**
- Migrated all .NET projects to **repo-wide Central Package Management** (`integrations/dotnet/Directory.Packages.props`), removing the per-tests `Directory.Packages.props` files and inline versions; one `VersionOverride` for the single genuine outlier.
- CI wiring: added the projects to `Scalar.slnx`, static-asset staging in `integration-builds.yml`, a NuGet `azure-functions-publish` job, and `azure_functions_package_changed` detection threaded through check-changes → `main.yml` → `release.yml` → `publish-integrations.yml`.
- Tests (processor unit tests + full-pipeline `HttpContext` integration tests), a runnable playground, docs, README, and changesets.

**Alternatives considered:** naming the package `Scalar.AspNetCore.Functions` (rejected — named by cloud platform so a future `Scalar.Aws.Lambda` reuses the same core); duplicating the rendering logic in the new package (rejected in favor of the shared core); shipping the zero-boilerplate auto-discovered function now (deferred pending reliable cross-assembly discovery).

## Checklist

- [x] I added a changeset. 
- [x] I added tests.


bielu:
@hanspagel  i think it is good one for next bigger release, as it fix biggest issue for scalar, that is not supported for azure functions yet! I am happy to create similar one for aws / google functions :)

hanspagel (approved):
nicee, let's get this in

PR #9622, opened 2026-06-29
ci: retry changeset version on transient GitHub API failures

hanspagel:
The release on `main` failed because `changeset version` hit a dropped connection to the GitHub GraphQL API (`ERR_STREAM_PREMATURE_CLOSE`) while generating changelog entries. Changesets escaped cleanly without touching any files, so nothing half-released — but the whole release job failed over one network blip.

This wraps `changeset version` in a small retry helper (3 attempts, 10s apart). The changelog generator is the only network-dependent step, so a transient blip now just retries instead of failing the release.

### DIFF
--- a/knip.jsonc
+++ b/knip.jsonc
@@ -24,6 +24,9 @@
     "packages/oas-utils/src/migrations/*/types.generated.ts"
   ],
   "ignoreDependencies": [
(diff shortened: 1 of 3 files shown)


hanspagel:
me don't like

PR #9625, opened 2026-06-29
fix(json-magic): resolve $dynamicRef inside the magic proxy

### LINKED ISSUE #9414: OpenAPI 3.1 schemas with $dynamicRef render as empty or unresolved shapes

hanspagel:
Resolve `$dynamicRef` in the magic proxy so shared schemas show the right fields. Repeated reads keep the same schema object.

In this example, `UserPage` now shows a `name` field for each item instead of an unresolved type.

<details>
<summary>OpenAPI example</summary>

```json
{
  "openapi": "3.1.0",
  "info": { "title": "Example", "version": "1.0.0" },
  "paths": {},
  "components": { "schemas": {
    "Page": {
      "$id": "urn:page",
      "$defs": { "item": { "$dynamicAnchor": "item" } },
      "type": "array", "items": { "$dynamicRef": "#item" }
    },
    "UserPage": {
      "$id": "urn:users", "$ref": "#/components/schemas/Page",
      "$defs": { "item": {
        "$dynamicAnchor": "item", "type": "object",
        "properties": { "name": { "type": "string" } }
      } }
    }
  } }
}
```

</details>

Dynamic bindings use a cache keyed by both scope and schema, so the same template can bind to User and Group independently while repeated reads keep stable identities. Recursive paths reuse those identities. The dynamic-reference probe runs at most once per document, lazily when a candidate resource is first read; ordinary anchor-free resources keep their shared cache even if another branch has a dynamic reference. Scoped cache entries are lazy and live with the document rather than using an eviction limit.

The [cache and performance notes](https://github.com/scalar/scalar/blob/fix/dynamic-anchor-nested/packages/json-magic/src/magic-proxy/README.md) document the scan cost, local ingestion/render measurements, and the 64-binding example. Rendering accesses the virtual binding through workspace-store.

## Ticket

Closes #9414
See #9883


marclave (approved):

DemonHa (changes requested):
I think we should integrate the $dynamicRef support on the magic proxy instead.

We shouldn't rely on the outside application to try and resolve the dynamic refs.

amritk (commented):
Conflicting with `main`, and it has been open since June with 14 commits — this needs a merge from base before it can go anywhere.

The design is the right one: `$dynamicRef` resolution belongs in the magic proxy, not threaded through Vue provide/inject in `api-reference`, and exposing the bound schema as a get-only `$dynamicRef-value` mirroring `$ref-value` is a clean way to express it. Collecting anchors anywhere inside a schema resource rather than only at the root or `$defs` is a real fix.

The part I want to look at hardest once this is mergeable, and the reason I am not approving on description alone, is this:

> The proxy cache is bypassed **only while a dynamic scope is active** (gated on `containsDynamicRef`), so the same shared template referenced from several bindings resolves to the right type per path — while ordinary documents keep their first-path-wins referential stability.

That gate is doing a lot of work. Proxy caching and referential stability are load-bearing for the whole store, and "ordinary documents are completely unaffected" is a claim about every document that does *not* contain `$dynamicRef` — which is almost all of them. Specifically I would want to confirm:

- `containsDynamicRef` is computed once per document rather than per access, and cannot be made true by a `$dynamicRef` appearing anywhere unrelated in the document (a single dynamic ref in one corner should not disable caching for the whole document, or this is a performance cliff).
- When the cache *is* bypassed, repeated access along the same path still returns a stable reference, or Vue reactivity and any `===` comparison downstream will churn.

Bugbot already flagged this as medium risk for the same reason, so it is worth a note in the description explaining the gating precisely.

Given the age and the conflict, a decision on this one would help: either bring it up to date and push it through, or close it and re-cut from current `main`. Fourteen commits of drift since June is likely more painful to reconcile than to redo.

---

amritk (commented):
Moving `$dynamicRef` resolution out of Vue `provide`/`inject` and into the magic proxy is the right call — the old design leaked a JSON Schema concern into component structure, and any consumer that did not sit under a `Schema` ancestor got the wrong answer. Exposing it as a virtual `$dynamicRef-value` mirroring `$ref-value` is a clean interface.

The `dynamic-ref.ts` tests are the strongest part of this: bookending (not borrowing an outer anchor when the innermost resource lacks the anchor), not crossing a nested `$id` resource boundary, outermost-first precedence, and cycle safety are the four things the spec makes easy to get wrong, and all four are pinned. Good.

What holds me back is that this lands in the magic proxy, which every document in the store goes through.

**Proxy identity and cost.** The changeset says active dynamic scopes use a scope-keyed proxy cache while anchor-free resources keep their shared identity. That mitigation is the right one, but I would like to see the numbers: for a large document with no `$dynamicRef` at all, what does the added scope threading cost on ingest and on render? `containsDynamicRef` appears to be the gate — is that a full document walk per proxy creation, or is it memoized? A whole-document scan to discover that the feature is unused would be a tax on the 99% case.

**Scope-keyed caching and memory.** A template referenced from many bindings now gets one proxy per scope. For a document that reuses a generic wrapper across dozens of endpoints, how many proxies is that in practice, and is the cache bounded or does it live for the document's lifetime?

**Layering.** `@scalar/api-reference` picks up a direct dependency on `@scalar/json-magic` just to import `DYNAMIC_REF_VALUE`. Everywhere else the reference renders through `@scalar/workspace-store`. Re-exporting the constant from workspace-store would keep the dependency direction as it was and avoid api-reference reaching past the store into its bundler internals.

**Nice catch** on binding a `$dynamicRef` that is an object property rather than only array items — and the comment in `SchemaObjectProperties.vue` explaining *why* binding must happen before `resolve.schema` (because coercion drops the proxy's virtual accessor) is exactly the kind of note that saves the next person an hour.

---

cameronrohani (approved):

amritk (commented):
Not approving — this is `blocked`, has been open since June across 15 commits, and at 1,335 additions over 20 files implementing `$dynamicRef` resolution it needs its assigned owner rather than a drive-by approval. @DemonHa is already assigned, which seems right.

The feature is worth having. `$dynamicRef` / `$dynamicAnchor` is the one JSON Schema 2020-12 mechanism that genuinely cannot be resolved lexically — resolution depends on the *dynamic scope* of the evaluation, so the correct target for `#item` in the `Page` schema differs depending on whether it was reached directly or through `UserPage`. The example in the description is exactly the canonical generic-container case, and getting an unresolved type where a `name` field belongs is the expected symptom of treating it as a static `$ref`.

Two things I would want the reviewer to press on, since they are where dynamic-scope implementations usually go wrong:

1. **"Repeated reads keep the same schema object" is doing a lot of work.** Identity stability is necessary for the renderer, but a `$dynamicRef` has no single correct resolution — the same node resolves differently under different dynamic scopes. If the proxy memoizes one resolution per node, the second scope silently gets the first one's answer. Worth confirming the cache is keyed by scope and not just by node, and that there is a test with the *same* `Page` reached through two different outer schemas asserting they resolve to different items.
2. **Cycle behavior.** `$dynamicAnchor` makes it easy to construct a schema whose dynamic scope never terminates. Worth checking there is a bound, especially in light of #9611, which is separately dealing with pathological cycles in the same general area.

Also worth a rebase before this is picked up — `main` has moved a long way since June, including several open PRs in this same batch that touch `json-magic` (#10274 refactors `join` onto a new module, #10206 changes reference resolution for `$self`). Landing those in either order against a three-month-old branch is going to be unpleasant.

---

amritk (commented):
Nice progress since the last round. Resolution now lives in the proxy, api-reference goes through `@scalar/workspace-store/resolve` instead of json-magic at runtime, and the README answers the scan-cost, cache-lifetime and memory questions. The proxy tests pin path-dependent binding, identity within and across scopes, and bookending. A few remaining points:

1. **Shared `anchorCache` with two kinds of callers** (`packages/json-magic/src/magic-proxy/dynamic-ref.ts:88`). The cache is module-level and keyed by the unwrapped raw object, but what it stores depends on the caller. The proxy path passes raw scope entries, so it stores raw anchor nodes. The workspace-store wrapper (`packages/workspace-store/src/helpers/dynamic-ref.ts:33`) is used by the example builder on `resolve.schema` output, which can still be a magic proxy, so it stores proxies or plain objects merged via `dereferenceSiblingRef`. Whichever caller runs first decides what the other gets. If the example builder fills the cache first, `$dynamicRef-value` takes the `isMagicProxyObject(resolved)` branch (`proxy.ts:253`) and returns a proxy created under another scope instead of wrapping the target with `childScope`. Depending on render order, that can mis-bind nested `$dynamicRef`s in recursive templates. Two options: always store raw anchor nodes and dereference sibling `$ref`s at the call site, or keep a separate cache for each `unwrap` mode.
2. **The probe and anchor cache never invalidate.** `dynamicRefsProbe` is memoized for the life of the root proxy (`proxy.ts:188`), and `anchorCache` is never invalidated. If a document is edited through the proxy's `set` trap (for example in the client), a `$dynamicRef` or `$dynamicAnchor` added later will not be picked up. This is probably fine for the reference, but a short note in the README would help.
3. **Ordinary documents pay a small cost.** The README's own numbers show cold and cached reads about 17% and 11% slower for documents with no `$id`, `$defs` or `$dynamicRef`. The absolute numbers are small, but this is the hot path for every document. Could the per-get `DYNAMIC_REF_VALUE` check and the per-proxy `carriesDynamicAnchor` work short-circuit once the probe comes back negative?
4. **Nit:** `@scalar/json-magic/magic-proxy` now exports new public API (`DYNAMIC_REF_VALUE`, `collectDynamicAnchors`, `isDynamicRef`, `pushDynamicScope`, `resolveDynamicRef`, `DynamicScope`) under a `patch` changeset. `minor` might be a better fit.

Also, @DemonHa's changes-requested review from June is still what blocks this. The PR now does what that review asked for, so a re-review or dismissal would unblock it.

---

PR #9626, opened 2026-06-29
fix(schemas): preserve AsyncAPI security scheme types on ingestion

hanspagel:
AsyncAPI security schemes referenced from a server (`{ $ref: ... }`) all got their `type` overwritten to `userPassword` after ingestion — so `http`, `oauth2`, `openIdConnect` and `apiKey` collapsed into the same type.

The reference coercion was filling in a default value for unresolved `$ref` objects, and because the magic proxy shares one object between a `$ref-value` and the thing it points at, that default leaked over the real definition.

Fix: make `$ref-value` optional so unresolved refs pass through untouched. Resolved documents still work as before.

Found while verifying #9618 (document-wide AsyncAPI auth), which needs the scheme types to survive ingestion.

## Ticket

Part of #9618


bielu:
oh i missed it when testing my changes lol

marclave (approved):

PR #9627, opened 2026-06-29
chore(dotnet): adopt repo-wide Central Package Management

bielu:
## Problem

NuGet package versions for the .NET integrations were scattered across every `.csproj` (inline `Version=` attributes) plus per-tests `Directory.Packages.props` files. There was no single source of truth, so the same package could drift to different versions across projects, and bumping a shared dependency meant editing many files by hand. This makes it harder to keep versions aligned and review dependency changes — and it's a prerequisite for cleanly adding further .NET integrations.

This migration was originally bundled into #9620 (`feat(azure-functions): add Scalar.Azure.Functions integration`), but touching every `.csproj` added a lot of noise to that diff and made the feature hard to review. Splitting it into its own PR keeps both focused; #9620 will be rebased on top of this once it lands.

## Solution

Adopt **Central Package Management (CPM)** for the whole `integrations/dotnet` tree:

- Add `integrations/dotnet/Directory.Packages.props` as the single source of truth (`ManagePackageVersionsCentrally=true`) listing every package version once.
- Remove the inline `Version=` attributes from every `PackageReference` across `aspnetcore`, `aspire`, `shared`, and their tests/playgrounds.
- Delete the per-tests `Directory.Packages.props` files (`aspnetcore/tests`, `aspire/tests`); their versions now live in the root file.
- Framework-aligned packages whose version must track the target framework (`Microsoft.AspNetCore.Mvc.Testing`, `Microsoft.AspNetCore.OpenApi`) are pinned in per-`$(TargetFramework)` `ItemGroup`s.
- The one genuine outlier — the Aspire `BookService` playground, which intentionally uses `Microsoft.AspNetCore.OpenApi` `10.0.5` while the libraries pin `10.0.0` for net10 — keeps its version via `VersionOverride`.

This is **build-infrastructure only**: no package version changes for any shipped artifact, and no public API or behavior change.

**Alternatives considered:** keeping per-tests CPM files alongside a root file (rejected — NuGet uses the nearest file, so they'd shadow the root and defeat the point); leaving src projects on inline versions and only centralizing tests (rejected — half-measure, version drift still possible in src).

**Verified:** `Scalar.slnx` builds clean, and all .NET test suites pass on net9 and net10 (Shared 44, AspNetCore 39, Microsoft 5, Swashbuckle 4, Aspire 35, Aspire.Service 8); the `BookService` playground builds with the `VersionOverride`.

## Checklist

- [ ] I added a changeset. 
- [ ] I added tests. 


hanspagel:
wait, I think we're managing those versions through [changesets](https://github.com/changesets/changesets), basically dumping the version number from the package.json in there

but maybe I misunderstood, need to look at it

bielu:
@hanspagel i  would be suprised if you would, you managed them directly in csproj, unless you have magic script somewhere which i didnt found! Anyway, the packages should be centralised so if you do have script i will adjust it, becasue restoring nuget with legacy approach cause nuget to query for each project the same package from cache/nuget server

hanspagel:
ah, we actually have this [1], but it's unrelated to the PR

[1] https://github.com/scalar/scalar/tree/main/.github/workflows/publish-integrations.yml

```yaml
      - name: Get version from package.json
        id: package-version
        working-directory: integrations/dotnet/aspnetcore
        run: echo "VERSION=$(node -p "require('./package.json').version")" >> "$GITHUB_OUTPUT"
```

bielu:
AHH so release version, yeah that's different!

hanspagel (approved):

PR #9634, opened 2026-07-01
fix(api-client): let the test client attach any auth scheme

### LINKED ISSUE #9632: in new version 2.16.7 the Authentication section have UI issue and Auth Type can't opened
### What happens?

In new version 2.16.7 when i opened app. Scalar frontend UI had issue on Authentication panel. I want to set Auth bearer token in header and the Auth Type combo could not opened. I think is open behind the section panel.

Image

in older version 2.16.6 it was correct.

Image

### OpenAPI Document

_No response_

hanspagel:
The Auth Type dropdown in the embedded "Test Request" client was empty, so you could not set a Bearer token. The standalone client was fine.

What I found:

- It is not a Safari z-index bug. The dropdown is genuinely empty, the caret just flips open over nothing.
- Since #9592 the embedded modal respects the operation's declared `security`. Operations that opt out with `security: []` (like the galaxy `POST /user/signup`) or that declare no security therefore offer nothing.
- But the embedded client is a testing tool, exactly like the standalone client — which already lets you attach any defined scheme via `createAnySecurityScheme`.

The fix:

- Treat the embedded modal like the standalone client: it can attach any defined scheme, even when an operation opts out with `security: []`.
- The reference docs auth block is unchanged and still respects the declared security (#7400).

Tests:

- Added a failing test first: the modal-layout auth selector now gets `createAnySecurityScheme: true`.

## Ticket

Fixes #9632

### DIFF
--- a/.changeset/wise-auth-schemes-return.md
+++ b/.changeset/wise-auth-schemes-return.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
+
+Let the embedded API client (the "Test Request" modal) attach any defined security scheme, like the standalone client already does. Since #9592 the modal respected the operation's declared `security`, so operations that opt out with `security: []` (or a subset) offered no way to set a token while testing. The reference docs auth block still respects the declared security.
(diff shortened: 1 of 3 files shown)


hanspagel:
not needed anymore I think, because reverted the PR that this was trying to patch

PR #9638, opened 2026-07-01
fix(workspace-store): keep $ref-value optional in reference() helper

hanspagel:
Follow-up to #9618. The shared `reference()` helper required `$ref-value`, so coercing an unresolved `{ $ref }` synthesized a default and dropped the reference. Make it optional in both OpenAPI 3.1 and 3.2 helpers, matching the schema position that already does this. Also align the duplicate `RefNode` in `get-resolved-ref-deep.ts`.

## Rebase follow-up

Rebased on main `94091b585f`. Main now uses OpenAPI 3.2 schemas, so the fix and unresolved-reference regression also cover the 3.2 helper. Removed the now-unnecessary type-error suppression from the unresolved request-body fixture. There were no submitted reviews or review threads to address.

## Validation

- 746 schema/reference tests and 62 downstream CodeExample/document-outline tests passed (808 total; 5 todo).
- One schema-coercion integration test requires Google Cloud fixture storage and was blocked by network access.
- Workspace-store type check passed using temporary source aliases to current workspace files. Standard resolution initially encountered stale/missing installed declarations; temporary validation configuration is not committed.
- Changed-file Biome, changeset formatting, scoped Knip, changeset status, and whitespace checks passed.

## Ticket

Part of #9618
See #9626

### DIFF
--- a/.changeset/optional-reference-ref-value.md
+++ b/.changeset/optional-reference-ref-value.md
@@ -1,0 +1,5 @@
+---
+'@scalar/workspace-store': patch
+---
+
+Keep `$ref-value` optional in the shared `reference()` schema helper. An unresolved `{ $ref }` (for example a sparse chunk from the server store) now passes through coercion untouched instead of being coerced into a synthesized default that dropped the reference. This aligns the helper with the schema position that already made this choice.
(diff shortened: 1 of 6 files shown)


amritk (approved):
The reasoning holds: with `$ref-value` required, an unresolved `{ $ref }` matches neither union branch, so coercion falls through to the object branch and synthesises a default — which is exactly the phantom `{ description: '' }` the `responses.test.ts` update removes. Making it optional is the fix, and it aligns the helper with the schema position that already made that choice.

The type-level fallout is the part worth watching, since every `$ref-value` access is now optional. `types` is green and the two direct accesses in the diff (`get-resolved-ref-deep.ts`, the `CodeExample` test) are handled, and runtime reads already go through `getResolvedRef`, so I am satisfied. Applying it to both 3.1 and 3.2 in one go is right.

---

PR #9639, opened 2026-07-02
feat(api-reference): expose global authentication state to plugins

bielu:
## Problem

Plugins had no way to read the user's authentication state. A plugin (e.g. a
custom auth helper, a "copy as cURL" view, an analytics or diagnostics panel)
could receive the resolved `config` in its lifecycle hooks but could not see
which security schemes the user selected or the secrets they entered (tokens,
API keys, OAuth credentials). That state lives in the workspace store and was
never surfaced through the plugin API, so plugins that need to react to auth
had no supported entry point.

## Solution

Add a **read-only** accessor for the global authentication state and expose it
through both plugin surfaces:

- **Lifecycle hooks** — `onInit` and `onConfigChange` now receive `auth`
  alongside `config`.
- **View components** — the plugin manager exposes `getAuthState()`, reachable
  via `usePluginManager()`.

The accessor exposes only the read methods, so plugins can inspect but not
mutate auth:

- `auth.export()` — snapshot of the entire auth state, keyed by document name
- `auth.getAuthSecrets(documentName, schemeName)`
- `auth.getAuthSelectedSchemas(payload)` — document- or operation-level

Implementation notes:

- The accessor in `ApiReference.vue` wraps `workspaceStore.auth` (rather than
  passing the store through) so the setters stay out of the plugin API.
- `@scalar/types` defines the auth types **structurally** because it can't
  depend on `@scalar/workspace-store` (that would be a dependency cycle).
- `pluginManager` creation was moved to **after** `workspaceStore` is created:
  `notifyInit` runs synchronously and a plugin's `onInit` may call `auth`, so
  the previous ordering risked a temporal-dead-zone `ReferenceError`.
- `auth` is **required** in the hook input (the manager always provides it), so
  plugin authors always get it without optional chaining.

## Checklist

- [x] I added a changeset.
- [x] I added tests.

### DIFF
--- a/.changeset/plugin-auth-state.md
+++ b/.changeset/plugin-auth-state.md
@@ -1,0 +1,7 @@
+---
+'@scalar/api-reference': minor
+'@scalar/types': minor
+'@scalar/schemas': patch
+---
+
+Add a read-only accessor for the global authentication state to the plugin API. Plugin lifecycle hooks (`onInit`, `onConfigChange`) now receive an `auth` accessor alongside `config`, and the plugin manager exposes `getAuthState()` for view components. Plugins can read stored secrets and the selected security schemes via `auth.export()`, `auth.getAuthSecrets(documentName, schemeName)`, and `auth.getAuthSelectedSchemas(payload)` without being able to mutate auth.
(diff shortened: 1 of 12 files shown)


hanspagel (approved):
super cool

bielu:
Thanks @hanspagel one less blocking for signal r and grpc console plugin in first beta!

PR #9642, opened 2026-07-02
ci: upload Playwright reports to Netlify only on failure

hanspagel:
The Playwright report jobs deploy their HTML report to Netlify and drop a PR comment on **every** run — including green ones, where there is nothing to look at. That upload sits on the critical path.

This gates the Netlify deploy + comment steps behind `failure()` in the four report jobs (component snapshots, api-reference e2e, scalar-app e2e, CDN snapshots), so a passing run skips them. When a test actually fails or a snapshot changes, the report still uploads — that is exactly when you want it.

### Before / after

The two report jobs are the last things a green PR run waits on. On a recent run each spent ~20s on the Netlify upload after tests already passed:

| Job | Netlify upload (green run) | After |
|-----|------|-------|
| `test-component-snapshots` (tail job) | ~20s | skipped |
| `test-e2e-api-reference` | ~19s | skipped |

`test-component-snapshots` is what sets the finish line, so this trims ~20s off total wall-clock for every passing PR (~3m40s → ~3m20s), and stops posting empty report comments.

---

> [!NOTE]
> **Low Risk**
> CI-only workflow condition changes; failure paths still upload reports, with slightly less visibility if a later step fails after tests pass.
> 
> **Overview**
> **PR CI** now runs Netlify Playwright report uploads and PR summary comments only when the preceding test step **fails**, instead of on every non-cancelled run.
> 
> The condition on both follow-up steps in **`test-e2e-api-reference`**, **`test-e2e-scalar-app`**, **`test-component-snapshots`**, and **`test-cdn-snapshots`** changes from `!cancelled()` to **`failure()`**. Same-repo and non-bot guards are unchanged. Passing runs skip deploy and comments, which trims wall-clock time and avoids empty report comments; failed or snapshot-mismatch runs still get the hosted report and PR link.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 73c02d74599ddc7305e5571e1b72fabf9dc34171. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -331,7 +331,7 @@ jobs:
       - name: Run e2e tests (@scalar/api-reference)
         run: pnpm --filter api-reference test:e2e:ci
         # Only when the PR is from the same repository
-      - if: ${{ !cancelled() && github.event.pull_request.head.repo.full_name == github.repository && !contains(github.actor, '[bot]') }}
+      - if: ${{ failure() && github.event.pull_request.head.repo.full_name == github.repository && !contains(github.actor, '[bot]') }}
         name: Deploy Playwright Report to Netlify
         run: |
           pnpx netlify deploy --dir "./packages/api-reference/playwright-report" \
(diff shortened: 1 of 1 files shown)


hwkr (approved):
This is probably fine but we sometimes the success reports can be helpful as well as they can contain screenshots or other diagnostic info. There is also a way to store a GitHub actions artifact, maybe that would be faster?

I'm done to turn it off for not to speed up build and turn it back on later if needed.

hanspagel:
we can keep it, too, that's fine.

I think artifacts would actually be cool, because my LLMs tried to access those for making CI pass but were disappointed to not find any

hanspagel:
let's just keep it

PR #9646, opened 2026-07-02
fix(api-reference): deep links to response properties

### LINKED ISSUE #9364: Property links not working
### What happens?

Some direct property links aren't working when you open them in a new page and just load the top of the page (they do work if it's the same tab). I can't seem to work out why. For me it seems to happen in responses and not requests, so I checked a [Mersk API](https://developer.maersk.com/api-catalogue/Locations) and it's happening there in requests so I've attached a video.

**Edit**: I just did some more testing on my APIs. When I don't expand all responses, there is no property link. Is that intended?

**Edit 2**: So when `expandAllResponses: true`, that then enables property links in responses at the top level, but it doesn't seem to work. Tested with your OpenAPI and HTML examples, just adding `expandAllResponses: true` to the config. See HTML file attached.

[scalar.html](https://github.com/user-attachments/files/28406552/scalar.html)

https://github.com/user-attachments/assets/7723a4c7-dba0-4978-826c-11f5144ff10e

Direct property links should link directly to the property.

### OpenAPI Document

_No response_

### LINKED ISSUE #9636: Property links still not working
### What happens?

There was this ticket for property links not working, a PR was done and it was closed 2 weeks ago, but the issue isn't fixed: https://github.com/scalar/scalar/issues/9364

I posted in the ticket 2 weeks ago, but it still remains closed. I couldn't see where to reopen it (I assume that would be the workflow) so I've created this ticket. Apologies if I should do different.

The example HTML I provided (and also below), property links still don't work in **responses** when **opened in a new tab**.  As we want to link to properties in our user guides, this will open the link in a new tab directly to the property.

[scalar.html](https://github.com/user-attachments/files/29542915/scalar.html)

Property links to responses should work when opened in a new tab, as we want to link to properties in our user guides, and this will open the endpoint property in a new tab.

### OpenAPI Document

_No response_

hanspagel:
Deep links to response properties did not scroll when opened in a new tab, while request body links did. This fixes both cases (expanded and collapsed responses).

**Scrolling to the target.** Response property anchors had no marker segment (`operation.200.name`), so we couldn't map them back to an operation. On a fresh load the target operation stayed a lazy placeholder — nothing to scroll to. Anchors now carry a `responses` marker (`operation.responses.200.name`), the same way request bodies use `body`.

**Collapsed responses.** With `expandAllResponses` off, responses start collapsed and had no copy-link at all (the schema was rendered without a name to avoid a duplicate heading, which also dropped the breadcrumb). The breadcrumb now carries the status code instead, so properties stay linkable, and a deep link opens the collapsed response on load the same way collapsible schema sections already do.

Also replaced the marker-matching regex with a plain string scan to avoid the polynomial-backtracking (ReDoS) pattern CodeQL flagged — behavior is unchanged.

Tests: unit coverage for the marker parsing, plus e2e for both the expanded and collapsed deep-link flows.

## Ticket

Fixes #9636
Closes #9364

---

> [!NOTE]
> **Low Risk**
> Navigation and disclosure behavior in api-reference only; no auth or data changes, with targeted unit and e2e coverage.
> 
> **Overview**
> Fixes **response property deep links** so they behave like request-body links on fresh navigation.
> 
> Anchors now include a **`responses`** segment (e.g. `…/list.responses.200.name`), and **`getSchemaParamsFromId`** treats `.responses.` like `.body.` / parameter markers so the lazy bus can resolve the parent operation and scroll. Marker detection was switched from a greedy regex to a **first-marker string scan** (same behavior, avoids ReDoS).
> 
> **UI wiring:** response list items append `responses` (and status code on the schema breadcrumb when collapsible), **`SchemaProperty`** keeps breadcrumbs linkable for nameless response schema wrappers, and collapsed responses **auto-open** when `scrollTargetId` matches the schema path.
> 
> Coverage: unit tests for parsing edge cases and Playwright e2e for lazy-load scroll + collapsed-response reveal.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bb190dedcb6625c60f0db0af47f3773c5b809ec4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/api-reference/src/components/Content/Schema/SchemaProperty.vue
+++ b/packages/api-reference/src/components/Content/Schema/SchemaProperty.vue
@@ -100,8 +100,14 @@ const optimizedValue = computed(() =>
 )
 
 const childBreadcrumb = computed<string[] | undefined>(() =>
-  props.breadcrumb && props.name
-    ? [...props.breadcrumb, props.name]
+  props.breadcrumb
+    ? // A named property extends the breadcrumb with its own name. A nameless
+      // container (e.g. a response object, rendered without a name to avoid a
+      // duplicate heading) passes its breadcrumb straight through so its
+      // properties stay linkable.
+      props.name
+      ? [...props.breadcrumb, props.name]
+      : props.breadcrumb
     : undefined,
 )
 
--- a/.changeset/fix-collapsed-response-anchors.md
+++ b/.changeset/fix-collapsed-response-anchors.md
@@ -1,5 +1,0 @@
----
-'@scalar/api-reference': patch
----
-
-Add anchor links to response properties and open collapsed responses on deep link. Response properties are now linkable even when `expandAllResponses` is off, and a deep link into a collapsed response expands it and scrolls the property into view.
--- a/.changeset/fix-collapsed-response-anchors.md
+++ b/.changeset/fix-collapsed-response-anchors.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
+Add anchor links to response properties and open collapsed responses on deep link. Response properties are now linkable even when `expandAllResponses` is off, and a deep link into a collapsed response expands it and scrolls the property into view.
--- a/.changeset/fix-response-deep-links.md
+++ b/.changeset/fix-response-deep-links.md
@@ -2,4 +2,4 @@
 '@scalar/api-reference': patch
 ---
 
-Fix deep links to response properties not scrolling when opened in a new tab. Response property anchors now carry a `responses` marker so the target operation is found and rendered on a fresh load, just like request body links.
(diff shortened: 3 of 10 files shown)


marclave (commented):
gunna let @amritk take a review!

packages/api-reference/src/components/Content/Schema/SchemaProperty.vue
```diff
@@ -100,8 +100,14 @@ const optimizedValue = computed(() =>
 )
 
 const childBreadcrumb = computed<string[] | undefined>(() =>
-  props.breadcrumb && props.name
-    ? [...props.breadcrumb, props.name]
+  props.breadcrumb
```
amritk, line 103: Logic is fine but normally we don't like to nest ternaries and prefer helpers with early returns for clarity. I wonder if we can add a biome rule for it

amritk (approved):
Rest looks pretty good, did not test

PR #9647, opened 2026-07-02
fix(api-reference): link response properties in collapsed responses

### LINKED ISSUE #9364: Property links not working

### LINKED ISSUE #9636: Property links still not working

hanspagel:
Follow-up to #9646. That PR fixed response deep links when `expandAllResponses` is on. This one covers the default case, where responses start collapsed.

Two things were missing:

- Collapsed responses rendered their schema without a name (to avoid a duplicate heading), which also dropped the breadcrumb, so response properties had no copy-link at all. The breadcrumb now carries the status code instead, so properties stay linkable without the extra heading.
- A deep link into a collapsed response did nothing, because the response never opened. It now opens on load when the link points inside it, the same way collapsible schema sections already do.

Stacked on #9646 — the `responses` marker from that PR is what lets these links resolve.

## Ticket

Fixes #9636
Closes #9364

### DIFF
--- a/.changeset/fix-collapsed-response-anchors.md
+++ b/.changeset/fix-collapsed-response-anchors.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
+Add anchor links to response properties and open collapsed responses on deep link. Response properties are now linkable even when `expandAllResponses` is off, and a deep link into a collapsed response expands it and scrolls the property into view.
--- a/packages/api-reference/src/components/Content/Schema/SchemaProperty.vue
+++ b/packages/api-reference/src/components/Content/Schema/SchemaProperty.vue
@@ -100,8 +100,14 @@ const optimizedValue = computed(() =>
 )
 
 const childBreadcrumb = computed<string[] | undefined>(() =>
-  props.breadcrumb && props.name
-    ? [...props.breadcrumb, props.name]
+  props.breadcrumb
+    ? // A named property extends the breadcrumb with its own name. A nameless
+      // container (e.g. a response object, rendered without a name to avoid a
+      // duplicate heading) passes its breadcrumb straight through so its
+      // properties stay linkable.
+      props.name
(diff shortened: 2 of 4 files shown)


hanspagel:
Folded into #9646 — the two changes have to ship together (the collapsed-response links rely on the `responses` marker), so it is simpler as one PR.

PR #9648, opened 2026-07-02
ci: build playwright-runner image with Chromium only

hanspagel:
Every Playwright config in this repo only runs Chromium, but the runner image installs Chromium, Firefox, and WebKit. The unused browsers make the image ~1.7 GB, and each of the Playwright CI jobs spends ~20–30s pulling and extracting them before any test runs.

This changes both Dockerfiles to `playwright install --with-deps chromium` and drops the leftover Firefox project from the Nuxt config, so nothing depends on a non-Chromium browser anymore.

### Prep only

This does not switch CI to a new image yet. The workflow image tags, the `@playwright/test` catalog, and `DEFAULT_PLAYWRIGHT_VERSION` all stay at the published `1.59.1`, so CI is unchanged and stays green. The only version bump is the build tag in `update-playwright-docker`, set to `1.62.1` for the next publish.

### To actually adopt it (follow-up)

1. Build and push the Chromium-only image: `pnpm script run update-playwright-docker` (needs the `scalarapi` Docker Hub login).
2. In a separate PR, point the workflow `image:` tags, the `@playwright/test` catalog, and `DEFAULT_PLAYWRIGHT_VERSION` at the new tag. These three must move together.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -157,7 +157,7 @@ jobs:
     needs: [build]
     runs-on: blacksmith-4vcpu-ubuntu-2204
     container:
-      image: scalarapi/playwright-runner:1.61.1
+      image: scalarapi/playwright-runner:1.59.1
       options: --privileged
       env:
         VM_ID: ${{ env.VM_ID }}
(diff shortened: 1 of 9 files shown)


hwkr:
@geoffgscott is able to push the new image, if we're going to change this let's also update to the latest playwright (needs to be updated on the image and in pnpm workspace) so the container version number matches playwright

hanspagel:
I can push too, I’d even consider to hook it into the release process?

marclave (approved):

PR #9656, opened 2026-07-02
test(workspace-store): cover optional operation security

kriptoburak:
## Problem

Operation schema tests cover a required security requirement, but not the OpenAPI optional auth form that uses `{}` as one alternative.

## Solution

Add a regression test for an operation security array with anonymous access and an API key alternative.

## Validation

- Git diff whitespace check passed.
- `pnpm test packages/workspace-store/src/schemas/v3.1/strict/operation.test.ts` could not start because this temporary checkout has no node_modules.

## Checklist

- [x] I added tests.
- [ ] I added a changeset. Not needed for a test-only change.
- [ ] I updated the documentation. Not needed for a test-only change.


hanspagel (approved):
ok, cool

PR #9661, opened 2026-07-04
test(scalar-app): add snapshot tests for the sync conflict editor

amritk:
## What

Adds visual snapshot coverage for the sync conflict (three-way merge) editor in scalar-app:

- A test-only harness route (`/@local/default/test/sync-conflict-editor`) that renders `SyncConflictResolutionEditor` with a deterministic fixture: two conflicts (`info.title`, `info.version`) plus a non-conflicting remote change that adds a `servers` entry. The component is lazy-loaded, so it stays out of the main bundle.
- A Playwright suite that:
  - snapshots the initial conflict state (three panes, conflict decorations, code lenses)
  - resolves one conflict as local and one as remote via the code lenses, snapshots the resolved state, and asserts the exact applied merge payload

## Why

The merge editor is being extracted into a standalone package in a follow-up PR. Landing these snapshots on `main` first means that PR's CI will verify the extracted editor renders identically to the current in-app implementation.

## Notes

- Snapshots are taken as element screenshots of the harness container with the Monaco overview ruler masked (it paints with variable timing) and a 0.001 max diff pixel ratio to absorb sub-pixel antialiasing noise in Monaco's text rendering.
- The suite runs in the existing `test-e2e-scalar-app` job; verified stable across 10+ consecutive local runs against the dockerized Playwright runner.


amritk:
nobody said open a PR!

PR #9662, opened 2026-07-04
feat(types): add documentType to source configuration

bielu:
## Problem

.NET already lets a document declare `DocumentType.AsyncApi` (added in #9413), but that information never reaches the browser — the JS `SourceConfiguration` has no equivalent field. Every other integration (fastapi, django-ninja, Java, Rust) has the same gap: there's no explicit, typed way to say "this source is AsyncAPI" in the config that ends up in the rendered page. The renderer auto-detects the type from document content today, so this isn't a functional bug, but it blocks two things: integrations that want to be explicit (e.g. to pick a route pattern or label before the document has loaded), and parity across languages.

## Solution

Added an optional `documentType: 'openapi' | 'asyncapi'` field to the shared source configuration, in the three places it's defined: `packages/schemas/src/api-reference/source-configuration.ts` (the `@scalar/validation` schema used by `createApiReference` and `normalizeConfigurations`), `packages/types/src/api-reference/source-configuration.ts` (the zod schema), and the hand-written `SourceConfiguration` type in `packages/types/src/api-reference/types.ts`.

Deliberately scoped as pass-through only: nothing in `packages/workspace-store` consumes it yet. The renderer keeps auto-detecting the type from the document's own `asyncapi`/`openapi` version field, so this field is a forward-compatible hint, not a requirement. This also avoids overlapping with the currently-open document-selector type-postfix PR (#9602), which derives the type post-fetch.

Considered making the field required or consuming it immediately in the document selector, but that would duplicate work already in flight on #9602 and add a config requirement where none exists today.

**Included in this branch:** an unrelated fix discovered while committing — `packages/schemas/scripts/generate-types.ts`, `stage-generated-types.ts`, and `packages/release-notes/scripts/stage-generated-schema.ts` call `spawnSync('pnpm', ...)` without `shell: true`, which fails with `ENOENT` on Windows (Node doesn't do `PATHEXT` resolution for a bare command without a shell). This broke the `schemas-types-generate` pre-commit hook on Windows dev machines for every contributor, not just this change, so it's fixed as its own commit ahead of the `documentType` commit.

## Tests

- `packages/schemas/src/api-reference/api-reference-configuration.test.ts` — coercion preserves a valid `documentType` and falls back to the first-listed literal (`'openapi'`) for an invalid one, matching this schema's existing union-literal fallback behavior.
- `packages/types/src/api-reference/api-reference-configuration.test.ts` — zod parse preserves the field.
- `packages/api-reference/src/helpers/normalize-configuration.test.ts` — a source's `documentType` survives into the normalized per-document config.
- `packages/client-side-rendering/src/html-rendering.test.ts` — `documentType` is serialized into the injected `<script>` config.

All four suites pass (58 + 130 + 21 + 39 tests).

## Checklist

- [x] I added a changeset. (`.changeset/source-document-type.md`)
- [x] I added tests.
- [ ] I updated the documentation. (Covered in the separate docs PR, `docs/asyncapi-integrations`, which depends on this one merging first.)

### DIFF
--- a/.changeset/source-document-type.md
+++ b/.changeset/source-document-type.md
@@ -1,0 +1,8 @@
+---
+'@scalar/schemas': minor
+'@scalar/types': minor
+---
+
+feat: add an optional `documentType` field (`'openapi' | 'asyncapi'`) to the source configuration
+
+Sources can now declare their document type explicitly, e.g. `sources: [{ url: '/asyncapi.json', documentType: 'asyncapi' }]`. The field is a forward-compatible hint: the renderer keeps auto-detecting the type from the document content, so it stays optional.
--- a/packages/api-reference/src/helpers/normalize-configuration.test.ts
+++ b/packages/api-reference/src/helpers/normalize-configuration.test.ts
@@ -1,6 +1,36 @@
 import { describe, expect, it } from 'vitest'
 
-import { normalizeContent } from './normalize-configurations'
+import { normalizeConfigurations, normalizeContent } from './normalize-configurations'
+
+describe('normalizeConfigurations', () => {
+  it('preserves the documentType of a source in the normalized config', () => {
+    const normalized = normalizeConfigurations({
+      sources: [
+        {
+          title: 'Streaming API',
+          url: 'https://example.com/asyncapi.json',
+          documentType: 'asyncapi',
+        },
+      ],
+    })
+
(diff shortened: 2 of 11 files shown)


hanspagel:
sorry, can you explain in simple terms why we'd need this?

bielu:
It is more about performance than actual need, we suggest to detection document type so detection can be faster

hanspagel:
I think we should live without the additional property ✌️

PR #9663, opened 2026-07-04
fix(docker): serve AsyncAPI documents

bielu:
## Problem

The Docker integration only scanned mounted files for OpenAPI/Swagger. AsyncAPI documents dropped into `/docs` were silently skipped, so they never showed up in the generated `configuration.json` — even though Caddy already serves the whole mount.

## Solution

The document scanner now also recognizes AsyncAPI files (`"asyncapi"` in JSON, `asyncapi:` in YAML) next to OpenAPI/Swagger. They are served at the existing `/openapi/{filename}` path, which is kept for both document types so existing deployments do not need to change their mount or proxy setup.

Also added `integrations/docker/example/docs/events-api.yaml` as a small AsyncAPI sample so the example directory covers both document types.

## Tests

Added `integrations/docker/scan-documents.test.ts`, a Vitest suite that runs the scanner against temporary fixture directories and checks the generated configuration. It covers OpenAPI, Swagger, and AsyncAPI detection (both JSON and YAML), skipping files without a version marker, default selection, titles and slugs from nested directories, and the `BASE_PATH` prefix.

To make the script testable outside Docker, `MOUNT_DIR` and `CONFIG_FILE` are now read from the environment with the existing container paths as defaults, so runtime behavior is unchanged.

## Checklist

- [x] I added a changeset.
- [x] I added tests.

### DIFF
--- a/.changeset/docker-asyncapi-scan.md
+++ b/.changeset/docker-asyncapi-scan.md
@@ -1,0 +1,7 @@
+---
+'@scalarapi/docker-api-reference': patch
+---
+
+fix: detect and serve mounted AsyncAPI documents
+
+The document scanner only recognized OpenAPI/Swagger files, so AsyncAPI documents mounted into `/docs` were silently skipped. AsyncAPI files (`.json`/`.yaml`/`.yml` with an `asyncapi` version field) are now detected and included in the generated configuration with `documentType: "asyncapi"`.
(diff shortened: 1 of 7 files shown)


bielu:
i noticed you put it on hold, @hanspagel do you want me to hold on with raising other prs for now? i have few more 😓 prepared but yeah i hit limit of open prs 😂

.changeset/docker-asyncapi-scan.md
```diff
@@ -0,0 +1,7 @@
+---
+'@scalarapi/docker-api-reference': patch
```
hanspagel:
```suggestion
'@scalarapi/docker-api-reference': minor
```

hanspagel (approved):
let's get this in! (finally)

PR #9664, opened 2026-07-05
fix(api-reference): keep base allOf properties when merging oneOf/anyOf branches

### LINKED ISSUE #9657: Object properties not showing with allOf
### What happens?

My spec uses an allOf for the properties on a schema, with a oneOf within that - when I render it, only the oneOf properties are showing:

Image

This is a minimal spec that reproduces it - this same spec does show OK in SwaggerEditor.

I expected to see all the properties, and not "object"!

### OpenAPI Document

[api-1 (1).json](https://github.com/user-attachments/files/29623502/api-1.1.json)

momomuchu:
When a schema factors shared object properties into an allOf base next to a oneOf/anyOf, each branch's own properties/required replaced the base ones instead of combining with them. Only the last branch's fields showed up, the shared fields disappeared.

Looks like a side effect of #9557: once mergeAllOfSchemas started keeping oneOf/anyOf as a sibling array instead of flattening it, optimizeValueForDisplay still spread the branch schema over the root properties, so a branch with its own properties key wiped out the base ones. Added mergeSchemaProperties to union properties/required instead of overwriting.

Test mirrors the repro attached to the issue.

Closes #9657

### DIFF
--- a/.changeset/short-cobras-relax.md
+++ b/.changeset/short-cobras-relax.md
@@ -1,0 +1,7 @@
+---
+'@scalar/api-reference': patch
+---
+
+fix: keep base allOf properties when merging oneOf/anyOf branches
+
+When a schema used `allOf` to factor out shared object properties next to a `oneOf`/`anyOf`, each branch's own `properties`/`required` overwrote the shared base fields instead of being combined with them. The base fields now stay visible alongside each branch's own fields.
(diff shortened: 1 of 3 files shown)


hanspagel (approved):
really good PR, thank you 👏

PR #9693, opened 2026-07-10
fix(api-reference): transfer SSR workspace state to the client

hanspagel:
When the API reference is server-rendered, the server loads the document before rendering, but the client loaded it asynchronously *after* the first render. So the client's first paint was an empty loading state that did not match the server markup, and Vue reported hydration mismatches (part of #4458).

This ships the loaded document alongside the markup: the server writes the store state into a small `<script type="application/json">` payload, and the client reads it back synchronously before its first render. Both sides now render the same thing, so those mismatches are gone. The old async load on the client becomes a no-op because the document is already there.

It's self-contained — the payload lives inside the reference's own markup, so there's no host wiring to do. On a plain client-only app (no SSR) nothing changes: no payload is written and nothing is read.

## Notes

This does not fully green the SSR regression test in #9544 yet. It clears the big "document is missing on the client" mismatch. A separate, deeper layer remains in the developer-tools popover (`ScalarTeleport`/`ScalarPopover` render `display:contents` on the client but not the server) — that one lives in the components package and will be a follow-up.

## Ticket

Part of #4458

### DIFF
--- a/.changeset/ssr-workspace-state-transfer.md
+++ b/.changeset/ssr-workspace-state-transfer.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
+Pass the loaded document from the server to the client during SSR so the first client render matches the server markup. Previously the server rendered the full reference while the client hydrated an empty loading state, causing hydration mismatches.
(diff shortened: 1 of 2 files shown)


hanspagel:
Maybe we can just transfer part of it … to improve the experience for bigger documents.

PR #9697, opened 2026-07-11
feat(api-client): add credential inputs for AsyncAPI broker security schemes

bielu:
## Problem

Document-wide auth for AsyncAPI (#9618) put every scheme from `components.securitySchemes` into the Authentication selector, but the eight broker-specific scheme types — `userPassword`, `plain`, `scramSha256`, `scramSha512`, `X509`, `gssapi`, `symmetricEncryption`, `asymmetricEncryption` — had no credential input UI. Selecting one showed "The `<type>` security scheme type is not supported yet.", which made the selector a dead end for exactly the schemes most broker-centric documents (Kafka, MQTT, AMQP) actually use. `documentation/asyncapi.md` tracked this as a known gap.

There was also a latent store bug hiding behind the missing UI: `setAuthSecrets` coerces every payload through `SecretsAuthUnionSchema`, which only modeled the four OpenAPI shapes — so even if broker credentials had been sent, they would have been laundered into the `apiKey` secret shape and lost.

## Solution

All eight types get credential inputs, grouped by shape rather than one-off per type:

| Group | Types | Inputs |
| --- | --- | --- |
| SASL | `userPassword`, `plain`, `scramSha256`, `scramSha512` | Username + Password |
| X509 | `X509` | Client certificate + Private key (PEM, masked) |
| Encryption | `symmetricEncryption`, `asymmetricEncryption` | single Key field |
| GSSAPI | `gssapi` | Service name |

The data path was already generic (secrets flow through `auth:update:security-scheme-secrets`; the mutators never branch on type), so the change is additive at the three ends of the pipe: new members in the auth-store secrets union (with `x-scalar-secret-client-certificate`, `x-scalar-secret-private-key`, `x-scalar-secret-service-name` extensions), matching branches in `extractSecuritySchemeSecrets` with the same store-over-config precedence as HTTP basic (and type-matched, so stored `plain` secrets don't bleed into a `scramSha256` scheme), and new template branches in `RequestAuthTab.vue` mirroring the HTTP basic form. The "not supported yet" fallback stays for genuinely unknown types.

Broker scheme object types are modeled as a separate minimal `AsyncApiBrokerScheme` base in `secret-types.ts` rather than widening the strict OpenAPI `SecuritySchemeObject` union, since these types are AsyncAPI-only and must not leak into the OpenAPI schema. The HTTP request builders narrow with `if (type === ...)` and fall through for broker types, which is the intended degrade — they never build HTTP requests.

**Alternatives considered:** informational messages instead of inputs for the non-SASL types (rejected: X509/encryption/gssapi documents would still dead-end), and per-protocol typed forms (deferred: the four shape groups cover all 20 binding protocols today).

The Galaxy AsyncAPI sample now defines one scheme of each group, referenced from the production server's `security`, so the selector can be exercised manually.

## Checklist

- [x] I added a changeset. 
- [x] I added tests. 


hanspagel (approved):
v cool

PR #9700, opened 2026-07-11
feat(helpers): add query as an http method

amritk:
## Problem

`QUERY` is a safe, idempotent HTTP method that carries its query in the request body (per the IETF HTTP QUERY method and OpenAPI 3.2). The v3.2 OpenAPI schema in this repo already recognizes `QUERY`, but the shared HTTP helpers and the workspace store did not, so QUERY operations could not be recognized, rendered, or sent with a body by the API reference and API client.

## Solution

Add `query` as a first-class HTTP method and make sure the reference and client handle it correctly.

**`@scalar/helpers/http`**
- Add `'query'` to `HTTP_METHODS`, so it flows through `isHttpMethod`, `normalizeHttpMethod`, and the `httpMethods` set.
- Add a `REQUEST_METHODS.query` entry with a distinct pink badge (the only unused color in the existing palette).
- Allow QUERY to have a request body via `canMethodHaveBody` (added to `BODY_METHODS`), so `buildSafeBodyRequest` preserves the body.

**`@scalar/workspace-store`**
- Add `query` to the v3.1 path-item schemas (strict typebox schema + type, and the non-strict generator) so QUERY operations are traversed and rendered.
- Replace a duplicated local method `Set` in `server.ts` with the canonical `isHttpMethod` helper, which also picks up `query`.

**Tests** were added across helpers (`http-methods`, `http-info`, `is-http-method`, `normalize-http-method`, `can-method-have-body`), the workspace store (`for-each-path-item-operation`, `server`, `bundler`), and the api-reference, blocks, and api-client `HttpMethod` components (including a new api-client test verifying QUERY is selectable and emits correctly).

Verification: full helpers (863) and workspace-store (2566) suites pass; type checks pass across all consuming packages; Biome + Prettier clean.

## Checklist

- [x] I added a changeset.
- [x] I added tests.

---
_Generated by [Claude Code](https://claude.ai/code/session_01HX6GmipCVdHBFoL1eG8xJQ)_

### DIFF
--- a/packages/helpers/src/http/can-method-have-body.ts
+++ b/packages/helpers/src/http/can-method-have-body.ts
@@ -1,7 +1,12 @@
 import { isElectron } from '@/general/is-electron'
 
-/** HTTP Methods which can have a body */
-export const BODY_METHODS = new Set(['post', 'put', 'patch', 'delete'])
+/**
+ * HTTP Methods which can have a body
+ *
+ * QUERY is included because it is a safe, idempotent method that carries its query in the request body,
+ * which is the whole point of the method (see the IETF HTTP QUERY method and OpenAPI 3.2).
+ */
+export const BODY_METHODS = new Set(['post', 'put', 'patch', 'delete', 'query'])
 
 /**
  * Makes a check to see if this method CAN have a body.
--- a/.changeset/add-query-http-method.md
+++ b/.changeset/add-query-http-method.md
@@ -1,0 +1,6 @@
+---
+'@scalar/helpers': minor
+'@scalar/workspace-store': patch
+---
+
+Add `query` as a supported HTTP method. The QUERY method (a safe, idempotent method that carries its query in the request body, per the IETF HTTP QUERY method and OpenAPI 3.2) is now part of `HTTP_METHODS`, so it is recognized by `isHttpMethod`, `normalizeHttpMethod`, and `getHttpMethodInfo` (rendered with a distinct pink badge), and it is allowed to have a request body via `canMethodHaveBody`. The workspace store now traverses and renders `query` operations on path items, so the API reference and API client handle QUERY operations correctly.
--- a/packages/api-client/src/components/HttpMethod/HttpMethod.test.ts
+++ b/packages/api-client/src/components/HttpMethod/HttpMethod.test.ts
@@ -1,0 +1,38 @@
+import { mount } from '@vue/test-utils'
+import { describe, expect, it } from 'vitest'
+
+import HttpMethod from './HttpMethod.vue'
+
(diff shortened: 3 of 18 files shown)


CaCTuCaTu4ECKuu:
When will next version be released?
When will we have it?

bielu:
I have only question how it will work with current browsers, as non of main browsers provided query support yet, are we going to polifill it? or are we doing custom wrapper? because if you add support for query support now without doing any of that, the method will fail on browser client if I am not wrong? 🤔

CaCTuCaTu4ECKuu:
Well, I guess you're right, rn only server side more or less suports it.
But still if it is supported on server side I wanna support it and see it 
Lets go

amritk:
Yea this PR was just supposed to be for the docs, we can actually support it in our api client app since we do a custom fetch implementation. But yea browser support will need to wait

amritk:
@CaCTuCaTu4ECKuu we can probably release soon as @hanspagel reviews it

packages/helpers/src/http/can-method-have-body.ts
```diff
@@ -1,7 +1,12 @@
 import { isElectron } from '@/general/is-electron'
 
-/** HTTP Methods which can have a body */
-export const BODY_METHODS = new Set(['post', 'put', 'patch', 'delete'])
+/**
+ * HTTP Methods which can have a body
+ *
+ * QUERY is included because it is a safe, idempotent method that carries its query in the request body,
+ * which is the whole point of the method (see the IETF HTTP QUERY method and OpenAPI 3.2).
+ */
```
Coding-Kiwi, line 8: Why is there dedicated documentation for "query"?

Coding-Kiwi:
Thanks for the fast pull request addressing the discussion :)

https://github.com/scalar/scalar/discussions/9698

packages/helpers/src/http/can-method-have-body.ts
amritk: just because its new, can remove it once we fully support openapi 3.2

hanspagel:
we don't really support OpenAPI 3.2 yet (it'll come v soon), sooo this adds a bit of the 3.2 stuff to the 3.1 schemas. let me close this for now and revisit once we added 3.2

amritk:
> we don't really support OpenAPI 3.2 yet (it'll come v soon), sooo this adds a bit of the 3.2 stuff to the 3.1 schemas. let me close this for now and revisit once we added 3.2

@hanspagel thats fine but just one thing to note the reason we used our own HTTP_METHODS instead of just the openapi spec is the spec didn't have every http method we wanted to support, so http methods sort of grew ahead of the openapi version

PR #9702, opened 2026-07-12
feat(api-client): pluggable transports (http + channel) for client plugins

bielu:
## What

Adds a plugin extension point to replace the request execution engine ("try it" client) with a custom one. Plugins can now register **transports** on `ClientPlugin`, selected by **document type** (`openapi` / `asyncapi`) and the **protocol** of the target server. Two kinds are supported:

**`kind: 'http'`** — one-shot request/response, replaces the built-in fetch engine:

```ts
const customClientPlugin: ClientPlugin = {
  transports: [
    {
      documentType: 'openapi', // optional — matches both when omitted
      protocols: ['http', 'https'],
      transport: {
        kind: 'http',
        send: async (request, { documentType, protocol }) =>
          myCustomHttpClient.execute(request),
      },
    },
  ],
}
```

**`kind: 'channel'`** — long-lived bidirectional connections (SignalR, gRPC streaming, MQTT, …), replaces the built-in WebSocket client:

```ts
const signalRPlugin: ClientPlugin = {
  transports: [
    {
      documentType: 'asyncapi',
      protocols: ['wss'],
      transport: {
        kind: 'channel',
        connect: async ({ url, handlers }) => {
          const hub = new HubConnectionBuilder().withUrl(url).build()
          hub.on('message', (data) => handlers.onMessage(JSON.stringify(data)))
          hub.onclose((error) => handlers.onClose({ code: error ? 1006 : 1000, wasClean: !error }))
          await hub.start()
          return { send: (data) => void hub.send('message', data), close: () => void hub.stop() }
        },
      },
    },
  ],
}
```

## Why

Today the only way to swap the client is the app-level `customFetch` option — a single global override, HTTP-only, and not available to plugins. Keying registrations by document type **and** protocol lets a plugin take over exactly the combinations it understands, and lets plugins provide clients for protocols the API client has no built-in support for (SignalR, gRPC, MQTT, Kafka, …).

## How

- **`@scalar/oas-utils`**: new `client-transports.ts` with the registration types (`ClientTransport`, `HttpTransport`, `ChannelTransport`, `ClientTransportContext`), protocol normalization (case-insensitive, trailing `:` ignored so both `https` and `URL.protocol`'s `https:` form match), and kind-aware resolvers (`resolveHttpTransport`, `resolveChannelTransport`) sharing one matching pass. `ClientPlugin` gains an optional `transports` field.
- **`@scalar/api-client`, HTTP path**: new `request-transport.ts` helper resolves the target protocol from the fully resolved request URL (environment substitution applied, no proxy rewriting — so `{protocol}://{host}` server templates work) and adapts a matching transport to the existing `customFetch` seam in `sendRequest`. Wired into `OperationBlock.handleExecute`.
- **`@scalar/api-client`, channel path**: new `channel-transport-socket.ts` bridges a `ChannelTransport` into the socket surface the WebSocket session already drives — the session state machine, frame log, and WebSocket plugin hooks (`beforeConnect`, `onWebSocketMessage`, `onWebSocketClose`) work identically for custom transports and native sockets. `connectWebSocket` resolves plugin transports by URL protocol after the `beforeConnect` hook (so a hook-rewritten URL selects the transport). The session's socket typing is widened to a minimal structural `WebSocketLike` that both the native `WebSocket` and adapters satisfy.

An `http` transport receives the exact fetch `Request` the client built and returns a standard `Response`, so the entire existing pipeline keeps working unchanged: streaming detection, cookie persistence, response body decoding (`responseBody` handlers), and all plugin hooks.

### Resolution rules

1. App-level overrides take precedence over plugin transports: `customFetch` for requests, an explicit `customWebSocket` for connections (backwards compatible, the desktop app keeps its overrides).
2. First matching registration in plugin order wins; each execution model only considers registrations of its own `kind`.
3. No match → built-in engine (global `fetch` / native `WebSocket`).
4. Requests executed through a plugin transport **bypass the CORS proxy**: transports own their I/O, and the proxy exists to work around browser `fetch` limitations a custom client does not have (it would also hand the transport the proxy URL instead of the real target).

### Channel transport contract

`connect(options, context)` resolves once the connection is established (e.g. after `HubConnection.start()`), rejects when it cannot be (mapped to an abnormal close, mirroring a failed native WebSocket handshake). Incoming traffic is reported through `handlers`; outgoing traffic and shutdown go through the returned `{ send, close }`. A close requested while the transport is still connecting is queued and applied once connect settles.

### Design notes

- The operation block passes `documentType: 'openapi'` and `connectWebSocket` passes `asyncapi` — matching what each path executes today.
- Channel transports are wired at the `connectWebSocket` layer, so they light up fully when the interactive AsyncAPI channel client UI lands; the contract is exercised end-to-end through the session tests today.

## Testing

- `client-transports.test.ts` (oas-utils, 17 tests): normalization + resolver matrix (protocol matching, documentType scoping, plugin order, kind separation — including a plugin registering both kinds for the same protocol).
- `request-transport.test.ts` (api-client, 10 tests): protocol resolution from real `requestFactory` builders (absolute URLs, env-templated server URLs, proxy ignored, page-origin fallback), `CustomFetch` adaptation, end-to-end resolve-and-wrap.
- `channel-transport-socket.test.ts` (api-client, 10 tests): full session lifecycle through a fake transport (open, incoming/outgoing frames, close info propagation, connect rejection → abnormal close, close-while-connecting queuing) plus `connectWebSocket` resolution (plugin transport used, failure path, protocol mismatch ignored, explicit `customWebSocket` precedence).
- Pre-existing websocket-session and connect-websocket suites pass unchanged (43 tests in the channel block total).
- `vue-tsc` on api-client shows only the two pre-existing errors also present on `main` (AuthSelector.vue / OAuth2.vue, unrelated broker-auth typing).

## Follow-ups (out of scope)

- Replace the hardcoded ws/wss connectability gate in the (upcoming) channel client UI with "built-in **or** plugin-registered transport", so non-WebSocket AsyncAPI protocols become connectable when a plugin serves them.

### DIFF
--- a/.changeset/pluggable-http-transports.md
+++ b/.changeset/pluggable-http-transports.md
@@ -1,0 +1,6 @@
+---
+'@scalar/oas-utils': minor
+'@scalar/api-client': minor
+---
+
+feat: pluggable request transports for client plugins. Plugins can now register custom transports via the new `transports` field on `ClientPlugin`, keyed by document type (`openapi` / `asyncapi`) and target protocol (`http`, `https`, …). When a registration matches the request being executed, its `send(request, context)` implementation replaces the built-in fetch engine while the rest of the pipeline (streaming detection, cookies, response body decoding, plugin hooks) keeps working unchanged. The first matching registration in plugin order wins, an app-level `customFetch` option still takes precedence, and requests executed through a plugin transport bypass the CORS proxy. The `kind: 'http'` discriminator leaves room for session-based channel transports (MQTT, Kafka, …) as a follow-up once the interactive AsyncAPI channel client lands.
--- a/packages/api-client/README.md
+++ b/packages/api-client/README.md
@@ -206,6 +206,53 @@ Each handler in `responseBody` supports:
 | `rawComponent` | A custom Vue component for the raw view (mutually exclusive with `language`). |
 | `previewComponent` | A custom Vue component for the preview view. |
 
+### Custom Transports
+
+By default, requests are executed with the global `fetch`. A plugin can register a custom transport to take over request execution, selected by document type and the protocol of the target server:
+
+```ts
+import type { ClientPlugin } from '@scalar/oas-utils/helpers'
+
+const customClientPlugin: ClientPlugin = {
+  transports: [
+    {
+      // Optional: restrict to 'openapi' or 'asyncapi' documents (matches both when omitted)
+      documentType: 'openapi',
+      // Protocols this transport serves
+      protocols: ['http', 'https'],
+      transport: {
+        kind: 'http',
+        // Receives the exact fetch Request the client built and returns a standard Response
+        send: async (request, { documentType, protocol }) => {
+          return myCustomHttpClient.execute(request)
+        },
+      },
+    },
+  ],
+}
+
+createApiClientApp(el, {
+  layout: 'web',
+  plugins: [customClientPlugin],
+})
+```
+
+Each registration in `transports` supports:
+
+| Property | Description |
+|---|---|
+| `documentType` | Restrict the transport to `openapi` or `asyncapi` documents. Matches both when omitted. |
+| `protocols` | Protocols the transport serves (case-insensitive, a trailing `:` is ignored), for example `['http', 'https']`. |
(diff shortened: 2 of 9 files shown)


bielu:
didnt want open pr yet, but claude went rogue here, lol

PR #9707, opened 2026-07-13
feat(api-reference): support OpenAPI 3.2 additional operations

cmmoran:
## Problem

Scalar currently assumes the fixed OpenAPI operation method set when discovering, validating, navigating, and rendering operations. OpenAPI 3.2 adds Path Item `additionalOperations`, allowing custom HTTP methods and operation-level behavior that would otherwise be omitted or normalized incorrectly.

## Solution

Extend the shared HTTP method helpers, Path Item traversal, OpenAPI 3.2 schema handling, navigation, request controls, and API Reference rendering to preserve and display `additionalOperations`. Add regression tests across the affected packages.

The PR is intentionally source-only: generated package output and temporary fork dependency changes are excluded.

## Checklist

- [x] I added a changeset.
- [x] I added tests.
- [ ] I updated the documentation. The change follows the OpenAPI 3.2 specification and is covered by tests; user-facing documentation can be refined separately if maintainers prefer.

Reference: [OpenAPI 3.2 Path Item Object](https://spec.openapis.org/oas/v3.2.0.html#path-item-object).

Validation included `npx pnpm@10.16.1 test` and repository Prettier checks on changed Vue files.


hanspagel:
we don't really support OpenAPI 3.2 yet (it'll come v soon), sooo this adds a bit of the 3.2 stuff to the 3.1 schemas. let me close this for now and revisit once we added 3.2

thanks for bringing this up!

PR #9709, opened 2026-07-15
feat(api-client): OAuth2 token-acquisition for bearer schemes

aqeelat:
## What this does

Treats a config-defined OAuth2 flow as a **token-acquisition source** for a declared HTTP bearer scheme, rather than a competing auth method.

- An `oauth2` scheme that no operation declares is **kept out of the auth dropdown**, so it can't be multi-selected alongside bearer (which would collide on `Authorization`). A new `getOauth2AcquisitionTarget` helper finds the first interactive (authorization-code / implicit) flow.
- The bearer scheme's form gains an inline **"Authorize via OAuth2"** shortcut (plus **Refresh** for authorization-code, plus a **gear** that opens the OAuth2 configuration in a modal). The flow's access token is written onto the **bearer** scheme, so the panel never switches to oauth2 and the request sends `Authorization: Bearer`. The refresh token stays on the oauth2 scheme — the only place that knows how to refresh.
- `runOAuth2Authorize` + `storeOAuth2Tokens` route the access token to the bearer scheme and the refresh token to the oauth2 scheme.

## Why config-side

Our gateway serves one merged spec declaring only `BearerAuth`. Declaring oauth2 in the spec isn't safe across generators: openapi-generator emits OAuth2 token-acquisition infrastructure for every consumer the moment `oauth2` appears in `components.securitySchemes`, and kiota's plugin generator throws on multiple security requirements. So the flow has to live in Scalar config (`authentication.oauth2`) — which is what made it surface as a confusing peer "auth method" in the first place.

## Related

- #9592 / #9637 — the per-operation security gating that previously hid a config-defined oauth2 scheme.

### DIFF
--- a/.changeset/oauth2-token-acquisition.md
+++ b/.changeset/oauth2-token-acquisition.md
@@ -1,0 +1,18 @@
+---
+'@scalar/api-client': minor
+---
+
+Add an OAuth2 token-acquisition shortcut to HTTP bearer schemes, so a bearer
+token can be obtained through an OAuth2 flow without switching auth methods.
+
+- The bearer scheme's form gains an inline **Authorize via OAuth2** shortcut
+  (and, for authorization-code, **Refresh**) that runs the flow and writes the
+  resulting access token onto the bearer scheme — so the panel never switches
+  to oauth2 and the request sends `Authorization: Bearer`. A gear opens the
+  oauth2 configuration in a modal (`OAuth2.hideActions`).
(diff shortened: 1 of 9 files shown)


hanspagel:
omg so cool

hanspagel:
nice

https://github.com/user-attachments/assets/4507c6af-8e49-4ed6-b752-6f7b32243a73

hanspagel (approved):

PR #9711, opened 2026-07-15
feat(api-reference): load plugins from URLs via new pluginUrls configuration option

bielu:
## Problem

API Reference plugins are functions, so they can only be registered by code that calls
`createApiReference` (or renders the component) itself. Every integration that passes its
configuration as **JSON** — the Docker container (`API_REFERENCE_CONFIG`), Scalar for Aspire,
the .NET/Java/Python server-side integrations — has no way to register a plugin at all.

Today the only workaround is to replace the whole standalone bundle with a custom build that
bakes the plugin in (for example via `ScalarAspireOptions.BundleUrl`). That means forking the
bundle, pinning it to a Scalar version, and rebuilding it for every Scalar release — just to
add one plugin.

## Solution

Add a JSON-serializable `pluginUrls: string[]` configuration option. Each entry points to an
ESM module that exports a plugin (the same shape as the `plugins` entries) as its default
export:

```js
  url: '/openapi.json',
  pluginUrls: ['https://cdn.jsdelivr.net/npm/@example/scalar-plugin/dist/plugin.js'],
})
```

- The standalone build (`Scalar.createApiReference`) imports the modules with a browser-native
  dynamic `import()` **before mounting**, because plugin registration happens once at first
  render and is not reactive. The loaded default exports are appended to the configuration's
  `plugins`.
- The import specifier is fully dynamic, so bundlers leave the `import()` call as-is in the
  UMD output (`/* @vite-ignore */` silences the analysis warning).
- Failures (network errors, modules without a function default export) are logged and
  swallowed, so a broken plugin URL never prevents the API reference itself from mounting.
- If the instance is destroyed while plugin modules are still loading, mounting is skipped.
- Each URL is imported only once, even when multiple configurations reference it.

Because the option is plain JSON, it flows through every integration that serializes its
configuration — no per-integration bundle swapping needed.

### Alternatives considered

- **Global plugin registry + extra `<script>` tags** (`window.Scalar.registerPlugin(...)`):
  requires changes in every integration's HTML template and careful script ordering; the
  config-driven approach needs no HTML changes anywhere.
- **Making plugin registration reactive** so plugins can arrive after mount: much more
  invasive (plugin manager, views, sidebar all read plugins once) and unnecessary for this
  use case.

### Changes

**Core (JS):**
- `@scalar/types`, `@scalar/schemas`: new optional `pluginUrls` field on the API Reference
  configuration schemas.
- `@scalar/api-reference`: new `load-plugins-from-urls.ts` loader; `createApiReference`
  defers mounting until URL plugins are resolved.

**Integrations with typed configuration models (exposing the new option):**
- **.NET** (`Scalar.AspNetCore` / `Scalar.Aspire` / `Scalar.Azure.Functions` via
  `Scalar.Shared`): `ScalarOptions.PluginUrls`, `ScalarConfiguration.PluginUrls`, mapper, and
  a fluent `AddPluginUrl(...)` extension.
- **Java** (`scalar-core`): `ScalarProperties.pluginUrls` → `ScalarConfiguration.pluginUrls`
  (`@JsonProperty("pluginUrls")`) + mapper.
- **FastAPI** (`scalar_fastapi`): new `plugin_urls` parameter → `pluginUrls` config key.
- **Django Ninja** (`scalar_ninja`): new `plugin_urls` field → `pluginUrls` config key.

No changes needed for: JS framework integrations (they reuse `ApiReferenceConfiguration` from
`@scalar/types`), the Docker container (passes `API_REFERENCE_CONFIG` through as-is), and the
Rust integration (freeform `serde_json::Value` configuration).

**Docs:** new "Loading a Plugin from a URL" section in `documentation/plugins.md` and a
`pluginUrls` entry in `documentation/configuration.md`.

### Tests

- `load-plugins-from-urls.test.ts`: unit tests for the loader (loading via `data:` module
  URLs, preserving directly-passed plugins, multi-configuration lists, non-function default
  exports, failed imports).
- `html-api.test.ts`: integration tests that the standalone mount is deferred until plugins
  are loaded and registered with the plugin manager, and that a destroyed instance does not
  mount late.
- .NET: `ScalarOptionsMapperTests` and `ScalarOptionsExtensionsTests` cover the new property,
  mapping, and fluent extension (green on net8.0/net9.0/net10.0).
- Note: the Java change follows the existing property/configuration/mapper pattern but was
  not compiled locally (no Maven toolchain on the machine this was authored on).

## Checklist

- [x] I added a changeset. 
- [x] I added tests.

### DIFF
--- a/.changeset/plugin-urls.md
+++ b/.changeset/plugin-urls.md
@@ -1,0 +1,9 @@
+---
+'@scalar/api-reference': minor
+'@scalar/schemas': minor
+'@scalar/types': minor
+---
+
+feat: add `pluginUrls` configuration option to load API Reference plugins from URLs
+
+Each entry must point to an ESM module that exports a plugin (the same shape as the `plugins` entries) as its default export. The standalone build (`Scalar.createApiReference`) imports the modules before the API reference mounts and registers their default exports alongside the plugins passed directly. Unlike `plugins`, the new option is JSON-serializable, so integrations that pass their configuration as JSON (for example the Docker container or Scalar for Aspire) can load plugins without replacing the whole bundle.
--- a/documentation/configuration.md
+++ b/documentation/configuration.md
@@ -983,6 +983,20 @@ Pass an array of custom plugins that you want. [Read more about plugins here.](p
 }
 ```
 
+#### pluginUrls
+
+**Type:** `string[]`
+
+Pass URLs of ESM modules that export a plugin as their default export. The modules are imported before the API reference mounts and their default exports are registered alongside the plugins passed via `plugins`. Unlike `plugins`, this option is JSON-serializable, so it also works in integrations that pass their configuration as JSON. Only supported by the standalone browser build (`Scalar.createApiReference`). [Read more about plugins here.](plugins.md)
+
+```javascript
+{
+  pluginUrls: [
+    'https://cdn.jsdelivr.net/npm/@example/scalar-plugin/dist/plugin.js',
+  ],
+}
+```
+
 #### proxyUrl
 
 **Type:** `string`
--- a/documentation/plugins.md
+++ b/documentation/plugins.md
@@ -15,6 +15,39 @@ const configuration = {
 }
 ```
 
+## Loading a Plugin from a URL
+
+When you use the standalone browser build (`Scalar.createApiReference`), you can also reference plugins by URL.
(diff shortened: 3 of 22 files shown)


hanspagel (approved):
nice! creative solution

PR #9726, opened 2026-07-20
feat(api-client): switchable form view for JSON and YAML request bodies

bielu:
## Problem

The request body editor only ever shows JSON/YAML bodies as raw text in a code editor, even
though the OpenAPI schema for the body is usually known. Users who just want to fill in a few
fields have to hand-write JSON/YAML — remembering property names, valid enum values, and
required fields themselves — instead of getting a structured, guided editor the way
`multipart/form-data` and `application/x-www-form-urlencoded` bodies already do.

## Solution

Add a **Form / Raw** toggle next to the content-type selector in `RequestBody.vue`. It appears
whenever the selected content type is JSON or YAML (including `+json` / `+yaml` vendor types)
and the body is — or per its schema should be — an object at the root. It's disabled (with a
tooltip) when the current raw text doesn't parse.

The Form view renders one row per schema property, reusing the existing form-table components
(`RequestTable` / `RequestTableRow`) that already power the multipart editor: enum dropdowns,
required badges, defaults, schema examples, and per-field validation come for free. Nested
object schemas flatten into dotted-path rows (`meta.created`), matching the multipart editor's
convention.

Edits fold back into a typed object using the property's declared schema type — numbers,
booleans, arrays, and nested objects survive the round trip instead of being flattened to
strings — then get serialized (JSON or YAML, matching the selected content type) and emitted
through the existing `update:value` event. No new store mutator or event was needed: the store,
raw editor, request sending, and code snippet generation all already consume that event.

```
Form view:                          Raw view:
┌────────────────────────┐         ┌────────────────────────┐
│ name     │ Ada          │         │ {                       │
│ age      │ 36           │   <->   │   "name": "Ada",        │
│ role     │ [admin ▾]    │         │   "age": 36,             │
│ meta.tag │ beta         │         │   "role": "admin",       │
└────────────────────────┘         │   "meta": { "tag": ... } │
                                     │ }                        │
                                     └────────────────────────┘
```

### Design decisions

- **Default is Raw** — no behavior change for existing users; the toggle is purely additive.
- **Storage format unchanged** — the form view emits a serialized string through the existing
  `update:value` event, so no new persistence format or mutator was introduced.
- **Non-object roots** (arrays, primitives) stay raw-only in this iteration — the form view only
  makes sense for object-shaped bodies.
- **Formatting isn't preserved across Form → Raw** — switching back re-serializes the body, so
  original key order / whitespace / YAML comments are not retained. Same trade-off the
  multipart editor already accepts.
- **Undeclared properties** get best-effort JSON-shape coercion (`"5"` → `5`, `"true"` → `true`)
  so untouched values don't silently become strings on their first edit, while schema-declared
  properties always use the schema's type as the source of truth.

### Changes

**`@scalar/workspace-store`:**
- Factored the type-coercion logic that already existed inside the multipart request-body
  builder out into a new shared, exported module: `builder/body/schema-value-coercion.ts`
  (`buildDottedNestedRowPredicate`, `coerceLeafValueToSchemaType`, `resolveLeafSchema`), plus a
  new `coerceUntypedValue` for schema-less properties. Re-exported from
  `@scalar/workspace-store/request-example`. `build-request-body.ts` now imports these instead
  of defining them privately — behavior-neutral refactor, covered by the existing test suite.

**`@scalar/api-client`:**
- `helpers/structured-body-codec.ts` — JSON/YAML parse+stringify pair, selected by content-type
  essence.
- `helpers/get-structured-body-rows.ts` — builds form rows from a parsed body + schema (reuses
  the `collectExampleRows` / `collectLeafProperties` walkers, now exported from
  `get-form-body-rows.ts`).
- `helpers/fold-structured-body-rows.ts` — folds edited rows back into a typed nested object.
- `components/RequestBodyStructured.vue` — the form view component (mirrors
  `RequestBodyForm.vue`); guards against rebuilding rows from its own just-emitted value so
  typing doesn't churn focus and a just-cleared field doesn't reappear.
- `components/RequestBodyViewToggle.vue` — the Form/Raw toggle (same pattern as the existing
  `ResponseBodyToggle.vue`).
- `components/RequestBody.vue` — wires the toggle, parses/validates the raw body for form
  eligibility, and falls back to Raw automatically if the form view stops being available.

**Docs:** new feature bullet in `packages/api-client/README.md`.

### Tests

- `schema-value-coercion.test.ts` (workspace-store): unit tests for the newly-exported
  coercion helpers, including the new `coerceUntypedValue`.
- `get-structured-body-rows.test.ts`, `fold-structured-body-rows.test.ts`: row building and
  fold-back, including **JSON and YAML round-trip stability tests** (parse → rows → fold →
  serialize → parse equals the original, with types preserved).
- `RequestBodyStructured.test.ts`: component tests for row rendering, emitting serialized
  JSON/YAML, and the self-echo guard.
- Existing suites (`build-request-body.test.ts`, `RequestBodyForm.test.ts`,
  `get-form-body-rows.test.ts`, `RequestBody.test.ts`) all pass unchanged, confirming the
  refactor didn't alter multipart/urlencoded behavior.
- `vue-tsc --noEmit` passes clean on `@scalar/api-client`.

## Checklist

- [x] I added a changeset. 
- [x] I added tests.


hanspagel (approved):
sick feature!!

PR #9730, opened 2026-07-20
feat(components): add ScalarDialog native dialog component

hwkr:
> WIP — first piece of the dialog overhaul. Migrating call sites + the command palette are separate follow-ups.

Adds `ScalarDialog`, a new modal built on the native `<dialog>` element + `showModal()`, shipped alongside the existing `ScalarModal`. Going native means the platform handles focus trap/restore, background inert, Escape, top-layer stacking, and nesting for free — so we can eventually drop the HeadlessUI workarounds `ScalarModal` still carries.

- `v-model:open` + default slot, composition-first (no `size`/`variant` — restyle via `class`)
- `watchPostEffect` puppet bridge drives `showModal()`/`close()` and mirrors native closes back to state
- `useBackdropClick` composable for backdrop dismissal (VueUse's `onClickOutside` can't — the `::backdrop` click targets the dialog itself); handles text-selection drags + keyboard clicks
- Storybook stories + 12 unit tests

## Visual

Storybook: `pnpm --filter @scalar/components dev` → **Components/ScalarDialog**. _(screenshots to follow)_

## Ticket

Part of DOC-3107

### DIFF
--- a/.changeset/doc-3107-scalar-dialog.md
+++ b/.changeset/doc-3107-scalar-dialog.md
@@ -1,0 +1,5 @@
+---
+'@scalar/components': minor
+---
+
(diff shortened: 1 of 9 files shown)


cameronrohani (approved):

hwkr:
Closing for now, it will return

PR #9732, opened 2026-07-21
fix(api-reference): render every oneOf/anyOf group of an allOf in place

### LINKED ISSUE #9727: Multiple `oneOf` groups on one object (via `allOf`)
### What happens?

Hey team 👋 I was porting a large legacy API doc into Scalar and hit a wall: when a single object has more than one `oneOf` group (composed with `allOf`), Scalar renders only the **first** group — the others silently disappear (no picker, no fields), even though the document is valid.

Minimal repro:
[Sandbox](https://sandbox.scalar.com/files/j37ME/openapi.json)
​
```
Thing:
  allOf:
    - type: object
      properties: { id: { type: string } }
    - oneOf:                    # group A → renders ✅
        - { title: foo, type: object, required: [foo], properties: { foo: { type: string } } }
        - { title: bar, type: object, required: [bar], properties: { bar: { type: string } } }
    - oneOf:                    # group B → silently dropped ❌
        - { title: baz, type: object, required: [baz], properties: { baz: { type: number } } }
        - { title: qux, type: object, required: [qux], properties: { qux: { type: boolean } } }
```

Only the `foo/bar` picker shows; `baz/qux` never appears. Looks like
`merge-all-of-schemas.ts` keeps just the first `oneOf`/`anyOf`.

each `oneOf` group renders its own picker. It could be also plain "OR" list (will show examples below); request example staying in sync per group.

### Examples 

Here's its working real case scenario, notice multiple OR inside one body:
image 

Current [sandbox](https://sandbox.scalar.com/files/j37ME/openapi.json): 
image

Expected:

Image

### OpenAPI Document

https://sandbox.scalar.com/files/Dz6wn/openapi.json

lavagri:
## What

When an object composes several mutually-exclusive choices as sibling `oneOf`/`anyOf` members under `allOf`, Scalar renders only the **first** group — the others are silently dropped (no selector, no fields), even for a valid document.

This splits the `allOf` into an ordered list of segments and renders each `oneOf`/`anyOf` group as its own selector, **in the position it was declared**, with the surrounding plain fields kept around it. The generated request example is threaded with the per-group composition selection so it stays in sync when you switch any group's variant, and constraint-only members (`not`, `if`/`then`/`else`) no longer blank the example.

Fixes #9727

## Changes

- **`partition-all-of-compositions.ts`** (new) — walk an `allOf` in source order, coalescing runs of object members and emitting each `oneOf`/`anyOf` as its own segment (so no group is dropped and ordering is preserved).
- **`SchemaComposition.vue`** — render those segments in order: a selector per choice group, in place, flowing inline with the surrounding fields.
- **`get-example-from-schema.ts`** (`@scalar/workspace-store`) — thread a per-group composition-selection key through the `allOf` example builder so switching a group updates the generated example; a `null` from a constraint-only member no longer clobbers the merged example.

## Tests

- New unit tests for the partition helper (order preserved, all groups kept, coalescing, nested `allOf`, constraint stripping).
- Updated the #5577 composition test to the new per-group-selector structure.
- `@scalar/api-reference` Schema suite (508) and `@scalar/workspace-store` request-example suite (855) pass; Prettier, Biome, and ESLint clean.

### DIFF
--- a/.changeset/allof-multiple-oneof.md
+++ b/.changeset/allof-multiple-oneof.md
@@ -1,0 +1,6 @@
+---
+"@scalar/api-reference": patch
+"@scalar/workspace-store": patch
+---
+
+Render every `oneOf`/`anyOf` group of an `allOf` in place. Previously, when one object composed several mutually-exclusive choices as sibling `oneOf`/`anyOf` under `allOf`, only the first group was shown and the rest were silently dropped. Each choice group now renders its own selector in the position it was declared, and the generated request example stays in sync per group.
--- a/packages/api-reference/src/components/Content/Schema/SchemaComposition.test.ts
+++ b/packages/api-reference/src/components/Content/Schema/SchemaComposition.test.ts
@@ -444,13 +444,21 @@ describe('SchemaComposition', () => {
       },
     })
 
-    // The merged schema handed to Schema keeps both the base property and the
-    // oneOf variants, so the variant selector still renders.
-    const schemaValue = wrapper.findComponent({ name: 'Schema' }).props('schema') as any
+    // The base object property is rendered by the merged Schema...
+    const baseSchema = wrapper.findComponent({ name: 'Schema' }).props('schema') as any
+    expect(baseSchema.properties).toMatchObject({ customerComment: { type: 'string' } })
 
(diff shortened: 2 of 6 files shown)


hanspagel (approved):
NICE

PR #9733, opened 2026-07-21
fix(openapi-types): versioned type declarations do not resolve with moduleResolution nodenext

### LINKED ISSUE #9731: @scalar/openapi-types does not resolve when typescript module is nodenext
### What happens?

When `module` is `NodeNext`
Image

When `module` is `esnext`
Image

### index.ts
```ts
import type { Document } from "@scalar/openapi-types/3.2";
```

### package.json
```json
{
  "name": "test",
  "version": "1.0.0",
  "type": "module",
  "devDependencies": {
    "typescript": "^7.0.2"
  },
  "dependencies": {
    "@scalar/openapi-types": "^0.9.3"
  }
}
```

### tsconfig.json

```json
{
  "compilerOptions": {
    "target": "es2024",
    "module": "NodeNext",
    "moduleResolution": "NodeNext"
  }
}
```

TypeScript type should resolve in all modules settings.

For example,
https://github.com/scalar/scalar/blob/be8106cd1ebc6dfa589033de0b12bb24bd790304/packages/openapi-types/3.2/index.d.ts#L1-L3
should be
```ts
export type { AuthorizationCodeObject } from './authorization-code.js' 
export type { CallbacksObject } from './callbacks.js' 
export type { ClientCredentialsObject } from './client-credentials.js' 
```
File extension is needed for it to resolve in all module settings.

### OpenAPI Document

_No response_

amritk:
## Problem

Importing the hand-published versioned type declarations (for example `import type { Document } from '@scalar/openapi-types/3.2'`) fails for consumers whose `tsconfig.json` uses `"module": "NodeNext"` / `"moduleResolution": "NodeNext"`. The relative import paths inside the checked-in `2.0`, `3.0`, `3.1` and `3.2` declaration files have no file extensions, and NodeNext/Node16 resolution requires explicit extensions in ESM declaration files:

```
error TS2834: Relative import paths need explicit file extensions in ECMAScript imports
when '--moduleResolution' is 'node16' or 'nodenext'.
```

The types resolved fine with `moduleResolution: bundler`, but broke for anyone on NodeNext.

## Solution

- Added explicit `.js` extensions to every relative import/export specifier in the versioned `.d.ts` files (123 files across `2.0`, `3.0`, `3.1` and `3.2`). TypeScript resolves the `.js` specifiers to the matching `.d.ts` files under every `moduleResolution` setting, so this stays compatible with bundler-mode consumers. The `dist` output was already correct (`tsc-alias` appends extensions there), so only the hand-written declarations needed the change.
- Fixed a strict-mode error the new test surfaced in `ResponsesObject` (`3.0`, `3.1`, `3.2`): the optional `default` property conflicted with the `[statusCode: string]` index signature. Moving `default` into its own intersection member removes the diagnostic without changing what the type accepts. This only affected consumers with `skipLibCheck: false`.
- Added `packages/openapi-types/src/versioned-types.test.ts`, which type-checks all versioned declaration files with the TypeScript compiler API under `module: NodeNext` / `moduleResolution: NodeNext` and asserts there are no diagnostics. Before the fix it reproduced the reported errors for all four versions; it now passes.
- Declared `typescript`, `vitest` and `@types/node` as dev dependencies of the package (the test needs them) and added `"types": ["node"]` to the package `tsconfig.json`, matching sibling packages, so `types:check` passes.

Verified end to end with a standalone project (`"type": "module"`, `module: NodeNext`, `skipLibCheck: false`) importing `Document` from all four versioned entry points — it now compiles cleanly.

## Checklist

- [x] I added a changeset.
- [x] I added tests.

## Ticket

Fixes #9731

https://claude.ai/code/session_01Dj6guzp3dwHbJxAMPRW8d2

---
_Generated by [Claude Code](https://claude.ai/code/session_01Dj6guzp3dwHbJxAMPRW8d2)_


hanspagel (approved):
okay

PR #9736, opened 2026-07-21
ci: move integration publishes off the release critical path

marclave:
## Motivation

Release pushes to `main` take ~18–28 min end to end versus ~13 min for ordinary pushes. Two things stack up on the critical path:

- `release / npm-publish` takes 6–9.5 min. It must run on a GitHub-hosted runner because npm provenance (OIDC) rejects self-hosted runners, so the Blacksmith stickydisk turbo cache in `.github/actions/setup-node` (gated on `startsWith(runner.name, 'blacksmith')`) never applies there and every main push rebuilds the monorepo cold — in run 29537719010 its Build step took 4m40s and its publish step 4m26s.
- The deploy jobs (`deploy-api-client` / `deploy-galaxy` / `deploy-galaxy-production`) have `needs: [release]` and read `release.yml` outputs. A reusable workflow's outputs only become available when **all** of its jobs finish. In run 29778444565 the deploys sat queued from 21:23:25 (npm-publish done) to 21:27:36 waiting for `release / publish-integrations / aspire-publish` — a ~4 min QEMU multi-arch Docker build — even though the deploys already gate on `npm_publish_status` specifically.

The cold npm-publish build is structural; the ~4 min deploy queue time behind the integration fan-out is not.

## Change

Hoist the two slow fan-out invocations out of `release.yml` into `main.yml` as sibling jobs of the deploys:

- `main.yml` gains `publish-integrations` and `todesktop` jobs, both gated on `!cancelled() && github.ref == 'refs/heads/main' && needs.release.outputs.npm_publish_status == 'success'`, with `secrets: inherit`. `publish-integrations` takes the `check-changes` / `integration-builds` outputs directly from `main.yml` instead of routing them through `release.yml` inputs.
- `release.yml` loses those two jobs and its entire `inputs:` block (all 15 inputs existed only to be forwarded to `publish-integrations.yml`). It now finishes right after `npm-publish` plus two small jobs (`create-github-release` ~20s, `push-to-scalar-registry` ~45s), so its outputs — and therefore the deploys — become available ~4 min earlier on release commits.
- The `release` caller job in `main.yml` drops its `with:` block and shrinks `needs` to `[required-jobs-main]`; its permissions block and `secrets: inherit` are unchanged.
- Header comments in `release.yml`, `publish-integrations.yml`, and `todesktop.yml` are updated to describe the new call sites, and the stale gating comments in `main.yml` are rewritten.

## Semantics preserved

- **Publish gating**: `publish-integrations` previously ran iff the `npm-publish` job succeeded (`needs: [npm-publish]`; sibling failures irrelevant). The new gate `npm_publish_status == 'success'` is equivalent — that output records the npm-publish job result via an `always()` step, and a failed or skipped release workflow leaves it `'failure'` or empty, so the publishes still skip. Same for `todesktop`.
- **Artifact flow**: the publish jobs download the `scalar-js` and `mock-server-docker-entrypoint` artifacts uploaded by `main.yml`'s `build` job. Reusable workflows share the caller's run, so the artifacts remain reachable exactly as before.
- **Degraded-input behavior**: a failed or skipped `check-changes` / `integration-builds` previously reached `publish-integrations.yml` as empty strings via `release.yml`'s inputs. The new `needs: [check-changes, integration-builds, release]` plus `!cancelled()` reproduces exactly that — empty outputs mean each publish job sees its package as unchanged or its build-test as not passed and skips itself.
- **Secrets**: both hoisted workflows receive `secrets: inherit` from `main.yml`, which is what they effectively received via `release.yml` (itself called with `secrets: inherit`). The PR-vs-main security boundary is untouched: `main.yml` still only runs on push to `main`, and `pr.yml` references none of these workflows.
- **Permissions**: every job in `publish-integrations.yml` and `todesktop.yml` requests only `contents: read`, which `main.yml`'s workflow-level default already grants to the calling jobs.
- **Outputs contract**: `release.yml`'s outputs (`published`, `scalar_app_released`, `npm_publish_status`) are unchanged, so the deploy gating expressions keep working as-is.

## Risks

- The `release` job no longer waits on `check-changes` / `integration-builds` before starting (it only forwarded their outputs), so `npm-publish` can start slightly earlier than before. Publish ordering is unaffected: `publish-integrations` still needs both jobs plus `release`.
- On the workflow-run UI, integration publishes and the ToDesktop build now appear as top-level `main` jobs instead of nested under `release` — cosmetic, but worth knowing when scanning run pages.
- Validated with Prettier and YAML parsing; `actionlint` was not available in the environment, so the expressions were reviewed manually against the previous wiring.

---

> [!NOTE]
> **Medium Risk**
> Restructures main-branch release and production deploy orchestration; behavior is documented as preserved but a wiring mistake could change when publishes or deploys run.
> 
> **Overview**
> **Moves slow integration registry publishes and the ToDesktop desktop build out of `release.yml` into sibling jobs on `main.yml`**, so web and Galaxy deploys can read `release` workflow outputs as soon as npm publish and the small follow-ups finish, instead of waiting behind multi-arch Docker builds (~4 minutes on release commits).
> 
> The `release` reusable workflow is narrowed to **npm publish**, GitHub release creation, and Scalar registry push; its **15 forwarded `workflow_call` inputs are removed**. `main.yml` now calls `publish-integrations.yml` and `todesktop.yml` directly (still gated on `npm_publish_status == 'success'`, with `check-changes` / `integration-builds` outputs passed in and `!cancelled()` preserving the old degraded-input skip behavior). The `release` caller job depends only on `required-jobs-main` and uses `secrets: inherit`.
> 
> Header comments in the touched workflows are updated to describe the new orchestration; **publish gating, secrets boundary (main-only), and deploy `if` expressions are intended to stay equivalent**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0a3287f6b011fe58c1cc4b5475074d3c621d9295. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -3,9 +3,12 @@ name: Main
 # Entry workflow for pushes to the default branch. Runs the same required tests
 # as pr.yml, then — only if every required test passes — invokes the reusable
 # release.yml (npm publish, GitHub release, Scalar registry) alongside the
-# integration publishes, the ToDesktop build, and the deploys. PRs never run
-# this file: the pull_request / push:main trigger split is the boundary that
-# keeps production publish secrets out of reach of pull requests.
+# integration publishes and the ToDesktop build. The deploys are not gated the
+# same way: they run on every main push — production for release commits,
+# staging otherwise — with their exact conditions documented on each deploy
+# job below. PRs never run this file: the pull_request / push:main trigger
+# split is the boundary that keeps production publish secrets out of reach of
+# pull requests.
 
 on:
   push:
@@ -292,9 +295,11 @@ jobs:
   release:
     # Runs only on main, only when required-jobs-main passed — mirroring
     # npm-publish's old `needs: [required-jobs-main]` + `if: refs/heads/main`.
-    # `!cancelled()` lets release proceed even when non-required jobs failed or
-    # were skipped: a broken change-detection or integration build must NOT
-    # block the npm release.
+    # With required-jobs-main as the only dependency, no other job can block
+    # the npm release. `!cancelled()` is kept purely as a defensive guard so
+    # the explicit `needs.required-jobs-main.result == 'success'` check below
+    # stays the single gate, instead of relying on the implicit success()
+    # GitHub applies when an `if:` contains no status-check function.
     if: |
       !cancelled() &&
       github.ref == 'refs/heads/main' &&
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -2,9 +2,10 @@ name: Main
 
 # Entry workflow for pushes to the default branch. Runs the same required tests
 # as pr.yml, then — only if every required test passes — invokes the reusable
-# release.yml which performs every production publish. PRs never run this file:
-# the pull_request / push:main trigger split is the boundary that keeps
-# production publish secrets out of reach of pull requests.
+# release.yml (npm publish, GitHub release, Scalar registry) alongside the
+# integration publishes, the ToDesktop build, and the deploys. PRs never run
+# this file: the pull_request / push:main trigger split is the boundary that
+# keeps production publish secrets out of reach of pull requests.
 
 on:
   push:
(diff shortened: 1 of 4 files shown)


hanspagel (approved):
LGTM

PR #9751, opened 2026-07-24
fix(api-reference): keep even heading-to-description spacing in narrow layout

hwkr:
Cam flagged that the gap between an operation heading and its markdown description looks uneven in the reference. Turns out the narrow (single-column) operation layout was overriding the title's bottom margin up to `24px`, while wide screens use `12px` — so the spacing jumped as soon as the layout collapsed.

This removes that narrow-only override so the operation title keeps the shared `12px` bottom margin at every width.

## Changes

- Drop the `.operation-title { margin-bottom: 24px }` override from the `narrow-references-container` query in `ModernLayout.vue`. The base rule's `12px` now applies everywhere.

Before / After

CleanShot 2026-07-24 at 14 37 54@2x
CleanShot 2026-07-24 at 14 38 02@2x

## Ticket

Fixes [DOC-5879](https://linear.app/api-docs/issue/DOC-5879)

### DIFF
--- a/.changeset/operation-title-narrow-spacing.md
+++ b/.changeset/operation-title-narrow-spacing.md
@@ -1,0 +1,7 @@
+---
+'@scalar/api-reference': patch
+---
+
+fix: keep even heading-to-description spacing on the operation title in narrow layouts
+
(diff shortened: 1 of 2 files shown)


cameronrohani (approved):

amritk (approved):
cool beans

PR #9757, opened 2026-07-25
fix(api-reference): read x-enum-varnames from array items

### LINKED ISSUE #9722: x-enum-varnames is ignored when the enum is defined in array items
### What happens?

Scalar API Reference does not render `x-enum-varnames` when an enum is defined inside an array schema's `items`.

Environment:

- Scalar.AspNetCore 2.16.10
- OpenAPI 3.0.4

Steps to reproduce:

1. Load the following OpenAPI document in Scalar.
2. Open `GET /example`.
3. Inspect the `statuses` query parameter.

```yaml
openapi: 3.0.4
info:
  title: Array enum metadata reproduction
  version: 1.0.0

paths:
  /example:
    get:
      parameters:
        - name: statuses
          in: query
          schema:
            type: array
            items:
              type: integer
              format: int32
              enum:
                - 1
                - 2
                - 3
              x-enum-varnames:
                - Pending
                - Approved
                - Rejected
      responses:
        "200":
          description: OK
```

The raw OpenAPI document correctly contains `items.x-enum-varnames`, but Scalar renders only:

```text
values
1
2
3
```

The same extension works when `enum` and `x-enum-varnames` are placed directly on a scalar schema.

A possible cause is visible in the current implementation:

- `SchemaEnums.vue` resolves enum values from `value.items`.
- `formatEnumValueWithName` still reads `x-enum-varnames` only from the outer `value`.

As a result, the enum values are found in `items`, but their corresponding metadata is not.

Related work:

- #5578 / #6109 added general `x-enum-varnames` rendering.
- #7682 / #7745 added nested array handling, but does not appear to cover enum metadata stored in `items`.

Scalar should read the enum values and their metadata from the same schema.

For an array whose `items` contains both `enum` and `x-enum-varnames`, the query parameter should be rendered as:

```text
1 = Pending
2 = Approved
3 = Rejected
```

lazerg:
## Problem

`x-enum-varnames` (and `x-enumNames` / `x-enumDescriptions`) are ignored when the enum is defined inside an array schema's `items`. The enum values render, but each one shows up as a bare number instead of `1 = Pending`.

`SchemaEnums.vue` resolves the enum values from `items`, but still reads the varnames and descriptions from the outer schema, so the values and their metadata never line up.

## Solution

Resolve the schema that actually carries the enum once, and read the values, varnames, and descriptions all from it. For an array that means the `items` schema; for a scalar it stays the schema itself.

Added tests covering `x-enum-varnames` and `x-enum-descriptions` defined on array `items`.

## Checklist

- [x] I added a changeset.
- [x] I added tests.

## Ticket

Fixes #9722

### DIFF
--- a/.changeset/array-items-enum-varnames.md
+++ b/.changeset/array-items-enum-varnames.md
@@ -1,0 +1,7 @@
+---
+'@scalar/api-reference': patch
+---
+
+fix: read enum metadata from array items
+
+When an enum is defined inside an array schema's `items`, the enum values were resolved from `items` but their `x-enum-varnames`, `x-enumNames`, and `x-enumDescriptions` were still read from the outer schema, so the metadata was dropped. Both the values and their metadata are now read from the same schema.
--- a/packages/api-reference/src/components/Content/Schema/SchemaEnums.test.ts
+++ b/packages/api-reference/src/components/Content/Schema/SchemaEnums.test.ts
@@ -134,6 +134,24 @@ describe('SchemaEnumValues', () => {
       expect(wrapper.text()).not.toContain(' = ')
     })
 
+    it('formats enum values with x-enum-varnames from array items', () => {
+      const wrapper = mount(SchemaEnumValues, {
+        props: {
+          value: coerceValue(SchemaObjectSchema, {
+            type: 'array',
+            items: {
+              enum: [1, 2, 3],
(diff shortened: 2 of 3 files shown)


hanspagel (approved):
cool, thanks!

PR #9758, opened 2026-07-25
fix(openapi-upgrader): keep schema properties named "example" during 3.0 to 3.1 upgrade

### LINKED ISSUE #9756: `@scalar/openapi-parser`: 3.0 -> 3.1 transform runs on map and payload nodes, corrupting keyword-named keys
### What happens?

The 3.0-> 3.1 upgrade transform is applied to every object node in the document, with no notion of what kind of node it is. So map entries and example payload data get treated as schema keywords.

 Given a schema with a property named example:
```ts
 "properties": {
   "documentation": { "type": "string", "nullable": true },
   "example":       { "type": "string", "nullable": true },
   "platform":      { "type": "string", "nullable": true }
 }
```

 the output becomes:
```ts
 "properties": {
   "documentation": { "type": ["string", "null"] },
   "platform":      { "type": ["string", "null"] },
   "examples":      [ { "type": ["string", "null"] } ]
 }
```

```ts
 "properties": {
   "documentation": { "type": ["string", "null"] },
   "example":       { "type": ["string", "null"] },
   "platform":      { "type": ["string", "null"] }
 }
```

### OpenAPI Document

https://sandbox.scalar.com/e/TIVer

lazerg:
## Problem

The 3.0 → 3.1 upgrade applies its `example` → `examples` conversion to every object node, with no notion of what kind of node it is. So when a schema has a property literally named `example`, that property gets treated as the `example` schema keyword and rewritten into an `examples` array. A `properties` map like `{ documentation, example, platform }` loses its `example` entry, which reappears as `examples: [...]`.

Fixes #9756

## Solution

Skip the conversion when the current node is a `properties` map, since its keys are property names rather than schema keywords. The `example` keyword only lives inside a schema object, never on the map that holds the named subschemas.

## Checklist

- [x] I added a changeset.
- [x] I added tests.

### DIFF
--- a/packages/openapi-upgrader/src/3.0-to-3.1/upgrade-from-three-to-three-one.ts
+++ b/packages/openapi-upgrader/src/3.0-to-3.1/upgrade-from-three-to-three-one.ts
@@ -97,7 +97,11 @@ const applyChangesToDocument = (schema: UnknownObject, path?: string[]) => {
     return false
   })
 
-  if (schema.example !== undefined && !isInsideExamplesMap) {
+  // A node whose own path ends in 'properties' is a map of named subschemas, so a
+  // key called 'example' is a property name and must not be treated as the schema keyword.
+  const isPropertiesMap = path?.[path.length - 1] === 'properties'
+
+  if (schema.example !== undefined && !isInsideExamplesMap && !isPropertiesMap) {
     if (isSchemaPath(path)) {
       schema.examples = [schema.example]
     } else {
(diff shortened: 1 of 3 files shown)


packages/openapi-upgrader/src/3.0-to-3.1/upgrade-from-three-to-three-one.ts
```diff
@@ -97,7 +97,11 @@ const applyChangesToDocument = (schema: UnknownObject, path?: string[]) => {
     return false
   })
 
-  if (schema.example !== undefined && !isInsideExamplesMap) {
+  // A node whose own path ends in 'properties' is a map of named subschemas, so a
+  // key called 'example' is a property name and must not be treated as the schema keyword.
+  const isPropertiesMap = path?.[path.length - 1] === 'properties'
```
daugvinasr: Not sure is the right way to fix this as the deeper issue is still present. Traversal should be fixed as now descends into values it has no business reading. `x-webhooks` still gets renamed. Inside payload `exclusiveMinimum`, `nullable`, `examples` and many others still get mangled.

hanspagel (approved):
right! thanks

PR #9765, opened 2026-07-26
feat(schemas): add Arazzo 1.1 schema

bielu:
## Problem

Scalar renders OpenAPI and (recently) AsyncAPI documents, but has no support for
[Arazzo](https://spec.openapis.org/arazzo/latest.html) — the OpenAPI Initiative's specification for
describing deterministic sequences of API calls ("workflows") across one or more API descriptions.
An API description tells you what the endpoints are; an Arazzo description tells you how to
accomplish something with them. There is currently zero Arazzo support anywhere in the repo.

This is the first of several small PRs landing Arazzo support. It ships the schema only — nothing
downstream consumes it yet — so it can be reviewed independently of the workspace-store and
rendering work that builds on it.

## Solution

Add a hand-written `@scalar/validation` schema for the Arazzo Specification **1.1.0** (17 May
2026), wired through the existing typegen pipeline so `ArazzoDocument` and its constituent types
become available from `@scalar/types/arazzo/1.1`, the same way `AsyncApiDocument` already works.

### Design decisions

- **`@scalar/validation`, not TypeBox.** `packages/schemas` is written against
  `@scalar/validation` (`object`/`union`/`record`/`intersection`/…); TypeBox is a different layer
  used one level up in `packages/workspace-store` and is out of scope here.
- **No official JSON Schema exists.** Both candidate paths in the `OAI/Arazzo-Specification` repo
  return 404, and the `schemas/` directory itself doesn't exist upstream. The schema is
  hand-written from the spec text, the same way the AsyncAPI schema was done.
- **No upgrader package.** Arazzo 1.1.0 is purely additive over 1.0.x (`$self`,
  `$sourceDescriptions.<name>.<ref>`, `$components.`, `$message.`, `querystring`, Selector
  Objects, `"asyncapi"` as a `sourceDescriptions[].type`), so unlike AsyncAPI there's no
  `@scalar/arazzo-upgrader` — a 1.0 document validates unchanged against the 1.1 schema.
- **`Reusable Object` union, not a discriminator.** The three positions that accept either a
  concrete object or a `{ reference, value }` Reusable Object (parameters, success actions,
  failure actions) are modeled as `union([xObject, arazzoReusableObject])`. No discriminator field
  is needed — `coerce`'s structural `scoreUnion` picks the right branch, the same mechanism the
  AsyncAPI schema already relies on for `object | { $ref }`.
- **Mutual exclusivity left to tests, not the schema.** A Step Object's `operationId` /
  `operationPath` / `channelPath` / `workflowId` are mutually exclusive by spec prose only. All
  four stay `optional()` so a document that violates it still renders; each combination is
  asserted independently in tests instead.
- **Arbitrary JSON Schema left untyped.** `workflow.inputs` and `components.inputs` are JSON
  Schema 2020-12 objects — typed as `unknown()` rather than modeled, matching how the AsyncAPI
  schema treats payload schemas.

### Changes

**`packages/schemas/src/arazzo/1.1/`** (new) — one file per Arazzo object:
`arazzo-object.ts` (root document), `info.ts`, `source-description.ts`, `workflow.ts`, `step.ts`,
`parameter.ts`, `criterion.ts`, `success-action.ts`, `failure-action.ts`, `reusable.ts`,
`selector.ts`, `expression-type.ts`, `request-body.ts`, `payload-replacement.ts`,
`components.ts`, and a barrel `index.ts`. The root object intersects the document core with
Scalar's existing document-level extensions (`XScalarNavigation`, `XScalarOriginalSourceUrl`,
`XScalarOriginalDocumentHash`, `XScalarIsDirty`, `XScalarWatchMode`, `XScalarRegistryMeta`).

**Typegen wiring:**
- `packages/schemas/scripts/generated-type-paths.ts` — added
  `packages/types/src/arazzo/1.1/index.generated.ts` to the output/staged path lists.
- `packages/schemas/scripts/generate-types.ts` — added `arazzoObjectSchema` as a fourth top-level
  typegen target (`ArazzoDocument`).
- `packages/types/src/arazzo/1.1/index.ts` (new, hand-written) — barrel re-exporting the generated
  types, mirroring `packages/types/src/asyncapi/3.1/index.ts`.
- `./arazzo/1.1` added to the `exports` map in both `packages/schemas/package.json` and
  `packages/types/package.json`.

### Tests

`arazzo-object.test.ts` plus five per-object test files (157 tests total in `@scalar/schemas`,
all passing):

- Coerces/validates the spec's own pet-purchase example (`loginUserAndRetrievePet` workflow from
  §5.8.1), plus rejection cases for missing `arazzo`/`info`/`sourceDescriptions`/`workflows`.
- Every `criterion.type` flavor: bare `simple`, `regex`, `jsonpath`, `xpath`, and an Expression
  Type Object.
- Each of the four mutually-exclusive step target fields (`operationId`, `operationPath`,
  `channelPath`, `workflowId`) validating independently.
- A Selector Object used in `step.outputs`, alongside the plain runtime-expression-string form.
- A Reusable Object in each of the three positions that accept one (step parameters, `onSuccess`,
  `onFailure`), alongside the concrete object form.

Verified: `pnpm vitest packages/schemas --run` (157/157), `pnpm --filter @scalar/schemas
types:check`, `pnpm --filter @scalar/types types:check` — all clean.

## Checklist

- [x] I added a changeset. 
- [x] I added tests.


hanspagel:
thanks! big arazzo fan here, not sure we want to add it *now* though
let me check with the team

bielu:
I thought it made sense based on that we already support async and openapi. But yeah understand it is additional thing to carry for

bielu:
btw @hanspagel i was looking also into https://spec.openapis.org/overlay/latest.html but yeah decided for me personally arazzo is more helpful, so I am happy to help to get overlays working as well at some point :)

hanspagel:
It hurts, but I’d tend to not work on the Arazzo (or Overlay) support yet. Let's improve the OpenAPI 3.2 and AsyncAPI support first, I don’t want us to neglect those.

Eventually, we'll add Arazzo for sure though. So thanks for taking the initiative here. ✌️

bielu:
Yeah I use all 3 so arazzo would be perfect completion of Documentation, but yeah for now i will use read only ui from OpenApi :). Yeah we also have few prs for asyncapi open, so lets maybe go back to them

Issue #9725, opened 2026-07-20
Accessibility: ARIA violations in rendered api-reference DOM (button-name, aria-allowed-attr, aria-required-children, link-name)


srisxnth:
### What happens?

axe-core (WCAG 2.1 AA) reports the following ARIA violations on the rendered @scalar/api-reference-react DOM. These are not reachable from consumer theming — our integration is config-only (theme:'none' + CSS-variable overrides), so the markup, roles, and ARIA below are all Scalar-generated.

Violations:
- critical - aria-allowed-attr
- critical - aria-required-children
- critical - button-name: collapsible section toggles (button[aria-controls="scalar-refs-…"]) have no accessible name
- serious - link-name: an empty anchor (a[href=""]) with no accessible name
- minor - aria-allowed-role (2 nodes: .t-doc__sidebar, .pr-1)

Environment:
- @scalar/api-reference-react 0.9.58 (reproduced identical on 0.9.54)
- Next.js 15, React 19
- Mount: theme:'none', layout:'modern', withDefaultFonts:false, plus --scalar-* CSS variable overrides. No ARIA/role/markup injected by us.

Reproduce: mount api-reference-react with theme:'none' and run axe-core (WCAG 2.1 AA) against the rendered page.

### What did you expect to happen?

The rendered reference should pass axe-core WCAG 2.1 AA with no serious/critical violations:
- section-toggle buttons have an accessible name (aria-label)
- sidebar rows use valid role/attribute combinations (aria-allowed-attr / aria-allowed-role)
- the client-libraries list satisfies aria-required-children
- the empty anchor has an accessible name or renders as a non-anchor control

These aren't consumer-remediable, so an a11y pass on the reference component's interactive controls would resolve it. Happy to provide the full axe JSON.

Issue #6974, opened 2025-09-30
Schema description is not merged but not overriden


codenomnom:
### What happens?

I have one schema extend another (with a typical `allOf`), but only the first description is shown:

```yaml
components:
  schemas:
    UserBase:
      type: object
      description: Base user information
      properties:
        id:
          type: integer
          description: Unique user ID
        email:
          type: string
          description: User email address
        userName:
          type: string
          description: Username
        name:
          type: string
          description: Display name
        status:
          type: integer
          description: User status
    User:
      allOf:
        - $ref: '#/components/schemas/UserBase'
        - type: object
          description: Full user information
          properties:
            address:
              type: string
              description: Street address
```

Here's the result:

<img width="343" height="568" alt="Image" src="https://github.com/user-attachments/assets/90e492f4-97c2-4f8c-9e0a-d1bee394ec4f" />

I can surely swap the `allOf` order, or put description on top level, but it seems like a merging issue.

### What did you expect to happen?

The description of the second set of `allOf` to override the first one. Swagger works like that.

jvandevelde-conquest:
+1 I've also just recently run into this issue. For now, I'm working around it by  re-ordering the items under `allOf` but am a bit wary that this may impact some tooling like OpenApi Generator. Iirc, there are cases where generators will use the first item in an allof to implement inheritance.

The `title` also seems to behave in a similar way, the first item's title is displayed in place of `object` in the expand details button. Having a title at the schemas root impacts the schemas display name, but having the button show the first `allOf` title can lead to some confusion because everything is listed once it's expanded. Perhaps using the root title would make more sense in this case (ie: `UserDetailsHeader` instead of `UserDetails` or `BaseUser`)

```yml
openapi: 3.0.1
info:
  title: AllOf Display Issue
  version: 1.0
components:
  schemas:
    UserBase:
      type: object
      description: Base user information
      title: BaseUser
      properties:
        id:
          type: integer
          description: Unique user ID
        email:
          type: string
          description: User email address
        userName:
          type: string
          description: Username
        name:
          type: string
          description: Display name
        status:
          type: integer
          description: User status
    User:
      title: UserDetailsHeader
      allOf:
        - $ref: '#/components/schemas/UserBase'
        - type: object
          description: Full user information
          title: UserDetails
          properties:
            address:
              type: string
              description: Street address
```

[...]

PR #9766, opened 2026-07-26
fix(api-reference): name the collapsible section trigger and fix its aria-controls

guillaume-flambard:
## Problem

Every model section in the API reference renders a collapse trigger through `CompactSection`. That button was doing two invalid things at once:

1. It set `aria-controls` to its own `id`. The `id` is the deep link target for the section, so the button was declaring that it controls itself, and the collapsible region it actually controls had no id at all.
2. It exposed no accessible name of its own. The name had to be inferred from its contents, which is a caret icon plus the heading slot plus the screen reader text of the nested copy-link button.

Screen reader users hear an unnamed button and get a control relationship that points nowhere. axe-core (WCAG 2.1 AA) reports `button-name` and `aria-allowed-attr` on every model section, as noted in #9725.

## Solution

`aria-controls` now points at the collapsible region, which gets an id derived from the trigger id. While the section is collapsed the attribute is dropped rather than left pointing at an element that is not rendered.

The accessible name comes from a new optional `triggerLabel` prop, and `Models/ModernLayout` passes the title it actually renders (`schema.title ?? name`), so the name always contains the visible heading.

Attribute changes only — no layout, styling, or behaviour change, so there is nothing to show in a screenshot.

Alternatives considered:

- Reusing the existing `id` for the region and giving the trigger a new one instead. Rejected because `id` is the deep link anchor target that the sidebar scrolls to, so moving it would break navigation.
- Naming the trigger from the `label` prop the component already receives. This is what the first version of this PR did, and it was wrong: models set `label` to the schema key, so a schema whose `title` differs from its key was announced under a name that never appears on screen — a WCAG 2.5.3 (Label in Name) failure, and it broke the `allOf` override e2e test, which locates that button by accessible name. Widening `label` to carry the title instead was also rejected, because `test/snapshots/models.e2e.ts` locates model regions by accessible name.

### Not addressed here

#9725 lists three more violations that need changes I did not want to make unilaterally:

- `aria-required-children` / nested interactive — the trigger `<button>` wraps `Anchor`, which renders the copy-link `ScalarButton`, so there is a button inside a button. Fixing it means either pulling the anchor out of the trigger, which loses click-the-heading-to-toggle, or reworking the copy affordance. Happy to do it in a follow-up once you tell me which direction you prefer.
- `link-name` on the empty `a[href=""]` and `aria-allowed-role` on `.t-doc__sidebar` / `.pr-1` live outside this component.

## Checklist

- [x] I added a changeset.
- [x] I added tests.

## Ticket

Part of #9725

### DIFF
--- a/.changeset/compact-section-trigger-a11y.md
+++ b/.changeset/compact-section-trigger-a11y.md
@@ -1,0 +1,7 @@
+---
+'@scalar/api-reference': patch
+---
+
+fix: give the collapsible section trigger an accessible name and a valid `aria-controls` target
+
+The trigger button rendered by `CompactSection` carried `aria-controls` set to its own `id`, so it declared that it controls itself, and it exposed no accessible name of its own. Screen reader users heard an unnamed button, and axe-core reported `button-name` and `aria-allowed-attr` on every model section. The trigger now takes its name from the section `label` and points `aria-controls` at the collapsible region, which carries its own id. While the section is collapsed the attribute is dropped entirely rather than left pointing at an element that is not rendered.
--- a/.changeset/compact-section-trigger-a11y.md
+++ b/.changeset/compact-section-trigger-a11y.md
@@ -8,4 +8,4 @@ The trigger button rendered by `CompactSection` carried `aria-controls` set to i
 
 The trigger now points `aria-controls` at the collapsible region, which carries its own id. While the section is collapsed the attribute is dropped entirely rather than left pointing at an element that is not rendered.
 
-The accessible name comes from a new optional `triggerLabel` prop. Models pass the title they actually render (`schema.title ?? name`), so the name always contains the visible heading. The section `label` is deliberately not reused here: a model whose `title` differs from its key would have been announced under a name that never appears on screen, which is the WCAG 2.5.3 (Label in Name) failure the previous approach introduced. When `triggerLabel` is omitted, the heading content names the trigger as before.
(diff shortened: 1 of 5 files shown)


guillaume-flambard:
Pushed a fix — the `test-e2e-api-reference` failure on the previous commit was mine, not a flake.

`schema rendering › lets a later allOf member override the description and title (#6974)` timed out on `modelsRegion.getByRole('button', { name: 'UserDetailsHeader' }).click()`. That fixture's schema is keyed `User` but carries `title: 'UserDetailsHeader'`, and the heading renders `schema.title ?? name`. Models were passing `label` (the schema key) into the new `aria-label`, and an `aria-label` overrides the name derived from content — so the button was announced as "User" while the screen said "UserDetailsHeader". The locator stopped matching, and more importantly that is a WCAG 2.5.3 (Label in Name) failure. The first version of this PR traded one violation for another.

The accessible name now comes from a dedicated optional `triggerLabel` prop, and `Models/ModernLayout` passes the same title it renders. I deliberately did not reuse `label` for both: `test/snapshots/models.e2e.ts` locates model regions by accessible name (`User`, `Satellite`, `CelestialBody`), so widening `label` would have moved the breakage rather than fixed it. When `triggerLabel` is omitted the heading content names the trigger, as before.

Added a `Model.test.ts` case mounting a schema whose `title` differs from its key — the exact shape that broke, and one the unit tests did not cover.

Verified locally: unit tests, `vue-tsc`, biome and prettier. I could not run the Playwright suite here — `getDockerServer` uses `--network=host`, which macOS does not support, so the containerised browser cannot reach the ephemeral host port `serveExample` binds to (`ERR_CONNECTION_REFUSED`). Only `lint` has run on this commit; the rest of the workflows are sitting at `action_required`, so the e2e result depends on a maintainer approving the run.

I updated the description above so it matches the current diff, and recorded the `label`-based approach under the rejected alternatives.

hanspagel (approved):
sick

PR #9767, opened 2026-07-26
fix(workspace-store): read x-scalar-secret extensions on apiKey and http schemes

### LINKED ISSUE #9724: x-scalar-secret-token is not used
### What happens?

When using `x-scalar-secret-token` with api_key auth the value should be used to populate the token value. Instead it shows `YOUR_SECRET_TOKEN`. 

Image

That the value provided in `x-scalar-secret-token` is used. 

I modified the rendered URL to show what it should look like
Image

### OpenAPI Document

https://sandbox.scalar.com/e/XA2BY

guillaume-flambard:
## Problem

A secret declared directly on an `apiKey` security scheme is ignored:

```yaml
components:
  securitySchemes:
    apiKeyAuth:
      type: apiKey
      in: query
      name: token
      x-scalar-secret-token: my-token
```

The request keeps rendering the `YOUR_SECRET_TOKEN` placeholder instead of the value from the document (#9724).

The cause is a gap in `extractSecuritySchemeSecrets`. The OAuth flows resolve their secrets from three sources, in order: the auth store, the `x-scalar-secret-*` extension on the document, then the config input field — that is what `mergeFlowSecrets` does with its `configSecrets` argument. The `apiKey` and `http` branches only consult two of them, the auth store and the config field (`value`, `token`, `username`, `password`). Whatever the document declares is never read, so it cannot win over an empty store.

`http` has the same defect as `apiKey`, on all three of its secrets. I fixed both because it is one cause, not two.

## Solution

The `apiKey` and `http` branches now follow the same precedence as the OAuth flows: auth store, then document extension, then config field.

The extension is read through a small `documentSecret` helper that mirrors how `mergeFlowSecrets` reads flow secrets, via an index on the object rather than a declared property. That keeps the strict `ApiKeyObject` / `HttpObject` schema types untouched — the alternative was adding `x-scalar-secret-*` to those schemas, which would turn the extension into published API surface and is a bigger decision than this bug warrants. Happy to switch to that if you would rather have the extensions declared.

Three tests cover it: the document value is used for `apiKey` and for all three `http` secrets, and the auth store still wins over the document.

## Checklist

- [x] I added a changeset.
- [x] I added tests.

## Ticket

Closes #9724

### DIFF
--- a/.changeset/api-key-http-document-secrets.md
+++ b/.changeset/api-key-http-document-secrets.md
@@ -1,0 +1,7 @@
+---
+'@scalar/workspace-store': patch
+---
+
+fix: honour document level `x-scalar-secret-*` extensions on apiKey and http security schemes
+
+OAuth flows already read their secrets from the OpenAPI document, but the apiKey and http branches only looked at the auth store and at the config input fields (`value`, `token`, `username`, `password`). A token declared as `x-scalar-secret-token` directly on the security scheme was therefore ignored, and the request kept rendering the `YOUR_SECRET_TOKEN` placeholder. Both branches now follow the same precedence as the OAuth flows: auth store, then the document extension, then the config field.
--- a/.changeset/api-key-http-document-secrets.md
+++ b/.changeset/api-key-http-document-secrets.md
@@ -2,6 +2,6 @@
 '@scalar/workspace-store': patch
 ---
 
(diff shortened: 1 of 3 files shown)


hanspagel (approved):
nice, thanks!!

PR #9775, opened 2026-07-27
feat(api-client): show request timing waterfall for proxied requests

hanspagel:
WIP

- [ ] needs love
- [ ] check how it's working with SSE

Adds a **Timing** tab to the API client response view that shows how long each network phase took: DNS lookup, connection, TLS, and waiting for the server.

Browsers hide these details for cross-origin requests, so the proxy measures them and sends them back in a `Server-Timing` header. The client reads that header and draws the waterfall. The tab only appears for requests that go through the proxy.

**Preview**

image

### DIFF
--- a/.changeset/proxy-request-timing.md
+++ b/.changeset/proxy-request-timing.md
@@ -1,0 +1,7 @@
+---
+'@scalar/api-client': minor
+---
+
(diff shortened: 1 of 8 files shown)


hanspagel:
Not a priority, maybe later.

PR #9782, opened 2026-07-28
fix(components): tooltip renders dark in light mode

cameronrohani:
image

image

The light-mode tooltip tokens were hard-coded to a dark fill with light text, so the tooltip stayed black no matter which color mode was active. They now derive from --scalar-background-1 and --scalar-color-1 so the tooltip follows the theme in both modes.

Also polishes the tooltip surface to match the rest of the system: a hairline border and base shadow in both color modes (replacing the dark-mode-only inset border), a large radius, regular font weight, and tighter vertical padding. Horizontal and vertical padding are split into separate variables so the placement offset applies to the correct axis.

## Checklist

- [ ] I added a changeset. 
- [ ] I added tests.


marclave (approved):

hwkr (changes requested):
Couple things:
1. We're changing the design? The dark tooltip language matches a lot of other design systems and doing something different might not read as a tooltip
2. The hotkeys look like they need some TLC before updating

hwkr (approved):
Ty!

PR #9783, opened 2026-07-28
fix(api-reference): resolve axe ARIA violations in sidebar and client tabs

guillaume-flambard:
## Problem

axe-core (WCAG 2.1 AA) reports critical/serious ARIA violations on the rendered `@scalar/api-reference` DOM (#9725): invalid `aria-selected` / `role="search"` / `role="navigation"` on `<aside>`, a non-tab control inside `role="tablist"`, and empty `a[href=""]` links for MCP install when no config URL exists.

`button-name` on CompactSection was already fixed in #9766.

## Solution

- Sidebar items use `aria-current="page"` instead of `aria-selected`
- Search trigger is a plain named button (no `role="search"`)
- Sidebar landmark is a `<nav>` (valid navigation role) instead of `<aside role="navigation">`
- Client libraries / SDK installation “More” combobox sits beside the tablist, not inside it
- MCP install controls without a target URL render as buttons

Alternatives considered: keeping `<aside>` and dropping the landmark entirely (worse for screen-reader navigation), or wrapping the More control in `role="presentation"` (still fails required children for interactive content).

## Checklist

- [x] I explained why the change is needed.
- [x] I added a changeset. 
- [x] I added tests.
- [ ] I updated the documentation.

Fixes #9725

Made with [Cursor](https://cursor.com)

PR #9785, opened 2026-07-28
fix(api-reference): resolve axe ARIA violations in sidebar and client tabs

hwkr:
> **Draft — CI validation only.** This duplicates community PR #9783 (by @guillaume-flambard) with three layout regressions fixed on top. It is not meant to merge as-is; the point is a clean CI run before the fixes go back to that PR.

## What's here

`f3f2747`, `3ae5ff2` and `f1d3558` are @guillaume-flambard's commits from #9783, unchanged. `87ba9fe` is a follow-up fixing three regressions the ARIA restructure introduced.

### The a11y fixes (his work)

- Sidebar items use `aria-current="page"` instead of `aria-selected`, which is invalid on link and button roles
- The search trigger drops `role="search"` — a landmark role, not a control role — and keeps its accessible name
- The sidebar `<aside>` drops `role="navigation"`, which `<aside>` does not permit
- The client libraries and SDK installation "More" comboboxes move out of `role="tablist"`, which requires tab-only children
- MCP install controls render as `<button>` when there is no target URL, instead of an `<a>` with no `href`

### The regressions fixed on top

| Where | Problem |
|---|---|
| `ClientDropdown` / `ClientSelector` | `flex: 1` on the "More" wrapper gave it half the row; the five tabs overflowed their box by 138px and painted over it, leaving the trigger unreachable |
| `SdkInstallationInstructions` | `availableWidth` was measured on the tab strip, whose width already excludes the "More" trigger — so that width was reserved twice and a tab that fits got pushed into the dropdown. The `ResizeObserver` also watched an element whose size depended on its own output |
| `OpenMCPButton` | `font: inherit` sat after `font-size` and `line-height` in the same rule; the shorthand resets both, enlarging every MCP link — including the "Generate MCP" `<div>`, which was never a button |

## Testing

Against the containerised Playwright runner, `--workers=1`:

- `test/snapshots` — **78/78** across 12 files, against unmodified baselines
- `test/configuration` — **112 pass**, covering every rewritten `complementary` query
- `test/features` + `test/standalone` — **13 pass**
- `pnpm vitest --run` — **1625 pass** (125 files)
- `vue-tsc`, biome, prettier — clean

## Visual

No screenshots, deliberately: the snapshot suite passing against `main`'s **unmodified** baselines is the stronger claim. This branch fixes ARIA without moving a pixel, so there is no before/after to show.

That result is load-bearing rather than incidental. Before the fixes on `87ba9fe`, `tag-groups › classic` and all 28 `themes › * › Classic *` failed with ~6000px diffs, and `sidebar › Mobile Sidebar` failed separately on the MCP font drift. No baselines were regenerated.

## Ticket

Part of DOC-5912
See #9725
Related to #9783

### DIFF
--- a/.changeset/fix-9725-api-reference-a11y.md
+++ b/.changeset/fix-9725-api-reference-a11y.md
@@ -1,0 +1,10 @@
+---
+'@scalar/api-reference': patch
+'@scalar/components': patch
+---
+
+fix: resolve axe-core ARIA violations in the API reference sidebar and client tabs
+
+Sidebar items used `aria-selected` on links/buttons (invalid for those roles) and the search trigger used `role="search"` on a button. The sidebar landmark is now a `<nav>`, selected items use `aria-current="page"`, and the search control is a plain named button.
+
+Client library and SDK installation "More" comboboxes sat inside `role="tablist"`, which fails `aria-required-children`. They now sit beside the tablist. MCP install controls without a target URL render as buttons instead of empty `a[href=""]` links.
--- a/.changeset/fix-9725-api-reference-a11y.md
+++ b/.changeset/fix-9725-api-reference-a11y.md
@@ -5,6 +5,6 @@
 
 fix: resolve axe-core ARIA violations in the API reference sidebar and client tabs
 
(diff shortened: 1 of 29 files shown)


hwkr:
See https://github.com/scalar/scalar/pull/9783

Issue #9598, opened 2026-06-23
CSS fixes for printing


sanscontext:
### What happens?

Based on [a quick convo on the Discord](https://discord.com/channels/1135330207960678410/1518667156630409286/1518667156630409286).

I'm trying to print a basic reference of our API to PDF to give to a potential customer. I don't need the code snippets etc to render here, but I do need the available parameters, request schema, and ideally the successful response schema.

When I print to PDF, the preview shows the example response schema text overlapping the next entry, some now-unselectable tabs, empty frames, etc.

<img width="1170" height="1164" alt="Image" src="https://github.com/user-attachments/assets/59432304-0316-4f88-9b92-8d6722207ea3" />

### What did you expect to happen?

Expected:
- Any long text wraps instead of printing a picture of the horizontal scroll bar.
- Each tab of the response schema is in its own separate frame, and the content there also wraps.
- Examples shouldn't render over the next page of content :)

hwkr:
Fixed in #9788

PR #9788, opened 2026-07-28
fix(api-reference): stop expanded content overlapping when printing

hwkr:
## Problem

We have no print styles at all, so printing the reference to PDF renders expanded content over the text that follows it. The reference lays itself out as a fixed-viewport app — a `100dvh` grid with sticky columns pinned to a JS-measured screen height — and none of that means anything on paper.

## Solution

Adds a `print.css` that flattens the shell back into a document:

* Un-sticks the example columns, callback titles and list headers. Sticky resolves against the *page* box when printing, so these were getting pinned at a stale `--refs-viewport-offset` and painting over the content after them (`Callback.vue`'s title also carries a `z-index`, which is what made it overlap rather than just sit in the wrong place)
* Zeroes the viewport-derived offsets and drops the `max-height` caps, which were silently truncating long examples to one screen height
* Hides the sidebar, header, footer and `.section-flare` — fixed elements otherwise repeat on every page
* `break-inside: avoid` on small units only (properties, parameters, cards). Deliberately not on whole operations, since anything taller than a page would leave a page-sized gap

Everything in there is `!important`, which looks heavy-handed but is load-bearing: the rules it overrides live in scoped `<style>` blocks that compile to `.examples[data-v-hash]`, so they tie on specificity with a two-class selector and land in a separate stylesheet with no guaranteed load order. It's print-only, so it can't affect screen rendering.

Not a complete fix — it's a solid step, not the whole thing. `Lazy.vue` only renders sections once an `IntersectionObserver` fires and Chrome doesn't re-run those against the paginated layout, so a long document will still print blank placeholder boxes for anything you never scrolled past. CSS can't reach that one, future thing and probably its own ticket.

Before / After

CleanShot 2026-07-28 at 16 11 46@2x-paddedCleanShot 2026-07-28 at 16 12 56@2x

## Ticket

Fixes [DOC-5894](https://linear.app/api-docs/issue/DOC-5894)
See #9598

## Checklist

- [x] I added a changeset.
- [ ] I added tests.

### DIFF
--- a/.changeset/print-styles-sticky-overlap.md
+++ b/.changeset/print-styles-sticky-overlap.md
@@ -1,0 +1,5 @@
+---
+"@scalar/api-reference": patch
+---
+
+Add print styles so printing (or saving to PDF) no longer renders expanded content over the text that follows it. The reference lays itself out as a fixed-viewport application, and its sticky columns were pinned to a screen measurement that is meaningless on paper. Printing now flattens that shell into ordinary document flow: navigation and floating chrome are hidden, sticky positioning and viewport-derived height caps are dropped so long examples are no longer truncated, and small units such as properties and cards avoid breaking across pages.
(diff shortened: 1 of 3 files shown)


hanspagel (approved):
sick improvement for my buddies using a printer

PR #9791, opened 2026-07-29
fix(api-reference): stop recursive discriminator rendering for allOf variants

### LINKED ISSUE #9771: Recursive rendering when using allOf/discriminator
### What happens?

I am using OpenAPI generator and want to follow [this](https://github.com/OpenAPITools/openapi-generator/pull/24174) pattern. The problem is that Scalar then just renders the recursion:

Image

_Version: @scalar/api-reference@1.63.0_

The base schema should render once, showing the discriminator property.

Every child should render all `allOf` properties and ignore the discriminator in the base schema.

### OpenAPI Document

https://sandbox.scalar.com/e/IgFNY

guillaume-flambard:
## Problem

Schemas that use `discriminator.mapping` on a base type with variants that `allOf` back to that base (OpenAPI Generator inheritance) rendered recursively: selecting a child re-inferred the mapping from the merged base and nested another "One of" selector forever.

Repro: https://sandbox.scalar.com/e/IgFNY

## Solution

Thread the parent `discriminator` through `Schema` → `SchemaProperty` → `SchemaComposition`, preferring the schema's own discriminator when present so nested independent polymorphic properties keep their context. The merged base then skips mapping inference and flattens inherited properties once.

## Ticket

Closes #9771

### DIFF
--- a/.changeset/fix-9771-discriminator-allof-recursion.md
+++ b/.changeset/fix-9771-discriminator-allof-recursion.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
+Stop recursive schema rendering when a discriminator variant `allOf`s back to its base type. The selected child now inherits the parent's discriminator context so the mapping is not re-inferred on every nest.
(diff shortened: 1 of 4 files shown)


hanspagel (approved):
thanks!!

PR #9793, opened 2026-07-29
feat(oas-utils): expose server and customFetch in the beforeRequest hook payload

aqeelat:
## What this does

Adds two optional fields to the `ClientPlugin` `beforeRequest` payload:

- `server?: ServerObject | null`: the active server, for resolving relative URLs.
- `customFetch?: typeof fetch`: the host-provided fetch implementation, including the desktop IPC-backed fetch.

`OperationBlock` forwards both values immediately before the request is built, so plugins can use them while mutating the request builder.

## Why

Plugins that inject or refresh OAuth2 bearer tokens need the active server to resolve a relative token endpoint and the host fetch to make that refresh through the same network channel as the outgoing request. Without these values, plugins must capture external state during registration, which is fragile and does not work for every host integration.

## Compatibility

This is additive. Both fields are optional, so existing client plugins continue to work unchanged.

## Verification

- Added an `OperationBlock` regression test proving both values reach a `beforeRequest` plugin.
- Retained the OAS-utils hook payload test.

## Related

Prerequisite for token-injection and auto-refresh client plugins.

### DIFF
--- a/.changeset/before-request-payload.md
+++ b/.changeset/before-request-payload.md
@@ -1,0 +1,9 @@
+---
+'@scalar/oas-utils': patch
+'@scalar/api-client': patch
+---
+
+Expose `server` and `customFetch` in the ClientPlugin `beforeRequest` hook
+payload (both optional, additive) so plugins can resolve relative URLs and run
+network calls — e.g. a token refresh — through the host fetch without closing
+over external state.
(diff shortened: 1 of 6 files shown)


hanspagel (changes requested):
nice, we just need to update the docs too

aqeelat:
I added the direct `beforeRequest` documentation update in a small commit. I also prepared a second, optional commit that adds a complete API Client lifecycle-hook reference so the new fields are documented alongside the other hook payloads. If you would prefer this PR to stay scoped to `server` and `customFetch`, I can drop the second commit.

hanspagel (approved):
cool, thanks

PR #8541, opened 2026-03-21
fix(build): add .js extensions to declaration file imports

marclave:
## Problem

Declaration files generated by `vue-tsc` were missing file extensions in relative imports. This caused TypeScript errors for consumers using `moduleResolution: "node16"` or `"nodenext"`:

```
error TS2834: Relative import paths need explicit file extensions in ECMAScript imports when '--moduleResolution' is 'node16' or 'nodenext'. Consider adding an extension to the import path.
```

This issue blocked Node.js ESM module resolution for consumers of Scalar packages.

## Solution

Added `ts-add-js-extension` to the build process for all Vue packages. This tool runs after `vue-tsc` (and `tsc-alias` where applicable) to add `.js` extensions to all relative imports in declaration files.

**Packages updated:**
- `@scalar/agent-chat`
- `@scalar/api-client`
- `@scalar/api-client-react`
- `@scalar/api-reference`
- `@scalar/api-reference-react`
- `@scalar/components`
- `@scalar/draggable`
- `@scalar/icons`
- `@scalar/openapi-to-markdown`
- `@scalar/pre-post-request-scripts`
- `@scalar/react-renderer`
- `@scalar/sidebar`
- `@scalar/themes`
- `@scalar/use-toasts`

**Additional fixes:**
- Fixed source files in `@scalar/sidebar` and `@scalar/oas-utils` that were using `.ts` extensions in type-only imports

## Validation

- Built all packages with `pnpm build:packages`
- Verified declaration files have proper `.js` extensions using `rg "from ['\"]\\./" packages/*/dist --glob "**/*.d.ts" | rg -v '\.(js|vue)'`
- Passed `pnpm types:check`
- Passed `pnpm lint:check`
- Passed `pnpm format:check`

## Checklist

- [x] I explained why the change is needed.
- [x] I added a changeset.
- [ ] I added tests.
- [ ] I updated the documentation.

[Slack Thread](https://apidocumentation.slack.com/archives/C07SWER07G9/p1774055610799499?thread_ts=1774055610.799499&cid=C07SWER07G9)

[...]

PR #9796, opened 2026-07-30
fix(fastify-api-reference): emit .js extensions in declaration files

### LINKED ISSUE #9795: fastify-api-reference: .d.ts lost .js extensions in 1.62.1, breaking moduleResolution nodenext (regression of #4523)
### What happens?

Since `1.62.1`, the declaration files published by `@scalar/fastify-api-reference` use **extensionless relative re-exports**, which ESM resolution rejects under `moduleResolution: node16`/`nodenext`:

```ts
// node_modules/@scalar/fastify-api-reference/dist/index.d.ts
export { default } from './fastifyApiReference';   // ❌ needs ./fastifyApiReference.js
export type { … } from './types';                  // ❌ needs ./types.js

// dist/fastifyApiReference.d.ts
import type { FastifyApiReferenceOptions } from './types';  // ❌
```

The package is `"type": "module"`, so these need explicit extensions. With `skipLibCheck: false`:

```
node_modules/@scalar/fastify-api-reference/dist/index.d.ts(1,25): error TS2834:
  Relative import paths need explicit file extensions in ECMAScript imports when
  '--moduleResolution' is 'node16' or 'nodenext'.
node_modules/@scalar/fastify-api-reference/dist/index.d.ts(2,110): error TS2834: …
```

This is a **regression** — it was reported as #4523 and fixed by #4528 in Feb 2025. I bisected the published tarballs (`dist/index.d.ts`, first line):

| version | emitted specifier | |
|---|---|---|
| 1.25.116 | `'./fastifyApiReference.ts'` | (the #4528 fix) |
| 1.31.0 – **1.62.0** | `'./fastifyApiReference.js'` | ✅ |
| **1.62.1** – 1.63.0 | `'./fastifyApiReference'` | ❌ |

**Last good: `1.62.0` (2026-06-26). First bad: `1.62.1` (2026-06-30).** Latest (`1.63.0`) is still affected.

`1.62.1` is #9539, which replaced `tsc + copy:standalone` with a Vite/Rolldown lib build so the standalone script is inlined (a good fix for a real bug — not disputing the change). The side effect looks like the declaration emit moving to `vite-plugin-dts`, where the previous `tsc-alias` step was what appended `.js`. Contrast a sibling package with the identical `exports` shape, which still emits correctly:

```ts
// @scalar/client-side-rendering@0.3.2/dist/index.d.ts
export { … } from './html-rendering.js'   // ✅
```

Worth noting why CI didn't catch it: #9539's checks (`types:check`, Biome, knip, package tests) all run inside the monorepo under `moduleResolution: bundler`, which tolerates extensionless specifiers. The failure is only observable from a consumer using `node16`/`nodenext`.

**Why it bites even with `skipLibCheck: true`** (the common case, so this is easy to miss): `skipLibCheck` hides the TS2834 diagnostic but does not repair resolution, so the default export degrades to an error type in consumer code. `tsc` then has nothing to say about passing an error/`any` value into a typed parameter — but type-aware linting does:

```ts
import scalarApiReference from '@scalar/fastify-api-reference';

// typescript-eslint / oxlint: no-unsafe-argument
//   "Unsafe argument of type error typed assigned to a parameter of type
//    FastifyPluginCallback<…>"
app.register(scalarApiReference, { routePrefix: '/docs', configuration: { url: '/openapi.json' } });
```

So consumers silently lose all types for the plugin, and anyone running type-aware lint rules gets a hard failure they can only suppress. Runtime is unaffected — #9539's actual goal works fine; this is type-only.

### Reproduction

```bash
npm i @scalar/fastify-api-reference@1.63.0 fastify typescript
# tsconfig: { "module": "NodeNext", "moduleResolution": "NodeNext", "skipLibCheck": false }
npx tsc --noEmit    # → TS2834 in dist/index.d.ts

# and the last good release, for contrast:
npm i @scalar/fastify-api-reference@1.62.0 && npx tsc --noEmit   # clean
```

`dist/**/*.d.ts` to emit relative specifiers with explicit `.js` extensions, as `1.62.0` and earlier did, so the package's types resolve under `moduleResolution: node16`/`nodenext`.

Two suggestions:

1. Restore `.js` extensions in the declaration output of the Vite build path (`vite-plugin-dts` can be configured to do this, or re-apply the `tsc-alias` step to the emitted `.d.ts`). #8541 proposed a repo-wide version of this and was closed with *"there's a better approach for sure"* — happy to send a PR scoped to `integrations/fastify` only if you'd prefer that shape. Same class of bug was also just fixed for `@scalar/openapi-types` in #9733.
2. Add a regression guard: a tiny consumer fixture that type-checks against the built package under `moduleResolution: nodenext`. Without one this has now shipped twice (#4523, and again here), and in-repo `bundler` resolution structurally cannot catch it.

Related: #4523, #4528, #5859, #8541, #9733.

### OpenAPI Document

N/A — build/type-level issue, independent of the OpenAPI document.

MRocholl:
## Problem

Since `1.62.1`, the published declaration files re-export relative modules without a file extension:

```ts
// dist/index.d.ts
export { default } from './fastifyApiReference';   // ❌
export type { … } from './types';                  // ❌
```

The package is `"type": "module"`, so ESM resolution rejects these under `moduleResolution: node16`/`nodenext` with TS2834. The practical damage is quiet: `skipLibCheck: true` (the common default) hides the diagnostic but does **not** repair resolution, so the default export degrades to an error/`any` type in consumer code. `tsc` has no complaint about passing `any` into a typed parameter — but type-aware lint rules do, and consumers lose all types for the plugin:

```
typescript-eslint / oxlint  no-unsafe-argument
  Unsafe argument of type error typed assigned to a parameter of type FastifyPluginCallback<…>
```

Runtime is unaffected — this is type-only.

Reported as #9795. Previously reported as #4523 and fixed by #4528; the same class of bug was just fixed for `@scalar/openapi-types` in #9733.

## Root cause

`tsc` copies relative specifiers into the emitted `.d.ts` verbatim, so extensionless source imports produce extensionless declarations. `tsc-alias --resolveFullPaths` used to paper over that — but **it only appends `.js` when the target file exists in `outDir`**.

#9539 (a good fix for a real bug — bundling the standalone script so consumers can bundle their app) changed the build to `vite build` + `emitDeclarationOnly: true`. That collapsed the output to a single `dist/index.js`, so the per-module files `tsc-alias` looked for no longer exist and the specifiers stay bare:

| | `dist/*.js` shipped | emitted specifier |
|---|---|---|
| 1.62.0 | `index.js`, `fastifyApiReference.js`, `types.js`, `utils/getJavaScriptFile.js` | `'./fastifyApiReference.js'` ✅ |
| 1.62.1 – 1.63.0 | `index.js` only | `'./fastifyApiReference'` ❌ |

I verified this mechanism in isolation with your exact `tsconfig` + `tsc-alias@1.8.10` pipeline: flipping `emitDeclarationOnly` alone flips the emitted specifier, with the source unchanged.

## Solution

Carry the extensions in source. `tsc` then emits correct declarations regardless of how the bundle is produced — no post-processing step to keep in sync, and nothing added to the toolchain. `moduleResolution: Bundler` resolves `./types.js` → `./types.ts` fine, as does Vite, so the source and the bundling are unaffected.

This is deliberately narrower than #8541, which proposed a repo-wide `ts-add-js-extension` step and was closed with *"there's a better approach for sure"* — scoped to `integrations/fastify` and dependency-free. Happy to reshape it if you'd prefer a different approach.

Added `src/moduleSpecifiers.test.ts` as the regression guard: it asserts every relative specifier in published source carries an extension. This has now shipped broken twice (#4523, and again here) precisely because in-repo checks run under `moduleResolution: Bundler`, which tolerates extensionless specifiers — the failure is only observable from a consumer on `node16`/`nodenext`, so a source-level guard is the cheapest thing that actually holds.

## Verification

Ran in this repo, `pnpm --filter "@scalar/fastify-api-reference..." build`:

```
dist/index.d.ts:               export { default } from './fastifyApiReference.js'  ✅
                               export type { … } from './types.js'                 ✅
dist/fastifyApiReference.d.ts: import type { … } from './types.js'                 ✅
dist/index.js                  3,772,905 bytes  (standalone still inlined ✅)
```

- `pnpm --filter @scalar/fastify-api-reference test` → **50 passed** (46 existing + 4 new)
- `types:check` (`tsgo --noEmit`) clean
- Biome `check` clean on all changed files
- `tsc --moduleResolution nodenext --skipLibCheck false` over the built `dist/index.d.ts` → **0 TS2834**; reverting just the extensions in that emitted file reproduces the original error at the exact same coordinates (`index.d.ts(1,25)`, `(2,110)`)
- Confirmed the new test fails if the extensions are dropped again

## Ticket

Closes #9795


hanspagel (changes requested):
gna! good catch.
I think we had another way to solve this without touching the source files though

MRocholl:
Let me know if you want me to change anything.

MRocholl:
Hey Hans! I think the other way was #4528, which used `tsc-alias --resolveFullPaths`. That flag only adds `.js` when the file exists in `outDir`, and since #9539 the build emits a single `dist/index.js` — so there's nothing left for it to find. I can't see a way to bring it back without giving up the bundling.

There is a source-free option already in the repo though: `integrations/nextjs` builds types with `dts({ insertTypesEntry: true, rollupTypes: true })`. Same thing here would roll everything into one `dist/index.d.ts` with no relative imports at all, so the problem can't come back. Build script gets shorter too:

```diff
- "build": "pnpm types:check && vite build && tsc -p tsconfig.build.json && tsc-alias -p tsconfig.build.json"
+ "build": "pnpm types:check && vite build"
```

One downside: the test I added checks the source specifiers, so it wouldn't make sense anymore — it would have to become a post-build check on `dist/index.d.ts`. Since this has broken twice now I'd like to keep some kind of guard.

Happy to swap it over if you prefer that approach. Just let me know which one you want and I'll push it. Cheers :)

hanspagel:
Thanks a lot for the super detailed writeup, it made this really easy to track down! We went with a slightly different fix in #9835 that keeps the source imports extensionless and bundles the types into a single index.d.ts. Closing this one in favor of that.

PR #9798, opened 2026-07-30
fix(api-client): header params with x-disabled: false stay enabled on…

AshishKumarPathak:
## Problem

Optional header parameters with `x-disabled: false` explicitly set in their OpenAPI example (e.g. `x-scenario-id`) were being reset to enabled while editing, losing their correct initial disabled state.

## Fix

`RequestTableRow`'s `handleUpdateRow` unconditionally ran:

```ts
isDisabled.value = payload.isDisabled ?? false
```

When `CodeInputLite` fires `@update:modelValue` for the name or value field it calls `handleUpdateRow({ name: v })` with no `isDisabled` in the payload, so the `?? false` fallback silently reset the row to enabled on every keystroke, overriding the state read from `x-disabled`.

Now `isDisabled` is only updated when it is explicitly present in the payload:

```ts
if (payload.isDisabled !== undefined) {
  isDisabled.value = payload.isDisabled
}
```

## Notes

An earlier revision also changed the `RequestTable` `v-for` `:key` from `index` to a `name`-based key. That was dropped: local verification could not reproduce the reported GET/DELETE staleness through it (the existing `watch(() => data.isDisabled)` already re-syncs on row reuse, and disabled optional params are never rendered as rows), and a `name`-based key remounts the row on every rename. This PR is now the `handleUpdateRow` fix only.

## Checklist

- [x] I added a changeset.
- [x] I added tests.

### DIFF
--- a/.changeset/fix-header-param-disabled-state.md
+++ b/.changeset/fix-header-param-disabled-state.md
@@ -1,0 +1,7 @@
+---
+'@scalar/api-client': patch
+---
+
+fix: header parameters with `x-disabled: false` (e.g. `x-scenario-id`) now stay enabled while editing
+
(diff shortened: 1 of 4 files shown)


hanspagel (approved):
thanks!

PR #9799, opened 2026-07-30
feat(api-client): show copy button on large request bodies in request panel

AshishKumarPathak:
# feat(api-client): show copy button on large request bodies in the request panel

## Problem

When a request body is very large (e.g. a bulk import payload or a large JSON fixture), the CodeMirror editor renders the full content eagerly. This causes noticeable UI lag and makes it impossible to copy the body without selecting all text manually. The response panel already solved this with `ScalarVirtualCodeBlock` — the request panel had no equivalent.

## Solution

When a raw request body exceeds **20 000 characters**, the CodeMirror editor is replaced with `ScalarVirtualCodeBlock` — the same virtualized renderer with a copy button already used in `ExampleResponse.vue` and `ExampleSchema.vue`. Bodies below the threshold continue to use the full CodeMirror editor unchanged.

The threshold is a single named constant (`VIRTUALIZE_THRESHOLD = 20000`) so it is easy to adjust.

No alternative was seriously considered — reusing the existing `ScalarVirtualCodeBlock` component is the obvious and consistent approach given it was purpose-built for this exact use case in the response panel.

## Files Changed

| File | Change |
|---|---|
| `packages/api-client/src/v2/blocks/request-block/components/RequestBody.vue` | Import `ScalarVirtualCodeBlock`, add `shouldVirtualizeBody` computed, add `v-else-if` branch before `CodeInput` |
| `packages/api-client/src/v2/blocks/request-block/components/RequestBody.test.ts` | Add `virtualized body rendering` test suite (4 tests) |
| `.changeset/request-body-virtualize-copy-button.md` | Patch bump for `@scalar/api-client` |

## Checklist

- [x] I added a changeset.
- [x] I added tests.

### DIFF
--- a/.changeset/fix-header-param-disabled-state.md
+++ b/.changeset/fix-header-param-disabled-state.md
@@ -1,0 +1,11 @@
+---
+'@scalar/api-client': patch
+---
+
+fix: header parameters with `x-disabled: false` (e.g. `x-scenario-id`) now stay enabled on GET and DELETE endpoints
+
+Two bugs caused optional header parameters to appear disabled even when the OpenAPI spec explicitly set `x-disabled: false` on their example:
+
+1. **`RequestTable.vue`** — the `v-for` loop used `:key="index"`, so Vue reused the same `RequestTableRow` component instance when the row list changed length (GET/DELETE have fewer default headers than POST). The stale instance kept its previous `isDisabled` ref value. Fixed to `:key="row.name ? \`${row.name}-${index}\` : index"` so named rows always get a stable, unique key.
+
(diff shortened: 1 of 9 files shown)


AshishKumarPathak:
hi @hanspagel, just wanted to know if any changes required in this PR can you please review this

hanspagel:
Thank you for this, @AshishKumarPathak, and sorry for the slow turnaround. The copy button is a great idea and we want it.

After digging in we decided to go a slightly different way. This version replaced the editor with a read-only `ScalarCodeBlock` above 20k characters, which has two downsides: it makes large request bodies non-editable, and `ScalarCodeBlock` is the non-virtualized renderer (the response panel uses `ScalarVirtualCodeBlock` above that threshold precisely to avoid highlighting huge payloads). Since the request body is meant to be edited, we would rather keep the editor and just add the copy button on top of it.

I have opened #9889 which does that, and credited you there. Closing this one in favour of that, but please keep the contributions coming. Really appreciate you taking the time.

PR #9800, opened 2026-07-30
feat(api-reference): add copy as markdown for LLM button on operations

AshishKumarPathak:
# feat(api-reference): add copy as markdown for LLM button on operations

## Problem

When developers are working with LLMs (ChatGPT, Claude, Copilot, etc.) and want help with a specific API endpoint — writing a fetch call, debugging a request, understanding parameters — they have no quick way to get the operation's full contract into the LLM's context.

The only option today is to manually copy the path, scroll through the parameters, read the request body schema, and paste it all together by hand. This is slow and error-prone, especially for operations with many parameters or complex request bodies.

## Solution

Added a **"Copy as Markdown for LLM"** button (brain icon) on every operation in both the Classic and Modern layouts. Clicking it copies a clean, structured Markdown representation of the operation to the clipboard — ready to paste directly into any LLM chat.

The copied Markdown includes:

- HTTP method + path as a heading (`## POST /pets`)
- Summary and description
- Deprecation notice if applicable
- Parameters table with name, location (`query`/`path`/`header`/`cookie`), type, required flag, and description
- Request body with content type and a properties table (name, type, required, description)
- Responses list with status codes and descriptions

**Example output:**

```markdown
## POST /pets

**Create a pet**

Creates a new pet in the store.

### Parameters

| Name | In | Type | Required | Description |
|------|----|------|----------|-------------|
| `x-api-key` | header | string | Yes | API key |

### Request Body

**Content-Type:** `application/json`

| Name | Type | Required | Description |
|------|------|----------|-------------|
| `name` | string | Yes | Pet name |
| `age` | integer | No | Pet age |
| `tags` | string[] | No | List of tags |
| `status` | "active" \| "inactive" | No | |

### Responses

- `201` — Pet created successfully
- `400` — Invalid input
- `401` — Unauthorized
```

### Why a button on each operation, not a single "copy whole API" button?

LLM prompts work best when scoped to one operation. Dumping an entire API spec would be thousands of tokens and dilute the context. Per-operation copy gives the user exactly what they need for the task at hand.

### Alternative considered

Serializing the raw OpenAPI JSON/YAML for the operation — rejected because LLMs handle prose + tables better than raw spec syntax, and it would expose internal `x-scalar-*` extension fields that are irrelevant to the user's question.

## Files Changed

| File | Change |
|------|--------|
| `packages/api-reference/src/features/Operation/helpers/operation-to-markdown.ts` | **New** — pure `operationToMarkdown()` helper, no Vue dependency |
| `packages/api-reference/src/features/Operation/helpers/operation-to-markdown.test.ts` | **New** — 9 unit tests covering all sections and edge cases |
| `packages/api-reference/src/features/Operation/components/CopyAsMarkdownButton.vue` | **New** — `ScalarIconButton` with brain icon, calls helper + clipboard |
| `packages/api-reference/src/features/Operation/layouts/ClassicLayout.vue` | **Modified** — button added after the existing copy-URL icon in `#actions` |
| `packages/api-reference/src/features/Operation/layouts/ModernLayout.vue` | **Modified** — button added in the top-right badges row |

## Checklist

- [ ] I added a changeset. 
- [x] I added tests.


hanspagel (commented):
Thanks for this! I rebased onto latest `main` and pushed a couple of fixups so CI goes green — all checks are passing now:

- **`fix(api-reference)`** — the `build`/`types` jobs were failing because `SchemaObject` from the workspace store is a strict discriminated union, so `type`/`items`/`properties`/`required` can't be read directly. Narrowed those with `in` checks and resolved ref-able property/items values via `getResolvedRef`. Also gave one test fixture a `type` so the test file type-checks.
- **`test(api-reference)`** — regenerated the visual snapshots that shifted because of the new button (Modern-layout theme screenshots, client, and the Long Strings operation), using the `scalarapi/playwright-runner` image so they match CI.

The feature itself is nice and the helper is well-tested. A few non-blocking notes for a follow-up:

1. **Localize the label.** The neighbouring copy-URL button uses `translate('actions.copyEndpointUrl')`, but this one hard-codes `label="Copy as Markdown for LLM"`. Worth routing through the localization system for consistency.
2. **Top-level `$ref`s.** Inside the helper I resolve refs for property/item schemas, but `operation.parameters`, `requestBody`, and each `response` are still cast rather than resolved. For documents that reference shared components (`parameters: [{ $ref }]`, `requestBody: { $ref }`, `responses: { '200': { $ref } }`) those sections would render empty. Running each through `getResolvedRef` would make it robust.
3. **Table escaping.** A description containing a literal `|` will break the Markdown table row. Newlines are already stripped; escaping `|` (or wrapping in backticks) would round it out. Low stakes for LLM consumption, but cheap to add.
4. **Opt-out.** Consider a configuration flag to hide it (like `hideTestRequestButton`), since it now appears on every operation.

None of these block merge — just ideas. 🧠

AshishKumarPathak:
Thanks for the fixups and for getting CI green!
On the follow-up notes: all four are already implemented in the current branch. The label is routed through translate('actions.copyAsMarkdownForLlm') with translations in all 7 locale files, getResolvedRef is called on parameters, requestBody, and responses (not just property/item schemas), | is escaped via an escapeTableCell() helper, and hideCopyAsMarkdownButton is wired up through the config schema and both layouts. Tests cover the pipe escaping and $ref resolution cases too.

Let me know if anything needs adjusting!

AshishKumarPathak:
Hi @hanspagel, I have made the changes according to the suggestion provided can you please re-review the PR

hanspagel:
thank you so much! needs some design work, I’m gonna look into it next week 🤞

marclave (approved):

AshishKumarPathak:
Hi @hanspagel , the pipeline is failing, which is preventing me from merging the pull request. Could you please help me with this?

hanspagel:
the approval was accidental, but I’m working on it this week ✌️

hanspagel:
It’s coming! I’ll do a fresh attempt.

PR #9804, opened 2026-07-31
fix(api-client): truncate long multipart file names instead of overflowing

### LINKED ISSUE #9803: Multipart Form: Value is not entirely visible
### What happens?

Hello,

In the attached screenshot, the value in the red box is not fully displayed. Is there a way to fix this? Would it be possible to add a horizontal scroll bar somewhere or increase the width of the **Post Analysis Request** section?

Image

The value is fully visible, as shown below. 
Perhaps we should use the `text-overflow: ellipsis` property in the **Select a File** section, along with a tooltip?
In my opinion, it’s more important that the value field be displayed in its entirety than the file field.

Image

### OpenAPI Document

_No response_

lazerg:
### Before

Before: the long file name is clipped on both sides with no ellipsis

### After

After: the file name truncates with an ellipsis

## Problem

Pick a file with a long name in a multipart form and the file column grows to fit the whole name. It runs past the request panel, so the name gets cut off at the edge. There is no ellipsis and no tooltip, so you cannot read the full name.

## Solution

The file column now shares the table width with the other columns instead of growing to its content. When the name is too long it truncates with an ellipsis, and the full name shows in a tooltip on hover.

Closes #9803

### DIFF
--- a/.changeset/multipart-file-name-overflow.md
+++ b/.changeset/multipart-file-name-overflow.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
+
+Long file names in the multipart form no longer stretch the request table past its container. The name now truncates with an ellipsis, and hovering shows the full name.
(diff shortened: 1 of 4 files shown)


hanspagel (approved):
Thanks for the fix!

PR #9815, opened 2026-08-03
feat(api-reference): add Portuguese locale

Joab0:
## Problem

Adds a Portuguese (pt) locale, which wasn't previously available, so Portuguese-speaking users can use the API Reference UI in their native language.

## Solution

Added a pt.ts locale file translating all UI strings, following the same structure as existing locales.

## Checklist

- [ ] I added a changeset. 
- [ ] I added tests.


hanspagel (approved):
cool! thanks for the PR

PR #9817, opened 2026-08-03
fix(workspace-store): resolve relative externalValue URLs

hanspagel:
Example objects can use `externalValue` to point at an external payload instead of embedding it. When that URL was relative (e.g. `/examples/pet.json`), Scalar never fetched it, because the resolver handed the raw relative path to the URL loader, which only accepts absolute URLs.

This resolves the `externalValue` against the document origin first, so relative external examples load the same way relative `$ref`s already do.

To make this possible, the bundler now exposes the current `origin` on the node context for lifecycle plugins.

## How to test

- Add an example with a relative `externalValue` and a document served from a URL.
- The example value now loads.

Part of #9784

---

> [!NOTE]
> **Low Risk**
> Scoped to example `externalValue` fetching and a new optional context field for lifecycle plugins; behavior for absolute URLs is unchanged.
> 
> **Overview**
> Fixes **OpenAPI examples** that point at external payloads via a relative `externalValue` (e.g. `/examples/pet.json`). Those paths were passed straight to URL loaders, which only accept absolute URLs, so the example body never loaded.
> 
> The json-magic bundler now puts **`origin`** on lifecycle **node context** (document URL/path, updated when a subtree has its own `$id`, same idea as `$ref` resolution). **`externalValueResolver`** resolves `externalValue` with **`resolveReferencePath(context.origin, …)`** before validate/fetch/cache, and **`resolveReferencePath`** is re-exported from `@scalar/json-magic/bundle`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 72014b5be09a12df5ca62bab9180fac3bc64e3c9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/external-value-relative-urls.md
+++ b/.changeset/external-value-relative-urls.md
@@ -1,0 +1,6 @@
+---
+'@scalar/json-magic': patch
+'@scalar/workspace-store': patch
+---
+
+Resolve relative `externalValue` URLs on example objects against the document origin, so external request and response examples load even when referenced with a relative path
--- a/packages/json-magic/src/bundle/bundle.test.ts
+++ b/packages/json-magic/src/bundle/bundle.test.ts
@@ -2612,6 +2612,7 @@ describe('bundle', () => {
             },
           },
           loaders: [],
+          origin: '/',
         },
       )
       expect(onBeforeNodeProcess).toHaveBeenNthCalledWith(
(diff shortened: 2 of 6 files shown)


MiroKov:
Could you merge this? We would like to use this in production...

marclave (approved):

PR #9823, opened 2026-08-03
fix(api-client): keep multipart form values readable

### LINKED ISSUE #9803: Multipart Form: Value is not entirely visible

sidsri14:
## Summary

- bound the multipart upload column so long filenames cannot shrink the value field
- give value cells twice the flexible space of key cells
- truncate selected filenames visually while preserving the full name on hover

## Visual

Before ([reported in #9803](https://github.com/scalar/scalar/issues/9803)):

Multipart value obscured before the fix

After (long value and filename at 1400x900):

Multipart value remains readable and the filename truncates after the fix

## Validation

- `pnpm --filter @scalar/api-client types:check`
- focused component tests: 2 files, 61 tests passed
- API client suite across two compatible pools: 94 files / 1,652 tests passed in the thread run; 6 worker-dependent files / 76 tests passed under the default fork pool
- scoped Prettier, Biome, and ESLint checks
- Chromium playground at 1400x900: tracks resolved to `36px 184.5px 369px 104px`; filename ellipsis and title verified

`knip --workspace @scalar/api-client` could not start in the sparse checkout because an unrelated API Reference config imports the omitted `rollup-plugin-webpack-stats` dependency.

Fixes #9803

### DIFF
--- a/.changeset/bright-forms-expand.md
+++ b/.changeset/bright-forms-expand.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
+
+Keep multipart form values readable by bounding the upload column and exposing full filenames on hover.
--- a/packages/api-client/src/v2/blocks/request-block/components/RequestTable.test.ts
+++ b/packages/api-client/src/v2/blocks/request-block/components/RequestTable.test.ts
@@ -189,6 +189,19 @@ describe('RequestTable', () => {
     expect(row.props('showUploadButton')).toBe(true)
   })
 
+  it('reserves more space for multipart values than file controls', () => {
+    const wrapper = mount(RequestTable, {
+      props: {
+        data: [{ name: 'file', value: '', isEnabled: true }],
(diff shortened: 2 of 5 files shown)


hanspagel:
Thank you so much for this, and sorry for the duplicate! This fixes the exact same issue (#9803) as #9804, which we're going with since it keeps the change a little smaller. Really appreciate the effort and the detailed write-up, hope to see more PRs from you! 🙏

PR #9824, opened 2026-08-03
docs(sdks): note the skipped release-title-edit workflow on older SDK repos

amritk:
## Problem

The SDK publishing docs tell maintainers that editing a release pull request's title is how you ship an exact version, and that the `Release PR version` check goes red until the title and the committed version agree.

On every SDK repo generated so far, neither happens. The generated `release-title-edit.yml` guarded on `startsWith(head.ref, 'release-please--branches--scalar-next--')` — the trailing separator of release-please's *multi-component* branch form — but the emitted `release-please-config.json` declares a single root package with no `component`, so release-please renders the componentless `release-please--branches--scalar-next`. The guard matched nothing, both jobs skipped, and:

- retitling pushed no `Release-As` commit, so the pull request was never re-rendered and merging shipped the computed version;
- the `Release PR version` check never reported at all, so nothing blocked that merge.

A skipped job is not a failed one, so this surfaced only as a workflow run with no jobs in it — nothing in the UI reads as broken. A maintainer following the docs exactly gets the wrong version with no signal. Seen in the wild on a customer SDK repo, where a retitle from `release: 0.3.0` to `release: 0.5.0` did nothing.

## Solution

One `> [!IMPORTANT]` callout in `documentation/guides/sdks/publishing/overview.md`, in the existing `## Versioning and releases` section directly under the retitling instructions it qualifies. It covers how to recognise the case (run reports *skipped*, check never appears), why nothing goes red, the fix (rebuild the SDK), and the workaround that still works meanwhile (the `Release-As` commit described in the paragraph above it).

Kept to the existing section rather than a new page — the surrounding prose already documents both the retitle flow and the `Release-As` equivalent, and this is a caveat on them, not a new topic.

## Checklist

- [ ] I added a changeset. 
- [ ] I added tests.

Docs-only prose change under `documentation/`, so no changeset and no tests.

---
_Generated by [Claude Code](https://claude.ai/code/session_01LCjgX5B4vibhrt4MPPbihz)_

### DIFF
--- a/documentation/guides/sdks/publishing/overview.md
+++ b/documentation/guides/sdks/publishing/overview.md
@@ -114,6 +114,9 @@ release: 1.0.0
 
 The `Release PR version` check goes red while the title and the committed version disagree, Scalar re-renders the pull request at your version, and the check goes green. Wait for green before merging. The git-native equivalent is an empty commit on `scalar-next` with a `Release-As: 1.0.0` footer.
 
+> [!IMPORTANT]
+> If retitling appears to do nothing and the `Release PR version` check never reports — the workflow run shows as **skipped**, with neither of its jobs executed — your repository has an early copy of `release-title-edit.yml` whose branch guard never matched the release pull request. Both of its jobs skip on every release pull request, and because a skipped job is not a failed one, nothing turns red to tell you: merging ships the version release-please computed from your commits, not the one in the title. Rebuild the SDK to pick up the corrected workflow. Until you do, set the version with the `Release-As` commit above, which never depended on that workflow.
+
 Every release gets a `vX.Y.Z` tag, a GitHub Release, and an entry in the repository's `CHANGELOG.md`, so your release history lives in your repo next to the code. The generated `VERSIONING.md` documents all of this for your maintainers.
 
 ## Permissions


amritk:
The `types` check (and therefore `required-jobs-pull-requests`) is failing on `@scalar/fastify-api-reference#types:check`. It is not caused by this PR, and I'm not fixing it here — details below.

**What's failing**

`integrations/fastify` fails to type-check against Fastify **5** types while its source targets Fastify **4**:

```
src/fastifyApiReference.ts(28,3):  error TS2353: 'hide' does not exist in type 'FastifySchema'
src/fastifyApiReference.ts(64,3):  error TS2739: 'FastifyTypeProviderDefault' is missing schema, validator, serializer
src/fastifyApiReference.ts(184,7): error TS2578: Unused '@ts-expect-error' directive
src/fastifyApiReference.test.ts(99,30): error TS2769: FastifyPluginCallback not assignable to FastifyPluginAsync
```

**Why it isn't from this PR**

This PR's entire diff is one markdown paragraph in `documentation/guides/sdks/publishing/overview.md`. It contains no TypeScript and does not touch `integrations/fastify`, its dependencies, or the lockfile.

The root cause is a version split already on `main`:

- `integrations/fastify/package.json` → `devDependencies.fastify: "^4.0.0"`
- `pnpm-workspace.yaml` catalog → `fastify: ^5.8.1` (used by the `@fastify/*` packages alongside it)
- `pnpm-lock.yaml` carries `fastify@4.28.0`, `fastify@5.7.4`, and `fastify@5.8.4`

On a full-workspace install the v5 types win for this package, which produces exactly the errors above. The `TS2578` is the clearest tell: line 184's directive is commented *"We're still on Fastify 4, this is introduced in Fastify 5"*, and it only reads as **unused** when v5 types are in scope. I confirmed the pinned `fastify@4.28.0` has no `routerOptions` in its types, so under v4 that directive is required — the type-check is resolving v5.

Worth noting this run reported `51 cached, 120 total`: earlier PRs likely had a cached green result for this task, so a cache miss here surfaced a pre-existing breakage rather than introducing one.

**Why I'm not fixing it**

The fix is a call about what `@scalar/fastify-api-reference` supports — align it onto the v5 catalog and update the source for the v5 type-provider and `FastifySchema` changes, or pin it off the catalog so it resolves v4 consistently. That changes a shipped integration package's supported Fastify range, which is a product decision well outside a docs note, and I'd rather not make it unilaterally inside this PR.

Happy to take it on as a separate PR if someone confirms which direction is wanted.

---

PR #9827, opened 2026-08-04
fix(openapi-upgrader): convert nullable $ref patterns when upgrading to 3.1

### LINKED ISSUE #9694: `@scalar/openapi-parser` upgrade leaves OAS 3.0 nullable `$ref` patterns unconverted in OAS 3.1 output
### What happens?

When upgrading an OpenAPI **3.0** document to **3.1** with `upgrade()` from `@scalar/openapi-parser`, <ins>nullable reference patterns</ins> are not normalized to OAS 3.1 null-union syntax. The upgrader converts inline scalar `nullable: true` correctly, but leaves nullable `$ref` shapes unchanged — including shapes that still contain the removed OAS 3.0 `nullable` keyword after upgrade.

### Nullable reference patterns

This is the pattern recommended in [OpenAPI-Specification#1368](https://github.com/OAI/OpenAPI-Specification/issues/1368) when nullable must be expressed alongside a `$ref` without modifying the referenced schema:

```yaml
pet:
  nullable: true
  allOf:
    - $ref: '#/components/schemas/Pet'
```

### Environment

| Package                                 | Version  |
| --------------------------------------- | -------- |
| `@scalar/openapi-parser`                | `0.28.7` |
| `@scalar/openapi-upgrader` (transitive) | `0.2.9`  |
| `@scalar/openapi-types`                 | `0.9.1`  |

Reproduction uses `upgrade()` from `@scalar/openapi-parser`.

```ts
import { upgrade } from '@scalar/openapi-parser';

const { specification } = upgrade(spec);
```

Input spec:

```yaml
openapi: 3.0.3
info:
  title: Nullable allOf ref upgrade test
  version: 1.0.0
paths: {}
components:
  schemas:
    Pet:
      type: object
      properties:
        id:
          type: integer
    Wrapper:
      type: object
      properties:
        pet:
          nullable: true
          allOf:
            - $ref: '#/components/schemas/Pet'
```

Upgraded output:

```yaml
openapi: 3.1.1
info:
  title: Nullable allOf ref upgrade test
  version: 1.0.0
paths: {}
components:
  schemas:
    Pet:
      type: object
      properties:
        id:
          type: integer
    Wrapper:
      type: object
      properties:
        pet:
          nullable: true
          allOf:
            - $ref: '#/components/schemas/Pet'
```

I expected something like this

```yaml
openapi: 3.1.1
info:
  title: Nullable allOf ref upgrade test
  version: 1.0.0
paths: {}
components:
  schemas:
    Pet:
      type: object
      properties:
        id:
          type: integer
    Wrapper:
      type: object
      properties:
        pet:
          anyOf:
            - $ref: '#/components/schemas/Pet'
            - type: 'null'
```

The nullable reference pattern being converted

before:
```yaml
pet:
  nullable: true
  allOf:
    - $ref: '#/components/schemas/Pet'
```
after:
```yaml
pet:
  anyOf:
    - $ref: '#/components/schemas/Pet'
    - type: 'null'
```

### OpenAPI Document

_No response_

hanspagel:
When upgrading from OpenAPI 3.0 to 3.1, nullable `$ref` shapes were left unchanged. That kept the `nullable` keyword in the output, which does not exist in 3.1, so the `$ref` was no longer nullable.

Now these get rewritten to the 3.1 null-union form:

```yaml
# before
pet:
  nullable: true
  allOf:
    - $ref: '#/components/schemas/Pet'

# after
pet:
  anyOf:
    - $ref: '#/components/schemas/Pet'
    - type: 'null'
```

This covers both `nullable` next to an `allOf` and `nullable` next to a `$ref`. A multi-member `allOf` is kept whole inside the union so no members are lost. Schemas with `type` are untouched.

## Ticket

Fixes #9694

### DIFF
--- a/packages/openapi-upgrader/src/3.0-to-3.1/upgrade-from-three-to-three-one.test.ts
+++ b/packages/openapi-upgrader/src/3.0-to-3.1/upgrade-from-three-to-three-one.test.ts
@@ -139,6 +139,116 @@ describe('upgradeFromThreeToThreeOne', () => {
         },
       })
     })
+
+    it('migrates a nullable `allOf` wrapping a `$ref`', () => {
+      const result: OpenAPIV3_1.Document = upgradeFromThreeToThreeOne({
+        openapi: '3.0.3',
+        info: { title: 'Hello World', version: '1.0.0' },
+        paths: {},
+        components: {
+          schemas: {
+            Pet: { type: 'object', properties: { id: { type: 'integer' } } },
+            Wrapper: {
+              type: 'object',
+              properties: {
+                pet: {
+                  nullable: true,
+                  allOf: [{ $ref: '#/components/schemas/Pet' }],
+                },
+              },
+            },
+          },
+        },
+      })
+
+      expect(result.components?.schemas?.Wrapper?.properties?.pet).toEqual({
+        anyOf: [{ $ref: '#/components/schemas/Pet' }, { type: 'null' }],
+      })
+    })
+
+    it('migrates a nullable `$ref` sibling', () => {
+      const result: OpenAPIV3_1.Document = upgradeFromThreeToThreeOne({
+        openapi: '3.0.3',
+        info: { title: 'Hello World', version: '1.0.0' },
+        paths: {},
+        components: {
+          schemas: {
+            Pet: { type: 'object', properties: { id: { type: 'integer' } } },
+            Wrapper: {
+              type: 'object',
+              properties: {
+                pet: {
+                  $ref: '#/components/schemas/Pet',
+                  nullable: true,
+                },
+              },
+            },
+          },
+        },
+      })
+
+      expect(result.components?.schemas?.Wrapper?.properties?.pet).toEqual({
+        anyOf: [{ $ref: '#/components/schemas/Pet' }, { type: 'null' }],
+      })
+    })
+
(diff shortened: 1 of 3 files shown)


marclave (approved):

packages/openapi-upgrader/src/3.0-to-3.1/upgrade-from-three-to-three-one.test.ts
```diff
@@ -139,6 +139,116 @@ describe('upgradeFromThreeToThreeOne', () => {
...
+        paths: {},
+        components: {
+          schemas: {
+            Pet: { type: 'object', properties: { id: { type: 'integer' } } },
+            Extra: { type: 'object', properties: { tag: { type: 'string' } } },
+            Wrapper: {
+              type: 'object',
+              properties: {
+                pet: {
+                  nullable: true,
```
Youssef1313, line 236: This doesn't sound correct to me. `nullable: true` should have no effect per the spec.

Youssef1313: https://spec.openapis.org/oas/v3.0.3.html#fixed-fields-19

> A true value adds "null" to the allowed type specified by the type keyword, only if type is explicitly defined within the same Schema Object. Other Schema Object constraints retain their defined behavior, and therefore may disallow the use of null as a value.

PR #9831, opened 2026-08-04
docs(docs): document the changelog RSS feed site config

amritk:
Documents `siteConfig.rss`, the changelog RSS feed being added in scalar-org#5395.

## What's covered

- Where the feed is written — `path` plus `rss.xml`, so `/changelog` publishes at `/changelog/rss.xml`
- The properties table, including that `title` defaults to your site title plus "Changelog"
- **Writing entries** — the `## 1.2.0 (2026-07-24)` dated-heading format, that undated subheadings belong to the entry above them, and what happens on an unparseable date
- **Multiple products** — a changelog split across several pages merges newest-first; an index page with no dated headings contributes nothing; hidden pages are skipped
- **Discovery** — the `<link rel="alternate">` tag on every page, and the subscribe button in the page header

## Scope

Documentation only. This adds one section to `site-config.md` and touches nothing else — no changelog or release-notes files, and no change to `scalar.config.json`. Enabling the feed on scalar.com is a separate change.

## Sequencing

**Do not merge before the feature ships.** The `rss` key is not in the published config schema yet, so this documents an option users cannot use until scalar-org#5395 lands and deploys.

### DIFF
--- a/.agents/skills/scalar-docs/SKILL.md
+++ b/.agents/skills/scalar-docs/SKILL.md
@@ -322,6 +322,26 @@ For `scripts` and `styles`: path relative to config root. For `links` (favicon):
 }
 ```
 
+### rss
+
+Publishes an RSS feed for your changelog so readers can subscribe. Written to `<path>/rss.xml` — a changelog at `/changelog` publishes its feed at `/changelog/rss.xml`.
+
+```json
+"rss": {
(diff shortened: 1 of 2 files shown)


hanspagel:
should update the docs skill too

hanspagel (approved):

PR #9833, opened 2026-08-05
Feat/nested sidebar sub tag path

AshishKumarPathak:
# feat(workspace-store): add `x-subTagPath` for unlimited sidebar nesting depth

## Problem

The sidebar currently supports only one level of nesting beneath a tag (via `x-tagGroups`), giving a maximum structure of:

```
Group (section header)
  └── Tag
        └── Endpoint
```

There was no way to add intermediate collapsible groupings between a tag and its endpoints. For API references with many operations under a single tag (e.g. a healthcare payer API with dozens of endpoints across logical sub-categories), this made the sidebar flat and hard to navigate.

## Solution

Added a new OpenAPI extension `x-subTagPath` on operations — an ordered array of strings where each element becomes one collapsible nesting level in the sidebar beneath the parent tag.

The resulting structure supports any number of levels:

```
x-tagGroups group          ← non-collapsible section header
  └── Tag                  ← collapsible
        └── Level 1        ← collapsible  (x-subTagPath[0])
              └── Level 2  ← collapsible  (x-subTagPath[1])
                    └── Level 3  ← collapsible  (x-subTagPath[2])
                          └── Endpoint
```

**How it works:**

- A new `groupBySubTagPath` recursive helper in `traverse-tags.ts` walks the flat list of operations under a tag and builds a nested tag tree based on each operation's `xSubTagPath` array
- At each depth level it buckets operations by their path segment at that depth, recurses into each bucket, and leaves operations with no segment at that depth flat
- Operations without `x-subTagPath` (or with an empty array) render flat directly under their parent tag — fully backward compatible

Alternative considered: hardcoding `x-subTag1`, `x-subTag2`, `x-subTag3` fields — rejected because it requires code changes to add more levels and is harder to use.

## Checklist

- [x] I added a changeset.
- [x] I added tests.

### DIFF
--- a/.changeset/copy-as-markdown-llm-button.md
+++ b/.changeset/copy-as-markdown-llm-button.md
@@ -1,7 +1,15 @@
 ---
 '@scalar/api-reference': minor
+'@scalar/types': minor
+'@scalar/schemas': minor
 ---
 
 feat: add "Copy as Markdown for LLM" button on every operation in both Classic and Modern layouts
 
 Clicking the brain icon copies a structured Markdown snapshot of the operation — method, path, summary, description, parameters table, request body schema, and responses — to the clipboard, ready to paste into any LLM chat.
+
+Follow-up improvements:
+- **Localized label**: button label is now routed through the localization system (`actions.copyAsMarkdownForLlm`) with translations for all 7 built-in locales (en, ar, de, es, fr, ru, zh-CN)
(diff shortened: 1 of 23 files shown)


hanspagel:
thanks for putting this together, I think we’re going with the OpenAPI 3.2 native nested tags, though

PR #9841, opened 2026-08-06
docs(documentation): add user management guide

cameronrohani:
> [!NOTE]
> This PR recreates the changes from #9840 (originally authored by @marclave) on a fresh branch.

## Problem

We have no public documentation on how teams and user management work in Scalar. Support and sales get repeat questions about it: what the roles actually allow, who can invite people, whether inviting adds a billing seat, and — most often — the difference between "team members" (people with dashboard access) and "access groups" (people allowed to read private docs). Right now the only place that answers any of this is the SSO guide, and only in passing.

## Solution

Adds a single guide at `documentation/guides/team/user-management.md`, published under **Resources** next to SSO/SAML since it is team-level rather than tied to one product.

It covers:

- What a team is, and that members/billing/projects are never shared across teams
- The Owner / Admin / Editor roles, with a plain-language table and a permission matrix
- Inviting members, the three-day invite expiry, and cancelling pending invites
- Changing roles, removing members, leaving a team, and the "at least one Owner" rule
- Billing seats on paid plans
- Domain restrictions and SSO for controlling who can join
- Team members vs. access groups, with a pointer to the Private Docs guide

Everything is grounded in current behavior rather than assumed — sourced from the team entity schema and permission map, the core team service (invites, role updates, error cases), and the dashboard team views.

### Notes for the reviewer

Three judgment calls worth a second opinion (carried over from the original PR):

1. **The `Viewer` role is intentionally left out.** It exists in `TeamUserRole`, but both the invite modal and the change-role modal filter it out, so nobody can be assigned it today. Documenting an unassignable role seemed more confusing than helpful. Happy to add a footnote if we would rather be exhaustive.
2. **Billing is documented as Owner *and* Admin.** The role descriptor copy in the dashboard implies billing is Owner-only, but the actual gate on the billing plan grid is the `manageTeam` permission, which Admins have. I documented the real behavior — it may be the UI copy that needs fixing rather than the docs.
3. **No screenshots yet.** The Private Docs guide is screenshot-heavy and this one is not. I can add shots of the Members page and the invite modal if we want the styles to match.

## Checklist

- [ ] I added a changeset. 
- [ ] I added tests. 

### DIFF
--- a/documentation/guides/team/user-management.md
+++ b/documentation/guides/team/user-management.md
@@ -1,0 +1,100 @@
+# User Management
+
+Everything in Scalar belongs to a **team**. Your docs, registry APIs, SDKs, themes and billing all live on the team, and the people you invite to that team get access to them based on their **role**.
+
+This guide covers how to set up your team, invite people, and choose the right role for each of them.
+
+## Teams
+
+A team is the top-level container in Scalar. When you sign up, we create one for you automatically.
+
+Every team has:
+
+- A **name** and a **slug** (the unique identifier used in URLs)
+- A **namespace** used by the [Registry](../registry/getting-started.md)
+- A **billing plan** shared by everyone on the team
+- A list of **members**, each with a role
+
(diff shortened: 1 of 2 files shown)


marclave (approved):

DemonHa (approved):
GH is down lol that's why.. its not marc's branch 😂

PR #9842, opened 2026-08-07
fix(components): show tooltips inside modal dialogs

hwkr:
## Problem

`ScalarTooltip` was invisible when its trigger sat inside a `<dialog>` opened with `showModal()`. Those get promoted to the browser's top layer, which paints above the rest of the document no matter what `z-index` anything else carries — and the tooltip element is a shared singleton parented to `<body>`, so it rendered behind the dialog.

Came up migrating the dashboard's popups from `ScalarModal` to native `<dialog>` (DOC-5872); hits the MCP add-tool flow and the SDK / docs-project repo-link steps.

## Solution

The tooltip now moves into the dialog while it's pointing at a target in there, and moves back to `<body>` once it's hidden.

* Detection is `closest('dialog')` gated on `:modal` — `.open` is also true for `dialog.show()`, which stays in normal flow and already worked
* The move happens *before* the config assignment, since Floating UI watches the target with `flush: 'sync'` and measures the position the moment it lands
* Moving back on hide matters more than it looks: left inside a dialog, the singleton gets removed from the document when that dialog unmounts and never comes back
* Widened the tooltip CSS from `:where(body) >` to `:where(body, dialog) >` — every tooltip style was scoped to being a direct child of `body`, so reparenting alone would've given us a visible but completely unstyled tooltip

Added a Playwright harness for it, with a `MockDialog` to swap out when the modal overhaul lands. One thing to know if you touch those tests: `toBeVisible()` **passes** on the broken tooltip (real 106×32 box, full opacity, just painted underneath), so the assertion is an `elementFromPoint` hit test instead.

## Checklist

- [x] I added a changeset.
- [x] I added tests.

## Ticket

Fixes [DOC-5944](https://linear.app/api-docs/issue/DOC-5944)


hanspagel (approved):
👍 👍 👍 👍 👍

PR #9852, opened 2026-08-07
feat(api-reference): render OpenAPI 3.2 display fields

hanspagel:
Renders three OpenAPI 3.2 display fields:

- Response `summary` above the response description.
- Server `name` in the API reference and API client server pickers. Named options show the name alone, falling back to the URL when the name is absent or empty. The selected URL remains visible in the trigger, and selection still uses the URL.
- A strikethrough on deprecated security schemes.


amritk (commented):
The code here is in good shape — 57 additions, four small independent renderings of 3.2 display fields, each with a test, and all of them degrade cleanly on documents that do not set the new fields (`server.name ? … : server.url`, `v-if="currentResponse?.summary"`, `:class="{ 'line-through': scheme.scheme?.deprecated }"`). Extending the `ScalarCardFooter` `v-if` to include `summary` is the easy thing to miss, and it is there. Nothing about the diff worries me.

I am not approving because this carries the **`on hold`** label. If the hold is stale, drop the label and I will approve; if it is deliberate, this comment is just a note that the code side looks ready whenever the hold lifts.

Two things worth settling while it waits, both about the server label rather than correctness:

**`Production (https://api.example.com)` puts the whole URL in the option label.** For servers with long URLs this will be a wide dropdown, and the URL is often already visible elsewhere in the UI. Worth checking with design whether the name alone, or name with the URL as secondary text, reads better — `ServerDropdown.vue` and `Selector.vue` both do it the same way, so it is one decision in two places. The tests assert the combined string, so changing it later means touching them.

**The description mentions tag `summary` as the section/nav title, but I do not see it in the diff.** The changeset lists only response summaries, server names and the deprecated strikethrough. Either the tag-summary part was dropped and the description needs updating, or it is missing from the branch.

---

amritk (approved):
Small and well scoped. The `server.name || server.url` fallback handles the empty-string name case correctly (tested), and both selectors keep `id: server.url` so selection values are unchanged — only the label moves. The response `summary` is additive and the footer `v-if` was widened to match.

---

PR #9856, opened 2026-08-08
fix(api-reference): render model names as plain text when models are hidden

### LINKED ISSUE #9854: With hideModels: true, model-name links next to types scroll to non-existent section
### What happens?

The model-name links next to types are underlined and have an on-hover color change indicating they can be clicked, but when `hideModels: true` is set in the API reference configuration, the links don't scroll anywhere, causing user confusion and the feeling that something is broken.

When `hideModels: true` is set, the model-name links should render as plain-text to indicate that they are not clickable.

### OpenAPI Document

_No response_

lazerg:
## Problem

With `hideModels: true` there is no models section, but the model names next to types and on the request body heading still render as links. They are underlined, they change color on hover, and clicking one does nothing, so the reference feels broken.

### Before

Before: the Satellite[] model name is underlined and looks clickable

Before: the request body model name Planet is underlined and looks clickable

### After

After: the Satellite[] model name renders as plain text

After: the request body model name Planet renders as plain text

## Solution

The names stay visible, they just lose the link when the models section is hidden. `hideModels` now rides along with the schema options down to the property heading, and the request body heading reads it as well.

## Checklist

- [x] I added a changeset. 
- [x] I added tests.

Closes #9854

### DIFF
--- a/.changeset/fix-9854-hide-models-plain-text.md
+++ b/.changeset/fix-9854-hide-models-plain-text.md
@@ -2,4 +2,4 @@
 '@scalar/api-reference': patch
 ---
 
-Render model names as plain text when `hideModels` is enabled. The names next to types and on the request body heading used to be links, but with the models section hidden there was nothing to scroll to, so clicking them did nothing.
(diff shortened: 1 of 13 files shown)


hanspagel (approved):
nice! I added a few minor things, e.g. to account for individually hidden models

PR #9859, opened 2026-08-10
fix(api-reference): empty popup when hovering Generate MCP without an MCP config

haidargit:
## Problem

Hovering the **Generate MCP** button in the sidebar footer opens an empty popup when no MCP config is set. The VS code/cursor/ Generate MCP rows are meant to fan out on hover, but the container just expands and nothing appears inside it.

https://github.com/scalar/scalar/blob/8cbc745b18f6076934d97e4ef12b20d24a203a7c/packages/api-reference/src/components/AgentScalar/OpenMCPButton.vue#L94-L96

### Steps to reproduce:
1. drop custom/test API to `packages/api-reference/public/`
2. edit index.html
`<script type="module">
    url: '/your_test_api.yaml',
    persistAuth: true,
  })
</script>`
3. build and run

**Clip/Before**
2026-08-10 at 13 54 10

## Solution

Point the three hover fan-out selectors at the shared `.scalar-mcp-layer-link` class. So they apply whether the rows are `<a>` (when the user provides an MCP config) or `<button>` (without one).

Also reduce the hovered container height from `172px` to `97px`. The height of the three rows that render (VS code, cursor, generate/connect). context: at `172px` the hover revealed an empty gap above the top row. e2e test added.

**Clip/After**
2026-08-10 at 16 08 42

If the extra height is intentional, happy to drop this part. ✌🏻 

## Checklist

- [x] I added a changeset. 
- [x] I added tests.

### DIFF
--- a/.changeset/mcp-button-empty-popup.md
+++ b/.changeset/mcp-button-empty-popup.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
+Fix empty popup when hovering the "Generate MCP" button without an MCP config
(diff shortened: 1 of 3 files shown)


hanspagel (approved):
thanks!

PR #9863, opened 2026-08-10
fix(api-reference): stop rendering discriminator base property twice

### LINKED ISSUE #9861: Property get's rendered twice when using `allOf` with `discriminator`
### What happens?

With #9771 the recursive rendering got solved. But now it renders `formatVersion` a second time outside of the one of block.

Image

---

Another visual bug: when the 'oneOf' is directly on the body level the spacing does not look correct.

Image

see: https://sandbox.scalar.com/e/JJGQB

No `formatVersion` outside of the `oneOf` block rendered.

### OpenAPI Document

https://sandbox.scalar.com/e/XgrAT

hanspagel:
Fixes both issues in #9861, which both stem from how a discriminator variant that `allOf`s back to its base is rendered.

## 1. Property rendered twice

When a property's schema is a bare `discriminator.mapping` base (an object with `properties` + `discriminator` but no explicit `oneOf`/`anyOf`), `SchemaProperty` rendered the base properties twice: once as a plain object block, and again inside every inferred `oneOf` variant (the variants `allOf` back to the base, so they already carry those fields). The discriminator field (`formatVersion` in the report) showed up outside the selector and inside it.

`Schema.vue` already renders the inferred composition and the object block mutually exclusively; `SchemaProperty` now does the same, so the property and the model render the base identically. A base that composes itself via its own `allOf` is excluded, since those members get flattened into `properties` and are not present in the variants.

## 2. Body-level `oneOf` spacing

A discriminator variant renders its merged object in a nested card inside the composition panel, which kept its own border/radius and read as a **detached box** floating below the selector. The connecting styles lived in `SchemaProperty`, but a discriminator-inferred `oneOf` is rendered straight from `Schema.vue` (bypassing `SchemaProperty`), so they never applied. Moved the connecting rules to `SchemaComposition` (which owns the panel), removed the now-redundant `SchemaProperty` rule, and dropped the nested card's border so the variant renders flush under the selector — like a plain `oneOf`.

## Verified

Plain `oneOf`, a variant with a nested object property, a variant with a nested `oneOf` property (keeps its own frame), and a variant that renders a description before its content all still render correctly. The nested-card rule keys off an absent property heading (not element position) to stay correct in those cases.

One intentional behavior worth noting: the base is now shown only through its variants, matching `Schema.vue`. A malformed mapping whose variants do not extend the base would surface base-only fields solely via the variants — this matches the existing model rendering and is documented in the code.

## Ticket

Closes #9861

### DIFF
--- a/.changeset/fix-9861-discriminator-property-duplicate.md
+++ b/.changeset/fix-9861-discriminator-property-duplicate.md
@@ -2,4 +2,7 @@
 '@scalar/api-reference': patch
 ---
 
-Stop rendering a discriminator base property twice. When a property's schema is a bare `discriminator.mapping` base (an object with `properties` and a `discriminator` but no explicit `oneOf`/`anyOf`), the inferred variant selector already shows those properties inside each variant (the variants `allOf` back to the base). The base object block is no longer rendered alongside the selector, so properties like the discriminator field are shown once instead of twice.
+Fix two rendering issues with discriminator schemas.
+
+- Stop rendering a discriminator base property twice. When a property's schema is a bare `discriminator.mapping` base (an object with `properties` and a `discriminator` but no explicit `oneOf`/`anyOf`), the inferred variant selector already shows those properties inside each variant (the variants `allOf` back to the base). The base object block is no longer rendered alongside the selector, so properties like the discriminator field are shown once instead of twice.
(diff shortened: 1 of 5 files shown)


hanspagel:
back to draft, not sure this PR is good (yet)

hanspagel:
selectors target schema-renderer DOM classes (e.g. schema-card--level-1, schema-card-description) that main deleted/renamed in the recent tree-layout refactors (#10074)

PR #9867, opened 2026-08-10
fix(api-client): header params showing as empty rows in the client

hanspagel:
In the API client, a header parameter could show up as an empty row (and not get sent) when an earlier operation in the document had a request body.

The row data was correct — the input just never painted its text because the editor component gets reused across rows and the paint step ran before its editor element existed. Now it paints when the element shows up, without overwriting anything you have typed.

## Ticket

Fixes #9797

### DIFF
--- a/packages/api-client/src/v2/components/code-input/CodeInputLite.vue
+++ b/packages/api-client/src/v2/components/code-input/CodeInputLite.vue
@@ -752,6 +752,26 @@ watch(
   },
 )
 
+// The editable surface only exists in "editor mode" (see template). When a row switches into editor
+// mode after mount — most notably when this component instance is reused for a different table row —
+// `onMounted` has already run and the `modelValue` watch fired while `editorRef` was still null, so
+// the model was never painted into the freshly created element and the cell renders empty.
+//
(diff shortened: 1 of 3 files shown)


packages/api-client/src/v2/components/code-input/CodeInputLite.vue
```diff
@@ -752,6 +752,26 @@ watch(
...
+  if (!editor || !isBlankValue(serializeEditor())) {
+    return
+  }
+  const serialized = serializeValue(modelValue)
+  if (serialized === '') {
+    return
+  }
+  lastPillSignature = pillSignature(serialized, withVariables)
+  renderModel(serialized)
+})
```
amritk:
```suggestion
}, { flush: 'post' })
```
can you test this?

amritk (changes requested):

packages/api-client/src/v2/components/code-input/CodeInputLite.vue
amritk: actually let me get an agent to add some tests

amritk (approved):

PR #9345, opened 2026-05-27
feat(api-client, workspace-store): add WebSocket channel operation UI for AsyncAPI documents

DemonHa:
Introduce WebSocket testing experience for AsyncAPI channels in the API client.

image

------

image

------

image

## Checklist

- [x] I explained why the change is needed.
- [x] I added a changeset. 
- [x] I added tests.
- [ ] I updated the documentation.

PR #9882, opened 2026-08-11
feat(api-client): add WebSocket channel operations for AsyncAPI

hanspagel:
Adds basic WebSocket support for AsyncAPI channels in the API client. Pick a channel, connect, send messages, and watch the live message log.

Builds on #9345.


hanspagel:
too much slop, let's keep the branch only

Issue #9883, opened 2026-08-12
Cannot show the document of the dynamicRef correctly


universorum:
### What happens?

Since #9468, Scalar can now generate the example body for entities that use `$dynamicRef`. However, the document cannot be shown correctly.

In paths:
<img width="349" height="602" alt="Image" src="https://github.com/user-attachments/assets/51985e4e-4b2a-4e6b-8275-20a4ea4fd364" />

and schemas:
<img width="215" height="97" alt="Image" src="https://github.com/user-attachments/assets/a8542551-9807-4cd8-aa07-5ab8024debd2" />

### What did you expect to happen?

It should contain the user's definitions, such as properties and descriptions

### OpenAPI Document

https://sandbox.scalar.com/e/c3als

hanspagel:
True! Working on a fix.

hanspagel:
Submitted two PRs: 

* https://github.com/scalar/scalar/pull/9885
* https://github.com/scalar/scalar/pull/9886

hanspagel:
Fixed, merged, released … I think we're done here. :)

PR #9886, opened 2026-08-12
fix(api-reference): render $dynamicRef model items bound to the referenced type

hanspagel:
Same bug in the Models section: a model that binds a template's item type through a `$ref` sibling `$defs` rendered the item empty. It now keeps the binding and shows the bound type.

Part of #9883

## Review follow-up

- The model guard distinguishes unresolved sparse refs while retaining resolved reference wrappers and dynamic-anchor siblings. The openapi-types guard documentation explains why its plain-reference semantics are unsuitable here.
- The binding fixture uses boundItemType with an explanatory comment; model lookup is cached.
- The regression test now checks a description unique to the bound User schema alongside email and the container name. Nested type titles are not rendered in this view, so checking that title would test unsupported presentation; the unique bound description pins the same binding without altering UI behavior.
- Validation: all 22 TraversedEntry tests, api-reference type check, changed-file lint/format, scoped Knip, and changeset status.


amritk (dismissed):
dont like the type cast, can we type guard?

hanspagel:
Good call — replaced the `as SchemaObject` cast with an `isSchemaObject` type guard (same pattern we already use in `get-async-api-message-payload-schema.ts`), so the `SchemaObject | ReferenceObject` union is narrowed instead of asserted. Also rebased on latest `main`.

amritk (commented):
The fix is the right shape — not resolving the `$ref` wrapper away is exactly what keeps the `$dynamicAnchor` sibling `$defs` in scope, and the test reproduces the reported `Paginated<User>` case.

### On reusing `packages/openapi-types/src/helpers/` for `isSchemaObject`

I looked at whether those guards fit here, and the honest answer is **mostly no — and `isDereferenced` in particular would reintroduce this bug.**

`isDereferenced` is the closest match by intent, but it is defined as "not a `ReferenceObject`", implemented as `typeof value.$ref !== 'string'`. `PaginatedUserResponse` in your own test *is* a schema with a sibling `$ref`:

```js
PaginatedUserResponse: {
  $defs: { itemTypeAAA: { $dynamicAnchor: 'itemType', $ref: '...' } },
  $ref: '#/components/schemas/PaginatedTemplate',
}
```

so `isDereferenced` returns `false`, `getModelSchema` returns `undefined`, and the model stops rendering entirely. Swapping it in would turn a rendering bug into a disappearing section. The other guards (`isObjectSchema`, `isArraySchema`, …) discriminate on `type`, which a `$ref` wrapper does not have, so they do not apply either. `isBooleanJsonSchema` is not needed here: `components.schemas` is typed `Union([SchemaObjectRef, reference(SchemaObjectRef)])` in the workspace store and coercion turns a bare `true`/`false` into an object, so a boolean JSON Schema never reaches this code path.

The reusable thing worth doing is the opposite direction: `isDereferenced` lives in `@scalar/openapi-types` and assumes plain OpenAPI semantics, while the workspace store models a reference as `{ $ref, '$ref-value' }` with meaningful siblings. Those two models genuinely disagree, and it would be worth a short note on `isDereferenced` saying it is not safe against workspace-store nodes — otherwise someone will make this substitution later and hit exactly the regression above.

### What I would change in this PR

1. **`isSchemaObject` overstates what it does.** The JSDoc says "we guard the union instead of casting it away", but the body is `isObject(value)` and a `ReferenceObject` is an object, so the union is not actually narrowed — this is a cast with extra steps. Either drop the wrapper and call `isObject` directly at the use site, or make it honest about the one case it really distinguishes, e.g.

   ```ts
   const isSchemaObject = (value: unknown): value is SchemaObject =>
     isObject(value) && (!('$ref' in value) || '$ref-value' in value)
   ```

   which at least screens out an unresolved sparse-chunk `{ $ref }` with no resolved value. That case is a real behaviour change from this PR: previously `getResolvedRef` handed `Model` `undefined` for such a node, and now it receives the bare `{ $ref }`.

2. **`$defs: { itemTypeAAA: ... }`** reads like leftover debugging. I assume the intent is to prove the `$defs` key is irrelevant and only `$dynamicAnchor` matters — if so, name it something like `boundItemType` and say that in a comment, otherwise it just looks like a typo someone will "fix" later.

3. **`getModelSchema(entry.name)` is called twice per model** (once in the `v-else-if`, once for `:schema`). Harmless at this size, but a small `computed` map keyed by entry name would avoid the double lookup and read better.

4. The assertion `expect(wrapper.text()).toContain('email')` is fine as a behavioural check, but it would pass for several wrong renderings too. Asserting the bound type name is present alongside `email` would pin the actual regression more tightly.

None of this blocks the fix — (1) is the only one I would want addressed before merge, and it is a two-line change.

---

amritk (approved):
The fix is the right shape: the bug was `getResolvedRef` resolving the wrapper *away*, which discarded the sibling `$defs`/`$dynamicAnchor` that carry the binding. Passing the wrapper through and letting the shared Schema renderer resolve it keeps the information the renderer needs.

`isSchemaObject` is the part that earns its keep — `!('$ref' in value) || value['$ref-value'] !== undefined` means an unresolved sparse-chunk reference is excluded rather than rendered as an empty model, which is what dropping `getResolvedRef` would otherwise have caused. Good that there is a test for it that actually constructs the sparse state (`'$ref-value' in sparse` true, value `undefined`) instead of mocking it.

The fixture is well chosen: `boundItemType` as the `$defs` key while the anchor is `itemType` proves the binding follows `$dynamicAnchor` and not the key name, which is exactly the thing an implementation could accidentally get right for the wrong reason. Asserting on the unique `Bound user record` description rather than a nested title is the correct call too — the reasoning in your follow-up note (nested type titles are not rendered in this view, so asserting one would be testing unsupported presentation) is right.

Adding the "do not use this guard for workspace-store nodes" warning to `is-dereferenced.ts` is a good defensive touch, since that helper's plain-reference semantics are precisely what would reintroduce this bug.

`required-jobs-pull-requests` green on `ab8a6f0`. Changeset present.

Approving.

One thing to keep an eye on rather than change here: dropping `getResolvedRef` affects *every* model, not just dynamic-ref ones. In practice `components.schemas[name]` is usually an inline schema so this is a no-op, but a pure alias (`Foo: { $ref: '#/components/schemas/Bar' }`) now reaches `Model` as the wrapper rather than the resolved target. The two tests here cover the dynamic-ref and sparse cases but not that one — worth a quick look at an alias model in the playground before merge, or a third test case if it is cheap.

---

PR #9889, opened 2026-08-12
feat(api-client): add copy button to the request body editor

hanspagel:
The raw request body editor had no way to copy the whole body without selecting the text by hand. This adds the same hover copy button the response panel already uses (`ScalarCodeBlockCopy`), so you can grab the body in one click. The editor stays fully editable at any size.

Supersedes #9799 (thanks @AshishKumarPathak for kicking this off). That version replaced the editor with a read-only `ScalarCodeBlock` above 20k characters, which made large bodies non-editable and relied on the non-virtualized `ScalarCodeBlock`. This keeps the editor and just adds the button.

## Rebase and validation

Rebased on main `94091b585f`. Resolved the RequestBody test conflict by retaining both main's generate-example coverage and this PR's clipboard tests. There were no submitted reviews or review threads to address; the rebase adds no new runtime UI behavior beyond the original PR.

- All 24 RequestBody tests passed under Node 24, including real clipboard behavior and the empty-body case.
- Changed-file Biome, Vue ESLint, formatting, scoped Knip, changeset status, and diff whitespace checks passed.
- Full API Client type checking is blocked by stale installed workspace declarations, including `executeResponseHook`, `getUniqueExampleName`, and QUERY method types. Tests ran with temporary source aliases, which are not committed.
- A fresh browser screenshot was attempted using the real RequestBody component, but the temporary preview could not load linked Vue dependencies. No fresh visual artifact is claimed; hover positioning still needs visual confirmation with the current workspace build.


amritk (approved):
Additive UI change, contained to `RequestBody.vue`. Hoisting the language into `selectedLanguage` removes the duplicated inline lookup, and gating the button on `bodyValue` keeps it out of the way for an empty editor. Tests cover both the copy path and the hidden-when-empty case.

---

PR #9892, opened 2026-08-13
fix(helpers): expand a reference that is used more than once

### LINKED ISSUE #9890: Show Schema renders [Circular] when a ref is used more than once
### What happens?

Given a schema like

```json
{
  "components": {
    "parameters": {},
    "schemas": {
      "ID": {
        "example": "12345678-1234-1234-1234-123456789012",
        "type": "string"
      },
      "UserRecord": {
        "properties": {
          "Id": { "$ref": "#/components/schemas/ID" },
          "ClientId": { "$ref": "#/components/schemas/ID" },
          "Name": {
            "description": "Name",
            "type": "string"
          }
        },
        "required": [],
        "type": "object"
      }
    }
  },
  "openapi": "3.0.2",
  "paths": {
    "/users/{user_id}": {
      "get": {
        "parameters": [{ "user_id": { "name": "user_id", "type": "string" } }],
        "responses": {
          "200": {
            "content": {
              "application/json": {
                "schema": { "$ref": "#/components/schemas/UserRecord" }
              }
            },
            "description": "Success."
          }
        },
        "summary": "Get user"
      }
    }
  }
}

```

Note how `UserRecord` has
```json
          "Id": { "$ref": "#/components/schemas/ID" },
          "ClientId": { "$ref": "#/components/schemas/ID" },
```

In this case the docs will render `[Circular]` after the first reference.

Image

Render the schema e.g. 
```json
{
  "properties": {
    "Id": {
      "type": "string",
      "examples": [
        "12345678-1234-1234-1234-123456789012"
      ]
    },
    "ClientId":  {
      "type": "string",
      "examples": [
        "12345678-1234-1234-1234-123456789012"
      ]
    },
    "Name": {
      "description": "Name",
      "type": "string"
    }
  },
  "required": [],
  "type": "object"
}
```

### OpenAPI Document

_No response_

lazerg:
## Problem

When two properties point at the same `$ref`, only the first one shows up in "Show Schema". Every one after that renders as `[Circular]`, even though nothing is circular.

The deep-resolved schema hands the same object to both properties, and pretty-printing collapsed *any* repeated object reference, not just the ones that form a cycle.

### Before

Before: ClientId renders as [Circular]

### After

After: ClientId renders the full ID schema

## Solution

Only collapse a reference that is already on the current path, so a shared type is shown in full everywhere it is used.

The repeated-reference collapsing is still there as a fallback, since that is what keeps #9377 from coming back: a deeply shared graph expands exponentially and freezes the tab (one real schema hit ~244 MB). So we first count the nodes a full expansion would emit, and only fall back to collapsing when that count gets out of hand. The count is memoized, so the walk stays cheap even when the number is astronomical.

## Checklist

- [x] I added a changeset. 
- [x] I added tests.

Closes #9890

### DIFF
--- a/.changeset/pretty-print-repeated-references.md
+++ b/.changeset/pretty-print-repeated-references.md
@@ -1,0 +1,5 @@
+---
+'@scalar/helpers': patch
+---
+
+Show a schema type that is referenced more than once in full, instead of rendering `[Circular]` after the first occurrence. Pretty-printing used to collapse every repeated object reference, which also hit two sibling properties pointing at the same `$ref`. Repeated references are now expanded, and only true circular references are collapsed. Very large shared graphs still fall back to collapsing repeated references, so "Show Schema" does not freeze the tab.
--- a/packages/helpers/src/json/pretty-print-json.test.ts
+++ b/packages/helpers/src/json/pretty-print-json.test.ts
@@ -99,4 +99,32 @@ describe('prettyPrintJson', () => {
     expect(result.length).toBeLessThan(10_000)
     expect(result).toContain('[Circular]')
   })
+
+  it('does not explode on a graph that mixes real cycles with heavy sharing', () => {
+    /*
+     * Measuring the expanded size up front can under-count this shape: a node first reached through a
+     * cycle is measured with its back-edge cut, yet it expands in full wherever the cycle is absent, so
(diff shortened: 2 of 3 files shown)


hanspagel (approved):
good catch! thanks for the fix

PR #9893, opened 2026-08-13
feat(api-reference): show schema pattern as hover dropdown

AshishKumarPathak:
# feat(api-reference): show schema `pattern` as hover dropdown

## Problem

Long regex patterns in API schema properties were being truncated inline in the API reference UI, making them unreadable. There was no way for users to see the full pattern value without inspecting the raw spec.

## Solution

Replaced the inline truncated `pattern` display with a hover dropdown — consistent with how `examples` are already handled via `SchemaPropertyExamples.vue`.

- Added a new `SchemaPropertyPattern.vue` component that renders a "Pattern" link button. On hover/focus, an absolute-positioned popup appears showing the full regex in a `<code>` block (long patterns wrap so they stay fully readable) with a clipboard copy button.
- Removed `pattern` from the inline `validationProperties` list in `SchemaPropertyHeading.vue` and replaced it with `<SchemaPropertyPattern>`.
- Added unit tests (`SchemaPropertyPattern.test.ts`) and updated `SchemaPropertyHeading.test.ts` to match the new hover behavior.
- Added Playwright e2e snapshot tests (`pattern.e2e.ts`) covering: pattern button visibility, popup on hover, long complex regex, and full request body snapshot.
- Fixed `serve-example.ts` to use `fileURLToPath` instead of `.pathname` to handle URL-encoded spaces in directory paths.

## Review follow-ups

- Added an invisible hover "bridge" so the pointer can reach the popup without dropping `:hover` (matches `SchemaPropertyExamples`).
- Capped the popup width at `min(500px, calc(100vw - 32px))` so it never overflows a narrow viewport (unchanged on the wide e2e viewport, so snapshots stay valid).
- Gave the copy button an `aria-label` (new `common.copyPattern` translation) and marked the clipboard icon `aria-hidden`, so screen readers announce the copy action.

## Checklist

- [x] I added a changeset. 
- [x] I added tests.


AshishKumarPathak:
The snapshots in pattern.e2e.ts.snapshots/ were generated on macOS and will fail on Linux CI due to font/rendering differences. Please regenerate them by running pnpm test:e2e --update-snapshots -- test/snapshots/pattern.e2e.ts inside a Linux environment or via Docker.

AshishKumarPathak:
hi @hanspagel can you please review this PR

hanspagel (approved):
nice, thanks!

PR #9896, opened 2026-08-13
fix(api-client): parameter name blanks in Try It panel on first open

AshishKumarPathak:
# fix(api-client): parameter name blanks in Try It panel on first open

## Problem

When opening the Try It panel for a GET endpoint that has a header parameter with `examples` defined (e.g. `x-scenario-id`), the parameter name appears blank on the **first** open. After clicking Send, closing, and reopening the panel, the name shows correctly.

The bug does **not** reproduce when Scalar loads the spec via `url:` — it only affects the `spec: { content: ... }` path (inline pre-fetched spec object), which goes through a different initialization sequence in the workspace store and triggers an extra reactive re-render cycle.

Root cause was three compounding issues:

1. **`RequestTable` used `key: index`** — Vue's virtual DOM diffing reuses component instances when only the data changes. `displayData` always appends a placeholder row `{ name: '', value: '' }` at the end of the list. When the list recomputed (e.g. after the workspace store initialized the parameter's example value), the array length or order shifted, and Vue mapped the existing `RequestTableRow` instance for `x-scenario-id` onto the placeholder row — passing `data.name = ''` as a prop update.

2. **`RequestTableRow` watch and blur forwarded the empty name** — the `watch:name` handler unconditionally synced the local `name` ref to the incoming prop (including `''`), and `handleKeyBlur` forwarded a blank name emitted by `CodeInputLite` before it had rendered its initial value into the contenteditable div. Both paths called `emit('upsertRow', { name: '' })`.

3. **`upsertOperationParameter` mutated `param.name` unconditionally** — the store mutator did `param.name = payload.name` without checking whether the payload name was empty. Because `param` is the live reactive spec object, setting `param.name = ''` immediately caused `createParameterRows` to recompute and produce a row with `name: ''`, which flowed back down as a prop — completing the feedback loop.

## Solution

**Fix 1 — stable `key` in `RequestTable.vue`** (primary fix):

Replace `key: index` with a key derived from the parameter's identity (`originalParameter.name + sourceParameterValuePath`). This ensures Vue never reuses a `RequestTableRow` instance for a different parameter or the placeholder row, eliminating the prop-swap that started the chain.

**Fix 2 — guards in `RequestTableRow.vue`**:

- `watch:name`: skip the local ref update when the incoming prop is `''` but the current local name is valid — prevents a transient empty prop from blanking the display.
- `handleKeyBlur`: skip emitting `upsertRow` when `newName` is `''` but `data.name` is valid — prevents a stale `CodeInputLite` blur from propagating an empty name into the store.

**Fix 3 — guard in `upsertOperationParameter` (`parameters.ts`)**:

Skip `param.name = payload.name` when `payload.name` is `''` and the parameter already has a non-empty name. A value-only update should never blank the parameter name.

Fixes 2 and 3 are defensive layers. Fix 1 is the structural root cause fix.

## Checklist

- [x] I added a changeset.
- [x] I added tests.


AshishKumarPathak:
hi @hanspagel  can you please review this PR

hanspagel (approved):
cool, thanks!

PR #9898, opened 2026-08-13
fix(json-magic): treat array/object container mismatches as updates in diff

geoffgscott:
## Problem

`diff(a, b)` from `@scalar/json-magic/diff` corrupts documents when the same key holds a plain object on one side and an array on the other. Both containers satisfy `typeof === 'object'`, so the comparison recurses key-by-key and walks the array's numeric indices as object keys:

```js
diff({ tags: {} }, { tags: ['x', 'y'] })
// [
//   { path: ['tags', '0'], changes: 'x', type: 'add' },
//   { path: ['tags', '1'], changes: 'y', type: 'add' }
// ]
```

`apply` then writes those numeric string properties onto the existing object and produces `{ tags: { '0': 'x', '1': 'y' } }` — a corrupted hybrid that is neither side. The reverse direction (array → object) is broken the same way: it emits per-index deletes plus per-key adds, and applying them splices the array instead of replacing it.

## Solution

A container type change is not a set of per-key edits — it is one change. When one side is an array and the other a plain object (either direction), `diff` now emits a single `update` difference for that path carrying the full new value, the same shape it already uses for scalar type changes, instead of descending. `apply` already handles that update by assignment, so the container is replaced correctly:

```js
diff({ tags: {} }, { tags: ['x', 'y'] })
// [{ path: ['tags'], changes: ['x', 'y'], type: 'update' }]

apply({ tags: {} }, ...)
// { tags: ['x', 'y'] }   Array.isArray → true
```

Behavioral scope: only inputs that previously produced corrupted results change behavior. Both-arrays element diffs and both-objects key diffs are untouched, and `merge`/`Trie` operate on difference lists so they need no changes. Consumers that filter to `add`-only differences now correctly skip container type changes instead of mangling the existing container.

## Checklist

- [x] I added a changeset.
- [x] I added tests.

### DIFF
--- a/.changeset/json-magic-diff-container-type-mismatch.md
+++ b/.changeset/json-magic-diff-container-type-mismatch.md
@@ -1,0 +1,5 @@
+---
+'@scalar/json-magic': patch
+---
+
+Treat array/object container mismatches as a single update in diff instead of descending into per-index adds. Before, `diff({ tags: {} }, { tags: ['x', 'y'] })` walked the array's indices as object keys and emitted per-index `add` differences, so `apply` wrote numeric string properties onto the existing object and produced the corrupted hybrid `{ tags: { '0': 'x', '1': 'y' } }`. A container type change (either direction) is now one `update` difference carrying the full new value, and applying it replaces the container correctly.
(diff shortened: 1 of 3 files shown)


DemonHa (approved):
Yep!

PR #9907, opened 2026-08-14
docs(json-magic): document reference-sharing in diff, merge and apply

DemonHa:
## What

Fifth PR of the `json-magic/diff` review series. It covers the two findings we deliberately **document rather than fix** (`merge` mutating the caller's document, `apply` aliasing its changes into the result), plus the confusing error a root-level change used to produce.

### Documentation

The diff primitives work on the documents themselves, never on copies, and nothing said so:

- `diff` — the `changes` it returns are live references into the documents it compared (`doc2` for add/update, `doc1` for delete).
- `merge` — folds two compatible changes together with `mergeObjects`, which writes in place, so merging writes into the document behind `diff2`.
- `mergeObjects` — mutates its target and shares the subtrees the source contributes.
- `apply` — mutates the document it is given and leaves the result structurally shared with the document the diff was built from.
- `detectDocumentConflicts` in `scalar-app` claimed "The function is pure". It is not: detecting conflicts writes local values into the `remote` document. The comment now says so.
- `README.md` told readers to clone the document before applying, which is not enough on its own. It now describes the real contract and lists all three `InvalidChangesDetectedError` triggers.

The aliasing itself stays as it is, per the plan for this series. Characterization tests pin it so a future switch to cloning is a deliberate break rather than a silent one.

### One behavior change

A change with an empty path asks to replace the document itself, which `apply` cannot do since it writes through the parent container of each path. It used to fail mid-changeset with a confusing message, leaving the document half updated:

```ts
apply({ name: 'John' }, [{ path: [], changes: [1, 2], type: 'update' }])
// before: InvalidChangesDetectedError: Process aborted. Path  at depth 0 is undefined, check diff object
// after:  InvalidChangesDetectedError: Process aborted. Root-level replacement is not supported,
//         the change targets the document itself instead of a property inside it
```

The check now runs up front, next to the prototype-pollution guard, so such a changeset is turned away before any of its entries touch the document. The set of rejected inputs is unchanged — only the message and the timing. No consumer matches on the old message.

`diff` deliberately keeps emitting these root-level changes: it is called for `.length` checks in `workspace-store`, and throwing there would be a wider behavior change than this bug warrants.

## Testing

- `pnpm vitest --run` in `packages/json-magic` — 631 passed
- `pnpm vitest --run` in `packages/workspace-store` — 2904 passed
- `pnpm --filter @scalar/json-magic types:check`, `pnpm --filter scalar-app types:check`, `pnpm knip`, `pnpm changeset status` — clean

No visual surface is touched: this PR only changes comments, one error message and tests.

### DIFF
--- a/.changeset/json-magic-diff-reference-sharing.md
+++ b/.changeset/json-magic-diff-reference-sharing.md
@@ -1,0 +1,16 @@
+---
+'@scalar/json-magic': patch
+---
+
+Explain the root-level change `apply` cannot handle, and document that `diff`, `merge` and `apply` share references with the documents they work on.
+
+A change with an empty path asks to replace the document itself, which `apply` cannot do since it writes through the parent container of each path. It used to fail with `Process aborted. Path  at depth 0 is undefined, check diff object` halfway through a changeset, and now says so up front and leaves the document untouched:
+
+```ts
+// `diff` emits a change with an empty path whenever the two documents differ at the root
+apply({ name: 'John' }, [{ path: [], changes: [1, 2], type: 'update' }])
+// InvalidChangesDetectedError: Process aborted. Root-level replacement is not supported, the
+// change targets the document itself instead of a property inside it
+```
+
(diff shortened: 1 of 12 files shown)


hanspagel (approved):
thanks

PR #9909, opened 2026-08-14
docs(sdks): document AsyncAPI support in the SDK Generator

DemonHa:
## Problem

The SDK Generator has grown real AsyncAPI support — a dedicated loader, an IR lowering for channels, a default-config generator, and thirteen AsyncAPI-specific diagnostic rules — and none of it is documented on scalar.com. The only AsyncAPI page we ship covers rendering in API References, so someone with an AsyncAPI description has no way to find out that it generates a client at all, let alone which protocols work, which targets ship a WebSocket runtime, or why their `type: apiKey` scheme produced no credential.

Two of those questions are actively expensive to get wrong:

- **The `action` inversion.** AsyncAPI 3's `action` is written from the provider's point of view, so `action: send` is a message the SDK *receives*. The generator inverts by default and exposes a `perspective` option to opt out. Getting it backwards swaps every send and receive type in an SDK while still type-checking cleanly.
- **The `apiKey` trap.** AsyncAPI spends the name `apiKey` on a broker-scoped credential with no HTTP position, and spells the ordinary API key `httpApiKey`. A document using the former silently generates no credential.

## Solution

Read through the AsyncAPI paths (loader, protocol table, IR compiler channel/message/streaming/security lowering, default-config generation, the diagnostics catalogue, and the `asyncapi-chat` fixture and its committed TypeScript snapshot), then wrote the page from what the code actually does.

New page at `documentation/guides/sdks/asyncapi.md`, registered under the SDK Generator section:

- **Supported versions** — 1.x/2.x/3.x, with anything below 3.x upgraded on load.
- **Protocol support** — `ws`/`wss` lower to connect methods, `http`/`https` to HTTP and streaming methods, everything else is skipped and reported. Includes the load-fails-on-no-supported-channel behavior and the document-order tie-break when a channel's servers disagree.
- **Send and receive** — the provider/client inversion and the `perspective` option.
- **Messages and payloads** — union lowering with discriminators from `const`/single-value `enum`, payload deduplication, empty payloads, and the JSON Schema `schemaFormat` families that are read.
- **WebSocket channels** — address parameters, `ws` binding handshake query and headers, and the four targets that ship a WebSocket runtime (TypeScript, Python, Rust, CLI).
- **HTTP channels** — how the message media type decides SSE, JSON Lines, or a buffered response, plus the `defaultContentType` fallback chain.
- **Servers, environments, and authentication** — URL reassembly from `protocol`/`host`/`pathname`, environment naming, the server-derived security union and required-versus-optional credentials, the four scheme families that lower, and the inline-scheme pitfall.
- **Configuration** — the `asyncapi` method block (`channel`, `operations`, `perspective`) and what the generated starter config looks like.
- **Diagnostics** — all thirteen AsyncAPI rules with default severities, gradable through the existing `diagnostics` config.
- **Limitations** — request/reply, traits, message headers, correlation ids, HTTP bindings, form media types, and unbound messages; the two known rough edges in the generated starter config for HTTP channels; and the absence of an augmented AsyncAPI document.

Also linked the new page from the SDK Generator overview (feature list and the "start from your document" step) and from the existing API References AsyncAPI page.

No changeset: this touches `documentation/` and `scalar.config.json` only, no package source.

## Checklist

- [ ] I added a changeset. 
- [ ] I added tests. 

https://claude.ai/code/session_01Pfm9KSPxVJ3j39mDpYZbfB

---
_Generated by [Claude Code](https://claude.ai/code/session_01Pfm9KSPxVJ3j39mDpYZbfB)_

### DIFF
--- a/documentation/asyncapi.md
+++ b/documentation/asyncapi.md
@@ -64,5 +64,9 @@ The broker-specific scheme types have dedicated credential inputs too:
 
 Credentials are kept in the auth store alongside the OpenAPI ones, so they persist while you browse and never leave the browser.
 
+## Generating SDKs
+
+The same document can generate a client. The [Scalar SDK Generator](guides/sdks/asyncapi.md) turns each channel into a WebSocket connect method or an HTTP streaming method, with typed send and receive events.
+
 > [!NOTE]
 > AsyncAPI support is still a work in progress, so not every part of the specification is rendered yet. The progress is tracked on GitHub in [issue #7080](https://github.com/scalar/scalar/issues/7080) — subscribe there to receive updates.
--- a/documentation/asyncapi.md
+++ b/documentation/asyncapi.md
@@ -66,7 +66,7 @@ Credentials are kept in the auth store alongside the OpenAPI ones, so they persi
 
 ## Generating SDKs
 
-The same document can generate a client. The [Scalar SDK Generator](guides/sdks/asyncapi.md) turns each channel into a WebSocket connect method or an HTTP streaming method, with typed send and receive events.
+The same document can generate a client. The [Scalar SDK Generator](guides/sdks/asyncapi.md) turns each channel into a WebSocket connect method or an HTTP streaming method, with typed send and receive events. AsyncAPI input to the SDK Generator is experimental.
 
 > [!NOTE]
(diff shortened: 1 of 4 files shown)


hanspagel (approved):
niiiice

PR #9913, opened 2026-08-16
feat(snippetz): add support for julia language

JosiahParry:
## Problem

This PR adds support for the [Julia](https://julialang.org/) programming language to the snippetz feature of Scalar.

We use Scalar to generate our API documentation for [ricochet](https://try.ricochet.rs/api/docs#tag/content).

Our tools target developers of the `R`, `julia`, and `python` programming languages. 

Scalar's API Snippets do not include Julia as a language. It would be great to support this small but mighty programming language community.

## Solution

We added support for [HTTP.jl](https://juliaweb.github.io/HTTP.jl/stable/api/client/) as a drop down value. 

<details><summary>AI Summary</summary>

## What

  Adds Julia as a code example target, using [HTTP.jl](https://juliaweb.github.io/HTTP.jl/stable/api/client/)
  as its client (`julia/http`).

  ## How

  The generator lives in `packages/snippetz/src/plugins/julia/http`, with the string
  escaping and collection formatting split into `src/libs/julia.ts` the way
  `libs/rust.ts` does it for reqwest. Registering the target in `GROUPED_CLIENTS`
  (`@scalar/types`) drives everything else: the clients array, the package export,
  the highlight.js language, the icon, the picker labels, and the generated .NET/Java
  enums and configuration docs.

  Decisions worth reviewing:

  - **Verb helpers where they exist.** `HTTP.get`/`post`/`put`/… for the standard
    methods, falling back to `HTTP.request("PURGE", …)` otherwise. `get`, `head` and
    `options` take no body argument in HTTP.jl, so a body on those methods is passed
    as the `body =` keyword instead.
  - **JSON bodies become `Dict`s** serialized with `JSON.json`, rather than a raw
    JSON string literal, so the snippet can be edited as Julia. Invalid JSON falls
    back to the literal text.
  - **Form bodies use the native types.** `Dict` for `x-www-form-urlencoded` (HTTP.jl
    derives the content type from it) and `HTTP.Form` / `HTTP.Multipart` for
    multipart. For multipart the request's `Content-Type` header is dropped, because
    `HTTP.Form` generates its own boundary and keeping the incoming one would produce
    a body the server cannot parse.
  - **Query parameters, cookies and basic auth are keyword arguments**
    (`query =`, `cookies =`, `basicauth =`) instead of being folded into the URL or
    a header, which is what the HTTP.jl client API is built around.
  - `$` is escaped in string literals, since Julia interpolates it.

  ## Alternatives considered

  - **Raw JSON string body** instead of `Dict` + `JSON.json`. Simpler and drops the
    `using JSON` line, but the snippet stops being editable as Julia — the same
    trade-off `r/httr2` resolves in favour of native structures.
  - **Appending the query string to the URL**, as `rust/reqwest` does. Correct, but
    hides the `query =` keyword that HTTP.jl users expect.
  - **An `Authorization: Basic …` header** rather than `basicauth = (user, pass)`.
    The header form also works on HTTP.jl 1.x, where `basicauth` is a `Bool`; the
    keyword form matches the current client API and reads better.
  - **`julia/httpjl` as the client id.** Kept `http`, matching `dart/http` and the
    package's actual name.

</details>

## Checklist

- [x] I added a changeset. 
- [x] I added tests.


JosiahParry:
Thanks all! Please let me know what needs to be done to support @JuliaLang in Scalar docs! <3

hanspagel (approved):
so cool!

JosiahParry:
<3 🚀

PR #9918, opened 2026-08-17
feat(sidebar): render sidebar navigation as links instead of buttons

marclave:
## What

Sidebar entries were `<button>` elements, so search engines could not crawl the navigation and users could not open an entry in a new tab. They now render as real `<a href>` links whenever a URL is available.

This is the sidebar used by the docs product on scalar.com, so the same `getHref`/`href` API applies there.

- ~~**``@scalar/components`** — `ScalarSidebarItem`, `ScalarSidebarButton` and `ScalarSidebarGroup` accept a new `href` prop. It has to be a *declared* prop rather than a fallthrough attribute, because `useBindCx` puts fallthrough attrs on the wrapping `<li>` instead of the anchor. Group labels were hard-coded to `is="button"` and now switch to an anchor when an href is set. The href is dropped when the item is `disabled` (an anchor cannot be disabled the way a button can) or when the rendered element cannot be a link.~~ Removed as not needed
- **`@scalar/sidebar`** — `ScalarSidebar` and `SidebarItem` accept a `getHref(item)` callback. Plain left clicks on the link still emit `selectItem` and drive in-app navigation; modified clicks (cmd/ctrl/shift/alt/middle) are left to the browser, so **open-in-new-tab now works**; clicks on decorator content outside the link keep their native behavior.
- **`@scalar/api-reference`** — passes `getHref` built on a new SSR-safe `makeHrefFromId`, so hrefs are present in server-rendered HTML and are spelled the same way as the URLs pushed to history (no duplicate URL spellings for one section).
- **`@scalar/helpers`** — new `isPlainLeftClick` helper.

Consumers that pass no href — the API client and the desktop app — render exactly as before, apart from two deliberate style fixes that apply to buttons too (inset focus ring, `cursor-default` on disabled/selected).

## Notes for reviewers

- **Hash routing gets no crawl benefit.** Search engines do not treat fragments as separate URLs, so the SEO win requires `pathRouting`. In hash modes the anchors still improve link semantics and open-in-new-tab. This is stated in the changeset rather than being implied away.
- **Keyboard semantics change**, as they must for links: Enter activates an entry, Space scrolls the page. No key handler was added to fake button behavior on an element announced as a link.
- **Collapsed groups are still absent from SSR HTML** (`<ul v-if="open">`), so deep links are only crawlable for groups expanded during SSR, e.g. via `defaultOpenAllTags`. Out of scope here and tracked separately.
- `makeHrefFromId` normalizes a degenerate id that would otherwise produce a protocol-relative (off-site) href — reachable when a document title slugifies to an empty string and the base path is empty.

## Visual

The change is intentionally pixel-neutral; the meaningful artifact is the rendered DOM. Verified live in the `api-reference` playground against the Scalar Galaxy document:

Before — every entry a button:

```html
<li><button type="button">Get a token</button></li>
```

After — every entry a crawlable link, with the discrete caret correctly left as a button:

```html
<li><a href="#scalar-galaxy/tag/authentication/POST/auth/token">Get a token</a></li>
```

Clicking still performs SPA navigation to the identical URL the href points at, the page title updates, and `aria-current="page"` lands on the selected entry.

## Testing

- 1710 api-reference, 207 sidebar, 73 components and 34 helpers unit tests pass.
- Full Playwright e2e suite (206 tests) passes; sidebar-entry locators moved from `getByRole('button')` to `getByRole('link')`, with a link/button union where the discrete caret is still a button.
- `types:check`, `biome`, `prettier`, `lint:vue` and `knip` clean; `changeset status` bumps the four changed packages.
- The branch went through five rounds of adversarial review-and-fix; the last round produced no findings.


packages/components/src/components/ScalarSidebar/ScalarSidebarButton.vue
```diff
@@ -25,25 +25,53 @@ export default {}
...
+  return href || undefined
+})
+
 const variants = cva({
   base: [
     'group/button peer/button flex items-stretch rounded p-2',
     'font-sidebar text-base/4  text-sidebar-c-2 no-underline wrap-break-word',
+    // Match the inset focus ring the theme reset gives buttons so the ring
+    // looks the same when the item renders as an anchor
+    'focus-visible:-outline-offset-1',
```
hwkr: I thought the reset already did this 🤔 If not it should

marclave: it defe doesn't and scalar.com sidebars aren't crawlable by SEO rip

packages/components/src/components/ScalarSidebar/ScalarSidebarGroup.vue
```diff
@@ -62,12 +64,27 @@ const { cx } = useBindCx()
...
+
+  // When the label is a link the browser navigates away unless the consumer
+  // prevented it, so the group only toggles once navigation is off the table.
+  // A modified click (meta, ctrl, shift, alt or a non-primary button) opens
+  // the link in a new tab and must not mutate the state of this tab.
+  if (href && !(isPlainClick && event.defaultPrevented)) {
+    return
+  }
+
+  // Only toggle the open state if the group is uncontrolled and not discrete
```
hwkr: This comment should probably be on the if up above

packages/components/src/components/ScalarSidebar/types.ts
```diff
@@ -8,6 +8,30 @@ import type { SidebarGroupLevel } from './useSidebarGroups'
...
+   * on the label follows the link. The caret sits inside that link, so a group
+   * with an href should be `controlled` or `discrete` for its toggle to be
+   * operable: `discrete` renders the caret as a separate button beside the
+   * link, while an uncontrolled group only toggles when the consumer prevents
+   * the default navigation in its own click handler.
+   *
+   * ScalarSidebarSection and ScalarSidebarNestedItems never render a link, so
+   * they ignore the href entirely.
+   */
+  href?: string
```
hwkr: So I don't think we actually need all this, the group should expose a slot so the button can be manipulated directly allowing href to be set on it (in @scalar/sidebar) presumably

hwkr (commented):
Is it cool if I have a look at this tomorrow? I don't know if we actually need all these props for the component changes

marclave:
ye sg

amritk (approved):
code looks good, lotta tests did not test myself

hwkr (approved):

PR #9922, opened 2026-08-18
fix(sidebar): use discrete group toggles where applicable

hwkr:
## Problem

Group labels render as links now, so the caret next to them ended up as a `div` inside the anchor. One focusable thing per row means there's no way to collapse a group from the keyboard — Enter just follows the link (which expands it as a side effect). `aria-expanded` also sat on the link instead of on something that can toggle, and the caret's screen reader text ran into the link name ("Authentication Close Group").

## Solution

The caret is its own button next to the label now.

* Clicking behaves exactly the same, this just adds the ability to toggle the tags directly
* `aria-expanded` moved off the link and onto the toggle, so the markup is semantically correct
* Toggles are named after their group ("Open Group - Authentication") instead of twenty identical "Open Group" buttons
* Pulled the toggle out into `ScalarSidebarGroupToggleButton` so `@scalar/sidebar` stops hand-rolling its own copy

https://github.com/user-attachments/assets/74a857cc-bc04-4300-af31-0a3ed7d20544

## Checklist

- [x] I added a changeset.
- [x] I added tests.


DemonHa (approved):

amritk (approved):
I see tests, I approve

DemonHa (approved):

PR #9923, opened 2026-08-18
feat: add @scalar/highlight, a zero-dependency syntax highlighter

amritk:
three_light

three_dark

## Problem

`@scalar/code-highlight` carries `highlight.js` and `lowlight`, which is 127.34 KB gzipped for the language set `ScalarCodeBlock` uses. We want a lightweight, zero-dependency highlighter that tree-shakes properly and reads its colours from the theme already on the page.

## Solution

Moves the standalone `code-highlighter` repository in as `packages/highlight`.

- **Grammars are plain data.** Each language is a small state machine; every state compiles to one merged regular expression, and rendering goes straight to an HTML string without allocating token objects.
- **40 languages**, one module each, so importing `python` does not cost you `scala`.
- **Zero runtime dependencies.**
- **Theme integration.** Colours come from the `--scalar-*` variables `@scalar/themes` already defines, so a block follows the app's theme and switches with `.light-mode` / `.dark-mode` without being re-highlighted.
- **Compat layer.** `@scalar/highlight/compat` is a drop-in for `syntaxHighlight`, emitting the same `hljs-*` markup.

Nothing consumes it yet — this only lands the package. Swapping consumers over is a separate PR.

### What was dropped rather than duplicated

The Bun lockfile and config, the package-local Biome, lefthook, CI, editor and agent files, and the LICENSE. Tests are colocated as `src/**/*.test.ts` per repo convention, and the compat differential test imports the real lowlight pipeline from `@scalar/code-highlight` instead of the copy the standalone repo vendored — so it tracks the package it replaces rather than a snapshot of it.

### Bundle size

Bundled with Rolldown, minified, gzipped. Entry points *call* the API rather than re-exporting it, so tree-shaking runs the way it does in a real app.

| Bundle | Gzipped |
| --- | ---: |
| core (tokenize + render) | 2.32 KB |
| core + Python | 4.33 KB |
| lazy loader, initial chunk | 3.93 KB |
| `style.css` | 1.26 KB |
| compat + JSON/JavaScript/Bash | 5.36 KB |
| `@scalar/code-highlight` today | 127.34 KB |

Reproduce with `pnpm size` in the package.

### Testing

1098 tests across 32 files. Per-language suites, cross-language invariants over a shared sample corpus, a stylesheet suite that recomputes every resolved colour in oklab and fails below WCAG AA, and a linearity suite that fails a grammar whose per-character cost grows with line length.

The compat suite ports the four assertions `@scalar/code-highlight` ships today and adds a differential run against the real pipeline: rendered text, the `<code>` envelope and line alignment must match byte for byte.

## Visual

The only UI-visible change is the line-number gutter. It was a fixed `width: 2ch`, so past 99 lines the counter overflowed its box and the gutter column went ragged. `highlightBlock` now emits `--shl-line-digits` for the whole block and the stylesheet sizes the gutter from it, so the code starts at the same column on line 9 and line 10.

Verified in Chromium against `@scalar/themes` at 8, 12 and 105 lines — before/after screenshots attached below. Blocks under 100 lines render identically to before (the 2ch floor is unchanged); the 105-line block goes from a ragged gutter to a right-aligned column.

## Review round

A review pass on the first commit surfaced issues that are fixed in the follow-up commits:

- `sideEffects: false` let bundlers drop the documented `import '@scalar/highlight/all'` — the bundle fell to 5.5 KB and threw at runtime. Narrowed to the entries that actually have side effects.
- Language names reaching a plain object lookup (```` ```constructor ````) threw where the package being replaced renders plain text. Guarded with `Object.hasOwn`; the compiled state map is now null-prototype, which also measured 12% faster on the tokenizer's transition lookup.
- 29 CodeQL alerts, all in test code — 30 duplicated tag-strip regexes replaced with one shared helper.
- A catalog entry for esbuild had moved 9 unrelated monorepo resolutions; pinned to the version already in the tree, which brought the lockfile diff down from 1465 lines to 148.
- The size benchmark's lone esbuild dependency was then swapped for Rolldown — the bundler Vite already builds on in this repo — so the package pulls in no esbuild at all. Rolldown is pinned to the version already in the tree, and the figures above are from the Rolldown run.

## Checklist

- [x] I added a changeset.
- [x] I added tests.

https://claude.ai/code/session_012xJPgGzT8bZZibQboFKhA4


hanspagel:
Pushed four commits addressing the ReDoS findings from review — three grammar rules were quadratic in line length, which is a client-side DoS on untrusted code blocks (a long pathological line froze the tab for seconds).

**Fixes** (each verified linear on the real tokenizer; before → after at 64k chars):
- **perl** `b8b8598` — `[$@%]$*NAME` ate a run of bare `$`, failed for want of a name, and retried at every position. Capped the leading deref run the way the grammar already caps its other scans. **3.1s → 4ms**
- **powershell** `90e4733` — `${ID}[\\w.]*` is two adjacent word-class runs (`\\w*[\\w.]*`); a `[` with no closer let the engine split the same characters O(n) ways. Rewrote as one quantifier. **6.3s → 0.9ms**
- **css** `718ec98` — inside `{ }` a bareword with no colon left the property rule to match-and-fail at every column of a long identifier. Added a fallback that consumes it. **2.6s → 0.1ms**
- **test** `ea38cc1` — added all three to the existing linearity suite (perl as a repeated-seed ratio, powershell/css as wall-clock budgets).

Verified byte-identical tokenization vs the originals across a real-world corpus (including `$$$ref`, nested types, 70-char custom properties, `&:hover`). Full suite: **1106 passing**, tsgo + biome clean.

hanspagel (approved):
sick work

PR #9927, opened 2026-08-19
fix(api-reference): show a composed schema's own description

### LINKED ISSUE #9926: allOf-composed schema shows first branch's description instead of its own top-level description
## Description

When a schema uses `allOf` composition and has its own top-level `description`, Scalar displays the *first allOf branch's* description instead of the schema's own — even though the schema's own description is present and correct in the spec.

This looks related to #6974 / #9546, but that fix does not appear to cover this case — the schema's own top-level `description` (sibling to `allOf`) still loses to the first branch's description.

## Reproduction

```yaml
openapi: 3.0.0
info:
  title: Repro
  version: "1.0"
paths:
  /jobs/{id}:
    get:
      responses:
        '200':
          description: OK
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/Job'
components:
  schemas:
    Entity:
      type: object
      description: Base class for all searchable entities.
      properties:
        entityType:
          type: string
    Job:
      description: A job.
      allOf:
        - $ref: '#/components/schemas/Entity'
        - type: object
          properties:
            jobNumber:
              type: string
```

Viewing the `Job` schema (standalone, in the Models/schema browser):

- **Expected:** "A job." (Job's own `description`, sibling to its `allOf`)
- **Actual:** "Base class for all searchable entities." (Entity's `description`, from the first `allOf` branch)

## Environment

- `@scalar/api-reference` 1.65.1 (via `Scalar.AspNetCore` 2.16.20)
- OpenAPI 3.0.0 spec

## Notes

- Confirmed via the raw spec content that `Job.description` is correctly set to `"A job."` at the top level — this is a rendering issue, not a spec-generation issue on our end.
- We observed this specifically when browsing the schema standalone (Models sidebar); we have not separately confirmed whether the same happens when the schema is rendered inline inside an operation's request/response body.
- #6974 (closed, fixed via #9546) describes a very similar symptom ("top-level description should take precedence over allOf member descriptions"), but doesn't appear to reach the code path exercised by this repro.

lazerg:
## Problem

A schema that composes with `allOf` and carries its own top-level `description` lost that description. Browsing such a schema standalone in the Models section showed the first `allOf` member's description instead, so a `Job` schema described as "A job." rendered as "Base class for all searchable entities." from the `Entity` base it extends.

The description was dropped because the schema itself has no `properties`, `patternProperties` or `additionalProperties`, and the card skips the description in that case on the assumption that the property rows will show it. For a composition there are no property rows on that card, so the text had nowhere to go.

### Before

Before

### After

After

## Solution

Keep the description when the schema has an `allOf`. The composition members still render their own cards with their own descriptions, so nothing is duplicated and nothing is lost.

## Checklist

- [x] I added a changeset. 
- [x] I added tests.

## Ticket

Closes #9926

### DIFF
--- a/.changeset/api-reference-allof-own-description.md
+++ b/.changeset/api-reference-allof-own-description.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
+Show a composed schema's own description when it has no properties of its own. A schema that only carries `allOf` plus a top-level `description` dropped that description and rendered the first `allOf` member's description instead, which was visible when browsing the schema standalone in the Models section.
--- a/packages/api-reference/src/components/Content/Schema/Schema.test.ts
+++ b/packages/api-reference/src/components/Content/Schema/Schema.test.ts
@@ -186,7 +186,14 @@ describe('Schema', () => {
         },
       })
 
-      expect(wrapper.text()).toContain('A job.')
+      const text = wrapper.text()
+
+      // The schema's own description renders exactly once. The member cards must
+      // not repeat it, otherwise the same text would show twice.
+      expect(text.split('A job.').length - 1).toBe(1)
+
+      // The allOf member still renders its own description, so nothing is lost.
+      expect(text).toContain('Base class for all searchable entities.')
     })
 
     it('shows the parent description for discriminator-based oneOf schemas', () => {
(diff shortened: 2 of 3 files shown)


hanspagel (approved):
thank you! 🙏

PR #9928, opened 2026-08-19
feat(Scalar.AspNetCore): Add async configure options overloads.

ff-gerard:
## Problem

I wanted to pre-fill the `ClientId` field for oauth using the logged in users credentials. These credentials are saved in the database, so I need to use async code to access these credentials, something like:

```csharp
app.MapScalarApiReference("/scalar", async (options, httpContext) =>
{
    if (httpContext.User.Identity?.IsAuthenticated == true)
    {
        var userManager = httpContext.RequestServices.GetRequiredService<UserManager<UserEntity>>();
        var dbUser = await userManager.GetUserAsync(httpContext.User);
        if (dbUser is not null)
        {
            options.AddOAuth2Authentication("OAuth2Token", oauth =>
            {
                oauth.Flows ??= new ScalarFlows();
                oauth.Flows.ClientCredentials ??= new ClientCredentialsFlow();
                oauth.Flows.ClientCredentials.ClientId = dbUser.ClientId;
            });
        }
    }
});
```

Except the `async`/`await` wasn't available here.

Another use case that currently "fails" would be:

```csharp
app.MapScalarApiReference(async options =>
{
    options.HeaderContent = await File.ReadAllTextAsync("header.html");
});
```

## Solution

I modified the scalar ASP.NET package to add overloads that do support async/await, and redirected the existing overloads to call the new async version with a `return Task.CompletedTask;` at the end. All existing code would keep working and keep using the existing overloads, but now async version are also supported.

## Checklist

- [X] I added a changeset. 
- [ ] I added tests. _(There were no existing test covering these method calls, and I am not sure what kind of tests would be necessary/useful here.)_
- [X] I updated the documentation.

### DIFF
--- a/.changeset/wise-clowns-notice.md
+++ b/.changeset/wise-clowns-notice.md
@@ -1,0 +1,5 @@
+---
+'@scalar/aspnetcore': minor
+---
+
+Add async parameter overloads to MapScalarApiReference, so async services can be used to configure options.
--- a/documentation/integrations/aspnetcore/integration.md
+++ b/documentation/integrations/aspnetcore/integration.md
@@ -134,6 +134,19 @@ app.MapScalarApiReference("/docs", (options, httpContext) =>
 });
 ```
 
+## MapScalarApiReference Async Overloads
+
+The `MapScalarApiReference` method provides several overloads to allow for asynchronous code in the `configureOptions` parameter.
+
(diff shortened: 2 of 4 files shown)


hanspagel (approved):
thanks for the PR 🙏

PR #9930, opened 2026-08-19
fix(api-client): new schema example will not render if previous schema is modified

xmrfate:
## Problem

If a request body schema object used oneOf (either with a discriminator or without), the request body example in the API client modal would not reset if another selection was made. This only occurred if the request body was changed.

For example:
- User selects Schema1 from the oneOf dropdown
- Test Request is clicked, example for Schema1 shows
- User selects Schema2 from the oneOf dropdown
- Test Request is clicked, example for Schema2 shows
- User modifies Schema2 example
- User selects Schema3 from the oneOf dropdown
- Test Request is clicked, example for Schema2 shows --> Here I expect Schema3 to be displayed in the request body.

## Solution

The current selection is tracked and compared to the previous selection. If the current selection is found to be different than the previous selection, an update event is emitted to update the modal request body.

## Checklist

- [x] I added a changeset.
- [x] I added tests.


hanspagel (approved):
thanks!

PR #9931, opened 2026-08-19
fix(components): keep the copy button off the end of one-line code blocks

DemonHa:
## Problem

On a single-line `ScalarCodeBlock` the copy button is centered over the only line there is (`top-1/2 right-1`), floating above the scroll area. The content stops at the block's own `p-2`, so once the line is longer than the block, its tail sits under the button — and under the label the button grows to on hover — no matter how far you scroll. A long install command ending in a repository URL can never be read to the end: it stops at `…-repository-`, the `name.git` is unreachable.

This turned up downstream (an SDK install command in the dashboard's preview pane) and was patched there with a wrapper class. The overlap is created entirely by this component's own layout, so every consumer of a one-line code block with a copy button has it — the fix belongs here, and the downstream patch can be dropped.

## Solution

One-line blocks that actually render a copy button now reserve trailing space on the `<pre>` (`pr-18`, 72px), so the end of the line can scroll clear of the button.

Where the space goes matters:

- **On the `<pre>`, not the scroll container** — the `<pre>` is `w-fit`, so its padding is reliably part of the scrollable width; browsers disagree about whether trailing padding on an `overflow-x` container is.
- **Not `scroll-padding` / `scroll-margin`** — these only change where scroll snapping and `scrollIntoView()` land an element, they do not extend the scrollable overflow. Measured in Chromium on this component (max `scrollLeft` on the overflowing one-liner): baseline `42`, `scroll-padding-right: 72px` on the container `42`, `scroll-margin-right: 72px` on the `<pre>` `42`, `padding-right: 72px` on the `<pre>` `114`. Only real box spacing moves it.
- **72px** clears the 24px button, its 4px offset, and the widest label it shows (`Copied`) with ~10px to spare, measured in the browser.

Multi-line blocks are untouched — there the button sits in the top corner and the padding is not applied. `v-if="copy && isContentValid"` became a `showCopy` computed so the padding and the button cannot disagree about whether a button is there.

A `Single Line Overflow` story was added to Storybook for the case. It is deliberately not in the e2e snapshot list; the existing `Single Line` story is unchanged (its content plus the reserved space still fits at the 640px snapshot viewport — `scrollWidth === clientWidth`, no new scrollbar), so no snapshots need updating.

## Visual

Verified in Storybook (`Components/ScalarCodeBlock` → `Single Line Overflow`), scrolled fully right with the copy label showing:

- **Before:** the line ends at `npm install git+https://github.com/scalar/a-rather-long-staging-repository-` — that is the end of the scroll range, `name.git` cannot be reached.
- **After:** the same block at full scroll reads `…a-rather-long-staging-repository-name.git`, with the `Copy` label sitting clear to its right.

Screenshots of both states were captured locally; this session has no way to upload images to GitHub, so they are with the author to paste in.

## Checklist

- [x] I added a changeset.
- [ ] I added tests. 


packages/components/src/components/ScalarCodeBlock/ScalarCodeBlock.stories.ts
```diff
@@ -139,6 +139,20 @@ export const SingleLine: Story = {
   }),
 }
 
+/**
+ * A one-line snippet that is wider than the block. The copy button floats over
+ * the end of the line, so the line reserves room to scroll clear of it.
+ */
+export const SingleLineOverflow: Story = {
```
hwkr: We can just update the normal one-line story to overflow, I think it's normal. And you can have the e2e test scroll to the end to make sure it works. (Remake the snapshots)

packages/components/src/components/ScalarCodeBlock/ScalarCodeBlock.vue
```diff
@@ -105,13 +108,20 @@ const { cx } = useBindCx()
     <div
       tabindex="0"
       class="custom-scroll overflow-x-auto p-2 -outline-offset-2 rounded-[inherit] min-h-0 min-w-0 flex-1">
+      <!-- On a one-liner the copy button sits centered over the only line there
+           is, so the end of a long line stays underneath it even at full scroll.
+           Trailing space keeps the line clear of the button, and it goes on the
+           `w-fit` <pre> because browsers disagree about whether padding on an
+           `overflow-x` container counts as scrollable. 4.5rem clears the 24px
+           button, its 4px offset and the label it grows to on hover. -->
```
hwkr: This feels like Claude over-explaining, also I think template comments like this leak into the DOM IIRC

packages/components/src/components/ScalarCodeBlock/ScalarCodeBlock.vue
```diff
@@ -90,6 +90,9 @@ const isContentValid = computed(() => {
   )
 })
 
+/** Whether the copy button is rendered on top of the code */
+const showCopy = computed(() => copy !== false && isContentValid.value)
```
hwkr: Nit: Since this is defaulted the strings will always be true so the only falsey value is the false
```suggestion
const showCopy = computed(() => copy && isContentValid.value)
```

packages/components/src/components/ScalarCodeBlock/ScalarCodeBlock.stories.ts
DemonHa: Done — the separate story is gone; `Single Line` now carries a command that overflows, and its e2e test snapshots it, scrolls to the end of the line, and snapshots again as `single-line-scrolled.png`. Both snapshots are regenerated in the commit.

One thing worth knowing about how they were made: this sandbox has no docker, so I ran the suite with `CI=1 playwright test` against the local browser instead of the runner image, and the webfonts are unreachable from here (`fonts.scalar.com` is blocked), which shifts every glyph. I served the woff2 files from disk for the run — with that, `json-string`, `bordered` and `ligatures` matched their committed baselines exactly, which is what gave me confidence the two new PNGs match what CI renders.

`base.png` is the one baseline I could not reproduce locally (it shows the copy button with its `JavaScript` label, and `=>` as a ligature, neither of which my render produces). It is untouched here and passes in CI, so I left it alone — but it may be worth a look separately.

---

packages/components/src/components/ScalarCodeBlock/ScalarCodeBlock.vue
DemonHa: Trimmed and moved into the script as a doc comment on a `reserveCopySpace` computed, so the template just reads `:class="{ 'pr-18': reserveCopySpace }"` and nothing ships to the DOM.

---

packages/components/src/components/ScalarCodeBlock/ScalarCodeBlock.vue
DemonHa: Applied.

---

packages/components/src/components/ScalarCodeBlock/ScalarCodeBlock.stories.ts
DemonHa: Confirmed by CI: the components snapshot job is green on `41027ae` (200 passed), so both regenerated PNGs match what the runner image renders.

---

packages/components/src/components/ScalarCodeBlock/snapshots/single-line-scrolled.png
hwkr, line 1: Code looks fine but `pr-18` might be a might too much, we probably need our spacing base is 4px so it comes out to 18*4. Probably need like `pr-6`? I think the icons are size-4 or something

DemonHa: Switched to `pr-6` and regenerated `single-line-scrolled.png`.

Measured what actually sits over the end of the line, from the block's right edge (real fonts, in the browser):

| state | space taken |
|---|---|
| at rest (button only) | 29px |
| hovered / focused (`Copy` + backdrop) | 75px |
| just copied (`Copied` + backdrop) | 87px |

The button is `size-6` (24px — the `size-4` is the icon inside it) at `right-1`, and the scroll container's own `p-2` already contributes 8px, so `pr-6` leaves the line clearing the button by 3px. The tail is readable at full scroll, which is the bug.

The trade-off worth knowing: 72px was sized for the label the button grows into, so at `pr-6` the last ~40px of the line sits under `Copy` whenever the block is hovered **or focused**. Hover is momentary and arguably fine — plenty of code blocks do exactly that — but the scroll container is `tabindex="0"`, so a keyboard user who tabs in to arrow-scroll holds `focus-within` the whole way and reads the tail through the label. `pr-17` (68px) is the measured minimum that keeps it clear. Happy to switch back up if you would rather pay the space — your call.

---

hwkr (approved):
Looks good! Got some tests failing?

DemonHa:
Thanks! The failure was a flake, and CI is green now — I re-ran the failed jobs and `test-js / test-packages (3, 3)` passed, along with the `required-jobs-pull-requests` roll-up.

What failed was `@scalar/workspace-store` → `client.test.ts` → *uses workspace fetcher to resolve documents if not specified per document*, with `Document 'default' could not be loaded`, i.e. the fetch against the test's own fastify server did not connect. Nothing in this PR can reach it: the diff is one Tailwind class plus a snapshot in `@scalar/components`, and `workspace-store` does not depend on that package. The same shard passed on both earlier commits of this branch.

Worth a separate look though — every server-backed test in that file binds the same hard-coded port (`const port = 9988`) and starts/stops fastify per test, which is exactly the shape that goes intermittent on a loaded runner.

---

PR #9932, opened 2026-08-19
fix(workspace-store): keep asyncapi documents intact in the server store

hwkr:
## Problem

`createServerWorkspaceStore` ingested every document as OpenAPI, so an AsyncAPI document lost its `channels` and `operations` before anything downstream could read them — published docs sites rendered only Introduction and Models. The editor was fine since it uses the client store.

## Solution

AsyncAPI documents now take the same path they already take in the client store.

* Runs the AsyncAPI upgrader instead of `upgrade(document, '3.1')`, and skips the OpenAPI coerce (it was injecting `openapi: ''` and breaking the type discriminator)
* Builds navigation with `traverseAsyncApiDocument` so `asyncapi-channel` and `asyncapi-operation` entries actually reach `x-scalar-navigation`
* Preserves `x-original-aas-version`, and keeps the document whole — there is no `paths` to externalize and consumers read channels/operations straight off it

future thing: the server store doesn't resolve `$ref`s at all (no magic proxy), which silently drops any operation behind a `$ref`'d path item — including the standard split-file style. Separate fix, tracked outside this PR.

## Checklist

- [x] I added a changeset. 
- [x] I added tests.


packages/workspace-store/src/server.ts
```diff
@@ -364,6 +368,48 @@ export async function createServerWorkspaceStore(
...
+
+    // Clone first: the upgrader and the traversal both write to the document they are handed, and
+    // the caller may keep using the object it passed in.
+    // The upgrader is typed against the loose `UnknownObject` shape; the result is a valid 3.x
+    // AsyncAPI document, so cast it back.
+    const asyncApiDocument = upgradeAsyncApi(deepClone(document)) as AsyncApiDocument
+
+    // Nothing is externalized, so the document owns no chunks. The empty entry keeps `get()` and
+    // chunk generation well defined for the document name.
+    assets[name] = {}
```
DemonHa, line 410: we don't need to support chunking for async api right now so this is fine

DemonHa (approved):

PR #386, opened 2023-10-27
chore: fix some linter warnings

hanspagel:
This PR fixes a few tiny linter warnings.

PR #9933, opened 2026-08-19
feat(agent): chat kit unification

marclave:
## What this adds

Scalar ships four AI chat surfaces that grew four separate implementations of the same
infrastructure — session state, IndexedDB persistence, streaming markdown, scroll
management, approval flows, error parsing. This PR introduces a shared chat kit and
recomposes `@scalar/agent-chat` onto it as the first adopter. Scalar's other product
surfaces adopt the same kit downstream.

### `@scalar/chat-protocol` (new)

The wire contract, with zero UI: tool schemas for every chat domain (OpenAPI,
MCP, editor, dynamic) with server-verbatim descriptions, the error envelope +
`parseChatError()` (one shared implementation of the JSON-in-`Error.message`
parsing every surface hand-rolled), a declarative per-tool approval policy
registry, tool-part state helpers and the tool-card status machine, prompt
limits, and scrubbed wire fixtures. Legacy rejection encodings shipped by
frozen clients are recognized permanently — that compatibility is a contract,
not debt.

### `@scalar/chat` (new)

The session core and component layer:

- `createChatSessions` / `createChatHistory` — lazy per-chat instances with the
  two-watcher persistence pair, IndexedDB adapter (shipped database names and
  record shapes preserved), hydration that can never clobber a stored
  conversation, tombstones against in-flight saves, epoch-guarded
  delete/clear/reset races, scoped indexes and per-scope last-chat pointers.
- `createApprovalStore` — dual-mode (native + legacy encodings) approval state
  over any chat instance, with per-tool executors.
- `ChatViewport` — anchor-pinned scrolling (a new exchange pins the user's
  message to the top; release never leaves a blank tail), epoch-owned scroll
  writes, user-scroll detection that survives observer timing.
- `ChatMarkdown` — streaming renderer that splits source into stable blocks so
  completed prose never re-parses (selection survives, screen readers announce
  each block exactly once via a `role="log"` live region). The splitter is
  CommonMark-aware: fences, loose lists, indented continuations, HTML blocks
  (types 1–5), link reference definitions (collected and invisibly appended so
  citations resolve across blocks), footnote bail-out — and every glue decision
  is stable from a line's first streamed character.
- `ChatComposer` (autogrow field, IME-safe Enter and `maxlength`, over-limit
  gate, focus-reclaim that never steals from other inputs, `inputStart` /
  `actionsStart` / `banners` / `footnote` docks, public
  `--chat-composer-input-bg` hook), `ChatApprovalBar`, `ChatToolCard` +
  `ChatDiff`, `ChatToolFallbackCard` (protocol-aware, handles dynamic parts and
  all three rejection encodings), `ChatStatusBadge`, `ChatRoot`
  (density axis + copy dictionary provider).
- A playground (`pnpm --filter @scalar/chat playground`) exercising all of it.

### `@scalar/agent-chat` (recomposed)

Nine status-badge components and the bespoke markdown renderer deleted in favor
of the kit; `PromptForm` recomposed onto `ChatComposer` with all surface docks
(upload, error, approval, payment, free messages) and verbatim labels/styles.
Approval decisions now come from the shared policy registry instead of an
inline heuristic. Deliberately deferred: the window-scroll page keeps its
scroll hook (adopting `ChatViewport` there is a page restructure), and this
surface has no history, so the session core is unused here.

### Version convergence

`ai` 6.0.33 → 6.0.168 and `@ai-sdk/vue` → 3.0.168 (purely additive; picks up
upstream approval-flow fixes), zod 4 for all new schemas.

## Review

The branch went through repeated adversarial review cycles (independent
collectors, then verifiers instructed to refute each finding, then fixes —
looped to convergence). Everything confirmed was fixed with regression tests;
notable hardening: hydration/deletion/reset races in the session core, the
streaming DOM-stability contract of the markdown splitter, focus and IME edge
cases in the composer, and rejection-reason visibility across all three wire
encodings.

## Tests

- `@scalar/chat`: 176 tests · `@scalar/chat-protocol`: 41 · `@scalar/agent-chat`: 5
- `tsc`/`vue-tsc` clean on all three packages, `pnpm knip` clean
- Changesets included (minor bumps: new packages + agent-chat recomposition)

## Visual


packages/agent-chat/src/views/Chat/Chat.vue
```diff
@@ -122,7 +135,7 @@ div + .userMessage {
   max-width: 744px;
   z-index: 1;
 }
-.chat :deep(.markdown) {
+.chat :deep(.assistantText) {
```
hwkr: Don't think we need deep anymore if in this component

hwkr:
## Review

> [!NOTE]
> Reviewed by Claude, on Brynn's behalf. Every finding below was verified against the source; the four in "Fix before merge" were each reproduced or traced to a concrete reachable path.

Reviewed the full `main...HEAD` range (134 files, +10,204 / −1,238) against merge-base `4ad9f97`.

The engineering here is unusually careful — the epoch guards, the tombstones, the two-watcher persistence pair and the CommonMark reasoning in the splitter are all well-argued in comments, and the tests are real. Findings below are mostly seams rather than design problems.

### Verified claims

| Claim | Result |
|---|---|
| `@scalar/chat` tests | ✅ 177 passed (description says 176) |
| `@scalar/chat-protocol` tests | ✅ 41 passed |
| `@scalar/agent-chat` tests | ✅ 5 passed |
| `tsc` / `vue-tsc` clean | ✅ chat, chat-protocol, agent-chat, api-reference |
| Changesets | ✅ all bumped packages covered; `@scalar/types` patch is right since the public config uses `PartialDeep` |
| Translations | ✅ all 6 new keys in all 8 locales |
| `pnpm knip` clean | ⚠️ couldn't verify — knip fails to load on an unrelated pre-existing `integrations/docusaurus` playground config |
| No `v-html` / `innerHTML` in new code | ✅ |

---

## Fix before merge

### 1. The agent panel can re-mount permanently empty

`packages/api-reference/src/components/AgentScalar/AgentScalarPanel.vue:64`

`hasOpened` — the latch gating `AgentScalarChatInterface` — is `ref(false)`, set only by a **non-`immediate`** watcher on `showAgent`. But `showAgent` lives in the shared agent context (`use-agent.ts:38`), not in the panel: it survives the panel unmounting and is only cleared by `closeAgent`/`toggleAgent`.

`agentEnabled` is a per-document computed (`ApiReference.vue:1268`, reading `configList[activeSlug]?.agent?.disabled`) and gates the panel's `v-if` at `ApiReference.vue:1586`. So on a multi-document reference where one document disables the agent:

1. Open the agent on doc A — `showAgent = true`, `hasOpened = true`.
2. Switch to doc B (agent disabled) — panel unmounts, `hasOpened` is lost.
3. Switch back to doc A — panel re-mounts, `showAgent` is *still* `true`, so the watcher never fires and `hasOpened` stays `false`.

The panel renders open, with its header and `--open` class, and an empty body — until the user manually closes and reopens it.

**Fix:** `{ immediate: true }` on the watcher, or `ref(Boolean(agentContext.value?.showAgent.value))`.

### 2. Send looks enabled on an empty composer

`packages/chat/src/components/ChatComposer.vue:179`

`:disabled="sendDisabled || overLimit"` omits the empty-draft case, but `submit()` bails on it (`ChatComposer.vue:78`). With an empty composer the Send button renders fully enabled — accent-filled — and clicking it silently does nothing.

Two things suggest this is unintended: `main`'s `PromptForm` had `missingInput` in `submitDisabled`, and this PR's own final commit (`15bca8f`, *"give the disabled send button the scalar.com muted treatment"*) styles precisely the state that is now unreachable on an empty draft. Highest reachability of anything in this review — every visitor hits it on first open.

**Fix:** `sendDisabled || overLimit || (!modelValue.trim() && !allowEmptySubmit)`

### 3. The just-sent message can get stuck in the composer

`packages/agent-chat/src/state/state.ts:241` · `packages/agent-chat/src/Chat.vue:74`

`handleSubmit()` is `await chat.sendMessage({ text: prompt.value })` — it never clears the model, so the textarea still holds the sent message through the whole `submitted` wait.

The new watcher clears the draft only when `prompt.value === sentText`. That works when the user types nothing. But any keystroke during the `submitted` window **appends to the sent text**, so the equality fails and the draft is never cleared: the user is left with their entire just-sent message in the composer, and a second Enter resends it verbatim.

The comment's premise — *"a draft the user started typing in the meantime"* — is unreachable as written, because the old text was never removed. Both halves of this are new in this PR: `main` disabled the textarea during `state.loading` (which covers `submitted`), so typing there was impossible.

**Fix:** clear the model in `handleSubmit` right after `sendMessage` is issued, and keep the equality guard for the genuine race.

### 4. Multi-line link reference definitions render visible junk in every block

`packages/chat/src/components/chat-markdown.ts:106`

`LINK_DEFINITION_LINE_RE` collects only the **first line** of a definition, but CommonMark permits the destination and/or title on following lines. The truncated fragment (`[foo]:`) is not a valid definition, and it gets appended to every block, where it parses as a paragraph and renders as literal text.

Reproduced by rendering the same source both ways through the real `ScalarMarkdown`:

```
source: "See [foo] here.\n\n[foo]:\n/url\n\nTrailing paragraph."

whole-document parse:  "See foo here.  Trailing paragraph."
ChatMarkdown (split):  "See foo here.  foo:  Trailing paragraph.  [foo]:"
```

Two visible leaks in one message, in the component that renders every assistant reply. This breaks the splitter's own stated contract that splitting is render-equivalent to the whole-document parse.

Lesser variant: a definition with its title on the next line loses the title in the appendix, so blocks other than the defining one render the link without it.

**Fix:** only collect a definition line carrying a destination on the same line; otherwise fall back to the whole-source bail-out already used for footnotes.

---

## Follow-ups

<details>
<summary><b>5. A completed link definition re-keys every block</b> — <code>chat-markdown.ts:280</code></summary>

Because `definitions` is appended to every block, the instant a citation line completes, every block's content — and so its hash key — changes. Every block re-mounts: selection collapses and the `role="log"` region re-announces the whole message. That is the exact failure mode the component exists to prevent, and it hits hardest on the case the description calls out (citations resolving across blocks). The `isTerminated` guard avoids re-keying on every *token*, but not the once-per-definition-line remount.

Either state it as an accepted tradeoff in the docblock, or append definitions only to blocks that actually reference an undefined label.
</details>

<details>
<summary><b>6. <code>openDb()</code> has no <code>onblocked</code> handler</b> — <code>create-chat-history.ts:92</code></summary>

An `indexedDB.open` that triggers `onupgradeneeded` while another tab holds a connection fires `onblocked`, which is unhandled — the promise never settles. The knock-on is worse than a stalled read: `refresh()` never returns → `ready` never resolves → `createChatSessions.hydrate()` (awaiting `history.ready.value` at `create-chat-sessions.ts:467`) hangs → `hydratedIds` is never populated → the surface **silently stops persisting anything**, with nothing logged. That breaks the package's own "storage errors degrade to warnings, never gate chatting" contract.

Only reachable on a version upgrade, and `version` is 1 everywhere today — but it is a public option and the code explicitly anticipates bumps. `request.onblocked = () => reject(...)` is enough.
</details>

<details>
<summary><b>7. <code>clearAllChats</code> has no failure path</b> — <code>create-chat-sessions.ts:390</code></summary>

Every instance is disposed **before** `await history.clearAll()`. `deleteChat` guards the symmetric case carefully (checks `chatList` for a surviving record and bails, leaving the chat fully alive); `clearAllChats` has no equivalent. If `clearAll()` rejects, the promise propagates to a caller that may not handle it, `currentChatId` is never reset, and the disposed instances have lost their persistence watchers — leaving live chats that no longer save. `clearAll()` swallows per-record errors so rejection is rare (an `openDb()` failure would do it), but the asymmetry looks unintentional.
</details>

<details>
<summary><b>8. Native approval fails silently when the method is missing</b> — <code>create-approval-store.ts:112</code></summary>

`await chat.addToolApprovalResponse?.({ id: approval.part.approval?.id ?? '', approved: true })` — the method is optional on `ApprovalChat`, so a shell that does not wire it makes Approve and Reject on an `approval-requested` part permanent silent no-ops: the bar keeps rendering the decision and the send gate never reopens. The `?? ''` id fallback has the same shape — an empty id matches no pending approval SDK-side. Both deserve at least a `console.warn`.
</details>

<details>
<summary><b>9. <code>reject()</code> with a custom reason renders as "failed", not "rejected"</b> — <code>create-approval-store.ts:141</code></summary>

`errorText: reason ?? LEGACY_REJECTION_MESSAGES[0]`, but `toolCardStatus` (`status.ts:88`) maps `output-error` → `rejected` only when the text exactly matches one of two hardcoded English strings. Any custom reason falls through to `failed` — and the reason itself is what made it unrecognizable. No in-repo caller passes a reason today, so it's latent; but `reason` is public API on a package published for downstream adoption, and the description bills rejection-reason visibility as hardened. A structured encoding (`{ rejected: true, reason }`) would be sturdier than string matching.
</details>

<details>
<summary><b>10. <code>@scalar/chat</code> declares peers it never imports</b> — <code>packages/chat/package.json</code></summary>

Neither `ai` nor `@ai-sdk/vue` is imported anywhere in `packages/chat/src` or the playground — `knip.jsonc:145` confirms it ("kept as devDependencies for future type-compat tests"). The kit deliberately types against the SDK *structurally* so it "does not force a specific SDK version on consumers"; declaring version-pinned peers does exactly that. They're `optional` in `peerDependenciesMeta` so installs won't fail, but pnpm still warns on mismatch.
</details>

<details>
<summary><b>11. <code>field-sizing: content</code> is the only autogrow mechanism</b> — <code>ChatComposer.vue:270</code></summary>

Shipped in Safari 26 (Sept 2025) and Firefox 145 (Nov 2025). On anything older the field silently stops growing — the stacked layout degrades acceptably (48px min-height, internal scroll), the inline layout sits at one row. `api-reference` is embedded on customer docs sites with long-tail browsers, so worth a conscious call rather than an implicit one.
</details>

<details>
<summary><b>12. The composer textarea lost its label</b> — a11y</summary>

`main`'s `PromptForm` had `<label for="agentTextarea">Type a Request To get Started</label>`. `ChatComposer`'s textarea has no `id`, no `<label>` and no `aria-label` — only a placeholder, which is a last-resort accessible-name source. An `ariaLabel` prop defaulted from the copy dictionary would close it.
</details>

---

## Observations (not defects)

**Risk shape.** Only 5 of ~20 kit exports have an in-repo consumer — `ChatRoot`, `ChatComposer`, `ChatMarkdown`, `ChatStatusBadge` and the copy types. `createChatSessions` (553 lines), `createChatHistory` (321), `ChatViewport` (419), `createApprovalStore`, `ChatToolCard`, `ChatToolFallbackCard`, `ChatDiff`, `ChatApprovalBar` have none. That's ~1,300 lines of the highest-risk code in the PR — the IndexedDB hydration/deletion races and the scroll-ownership epochs — shipping behind unit tests and a playground and nothing else. Consistent with the stated plan, but it does mean those races are currently verified only by tests written against the same mental model that produced them.

**Two of the four blocking findings are seams** — behavior the donor `PromptForm` held that neither `ChatComposer` nor the recomposed shell picked up. That's the characteristic failure mode of this kind of extraction, and it's where a second pass over the remaining donor surfaces would pay off most as the other three surfaces adopt the kit.

**`isLegacyRejectionOutput` is unscoped.** `status.ts:44` is documented as the editor's `write_file` encoding, but `toolCardStatus` applies it to *every* `output-available` payload. A dynamic MCP tool returning top-level `{ rejected: true }` would render as a user rejection. The OpenAPI surface is safe — it nests the API body under `data.responseBody`.

**`ChatViewport`'s MutationObserver** watches `characterData: true` across the whole subtree and calls `measureDistanceFromBottom` (a forced layout read) per batch while streaming. Batched per microtask so likely fine, but it's the one spot that could jank on very long replies.

**`useMcpActions` returns non-reactive values.** `hasConfig` / `cursorLink` / `vscodeLink` / `copyMcpUrl` are computed once at setup from `options.config`. Faithful to the previous `OpenMCPButton` code, so not a regression — but it's now a shared hook used by two components, and multi-document references swap `mergedConfig.mcp` at runtime, so both would keep the first document's install links.

**`AgentScalarPanel` declares `url` twice** — in `defineProps` and again via `defineModel<string>('url')`. Copied from `OpenMCPButton`; harmless today (the parent binds one-way to a computed), but a confusing pattern to propagate.

---

<sub>Unrelated to this PR, but noticed while verifying: `pnpm vitest packages/chat --run` from the repo root — the invocation `AGENTS.md` recommends — fails with `localStorage is not defined`, because the root `vitest.config.ts` sets no jsdom environment so the package's own config never applies. Per-package `pnpm test` works.</sub>

marclave:
## Re: review

> [!NOTE]
> Written by Claude, on Marc's behalf. All four blockers are fixed in f48382c3b; CI is green across the suite (34 checks, 0 failing).

### Fix before merge

**1. Panel re-mounting permanently empty** — fixed. The `hasOpened` watcher runs `{ immediate: true }` now, declared after the refs its else-branch writes — the naive placement fired during setup and hit a `showMcpList` TDZ crash that the api-reference unit suite caught, so the re-mount path is genuinely exercised by tests.

**2. Send enabled on an empty composer** — fixed with the suggested condition: `sendDisabled || overLimit || (!modelValue.trim() && !allowEmptySubmit)`. The duplicate shell-side empty-draft gate that dcafa67c2 had added to `PromptForm` is dropped, so the kit is the single source of the disabled treatment. A new component test asserts the disabled attribute flips with the draft.

**3. Just-sent message stuck in the composer** — fixed. `handleSubmit` clears the model right after the send is issued — not before, so a synchronous `sendMessage` throw keeps the draft — and the streaming watcher keeps its equality guard for the prefill path and drafts typed before the stream begins.

**4. Multi-line link reference definitions** — fixed as prescribed: collection now requires the destination on the definition line, and a terminated destination-less `[label]:` line bails the whole source to a single block, footnote-style. The bail respects the streaming-stability ruling — an unterminated last line never triggers it, and once the line is final the decision cannot flip back. Three new splitter tests, including the review's repro (`"See [foo] here.\n\n[foo]:\n/url\n\nTrailing paragraph."` now renders identically split or whole). The title-on-next-line variant stays as scoped: the appendix copy drops only the title, and first-definition-wins keeps it effective in the defining block.

### Inline comments

- `:deep(.assistantText)` — correct, the element lives in the same component now; dropped.
- CodeQL `js/bad-tag-filter` — stale: alert #386 is already in `fixed` state, resolved by abb6c1200's `/--!?>/`.

### Follow-ups (5–12)

Deliberately left for a follow-up round, per their non-blocking framing: the docblock for the definition re-keying tradeoff (5), the `openDb()` `onblocked` handler (6), the `clearAllChats` failure path (7), warns on missing native approval wiring (8), a structured rejection encoding over string matching (9), the unused `ai` / `@ai-sdk/vue` peers (10), a conscious call on `field-sizing: content` (11), and the composer `aria-label` (12). One addition to that list from this round: the `useMcpActions` non-reactivity observation now applies to the agent panel's MCP menu as well as `OpenMCPButton`.

hanspagel (changes requested):
needs more love/tokens

Image

hanspagel:
better

Screenshot 2026-08-20 at 16 35 57

hanspagel (approved):
nice, consolidate all the things

PR #9934, opened 2026-08-20
fix(workspace-store): resolve local references in the server store

hwkr:
Operations behind a `$ref`'d path item silently vanished from published docs — no sidebar entry, no chunk, no error. The server store never resolved `$ref`s at all, so it now wraps documents in `createMagicProxy` like the client store already does.

* Split-file docs were hit too — bundling rewrites external refs into `#/x-ext/…` and coercion was dropping that bucket
* Follows ref chains, since a split file that's only a `$ref` bundles to a bucket entry that's itself a `$ref`
* AsyncAPI message titles resolve now, so the sidebar reads `Planet Created` instead of `planetCreated`
* `deletePathItemOperation` clears every hop — deleting through a chain was silently no-oping

Also worth knowing: `x-ext` now ships in the workspace payload, since the chunks keep their `#/x-ext/…` pointers and the client resolves them against the document root.

## Checklist
- [x] I added a changeset.
- [x] I added tests.


hanspagel (approved):
thanks!

PR #9935, opened 2026-08-20
fix(api-client): drop of a declared header after opening an operation with a request body

### LINKED ISSUE #9903: Opening an operation with a requestBody permanently drops a declared header parameter from the client
### What happens?

Opening an operation that has a `requestBody` corrupts the API client's header
rows. Two separate effects:

1. **On that operation** - a declared header renders with its name but an empty
   value.
2. **On every operation opened afterwards** - the *first* declared header
   disappears from the panel entirely, and stays gone for the rest of the
   session.

The headers are present and correct in the OpenAPI document, and still appear in
the generated code snippet. Only the client's parameter table loses them.
Reopening the affected operation does not restore it.

## Reproduction

1. Open **Alpha** -> Test Request. All three headers show a name and a value.
2. Open **Gamma** (the POST) -> Test Request.
3. Return to **Alpha** -> Test Request.

Image

### Expected

Alpha shows the same three headers, with their values, every time.

### Actual

```
Alpha  (first)   x-tenant-id=acme   X-Request-Source=web   Accept-Version=1
Gamma  (POST)    x-tenant-id=acme   X-Request-Source=""    Accept-Version=1
Alpha  (again)                      X-Request-Source=web   Accept-Version=1     <- x-tenant-id gone
```

`x-tenant-id` is absent from every operation opened from then on.

### OpenAPI Document

https://sandbox.scalar.com/e/HfbcA

hanspagel:
Opening an operation with a request body could blank out a declared header, both on that operation and on later ones. It kept happening for the rest of the session.

The header table reuses input cells across operations. When a cell switched into edit mode, it was never repainted, so it showed up empty. It now repaints as soon as the editable field appears.

## Ticket

Fixes #9903

### DIFF
--- a/.changeset/fix-dropped-header-9903.md
+++ b/.changeset/fix-dropped-header-9903.md
@@ -1,0 +1,6 @@
+---
+'@scalar/api-client': patch
+---
+
+Fix a header row rendering blank in the client after opening an operation with
+a request body. Reused input cells now repaint when they switch into edit mode.
--- a/packages/api-client/src/v2/components/code-input/CodeInputLite.test.ts
+++ b/packages/api-client/src/v2/components/code-input/CodeInputLite.test.ts
@@ -1056,6 +1056,38 @@ describe('CodeInputLite', () => {
     })
   })
 
+  describe('mode switch into editor', () => {
+    // Reused table rows (RequestTable keys its v-for by index) can flip a cell
+    // from a non-editor mode into editor mode. The editor element only appears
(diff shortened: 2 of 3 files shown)


hanspagel:
Closing as redundant. The root cause here is the same as #9797, and it was already fixed on `main` by #9867 (merged today): `CodeInputLite` now repaints via a `watch(editorRef, …)` when a reused table cell switches into editor mode. Verified the #9903 repro (both the enum-select→editor and disabled-label→editor mode switches) passes against current `origin/main`.

PR #9936, opened 2026-08-20
fix(api-reference): add bottom padding to classic layout

### LINKED ISSUE #9911: Missing padding-bottom in Classic Layout (Models section)
### What happens?

When using ScalarLayout.Classic (or layout="classic" in Vue), the rendered page lacks a bottom padding when scrolled all the way down.

The last element on the page (usually the last Model in the Models section) gets glued to the very bottom edge of the viewport, which looks broken and makes it difficult to read.

Image

Using a workaround

Image

Image

The .references-rendered container in the Classic layout should have some bottom padding (e.g., 80px or 100px) so there is some breathing room at the end of the scroll, just like the Modern layout has.

Proposed Fix: I dug into the source code and this can be easily fixed by adding padding-bottom: 80px; to /packages/api-reference/src/components/ApiReference.vue (around line 1912).

Image

### OpenAPI Document

_No response_

hanspagel:
In the classic layout the last section was stuck to the bottom edge of the screen. This adds some space below it so there is room to breathe.

## Visual

Before: last model glued to the bottom:

Before

After: 80px of space below it:

After

## Ticket

Closes #9911


hwkr (approved):
Looks good!

PR #9939, opened 2026-08-20
feat(components): migrate `ScalarCodeBlock` to @scalar/highlight

amritk:
## Problem

`ScalarCodeBlock` uses `@scalar/code-highlight` (lowlight-based) for syntax highlighting, which ships ~127 KB gzipped and builds a fresh lowlight instance on every call. The new `@scalar/highlight` package is a state-machine highlighter covering the same ground in ~47 KB gzipped for all 40 grammars, and exposes a `compat` layer so the swap needs no stylesheet or consumer changes.

This branch does two things: it readies `@scalar/highlight` for its first npm release, and it moves `ScalarCodeBlock` onto it.

### 1. Ready `@scalar/highlight` to publish (`3979f76`)

- Fixed three stale docs that would have shipped to npm: the `highlightBlock` JSDoc and the scope-vocabulary comment still named the old `ch-` class prefix (the renderer emits `shl-`), and `style.css` pointed at `test/style.test.ts` instead of `src/style.test.ts`. The first two land in `dist/*.d.ts`, so consumers would have seen the wrong class name on hover.
- Added the npm/Discord badges, Community section and License wording the other published READMEs carry, plus light/dark screenshots.

Packaging itself already matched the repo — `files`, `engines`, `sideEffects`, the `tsconfig.build.json` exclude list, and no `publishConfig` (access comes from `.changeset/config.json`). `npm pack --dry-run` is clean: dist + README + package.json, no tests, no bench, no src.

> Note: this commit originally set the version to `0.0.0` so the pending changeset would debut the package at `0.1.0`. The release in #9901 shipped `@scalar/highlight@0.2.0` and consumed that changeset first, so the merge with `main` takes the published version and that part is now moot.

### 2. Migrate `ScalarCodeBlock` (`34dd6e4`)

- Imports `syntaxHighlight` and `standardLanguages` from `@scalar/highlight/compat`, and registers grammars via `@scalar/highlight/all`.
- Keeps the same `pre > code.hljs.language-x` envelope and `hljs-*` token classes, so `code.css` and every consumer of the rendered string are untouched. `standardLanguages` keeps its key set, so `StandardLanguageKey` — and therefore `ScalarCodeBlock` and `ScalarVirtualCodeBlock` prop types — is unchanged.
- `ScalarMarkdown` still uses `@scalar/code-highlight` for its remark/rehype pipeline, which the compat layer deliberately does not cover, so the dependency stays.

### 3. Grammar fixes the migration surfaced

Rendering the real stories through the new highlighter turned up three JavaScript scoping bugs. All three are fixed here, with unit tests in `packages/highlight/src/langs/javascript.test.ts`.

- **An identifier before `<` read as a generic call** (`19adfe1`). `count` in `if (count <= max)`, `b` in `b <= 10` and the text `hi` in `<button>hi</button>` were all painted as function names. The generic branch now requires an argument list that closes and is followed by `(`, so `useState<number>(0)` still scopes and a comparison does not.
- **A name bound to a function went unscoped** (`19adfe1`). Only a name sitting directly in front of a `(` was recognised, so `const compare = (a, b) => …` and `const greet = function () {}` lost the colour `function compare() {}` already had.
- **JSX element text ran through the expression rules** (`2a29981`). Nothing changed state at an opening tag's `>`, so the prose in `<button>Click me to open the Api Client</button>` had `Click`, `Api` and `Client` each painted as a type — and `</button>` was not recognised as a tag at all, because a tag may only open where the character before `<` is not a letter, and a closing tag has element text directly in front of it. Children are now their own state.

### 4. A line height bug the JSX fix exposed (`cb7c24c`)

`code.css` declared `line-height` on `:where(code.hljs) *`. That selector reaches elements and never text nodes, and a highlighter emits a token span only where it has a scope to give — so a line of *entirely* unscoped text carried no element, missed the rule, and rendered about two pixels shorter than its neighbours.

Line numbers hid it. That mode wraps every line in `span.line`, which is an element, so the same source rendered evenly with a gutter and unevenly without one — which is exactly why `Base` and `Bordered`, both built from `contentJs`, disagreed. Fixing the JSX scoping above made the prose line unscoped and so made the existing bug visible.

The declaration now sits on `code.hljs` itself, so it sets the line box for every line and the spans inherit it. Ink height of the committed baselines:

| | `main` | before | after |
|---|---|---|---|
| `base.png` | 846 | 842 | 860 |
| `bordered.png` | 895 | 895 | 895 |

The new baseline is *taller* than main's rather than equal to it, because the same bug also shortened blank lines and they are now full height — which is what the line-numbered baseline has always shown. Line pitch in `base.png` is now a uniform ~36 device px, matching `bordered.png`.

## Bundle size — this PR grows the API Reference bundle

The per-package figures above are real, but **they do not translate into a saving here, and this is what `relative-ci` is flagging.** `@scalar/code-highlight` is not removed by this change — `ScalarMarkdown` still needs it for its remark/rehype pipeline — so the API Reference standalone bundle now carries *both* highlighters.

Measured by building `packages/api-reference` `build:standalone` at three commits and gzipping `dist/browser/standalone.js`:

| commit | gzipped | vs merge-base |
|---|---|---|
| `c049841` (merge-base) | 1,078,679 B | — |
| `59a439b` (`main` at the time) | 1,078,679 B | ±0 B |
| `520b21f` (this branch) | 1,111,812 B | **+33,133 B (+3.07%)** |

`main` and the merge-base were byte-identical, so the commits this branch was behind accounted for none of it: the whole +33 kB is this PR. Confirmed by probing the built bundle — `jsx-children` (a `@scalar/highlight` state) appears on this branch and not on `main`, while `lowlight` and `registerAlias` appear identically on both. Nothing was dropped; something was added.

> These figures were measured before `main` was merged in, so the absolute byte counts are anchored to an older `main`. The attribution is unaffected — the delta is `@scalar/highlight` being added to a bundle that still ships lowlight — and `relative-ci` reports the same status on later heads.

The saving lands only once `ScalarMarkdown` moves too, which needs `@scalar/highlight` to grow a remark/rehype integration. That is a separate piece of work and is **not** in this PR. Worth deciding whether to take the +33 kB now and the saving later, or to hold this until the markdown side is ready.

## Token changes

`@scalar/highlight`'s scope vocabulary is finer than highlight.js's, so some tokens now reach `code.css` rules that lowlight's output never hit.

**Visible — colour changes:**

- JSON `true` / `false` / `null` render green rather than purple. highlight.js nested an `hljs-keyword` span *inside* an `hljs-literal` span, so the inner keyword colour won; we emit a single `hljs-literal` span, which picks up the green `code.css` already defines for literals. This one shows on every JSON response body.
- HTTP header values are coloured where they were previously unstyled.
- Python docstrings scope as comments instead of strings.

**Markup-only — no visual change:**

- Punctuation (`.`, `()`) gets its own `hljs-punctuation` spans, and adjacent same-scope ranges merge (a closing brace followed by a comma becomes one span). `code.css` defines no `.hljs-punctuation` rule and no `.language_` rule, so both render identically to before. The two exact-markup assertions in `ScalarCodeBlock.test.ts` are updated for this.

## Visual

The committed Playwright baselines are the artifact, so before/after is `main`'s PNG against this branch's:

| | |
|---|---|
| **Before** (`main`) | ScalarCodeBlock Base on main |
| **After** (this branch) | ScalarCodeBlock Base on this branch |

What moves in this story: `Wrapper` picks up a colour as a name bound to an arrow function, `Click me to open the Api Client` goes back to plain text at full line height, and `</button>` renders as a closing tag instead of three loose operators.

## Snapshots

Regenerated in the `scalarapi/playwright-runner:1.62.1` container:

```bash
pnpm --filter @scalar/components test:e2e:update
```

`base.png` and `bordered.png` in `520b21f` for the token changes, then `base.png` again in `cb7c24c` for the line height. Regenerating the whole suite each time confirmed the blast radius: 203/203 pass and no other component baseline moves. The `api-reference` baselines are unaffected.

Note for whoever touches this next: #10103 added `--font-render-hinting=none --disable-skia-runtime-opts --force-color-profile=srgb` to `packages/api-reference`'s Playwright config and regenerated its baselines. Those flags are **not** on `packages/components`, so the component baselines here match the components config as it stands. Adding them there later will require regenerating all 203 component snapshots.

## Verification

The `PR` workflow run on `b1fab15` — this branch with `main` merged in — concluded `success`:

- Snapshots: components **203/203**, api-reference e2e **218/218**, scalar-app 6 passed / 2 skipped. The merge moved no baseline; the full component suite was also regenerated locally against the merged tree and produced no change.
- `build`, `types`, `knip`, `format`, `lint`, `test-integrations`, `test-server-side-rendering`, `test-proxy-server`, all `test-js` shards, CodeQL, dependency-review: green.
- Locally: `@scalar/components` 553/553 unit tests; `@scalar/highlight` 1321 passed / 2 skipped, including the compat suite against the real lowlight pipeline.
- `test-cdn-snapshots`: job green. It reports 12 visual diffs; that suite is non-blocking and compares against the last *released* CDN build, so it flags any rendering change until it ships. Eleven arrived with `main`, and one is the line height fix.

The one status that is not green is `relative-ci`, for the bundle-size reason documented above.

## Checklist

- [x] I added a changeset.
- [x] I updated tests (exact-markup assertions reflect the new token boundaries; new grammar tests cover the three scoping fixes).
- [x] I updated documentation (README badges, screenshots, JSDoc, comments, types).

https://claude.ai/code/session_01NxHFt5aoffmQhXDDhuDGWd


amritk:
highlight_light
highlight_dark

amritk:
One snapshot still red on `034ff76`, and I can't regenerate it from my environment — flagging it so it doesn't sit unexplained.

**Failing:** `test-component-snapshots` → `ScalarCodeBlock.e2e.ts › ScalarCodeBlock › Ligatures` (1 of 203; the other 202 pass).

**It belongs to this PR.** The `Ligatures` story is a JavaScript sample built from `!==`, `>=` and `<=`, which is exactly the scoping the `@scalar/highlight` migration changes. It is also the only `ScalarCodeBlock` baseline that was never regenerated — `base.png` and `bordered.png` were updated in `be0894d`, but `ligatures.png` still dates from `3bf6d90`, before the migration.

It passed on `be0894d` and `fdf2de5` and started failing on `029519a`, so it was most likely sitting just under the `maxDiffPixelRatio: 0.001` threshold and tipped over. It has now failed twice, including on a freshly merged head, so it is not a flake.

**What blocks me fixing it:** I cannot reproduce CI's rasterization. I ran the official `scalarapi/playwright-runner:1.59.1` image with the matching chromium-1217 build on matching x86_64, including CI's own topology of running the whole job inside the container. Using untouched `ScalarButton` as a control, snapshots still differed by 27–495 pixels, ratio 0.01–0.03 against the 0.001 threshold — sub-pixel antialiasing on a component this PR does not touch. Baselines generated here would fail CI and bury the real change under font noise.

**The fix**, on a machine whose rendering matches CI:

```bash
pnpm --filter @scalar/components test:e2e:update
```

**Re-run status:** already done. Merging `main` into the branch (`034ff76`) re-ran the check, and it failed identically, which is what rules out the threshold-margin explanation.

Everything else is green on this head: `test-references-e2e` 212/212, `test-cdn-snapshots` 26/26, plus build, types, knip, lint, format and the unit suites.

---

amritk:
Four checks went red on `5308375` for one upstream reason, unrelated to this PR. Flagging because one of them gates merge, and because it will be hitting every open PR right now.

**Cause.** Netlify published `@netlify/dev@5.0.4` depending on `@netlify/ai@^1.0.1`, but the newest published `@netlify/ai` is `1.0.0`. Every job that installs the Netlify CLI dies at dependency resolution:

```
ERR_PNPM_NO_MATCHING_VERSION  No matching version found for @netlify/ai@^1.0.1
This error happened while installing the dependencies of netlify@27.4.0
 at @netlify/dev@5.0.4
The latest release of @netlify/ai is "1.0.0".
```

**Affected:**

| Check | Tests | Why it is red |
| --- | --- | --- |
| `test-e2e-scalar-app` | 6 passed, 2 skipped | Playwright report upload — **required job** |
| `test-cdn-snapshots` | 26 passed | Playwright report upload |
| `deploy-previews / deploy-components` | — | `pnpm install -g netlify` |
| `deploy-previews / deploy-examples` | — | `pnpm install -g netlify` |

In both test jobs the suite itself passed and the PR comment was posted; only the step that publishes the HTML report to `scalar-snapshots.netlify.app` failed. I read the failing step directly on `deploy-components` and `test-cdn-snapshots`; on `test-e2e-scalar-app` the tests and the report-comment step both succeeded, leaving the same upload step as the only candidate.

**Not re-running it.** This is a semver range with no satisfying published version, so it resolves identically every attempt. It clears when Netlify publishes `1.0.1` or relaxes the range.

If it persists and CI needs to be green sooner, pinning the CLI in those workflow steps (`netlify@27.3.x`) would sidestep it — but that is a repo-wide CI change and does not belong in this PR.

---

amritk:
Correcting my previous comment: **pinning the CLI to `netlify@27.3.x` does not work.** Do not spend time on it.

The broken range is transitive and floating. Every published `netlify@27.x` depends on `@netlify/dev@^5.0.1`, which resolves to the newest 5.x — `5.0.4`, the broken one:

| `netlify` | → `@netlify/dev` | resolves to | → `@netlify/ai` |
| --- | --- | --- | --- |
| `27.3.0` | `^5.0.1` | `5.0.4` | `^1.0.1` ❌ |
| `27.3.1` | `^5.0.1` | `5.0.4` | `^1.0.1` ❌ |
| `27.4.0` | `^5.0.1` | `5.0.4` | `^1.0.1` ❌ |

(`@netlify/dev` publishes `5.0.0`, `5.0.1`, `5.0.4` — no `5.0.2`/`5.0.3`. `5.0.1` asks for `@netlify/ai@^1.0.0` and resolves fine; `5.0.4` asks for `^1.0.1`, which does not exist.)

So no CLI version is safe. Sidestepping it requires forcing `@netlify/dev` down to `5.0.1`, and a workspace `pnpm.overrides` will not reach `pnpm install -g` / `pnpx` — the global install resolves independently of this repo's manifest.

**Also worth separating the two kinds of failure.** The deploy-preview jobs genuinely need the CLI. The four failing test jobs do not — they only use it to publish an HTML Playwright report, *after* the tests have already passed. A report upload to a third-party host should not be able to fail a test job, least of all a required one. Adding `continue-on-error: true` to the four `Deploy Playwright Report to Netlify` steps in `.github/workflows/pr.yml` (~lines 335, 383, 430, 476) fixes this outage and permanently removes npm-registry availability from the merge gate.

That is a repo-wide CI change and still does not belong in this PR — noting it here so the next person hitting this does not repeat the pinning dead end.

Re-run confirmed deterministic: `test-e2e-scalar-app` was re-run on `034ff76` and failed identically at the same step, tests passing both times.

---

packages/api-reference/test/snapshots/models.e2e.ts.snapshots/scalar-galaxy-model-discriminator.png
hwkr, line 1: @amritk these don't seem like snapshots that should change from this

packages/api-reference/test/snapshots/operations.e2e.ts.snapshots/scalar-galaxy-classic-request-example.png
hwkr, line 1: Yay purple!

packages/api-reference/test/snapshots/operations.e2e.ts.snapshots/scalar-galaxy-request-responses-open.png
hwkr, line 1: Looks like some underlines creeping in

packages/components/src/components/ScalarCodeBlock/snapshots/base.png
hwkr, line 1: The highlighting looks good but we lost ligatures and there's a line height shift before `<ApiClientReact />`

Probably css issues

packages/components/src/components/ScalarCodeBlock/snapshots/bordered.png
hwkr, line 1: No shift but also no ligatures

packages/api-reference/test/snapshots/models.e2e.ts.snapshots/scalar-galaxy-model-discriminator.png
amritk: sub pixel diff? looks exactly the same

packages/components/src/components/ScalarCodeBlock/snapshots/ligatures.png
hwkr, line 1: This file name sounds like it should test ligatures on but I think it tests off. Probably fine

hwkr (changes requested):
Looks like some small regressions in the snapshots (mostly the ligatures)

packages/api-reference/test/snapshots/operations.e2e.ts.snapshots/scalar-galaxy-classic-request-example.png
amritk: yea fixed it in another PR but merged it in here to fix the screenshots

packages/api-reference/test/snapshots/models.e2e.ts.snapshots/scalar-galaxy-model-discriminator.png
hwkr: Yeah try removing it from the PR and see if CI still passes

amritk:
Thanks for the careful pass @hwkr — I measured all five snapshots, and the short version is that two of them read the opposite way round, and the ligature and underline changes are corrections rather than regressions.

## The image diffs read inverted

The purple and the underlines are on the **base** side, not the head side:

| snapshot | main (base) | this PR (head) |
| --- | --- | --- |
| `classic-request-example` | `curl` purple, `POST` cyan | both plain grey |
| `request-responses-open` | `BadRequestError` underlined | no underline |

So the purple was being *lost*, not gained. That was a real regression — `curl` was mapped to the bash grammar, which scopes neither the command nor the method. **#10056 fixes it and is now merged into this branch**, so `curl` goes back to `hljs-keyword` and `POST` to `hljs-symbol`. Those two baselines now record the old grey output and need regenerating.

## Ligatures are off by design, and this PR did not change that

`ScalarCodeBlock.vue` disables them, and the diff does not touch that rule:

```css
/** Disable font ligatures so code renders with literal characters */
.scalar-code-block pre {
  font-variant-ligatures: none;
}
```

You spotted the tell yourself on the `ligatures.png` thread. That story exists to prove ligature characters render *literally*, and its doc comment says so. The oddity is that on `main`, two baselines of the same component disagree under that one rule:

- `main`'s `ligatures.png` renders `=>`, `!==`, `>=` literally
- `main`'s `base.png` renders `⇒`, `◇` as ligatures

Both cannot be right. Since the rule is present and unchanged on both branches, `base.png` on main is a stale artifact from before ligatures were disabled, and the regenerated baseline here is the one that matches the CSS.

## The line-height shift is a one-off offset, not line-height

Measuring every line in `base.png`:

- line box heights are **identical** in both (25/26/27 px)
- lines 0–8 sit at **identical** positions
- from `<ApiClientReact` down, the head sits **4–6 px higher**, and the offset does not accumulate

A line-height change would compound down the block; this does not, so it is something above that element rendering slightly shorter.

## The underline going away is correct

Both schemas are hidden in the Galaxy document:

```yaml
BadRequestError:
  x-scalar-ignore: true
ForbiddenError:
  x-scalar-ignore: true
```

`isModelLinkable` returns `!isHidden(schema)`, so both are non-linkable, and `SchemaPropertyHeading` renders the `v-else` plain-text branch instead of `LinkButton` — which is where the underline comes from. Plain text is the intended output; main's underlined baseline predates that behaviour. Both baselines pass CI only because the difference is ~2,800 px ≈ 0.075%, under the 0.001 `maxDiffPixelRatio`.

The other 13 changed regions in that file are the dotted underline under `Example`, present in **both** images with only a sub-pixel shift in dot phase.

## `model-discriminator`

You were right, and so was @amritk — exactly **1 pixel** of 2,027,392. Removed from the PR in 5356ac9.

## Where that leaves it

- `model-discriminator` — reverted ✅
- `request-example` / `classic-request-example` — need regenerating now that #10056 is in; the purple comes back
- `base.png` / `bordered.png` / `ligatures.png` — correct as regenerated; ligatures were already meant to be off
- `request-responses-open` — correct as regenerated

The one genuinely open item is the 4–6 px offset, which I have not root-caused yet.

---

amritk:
`test-component-snapshots` is red on `2092fd0` — **1 failed, 202 passed**, `ScalarCodeBlock › Ligatures`. It is expected, and it corrects something I got wrong in [my earlier comment](https://github.com/scalar/scalar/pull/9939#issuecomment-5457019937).

## I had the ligature question backwards

I told @hwkr that ligatures were off by design and that this PR did not change that. The first half was wrong. `--scalar-font-code` leads with JetBrains Mono, which ships coding ligatures; `font-variant-ligatures: none` was switching off the thing the font is there for. `2092fd0` sets it to `normal` in `ScalarCodeBlock.vue`, and in `@scalar/highlight`'s `style.css` so the component and the package theme agree.

The run history is the proof that the change lands where intended:

| commit | rule | `Ligatures` |
| --- | --- | --- |
| `ef0d85a` (baseline regenerated) | `none` | ✅ pass |
| `2092fd0` (flipped) | `normal` | ❌ fail |

Nothing else about that story changed between the two. The committed baseline records ligatures off; the render now has them on.

## Why this also explains the two baselines that disagreed

`base.png` and `bordered.png` **pass** on this run, which I had expected to fail. `contentJs` contains only three `=>` plus `<>` and `</>`, in an image of roughly 1.2M pixels, so the ligature delta lands under `maxDiffPixelRatio: 0.001`. The `Ligatures` story packs ten of them into a much smaller canvas.

That is the missing piece from the earlier thread. It means two `ScalarCodeBlock` baselines on `main` **can** disagree indefinitely under one stylesheet: when `font-variant-ligatures: none` was added, `ligatures.png` moved and `base.png` did not, because `base.png`'s share of the change was below the threshold and CI had nothing to report. So the contradiction was never evidence of anything odd — just a sub-threshold baseline that went un-regenerated.

## Fixing it

Not a flake, and not re-running it: this is a deterministic baseline mismatch that the CSS change deliberately causes. I still cannot regenerate baselines from this environment — as [documented earlier](https://github.com/scalar/scalar/pull/9939#issuecomment-5442275073), an untouched control component differs by 27–495 pixels here against a 0.001 threshold, so anything I generate would fail CI and bury the real change in font noise.

One command on a machine whose rendering matches CI closes it, and only `ligatures.png` should move:

```bash
pnpm --filter @scalar/components test:e2e:update
```

`test-component-snapshots` is not in `required-jobs-pull-requests`, so it is not blocking merge in the meantime.

## The curl regression is fixed and confirmed

`test-e2e-api-reference` is **212/212 green** with both curl baselines reverted to `main`'s copies, so the purple `curl` / cyan `POST` render is back end-to-end after #10056. `request-responses-open` is likewise back to `main`'s copy and passes.

Against a fresh `origin/main`, this branch now changes exactly three baselines, all in one component:

```
packages/components/src/components/ScalarCodeBlock/snapshots/{base,bordered,ligatures}.png
```

Every api-reference baseline is `main`'s, unmodified.

---

amritk:
@hwkr your review was right, and #10061 answered it better than I did. This branch now matches main.

## Reverted the ligature change

`d0ea244` puts `font-variant-ligatures: none` back in `ScalarCodeBlock.vue` and in `@scalar/highlight`'s `style.css`, restores the story doc, and drops the changeset. [My earlier comment](https://github.com/scalar/scalar/pull/9939#issuecomment-5457706106) argued for turning ligatures on — that is superseded. Your `addStyleTag` approach keeps the product behaviour and still captures the glyphs, and the `maxDiffPixels: 100` budget explains the thing I could not account for: a handful of ligature glyphs in a large image fall under the project-wide ratio, so `base.png` could sit un-regenerated after the rule was added and CI would never say so.

## The snapshots that stayed red were a real regression

After the revert, three remained: `Base`, `Bordered` and `Ligatures` — every one of them a JavaScript story. That is not antialiasing. Dumping the scopes for those exact story samples:

| token | scoped as | should be |
| --- | --- | --- |
| `count` in `if (count <= max)` | `hljs-title function_` | plain variable |
| `b` in `b <= 10` | `hljs-title function_` | plain variable |
| `hi` in `<button>hi</button>` | `hljs-title function_` | plain text |
| `compare` in `const compare = (a, b) => …` | **unscoped** | function name |

Two bugs. An identifier followed by `<` was enough to read as the start of a generic call, which is right for `useState<number>(0)` and wrong for every comparison and every piece of JSX text shaped like it. And a name bound to a function went unscoped, because only a name sitting directly in front of a `(` was recognised.

That last row is the `compare` colour loss visible in the screenshots on this thread. It was a shipping regression, not snapshot noise — and regenerating the baselines before finding it would have frozen it in.

`19adfe1` fixes both: the generic branch now requires an argument list that closes and is followed by `(`, and `const name = () => …` (plus the `function`-expression, `async` and single-parameter forms) scopes the name. Adds `javascript.test.ts`, which the package did not have — 13 tests pinning each case, including the ones that must keep working. Suite is 1314 passing, 2 skipped.

## Still needs a baseline regeneration

The three baselines should be regenerated **after** `19adfe1`, since the grammar fix moves those colours again. I cannot produce CI-matching Playwright output from my environment — an untouched control component differs by 27–495 pixels here against this threshold — so this needs a machine whose rendering matches CI:

```bash
pnpm --filter @scalar/components test:e2e:update
```

## One correction

I previously said `Single Line` was proof the failures were colour rather than ligatures. Wrong — it contains three `--` and a `://`, all JetBrains Mono ligatures. It failed on the ligature flip and the revert fixed it. It now passes, along with `JSON String`.

Everything else is green on this head: api-reference e2e 212/212, CDN snapshots 26/26, scalar-app 6 passed, plus build, types, knip, lint, format and the unit suites.

---

amritk:
@hwkr your changes-requested review is addressed — `test-component-snapshots` is **203/203** on `520b21f`, and the whole `PR` run concluded `success`.

## Correcting myself

In [my last comment](https://github.com/scalar/scalar/pull/9939#issuecomment-5486458933) I said I could not produce CI-matching Playwright output from this environment and that the baselines needed regenerating on someone else's machine. **That was wrong, and I have now regenerated them here.**

The failures were never antialiasing. `@scalar/themes` fetches JetBrains Mono and Inter from `fonts.scalar.com` at render time, and those requests were dying under the concurrency of a test run — so text rendered with fallback metrics and *every* snapshot drifted, which is what made it look like an unfixable environment difference. Fetching the fonts one at a time works fine. With them cached locally and served to the browser, the environment reproduced CI exactly:

- A no-update run failed on precisely `Base` and `Bordered` and passed `JSON String`, `Ligatures` and `Single Line` — the same split CI reported.
- On `--update`, only `base.png` and `bordered.png` were rewritten. The other three regenerated **byte for byte**, and both new images kept their exact `1280×960` dimensions.

That byte-for-byte match is the part that makes these baselines trustworthy rather than merely plausible.

## A third grammar bug, found before regenerating

Probing the story samples first turned up one more, and it is the one your ligature comment was circling:

| token | was | now |
| --- | --- | --- |
| `Click` / `Api` / `Client` in `<button>Click me to open the Api Client</button>` | `hljs-title class_` | plain text |
| `</button>` | `<` `/` `>` as three operators, `button` unscoped | a closing tag |

Both from one cause: nothing changed state at an opening tag's `>`, so element children ran through the expression rules. And a tag may only open where the character before `<` is not a letter — the guard that keeps `a < b` and `Array<string>` out of JSX — which a closing tag never satisfies, because element text sits directly in front of it. `2a29981` gives children their own state; 7 more tests in `javascript.test.ts` (20 in the file, suite 1321 passing).

I fixed it *before* regenerating rather than after, since it moves the same pixels.

## Where it stands

- Ligature revert: done in `d0ea244`, matching #10061. Your `addStyleTag` approach is what the suite uses.
- Three JavaScript scoping regressions: fixed, with tests.
- Baselines: regenerated; only `base.png` and `bordered.png` move. No api-reference baseline changes.
- Green on `520b21f`: components 203/203, api-reference e2e 212/212, CDN 26/26, scalar-app 6 passed.

One thing the PR body now documents that is worth your eye before this goes up for review: this change **adds ~33 kB gzipped** to the API Reference standalone bundle rather than saving anything, because `ScalarMarkdown` still needs `@scalar/code-highlight` and the bundle ends up carrying both highlighters. Measured at three commits, with `main` and the merge-base byte-identical, so it is entirely this PR. That is what `relative-ci` is flagging.

Leaving it in draft and not re-requesting review — that call is @amritk's.

---

packages/components/src/components/ScalarCodeBlock/snapshots/base.png
hwkr: Line height still squished

packages/components/src/components/ScalarCodeBlock/snapshots/bordered.png
hwkr: Fixed

hwkr (changes requested):

packages/components/src/components/ScalarCodeBlock/snapshots/base.png
amritk: Fixed in `cb7c24c`. And you were right the first time — back in August you said "line height shift before `<ApiClientReact />`, probably css issues", and I argued it was a one-off offset rather than line-height. That was wrong. It was line-height, and it was CSS.

Your two comments this round are what pinned it down. `Base` and `Bordered` render the *same* `contentJs`; the only difference is `lineNumbers: true` on `Bordered`. So:

```css
:where(code.hljs) * { line-height: 1.4; }
```

`*` reaches elements, never text nodes. A highlighter only emits a span where it has a scope to give, so a line of entirely unscoped text carries no element, misses that rule, and falls back to the surrounding line height. Line numbers hide it, because that mode wraps every line in `span.line` — an element. Hence squished without a gutter, fine with one.

The line that goes unscoped here is the JSX prose, `Click me to open the Api Client`, which this PR stopped painting as a type. Measured on the committed baselines, ink height:

| | main | before | after |
|---|---|---|---|
| `base.png` | 846 | 842 | 860 |
| `bordered.png` | 895 | 895 | 895 |

`cb7c24c` moves the declaration onto `code.hljs`, so it sets the line box for every line and the spans inherit it. Line pitch in `base.png` is now a uniform ~36 device px, matching `bordered.png`.

Worth a look before you re-approve: this also fixes the **blank** lines, which were short for the same reason (67/68 → 72/73 device px across a blank line). That is why the new baseline is taller than main's rather than equal to it — main was the uneven render. Blank lines have always been full height in the line-numbered baseline you called "Fixed", so the two stories now agree.

Regenerated the whole `@scalar/components` suite to check the blast radius: 203/203 pass and `base.png` is the only file that moved.

---

packages/api-reference/test/snapshots/operations.e2e.ts.snapshots/scalar-galaxy-request-responses-open.png
amritk: This one is moot now — the branch carries no `api-reference` baseline changes at all. Every image under `test/snapshots/` is `main`'s copy, so this file is untouched by the PR and the underlines are whatever `main` renders.

For the record on the underlines themselves: both schemas are `x-scalar-ignore: true` in the Galaxy document, `isModelLinkable` returns `!isHidden(schema)`, so `SchemaPropertyHeading` takes its plain-text branch instead of `LinkButton` — and the underline comes from `LinkButton`. Plain text was the intended output; the underlined baseline predated that behaviour. Nothing here needed changing either way.

---

packages/components/src/components/ScalarCodeBlock/snapshots/ligatures.png
amritk: You were right, and the name now matches what it tests. When you wrote this, the story rendered with the component's own `font-variant-ligatures: none` still in force, so a file called `ligatures.png` was capturing ligatures switched *off*.

Your #10061 is what fixed it: the test turns them back on with `addStyleTag` before capturing, so the baseline now shows the glyphs the name promises. That approach is what this branch uses — the ligature CSS flip I had added is reverted, and `ScalarCodeBlock` keeps `font-variant-ligatures: none` as the product behaviour.

---

PR #9941, opened 2026-08-20
ci(release): publish to npm with trusted publishing

amritk:
## Problem

The release for [run 32402755278](https://github.com/scalar/scalar/actions/runs/32402755278/job/96536740490) built cleanly (54/54 turbo tasks) and then failed on the very first package it tried to publish:

```
npm notice name: @scalar/code-highlight
npm notice publish Signed provenance statement ...   <- sigstore/OIDC fine
npm error code E404
npm error 404 Not Found - PUT https://registry.npmjs.org/@scalar%2fcode-highlight
```

A 404 on a `PUT` is npm's masked 403 — the registry reports a write it will not authorize as "not found". `@scalar/code-highlight` obviously exists, so this is `NPM_TOKEN` no longer having write access, not a missing package. Because `code-highlight` is first in the topological order, `pnpm -r publish` died there and **none of the 50 packages shipped**; npm still shows `api-reference@1.65.1` / `api-client@3.16.1` from 13 Aug, the last successful release.

The token is a granular token documented in this workflow as created with `Custom Expiration: 01-01-2100`. npm no longer permits long-lived tokens and is actively restricting tokens that bypass 2FA for direct publishing — the run logs print that banner right before the failure. Minting a replacement buys us the same failure on a timer, so this moves the publish off tokens entirely.

## Solution

Switch the npm publish to [trusted publishing (OIDC)](https://docs.npmjs.com/trusted-publishers/), where npm trades this job's short-lived GitHub id-token for a package-scoped publish token at publish time.

- **Bump `changesets/action` v1.5.3 → v1.9.0.** This one is load-bearing, not hygiene. v1.5.3 unconditionally writes ``//registry.npmjs.org/:_authToken=${process.env.NPM_TOKEN}`` into `~/.npmrc`; with the secret removed that interpolates to the literal string `undefined` and puts a garbage bearer token in front of the OIDC exchange. v1.9.0 gates the write on `NPM_TOKEN` being set and logs `No NPM_TOKEN found, but OIDC is available - using npm trusted publishing`. Inputs are unchanged, and `GITHUB_TOKEN` from the env still takes precedence over the new `github-token` input, so the release PR keeps its app-token attribution.
- **Drop `NPM_TOKEN` from the publish step**, with a comment explaining why re-adding it would defeat the exchange.
- **Assert npm >= 11.5.1 before publishing.** `pnpm publish` has no OIDC support of its own — it packs a tarball and shells out to `npm publish`, and npm performs the token exchange. 11.5.1 is where that landed, which is newer than the npm bundled with some Node 24.x releases, so the workflow now upgrades instead of inheriting whatever `.nvmrc` resolves to.
- **Document the setup in the workflow header**, including the `npm trust` invocation, so the next person does not have to rediscover it.

`id-token: write` was already granted on both the caller job (`main.yml`) and `npm-publish` here, which npm requires for `workflow_call` — no permission changes needed.

## Testing

A changeset patches all 53 public packages, so the first release after this merges exercises the OIDC path across the whole workspace instead of only the packages that happen to have changesets pending. If a package's trusted publisher is missing or misconfigured, it shows up as that package's 404 rather than surfacing weeks later.

`scalar-app` is deliberately excluded: it is private, so it never reaches npm, and bumping it would trigger a ToDesktop build for nothing.

### Required before merging

The workflow half is inert until the registry side exists. **53** public workspace packages each need a trusted publisher (npm allows one per package); the 10 private packages need nothing.

```bash
# npm CLI >= 11.10.0; account 2FA required — use the "skip 2FA for 5 minutes"
# toggle on npmjs.com so the loop does not prompt 51 times. npm's bulk guidance
# is a 2s sleep between calls; ~51 packages then fits inside that 5m window.
for pkg in $(jq -r 'select(.private != true) | .name' packages/*/package.json integrations/*/package.json); do
  npm trust github "$pkg" --repo scalar/scalar --file main.yml --env production-npm --allow-publish --yes
  sleep 2
done
```

Two details worth flagging:

- **`--file main.yml`, not `release.yml`.** When a publish runs inside a `workflow_call`, npm validates the *calling* workflow's filename ([npm/documentation#1755](https://github.com/npm/documentation/issues/1755)). `main.yml` is the only caller and `publish-integrations.yml` only touches NuGet/PyPI/crates/Maven/DockerHub, so one config per package covers everything.
- **51 of the 53 can be configured now; `@scalar/highlight` and `@scalar/starlight` cannot.** They are not on npm yet, and npm requires a package to exist before it can carry a trusted publisher (initial publish over OIDC is still open: [npm/cli#8544](https://github.com/npm/cli/issues/8544)). They need one token-based or manual publish first — the recovery release creates them, since `pnpm -r publish` publishes any workspace version missing from the registry.

Keep the `NPM_TOKEN` secret in place until a release goes green, then revoke it.

## Checklist

- [x] I added a changeset.
- [ ] I added tests. 

### DIFF
--- a/.changeset/verify-trusted-publishing.md
+++ b/.changeset/verify-trusted-publishing.md
@@ -1,0 +1,57 @@
+---
+"@scalar/agent-chat": patch
+"@scalar/api-client": patch
+"@scalar/api-client-react": patch
+"@scalar/api-reference": patch
+"@scalar/api-reference-react": patch
+"@scalar/astro": patch
+"@scalar/asyncapi-upgrader": patch
+"@scalar/blocks": patch
+"@scalar/client-side-rendering": patch
+"@scalar/code-highlight": patch
+"@scalar/components": patch
+"@scalar/core": patch
+"@scalar/docusaurus": patch
+"@scalar/draggable": patch
+"@scalar/express-api-reference": patch
+"@scalar/fastify-api-reference": patch
+"@scalar/galaxy": patch
+"@scalar/helpers": patch
+"@scalar/highlight": patch
+"@scalar/hono-api-reference": patch
+"@scalar/icons": patch
+"@scalar/import": patch
+"@scalar/json-magic": patch
+"@scalar/mock-server": patch
+"@scalar/nestjs-api-reference": patch
+"@scalar/nextjs-api-reference": patch
+"@scalar/nextjs-openapi": patch
+"@scalar/nuxt": patch
+"@scalar/oas-utils": patch
+"@scalar/object-utils": patch
+"@scalar/openapi-parser": patch
+"@scalar/openapi-to-markdown": patch
+"@scalar/openapi-types": patch
+"@scalar/openapi-upgrader": patch
+"@scalar/postman-to-openapi": patch
+"@scalar/pre-post-request-scripts": patch
+"@scalar/react-renderer": patch
+"@scalar/release-notes": patch
+"@scalar/schemas": patch
+"@scalar/server-side-rendering": patch
+"@scalar/sidebar": patch
+"@scalar/snippetz": patch
+"@scalar/starlight": patch
+"@scalar/sveltekit": patch
+"@scalar/themes": patch
+"@scalar/ts-to-openapi": patch
+"@scalar/types": patch
+"@scalar/use-codemirror": patch
+"@scalar/use-hooks": patch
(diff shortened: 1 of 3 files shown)


hwkr (approved):
Looks good! CDN failures are expected until release (and unrelated)

PR #9945, opened 2026-08-21
fix(proxy): omit Cookie header when X-Scalar-Cookie is not provided

### LINKED ISSUE #9943: Scalr-proxy some WAFs reject that outright as an HTTP protocol violation ("Header name with no header value").
### What happens?

// Set the cookie - PATCHED: only when non-empty. The upstream code always
	// called Set() even with an empty string, producing a literal `Cookie:`
	// header with no value on every request that didn't set X-Scalar-Cookie
	// (i.e. every normal doc-load) - some WAFs reject that outright as an
	// HTTP protocol violation ("Header name with no header value").
	if xScalarCookie != "" {
		outreq.Header.Set("Cookie", xScalarCookie)
	}
	// Remove the X-Scalar-Cookie header
	outreq.Header.Del("X-Scalar-Cookie")

https://github.com/scalar/scalar/blob/main/projects/proxy-scalar-com/main.go

Set the cookie only when non-empty. 

### OpenAPI Document

https://github.com/scalar/scalar/blob/main/projects/proxy-scalar-com/main.go

sidsri14:
## Summary

Fixes #9943.

When making requests through `proxy-scalar-com` without setting `X-Scalar-Cookie`, `outreq.Header.Set("Cookie", xScalarCookie)` was called with `""`. In Go's `net/http`, this set an empty `Cookie: ` header on the outbound request sent across the wire. Certain WAFs (Web Application Firewalls) and upstream reverse proxies reject empty header values as an HTTP protocol violation (`400 Bad Request: Header name with no header value`).

Additionally, browser cookies sent to the proxy should not be leaked to the upstream API by default.

## Changes

1. In `projects/proxy-scalar-com/main.go`:
   - Excluded `Cookie` from the header copy loop alongside `Origin`.
   - Only set `Cookie` on `outreq.Header` when `X-Scalar-Cookie` is non-empty.
2. In `projects/proxy-scalar-com/main_test.go`:
   - Updated test assertion in `Does not forward any cookies by default` to check that the `Cookie` key is omitted from request headers.
   - Added unit test `Omits Cookie header completely when X-Scalar-Cookie is empty string`.

## Verification

- `go test -v .` in `projects/proxy-scalar-com` passed 100% (all 26 tests passed).

### DIFF
--- a/projects/proxy-scalar-com/main.go
+++ b/projects/proxy-scalar-com/main.go
@@ -388,18 +388,21 @@ func (ps *ProxyServer) executeProxyRequest(w http.ResponseWriter, r *http.Reques
 		return err
 	}
 
-	// Copy the headers but exclude Origin. It's not required and might confuse some target servers.
+	// Copy the headers but exclude Origin and Cookie.
+	// Origin is not required and might confuse some target servers.
+	// Cookies are excluded by default so browser cookies are not leaked;
+	// users need to specifically set cookies via X-Scalar-Cookie.
 	for key, values := range r.Header {
-		if !strings.EqualFold(strings.ToLower(key), "origin") {
+		lowerKey := strings.ToLower(key)
+		if lowerKey != "origin" && lowerKey != "cookie" {
 			outreq.Header[key] = values
 		}
 	}
 
 	// Users need to specifically set their cookies in X-Scalar-Cookie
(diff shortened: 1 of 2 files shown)


hanspagel (approved):
oh, yes!! thanks

PR #9946, opened 2026-08-21
docs(docs): fix header item in the full config example

hanspagel:
The header item in the full config example was missing `type` and used `url` instead of `to`. This fixes it to match the schema.


hwkr (approved):
Looks good!

PR #9947, opened 2026-08-21
docs(docs): document version selector and group header items

hanspagel:
Documents the `version-selector` and `group` header item types, which were missing from the navigation docs.


hwkr:
Consolidated into #9956 — the `group` and `version-selector` item types are documented there, each with its own properties table. Thanks!

PR #9948, opened 2026-08-21
docs(docs): document asyncapi navigation entries

hanspagel:
You can now add an AsyncAPI document to your docs with a navigation entry of `type: "asyncapi"` (it behaves the same as `openapi`). This documents that.

### DIFF
--- a/documentation/guides/docs/configuration/navigation.md
+++ b/documentation/guides/docs/configuration/navigation.md
@@ -311,6 +311,8 @@ Scalar supports three ways to generate API references from OpenAPI documents:
 2. the [Registry](../../registry/index.md), or
 3. remote URLs.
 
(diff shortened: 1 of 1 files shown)


hwkr:
Just moving this to draft until we release production

hwkr:
Pushed a small wording pass on `navigation.md` (ca63f13a0) — happy to revert any of it if you disagree:

- **Trimmed the `type` paragraph.** It explained that `asyncapi` is an alias for `openapi` and that channels render like operations — that is how it works internally, not something a user needs to decide anything on. Now it just says which value to use and that the format is detected automatically.
- **Reworded the `type` table row.** It read `Must be "openapi" or "asyncapi" (both behave identically; …)`, which restated the Type column and blew out the column width. Now it says what the property does, and the row is aligned with the rest of the table again.
- **Generalized the OpenAPI-only wording in the section.** Since these entries now take AsyncAPI too, `filepath`, `url`, and the Files / Registry / URL / Hiding / config prose all said "OpenAPI document" where they mean either format. Those now say "API document".

There is a follow-up branch coming with the same generalization applied to the rest of the docs (`configuration.md`, api-references, mock-server, registry, docs guides). Keeping it out of this PR so this one stays focused — will stack it on top.

hwkr (approved):
Looks good!

marclave (approved):

marclave (approved):

PR #9962, opened 2026-08-23
feat(plugin-mermaid): add @scalar/plugin-mermaid

bielu:
## Problem

#9567 asked for the same thing this does — Mermaid diagrams rendering from spec content instead of a plain code block — and got closed because bundling Mermaid into `@scalar/api-reference` added +3,157 kB to the core bundle, even with the dynamic `import()` that PR used. Dynamic import alone doesn't help when Mermaid still ships as part of the same package's `dist/` output.

[One of the comments on that PR](https://github.com/scalar/scalar/pull/9567#issuecomment-4768933277) suggested an opt-in extension point instead, so consumers who want diagrams can plug in their own renderer without the core bundle paying for it.

## Solution

Turns out no new extension point is needed — the plugin API already has one. `ApiReferencePlugin.extensions` lets a plugin register a component for any `x-*` specification extension name, and `SpecificationExtension` already renders it wherever the surrounding UI passes extensions through (operations, tags, `info`, …). This PR is a new, separate package that uses exactly that, registering a component for `x-mermaid`.

```ts
import { createApiReference } from '@scalar/api-reference'
import { createMermaidPlugin } from '@scalar/plugin-mermaid'

createApiReference('#app', {
  url: '/openapi.json',
  plugins: [createMermaidPlugin()],
})
```

```yaml
paths:
  /orders:
    post:
      x-mermaid: |
        sequenceDiagram
          Client->>API: POST /orders
          API->>Client: 201 Created
```

Because this is its own package:

- `mermaid` is a dependency of `@scalar/plugin-mermaid` only — `@scalar/api-reference`'s `package.json` is untouched, so nothing changes for consumers who don't install this.
- Even within this package, `mermaid` is dynamically `import()`-ed inside the rendering component itself, only once a document actually carries an `x-mermaid` value and the UI mounts a component for it — not at module load.
- The diagram adapts to light/dark color mode via `@scalar/use-hooks`' `useColorMode`, and Mermaid runs with `securityLevel: 'strict'` (same as #9567's approach) so diagram sources can't inject HTML or click handlers.

## What this doesn't do

This does not touch fenced ` ```mermaid ` code blocks inside markdown descriptions (`ScalarMarkdown`) — that was the other half of #9567's original scope. `ScalarMarkdown` has no plugin extension point today, so that would need its own, separate change. Happy to look at that too if there's interest, but wanted to keep this PR to the part the existing API already supports cleanly.

## Checklist

- [x] I added a changeset.
- [x] I added tests.
- [x] I updated the documentation (package README).

### DIFF
--- a/.changeset/plugin-mermaid.md
+++ b/.changeset/plugin-mermaid.md
@@ -1,0 +1,5 @@
+---
+'@scalar/plugin-mermaid': minor
+---
+
+feat: add @scalar/plugin-mermaid, an optional plugin that renders `x-mermaid` specification extensions as Mermaid diagrams via the existing plugin API. Mermaid is a dependency of this package only and is dynamically imported at render time, so the core `@scalar/api-reference` bundle is unaffected unless a consumer installs and registers this plugin.
--- a/knip.jsonc
+++ b/knip.jsonc
@@ -174,6 +174,10 @@
       "entry": ["src/index.ts"],
       "ignoreDependencies": ["vue"]
     },
+    "packages/plugin-mermaid": {
+      "entry": ["src/index.ts"],
+      "ignoreDependencies": ["vue"]
+    },
     "packages/helpers": {
       "entry": ["src/*/*.ts", "!src/*/*.test.ts"]
     },
--- a/packages/plugin-mermaid/README.md
+++ b/packages/plugin-mermaid/README.md
@@ -47,3 +47,13 @@ paths:
 
 The diagram renders in place, adapts to light/dark color mode, and re-renders on a color-mode
 change.
+
+## Styling
+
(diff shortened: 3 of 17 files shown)


hanspagel:
Not convinced we should add `x-mermaid`, we probably want to extend the Markdown rendering. Let me look into that.

bielu:
@hanspagel it was part which i implement for arazzo, i am using custom fork for now:).
Unless you have better way of doing it as custom bit for other schemas :)
image
I also look into making markdowns (have branch, can find later today)

hanspagel:
Thanks for working on this! I’ll pick this up in a fresh PR with support for Mermaid across Markdown rendering.

PR #9965, opened 2026-08-24
fix(api-reference): keep the variant's name when flattening a single-$ref allOf

### LINKED ISSUE #9964: `oneOf` options inherit the shared `allOf` base schema name
### What happens?

With `@scalar/api-reference` 1.66.1, a `oneOf` whose referenced variants each extend the same base through a single-member `allOf` renders every selector option with the base schema's name.

In the reproduction below, the variants are `Cat` and `Dog`, and both extend `Animal`. Scalar renders the selector options as:

Image

The selector should preserve and display the `oneOf` branch component names:

```
Cat
Dog
```

### OpenAPI Document

https://sandbox.scalar.com/e/H8s_H

lazerg:
## Problem

A `oneOf` selector shows the shared `allOf` base name for every variant, instead of each variant's own name. This happens when a variant extends a common base through a single-`$ref` `allOf`, a common inheritance pattern.

### Before

Before: the oneOf selector shows Animal for every variant

### After

After: the oneOf selector shows the variant's own name, Cat

## Solution

`optimizeValueForDisplay` flattens a variant's single-item `allOf` by merging the resolved base schema into it. The resolved base keeps its own `$ref` (so refs still work elsewhere), and that `$ref` was merged in last, overwriting the variant's own identity. The base's `$ref` is now dropped before merging, so the variant keeps its own name.

## Checklist

- [x] I added a changeset.
- [x] I added tests.

Closes #9964

### DIFF
--- a/.changeset/fix-9964-oneof-variant-name.md
+++ b/.changeset/fix-9964-oneof-variant-name.md
@@ -2,4 +2,4 @@
 '@scalar/api-reference': patch
 ---
 
-Fix oneOf selector labels showing the shared allOf base name instead of the variant's own name. When a oneOf branch extended a common base through a single-`$ref` allOf (a common inheritance pattern), flattening that allOf for display left the base schema's own `$ref` on the flattened variant, so the selector picked up the base's name for every branch instead of each branch's own name.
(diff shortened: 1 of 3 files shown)


hanspagel (approved):
yep!

PR #9967, opened 2026-08-24
feat(openapi-validator, asyncapi-validator, json-schema-validator): add standalone validation packages

hanspagel:
Splits validation into its own packages, all built on one shared engine.

New packages:

- `@scalar/json-schema-validator`: the shared engine. It validates any document against any JSON Schema and returns readable errors. Works on its own.
- `@scalar/openapi-validator`: validate an OpenAPI document without pulling in the whole parser.
- `@scalar/asyncapi-validator`: validate AsyncAPI 2.x and 3.x documents, using the official `@asyncapi/specs` schemas.

`@scalar/openapi-parser` still works the same. It now uses `@scalar/openapi-validator` under the hood. Both spec validators are thin layers on the shared engine: they provide the schemas and pick the version, and the engine does the actual validation.

The validators are schema-only. They check one document and do not resolve `$ref`s, so bundle or dereference first if your files span multiple documents.

The OpenAPI tests moved into `@scalar/openapi-validator`.

Screenshot 2026-08-25 at 12 27 53


hwkr:
Ran an adversarial review loop over this branch — read-only collectors fan out over the diff, a second independent pass tries to **refute** every finding, confirmed ones get fixed, repeat until a round turns up nothing new. It converged after four rounds.

Fixes are pushed to `claude/openapi-validator` (`183582689..0d17b08c5`, nine commits). Every claim below was verified by executing it, not just by reading the code. Six other candidate findings were refuted during verification and dropped.

## Three that mattered

**`throwOnError: false` did not hold in `@scalar/json-schema-validator`.** A schema Ajv could not compile (bad `pattern`, unresolvable `$ref`, unknown dialect) or a malformed YAML/JSON string threw straight through the documented option. The old `Validator` had `getAjvValidator()` inside its outer try/catch; the extraction lost that guard. Both paths now return `{ valid: false, errors }` and only throw when asked to.

**A failing document could return `{ valid: false, errors: [] }`.** `JSON_POINTERS_REGEX` only matches ASCII-word segments, so an error on a non-ASCII property name (`/日本語`) matched nothing and `makeTree` dropped it via `if (!paths) continue`. Now it gets its own node keyed by the full instance path — deliberately *not* the root node, which would have exposed it to the root's `required` pruning and swallowed it again.

**Every `pattern` failure rendered `Property "undefined" must match pattern …`.** Ajv only populates `propertyName` for the `propertyNames` keyword, never for `pattern`. Ported faithfully from `validation-errors/pattern.js`, and none of the 67 fixtures exercise it. The name now comes from the instance path — and a numeric segment is only treated as an array index when the document actually holds an array there, so `responses.404` keeps its name.

## Also fixed

- Two enum-pruning regressions in the rewrite: the sibling check only looked at a sibling's *direct* errors (the old `.some(getErrors)` was accidentally checking "does a sibling exist at all", since `[]` is truthy), and `parent && key` is falsy for the root node's `''` key, so root-level enum errors could never be pruned.
- `validate(doc, true)` threw `TypeError: Invalid value used as weak map key` — boolean schemas are valid JSON Schema but cannot key a `WeakMap`. They now skip the cache.
- A schema that failed to compile was recompiled on every call (~1.5 ms each, full Ajv construction before failing). Failures are cached alongside successes now.
- `validate-path-parameters.ts` swapped a local `isRecord` for `@scalar/helpers`' `isObject`, which additionally requires a plain-object prototype — that silently skipped whole path items in hand-constructed documents. Restored the original semantics.

Each fix has a regression test; the suites are at 22 / 71 / 10 / 269 across the four packages.

## One thing to look at

I bumped `@scalar/openapi-parser` from `patch` to **`minor`** in the changeset, and reworded it. `ErrorObject.path` widened from `string[]` to `string | string[]`, and `ValidationOutcome` / `OpenApiDocument` are no longer exported — both are fine changes, but the changeset claimed the parser "keeps working exactly as before", which is not quite true at the type level. Happy to be overruled on the bump.

## Left open on purpose

- **`throwOnError` now throws the prettified, pruned message** rather than the raw Ajv one — `"if must match \"else\" schema"` where the old code gave `"must be equal to one of the allowed values: query, header, path, cookie"`. Strictly less actionable, but which one it *should* surface is a design call, so I left it.
- **The pointer tokenizer.** Unmatched paths get a flat node instead of nesting per segment, so they skip parent/child pruning. Fixing it properly means replacing `JSON_POINTERS_REGEX` with a real JSON Pointer tokenizer, which would change error grouping for every document — too broad for this PR.
- **`@scalar/types` is a devDependency of `@scalar/openapi-parser`** but appears in its published `.d.ts`. Pre-existing (already true via `normalize` and `traverse`), so out of scope here, but external consumers would hit `TS2307`. Worth a follow-up.

Also confirmed along the way: the four vendored spec schemas are byte-for-byte semantically identical to the parser's originals (only reformatted), and the 140 moved fixtures are pure renames with zero content changes.

---

<sub>🤖 Posted by Claude — automated review run via Claude Code, kicked off and reviewed by @hwkr.</sub>

hanspagel:
> @scalar/types is a devDependency of @scalar/openapi-parser

addressed in #9990

hwkr (approved):
I'm happy with this but wouldn't hurt to have @DemonHa and @amritk sign off since it's a big-ish change

marclave (approved):

PR #9973, opened 2026-08-24
fix(api-reference): only render oauth scopes for `oauth2`/`openIdConnect` schemes

### LINKED ISSUE #9972: "OAuth scopes" section/label shown for non-OAuth2 security schemes (e.g. http/bearer)
### Description

The "OAuth scopes" section (introduced in #9683, translation key `authentication.scopes`) is rendered for **any** security scheme that has a non-empty scope array in a `security` requirement - not just `oauth2`/`openIdConnect` schemes as the code comment states.

`packages/api-reference/src/features/Operation/helpers/get-required-security.ts` (`getRequiredScopeGroups`) says:

> Scopes are only collected for OAuth2 / OpenID Connect schemes, so alternatives without any scopes (for example API-key-only auth) are omitted...

but the implementation doesn't check `scheme.type` at all:

```ts
for (const group of requiredSecurity.requirements) {
  const collected = new Set<string>()
  for (const scheme of group.schemes)
    for (const scope of scheme.scopes) collected.add(scope)
  if (collected.size > 0) groups.push([...collected])
}
```

It just collects whatever strings are listed under any `security:` requirement entry, regardless of whether the named scheme is `oauth2`, `openIdConnect`, or (per spec, where such an array isn't semantically OAuth scopes at all) `http`/`apiKey`.

### Reproduction

```yaml
openapi: 3.0.3
info:
  title: Repro
  version: "1.0"
paths:
  /ping:
    get:
      security:
        - bearerToken:
            - some:custom:scope
      responses:
        "200":
          description: OK
components:
  securitySchemes:
    bearerToken:
      type: http
      scheme: bearer
      bearerFormat: JWT
```

`bearerToken` is a plain HTTP Bearer/JWT scheme - there is no `oauth2`/`openIdConnect` scheme in the document at all.

### Actual behavior

The operation page renders a section titled **"OAuth scopes"** listing `some:custom:scope`, even though there is no OAuth2 flow, authorization URL, token URL, or OIDC discovery document anywhere in the spec.

### Expected behavior

Either:
- Only collect/render scopes for `oauth2`/`openIdConnect` schemes (matching the existing code comment), and skip the section entirely for `http`/`apiKey` schemes, **or**
- If deliberately supporting non-OAuth2 schemes carrying an informal scope-like list, use a scheme-type-aware, more generic label (e.g. "Required scopes") instead of hardcoding "OAuth scopes".

### Environment

- `@scalar/api-reference` `1.65.1` (also reproduced on the Docker image `scalarapi/api-reference:0.5.60`)
- Confirmed by inspecting the published package: `dist/features/localization/locales/en.js` → `authentication.scopes: "OAuth scopes"`, and `dist/features/Operation/helpers/get-required-security.js`

lazerg:
## Problem

The "OAuth scopes" section renders for any security scheme that has a non-empty scope array, not just `oauth2`/`openIdConnect` schemes. An `http` bearer scheme with scope-like strings in its `security` requirement gets an "OAuth scopes" section too, even though there's no OAuth2 flow or OIDC discovery document anywhere in the spec.

## Solution

`getRequiredScopeGroups` now checks the resolved scheme's `type` before collecting its scopes, skipping anything that isn't `oauth2` or `openIdConnect`.

## Checklist

- [x] I added a changeset.
- [x] I added tests.

Closes #9972

### DIFF
--- a/.changeset/fix-9972-oauth-scopes-non-oauth-schemes.md
+++ b/.changeset/fix-9972-oauth-scopes-non-oauth-schemes.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
+Fix the "OAuth scopes" section rendering for any security scheme with a non-empty scope array, including `http` and `apiKey` schemes. Scopes are only meaningful for `oauth2` and `openIdConnect` schemes, so the section is now skipped when the resolved scheme is of a different type.
--- a/packages/api-reference/src/features/Operation/components/OperationScopes.test.ts
+++ b/packages/api-reference/src/features/Operation/components/OperationScopes.test.ts
@@ -51,6 +51,19 @@ const scopedOrScopeFree: RequiredSecurity = {
   ],
 }
 
+/**
+ * One OAuth2 alternative with scopes, plus an HTTP bearer alternative that carries
+ * scope-shaped strings. Those strings are not OAuth scopes, so the bearer alternative is
+ * still scope-free and the OAuth scopes remain optional.
+ */
+const scopedOrScopeShapedNonOauth: RequiredSecurity = {
+  state: 'required',
+  requirements: [
+    { schemes: [{ name: 'oauth2', scheme: { type: 'oauth2', flows: {} }, scopes: ['read:items'] }] },
+    { schemes: [{ name: 'bearerToken', scheme: { type: 'http', scheme: 'bearer' }, scopes: ['some:custom:scope'] }] },
+  ],
+}
+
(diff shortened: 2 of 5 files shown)


lazerg:
The openapi-to-markdown circular-reference test timeout on this run looks unrelated to this change and flaky, since it passed cleanly on the previous run with no changes to that package. Pushing an empty commit to retrigger.

hanspagel (approved):
niice, thanks!

PR #9975, opened 2026-08-24
chore(docs): use an SVG for the SOC 2 footer seal

cameronrohani:
## What

Replaces the AICPA SOC 2 seal in the global footer with a vector version, and drops the PNG.

## Why

Weight and reuse — **not** sharpness. At 68×68 the 748 px PNG is already heavily oversampled, so it looks fine on high-DPI today. The actual problems are that it is the heaviest asset in the footer by a wide margin, and that it cannot be reused at a larger size without a new export. The GDPR badge sitting next to it is already an SVG, so this also makes the pair consistent.

| | Raw | Gzipped |
| --- | --- | --- |
| `aicpa-soc-2-seal.png` (before) | 70,748 B | 68,790 B |
| `aicpa-soc-2-seal.svg` (after) | 21,756 B | **8,850 B** |
| `gdpr-compliant-badge.svg` (for scale) | 2,557 B | 779 B |

## How

Traced from the licensed PNG already in the repo, rather than pulled from a third-party logo site, since the seal is a trademarked certification mark.

The geometry was measured sub-pixel first, so the structural parts are real SVG primitives rather than traced blobs:

- Outer disc — `<circle>` at (374, 370), r = 364.54 (circle fit stdev 0.09 px)
- Inner disc — `<circle>` r = 283.1, `#231f20`
- Ring gradient — a horizontal power curve (γ ≈ 1.37, `#000000` → `#57585a`), reproduced with 8 stops
- Only the type and the ™ are traced outlines, from a 4× supersample

Path data is emitted as integers in a 10× coordinate space (`viewBox="0 0 7480 7430"`), which is both shorter and more precise than decimals in a 748-unit space, then minified with SVGO.

## Verification

Rendered the SVG and the original PNG and compared per pixel, in two independent renderers:

| Renderer | Mean diff | Pixels differing by >8 |
| --- | --- | --- |
| Chrome / Skia | 1.34 / 255 | 2.35% |
| librsvg / cairo | 1.24 / 255 | 2.48% |

The residual is confined to glyph antialiasing edges. Under librsvg the ring gradient endpoints sample to `(1,1,1)` and `(83,84,86)` and the inner disc to `(35,31,32)`, matching the source PNG exactly.

## Notes

- **No CSS change needed.** The badge is 68×68 with `object-fit: contain` in `documentation/assets/styles.css`, and the SVG keeps the PNG's 748×743 aspect ratio.
- **No changeset.** `documentation/` is not `packages/`, `integrations/`, or `projects/`.
- Prettier infers no parser for `.svg`, so the file is hand-formatted to match the sibling `gdpr-compliant-badge.svg`.
- AICPA supplies an official EPS/SVG to organizations holding a valid report. Worth requesting as the canonical asset — this is a faithful trace, but it is still a trace.
- Deleting the PNG retires the public URL `scalar.com/security/aicpa-soc-2-seal.png`. Nothing in this repo referenced it, but it will 404 for any external consumer after deploy.


hwkr (approved):
Noice

PR #9978, opened 2026-08-25
ci: adopt the Chromium-only playwright-runner 1.62.1 image

hanspagel:
Follow-up to #9648. That PR flips to the new image, all in lockstep so the local `@playwright/test` and the container always speak the same protocol:

- workflow container `image:` tags (`pr.yml`, `main.yml`) → `1.62.1`
- `@playwright/test` catalog (+ lockfile) → `1.62.1`
- `DEFAULT_PLAYWRIGHT_VERSION` in `@scalar/helpers` → `1.62.1`

Size: 1622.8 MB → 1105.6 MB (amd64, compressed): ~517 MB / 32% smaller per job pull.

### DIFF
--- a/.changeset/playwright-runner-1-62-1.md
+++ b/.changeset/playwright-runner-1-62-1.md
@@ -1,0 +1,7 @@
+---
+'@scalar/helpers': patch
+---
+
+chore: bump the default Playwright runner image to 1.62.1
+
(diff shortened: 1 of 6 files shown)


hwkr (approved):
Cool beans, we should do this in Org too

hanspagel:
@hwkr there's a PR over in org, too :)

PR #9981, opened 2026-08-25
feat(client-side-rendering): default to the ESM build

hanspagel:
Makes the modern, code-split ESM build the default for the generated API Reference HTML, and keeps the classic UMD bundle available as an opt-out.

- **Default (new):** loads the ESM build (`/esm.js`, added in #9871) as a `<script type="module">`.
- **`cdn: '…/@scalar/api-reference'`:** classic UMD bundle (e.g. to pin a version) — unchanged.
- **`bundle: false`:** classic UMD bundle without picking a URL.
- **`bundle: '…/esm.js'`:** a specific ESM build.

Works everywhere `renderApiReference` is used (Express, Hono, Fastify, Next.js, etc.) — the option flows through the shared config.

### CSP handled automatically

We can't read the CSP header during HTML generation, but a `nonce` is a reliable signal of a strict, nonce-based CSP. The ESM build loads its chunks with native `import`, which can't carry a nonce, so under that policy those requests would be blocked. So **when a `nonce` is set, the UMD bundle is used automatically** — a single nonced `<script>` with no follow-up requests. Set `bundle: true` to force ESM if your CSP uses `'strict-dynamic'`.

### Size

Total download is ~unchanged — the win is that less JavaScript blocks the first paint.

| | UMD | ESM (default) |
| --- | --- | --- |
| JS before first paint | ~1,071 kB | **~700 kB** |
| Total JS | ~1,071 kB | ~1,090 kB |

(Scalar Galaxy doc, live CDN, real browser.) The UMD build is one big file that must finish before rendering; the ESM build defers the heavy interactive parts (API client modal + AI chat, ≈ 370 kB) until after first paint. Renders identically.

### DIFF
--- a/.changeset/client-side-rendering-esm-cdn.md
+++ b/.changeset/client-side-rendering-esm-cdn.md
@@ -1,9 +1,11 @@
 ---
 '@scalar/client-side-rendering': minor
+'@scalar/schemas': minor
+'@scalar/types': minor
 ---
 
-Load the API Reference from the ESM CDN build by default
+Add a `bundle` option to load the modern ESM build of the API Reference
 
-When no `cdn` is set, the generated HTML now loads `@scalar/api-reference` from the short ESM entry (`/esm.js`, added in #9871) as a `<script type="module">` and calls `createApiReference` directly, instead of the monolithic UMD bundle via the `window.Scalar` global. The bundle is code-split, so the heavy interactive parts load after the page has painted and less JavaScript blocks the first render.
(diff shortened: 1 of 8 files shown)


marclave (approved):

marclave (commented):
plz talk to @geoffgscott and @amritk before merging : )

amritk (approved):
Yep ESM should be the one moving forward, its going to get smaller and smaller

PR #9982, opened 2026-08-25
fix(api-reference): avoid SSR hydration errors

hanspagel:
Load the initial API description before hydrating server-rendered content. Existing content stays readable during slow or failed loads, and successful hydration reuses its existing DOM nodes. Browser preferences, deep links, and localhost-only controls apply after hydration.

For example, render this API description on the server and configure the browser with the same URL:

```yaml
openapi: 3.1.0
info:
  title: Example API
  version: 1.0.0
paths:
  /things:
    get:
      summary: List things
      responses:
        '200':
          description: OK
```

Previously, the browser could start with an empty document and replace the server content during hydration. Now `createApiReference(element, configuration)` prepares the document first, including external references, configured servers, and environment defaults. Failed preparation leaves the server HTML intact; replacing the configuration can recover, and destroying a pending instance prevents mounting.

Custom Vue SSR integrations can await the exported `prepareApiReference` helper and pass its result as the `prepared` prop. Plain `<ApiReference>` client mounts still require this explicit preparation when hydrating server content.

## Ticket

Part of #4458


marclave (approved):

hanspagel:
not sure, this is good. need to research/test a bit more.

cameronrohani (approved):

PR #9984, opened 2026-08-25
docs: add an RSS feed for the blog

hanspagel:
Adds a blog feed at `/blog/rss.xml` alongside the existing changelog feed. The blog has a `/blog` landing page with a subscribe button; existing blog links redirect to their new routes.

Each post gets frontmatter `date` copied from its existing date-prefixed filename and matching the date already displayed in the blog index, preserving the chronology rather than using the date of this change. The feed currently contains the 17 dated posts already linked from the public blog index, newest first. The undated blog index contributes no item. Only configured pages under `/blog` are selected; there are no draft routes in that set. Future dated pages added under that path will become feed items, while undated pages and hidden pages are excluded.

The released Scalar CLI now supports the multiple-feed configuration and `entries: "pages"`; the earlier release blocker no longer applies.

Excerpt from the generated local preview feed, with other item fields omitted:

```xml
<item>
  <title>We wrote up the Stainless wind-down</title>
  <pubDate>Sun, 16 Aug 2026 00:00:00 GMT</pubDate>
</item>
<item>
  <title>Scalar's 2025 Open Source Pledge</title>
  <pubDate>Sat, 11 Apr 2026 00:00:00 GMT</pubDate>
</item>
```

### DIFF
--- a/documentation/blog/2024-09-01-oss-pledge.md
+++ b/documentation/blog/2024-09-01-oss-pledge.md
@@ -1,3 +1,7 @@
+---
+date: 2024-09-01
+---
+
 # Scalar Joins the Open-Source Pledge
 
 ## Sep 1st, 2024
--- a/documentation/blog/2025-03-05-how-net-9-and-scalar-solve-the-problem.md
+++ b/documentation/blog/2025-03-05-how-net-9-and-scalar-solve-the-problem.md
@@ -1,3 +1,7 @@
+---
+date: 2025-03-05
+---
+
 # How .NET 9 and Scalar solve the problem of under-documented APIs
 
 Stop me if you’ve heard this one before:
--- a/documentation/blog/2025-03-12-how-we-sped-up-our-api-docs-25x.md
+++ b/documentation/blog/2025-03-12-how-we-sped-up-our-api-docs-25x.md
@@ -1,3 +1,7 @@
+---
+date: 2025-03-12
+---
+
 # How we sped up our API docs 25x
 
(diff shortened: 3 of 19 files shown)


marclave (approved):
probs good to pick up again : )

amritk (commented):
Two things keep this from being mergeable right now, and the PR body already names the first:

1. **`on hold` label**, plus the body's own note: "❌ Config validation fails until the CLI is released." So this is gated on an external release, not on review.
2. **CI is failing** — GitHub reports the branch as `unstable`, which is consistent with the config-validation failure above.

Nothing to do on the review side until the CLI ships and validation passes. When it does, ping me and I will look properly; it is a contained change (a second feed alongside the changelog one, plus `date` frontmatter on each post) and should be quick.

Two things worth checking at that point, noting them now so they are not rediscovered later:

- **Adding `date` to every blog post's frontmatter is the risky half of this diff**, not the feed config. 230 additions against 149 deletions across 19 files for what is described as "a second RSS feed" suggests most of the churn is in the posts themselves. Worth confirming the dates match each post's actual publication date rather than the date the frontmatter was added — a feed is the one place where a wrong date is immediately visible to subscribers, and it determines item ordering.
- **`entries: "pages"` means every post becomes a feed item, including any future ones.** If there are drafts or unlisted posts under the same path, they would be published to the feed. Worth confirming the path scoping excludes them.

The preview link in the description (`x4jaonxr7lyip3dsg8aw2--scalar.apidocumentation.com/blog/`) is a per-deploy URL that will not survive — if you want a reviewable artifact after the hold lifts, paste the generated `rss.xml` for a couple of posts into the description instead.

---

hanspagel:
the CLI was released - finally

PR #9986, opened 2026-08-25
refactor(types): migrate security-scheme + nanoid off zod to @scalar/validation

hanspagel:
Migrates the two runtime modules in `@scalar/types` that build zod schema values at module load — `utils/nanoid.ts` and `entities/security-scheme.ts` — onto the zero-dependency `@scalar/validation`.

These two modules were the only thing pulling the `zod@4.3.5` runtime path into the `@scalar/api-reference` standalone browser bundle. After the change that path is gone.

### Bundle impact (standalone.js)

| | before | after |
|---|---|---|
| raw | 3,806,019 B | 3,735,844 B (−68.5 KB) |
| gzip | 1,092,517 B | 1,074,157 B (−17.9 KB) |
| `zod@4.3.5` modules | 22 (~141 KB) | **0** |

The separate `zod@4.4.3` copy in the bundle comes from a different dependency and is out of scope here.

### zod is not fully removed from `@scalar/types`

`zod` stays a dependency of `@scalar/types` because the type-only api-reference configuration schemas (`api-reference/*.ts`) still use it for `z.infer`. Those are already tree-shaken out of the reference bundle, so keeping them on zod does not affect bundle size. This PR removes zod only from the **runtime** bundle path.

### Notes on behaviour mapping

`@scalar/validation` has no `.parse()`/`.brand()`/`.transform()`/`.catch()`, so a few things were mapped faithfully rather than 1:1:

- Exported schema value names and inferred type shapes are preserved, so downstream `import type` consumers are unchanged. The only runtime `.parse()`/`.safeParse()` callers were tests; they now use `coerce()`/`validate()`.
- The branded `uid` is preserved at the type level via a small `Brand<T, B>` helper (validation has no runtime brand).
- zod's `.default(() => nanoid())` becomes an `evaluate(...)` that generates an id for missing/too-short values.
- The `.catch()`/`.transform()` behaviours (enum fallbacks, HTTP scheme lower-casing, default oauth flows, `x-default-scopes` → `selectedScopes`) are reproduced with `evaluate` and a post-coerce transform.
- The exported types are hand-written concrete shapes rather than `Static<typeof schema>`, because deriving them from validation's deep `Static` conditional — and feeding the result to `PartialDeep` downstream — trips TypeScript's `TS2589` "excessively deep" budget. This mirrors how `@scalar/workspace-store` pairs an explicit `SecuritySchemeObject` union with its runtime schema.

### Checks

- `@scalar/types`: build (tsc) + type-check (tsgo) clean, 56 tests pass
- `@scalar/workspace-store`, `@scalar/api-client`, `@scalar/oas-utils`: type-check clean; auth/security tests pass (workspace-store 133, api-client auth-selector 342, oas-utils migration 131)
- knip clean on both touched workspaces

### DIFF
--- a/.changeset/security-scheme-validation.md
+++ b/.changeset/security-scheme-validation.md
@@ -1,0 +1,6 @@
+---
+'@scalar/types': patch
+'@scalar/api-reference': patch
+---
+
+Migrate the `nanoid` and `security-scheme` entity schemas from `zod` to the zero-dependency `@scalar/validation`. These were the only two runtime modules in `@scalar/types` that constructed zod schema values at module load, so this removes the `zod@4.3.5` runtime path from the `@scalar/api-reference` standalone browser bundle (22 modules, ~141 KB raw). The standalone bundle drops from 3,806,019 to 3,735,844 bytes raw (1,092,517 to 1,074,157 bytes gzip). The exported schema value names and inferred types are preserved, so downstream `import type` consumers keep working unchanged. `zod` remains a dependency of `@scalar/types` because the type-only api-reference configuration schemas still use it (they are already tree-shaken out of the bundle).
--- a/packages/oas-utils/package.json
+++ b/packages/oas-utils/package.json
@@ -54,6 +54,7 @@
     "yaml": "catalog:*"
   },
   "devDependencies": {
+    "@scalar/validation": "workspace:*",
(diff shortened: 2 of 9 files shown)


hanspagel:
we don't need this

PR #10024, opened 2026-08-26
fix(api-client): keep the composition selection when reopening the same operation

### LINKED ISSUE #10023: api-client: "Test Request" overwrites an edited request body when the operation is already open
## Summary

Clicking "Test Request" for an operation that is already selected in the client sidebar can wipe out an edited request body. The cause is how `requestBodyCompositionSelection` gets applied on `ui:open:client-modal`: the sidebar's `handleSelectItem` skips routing when the operation is already selected, but the composition selection still gets overwritten regardless. `RequestBody.vue`'s watcher then reads "operation unchanged, selection changed" as a manual oneOf/anyOf branch switch and regenerates the body from the schema, discarding the edit.

## Steps to reproduce

1. Open a reference page for an operation whose request body schema has `oneOf`/`anyOf` with more than one branch.
2. Click "Test Request". The operation opens in the client and becomes selected in the sidebar.
3. Edit the request body so it no longer matches the schema-generated example.
4. Without closing the modal, go back to the reference page and view the same operation's example for a different composition branch (so the doc page's own `requestBodyCompositionSelection` for that operation changes).
5. Click "Test Request" again for the same operation.

Expected: the edit from step 3 survives, since the same operation is still open.
Actual: the edit is gone, replaced by the schema-generated example for whichever branch was showing in step 4. Ctrl+Z only recovers the previous example, not the edit, because two `update:value` emits happen back to back and undo only reverts the second one.

## Root cause

- [`TestRequestButton.vue`](https://github.com/scalar/scalar/blob/main/packages/api-reference/src/features/test-request-button/TestRequestButton.vue) emits `ui:open:client-modal` with the operation `id` and whatever `requestBodyCompositionSelection` the doc page currently has selected.
- [`modal-events.ts`](https://github.com/scalar/scalar/blob/main/packages/api-client/src/v2/features/modal/modal-events.ts#L82-L99) handles that event: it calls `sidebarState.handleSelectItem(targetId)`, then unconditionally runs `requestBodyCompositionSelection.value = nextRequestBodyCompositionSelection`. The comment above that line says it applies "after routing", which assumes routing always happens.
- It doesn't. [`use-modal-sidebar.ts`](https://github.com/scalar/scalar/blob/main/packages/api-client/src/v2/features/modal/hooks/use-modal-sidebar.ts#L157-L162) returns early, without calling `route()`, when the target is already selected:

  ```ts
  if (state.isSelected(id)) {
    state.setExpanded(id, !state.isExpanded(id))
    return
  }
  ```

  `path`, `method`, and `exampleKey` stay exactly as they were, while `requestBodyCompositionSelection` still gets the new value.
- [`RequestBody.vue`](https://github.com/scalar/scalar/blob/main/packages/api-client/src/v2/blocks/request-block/components/RequestBody.vue#L240-L280) watches `requestBody`, `exampleKey`, and the composition selection as a group, and only skips regenerating the body when `operationChanged` is true. Since the operation is the same but the selection changed, this looks exactly like a user manually switching branches within one operation, so it calls `getSchemaExampleFromBody(...)` and overwrites the edit.

## Minimal reproduction

This calls `initializeModalEvents` directly and fires two `ui:open:client-modal` events for the same operation id with different `requestBodyCompositionSelection` values, the same sequence a doc page produces when "Test Request" is clicked twice for one operation.

```ts
import { useModal } from '@scalar/components/modal'
import { createWorkspaceStore } from '@scalar/workspace-store/client'
import type { OpenApiDocument } from '@scalar/workspace-store/schemas/v3.1/strict/openapi-document'
import { describe, expect, it, vi } from 'vitest'
import { computed, nextTick, ref } from 'vue'

import 'fake-indexeddb/auto'

import { initializeModalEvents } from './modal-events'
import { useModalSidebar } from './hooks/use-modal-sidebar'

const getDocument = (overrides: Partial<OpenApiDocument> = {}): OpenApiDocument => ({
  openapi: '3.0.0',
  info: { title: 'Test API', version: '1.0.0' },
  paths: {},
  'x-scalar-original-document-hash': 'test-hash',
  ...overrides,
})

describe('modal-events repro', () => {
  it('overwrites requestBodyCompositionSelection even when the operation was already selected (no re-routing)', async () => {
    const store = createWorkspaceStore()
    await store.addDocument({
      name: 'test-doc',
      document: getDocument({
        paths: {
          '/pets': {
            post: {
              operationId: 'createPet',
              requestBody: {
                content: {
                  'application/json': {
                    schema: {
                      oneOf: [
                        { type: 'object', properties: { name: { type: 'string' } } },
                        {
                          type: 'object',
                          properties: { name: { type: 'string' }, hasRings: { type: 'boolean' } },
                        },
                      ],
                    },
                  },
                },
              },
            },
          },
        },
      }),
    })
    store.update('x-scalar-active-document', 'test-doc')

    const documentSlug = computed<string | undefined>(() => 'test-doc')
    const path = ref<string | undefined>(undefined)
    const method = ref<'get' | 'post' | undefined>(undefined)
    const exampleName = ref<string | undefined>(undefined)

    const routeSpy = vi.fn((payload: { path?: string; method?: string; example?: string }) => {
      path.value = payload.path
      method.value = payload.method as 'get' | 'post' | undefined
      exampleName.value = payload.example
    })

    const sidebarState = useModalSidebar({
      workspaceStore: store,
      documentSlug,
      path: computed(() => path.value),
      method: computed(() => method.value),
      exampleName: computed(() => exampleName.value),
      route: routeSpy,
    })

    // Capture the handlers `initializeModalEvents` registers, so we can call them directly
    // the same way the real WorkspaceEventBus would when `ui:open:client-modal` is emitted.
    const handlers: Record<string, (...args: any[]) => void> = {}
    const eventBus = {
      on: vi.fn((event: string, handler: any) => {
        handlers[event] = handler
        return vi.fn()
      }),
      once: vi.fn(() => vi.fn()),
      off: vi.fn(),
      onAny: vi.fn(() => vi.fn()),
      offAny: vi.fn(),
      emit: vi.fn(() => null),
      flushDebouncedEmits: vi.fn(),
    } as any

    const requestBodyCompositionSelection = ref<Record<string, number>>({})
    const modalState = useModal()

    initializeModalEvents({
      eventBus,
      isSidebarOpen: ref(false),
      requestBodyCompositionSelection,
      sidebarState,
      modalState,
      store,
    })

    const operationEntry = sidebarState.getEntryByLocation({
      document: 'test-doc',
      path: '/pets',
      method: 'post',
    })
    expect(operationEntry).toBeDefined()
    const operationId = operationEntry!.id

    // 1st click on "Test Request": operation is not yet selected, so this routes normally and
    // establishes the oneOf branch 0 selection.
    handlers['ui:open:client-modal']!({
      id: operationId,
      requestBodyCompositionSelection: { 'requestBody.oneOf': 0 },
    })
    await nextTick()

    expect(routeSpy).toHaveBeenCalledTimes(1)
    expect(requestBodyCompositionSelection.value).toEqual({ 'requestBody.oneOf': 0 })

    // 2nd click on "Test Request" for the SAME operation (e.g. the doc page's schema viewer had
    // branch 1 selected this time). The operation is already selected in the sidebar, so
    // `handleSelectItem` takes its early-return path and never calls `route()` again.
    handlers['ui:open:client-modal']!({
      id: operationId,
      requestBodyCompositionSelection: { 'requestBody.oneOf': 1 },
    })
    await nextTick()

    // route() was NOT called again, path/method/example are unchanged, so RequestBody's
    // `operationChanged` check will be false.
    expect(routeSpy).toHaveBeenCalledTimes(1)

    // BUG: requestBodyCompositionSelection was still overwritten unconditionally, even though no
    // routing happened. RequestBody.vue's watcher now sees operationChanged=false and
    // selectionChanged=true, which it can only interpret as the user manually switching oneOf
    // branches within the same operation, so it regenerates the body from the schema and
    // discards whatever the user had typed.
    expect(requestBodyCompositionSelection.value).toEqual({ 'requestBody.oneOf': 1 })
  })
})
```

Dropped this next to `modal-events.ts` and ran it against current `main` with `pnpm vitest run src/v2/features/modal/modal-events.repro.test.ts`: it passes, confirming `routeSpy` is called once while `requestBodyCompositionSelection` still updates on the second event.

## Affected files

- `packages/api-reference/src/features/test-request-button/TestRequestButton.vue`
- `packages/api-client/src/v2/features/modal/modal-events.ts` (the `ui:open:client-modal` handler)
- `packages/api-client/src/v2/features/modal/hooks/use-modal-sidebar.ts` (`handleSelectItem`'s early return)
- `packages/api-client/src/v2/blocks/request-block/components/RequestBody.vue` (the `compositionSelectionKey` watcher)

## Suggested fix direction

`handleSelectItem` (or the `ui:open:client-modal` handler) needs to tell "already on this operation, nothing to route" apart from "opening fresh", and only push a new `requestBodyCompositionSelection` in the latter case. Right now both paths update the ref the same way, which is what lets an unrelated doc-page selection reach into an already-open, already-edited request body.

lazerg:
## Problem

Clicking "Test Request" for an operation that is already open in the client throws away an edited request body. The modal handler applies the reference page's `requestBodyCompositionSelection` after routing, but `handleSelectItem` returns early without routing when the target entry is already selected. The body stays as the user typed it while the selection jumps to whichever `oneOf`/`anyOf` branch the reference page happened to show, and `RequestBody.vue` can only read that as a manual branch switch, so it regenerates the body from the schema.

Webhooks hit the same thing from the other side: they re-route with an identical path, method and example, so nothing the request body watches changes there either.

## Solution

Remember which sidebar entry was selected before routing, and skip the new selection when an open modal is already showing that entry. Opening a different operation or example still re-establishes the selection, and so does opening the modal again after it was closed, so the reference page keeps driving the branch in every case except the one where it would destroy an edit.

## Checklist

- [x] I added a changeset. 
- [x] I added tests.

## Ticket

Closes #10023

### DIFF
--- a/.changeset/fix-10023-modal-composition-selection.md
+++ b/.changeset/fix-10023-modal-composition-selection.md
@@ -2,4 +2,4 @@
 '@scalar/api-client': patch
 ---
 
-Fix "Test Request" discarding an edited request body when the operation it opens is already the one shown in the client. Opening the modal on the entry that is already selected does not route anywhere, so the request body kept its edited value while the composition selection was replaced with whatever the reference page had selected. The request body read that as a manual `oneOf`/`anyOf` branch switch and regenerated itself from the schema. The selection is now only applied when the modal actually moves to a different entry.
+Fix "Test Request" discarding an edited request body when the operation it opens is already the one on screen. Reopening the entry the modal already shows does not route anywhere, so the request body kept its edited value while the composition selection was replaced with whatever the reference page had selected. The request body read that as a manual `oneOf`/`anyOf` branch switch and regenerated itself from the schema. An open modal now keeps the selection it is already showing, while opening a different entry still re-establishes it.
--- a/.changeset/fix-10023-modal-composition-selection.md
+++ b/.changeset/fix-10023-modal-composition-selection.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
+
+Fix "Test Request" discarding an edited request body when the operation it opens is already the one shown in the client. Opening the modal on the entry that is already selected does not route anywhere, so the request body kept its edited value while the composition selection was replaced with whatever the reference page had selected. The request body read that as a manual `oneOf`/`anyOf` branch switch and regenerated itself from the schema. The selection is now only applied when the modal actually moves to a different entry.
--- a/packages/api-client/src/v2/features/modal/modal-events.test.ts
+++ b/packages/api-client/src/v2/features/modal/modal-events.test.ts
@@ -8,6 +8,8 @@ import { computed, nextTick, ref } from 'vue'
 
 import 'fake-indexeddb/auto'
 
+import { createMockEventBus } from '@/v2/helpers/test-utils'
+
 import { useModalSidebar } from './hooks/use-modal-sidebar'
 import { initializeModalEvents } from './modal-events'
 
@@ -23,10 +25,18 @@ const getDocument = (overrides: Partial<OpenApiDocument> = {}): OpenApiDocument 
 })
 
 /**
- * Wires up the modal events against a real sidebar and returns the pieces the tests assert on.
+ * Helper to wait for async operations and Vue updates.
+ */
+const waitForUpdates = async () => {
+  await nextTick()
+  await flushPromises()
+}
+
+/**
+ * Wires the modal events up to a real sidebar and returns the pieces the tests assert on.
  *
- * The event bus is a stub that records the registered handlers, so a test can call them the same
- * way the workspace event bus would when `ui:open:client-modal` is emitted.
+ * The mock event bus records the handlers that are registered, so a test can call them the same way
+ * the workspace event bus would when the event is emitted.
  */
 const createTestSetup = async () => {
   const store = createWorkspaceStore()
@@ -38,6 +48,9 @@ const createTestSetup = async () => {
         '/pets': { post: { operationId: 'createPet' } },
         '/users': { post: { operationId: 'createUser' } },
       },
+      webhooks: {
+        newPet: { post: { operationId: 'newPetWebhook' } },
+      },
     }),
   })
 
@@ -46,11 +59,13 @@ const createTestSetup = async () => {
   const path = ref<string | undefined>(undefined)
   const method = ref<HttpMethod | undefined>(undefined)
   const exampleName = ref<string | undefined>(undefined)
+  const isWebhook = ref(false)
 
-  const route = vi.fn((payload: { path?: string; method?: HttpMethod; example?: string }) => {
+  const route = vi.fn((payload: { path?: string; method?: HttpMethod; example?: string; isWebhook?: boolean }) => {
     path.value = payload.path
     method.value = payload.method
     exampleName.value = payload.example
+    isWebhook.value = payload.isWebhook ?? false
   })
 
   const sidebarState = useModalSidebar({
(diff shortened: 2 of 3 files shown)


hanspagel (approved):
thanks!

PR #10025, opened 2026-08-26
refactor(components): make ScalarHeader chrome-only and composable

hwkr:
## Problem

`ScalarHeader` owned its own three column layout, which nobody actually wanted. Every consumer overrode it (`*:first:flex-none`, re-declared padding and borders) and the newer dashboard surfaces skipped the component entirely. It also couldn't shrink, so the app header overflowed on mobile.

## Solution

`ScalarHeader` now just provides the chrome (element, height, background, border, text colours) and you compose `ScalarHeaderColumn` children yourself.

* Columns hug their content — give one `flex-1` to take the free space, or both sides `flex-1` to centre the middle
* `ScalarHeaderColumn` carries `min-w-0`, so long titles ellipsize instead of pushing the header past the viewport
* `start` / `center` / `end` slots are gone, with a deprecation warning since Vue drops unrendered slot content silently
* New `is` prop so it can render as a `div` when nested inside an existing `header` landmark

Supersedes #9223 — thanks @redis for the original diagnosis and the `openScalarMenuPanel` test helper, both carried over here. `ScalarMenu.vue` itself ends up unchanged: verified in real WebKit at 375px that `min-w-0` is what fixes the overflow.

## Checklist

- [x] I added a changeset. 
- [x] I added tests.

## Ticket

Fixes [DOC-5424](https://linear.app/api-docs/issue/DOC-5424)
Fixes [DOC-6025](https://linear.app/api-docs/issue/DOC-6025)


hwkr:
Drafted pending testing in Org to make sure it works without hiccups

cameronrohani (approved):

DemonHa (approved):

PR #10026, opened 2026-08-27
fix(themes): stop autofilled input text rendering hollow

hwkr:
Autofilled inputs render hollow and near-invisible in dark mode — the value is there, just outlined at very low contrast. Showed up on the dashboard login filling with 1Password on iOS.

* `background-clip: text` clips WebKit's autofill background to the glyphs, but WebKit also forces `-webkit-text-fill-color` on autofilled fields — without re-asserting it the glyph interiors show that clipped background instead of the theme colour
* `currentColor` doesn't work here (WebKit resolves it against its own forced colour), so it's pinned to `--scalar-color-1`
* Dropping the rule instead just brings the yellow/blue autofill box back, so it's half-written rather than obsolete

No test — autofill state isn't something Playwright can drive.

Before / After

IMG_2716
IMG_2715

## Checklist
- [x] I added a changeset.
- [ ] I added tests.

### DIFF
--- a/.changeset/fix-autofill-text-fill-color.md
+++ b/.changeset/fix-autofill-text-fill-color.md
@@ -1,0 +1,5 @@
+---
+'@scalar/themes': patch
+---
+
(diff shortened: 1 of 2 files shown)


hanspagel (approved):
cool

PR #10034, opened 2026-08-27
feat(mock-server): add a quiet option to suppress startup logging

DemonHa:
## Problem

`createMockServer()` prints authentication instructions for the security schemes of the document every time it starts, straight to `console.log`. That is helpful when somebody runs the mock server in a terminal, but it is noise when the mock server is embedded — a test harness, a fixture server, another program's output.

There was no way to turn it off. The only escape was to replace the global `console` around the call, which is what the SDK generator's smoke-test harness does today: it wraps every `startMockServer()` call in a `withSilencedLog` helper that monkey-patches `console.log` and restores it afterwards. Patching the global console to quiet a library is a workaround, not an API.

## Solution

Add a `quiet` option to `createMockServer()`:

```ts
const app = await createMockServer({ document, quiet: true })
```

`quiet` is handed down into `logAuthenticationInstructions()` rather than used to skip the call, which matters: that function also reports security schemes the mock server cannot handle (`❌ Unsupported API Key Location`, `Unsupported security scheme type: mutualTLS`, …). Only the informational lines go through the new sink, so those diagnostics — and the request validator and `x-seed` errors elsewhere in startup — are printed either way. Somebody who asks for a quiet startup still wants to hear about a scheme that will not work.

Default is `false`, so nothing changes for existing callers.

Alternatives considered:

- **A `logger` sink** (like `createAsyncApiMockServer` has) would be more flexible, but a boolean covers the actual need and keeps the two factories' defaults from diverging further. It can be added later without breaking `quiet`.
- **Skipping the `logAuthenticationInstructions()` call at the call site** was the first shape and is one line shorter, but it silently swallows the unsupported-scheme diagnostics along with the instructions.

`MockServerOptions` is now exported too, so the options object can be named outside of the package.

## Checklist

- [x] I added a changeset.
- [x] I added tests.

Tests: a new `src/utils/log-authentication-instructions.test.ts` covering the instruction output (including the one call that passes two arguments) and asserting the exact warnings and errors that survive `quiet: true`, plus wiring tests in `create-mock-server.test.ts`. Docs: a "Quiet Startup" section in the mock server guide, a pointer to it from "Authentication", and the option in the agent skill.

`pnpm vitest --run` in `packages/mock-server`: 30 files, 376 passed, 1 skipped. `types:check`, Biome and Prettier clean.

---
_Generated by [Claude Code](https://claude.ai/code/session_01WEwc53fPE3PR5JcoqeULpA)_


amritk (approved):

DemonHa:
`deploy-previews / deploy-components` and `deploy-previews / deploy-examples` are failing here, and neither failure belongs to this PR. This diff only touches `packages/mock-server` and docs.

**What is failing**

Both jobs die in `pnpm install -g netlify` (`.github/workflows/deploy-previews.yml:75` and `:150`):

```
ERR_PNPM_NO_MATCHING_VERSION  No matching version found for @netlify/ai@^1.0.1
This error happened while installing the dependencies of netlify@27.4.0
 at @netlify/dev@5.0.4
The latest release of @netlify/ai is "1.0.0".
```

The install is unpinned, so it resolves to the current latest `netlify@27.4.0` → `@netlify/dev@5.0.4`, which declares `@netlify/ai@^1.0.1`. That version was never published — the highest on npm is `1.0.0`. The range is unsatisfiable for everyone, on every branch.

**Evidence it is not this PR's**

- The most recent `main` run ([33110550412](https://github.com/scalar/scalar/actions/runs/33110550412), commit `f856436` — this PR's own base) has exactly one failed job: `deploy-components`, same error.
- `deploy-galaxy` passes on this PR. It is the one preview job that does not install the Netlify CLI globally.
- Reproduced off CI against the live registry: `npm install --dry-run netlify@27.3.0` → `ETARGET No matching version found for @netlify/ai@^1.0.1`. Pinning to `27.3.0` does not help, because its `@netlify/dev@^5.0.1` still floats up to the broken `5.0.4`.

**No fix exists to port**

There is no open or merged PR in this repo addressing it, and the root cause is a third-party publish, so nothing can be ported into this branch. I have not widened this PR with a CI change.

**Proposed patch** (for a separate CI PR) — pin the CLI to the last version that resolves, in the three places that install it: `.github/workflows/deploy-previews.yml:75`, `:150`, and `.github/workflows/main.yml:396`:

```diff
-        run: pnpm install -g netlify
+        run: pnpm install -g netlify@27.1.0
```

Verified: `npm install --dry-run netlify@27.1.0` resolves cleanly (it pulls `@netlify/ai@0.4.4` via `@netlify/dev@4.18.13`). Pinning `@netlify/dev` to `5.0.1` instead would also resolve, but a global install cannot carry a pnpm override, so pinning the CLI is the practical fix. Either way it should be reverted once Netlify publishes `@netlify/ai@1.0.1`.

I re-ran the failed jobs once in case that publish lands in the meantime. Everything else on this PR is green.

---

DemonHa:
Correcting the scope of my previous comment: the same Netlify break takes down more than the two preview jobs, and it now blocks the merge gate. My "everything else is green" was wrong — more results landed right after I posted.

Six jobs are failing, all on the same unsatisfiable `@netlify/ai@^1.0.1`:

| Job | How it hits the CLI | Tests |
|---|---|---|
| `deploy-previews / deploy-components` | `pnpm install -g netlify` | n/a |
| `deploy-previews / deploy-examples` | `pnpm install -g netlify` | n/a |
| `test-component-snapshots` | `pnpx netlify deploy` (`pr.yml:433`) | **203 passed** |
| `test-e2e-api-reference` | `pnpx netlify deploy` (`pr.yml:338`) | **212 passed** |
| `test-e2e-scalar-app` | `pnpx netlify deploy` (`pr.yml:386`) | **6 passed, 2 skipped** |
| `test-cdn-snapshots` | `pnpx netlify deploy` | **26 passed** |

`required-jobs-pull-requests` then fails because it aggregates those.

The four Playwright jobs **passed every test** — see the result comments above. What fails is the "Deploy Playwright Report to Netlify" step that runs afterwards under `!cancelled()`, so a fully green test run still red-flags the job.

Verified locally that `pnpx` is affected exactly like the global install:

```
$ pnpm dlx netlify --version
ERR_PNPM_NO_MATCHING_VERSION  No matching version found for @netlify/ai@^1.0.1
This error happened while installing the dependencies of netlify@27.4.0
 at @netlify/dev@5.0.4
```

So the proposed patch needs to cover the `pnpx` call sites too, not just the three global installs — pinning `netlify@27.1.0` (verified to resolve) at `pr.yml:338`, `:386`, `:433`, the CDN snapshots equivalent, `deploy-previews.yml:75`, `:150`, and `main.yml:396`. Still a repo-wide CI fix that does not belong in this PR, and still nothing to port — `main` is red on the same cause since [33110550412](https://github.com/scalar/scalar/actions/runs/33110550412).

I have spent my one re-run on this run. Happy to open the CI pin as its own PR if that is wanted.

---

hanspagel (approved):

PR #10036, opened 2026-08-27
ci: pin the Netlify CLI to 27.1.2

DemonHa:
## Problem

Every job that reaches for the Netlify CLI is failing at the install step, on every branch.

`netlify@27.3.0` and newer pull in `@netlify/dev@5.0.4`, which depends on an `@netlify/ai` version that was never published to npm. There is nothing to resolve, so `pnpm install -g netlify` and `pnpx netlify` both fail outright — and because none of the invocations were pinned, the breakage arrived on its own the moment upstream published, without a commit here.

The fix currently rides along inside https://github.com/scalar/scalar/pull/10033, which is a `@scalar/highlight` feature PR. It is repo-wide and unrelated to that change, so it should not have to wait for a feature review to land.

## Solution

Pins the CLI through a `NETLIFY_CLI_VERSION` workflow-level env var, set to the last working release, in each of the three workflows that use it:

- `pr.yml` — the four `pnpx netlify deploy` calls that upload Playwright reports
- `main.yml` — the global install in the Storybook production deploy
- `deploy-previews.yml` — the global installs in the examples and Storybook preview deploys

One variable per workflow rather than seven inline versions, so unpinning later is a single-line change per file. The comment on each says why the pin exists and what has to happen upstream before it can go.

Extracted verbatim from https://github.com/scalar/scalar/pull/10033 — the workflow diff here is identical, so whichever lands first, the other is a no-op merge.

## Checklist

- [ ] I added a changeset. 
- [ ] I added tests. 

---
_Generated by [Claude Code](https://claude.ai/code/session_01GoGPa2xNPk9QLEunosA91Q)_

### DIFF
--- a/.github/workflows/deploy-previews.yml
+++ b/.github/workflows/deploy-previews.yml
@@ -44,6 +44,13 @@ on:
 permissions:
   contents: read
 
+env:
+  # The Netlify CLI is pinned because netlify@27.3.0 and newer pull in
+  # @netlify/dev@5.0.4, which depends on an @netlify/ai version that was never
+  # published — every install fails to resolve. Drop the pin once upstream
+  # publishes a working release.
+  NETLIFY_CLI_VERSION: '27.1.2'
+
 jobs:
   deploy-examples:
     runs-on: blacksmith-2vcpu-ubuntu-2204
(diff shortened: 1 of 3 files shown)


marclave (approved):

amritk (approved):
ty

PR #10039, opened 2026-08-27
fix(mock-server): keep circular markers out of validator compilation

DemonHa:
## Problem

Request validation silently fell open for any operation whose schema references itself.

`getResolvedRefDeep` cuts `$ref` cycles by substituting the *string* `'[circular]'`, so a schema as ordinary as a tree node

```jsonc
{ "type": "object", "required": ["name"], "properties": {
  "name": { "type": "string" },
  "child": { "$ref": "#/components/schemas/Node" }   // ← becomes "[circular]"
}}
```

reaches Ajv carrying that string. Ajv rejects the whole schema (`data/properties/child must be object,boolean`), `compileSchema` logs and returns `null`, and validation is skipped — for the request body, or for every parameter in that location. A `POST` with a body missing its required `name` answered `200` instead of `422`, and each such operation emitted a `console.error` at startup. Measured on a real OpenAPI document with recursive schemas: 11 skipped validators.

## Solution

Rewrite the markers before compiling (`utils/replace-circular-markers.ts`): the recursion point becomes an empty, always-valid schema, so everything above it stays enforceable. Only Ajv sees the rewritten copy — the parameter descriptors keep reading the resolved schema, since dropping a keyword must not change how a parameter value is deserialized.

Because that rewrite only ever *loosens* a subschema, the walk tracks whether each subtree was relaxed. Under `not`, `if`, `oneOf`, or `contains`, a looser subschema makes the schema **stricter**, so those keywords are dropped rather than rewritten, and the keywords that only qualify them go too (`then`/`else`, `minContains`/`maxContains`, `additionalProperties`, `items`, `unevaluatedProperties`/`unevaluatedItems` beside a branch the cycle ran through). The result: the rewrite can lose a violation, but it never invents one. Without that rule a nullable or recursive union — `oneOf: [{ $ref: Node }, { type: 'null' }]` — would start rejecting bodies the document allows.

Alternatives considered: hoisting each cycle target into `$defs` and emitting a real `$ref`, which Ajv resolves natively. That would validate recursion exactly instead of approximating it, but it is a much larger change; worth a follow-up.

**Verification.** Beyond the unit and end-to-end tests: a differential harness generated 3000 recursive schemas and compared this rewrite against Ajv compiling the same schemas with real recursive `$ref`s, over 45,000 instance checks — **0 compile failures and 0 cases where the mock rejects what the contract accepts** (93 compile failures before the `dependencies` keyword was covered).

Known follow-up, pre-existing and unrelated to this rewrite: a body schema using `$dynamicRef` is mis-validated because `validate-request.ts` does not resolve dynamic refs. Recursive documents used to fall open past that; they now hit the same pre-existing gap a non-recursive `$dynamicRef` document already hits.

## Checklist

- [x] I added a changeset.
- [x] I added tests.

---
_Generated by [Claude Code](https://claude.ai/code/session_01KkQwiA4uYrnhT4h9Ex5DRp)_

### DIFF
--- a/.changeset/mock-server-circular-validator.md
+++ b/.changeset/mock-server-circular-validator.md
@@ -1,0 +1,5 @@
+---
+'@scalar/mock-server': patch
+---
+
+Fix request validation falling open for recursive schemas. Resolving a schema that references itself leaves a `'[circular]'` marker where the cycle was cut, which Ajv refused to compile, so the mock server logged an error and skipped validating the request body — or every parameter in that location. The recursion point now compiles as an always-valid schema, and a constraint that would flip that into a stricter one (`not`, `if`, `oneOf`, `contains`, and the keywords that only qualify them) is dropped, so the rest of the schema is enforced again without rejecting requests the document allows.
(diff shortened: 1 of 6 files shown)


hanspagel (approved):
👍 👍 👍

PR #10041, opened 2026-08-27
perf(code-highlight): cache markdown processors and memoize renders

amritk:
## Problem

`htmlFromMarkdown` was rebuilding the entire `unified` processor pipeline on every call, and `rehypeHighlight` was creating a fresh lowlight registry of ~55 standard grammars at plugin initialization for each render. This caused unnecessary overhead when rendering multiple descriptions or markdown blocks — across ~35 call sites, inside a Vue `computed`, including `agent-chat`'s `Text.vue` which re-renders on every streamed token.

## Solution

Refactored the markdown rendering pipeline to:

1. **Cache processors by tag allowlist** — Processors are now built once per unique tag configuration and reused across calls. In practice there are two: with and without `img`/`picture`.
2. **Share a single lowlight instance** — Created once at module load and reused by all `rehypeHighlight` invocations.
3. **Memoize renders without transforms** — Renders that don't use custom transforms are cached by a key combining the tag allowlist and markdown content.
4. **Bound the cache** — 200 entries, evicting the oldest on insert, so it cannot grow without limit.

The implementation uses a module-level `activeTransform` variable to pass the per-call transform to the `transformNodes` plugin, since `unified` processors are frozen after configuration but transforms arrive per invocation. `processSync` runs to completion without yielding, so nothing can interleave; the save/restore around it only matters if a `transform` were itself to render markdown.

**Renders that pass a `transform` are deliberately not cached.** A `transform` is a closure that can read state the cache key cannot see — `ScalarMarkdown`'s heading transform reads `anchorPrefix` off props, so the same function object and the same markdown can legitimately produce different ids. Caching those would be a stale-anchor bug.

## Benchmarks

Measured over a 55-document corpus (92.3 KB of markdown: the fixtures from `markdown.test.ts`, `info.description` blocks from `@scalar/galaxy`, five `documentation/*.md` files and the markdown playground). One pass over the whole corpus, averaged over 5 runs, Node 22.

`before` reconstructs the pipeline exactly as it shipped — a fresh `unified()` per call, and `rehypeHighlight` given `languages` so it builds its own lowlight registry at plugin-init.

| | Whole corpus | Per document | vs before |
| --- | ---: | ---: | ---: |
| before — fresh processor + lowlight per call | 571–679 ms | ~10.4–12.3 ms | 1× |
| after — cache miss (full pipeline) | 282–324 ms | ~5.1–5.9 ms | **~2×** |
| after — cache hit | 0.3 ms | ~0.005 ms | **~2000×** |

Two paths are reported because the change creates two: a cache miss still runs the full pipeline and benefits only from the shared processor and lowlight registry; a hit skips it entirely. Real usage is a mixture. The cache-miss row is the honest floor — a page of all-distinct descriptions still halves.

The repeated-render row is what `agent-chat` streaming hits: re-rendering the same description 100× went from 679 ms to 19 ms before this change was even cached, and is now served from the cache.

Ranges are min–max across two independent runs of the same benchmark.

### Not included

Dropping `rehypeFormat` was considered and left out. It is worth a further ~23% (293 ms → 224 ms on the same corpus), but it changes output whitespace — `markdown.test.ts` asserts the exact indented string for the `allowTags` case, and the Playwright snapshots could shift. That is a behaviour change and does not belong in a commit whose claim is byte-identical output. Worth doing separately.

## Verification

Output is unchanged. A snapshot harness rendered all 55 corpus documents through three option shapes — default, `removeTags: ['img', 'picture']`, and a heading `transform` — for **165 snapshots total**, captured against the pipeline before the change and compared after: all 165 byte-identical.

To be clear about what that is: a throwaway harness written for this change, not a test suite in the repo. The repo's own coverage is what CI runs.

- `@scalar/code-highlight`: 54/54 tests pass, `types:check` clean
- `@scalar/components`: 12/12 `ScalarMarkdown` + `ScalarMarkdownSummary` tests pass
- biome and prettier clean
- `pnpm changeset status` bumps `@scalar/code-highlight` and `@scalar/api-reference` at patch

## Checklist

- [x] I added a changeset.
- [x] I updated the documentation. (JSDoc comments explaining the caching strategy and why transforms are excluded)
- [ ] I added tests. No new tests: this is a pure refactor with no behaviour change, and correctness is established by the byte-identical snapshot comparison above plus the existing suites. Happy to land the benchmark as a committed script if that is worth having as a regression guard.

https://claude.ai/code/session_01DjtTZRWnPmdQUmHRWbaUA2


amritk:
Looks like @cameronrohani is ahead of it https://github.com/scalar/scalar/pull/10074

PR #10042, opened 2026-08-27
fix(mock-server): route path keys that carry a query string

DemonHa:
## Problem

An OpenAPI path key is allowed to carry a query string, and SDK generators use that to describe two operations that share a path — a stable one and a variant:

```jsonc
"/v1/messages":           { "post": { "operationId": "messages_post" } },
"/v1/messages?beta=true": { "post": { "operationId": "beta_messages_post" } }
```

`honoRouteFromPath` passed the `?` straight through to Hono. Hono routes on the pathname alone, so the `?` reached `RegExpRouter` as a regular expression quantifier and it threw while lazily building its matcher on the very first request:

```
SyntaxError: Invalid regular expression: /^/a/([^/]+)/b(?:/([^/]+)$()|?q=1$())/: Nothing to repeat
```

Because that is a raw `SyntaxError` rather than Hono's own `UnsupportedPath`, `SmartRouter` never falls back to `TrieRouter`. The server starts cleanly and then **every** request to the document fails with an empty `500` — including requests to paths that have nothing to do with the offending key.

This is not hypothetical: Anthropic's public OpenAPI document carries `?beta=true` on 80 of its 89 path keys. Against that document, all 131 operations threw on every request before this change; the empty-bodied 500 is also what made the cause hard to find.

## Solution

The query string is split off the path key (`parsePathKey`) and matched against the request instead of being fed to the router.

Path keys stay registered in **document order**, exactly as before — nothing is reordered and nothing is registered twice, so precedence between overlapping routes (whether `/pets/mine` or `/pets/{id}` answers `GET /pets/mine`) is decided the same way as in a document with no query strings. Which key of a shared route answers is decided per request instead, in `selectPathKey`:

- the key whose query parameters the request carries answers, the most specific one winning, ties going to the key declared first;
- when no key matches, the least specific one answers anyway. Such a query string tells two operations apart rather than describing something a client has to send — the beta opt-in on the real API is a header — and documents written this way often carry no plain key at all, so the operation would otherwise never answer its documented URL.

A route with a single path key gets its handler back untouched, so the common case carries no overhead.

Two alternatives were measured and rejected. Reordering registrations so query-bearing keys come first is the obvious approach, but it moves whole routes past each other: a key declared between two keys of another route gets hopped over, and a key can become unreachable from every URL. Requiring the query string strictly is also wrong — it leaves 72 of Anthropic's 89 keys answering 404 at their documented URL.

Escaping other characters was considered and deliberately left out. Hono already escapes `.`, `+`, `(`, `[`, `]`, `^`, `$` and `|` in literal path text itself, and its `*` is a documented wildcard that existing documents — and AsyncAPI channel addresses such as `sensors/*/temperature`, which share this helper — can rely on. Once the `?` is split off there is nothing left that needs escaping.

Verified against the real Anthropic document: **131/131 operations answer 200 with the correct `operationId`, both with and without `?beta=true`, and `onRequest` fires exactly once per request** (previously 131/131 threw).

## Checklist

- [x] I added a changeset.
- [x] I added tests.

---
_Generated by [Claude Code](https://claude.ai/code/session_01PQG3ULL66nep7LXa4LyAKV)_


DemonHa:
The `build` check is failing (and with it the `required-jobs-pull-requests` rollup), but the failure is not from this PR.

**What is failing:** `@scalar/fastify-api-reference#build` → `tsgo --noEmit` in `integrations/fastify`, with type errors against Fastify's own types in `src/fastifyApiReference.ts` and `src/fastifyApiReference.test.ts`:

```
src/fastifyApiReference.ts(28,3): error TS2353: 'hide' does not exist in type 'FastifySchema'
src/fastifyApiReference.ts(138,21): error TS2345: ... 'FastifyRequest<..., NoInfer<FastifySchema>, ...>' is missing
  the following properties ...: context, routeConfig, routerPath, routerMethod, connection
src/fastifyApiReference.ts(184,7): error TS2578: Unused '@ts-expect-error' directive
src/fastifyApiReference.test.ts(99,30): error TS2769: No overload matches this call
```

Every other check on this PR passes — `lint`, `format`, `check-changes`, `dependency-review`, CodeQL.

**Why it is not this PR's:** this branch changes only `packages/mock-server/**` and one changeset — `git diff --name-only origin/main...HEAD` returns nothing else, and the lockfile is untouched. `integrations/fastify` does not depend on `@scalar/mock-server`: it is absent from its `package.json`, from its sources, and from its Turbo build graph (`turbo build --filter '@scalar/fastify-api-reference...' --dry=json` lists no mock-server task). The Fastify integration's sources, its dependencies and `pnpm-lock.yaml` are byte-identical between this branch and `main`, so the type-check outcome cannot differ because of this change.

The errors themselves are the signature of Fastify v5 type declarations being resolved where v4 is expected — v5 dropped `request.context` / `routerPath` / `routerMethod` / `connection`, added `NoInfer<FastifySchema>` to `RouteShorthandHook`, and no longer accepts `hide` on `FastifySchema`. `integrations/fastify` declares `"fastify": "^4.0.0"` and resolves `fastify@4.28.0`, while `fastify@5.12.0` is present in the workspace transitively through the v5-era `@fastify/*` plugins. On a warm Turbo cache this package's build is restored rather than executed, which is why `main` stays green; this PR's run reported `Cached: 0 cached, 53 total` and `cache miss, executing` for the Fastify build.

**What I did:** re-ran the failed job once to confirm. I found no existing fix to port — no open PR addresses these Fastify type errors. I am not fixing them here, because doing so would mean changing `integrations/fastify` (or its dependency resolution), which is unrelated to this PR and would widen a mock-server change into the Fastify integration.

Happy to open a separate PR for the Fastify type resolution if that is useful — just say the word.


---

DemonHa:
Correction to my previous comment: the re-run passed and **CI is now fully green** — `build`, `types`, `knip`, `test-js`, `test-integrations`, the e2e and snapshot suites, `lint`, `format` and CodeQL all succeeded on `f43c6eb`.

So the Fastify type errors did not reproduce. They were a one-off in that run's install, not a standing breakage — please disregard the offer of a separate PR for them, there is nothing to fix. My reasoning about *why* it could not have been this PR's failure still holds (this branch touches only `packages/mock-server/**` plus a changeset, and `integrations/fastify` has no dependency or build-graph edge to it), but the "warm Turbo cache masks it on `main`" hypothesis was wrong, and I would rather retract it than leave someone chasing it.

The PR is green and waiting on review.


---

hanspagel:
Thanks, it seems the same bug got fixed in the meantime by #10044, which is already on main. :))

PR #10043, opened 2026-08-27
feat(api-reference): resolve heading levels from the page outline

hwkr:
Docs sites render one operation per page, so the title was coming out as an `h3` with no `h1` on the page at all. Blocks now resolve their own level instead of hardcoding one — rendered on its own a block is the `h1`, composed into a full reference nothing moves.

* New `useDocumentOutline` hook — a block declares its role, `Content` anchors the outline, the levels fall out
* Classic layout operation titles were a raw `h3` that skipped the heading components entirely
* Classic layout Models label passed `:level` to a component that doesn't take one, so it rendered no heading at all

No visual change — the themes reset already neutralizes UA heading sizing, so only the tag names move.

## Checklist
- [x] I added a changeset.
- [x] I added tests.

### DIFF
--- a/.changeset/document-outline.md
+++ b/.changeset/document-outline.md
@@ -1,0 +1,9 @@
+---
+'@scalar/api-reference': minor
+---
+
+Resolve heading levels from a block's place in the page outline instead of hardcoding them, so a block rendered on its own starts at `h1`.
+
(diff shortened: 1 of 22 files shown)


hanspagel (approved):
V NICE

PR #10051, opened 2026-08-28
feat(mock-server): replace the quiet option with a logger option

hanspagel:
`createMockServer` always printed startup logs. #10034 added `quiet: true` to hide them; this swaps that flag for a `logger` option before it ships:

- `logger: false` — no logs
- `logger: true` — logs to the console
- `logger: (line) => …` — send the logs somewhere else

`createAsyncApiMockServer` takes the same option now. Warnings and errors always print, whatever you pass.


hanspagel:
bypassed review, those were just a few uncommited changes to an already approved PR
I wanted to patch this before release

PR #10053, opened 2026-08-28
test(highlight): re-enable linear-cost performance test

hanspagel:
PR #9941 (a CI/release change) accidentally added `describe.skip` to the `cost stays linear in line length` test in `@scalar/highlight`. That skip had nothing to do with that PR, so this turns the test back on.

The test guards against a slow highlighter (a rule scanning a line over and over) on untrusted code blocks, so it is worth keeping on. I ran it three times and it passed every time.


amritk:
This test is flakey, I have it fixed in a draft pr

PR #10054, opened 2026-08-28
ci(release): fail fast on unpublishable new packages

hanspagel:
The release [failed halfway](https://github.com/scalar/scalar/actions/runs/33164977470): three new packages from #9967 were not on npm yet, and with trusted publishing (OIDC) you cannot set up a publisher until a package exists. So `pnpm -r publish` hit the first new package, errored, and stopped — most packages (including api-reference and api-client) never shipped.

This adds a check that runs before publishing, on release commits only. If a public package is not on npm yet, the release stops early and tells you how to publish it the first time — instead of failing halfway through.

Workflow-only change, so no changeset.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -182,6 +182,64 @@ jobs:
           fi
           echo "npm $(npm --version)"
 
+      # Refuse to start a publish that would abort halfway through. npm trusted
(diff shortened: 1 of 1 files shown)


hanspagel:
insane script

PR #10055, opened 2026-08-28
chore(tooling): add new-package npm bootstrap script

hanspagel:
When you add a brand-new package, the release cannot publish it the first time. We use trusted publishing (OIDC), and npm will not let you set up a publisher until the package already exists — so the first publish has to be done by hand.

This adds a `bootstrap-package` command that does the first-time setup for you: it publishes the package once, then sets up its trusted publisher so future releases just work. It skips anything already on npm, so it is safe to re-run, and `--dry-run` previews without touching the registry.

```bash
pnpm script bootstrap-package @scalar/your-new-package
```

AGENTS.md and the tooling README explain when to use it. No changeset (tooling + docs only).

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -347,6 +347,18 @@ If a package version should bump, add a changeset:
 pnpm changeset
 ```
 
+### Publishing a brand-new public package
+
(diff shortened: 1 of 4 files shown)


hanspagel:
maybe let's just add a script for the first publish of a script, not to scaffold it

PR #10056, opened 2026-08-28
feat(highlight): add a dedicated curl grammar

amritk:
## Problem

`@scalar/highlight` maps `curl` to the bash grammar. A curl invocation *is* a shell command line, so that reads as the honest answer — but it loses the two tokens anyone actually looks for in one.

Rendering the same curl block through both engines:

```
reference (highlight.js + Scalar's curl grammar)
<span class="hljs-keyword">curl</span> https://galaxy.scalar.com/planets \
  <span class="hljs-literal">--request <span class="hljs-symbol">POST</span></span> \
  <span class="hljs-literal">--header</span> <span class="hljs-string">'Content-Type: application/json'</span>

@scalar/highlight, before this change
curl https://galaxy.scalar.com/planets \
  <span class="hljs-variable">--request</span> POST \
  <span class="hljs-variable">--header</span> <span class="hljs-string">'Content-Type: application/json'</span>
```

`curl` and `POST` came out unscoped: the command name lost its keyword colour (purple) and the method lost its symbol colour (blue). The flags landed on `hljs-variable` rather than `hljs-literal`.

That last one matters more than it looks, because `code.css` has **curl-specific** rules written against the reference grammar:

```css
.hljs.language-curl .hljs-string  { color: var(--scalar-color-blue); }
.hljs.language-curl .hljs-literal { color: var(--scalar-color-1); }
```

So the mapping was not just coarser — it routed flags around a stylesheet rule written for them.

## Solution

Adds `src/langs/curl.ts`, mirroring the grammar `@scalar/code-highlight` ships:

- the command is a `keyword`
- `--request` / `-X` and other flags are `constant.builtin` → `hljs-literal`
- the method they carry is a `symbol` → `hljs-symbol`
- `$(…)` inside a double-quoted value reads as a `variable`
- single-quoted values stay literal, the way the shell treats them

Output now matches the reference class for class. The one structural difference: the reference nests the symbol *inside* the literal span, this emits them as siblings. The space between them is unstyled either way, so it renders identically.

Two details worth flagging for review:

- Flag rules are anchored to a word boundary, so the hyphens in a URL or in `Content-Type` are not mistaken for flags.
- The escaped-quote rule is ordered ahead of the opening-quote rule, so `\"` stays content rather than opening a string.

### The `symbol` scope

The method word needs a scope that maps to `hljs-symbol`, and the vocabulary had none. This adds one. It is a scope highlight.js already emits and `code.css` already styles, and it is the natural home for Ruby symbols and Erlang atoms if those grammars ever want it — so it earns its place beyond this one use.

`style.css` gives it a colour in our own theme, which the existing "styles every scope the grammars can emit" test enforces.

### Tests

- `src/langs/curl.test.ts` — 11 cases: round-trip, scope vocabulary, the command, the method apart from its flag, method case-insensitivity, long and short flags, hyphens inside a header name, quoted values, command substitution, escaped quotes, and that a bare `post` in a URL path is *not* a method.
- A curl sample joins the shared corpus in `test/samples.ts`, and `test/languages.ts` gains the reference grammar to differ against. The existing comparison suites — exact text round-trip, class-vocabulary, and the colour-agreement floor — now cover curl the way they cover every other language. **That is the check that would have caught this in the first place.**

`pnpm vitest packages/highlight --run` → 1269 passed, 34 skipped. `types:check` clean. Knip is unchanged by this diff (it exits 1 on a clean main too, on unrelated packages).

## Checklist

- [x] I added a changeset.
- [x] I added tests.

---
_Generated by [Claude Code](https://claude.ai/code/session_01NxHFt5aoffmQhXDDhuDGWd)_


amritk:
merging into other PR for screenshots

PR #10067, opened 2026-08-31
chore: update pnpm to v12

hanspagel:
Updates pnpm `10.16.1` → `12.1.0` and migrates the config pnpm 12 (the Rust rewrite) needs.

**What changed**

- `packageManager` + `engines` → pnpm 12.1.0
- Moved `pnpm.overrides` out of `package.json` into `pnpm-workspace.yaml` — pnpm 12 no longer reads the `package.json` `"pnpm"` field.
- Replaced `onlyBuiltDependencies` / `ignoredBuiltDependencies` with the new `allowBuilds` map (same packages, same intent). pnpm 12 now makes an un-approved build script a hard error instead of a warning.
- Regenerated the lockfile. The change is small (~100 lines): pnpm 12 just records its own binary — the dependency graph is unchanged.
- CI: the Netlify CLI is installed on the fly (`pnpm dlx` / `pnpm install -g`), and pnpm 12 failed those because netlify pulls in build scripts (esbuild, sharp, unix-dgram). Added `--allow-build` for exactly those so the deploy/snapshot jobs pass again.

### DIFF
--- a/.github/workflows/deploy-previews.yml
+++ b/.github/workflows/deploy-previews.yml
@@ -50,6 +50,10 @@ env:
   # published — every install fails to resolve. Drop the pin once upstream
   # publishes a working release.
   NETLIFY_CLI_VERSION: '27.1.2'
+  # netlify@27.1.2 pulls in esbuild, sharp and unix-dgram, which run build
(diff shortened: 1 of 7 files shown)


marclave (approved):

amritk (approved):
yesss

PR #10068, opened 2026-08-31
fix(workspace-store): show a uuid example for version-specific uuid formats

hanspagel:
FastAPI/Pydantic emit version-specific UUID formats like `uuid4`. We only knew `uuid`, so those fields got an empty example instead of a UUID.

Now `uuid1`, `uuid3`, `uuid4`, `uuid5` (and up to `uuid8`) all use the normal UUID example. Nothing else changes, the format label still shows the exact value from the document.

This comes out of discussion #10064. A full `x-formatHints` extension felt like too much for this; a small normalization fixes the real case with no config.

## Ticket

Ref #10064

### DIFF
--- a/.changeset/fix-uuid-version-format-example.md
+++ b/.changeset/fix-uuid-version-format-example.md
@@ -1,0 +1,5 @@
+---
+'@scalar/workspace-store': patch
+---
+
+Generate a proper UUID example for version-specific uuid formats (uuid1, uuid3, uuid4, uuid5), so documents from FastAPI/Pydantic get a UUID example instead of an empty one.
--- a/packages/workspace-store/src/request-example/builder/helpers/get-example-from-schema.test.ts
+++ b/packages/workspace-store/src/request-example/builder/helpers/get-example-from-schema.test.ts
@@ -267,6 +267,22 @@ describe('getExampleFromSchema', () => {
     expect(result).toBe('hello@example.com')
   })
 
+  it('uses a uuid example for version-specific uuid formats', () => {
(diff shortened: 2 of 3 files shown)


ShinonomeNoAlice:
Thanks for putting this together so quickly. This definitely solves the immediate issue for UUIDs, which is great.

My main hesitation with the current approach is around how it scales long-term. While the `uuid1`–`uuid8` behaviour has been stable in Pydantic (hasn't changed much [since 2019](https://github.com/pydantic/pydantic/pull/1103/changes#diff-f7a381c0918da7ec905bfbe6aff2ccc72c2d0fd6da0dd3011ef1da892e42d57aR458)), this approach still relies on framework-specific behavior. Since it's not a formally documented invariant, there's no strict guarantee it won't change in the future.

With `uuid8` in particular, it can carry application-defined semantics. For example, a project could have `uuid8-with-metadata-bitfield` and `uuid8-femtosecond-timestamp`. Treating every `uuid8-*` format as interchangeable might get messy down the line, since the version alone doesn't tell the consumer what the custom bits mean.

My main worry is that hardcoding these could accidentally lock Scalar into maintaining a growing list of framework-specific format names (which was the edge case I was hoping we could avoid in #10064). That said, I haven't seen another framework emit this exact `uuid1`–`uuid8` convention yet, so maybe this specific case is stable and common enough to warrant special handling.

In any case, I appreciate the fix, and I'm very much open to further discussion on the general case.

marclave (approved):

Issue #7219, opened 2025-10-31
show schemas in api client


hanspagel:
### Discussed in https://github.com/scalar/scalar/discussions/6982

<div type='discussions-op-text'>

<sup>Originally posted by **OrkanCom** October  1, 2025</sup>
## Problem
When adding custom examples to OpenAPI specification, Scalar's request dropdown only shows the custom examples. There's no way to access the auto-generated schema fields that Scalar originally displayed.

## Current Behavior
- Without custom examples: Scalar auto-generates request body from schema.
- With custom examples: Only custom examples appear in drop-down, no way to see all available schema fields.

## Expected Behavior
The dropdown should include both:
1. **"Schema"** or **"Default"** option - showing all fields auto-generated from OpenAPI schema.
2. Custom examples from the specification.

## Use Case
Developers need to:
- See all available fields in the API (schema-generated)
- Quickly test with pre-filled examples (custom examples)
- Switch between these two views easily

## Screenshot
<img width="1381" height="792" alt="image" src="https://github.com/user-attachments/assets/b5b5094e-c116-4651-8d71-cea98c9f2ade" />

## Proposed Solution
Add a reserved example name (e.g., `"$schema"`) or automatically prepend a schema-based example to the dropdown when custom examples exist.</div>

amritk:
A bonus feature would be use the example generated data as well

atharvamhaske:
hey can i know more on this, also are you accepting open source PRs

amritk:
hey we are always open to open source PRs! Basically you would just need to run this function on the schema to generate an example
`packages/api-client/src/v2/blocks/operation-code-sample/helpers/get-example-from-schema.ts`

atharvamhaske:
Sure on it

amritk:
hey client v2 is built on examples!

<img width="1252" height="930" alt="Image" src="https://github.com/user-attachments/assets/8ddd421e-8e66-4770-bcc1-9d6109d37877" />

amritk:
Re-opening because we are still missing the autogen from schema button! But that would be super easy to add now:
- add a button to "generate example from schema" @cameronrohani should I just toss it where hans had it in his drawing?
- hitting the button will create a new example using the generated values

DemonHa:
@amritk the new examples use the generated schemas by default

AniketR10:
@amritk  is this still open? i would like to work on it

amritk:
Yes for sure!

PR #10072, opened 2026-09-02
feat(api-client): add generate example from schema button

AniketR10:
## Problem

when an operation has custom examples in the openapi document, the api client only shows those examples. there is no way to see the auto-generated body built from the schema.

this is confusing because the client does show it when there are no custom examples. so adding examples to your document silently takes the feature away.

hand-written examples are usually short. the schema often has many more fields. before this change you had to go read the docs elsewhere and type the json yourself.

## Solution

added a button in the request body header. clicking it creates a new example filled with values generated from the schema, and opens it.

- the button only shows for json/yaml bodies that have a schema
- it reuses `getSchemaExampleFromBody`, the same helper the client already uses, so the output matches what you would get on an operation without custom examples
- the generated body is saved into the example instead of only rendered, so you can edit it and it survives an export
- names are deduped, so clicking twice gives `generated from schema (2)`

the generator itself already existed. it just had no example to attach to when every example key already had a value.

## Visual

https://github.com/user-attachments/assets/aa57f764-4ee9-4c66-b8df-05ded993f9d2

## Checklist

- [x] I added a changeset. 
- [x] I added tests.

## Ticket

See #7219


AniketR10:
@hanspagel @amritk @marclave @DemonHa  please review...

hanspagel:
I mean, real examples will always be better than our generated ones, right?

AniketR10:
> I mean, real examples will always be better than our generated ones, right?

@hanspagel yes but doesn't change that they stay first in the list and stay the default.

the thing i was trying to fix is that they answer different questions a real example shows what a typical req looks like. the generated one shows what fields the endpoint accepts. hand-written examples are minimal on purpose just like Marc's is 3 fields, the schema has more so there's no way to see the full set once custom eg exist

hanspagel:
I think we'd rather want a way to see the schema in the api client, that might be more helpful than the "best guess" generated payload:

#7219 

EDIT: Ah, I just see the issue describes the auto-generated payload, but I’d personally prefer to see the raw schema in the api client I think.

AniketR10:
> I think we'd rather want a way to see the schema in the api client, that might be more helpful than the "best guess" generated payload:
> 
> #7219
> 
> EDIT: Ah, I just see the issue describes the auto-generated payload, but I’d personally prefer to see the raw schema in the api client I think.

makes sense. as @amritk  re-opened this specifically for the generate btn and i picked it up on that basis, so i want to make sure i'm building the right thing also i don't think these conflict though. "what fields does this take" and "give me a req i can send" feel like different jobs and the second still needs a payload in the box

marclave (approved):

hanspagel (approved):

PR #10074, opened 2026-09-02
feat(api-reference): rail and gutter tree layout for the schema renderer

cameronrohani:
## Problem

The schema renderer drew every level of a document as a nested card — a box inside a box inside a box. Past two levels the chrome cost more horizontal space than the content it framed, and once a reader was a few levels deep nothing told them *where* they were or how to climb back out. Several surfaces never got the treatment at all: response headers, callbacks and AsyncAPI messages each had their own ad-hoc disclosure, so a property inside a callback body could not be linked to, `expandAllSchemaProperties` did not reach it, and find-in-page could not see it.

Two long-standing bugs turned up while working in here and are fixed in **both** layouts:

- The Examples / Default / Pattern popups were hover-only. They could not be opened by keyboard or by touch.
- The response-headers card carried a CSS syntax error that had silently disabled its own styling.

## Solution

A tree, drawn the way a file tree is drawn: one rail per level, and a disclosure toggle hanging in the gutter beside the text rather than wrapping it in a box.

The important part is that it is **one grammar**, composed by every surface instead of reimplemented per surface. Rows, response headers, callbacks, parameter rows, models and AsyncAPI messages all fold the same way, share the same expansion store, and produce the same anchors — so a property inside a callback body is addressable for the first time.

What a reader gets:

- **Depth you can see.** A rail per level that fades with depth; clicking a rail closes the row it belongs to.
- **Collapsed rows preview their children** — `{ compound, percentage }` — so folding hides elements, never information.
- **Types read as a token run.** `array of Satellite`, with the model name a link. A `$ref` *is* the type rather than an appended `· Account`.
- **Cycles say so.** A recursive property reads `array of Satellite · recursive` and stops, instead of being an ordinary-looking toggle that expands forever.
- **Map keys name themselves.** `additionalProperty · string` in the signature line, so the placeholder name can look like every other name instead of wearing a dashed box.
- **Deep links work.** Expansion lives in a store keyed by anchor path, so a link opens the row it points at, expand-all reaches every surface, and closed panels stay reachable by find-in-page through `hidden="until-found"`.

### The legacy layout is untouched

Every tree style is a Tailwind utility applied only under `schemaLayout: 'tree'`. The semantic class names (`property--tree`, `property-children`, `schema-glyph`, …) stay on the elements as styling hooks for anyone theming Scalar by hand, and legacy anchors keep the exact ids readers have bookmarked — including the `…responses.headers.headers.X` duplication, which the tree drops but legacy preserves on purpose.

`schemaLayout: 'tree'` is the new default; `schemaLayout: 'legacy'` restores the previous rendering byte for byte.

### Accessibility

The gutter toggle is the accessible control (`aria-expanded`, `aria-controls`, `aria-labelledby` pointing at the property name, child count on `aria-describedby`); the rail and the row heading are pointer-only conveniences on top of it. Tab order matches visual order. `schemaKeyboardNav` (default off) adds APG tree arrow-key navigation over the toggles.

## Reviewing this

It is a large diff, so here is where it comes from. Of **5,395 added lines**: 1,102 are unit tests, 604 are dev-only harness (bench, Storybook stories, Playwright e2e), 210 are config/locales/docs/changesets, and ~790 of the remaining source lines are comments. The genuinely new logic is about **2,300 lines**, concentrated in the schema renderer rewrite (~1,070), the other surfaces adopting it (~680), seven small new tree primitives (~580) and the expansion store (370).

It is also, honestly, five increments in one branch — the five changesets name them: popup accessibility, the expansion store, the tree layout, the remaining surfaces, and the default flip. The first two are independent of the redesign and could be split out if that would make review easier; say the word and I will stack them.

Worth a reviewer's attention:

- `SchemaProperty.vue` is now 1,139 lines and contains six near-identical `<Schema>` invocations that a single child-renderer component could collapse. Deliberately left for a follow-up rather than growing this diff further.
- The indentation model is documented at the top of `SchemaProperty.vue` and is easy to break: a row never indents itself, `SchemaRailPanel` pads its children by exactly one gutter, and floating controls hang at fixed negative offsets. Adding a per-row gutter column and cancelling it elsewhere re-introduces the first-level indent bug.
- Tailwind v4 compiles utilities before SFC scoped styles, so contested utilities carry a `!` suffix. There is a note explaining this at the top of `styles/tailwind.config.css`.

## Verification

- `@scalar/api-reference` unit tests: **1,860 passing** (143 files)
- `vue-tsc --noEmit` clean on `@scalar/api-reference`, `@scalar/types`, `@scalar/schemas`
- `pnpm knip` clean, `pnpm changeset status` clean
- Every layout decision measured in the browser against Scalar Galaxy rather than eyeballed — row rhythm, glyph centring, hover states and the open/close shift are all pixel-verified

## Checklist

- [x] I added a changeset.
- [x] I added tests.


hanspagel (approved):
sick work mate

PR #10088, opened 2026-09-04
fix(docs): correct how API reference sources stay up to date

amritk:
## Problem

The URL source for an API reference route was documented like this:

> Fetch an API document from a remote URL. **The document is fetched on each page load, keeping your documentation in sync with your live API**

That is not what happens. `loadLiveLinkOpenapiContent` runs in the SSG **preprocess** stage, so the document is fetched once while the site is being published and baked into that publish. A `url` reference is a snapshot, exactly like a `filepath` one.

Read literally, the old sentence tells you the opposite — that a `url` reference keeps itself current. Someone acting on it will pick `url` precisely when they want a reference that stays fresh, which is the one thing it does not do.

This is not hypothetical. A customer configured:

```json
"/reference": {
  "type": "openapi",
  "title": "API references",
  "url": "https://registry.scalar.com/@their-namespace/apis/their-api@latest"
}
```

A live link, pointing at our own Registry, at `@latest`. The intent could not be clearer: always serve the newest version. But because it is a `url` and not `namespace` + `slug`, `getConfigEmbeddedRegistryApis` collects nothing, the publish record is stamped with `registryApis: []`, and the `api-version-upserted` republish fan-out matches no record. New endpoints pushed to the Registry stayed invisible on their live site until whenever they next published — 22 hours in the case that surfaced this, and their publish gaps in that same week ranged from 1 to 40 hours.

It read as a caching bug for a day. It was a docs bug.

## Solution

In `documentation/guides/docs/configuration/navigation.md`:

- Replace the false claim in the URL section with what actually happens: the document is fetched while the site is being published, not per page load, and later changes at that URL do not reach the site on their own.
- Steer the specific trap: if the URL points at a Registry document, use `namespace` and `slug` instead. That is the shape the republish fan-out can see.
- State the timing up front, so it is visible before the reader picks a source.
- Add a **Keeping a reference up to date** table comparing all three sources, since "will this stay current?" is the actual question people bring to this page.
- Sharpen the Registry section: it is the only source that reaches a published site without publishing again. Document that `version` defaults to `latest`.
- Note in the properties table that `url` is fetched at publish time and that `version` defaults to `latest`.

Also added the same distinction to `.agents/skills/scalar-docs/SKILL.md`, which lists the three sources with no update semantics at all — an agent writing a config from it today would produce the same broken shape.

Timings are from production: registry-push-to-republish measured 0–39 seconds across 25 samples in three days, hence "usually within a minute".

## Checklist

- [ ] I added a changeset. 
- [ ] I added tests. 

https://claude.ai/code/session_01VTXbBBtofHsBYKmadCroJM

---
_Generated by [Claude Code](https://claude.ai/code/session_01VTXbBBtofHsBYKmadCroJM)_

### DIFF
--- a/.agents/skills/scalar-docs/SKILL.md
+++ b/.agents/skills/scalar-docs/SKILL.md
@@ -175,6 +175,8 @@ API reference from file, Registry, or URL:
 }
 ```
 
+All three sources are read at publish time. Only the Registry source republishes the site on its own when the document changes; `filepath` and `url` stay as they were until the next publish. Never point `url` at a Registry document — use `namespace` and `slug` so the reference keeps itself current.
+
 Display modes: `folder` (default), `flat`, `nested`.
 
 **Single page mode:** Set `singlePage: true` to render all operations on a single page instead of creating separate pages for each operation:
--- a/documentation/guides/docs/configuration/navigation.md
+++ b/documentation/guides/docs/configuration/navigation.md
@@ -390,6 +390,8 @@ Scalar supports three ways to generate API references:
 
 An API reference entry accepts `type: "openapi"` or `type: "asyncapi"`. Use whichever matches your document — the format is detected automatically.
 
+Whichever source you pick, the API document is read when your site is published, and the reference that gets rendered from it is part of that publish. Only the Registry updates your published site on its own — see [Keeping a reference up to date](#keeping-a-reference-up-to-date).
+
 ### 1. Files
 
 Reference an API document stored in your repository by specifying a relative path from your configuration root:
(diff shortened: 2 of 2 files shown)


amritk:
Closing in favour of fixing the behaviour instead of documenting the limitation.

This PR documented that a `url` reference is a snapshot and steered anyone pointing at a Registry document to use `namespace` and `slug`. `scalar/scalar-org#6267` makes that steer unnecessary: a route whose `url` addresses a Registry document is now recorded on the publish record, so a Registry update republishes the site the same way a `namespace` + `slug` route already did.

One thing this drops on the floor, noted so it is not lost: the sentence

> The document is fetched on each page load, keeping your documentation in sync with your live API

is still wrong for a `url` pointing anywhere other than the Registry. Those are fetched at publish time and frozen into the publish, and after #6267 they are the only URL source that does not update on its own — so the sentence stays misleading exactly where it is now hardest to notice. Worth a smaller follow-up.

---

PR #10090, opened 2026-09-04
fix(workspace-store): stop truncating large numeric strings in array parameters

### LINKED ISSUE #10087: Large numeric strings in query parameters (like List<string>) suffer from JavaScript precision loss (Number.MAX_SAFE_INTEGER)
### What happens?

When entering a long numeric string (e.g., a 20-digit bank account or reference number) into a query parameter input field that is explicitly defined as an array of strings, Scalar UI internally treats the value as a JavaScript Number.Because the value exceeds Number.MAX_SAFE_INTEGER, it suffers from IEEE 754 precision loss, truncating the last digits before compiling the final query string and sending the request.

### Minimal Reproducible Example (.NET 10 / C#)
```csharp
public record FilterTest
{
    // Explicitly defined as a list of strings
    public List<string> Values { get; set; }
}

[HttpGet("load-test")]
[ProducesResponseType(StatusCodes.Status200OK)]
public Task Test1([FromQuery] FilterTest request) => Task.CompletedTask;
```
### Steps to Reproduce
Open Scalar UI for the endpoint above.
In the Values query parameter field, enter the following 20-digit numeric value: 40702810506710000185 
Click "Send" or look at the generated URL preview

### Actual Behavior
The value is treated as a JS number and gets rounded/truncated:
```text
/load-test?Values=40702810506710000000
```
Consequently, the backend receives 40702810506710000000.

### Environment
Scalar Package: Scalar.AspNetCore.Microsoft (or core @scalar/api-reference)
Version: 2.14.14+
Framework: .NET 10 (C#)

### Additional Context / Workarounds
The issue occurs because the Scalar client uses aggressive type-guessing heuristics on the input text instead of strictly adhering to the type: string definition from the OpenAPI schema.
Current UI workarounds that bypass the bug: Entering values in quotes: "40702810506710000185" (sends correctly).
Appending a character: 40702810506710000185m (forces string fallback, sends correctly).

The generated URL should preserve the exact string value:
```text
/load-test?Values=40702810506710000185
```
And the backend should receive 40702810506710000185.

### OpenAPI Document

_No response_

lazerg:
## Problem

Query and header parameters typed as an array of strings go through `JSON.parse` before being sent, since that is how a comma-separated or JSON-array input gets turned into a real array. A bare numeric string is valid JSON on its own though, so typing a 20-digit value like `40702810506710000185` into a `type: string` array field parses it into a JS number and loses precision past `Number.MAX_SAFE_INTEGER`, sending `...000000` instead of the original digits.

## Solution

`deSerializeSchemaValue` now only trusts the `JSON.parse` result for an array-typed field when it actually parsed to an array. A quoted string (`"40702810506710000185"`) still parses to a single-item array, matching the escape hatch reported in the issue. Any other non-array result falls back to the same comma-split handling already used for plain non-JSON input like `foo,bar`.

## Checklist

- [x] I added a changeset.
- [x] I added tests.

Fixes #10087

### DIFF
--- a/.changeset/fix-10087-array-param-precision-loss.md
+++ b/.changeset/fix-10087-array-param-precision-loss.md
@@ -1,0 +1,6 @@
+---
+'@scalar/workspace-store': patch
+'@scalar/api-client': patch
+---
+
+Stop truncating large numeric strings entered into `type: string` array/object query and header parameters. A value like a 20-digit reference number was accepted as valid JSON, parsed into a JS number, and lost precision beyond `Number.MAX_SAFE_INTEGER` before being sent. Such values now fall back to the comma-split string handling instead of being parsed as a number.
--- a/.changeset/fix-10087-array-param-precision-loss.md
+++ b/.changeset/fix-10087-array-param-precision-loss.md
@@ -3,4 +3,4 @@
 '@scalar/api-client': patch
 ---
 
-Stop truncating large numeric strings entered into `type: string` array/object query and header parameters. A value like a 20-digit reference number was accepted as valid JSON, parsed into a JS number, and lost precision beyond `Number.MAX_SAFE_INTEGER` before being sent. Such values now fall back to the comma-split string handling instead of being parsed as a number.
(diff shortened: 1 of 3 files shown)


hanspagel (approved):
thanks, great PR!

PR #10094, opened 2026-09-06
fix(workspace-store): prevent precision loss for numeric strings in array schema parameters

### LINKED ISSUE #10087: Large numeric strings in query parameters (like List<string>) suffer from JavaScript precision loss (Number.MAX_SAFE_INTEGER)

Tyagiquamar:
## Summary
Fixes #10087

When entering a numeric string exceeding `Number.MAX_SAFE_INTEGER` (e.g. `40702810506710000185`) in an array query parameter (such as `type: array` with `items: { type: string }`), `deSerializeSchemaValue` previously ran `JSON.parse(example)`. Because a standalone numeric string is valid JSON, `JSON.parse` parsed it as a JavaScript number without throwing an error, causing IEEE 754 precision loss that truncated the trailing digits into `40702810506710000000`.

## Changes
- In `deSerializeSchemaValue`, verify that the result of `JSON.parse` for `type === "array"` is actually an array (`Array.isArray(parsed)`) before returning it. If it is a scalar/single value or comma-separated list, fall through to string splitting to preserve exact string values without precision loss.
- For `type === "object"`, verify that `parsed` is a non-null object before returning.
- Added unit tests in `de-serialize-parameter.test.ts` verifying precision preservation for single and multiple large numeric string parameters.

### DIFF
--- a/.changeset/kind-planes-hang.md
+++ b/.changeset/kind-planes-hang.md
@@ -1,0 +1,5 @@
+---
+'@scalar/workspace-store': patch
+---
+
+fix(workspace-store): preserve numeric-string precision when deserializing array schema parameters
--- a/packages/workspace-store/src/request-example/builder/header/de-serialize-parameter.test.ts
+++ b/packages/workspace-store/src/request-example/builder/header/de-serialize-parameter.test.ts
@@ -1,8 +1,6 @@
-import type {
-  ParameterObject,
-  ParameterWithSchemaObject,
-} from '@scalar/workspace-store/schemas/v3.1/strict/openapi-document'
 import { describe, expect, it } from 'vitest'
+
+import type { ParameterObject, ParameterWithSchemaObject } from '@/schemas/v3.1/strict/openapi-document'
 
 import { deSerializeParameter } from './de-serialize-parameter'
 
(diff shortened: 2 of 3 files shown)


hanspagel (approved):

hanspagel:
Thanks for this, @Tyagiquamar — and nice catch on the precision-loss bug! 🙏

Unfortunately this got fixed in parallel by #10090, which merged into `main` earlier today. It addresses the same numeric-string precision issue in `deSerializeSchemaValue`, just with a slightly different implementation.

I rebased this branch onto the latest `main` locally to check, and `main` already passes both of the regression tests you added here:
- single large numeric string → `['40702810506710000185']`
- comma-separated values → `['40702810506710000185', '40702810506710000186']`

So the core change is now redundant. I think it makes sense to close this one in favor of #10090 — but genuinely appreciate you digging in and fixing it. Hope to see more PRs from you! 💙

PR #10096, opened 2026-09-06
fix(api-reference): only show one-of hint above multiple oauth scopes

### LINKED ISSUE #10095: Generated docs for OAuth scopes uses one-of when only one scope is present
### What happens?

Details and screenshot in this discord thread: https://discord.com/channels/1135330207960678410/1544891184000925796

Sandbox example: https://sandbox.scalar.com/e/T8_U3#GET/profile

When an endpoint requires only one scope, there should be no "one-of".  Just list the scope.

In general, when the spec includes multiple scopes in the oauth2 list, it requires all of them, so there is really no case where one-of is appropriate.  Simply listing all of the scopes in the list, when its present.  Is sufficient.

### OpenAPI Document

 https://sandbox.scalar.com/e/T8_U3#GET/profile

lazerg:
## Problem

When an operation's security requirement lists a single OAuth scope alongside a scope-free alternative (an API key or bearer token), the reference shows a "one of:" hint above that lone scope. "One of" only makes sense when there's more than one item to pick from, so it reads as confusing noise on a single-scope list. See #10095.

## Solution

`OperationScopes` now only shows the hint when the total scope count across all groups is greater than one. A single scope is listed plainly, even when a scope-free alternative also satisfies auth.

## Checklist

- [x] I added a changeset.
- [x] I added tests.

Closes #10095


lazerg:
The failing test-e2e-api-reference check is the Models nested-property screenshot test, unrelated to this change and not in the repo's required-jobs list. It fails the same way on repeated runs with no code changes, so it looks like a pre-existing timing-sensitive snapshot, not something this PR caused.

hanspagel (approved):
cool!

PR #10102, opened 2026-09-07
feat(api-client): support response interception

hanspagel:
Allow response hooks to replace the body, status, and headers before Scalar displays them. Add `onResponseReceived` and keep token scripts working.

Tested with package tests, a browser test, type checks, lint, and Knip.

## Ticket

See [discussion #2810](https://github.com/scalar/scalar/discussions/2810).


marclave (approved):

amritk (commented):
The hard part of this — stream lifetime across `clone()` — is handled better than I expected. Worth saying explicitly, because it is the thing that would have leaked silently:

```ts
for (const body of [previousResponse.body, clone.body]) {
  if (body && body !== nextResponse?.body && !body.locked) {
    void body.cancel().catch(() => {})
  }
}
```

`clone()` tees, and a tee only propagates cancellation to the source once **both** branches are cancelled — so without this, a hook that ignores the body would leave the discarded branch open and the source would buffer forever with no consumer. Cancelling in `finally` (so it also runs when a hook throws), skipping the branch that became `nextResponse`, skipping locked branches, and deliberately not awaiting the cancel are all the right calls, and the `it.each(['forward', 'transform', 'partial read'])` test proves the source actually reaches `cancel()` in each shape. The reject case is covered too.

**The one thing I would fix before merge: the streaming guard in `toPostmanResponse` only matches SSE.**

```ts
const buffer = response.headers.get('content-type')?.startsWith('text/event-stream')
  ? new ArrayBuffer(0)
  : await response.arrayBuffer()
```

The comment is right that "event streams may never finish" — but so do the other streaming formats, and #10207 and #10176 are adding first-class support for exactly those right now: JSON Lines (`application/jsonl`, `application/x-ndjson`, `application/json-lines`), JSON sequences (`application/json-seq`, `+json-seq`), and streaming multipart. On any of those, `await response.arrayBuffer()` on a long-lived stream hangs, and since post-response scripts are awaited this blocks the response viewer — the exact failure this guard exists to prevent, just via a different content type.

`@scalar/helpers` is gaining a shared stream-format predicate in #10178/#10176 ("Mock and documentation stream serializers share one implementation"). Using that here instead of the `text/event-stream` prefix check would keep the three PRs consistent and stop this from regressing the moment they land. Also note `startsWith` is doing the `; charset=utf-8` tolerance by luck — the repo already has `parseMimeType` for this.

Two smaller things:

**`executeHook` now has a hook-specific branch inside a generic dispatcher.** The `if (hookName === 'responseReceived')` early return with its own loop and two casts works, but it means the function no longer does one thing, and the casts defeat the `HookPayloadMap` typing that the rest of it relies on. A separate exported `executeResponseHook` with a precise signature would read better and let `executeHook` stay generic. Not a correctness issue — a maintainability one, and this file is going to keep growing.

**A locked clone leaks its branch.** If a plugin takes a reader on the clone and never calls `releaseLock()`, `body.locked` stays true, `finally` skips it, and the tee never propagates cancellation. Nothing this code can do about it (you cannot cancel a locked stream), but it is a real footgun for plugin authors and the docs section is the place to say "release the lock or do not take one" — the docs currently only warn about reading unbounded streams.

Everything else reads well: the config callback is threaded through both `@scalar/types` and `@scalar/schemas` with sync-and-async coercion tests on each, `onRequestSent` and `onResponseReceived` compose correctly in `mapConfigPlugins` (including the removal case), the `response.url || fetchedResponse.url || requestPayload[0]` fallback keeps the real destination for a constructed replacement, and there is an actual e2e test driving the modal. `minor` bumps are right for a new public API.

Happy to approve once the streaming content-type check is generalised.

---

PR #10113, opened 2026-09-07
fix(openapi-parser): handle component keys safely

hanspagel:
Ignore unsafe component keys and check own properties when joining API descriptions.

### DIFF
--- a/packages/openapi-parser/src/utils/join/join.ts
+++ b/packages/openapi-parser/src/utils/join/join.ts
@@ -1,3 +1,4 @@
+import { isPollutionKey } from '@scalar/helpers/object/prevent-pollution'
 import { bundle } from '@scalar/json-magic/bundle'
 import type {
   ComponentsObject,
@@ -71,7 +72,11 @@ const mergePaths = (inputs: PathsObject[]) => {
     }
 
     for (const [path, pathItem] of Object.entries(paths)) {
-      if (!result[path]) {
+      if (isPollutionKey(path)) {
+        continue
+      }
+
+      if (!Object.hasOwn(result, path)) {
         // If the path does not exist, add it directly
         result[path] = pathItem
         continue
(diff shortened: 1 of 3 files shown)


packages/openapi-parser/src/utils/join/join.ts
```diff
@@ -71,7 +72,11 @@ const mergePaths = (inputs: PathsObject[]) => {
     }
 
     for (const [path, pathItem] of Object.entries(paths)) {
-      if (!result[path]) {
+      if (isPollutionKey(path)) {
+        continue
+      }
+
+      if (!Object.hasOwn(result, path)) {
```
amritk: ### Falsy path items now crash join

**Medium Severity**


`mergePaths` swaps `!result[path]` for `Object.hasOwn(result, path)`, so a path whose value is `null` or `undefined` no longer takes the "add it directly" branch; the next document with the same path falls through to `Object.keys(result[path])` and throws `TypeError: Cannot convert undefined or null to object`. Because `join` reverses the inputs before merging, this fires whenever a *lower-precedence* document carries such an entry — `join([{ paths: { '/a': { get: {} } } }, { paths: { '/a': undefined } }])` returns `{ ok: true }` on `main` and throws on this branch, and `webhooks` is affected identically since it reuses `mergePaths`. The matching `Object.hasOwn(result[key], name)` swap in `mergeComponents` turns the same shape into a spurious `component` conflict, so `join` returns `{ ok: false }` for documents it merged cleanly before. Own-property checks are the right fix for the inherited-key problem, but they also change how falsy values are treated, which is not what the changeset describes.




<details>
<summary>Additional Locations (1)</summary>

- [`packages/openapi-parser/src/utils/join/join.ts#L176`](https://github.com/scalar/scalar/blob/cec61ea9c460e3fb0029493528353956c88209af/packages/openapi-parser/src/utils/join/join.ts#L176)

</details>

<sup>Reviewed at cec61ea9c460e3fb0029493528353956c88209af</sup>

---

marclave (approved):

PR #10117, opened 2026-09-07
fix(mock-server): limit remote reference loading

hanspagel:
Limit remote reference time, size, and count. Adds configurable limits and tests.

Builds on #10112.


hanspagel:
I don't know man, who would use that … just pass an OpenAPI document that you trust.

PR #10120, opened 2026-09-07
fix(snippetz): disable curl globbing for URLs with square brackets

### LINKED ISSUE #10119: shell/curl snippet generator breaks on bracket-notation query params (e.g. JSON:API filter[x]=y)
**Package:** `@scalar/snippetz`
**File:** `packages/snippetz/src/plugins/shell/curl/curl.ts`
**Confirmed present on:** `main`, and in the currently published package (tested `@scalar/snippetz@latest` directly)

**What happens**

Given an OpenAPI operation with a query parameter literally named using bracket notation — e.g. `filter[user_id]` (the JSON:API filtering convention) — the generated cURL snippet is:

```
curl 'https://example.com/api/users?filter[user_id]=me' \
  --header 'Authorization: Bearer YOUR_SECRET_TOKEN'
```

Copy-pasting this into a shell and running it fails:

```
curl: (3) bad range in URL position 38:
https://example.com/api/users?filter[user_id]=me
                                     ^
```

**Why**

`curl`'s own argument parser treats unescaped `[` and `]` in a URL as its globbing/range syntax, independently of shell quoting. The generator's `isShellSafe` check in `curl.ts` correctly detects that the URL needs shell-quoting and wraps it in single quotes:

```ts
const isShellSafe = /^[A-Za-z0-9._~:/%@+,=-]*$/.test(url)
const urlPart = isShellSafe ? url : `'${escapeSingleQuotes(url)}'`
```

...but shell-quoting only protects against the *shell* interpreting the brackets — it does nothing about `curl`'s own URL/glob parser. Nothing in the file adds `-g`/`--globoff`, and nothing percent-encodes `[`/`]`.

There's also a query-string-building comment that doesn't match its actual behavior — it claims to encode, but doesn't:

```ts
.map((param) => {
  // Ensure both name and value are fully URI encoded
  return `${param.name}=${param.value}`
})
```

No encoding actually happens here.

**Why other targets aren't affected**

Every other language plugin in this package (JS/Node `fetch`, Node `undici`, Python, Ruby, Rust, C#, PHP/Laravel) shares `buildQueryString()` from `packages/snippetz/src/libs/http.ts` and renders fine with raw brackets, because `[`/`]` are ordinary, unreserved characters to every one of those HTTP client APIs. `shell/curl.ts` is the only target with its own string-building path, and the only one where `curl`'s glob parser makes the extra escaping matter.

**Suggested fix**

In `curl.ts`, either:
1. Always append `-g` / `--globoff` to the generated command, or
2. Percent-encode `[`, `]`, `{`, `}`, and `*` (curl's other glob-significant characters) when building the URL for this target specifically.

Option 1 is simpler and has no behavioral downside for API-snippet use cases — nobody wants curl's shell-glob expansion in a generated request example.

**Repro**

Reproduced directly against the published package:

```js
const { snippetz } = require('@scalar/snippetz')
snippetz().print('shell', 'curl', {
  method: 'GET',
  url: 'https://example.com/api/users',
  queryString: [{ name: 'filter[user_id]', value: 'me' }],
  headers: [{ name: 'Authorization', value: 'Bearer YOUR_SECRET_TOKEN' }],
})
```

produces the broken snippet above, and running that exact string through real `curl` reproduces the `bad range` error shown above.

Any OpenAPI 3.x document with a query parameter literally named `filter[user_id]` (or similar bracket-notation name) reproduces this in the "Test Request" cURL tab. Real-world example: CircleCI's public API v3 docs, e.g. https://circleci.com/docs/api/v3#tag/users/GET/api/v3/users.

lazerg:
## Problem

A query parameter whose name uses bracket notation, like the JSON:API `filter[user_id]` convention, produces a curl snippet that fails as soon as you paste it into a shell:

```
curl: (3) bad range in URL position 38:
https://example.com/api/users?filter[user_id]=me
                                     ^
```

curl reads `[` and `]` in a URL as its own range syntax, and it does that before libcurl ever sees the URL, so the single quotes the generator already adds do not help. Only curl itself can be told to stop.

Every other target in the package is fine here, because brackets are ordinary characters to those HTTP clients. `shell/curl` is the only one where curl's own parser gets in the way.

## Solution

Append `--globoff` when the URL contains a bracket. The URL itself stays untouched, so the snippet still reads like the endpoint it documents, and it now runs as it is. Snippets without brackets are unchanged.

Percent-encoding the brackets instead would work too, but it hides what the parameter is actually called, and #6466 showed that people rely on those names staying readable.

Curly braces have the same problem in curl, but I left them alone on purpose. They almost always mark a path placeholder like `/planets/{planetId}`, which you replace before running anything, so adding the flag there would put `--globoff` on nearly every snippet for no real gain. There is a test pinning that.

## Checklist

- [x] I added a changeset.
- [x] I added tests.

## Ticket

Closes #10119

### DIFF
--- a/.changeset/fix-curl-globbing.md
+++ b/.changeset/fix-curl-globbing.md
@@ -2,4 +2,4 @@
 '@scalar/snippetz': patch
 ---
 
-Add `--globoff` to generated curl commands whose URL contains square brackets, so snippets for bracket-notation query parameters such as `filter[user_id]=me` can be pasted into a shell and run as they are
+Add `--globoff` to generated curl commands whose URL contains square brackets, or curly braces in the query string, so snippets for bracket-notation query parameters such as `filter[user_id]=me` and glob-set values such as `ids={1,2,3}` can be pasted into a shell and run as they are. Curly braces in the path, which are almost always placeholders like `/users/{id}`, are left untouched.
--- a/.changeset/fix-curl-globbing.md
+++ b/.changeset/fix-curl-globbing.md
@@ -1,0 +1,5 @@
+---
+'@scalar/snippetz': patch
+---
+
+Add `--globoff` to generated curl commands whose URL contains square brackets, so snippets for bracket-notation query parameters such as `filter[user_id]=me` can be pasted into a shell and run as they are
--- a/packages/snippetz/src/plugins/shell/curl/curl.test.ts
+++ b/packages/snippetz/src/plugins/shell/curl/curl.test.ts
@@ -163,6 +163,15 @@ describe('shellCurl', () => {
     expect(result).toBe(`curl 'https://galaxy.scalar.com/planets/{planetId}?limit=10'`)
   })
 
+  it('turns off globbing for braces already present in the URL query', () => {
+    const result = shellCurl.generate({
+      url: 'https://example.com/api?ids={1,2,3}',
+    })
+
+    expect(result).toBe(`curl 'https://example.com/api?ids={1,2,3}' \\
+  --globoff`)
+  })
+
   it('has cookies', () => {
     const result = shellCurl.generate({
       url: 'https://example.com',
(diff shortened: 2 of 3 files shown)


hanspagel (approved):
thank you so much!

PR #10125, opened 2026-09-08
fix(api-reference): preserve standalone configuration sources

ryux1:
## What changed

- preserve a `sources` array supplied through the standalone `data-configuration` attribute
- keep the existing single-source parsing and migration path unchanged
- add regression coverage for two configured OpenAPI document sources

The HTML reader previously passed every parsed configuration through the single-source schema before mounting the reference. That schema removes the top-level `sources` property, leaving the normalized configuration list empty and producing `Document not found in configList`.

## Validation

- `pnpm test` in `packages/api-reference`: 1,787 passed, 1 skipped, 3 todo
- `pnpm types:check` in `packages/api-reference`
- `pnpm knip`
- `pnpm changeset status`
- `pnpm build:packages`

## Ticket

See https://github.com/scalar/scalar/discussions/10124

### DIFF
--- a/.changeset/clean-wombats-listen.md
+++ b/.changeset/clean-wombats-listen.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-reference': patch
+---
+
+Preserve multiple sources supplied through the standalone data-configuration attribute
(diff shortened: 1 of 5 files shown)


hanspagel (approved):
ah, thank you!

PR #10129, opened 2026-09-09
feat(release-notes): select a built-in provider by name in the config

DemonHa:
## What

Two gaps in `@scalar/release-notes`:

**1. A config file could not select a built-in provider.** `ReleaseNotesConfig.provider` only accepted a fully constructed provider object, so `release-notes.config.json` — a supported discovery target — could never pick one, because a provider carries a `generateJson` **function** and JSON cannot hold one. Worse, a JSON config with a `provider` key broke silently: `usesBuiltInProvider` evaluated to `false`, the missing-API-key guard was skipped, and the run died later with `provider.generateJson is not a function`.

**2. There was no way to turn off pull request context.** Every run scraped `#123` refs out of the changelog, fetched each PR from the GitHub REST API, and injected up to 4000 characters of PR body into the prompt. The only escape was unsetting `github.repo`, which also breaks changelog links.

## Changes

- `provider` accepts `'anthropic' | 'openai'` or a custom provider object; `model` and `apiKeyEnv` become config fields.
- Provider resolution is consolidated in `read-config.ts`, removing a duplicated block in `generate.ts` and with it the missing-guard bug. An unusable `provider` value now throws a clear error.
- `github.pullRequestContext: false` / `--no-pull-request-context` skips PR fetching entirely — no GitHub API calls, CHANGELOG-only prompt.
- The default system prompt no longer advertises a "Pull request context" block when none was fetched (a pre-existing wart, since the block was already absent whenever zero PRs came back).

### Settings follow their own provider

`model` and `apiKeyEnv` apply only to the provider selected in the same layer of configuration. Selecting a different provider — with `--provider`, or in a config file overriding a base config — drops the model id and API key environment variable that came with the provider it replaced. A config that names no provider counts as selecting the default one.

Without this, `--provider openai` against an Anthropic config would have sent the Anthropic API key to `api.openai.com` as a Bearer token. There is a test that stubs `fetch` and asserts the outbound `Authorization` header.

## Compatibility

Backward compatible at runtime: a config passing a provider object is unchanged. `ResolvedReleaseNotesConfig` changed shape and `PromptContext` gained a field — both type-level, both called out in the changeset, and nothing outside this package imports them. `minor` bump on a `0.1.x` package.

`sync-release-notes-markdown` also reads the config, so it now validates `provider` too even though it does no AI work. That is stricter validation for an AI-free command, and it surfaces a broken config earlier.

## Testing

53 tests pass (`pnpm vitest packages/release-notes --run`), up from 24. New coverage includes a new `read-config.test.ts` (the file had none). Every new guard was mutation-verified: reverting it makes its specific test fail.

Verified end to end against the real repo config and a real changelog, with a `fetch` spy recording outbound hosts:

- `ANTHROPIC_API_KEY= …/cli.js --all --dry-run` — the config-level provider name is accepted and the run skips cleanly with no key.
- Default run → `api.github.com` + `api.anthropic.com`.
- `--no-pull-request-context` → `api.anthropic.com` only, no GitHub calls.

`types:check`, `biome`, `prettier`, and `knip` are clean.

## Ticket

None.

### DIFF
--- a/.changeset/thick-pandas-shave.md
+++ b/.changeset/thick-pandas-shave.md
@@ -1,0 +1,15 @@
+---
+'@scalar/release-notes': minor
+---
+
+feat: select a built-in provider by name in the config and add a pull request context switch
+
+`provider` now accepts `'anthropic'` or `'openai'` in addition to a custom provider object, and `model` and `apiKeyEnv` are config fields, so a config file (including `release-notes.config.json`) can pick a built-in provider without importing a factory. A `provider` value that is neither a known name nor a provider object now throws a clear error instead of failing later with `provider.generateJson is not a function`.
+
+A config-level `model` now reaches the provider, custom providers included, where previously only `--model` did. Set `github.pullRequestContext: false` (or pass `--no-pull-request-context`) to skip fetching pull request titles and descriptions and generate from the CHANGELOG alone. The default system prompt no longer describes a pull request context block when none was fetched.
+
+`model` and `apiKeyEnv` only apply to the provider their own layer of configuration selects, so settings written for one provider are never handed to another. Selecting a different provider with `--provider`, or in a config file that overrides a base config, drops the model id and API key environment variable that came with the provider it replaced. A config that names no provider counts as selecting the default one.
+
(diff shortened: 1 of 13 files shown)


hanspagel (approved):
👍

PR #10150, opened 2026-09-09
fix(helpers): let the dockerized playwright server accept launch options

hwkr:
Since #10103 the font flags only ever applied in CI. Locally the suite connects to the browser server in the container, and `run-server` throws away `args` unless you start it with `--unsafe` — so `pnpm test:e2e:update-snapshots` was quietly producing baselines CI couldn't reproduce. One flag on the `docker run` fixes it.

* `models.e2e.ts` goes from 0/8 to 7/8 byte-identical against the committed baselines (the 8th differs by a hover-state `#` anchor, not rendering)
* Storybook `SchemaProperty` goes 0/3 → 3/3
* #10103's flags turn out to fix nothing measurable (CI is already bit-identical across six hosts) — but that's a separate call, and this makes local match CI whichever way it goes
* `api-reference` is the only package that sets `launchOptions`, so `components` and `scalar-app` renders are unaffected

## Checklist
- [x] I added a changeset.
- [ ] I added tests.


hanspagel (approved):
I see unsafe

I approve

👍

PR #10152, opened 2026-09-09
docs(sso): require the Entra ID signing option Scalar accepts

amritk:
## Problem

Customers who follow the Microsoft Entra ID guide to the letter end up with SAML sign-in failing at `https://identity.scalar.com/acs` with a 401 "Invalid SAML signature state".

Entra defaults the **Signing Option** in the SAML Certificates card to **Sign SAML assertion**. The guide has customers enable token encryption (step 7) but never mentions the signing option. With encryption on and only the assertion signed, Scalar's ACS handler cannot re-verify the assertion signature and rejects the login. The working Entra tenants all use **Sign SAML response and assertion**; the ones hitting the 401 were left on the default. The guide currently gives no way to discover this.

## Solution

- Add a new step 7, **Set the Signing Option**, telling customers to switch to **Sign SAML response and assertion** before enabling token encryption, with a callout explaining why the default fails.
- Renumber token encryption to step 8 and note that leaving it off sends users back to the Scalar login page (a different symptom from the signature error).
- Add a **Troubleshooting** section mapping the three symptoms seen in support to their fix: the signature-state error, the silent redirect to login, and users not matching an account because the Name ID claim is not an email address.

The new step has no screenshot yet. Every other step on this page has one, so a capture of the Signing Option dropdown in the SAML Certificates edit pane should be added before this leaves draft.

## Checklist

- [ ] I added a changeset. (docs only, no package change)
- [ ] I added tests. (docs only)

https://claude.ai/code/session_01EJL1TUmyd8LDy1B8xrZhoc

---
_Generated by [Claude Code](https://claude.ai/code/session_01EJL1TUmyd8LDy1B8xrZhoc)_

### DIFF
--- a/documentation/guides/sso/providers/microsoft-entra-id.md
+++ b/documentation/guides/sso/providers/microsoft-entra-id.md
@@ -74,9 +74,15 @@ Copy the IdP details from Microsoft Entra ID into Scalar:
 
 ![](../../../assets/sso/microsoft-entra-id/scalar-service-provider-details.png)
 
-## 7. Enable Token Encryption
+## 7. Set the Signing Option
 
-Scalar requires token encryption for SAML assertions. Download the encryption certificate from the Scalar and import it into Microsoft Entra ID under **Token encryption**:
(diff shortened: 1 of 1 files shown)


hanspagel (approved):
thanks

PR #10158, opened 2026-09-11
test(openapi-validator): check duplicate required properties

hanspagel:
Enable the test for duplicate required property names. Check the exact error message and schema path.


amritk (approved):
Test-only: turns a `describe.todo` into a real assertion, and pins both the message and the schema path rather than just the count. Changeset included, CI green.

Approving.

---

PR #10159, opened 2026-09-11
test(api-reference): cover nested schema selector titles

hanspagel:
Complete the nested schema selector test. Check option titles and displayed fields when changing outer and inner selections.

### DIFF
--- a/.changeset/nested-composition-test.md
+++ b/.changeset/nested-composition-test.md
@@ -1,0 +1,5 @@
+---
+"@scalar/api-reference": patch
+---
+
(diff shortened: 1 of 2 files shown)


amritk (approved):
Test-only. Replaces the `it.todo` placeholder with a test that actually exercises the nested `oneOf` → `allOf` → `anyOf` path: option titles, switching outer and inner selections, and that the sibling `allOf` field (`customerComment`) stays visible throughout. Much better than the stub it replaces.

Approving.

---

PR #10160, opened 2026-09-11
fix(pre-post-request-scripts): expose response duration

hanspagel:
Pass request duration to post-response scripts as `pm.response.responseTime` in milliseconds. Enable the response time example.

## Visual

Before:

Before

After, with the response time example enabled:

After


amritk (approved):
Clean plumbing change. `responseDuration` is optional at every hop — the `ClientPluginHooks.responseReceived` payload, `executePostResponseScript`, `executeInPostmanSandbox`, and `toPostmanResponse` — so existing plugins and callers that do not supply it are unaffected, and `pm.response.responseTime` is simply absent rather than `0`, which the new `sandbox-adapter` test pins.

The detail I checked specifically: `...(responseDuration === undefined ? {} : { responseTime: responseDuration })` is an explicit `undefined` check rather than a truthiness check, so a genuine `0 ms` duration still reaches the sandbox. The plugin test covers exactly that by parameterising over `[0, 125.5]`.

Enabling the previously commented-out "Check response time" example now that the value actually exists is the right follow-through, and the `scalar-app` snapshot is updated to match.

Changeset covers all three affected packages. CI green.

Approving.

---

PR #10161, opened 2026-09-11
test(openapi-to-markdown): add output evaluation harness

hanspagel:
Add a fixed evaluation corpus for Markdown features and output quality. Save scores and rendered Markdown for comparison. The baseline passes 27 of 37 checks.

### DIFF
--- a/.changeset/tidy-markdown-evaluation.md
+++ b/.changeset/tidy-markdown-evaluation.md
@@ -1,0 +1,5 @@
+---
+'@scalar/openapi-to-markdown': patch
+---
+
+Add a repeatable evaluation harness for Markdown output quality and feature coverage.
--- a/packages/openapi-to-markdown/evaluation/README.md
+++ b/packages/openapi-to-markdown/evaluation/README.md
@@ -1,0 +1,25 @@
+# Markdown evaluation
+
+Run from the repository root after installing dependencies and building packages:
+
+```sh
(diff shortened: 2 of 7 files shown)


amritk (commented):
Conflicting with `main` — needs a merge from base before this can land.

On the substance, the idea of a fixed evaluation corpus for Markdown output quality is good, and test-only changes are the cheapest kind to land. Two questions about the shape of the harness before I approve:

**Does a baseline of "27 of 37" fail CI?** A harness that records a score is only useful if regressing it is caught. If the 10 failures are known gaps that #10162 is meant to close, the harness should assert against a committed baseline number (or a per-check expectation file) so a drop from 27 breaks the build, rather than just writing a score somewhere. If it does already do that, say so in the description — it is the first thing a reviewer wonders.

**Are the rendered Markdown outputs committed as snapshots?** "Save scores and rendered Markdown for comparison" could mean either written to a gitignored directory at run time, or committed and diffed. Committed snapshots are much more useful for review — a reviewer of #10162 could then see the exact output change — but they need to be regenerable with a documented command.

Since this and #10162 are a pair, it is worth deciding which lands first. Landing this one first with the baseline pinned at 27 makes #10162's improvement visible as a score change in its diff, which is a genuinely nice review artifact.

Not approving while it conflicts.

---

hanspagel:
I think that's worthless. Just need to be tests.

PR #10162, opened 2026-09-11
fix(openapi-to-markdown): preserve examples and schema details

hanspagel:
Preserve supplied request and response examples, including falsy values and raw XML, and render response headers, links, and form/multipart encoding details. Improve schema output for composition siblings, constraints, readOnly/writeOnly properties, additional properties, discriminators, and recursive versus deep nonrecursive schemas.


amritk (commented):
Conflicting with `main` — needs a merge from base first.

This is 966 additions across 13 files and changes Markdown output in a lot of dimensions at once: example preservation (including falsy values and raw XML), response headers, links, form/multipart encoding, composition siblings, constraints, readOnly/writeOnly, additionalProperties, discriminators, and recursive-vs-deep-nonrecursive handling. Each of those is a defensible change; together they make it hard to tell which output difference came from which rule.

That is exactly what #10161's evaluation harness is for. Strong suggestion: **land #10161 first with its baseline pinned, then rebase this on top** so the diff shows the score moving and, if the harness commits rendered output, the actual Markdown changes are reviewable line by line. Right now a reviewer has to take the output changes on faith.

One thing worth calling out explicitly in the description regardless: preserving falsy example values is a behaviour change for existing documents — a `0`, `false` or `""` example that previously rendered as absent will now render. That is the correct fix, but anyone diffing generated Markdown after upgrading will see churn that has nothing to do with the new features, and it deserves a line in the changeset.

Happy to review properly once it merges cleanly and, ideally, sits on top of the harness.

---

hanspagel:
probably worthless with #10233

hanspagel:
reimplemented in https://github.com/scalar/scalar/pull/10275

PR #10165, opened 2026-09-12
fix(workspace-store): truncate deep examples in the declared type

DemonHa:
## Problem

`getExampleFromSchema` gives up below ten levels and, past that, returned the literal string `[Max Depth Exceeded]` no matter what the schema declared there. The stand-in is still read as an example of the schema it replaced, so the generated example contradicted its own document: a `type: object` field came back as a string.

That is not only cosmetic. `@scalar/mock-server` serves these examples as response bodies, so it answered a declared object with a string, and a strict consumer cannot parse it — a generated Rust SDK decoding a mocked response fails with `invalid type: string "[Max Depth Exceeded]", expected struct PriceFilter`. Before this change, a 30-deep response schema served:

```json
{"next":{"next":{ … {"next":"[Max Depth Exceeded]"} … }}}
```

and after it serves `…{"next":{}}}…`, which satisfies the schema it declares.

## Solution

The cap now describes the declared type instead of emitting one string for every case: `{}` for objects, `[]` for arrays, a placeholder of the declared primitive for scalars, and `null` where the schema names it. Only schemas that describe no shape at all keep the sentinel, so the signal that truncation happened is still reachable.

Four things fell out of making the stand-in faithful rather than fixed:

- **The cap moved below the value-precedence block.** It used to sit above `shouldOmitProperty`, `x-variable` and `examples`/`example`/`default`/`const`/`enum`, so a truncated `{ type: 'string', enum: ['active', 'archived'] }` was answered with a value its own enum forbids. A schema that states what it holds is now taken at its word at any depth, and a property excluded by `mode` is omitted there rather than replaced by a placeholder.
- **Composition wrappers are unwrapped.** `allOf: [$ref]` declares no type of its own, and it is the ordinary way a generated document attaches a description to a shared type — so it is described by the member the walk itself would have rendered, honouring an explicit `compositionSelection`, including a selection on an `allOf`-wrapped `oneOf`. Unwrapping runs outside the walk's own cycle guard, so it carries one; the guard set doubles as the path length, which bounds the stack against a pathological chain.
- **Type is read as a list**, so `type: ['object']` behaves like `type: 'object'` rather than falling through to a string.
- **Results built around a truncation are no longer cached.** This one is a pre-existing bug rather than something this branch introduced — `resultCache` keys carry no level, and an `allOf` member climbs a level without growing the schema path, so a schema rendered just above the cap in one chain shares its key with the same schema used at the top of a document, and the shallow use was served the truncated example. On `main` the same construction serves `[Max Depth Exceeded]` there. Replacing a loud wrong value with a plausible empty one would have hidden it, so storing stops for the rest of a call once anything actually drops children.

### Alternative considered

Filling `required` properties at the cap. A truncated `{}` still fails a decoder whose struct fields are not optional, so this does not close that case — but satisfying `required` means descending again, which is what the cap exists to prevent, and a required property that is itself a container regresses immediately. The doc comment records the limitation rather than pretending it away.

### Validation

Every suite that consumes this builder, on the final commit: `workspace-store` 3033 passed, `mock-server` 562, `api-reference` 1973, `api-client` 1825, `blocks` 481, `openapi-to-markdown` 40. Plus `types:check` and `biome check` on the changed files. Each new test was confirmed to fail against the pre-fix source; the cycle guard and the cache suppression were each confirmed load-bearing by removing them (the first overflows the stack, the second serves the poisoned value).

`pnpm knip` exits 2 both with and without this change — a pre-existing module-load error in the docusaurus integration playground, unrelated here.

### Known follow-ups

A composition member is described without going back through the walk, so `shouldOmitProperty` and `x-variable` are not applied to it, and `$dynamicRef` is not resolved for one. Every value produced is still type-valid and nothing that was previously omitted becomes present, so these are fidelity gaps rather than correctness breaks.

## Checklist

- [x] I added a changeset.
- [x] I added tests.

https://claude.ai/code/session_01WP3sifWhdGs9PJhrzQFEDG

---
_Generated by [Claude Code](https://claude.ai/code/session_01WP3sifWhdGs9PJhrzQFEDG)_


DemonHa:
CI is green — all checks pass, including `test-e2e-api-reference` (218), `test-component-snapshots` (203), `test-js`, `types`, `knip` and CodeQL.

One note on the non-blocking CDN snapshot diff, since this PR changes rendering and the count looks PR-specific at a glance. It reports 11 failures where the concurrently open #10158 and #10159 both report 10 — the same ten (Galaxy, Galaxy Registry, Petstore 2.0/3.0/3.1, Zoom, Cloudinary, Bolt, OpenStatus, Spotify), which is the usual drift against the released CDN build.

The one extra here is `AsyncAPI Sidebar Filters`, and it is not this change. That test loads `packages/api-reference/public/examples/asyncapi-filters.yaml`, whose three payload schemas are flat objects of scalar properties with no composition keywords anywhere in the document and no recursive `$ref`s — nothing in it reaches schema depth 3, while the truncation this PR touches fires only past level 10. There is no code path by which this diff can be caused by the change.

---

amritk (approved):

PR #10166, opened 2026-09-12
fix(scalar-app): make workspace settings options readable while pressed

DemonHa:
## Problem

The option rows on the workspace settings page (CORS Proxy, Themes, Framework Themes, Custom Themes, Appearance) turn into a solid black bar with unreadable dark text for as long as you hold the mouse down on them. It was reported as "they look weird when clicked in them" — the reporter's screenshot was captured mid-click.

The rows styled their selected state with `bg-primary`, **a class that does not exist in the Scalar Tailwind theme**: `packages/themes/src/tailwind/theme.css` resets `--color-*: initial` and never defines `--color-primary`. It produces no CSS at all, but tailwind-merge still classifies it as a background utility, so it stripped `ScalarButton`'s own `bg-b-btn` from the solid variant while leaving `active:bg-b-btn` untouched. Pressing a row therefore painted it `--scalar-button-1` (black) while the row's own `text-c-1` kept the label near-black.

Two consequences of the same mistake were visible at rest:

- The selected row ended up with **no background at all** (the phantom class replaced the real one), so the only selection affordance was the checkmark.
- `hover:bg-inherit` made the selected row adopt the page background on hover.

Separately, the theme color swatches carried `border-c-3` with no border-width utility, so the border never rendered and the white/light swatch of each theme was invisible against the white page — visible in the reporter's screenshot, where "Default" appears to have only two dots.

## Solution

Extract the rows into a shared `SettingsOption` component built on `ScalarButton`'s **ghost** variant, which has no dark backgrounds to override, and set `hover:` / `active:` backgrounds explicitly so no variant style can leak through again. The selected row sits on the plain raised background (`bg-b-2`) and the radio alone carries the selection, so no accent color appears on the row in any state. The radio indicator is the design system's `ScalarCheckbox type="radio"` rather than a hand-rolled circle. The swatches get a real border width so light swatches are visible.

This also removes the duplicated `cva` block and checkmark helper that `CollectionSettings.vue` and `Appearance.vue` each carried a copy of.

Alternatives considered:

- **`variant="outlined"` instead of ghost.** Its unselected styling is equivalent, but it draws its border with `shadow-border`, which makes a bordered selected state fiddlier than a plain `border`. Ghost plus an explicit border was simpler.
- **An accent border and tint for the selected row**, matching `packages/api-client/src/v2/features/environments/components/Environment.vue`. That is what this PR originally did; dropped in review in favor of the neutral treatment above.
- **Replacing the swatches with `ScalarThemeSwatches`.** Tempting, since `get-theme-colors.ts` hardcodes hex triples that can drift from `packages/themes/src/presets/*.css`, and the api-reference theme picker already uses it. Rendered, the 7-color strip came out ragged (themes whose CSS omits some variables leave gaps), Solarized read muddy, and the "None" pseudo-theme lost its swatch entirely — worse than what is there today. `get-theme-colors.ts` is left byte-identical to `main`; fixing the drift without the visual regression belongs in its own change.
- **`role="radiogroup"` instead of `aria-pressed`.** No precedent anywhere in the repo, and a half-implemented radiogroup (every option a tab stop, no arrow keys) is worse for keyboard users than the current buttons. `ScalarCheckboxRadioGroup` would fit the Appearance section but not the theme grids, which need per-option slots for swatches and logos.

### Verification

Reproduced the original failure in a browser against the real component, then confirmed the fix in light and dark mode. `pnpm --filter scalar-app types:check` passes; 169 tests pass in `projects/scalar-app/src/features/{settings,collection}`, including 5 new ones for `SettingsOption` — one of which asserts the dark button backgrounds (`bg-b-btn`, `active:bg-b-btn`, `hover:bg-h-btn`) never reach the rendered row, so this specific regression cannot come back silently.

`grep` confirms zero remaining `bg-primary` occurrences repo-wide.

The `workspace-tab-settings` tour baseline is refreshed because the rows changed shape; every other snapshot in that suite is unchanged.

## Checklist

- [x] I added a changeset.
- [x] I added tests.

https://claude.ai/code/session_0176auAokfKU3zqqFi1g3N4r


amritk (approved):

hwkr (approved):
Looks good! I think my only tweak would be not doing the blue state when active, but not required

DemonHa:
Thanks @hwkr — done in ab7c526.

"The blue state when active" could have meant either the selected row's accent border and tint, or the deeper accent tint on `:active` while the row is pressed, so I removed both. The rows now use `bg-b-2` when selected and `bg-b-1` otherwise, with `hover:`/`active:` landing on `bg-b-2` either way, so no accent color appears on the row in any state and the radio alone carries the selection. Verified in light and dark mode.

If you meant only one of the two and want the other kept, say which and I will put it back.

---

Issue #7986, opened 2026-01-29
`hiddenClients: true` doesn't hide code snippet section


silkfire:
### What happens?

Maybe this is not intended or perhaps I'm misunderstanding the option, but currently `hiddenClients: true` doesn't hide the code snippet section, instead it shows it as empty:

<img width="705" height="96" alt="Image" src="https://github.com/user-attachments/assets/140ecbc7-2347-4fe9-a8e8-db55fa7930aa" />

### What did you expect to happen?

I want it to hide the code snippet section entirely, or add a dedicated option for this.

amritk:
Not a bad idea eh @hanspagel, if there's no clients hide the code snippet

hanspagel:
yep! that’s what we used to do :D

silkfire:
Would this be considered a bug or does the option need to be tweaked in order to support both behaviors?

hanspagel:
just a bug! :) let me fix it

silkfire:
I don't think this is working as intended. I'm running `hiddenClients: true` (but with .NET) and the collapsible is still displayed. How do I make the array of clients empty? Providing an empty array means that it enables all clients.

amritk:
@hanspagel

silkfire:
@hanspagel Any status update?

timwhitlock:
With `hiddenClients: true` I see the "Code Snippets" tab with the contents just displaying `"1"`.  

This is version 1.57.1

PR #10167, opened 2026-09-12
fix(api-client): hide code snippet section when all clients are hidden

silkfire:
## Problem

image
<br><br>

With `hiddenClients: true`, the API client still renders an empty **Code Snippet** section: a collapsible with no client selected and a blank code block showing just line `1`.

This is [#7986](https://github.com/scalar/scalar/issues/7986), which [#7992](https://github.com/scalar/scalar/pull/7992) intended to fix. That PR's logic was right (no clients → hide the section), but the mechanism it used never took effect, so the issue is still reproducible on current `main` (also reported against 1.57.1 in [this comment](https://github.com/scalar/scalar/issues/7986#issuecomment-4458594325)).

The cause is two competing `v-show` directives on the same element:

- `RequestCodeSnippet.vue` puts `v-show="hasClients"` on its `CollapsibleSection` root.
- `RequestBlock.vue` puts `v-show="selectedFilter === 'All'"` on the `<RequestCodeSnippet>` component itself.

Both write `style.display` on the *same* root element, and the parent's directive is applied last. With the filter on `All` (the default) it resets `display` to `''`, wiping the `display: none` the component just set.

Verified by mounting: on its own the component gets `style="display: none"`; wrapped in a parent with a truthy `v-show` it gets `style=""` and the section becomes visible.

## Solution

Swap the component's `v-show="hasClients"` for `v-if`, so when there are no clients there is no element left for the parent's directive to un-hide.

The existing visibility test was asserting on a stub in isolation, which is why the clash was invisible to it. It now checks `.exists()`, and a regression test mounts the component under a truthy parent `v-show` to cover the actual failure mode.

## Checklist

- [x] I added a changeset. 
- [x] I added tests.

### DIFF
--- a/.changeset/plain-pianos-listen.md
+++ b/.changeset/plain-pianos-listen.md
@@ -1,0 +1,5 @@
+---
+'@scalar/api-client': patch
+---
+
+fix: hide the code snippet section when all clients are hidden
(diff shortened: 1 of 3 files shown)


hanspagel (approved):
ooops

PR #10168, opened 2026-09-14
fix(api-reference): resolve AsyncAPI message traits

hanspagel:
## Problem

AsyncAPI messages hide headers and other fields inherited from traits.

## Solution

Merge traits in order and keep message-level overrides.

## Visual

image

## Ticket

Related to [T-2002](https://app.plain.com/workspace/w_01JJA9M6A139BDX8ETB876C1K7/thread/th_01M1VK6MC4B8JQ8RVT6JXRN80S/).


cameronrohani (approved):

amritk (approved):
This is a careful JSON Merge Patch implementation and the test file reads like someone went looking for the traps rather than waiting to be told about them. The ones that matter:

- **Literal `$ref` keys are not mistaken for references.** The `MergeContext` machinery is the interesting part — `data` is absorbing (once inside `example`/`default`/`enum`/`x-*` you never resolve again), and `map` flips children back to `object` so a *property named* `$ref` under `properties` is a schema, not a pointer. The "preserves header names and literal data containing `$ref`" test pins exactly this, and it is the bug I would have expected a naive version to ship with.
- **`__proto__` cannot pollute.** `Object.defineProperty` with `enumerable/configurable/writable` instead of `result[key] = …`, plus a test asserting `Object.getPrototypeOf(result.headers) === Object.prototype`.
- **Cycles are reproduced rather than shared or infinite.** `ancestors` is a path-scoped map that is deleted on unwind, so a recursive header schema comes back deep-equal but not identical — `expect(result.headers).not.toBe(headers)` — while the source document is left untouched (`expect(message).toStrictEqual(original)`).
- **`null` removes a key** at both trait and message level, which is the part of merge-patch people usually skip.

I checked the one thing that looked like it could throw: `resolveAsyncApiMessage` now pipes through `resolveMessageTraits`, and `resolveMessageTraits` starts with `message.traits?.length` — which would blow up on `undefined`. It cannot get there, because the call uses `getResolvedRef(node, mergeSiblingReferences)` and `mergeSiblingReferences` never returns `undefined`: when `$ref-value` is missing or not an object it returns the siblings (`get-resolved-ref.ts:24`), so at minimum `{ $ref }`. The single-argument `getResolvedRef(trait)` inside the reduce *can* be `undefined`, and that one is guarded — covered by the "ignores unresolved traits" test.

The override layering is right too: only trait-supplied keys get merge-patched, everything the message owns exclusively (payload, and `traits` itself) passes through untouched, so message-level fields win without clobbering unrelated ones. The component test confirms the visible outcome — `Local title` wins over `Trait title` while the trait's description and `correlationId` header are still inherited alongside the message's own `tenantId`.

Changeset present, CI green.

Approving. One non-blocking note: `mergeTraitFields` deep-copies the whole trait path on every call, so if `resolveAsyncApiMessage` is ever called outside a `computed` this becomes per-render work on potentially large header schemas. Fine as is — just worth not losing the memoisation if this moves.

---

PR #10170, opened 2026-09-14
fix(api-client): match composed form checkboxes to sent fields

### LINKED ISSUE #10144: Optional form-body properties are shown unchecked but still sent when the body schema uses allOf / oneOf / anyOf
### What happens?

#10046 made optional `multipart/form-data` and `application/x-www-form-urlencoded` properties start unchecked in the Test Request panel. When the body schema uses composition (`allOf`, `oneOf` or `anyOf`) the panel unchecks them and the request builder sends them anyway. Press Send on a panel you haven't touched and it transmits fields the panel shows as disabled. No error, nothing on screen.

The two sides disagree about what optional means. The builder skips the drop as soon as the body schema carries a composition keyword:

https://github.com/scalar/scalar/blob/ba0caf253feda3c5c5eac57e62bc40e62dbb579f/packages/workspace-store/src/request-example/builder/body/build-request-body.ts#L92-L101

With `isComposed` true, `bodyProperties` is `undefined`, `isOptionalBodyProperty()` always returns `false`, and nothing is dropped.

The panel has no equivalent guard. It checks `isObjectSchema()` and then unchecks any top-level declared property that isn't in `required`:

https://github.com/scalar/scalar/blob/ba0caf253feda3c5c5eac57e62bc40e62dbb579f/packages/api-client/src/v2/blocks/request-block/helpers/get-form-body-rows.ts#L170-L182

The `isComposed` guard came out of the second review round on #10046 and the reasoning is sound: a property can be required inside a composition member that isn't merged at that point. The panel side never got the matching guard. The third bullet of that commit message shows the same failure mode was already understood for nested leaves:

> scope the panel's auto-disable to top-level declared properties so a nested leaf is not shown unchecked while still being transmitted

### Reproduction

https://sandbox.scalar.com/e/14nTy

Two endpoints with the same three properties: `name` (required), `note`, and `mode` (an `enum`). `PATCH /items` also carries an `anyOf` next to `properties`, which is a common way to say "exactly one of these two fields is required". `PATCH /things` is the same body without the composition.

Open the Test Request panel for `PATCH /items`. `note` and `mode` are unchecked. `name` is checked and empty, and `label`, required inside its `anyOf` branch, is checked and prefilled with `alpha`. Press Send without editing anything and the body goes out as:

```
name=&note=&mode=none&label=alpha
```

`note` and `mode` are transmitted even though both rows are unchecked.

`PATCH /things`, the same properties with no `anyOf`, sends only `name=`. Deleting the `anyOf` block from `/items` gives that same result, so the composition is what changes the behaviour.

`mode` is the one that hurts. Its value comes from the first `enum` member, so an untouched Send transmits a valid choice the user never made, and the server can't tell it apart from a deliberate one. `note` at least arrives empty, which a server can choose to read as "not provided".

Toggling any row's checkbox converts the example to the array form, which takes the `Form data - array format (from UI editor)` branch in `build-request-body.ts` and honours each row's own `isDisabled`. After that every unchecked row is dropped correctly, including the ones you never touched. So the bug only shows on a panel nobody has edited yet, which is the first thing someone reading the reference is likely to try.

### Environment

- `@scalar/api-reference` 1.68.0 / `@scalar/api-client` 3.18.0 (latest at the time of writing)
- Also present on `main` at `ba0caf253`; neither file has changed since 1.68.0 was published

The request to match the panel. Either the panel gets the same composition guard the builder has, so a composed schema keeps the pre-#10046 behaviour and shows everything checked, or the builder honours the per-row state for composed schemas instead of skipping the drop wholesale. The first is the smaller change and gets back to a consistent state, if a less convenient one.

On a write endpoint an untouched Send can change a setting nobody chose. Before #10046 the row was visibly checked, so at least the value being sent was discoverable.

### OpenAPI Document

https://sandbox.scalar.com/e/14nTy

hanspagel:
Keep form fields checked when the body schema uses `allOf`, `oneOf`, or `anyOf`, so the panel matches what Send includes. Explicit checkbox changes still work.

## Visual

Before
After

## Ticket

Fixes #10144


amritk (approved):
Small and correct. The new `isComposed` guard in `get-form-body-rows.ts` mirrors exactly what `build-request-body.ts` already does (`packages/workspace-store/src/request-example/builder/body/build-request-body.ts:120`), so the checkbox state and the body that Send actually produces now derive from the same rule instead of diverging. Explicit `isDisabled` from a stored form-row array still wins, which the test pins for both content types across all three composition keywords.

Changeset included, CI green.

Approving.

---

PR #10171, opened 2026-09-14
feat(api-reference): show AsyncAPI message examples

hanspagel:
Show AsyncAPI message examples with a picker and copy button. Keep duplicate names and empty values accessible.

Generated example labels receive a unique suffix when they collide with authored names, including names later in the list. Authored names and index-based selection are preserved.

Based on main, independent of #10168.

## Visual

Generated label collision after the fix:

Distinct generated example label

Desktop

<details><summary>Mobile</summary>

Mobile

</details>

## Validation

- API reference package: 2,018 tests passed across the suite and targeted reruns after enabling network access for SSR and building CSS; 1 skipped, 2 todo.
- AsyncAPI tests: 85 passed.
- Scoped lint, formatting, type checking, Knip, and changeset status passed.
- Browser check confirmed the generated label selects its own payload.

## Ticket

Related to T-2002.


cameronrohani (approved):

amritk (approved):
Self-contained feature — a new component and a new helper, no shared code touched — with the fiddly parts handled properly:

- **Keying by array index** rather than by `name` is the right call, and the test proves duplicate names stay independently selectable.
- **Falsy payloads survive.** `example.payload !== undefined ? payload : headers` plus the explicit `null`/`boolean` branch means `null`, `false`, `0` and `''` all render and copy rather than vanishing. Covered by `it.each([null, false, 0, ''])`.
- **The selection recovers when the list changes.** The `watch` resetting `selectedKey` to the first remaining entry is the thing that usually gets forgotten, and there is a test that swaps the examples out from under a selection.
- **Code block and clipboard share one formatted string**, including on the virtualised path — `ScalarVirtualCodeBlock` and `ScalarCopy` both get `content`, asserted at 20,001 characters.
- `prettyPrintJson` handles the recursive payload case rather than throwing.
- The `:has(> .message-details):has(> .message-examples)` container query degrades correctly: `MessageExamples` renders nothing when `content === undefined`, so the `:has` fails and the layout falls back to one column instead of leaving an empty grid track.

I checked the three localization keys — `schema.example`, `schema.examples` and `common.copyExample` all exist in `locales/en.ts`, so no missing-key fallbacks.

CI green, changeset present. Approving.

**One UX thing worth a follow-up**, which your own test actually documents:

```js
expect(picker.props('examples')).toStrictEqual({
  '0': { summary: 'Example 3' },
  '1': { summary: 'Example 3' },
  '2': { summary: 'Example 3' },
})
```

Entries 0 and 1 are named `Example 3` by the author; entry 2 is unnamed and the fallback generates `Example 3` from its index. So all three options in the picker read identically and a user cannot tell them apart. The keys keep them selectable, which is what the test is asserting, but the labels do not.

Cheap fix: when a generated label collides with one already in use, disambiguate it — e.g. fall back to `#{index + 1}`, or suffix generated labels only when the set has duplicates. Not blocking, since it needs an author to name an example exactly `Example N`, but it is a real collision rather than a hypothetical one.

Minor: the two screenshots add ~binary weight to `.github/assets/` permanently. `.github/pr-artifacts/` is used elsewhere in the repo for the same purpose (#10173), so worth being consistent about which one is the convention.

---

PR #10172, opened 2026-09-14
fix(workspace-store): preserve boolean schema semantics

hanspagel:
Boolean schemas currently lose their meaning during workspace ingestion: a `false` schema becomes an empty schema that accepts any value. Normalize boolean schema positions before client and server object coercion, preserving `true` as an empty schema and `false` as its negation.

```yaml
openapi: 3.1.2
info:
  title: Boolean schemas
  version: '1'
components:
  schemas:
    Anything: true
    Nothing: false
    Closed:
      type: object
      additionalProperties: false
```

Previously, `Nothing` became an unconstrained schema. It now becomes `{ not: {} }` in the working model and after saving, while `Anything` becomes `{}`. Original, unsaved exports retain authored booleans. **`additionalProperties: false` stays literal `false` in the working model and in original and saved JSON/YAML exports.**

Server normalization copies only changed schema containers and their ancestors. Unchanged bundled branches remain shared, caller input is not mutated, and shared/circular schema relationships are preserved. Traversal is iterative; other ingestion stages retain their existing depth and resource limits. Examples, defaults, enums, constants, and ordinary vendor extensions remain opaque.

Internal `getRaw` backing data retains the established `__scalar_` marker. Public proxy serialization and schema rendering hide it. Both saved export formats share the existing `getEditableDocument` → `removeExtraScalarKeys` pipeline, so this change introduces no per-format stripping logic. The helper documentation records measured normalization costs and their limits.

## Visual

Ingested schemas render without internal markers, and public serialization preserves the negative schema as `not: {}`. This existing capture documents that behavior; the image is excluded from the PR's source diff.

Boolean schema rendering and public serialization


amritk (commented):
The underlying bug is real and worth fixing — `false` collapsing to `{}` turns "accepts nothing" into "accepts anything", which is the worst possible direction for that failure to go. `normalizeBooleanSchemas` is carefully written: separate `visitedSchemas`/`visitedObjects` sets so a shared object reached in both a schema and a non-schema context is still handled correctly, the recursion terminates on self-referential objects, and `{ __scalar_: '' }` is the established disambiguation marker (`schemas/v3.2/strict/schema.ts:64`) that the magic proxy already hides, so it does not leak into client-visible output. The test matrix is genuinely good — `opaqueValues` keeping `example`/`default`/`const`/`enum` literal, and `components.parameters.example` not being mistaken for an example payload, are exactly the traps here.

**One thing I would not merge without a deliberate decision: `server.ts` now deep-clones every document at ingest.**

```ts
const normalizedDocument = normalizeBooleanSchemas(deepClone(upgradedDocument))
```

The comment this PR deletes from `preserveBundledExternals` was specifically an argument against doing that:

> It would also be the worst place to pay for it: `x-ext` holds every external document that was bundled in, so cloning it roughly doubles peak memory at ingest, and `deepClone` recurses per level and throws on input nested a few thousand deep.

The replacement comment ("The source has already been cloned before coercion, so these buckets do not alias caller data") is accurate about *aliasing*, but it reverses that earlier decision without addressing either cost it named. Concretely, a split-file document with a large `x-ext` bucket now has its entire bundle cloned on every server ingest, and a deeply nested document that previously ingested fine can now hit the `deepClone` recursion limit before it ever reaches coercion.

That may be an acceptable trade for correctness — but it is a memory and robustness change to the SSR ingest path, it is not mentioned in the PR description or the changeset, and the changeset is marked `patch`. Options I would consider, in order of preference:

1. Normalize in place on `upgradedDocument` without cloning, if the caller-aliasing concern can be handled the way the old comment suggested (cloning the caller's document as a whole, where the store already handles aliasing).
2. Keep the clone but scope it so `x-ext` is not duplicated.
3. Keep it as is, and say so explicitly in the PR description plus a note in the changeset, so the next person debugging an OOM or a stack overflow at ingest finds the reasoning.

Two smaller notes:

- `visit` skips any key starting with `x-` except `x-ext`, so a boolean schema inside a vendor extension is left alone. I think that is right, but it is worth a line of comment since the `x-ext` exception sitting right there invites the question.
- `additionalProperties: false` is deliberately left as a literal boolean while every other boolean schema position is normalized. The docstring explains why ("`additionalProperties` already supports booleans") and the test pins it, so this is fine — just calling out that it is the one asymmetry a reader will trip over.

Happy to approve once the `deepClone` question is settled.

---

amritk (commented):
Preserving boolean schema semantics is worth doing, and I appreciate that the changeset is honest about the costs rather than burying them. But those costs are what stops me approving this as-is — the changeset itself says this "deliberately increases peak memory by the cloned document graph and inherits the recursive clone helper's depth limit; very large bundles or deeply nested input may require more memory or encounter a stack overflow." That is a self-declared regression risk on the server ingestion path for exactly the documents most likely to hit it.

**The clone on the server path.** `server.ts` now does `normalizeBooleanSchemas(deepClone(upgradedDocument))`, and `x-ext` holds every bundled external document, so peak memory at ingest roughly doubles for split-file documents. The code comment this replaces argued the opposite position — that cloning `x-ext` was the worst place to pay for it — and I would like to understand what changed. Is there a measurement on a large real bundle (the openapi-parser corpus, or one of the big public documents in the snapshot tests) showing what this actually costs? The comment names a copy-on-write normalizer as the alternative; if that is the eventual fix, is it worth doing now rather than shipping the regression and revisiting?

**The traversal is name-based and therefore fragile.** `normalize-boolean-schemas.ts` decides what is a schema position from five hardcoded key sets plus `isSchemaPath`. Anything the sets miss silently keeps its boolean and hits coercion unnormalized; anything they over-match normalizes user data into `{ __scalar_: '' }`. `opaqueValues` guards `example`/`default`/`enum`/`const`, but for instance a `default` nested under `x-ext` reached via the `x-ext` exception, or `properties` used as a literal object key inside example data, are the shapes I would want pinned by tests. 278 lines of test is a decent start — does it cover a boolean schema reached only through an external `$ref` in `x-ext`, and a document where `properties`/`items` appear as literal data keys?

**`__scalar_` as the marker.** `true` becomes `{ __scalar_: '' }` and `false` becomes `{ __scalar_: '', not: { __scalar_: '' } }`. Since this lands in the stored document, it can surface anywhere downstream that renders or serialises schemas — schema rendering, code examples, the `getRaw`/export path. Is the marker stripped on the way out, and is that covered? A stray `__scalar_` in an exported document or a rendered schema would be a visible bug.

**Ordering.** `$ref` targets are normalized in place via `setValueAtPath` while the referring node is being visited, which means the result depends on traversal order for a target referenced from more than one place. `visitedSchemas` makes it idempotent, but it is worth a test that two different references to the same boolean target both end up normalized.

---

cameronrohani (approved):

amritk (commented):
Not approving, though this is close — the bug is real and the validation is genuinely thorough. Two decisions need a human call first.

**1. `false` round-trips as `not: {}` in exported documents.** The description says "the public serialized shape retains `false` semantics as `not: {}`". That is semantically exact, but it means an author who writes

```yaml
additionalProperties: false
```

and exports their document back out gets `additionalProperties: { not: {} }`. Equivalent to a validator, surprising to a human, and noisy in a diff if anyone round-trips their description through Scalar and commits the result. Since `false` is by far the most common boolean schema in real documents (`additionalProperties: false` is nearly idiomatic), this will be noticed. Preserving the literal `false` on export, if the serializer can, seems worth the effort — otherwise it is worth calling out in the changeset so it does not arrive as a surprise.

**2. Internal `__scalar_` markers now live in the ingested data.** The PR is explicit that `getRaw` backing data deliberately retains them while "public proxy serialization, rendered schema fields, and original/saved JSON/YAML API-description exports omit it", and there is marker-free rendering coverage. That is the right set of exits to have covered. But it is an invariant with several independent escape routes, and the failure mode — `__scalar_` appearing in a user's exported API description — is exactly the kind of thing that ships. A single chokepoint that strips markers, rather than three call sites that each remember to, would make this much harder to regress.

**What I checked and liked:** the underlying bug is unambiguous — `false` means "accepts nothing", and coercing it to an empty object turns it into "accepts everything", which is the most inverted a schema can get. The `examples` map-context guard is a sharp catch: without explicit map context, an external resource named `examples` would make literal `properties` or `items` payload keys look like schema positions, which is the same class of bug #10278 is fixing in the upgrader. And the benchmark write-up is a model of honesty — reporting that post-GC heap deltas did not improve and explicitly declining to claim a retained-heap saving is the opposite of the usual perf-PR framing.

Given it touches workspace-store ingestion for both client and server, I would want @DemonHa on it rather than approving myself.

---

PR #10173, opened 2026-09-14
feat(workspace-store): support QUERY request method

hanspagel:
Adds support for the OpenAPI 3.2 `QUERY` method, including request bodies, workspace navigation, and server chunks. The API client can select `QUERY`, edit its body, and send it.

## Validation

Rebased onto `origin/main` at `ddab296d58`.

- 138 HTTP helper tests passed using the package Vitest configuration.
- Themes build and type check passed.
- Changed-file Biome and Prettier checks passed; changeset status passed.
- Scoped themes Knip passed using a temporary configuration excluding unrelated workspaces and disabling the Lefthook plugin (its `.git/hooks` lookup does not support this worktree).
- Visually inspected the light/dark theme colors. Themes has no unit test files.

QUERY has a dedicated color in every preset and the custom theme starter, including Laserwave P3 variants.


amritk (commented):
Solid, well-tested change — the `isHttpMethod` refactor in `server.ts` is a genuine improvement over the local `httpMethods` set, and the test coverage across helpers, client, server store and bundler is thorough. A few things worth resolving before merge.

### Worth addressing

1. **`REQUEST_METHODS.query` is byte-for-byte identical to `get`** (`text-blue` / `var(--scalar-color-blue)` / `bg-blue/10`). In the method dropdown and the sidebar, GET and QUERY become visually indistinguishable. QUERY being "GET with a body" makes blue defensible, but identical blue is not — `head`/`trace` sharing `text-c-2` is the existing precedent for "de-emphasised", not for "two prominent methods look the same". I would give it its own token.

2. **No `## Visual` section.** `AGENTS.md` requires screenshots for UI changes, and this touches the address bar method picker plus every method badge surface. Given point 1, a screenshot of the dropdown is exactly what a reviewer needs here.

3. **Why does `query` need to land in the v3.1 schemas?** `packages/workspace-store/src/schemas/v3.1/strict/path-item.ts` and `v3.1/openapi/index.ts` both gain a field whose own doc comment says "OpenAPI 3.2 QUERY operation". The v3.2 strict path item already has `query` on `main`, documents are upgraded through `openapi-upgrader`'s 3.1→3.2 pass, and `fbd9010` just typed the store against the 3.2 document. If something still reaches the 3.1 schema with a `query` key I would like a comment saying what, because otherwise this makes the 3.1 type non-conformant to 3.1 for no consumer.

4. **`externalizePathReferences` quietly became case-insensitive.** It previously did `httpMethods.has(type)` against the raw key; `isHttpMethod(type)` lowercases. So a path item key `GET` is now externalized where it was skipped before, and the chunk `$ref` is built from the original casing (`.../operations/~1planets/GET#`). Probably fine and arguably better, but it is an unrelated behaviour change that becomes load-bearing once #10186 starts preserving custom-method casing — worth a line in the changeset at least.

### Nits

5. `for-each-path-item-operation.test.ts`: the new test covers `setPathItemOperation` / `forEachPathItemOperation` / `getPathItemOperation` / `deletePathItemOperation` but sits inside `describe('getResolvedPathItem')`. `AGENTS.md` asks for the top-level `describe` to match the file, so this belongs in its own block.

6. `can-method-have-body.test.ts`: the new `preserves a QUERY body in the final request` test is placed *above* the `beforeEach`/`afterEach` that set up the Electron mock. Vitest applies hooks to the whole block regardless of position so it works, but reading it suggests the test runs without the mock.

7. `REQUEST_METHODS` carries a "Do not change the order" comment and `query` was inserted between `head` and `trace`. Adding an entry is not reordering, so this is fine — but next to `trace` is an odd home for a method users will actually pick. Between `patch` and `delete`, or right after `get`, would match how prominent it is.

Nothing here is hard to fix; (1) and (3) are the two I would want answered.

---

amritk (approved):
Small and consistent. `query` is added everywhere it needs to be — `HTTP_METHODS`, `REQUEST_METHODS` (so the picker has a label and colour rather than an undefined lookup), and `BODY_METHODS` — and the coverage follows it through `canMethodHaveBody`, `buildSafeBodyRequest`, `requestFactory`, `sendRequest`, `RequestBlock`, `AddressBar`, and both the client and server stores.

The `server.ts` change is better than it looks in the diff. On `main` the two call sites disagreed: `filterHttpMethodsOnly` tested `method.toLowerCase()` while storing `filteredMethods[method]` in the authored case, but `externalizePathReferences` tested `httpMethods.has(type)` case-sensitively. So an uppercase method key produced a chunk that nothing ever pointed at. Routing both through `isHttpMethod` (which lowercases before the lookup) makes them agree, and since `externalizePathReferences` builds the `$ref` from `${type}` in the authored case it still matches the chunk key. That matches the changeset note about recognising uppercase keys while retaining their case.

Marking `@scalar/helpers` and `@scalar/themes` `minor` is right — widening the `HttpMethod` union is an API change for consumers.

One cosmetic note, not blocking: `--scalar-color-query` is added to `presets/default.css` only. All fourteen presets define their own `--scalar-color-blue`/`green`/`orange`/`purple` independently and none import each other, so QUERY will be the one method colour that is not theme-tuned. It still *renders* everywhere, because `style.css` pulls `presets/default.css` into the `scalar-base` layer and the other presets never override this var — so this is "QUERY is teal in every theme", not "QUERY is invisible in thirteen themes". Worth a follow-up to give the darker presets their own value.

Good catch including `@source inline("text-query bg-query/10")` — that would have been an easy one to miss with Tailwind v4 and a dynamically selected class.

CI green. Approving.

---

PR #10174, opened 2026-09-14
docs(api-reference): document OpenAPI 3.2 nested tags

hanspagel:
The OpenAPI guide previously described tag grouping only through `x-tagGroups`. Document native OpenAPI 3.2 nesting with `parent`, including a three-level example where a parent tag has its own operation.

Explain `summary` titles, `x-displayName` precedence, and when native nesting replaces `x-tagGroups` or falls back to it. Keep the existing extension example for older API descriptions and include OpenAPI 3.2 in the introduction.


amritk (commented):
Docs-only, and I checked every behavioural claim against the code rather than taking the prose at face value:

- `x-displayName` taking precedence over `summary` — confirmed in `packages/workspace-store/src/navigation/helpers/traverse-tags.ts`: `tag['x-displayName'] ?? tag.summary ?? tag.name ?? 'Untitled Tag'`.
- "Unknown parents, self-references, and circular relationships do not create nesting" — confirmed by `getParentMap` (`tag.parent !== tag.name && names.has(tag.parent)`) and `isCyclic`.
- "When at least one `parent` relationship points to a declared tag without forming a cycle, Scalar uses native nesting … otherwise falls back to `x-tagGroups`" — matches the guard at `traverse-tags.ts:322-325` and is covered by the existing tests `falls back to x-tagGroups when parents only form cycles` and `breaks circular parent references instead of looping`.
- "a parent tag can have operations of its own alongside its child tags" — matches the re-parenting logic that appends children after the parent's own entries and then re-applies `x-scalar-order`.

The three-level Galaxy → Planets → Moons example is accurate and the `x-tagGroups` section correctly keeps the extension documented for pre-3.2 documents instead of deleting it.

---

amritk (approved):

PR #10175, opened 2026-09-14
feat(api-client): add localization support

### LINKED ISSUE #9655: Allow customizing the "Test Request" button label
Currently the "Test Request" button in the API client panel has a hardcoded label. For documentation targeting production APIs — where clicking the button sends a real request, not a sandboxed test — the word "Test" is misleading and can cause confusion.

**Feature request:** Allow the button label to be configured, either:
- Via `scalar.config.json` (e.g. a `localization.translations` key), or
- A dedicated `requestButtonLabel` config option

**Example use case:** A label like "Run Request" or "Send Request" better communicates that the action executes against a live production API, not a mock or test environment. This is especially important for APIs where a single call can trigger real-world side effects (order processing, inventory changes, etc.).

---

> *This issue was written with the assistance of an AI (Claude) on behalf of a Scalar Docs user.*

hanspagel:
Localize the API Client in the same eight languages as API Reference. Users can also override client strings through the API Reference configuration. Authentication labels, request fields, response controls, placeholders, accessible names, and empty states follow the selected locale.

## Translation scope and provenance

The component changes replace presentation copy with grouped `apiClient` keys. Protocol values, HTTP methods, code examples, user-provided API description content, and persisted example names retain their original values. The application shell and surfaces outside the converted client components still contain English copy. Future client labels should use the localization API. The still-open #10189 introduces a Query String title and helper paragraph; those labels need localization when its behavior and this PR are integrated.

The available history does not establish whether the original seven non-English dictionaries were generated or reviewed by native speakers. Newly added device authorization labels were generated during the rebase. Speaker review is unverified; key and placeholder parity does not establish linguistic accuracy.

## Localization plumbing

Client consumers inherit the reference locale and overrides while supplying client dictionary defaults. Cached contexts use a detached Vue effect scope so unmounting their first consumer does not stop updates for later consumers.

Prerequisite: #10311 fixes literal-dollar interpolation for existing API Reference strings and can land independently. This branch retains that fix and the client scope-name coverage until the prerequisite is merged, so scope names containing `$&` or `$'` remain intact.

## Bundle impact

The previous CI bundle comparison reported **88.05 KiB (+2.55%)** additional standalone JavaScript, mainly from the eight API Client translation dictionaries. This measurement predates the latest rebase and the lazy client changes now on main. All supported locales are available synchronously for locale selection and English fallback. See the [bundle comparison](https://app.relative-ci.com/projects/WjV7m24scsHglgIEsSBX/jobs/22765-uBklnuMufcseFql13IwK) for the size breakdown.

## Visual

These are actual CI report assets captured at `aa1e359e8c`, showing the English-to-Arabic authentication labels behind the search overlay. The empty search snapshot had retained the older untranslated background while the reference and search-results snapshots already included translations. The reference root uses RTL direction; URL fields retain LTR direction for URL syntax.

Before:

Arabic search before

After:

Arabic search after

Long-URL before:

Long URL before

Long-URL after:

Long URL after

The long URL remains contained in its address bar. The latest screenshot difference is confined to the script example buttons and displayed application version, with no address-bar or split-pane geometry change.

## Ticket

Fixes #9655


amritk (commented):
First, credit where it is due: I spot-checked `de.ts`, `zh-cn.ts` and `ar.ts` and these are **real translations**, not English placeholders — `'Ausgewählt:'`, `'已选择：'` (with the full-width colon), `'ملفات تعريف الارتباط'` for cookies. Eight locales × ~385 keys of actual translated content is the bulk of the 5,061 additions, and it is the part that would have been easiest to fake.

**There is a genuine bug fix buried in here that deserves its own changeset line:**

```diff
-(result, [param, paramValue]) => result.replaceAll(`{${param}}`, String(paramValue)),
+(result, [param, paramValue]) => result.replaceAll(`{${param}}`, () => String(paramValue)),
```

A string replacement argument gives `$&`, `$'`, `` $` `` and `$1` special meaning, so any interpolated *value* containing those sequences was being mangled. Switching to a function replacer disables that. This is in `@scalar/localization`, so it fixes existing **API Reference** interpolation too, not just the new client strings — it should not be invisible inside a "add localization support" changeset.

**The part I would most like a second opinion on is the derived-context cache in `useLocalization`:**

```ts
const cached = inheritedContexts.get(inherited)
if (cached) return cached
const context = createContext(() => ({ … }))
inheritedContexts.set(inherited, context)
```

The mechanism itself is sound — a component under an API Reference provider needs the *client's* dictionary defaults with the reference's overrides layered on, and computing that once per provider rather than per consumer is the right instinct.

What I want confirmed is lifecycle. `createContext` builds computeds, and `useLocalization` is called from `setup()`, so those computeds are registered with **the first consuming component's effect scope**. When that component unmounts its scope is stopped, but the WeakMap keeps handing the same context to every later consumer. Depending on Vue's behaviour for a stopped computed, later consumers may get a context that no longer tracks locale changes — which would show up as "switching language stops working after navigating away from whichever component mounted first", i.e. exactly the kind of bug that survives a test suite and shows up in a demo.

If `createContext` is already called outside a component scope (or wrapped in a detached `effectScope`), say so in a comment; otherwise wrapping the cached construction in `effectScope(true)` would make it explicitly owner-less. A test that mounts consumer A, unmounts it, mounts consumer B, then changes the locale and asserts B updates would pin it.

**Two smaller things:**

`rtlLocales: new Set(['ar', 'fa', 'he', 'ur'])` — only `ar` ships here, but it means the API Client now has an RTL locale. The API Reference presumably already handles `direction`, but the client has its own layout (address bar, split panes, the request/response table). Worth confirming the client actually flips rather than just setting an attribute; an RTL screenshot in the description would make that reviewable.

**On scope:** this touches ~30 Vue components mechanically, and that is where I would expect defects rather than in the dictionaries — a key pointed at the wrong group, a lost `{count}` parameter, an `aria-label` or `placeholder` left hard-coded while its visible sibling was converted. Those are invisible to type checking when the key exists but means something slightly different. `Fixes #9655` and the dictionaries look right; what would make the component half reviewable is a short note in the description on how the extraction was done and what was deliberately left hard-coded. #10189's follow-up mentions "hard-coded api-client labels remain consistent with that surface's existing localization policy" — after this PR that policy changes, so those two need to agree.

Not blocking on the smaller points — the derived-context lifecycle is the one I would settle before merging.

---

amritk (commented):
Bringing the client to parity with the reference on translations is clearly worth doing, and matching the same eight locales is the right target. Three things before approving.

**The `$&` / `$'` / `$1` fix should not be in here.** `create-localization.ts` gaining literal-dollar handling is a real bug fix for *existing* API Reference translations — anyone whose translated string contains `$&` is getting corrupted output today. That is a small, high-value patch that should land on its own, immediately, rather than waiting behind a 96-file feature. It also has its own test coverage in `create-localization.test.ts` already.

**Translation provenance.** ~2,700 lines of new locale strings across seven languages. Were these reviewed by speakers, or generated? Either is fine, but it should be stated somewhere, because the review signal is completely different — nobody on this PR can meaningfully check the Arabic or Simplified Chinese strings by reading the diff, and a wrong translation in an auth or destructive-action label is worse than an English fallback.

**The RTL snapshots moved.** `arabic-reference.png` and `arabic-search-results.png` both changed. That is expected given new translated strings, but Arabic is where layout regressions actually show up. Could you post the before/after in the description per the visual-artifact guidance? Same for the `operation-example-long-request-url.png` change, which I would not have expected from a localization PR.

**Scope note:** the ~1,900 lines of non-locale changes across 82 files are mostly mechanical string extraction, which is fine, but bundled with the localization plumbing, the `@scalar/localization` fix and the locale data it is hard to see the mechanical part separately from the parts that need judgement. Splitting the plumbing plus English from the seven translations would make both halves much easier to review.

---

cameronrohani (approved):

PR #10176, opened 2026-09-14
feat(mock-server): support streaming item schemas

hanspagel:
Generate finite mock streams from OpenAPI 3.2 `itemSchema` definitions for SSE, JSON Lines, NDJSON, and JSON Sequences. Previously, responses with only `itemSchema` had no schema to generate a body from. They now return three generated items and close the stream.

For example, `/events` returns `data: "hello"` followed by a blank line, three times:

```yaml
openapi: 3.2.1
info:
  title: Stream example
  version: '1.0'
paths:
  /events:
    get:
      responses:
        '200':
          description: Events
          content:
            text/event-stream:
              itemSchema:
                type: string
                const: hello
```

Custom handlers use the same framing and retain their HTTP status. The PR reuses the shared classifier and serializer from main. Its changeset targets only mock-server.


amritk (commented):
Nicely scoped and the test suite is genuinely good — the `
` framing for `json-seq`, the `+json-seq` structured suffix, the SSE field-injection guard (`event: 'hello\ndata: bad'` being dropped rather than smuggling a field), and `false` schemas producing no items are all the cases I would have gone looking for.

### The one I would fix before merge

**SSE events whose `data` is not a string are silently dropped.** `serializeEvent` does:

```ts
if (typeof value !== 'string' || (field === 'id' && value.includes('\0'))) {
  return []
}
```

So for the single most common real SSE shape —

```yaml
content:
  text/event-stream:
    itemSchema:
      type: object
      properties:
        event: { type: string }
        data:  { type: object, properties: { id: { type: string } } }
```

— `data` generates an object, fails the `typeof === 'string'` test, and the event is emitted with *no* `data:` line at all. And because `serializeEvent` returns `''` when no field survives, an `itemSchema` that is a bare `{ type: 'string' }` yields three empty chunks: a 200 with `Content-Type: text/event-stream` and an empty body. Both cases look like the mock server is broken rather than like a documentation problem.

I would JSON-encode a non-string `data` (that is what SSE producers actually do on the wire) and fall back to `data: <json>\n\n` when the item is not an object, rather than returning `''`.

### Worth addressing

1. **The `.some()` pre-check in `mock-handler-response.ts` is redundant and re-implements content negotiation.** `getStreamingResponse` already returns `undefined` for a media type without `itemSchema`, so the `contentTypes.some(type => 'itemSchema' in ...)` guard buys nothing — and inside it you rebuild the accept negotiation with a hand-rolled default (`application/json` if present, else `contentTypes[0]`) that `mockAnyResponse` already computes. Extracting the negotiation both routes need would drop a fair chunk of this.

2. **`itemSchema: true` is untested and probably wrong.** The code checks `itemSchema === false` but not `=== true`, so a `true` schema reaches `generateResponseExample(true as ExampleSchema)`. `true` is a legal JSON Schema meaning "anything", and `packages/openapi-types/src/helpers/is-boolean-json-schema.ts` exists precisely for this. Worth a test either way so the behaviour is pinned.

3. **The type signature fights the migration.** `OpenAPIV3_1.MediaTypeObject & { itemSchema?: OpenAPIV3_2.MediaTypeObject['itemSchema'] | boolean }` bolts 3.2 onto 3.1. With #10177 moving consumers to 3.2 schemas, this should just be `OpenAPIV3_2.MediaTypeObject` — and that would also remove several of the `as ExampleSchema` casts, which `AGENTS.md` asks us to avoid.

### Nits

4. `parseMimeType` handles the `+json-seq` structured suffix but `application/jsonl` / `application/x-ndjson` are matched as exact essence strings. Inconsistent, and `application/json-lines` (seen in the wild) falls through.

5. `tsconfig.json` gaining `"types": ["node"]` is unexplained — if `hono/streaming` needs it, a comment saves the next person the bisect.

6. No `X-Accel-Buffering: no` alongside `Cache-Control: no-cache`. Not needed for the in-process `server.request()` tests, but it is what makes SSE work behind a proxy.

### Cross-PR

#10178 adds `packages/workspace-store/src/helpers/serialize-stream-example.ts`, which reimplements this file's SSE/JSONL/json-seq serialisation almost line for line (with a cosmetic divergence: `/[\r\n]/.test(value)` here vs `lines.length === 1` there). `streaming-response.ts` already imports from `@scalar/workspace-store`, so whichever of the two lands second should delete its copy and import the other. Worth agreeing on the direction now rather than after both merge.

---

amritk (commented):
The SSE serializer is the strongest part of this — it is spec-correct in the places that usually go wrong:

- `data` split on `\r\n|\r|\n` with a `data:` line per segment, including the trailing empty one, so `'a\r\nb\n'` becomes three `data:` lines rather than a frame-breaking literal newline.
- `event` and `id` rejected outright if they contain any line ending, and `id` additionally rejected for `\0`, which is exactly what the WHATWG spec requires.
- `retry` accepted only as a non-negative integer.
- A frame with no valid fields emits nothing rather than an empty `\n\n`.

The "ignores invalid SSE fields without injecting new fields" test — `event: 'hello\ndata: bad'` producing only `data: safe` — is a genuine injection case and good to see pinned.

Also good: `writer.aborted` checked before each write, `X-Accel-Buffering: no` and `Cache-Control: no-cache` set, and `selectResponseExample` generalised to a `T extends Pick<…>` so handler media types work without a cast.

**The thing I most want changed is not in this PR alone: `getStreamFormat` is now the third copy of the same predicate.**

- Here: `packages/workspace-store/src/helpers/serialize-stream-example.ts` → `isStreamingMediaType`
- #10207: `packages/api-client/src/v2/blocks/response-block/helpers/response-stream.ts` → `getResponseStreamFormat` (same media types, plus `multipart`)
- #10102 needs the same thing and currently ships `contentType?.startsWith('text/event-stream')` in `toPostmanResponse`, which will silently miss every format this PR adds

Three PRs in flight, one concept, and they will drift the first time someone adds a media type to one of them. This PR's version is in the best place already — `@scalar/workspace-store/helpers` or `@scalar/helpers` — so the ask is mainly to say in the description that this is the canonical one and have the others import it. Your own follow-up note says "Mock and documentation stream serializers share one implementation across #10176/#10178", so that is already half the plan; it just needs to extend to the client-side copies.

**One thing I think may be a bug: the handler path does not set the status.**

`mockAnyResponse` does this:

```ts
c.status(statusCode)
return sendStreamingResponse(c, streamingResponse)
```

`mockHandlerResponse` does not:

```ts
const streamingResponse = getStreamingResponse(response?.content?.[contentType], contentType, {…})
if (streamingResponse) {
  return sendStreamingResponse(c, streamingResponse)
}
```

`statusCode` is in scope there — it is used on the line above to look the response up — so the asymmetry looks unintentional. The tests do not catch it because every `x-handler` streaming test asserts 200; the only non-200 streaming assertion (`Prefer: code=201`) goes through `mockAnyResponse`. Worth either adding `c.status(statusCode)` or a test with `x-handler` plus a 201 stream to show it is already handled elsewhere in that function.

**Two smaller notes:**

`"types": ["node"]` in `packages/mock-server/tsconfig.json` pulls the whole Node global surface into the package to satisfy one `hono/streaming` declaration. That also means an accidental `process.env` or `Buffer` in mock-server source now type-checks silently, which matters for a package that can run on non-Node runtimes. A triple-slash `/// <reference types="node" />` in a single `.d.ts`, scoped to where the streaming types are needed, keeps the rest of the package honest.

`itemSchema: true` generating `null` records is legal (a `true` schema accepts anything) but makes for a dull mock. `generateResponseExample` on an empty schema might produce something more representative — minor, and the current behaviour is at least documented by a test.

---

amritk (commented):
The mock-server side is the smallest and most contained of the streaming PRs in flight, and content negotiation plus finite generation from `itemSchema` is the right shape for a mock.

**This overlaps #10178 file-for-file.** Both PRs add `packages/helpers/src/http/is-streaming-media-type.ts` (18 lines) and `packages/workspace-store/src/helpers/serialize-stream-example.ts` (49 lines), apparently identical in both. Neither conflicts with `main`, so both are green, but the second to merge will conflict on two whole new files — and a hand-resolution of "two identical new files" is exactly where a subtly older copy of the SSE serializer survives without anyone noticing.

The same pattern shows up between #10178 and #10189, which both add `is-json-media-type.ts`. Landing the shared helpers as one small PR first and rebasing the feature PRs onto it would remove the whole class of problem and make each feature PR smaller. Worth doing across this stack rather than PR by PR.

**Also worth checking:** `@scalar/mock-server` and `@scalar/workspace-store` are both bumped `minor` here and `@scalar/workspace-store` is bumped `minor` in #10177 too. Confirm `pnpm changeset status` gives the version you expect once the stack lands, since the ordering affects which bump wins.

Happy to approve once the shared-helper duplication is resolved one way or another — the mock-server logic itself I have no objection to.

---

cameronrohani (approved):

amritk (commented):
Not approving, but this is the healthiest of the streaming cluster — clean merge state, 14 files, and a description that is honest about what was and was not verified.

**Two ordering dependencies worth making explicit before anyone merges:**

1. **The shared helper collides with #10178.** Both PRs ship their own copy of `@scalar/helpers/http/is-streaming-media-type` (`getStreamFormat` / `isStreamingMediaType`) plus the stream serializer. The description says either can land first, which is right, but whichever goes second will conflict on those files and needs its duplicate dropped rather than merged. #10207 and #10102 are also supposed to import the same helper, so this is a four-way convergence point — worth landing one of #10176/#10178 quickly and rebasing the rest onto it.

2. **The failing CORS test looks like it needs #10290.** The description notes "the remaining CORS test expects QUERY support from current Hono; the older installed Hono omits QUERY". #10290 is the PR that bumps Hono (4.13.7 → 4.13.8, and `@hono/node-server` 1 → 2). If that is the same gap, this test should pass on CI against the lockfile and the note is stale — but it is worth confirming rather than assuming, because "passes in CI, fails locally" and "depends on an unmerged PR" look identical from here.

**Things I liked in the write-up:** being explicit that an unconstrained `true` item schema deliberately generates `null` as a deterministic value, restricting line formats to `jsonl` / `x-ndjson` / `json-lines` rather than inferring arbitrary suffixes, disabling proxy buffering, and the regression proving a streaming handler keeps its already-set HTTP 201. Those are exactly the details that get quietly broken later.

The honest note that type checks ran against temporary aliases because pnpm 12.2.1 could not be installed is appreciated — CI covers it, so I am not treating it as a gap here, just flagging that I could not independently confirm it.

I would rather @DemonHa sign off on the streaming semantics than approve this myself.

---

PR #10177, opened 2026-09-14
fix(workspace-store): complete OpenAPI 3.2 schema migration

hanspagel:
Complete the remaining OpenAPI 3.2 schema imports and preserve 3.2 fields in the loose workspace schema. For example, this API description now retains the server name, response summary, and streamed item schema:

```yaml
openapi: 3.2.0
info:
  title: User stream
  version: 1.0.0
servers:
  - url: https://example.com
    name: Production
paths:
  /users:
    get:
      responses:
        '200':
          summary: User stream
          content:
            application/jsonl:
              itemSchema:
                type: object
```

The browser and server stores continue to call `upgrade(document, '3.1')`. This PR does not change that target to 3.2 or enter the strict XML migration discussed in #10211. Existing 3.1 documents with unnamed inline XML bodies remain loadable and retain XML attribute metadata. Adopting the 3.2 upgrader later still requires handling its migration errors.

The editor offers 3.2 validation and completion, accepts existing 3.1 documents, and no longer warns about valid leading-slash paths. Its intentionally permissive authoring behavior is documented in the user-facing OpenAPI docs and app README: authors must explicitly migrate and declare 3.2 before relying on 3.2-only fields in other tools.

Remove the unused 3.1 loose-schema generator. The changeset names its removed imports and 3.2 replacements; the schema export wildcard and 3.1 strict-schema exports remain available. Existing schema-injection and dynamic-reference coverage remains under `v3.2`. The generated namespace is `OpenAPIV3_2`, with no remaining internal reference to the old generated namespace.

## Visual

Same 3.1 document before and after. JSON and YAML show no problems after the change.

| Before | After |
| --- | --- |
| Before | After |

## Ticket

See #10211


amritk (commented):
I spot-checked the loose-schema additions against the strict 3.2 schemas already on `main` and they line up exactly — `xml.nodeType`, `discriminator.defaultMapping`, `tag.summary`/`parent`/`kind`, `server.name`, `securityScheme.deprecated`, `example.dataValue`/`serializedValue`, `components.mediaTypes`, `mediaType.itemSchema`, `pathItem.query`/`additionalOperations`, nested `encoding`, `parameter.in: querystring`, `$self`, and `response.summary` with `description` becoming optional (which matches `v3.2/strict/response.ts`'s "Optional as of OpenAPI 3.2"). Bringing the loose schema to parity with the strict one is the right call and this looks faithful.

### Worth addressing

1. **`v3.1/openapi/index.ts` becomes dead code and the PR keeps it alive with a knip ignore.** After moving `generate-openapi-types.ts` to the 3.2 schema, the only remaining references to `@/schemas/v3.1/openapi` are its own two test files. That is ~46KB of schema builder with no production consumer, and the knip entry swap (`v3.2/openapi/reference.ts` → `v3.1/openapi/reference.ts`) papers over it. I would delete `v3.1/openapi/` plus its tests in this PR — as it stands it will silently drift from the 3.2 copy and the next person cannot tell whether it matters.

2. **The `$defs.paths` override in `openapi-editor-schema.ts` discards everything except `patternProperties`.**

   ```ts
   paths: {
     type: 'object',
     patternProperties: { ...openApiSchema.$defs.paths.patternProperties, '^x-': true },
     additionalProperties: false,
   }
   ```

   Any other keyword the upstream `$defs.paths` carries — a `$ref` to the specification-extensions definition, `unevaluatedProperties`, `$comment` — is dropped, and `additionalProperties: false` then becomes the sole authority. The two new tests cover `^x-` and the leading-slash rule, so the stated intent is verified, but nothing catches the loss. Spreading the whole definition and overriding only `patternProperties` would be safer:

   ```ts
   paths: { ...openApiSchema.$defs.paths, patternProperties: { ... }, additionalProperties: false }
   ```

3. **`openapi-3.2-schema.json` has no provenance.** A 437-line-diff vendored upstream schema with no source URL or regeneration script means nobody can tell later whether it is current or hand-edited. A one-line comment in `openapi-editor-schema.ts` pointing at the release it came from would fix this.

4. **The `openapi` pattern loosening deserves a note.** `^3\\.(1|2)\\.\\d+(-.+)?$` means a document declaring `3.1.0` now gets 3.2-only completion and no error for 3.2-only fields. That is the pragmatic editor choice, but the changeset says "continuing to accept existing 3.1 documents" without mentioning that 3.1 documents also lose the "this field does not exist in your version" signal.

### Nits

5. `generate-openapi-types.ts` renames the namespace `OpenAPIV3_1` → `OpenAPIV3_2`. The outputs (`scripts/openapi-types.ts`, `scripts/openapi-proxy-types.ts`) are not checked in so there is no drift, but anything downstream referencing the old namespace name breaks — worth confirming nothing does.

6. The changeset marks `@scalar/workspace-store` as `patch`, but the loose schema newly preserves a large set of 3.2 fields that it previously dropped. That is additive behaviour a consumer can depend on; `minor` reads more accurately.

Good, careful PR overall. (1) is the one I would not want to defer.

---

amritk (commented):
Net −964 lines and the removal of a 1,211-line parallel schema generator is a good trade. Updating `generate-openapi-types.ts` and dropping the `knip.jsonc` exemption in the same PR is the right hygiene — a deletion like this is only finished when the tooling stops pretending the code is still needed.

Three things I would like settled first.

**1. This removes a public export path, and the changeset does not say so.**

`packages/workspace-store/package.json` exports `./schemas/*` as a wildcard, so `@scalar/workspace-store/schemas/v3.1/openapi` resolves today for any external consumer. Deleting `src/schemas/v3.1/openapi/index.ts` and `reference.ts` makes that import fail at build time.

Your PR body acknowledges it — "external consumers of locally generated files must adopt the new namespace" — and that is the right call for the repo. But the PR body is not what lands in the changelog. The changeset is what a consumer reads when their build breaks, and it needs to name the removed path and the replacement explicitly. `minor` is also generous for removing an export; I realise the repo policy rules out `major`, which makes the changeset text carry all the weight.

Relatedly, `package.json` gains 10 lines of what I assume are explicit export entries — worth confirming those cover everything the wildcard previously exposed, so this does not quietly narrow other paths at the same time.

**2. Two deleted test files, one with a name I would not delete without checking.**

```
rem packages/workspace-store/src/schemas/v3.1/openapi/schema-object-injection.test.ts  -65
rem packages/workspace-store/src/schemas/v3.1/openapi/dynamic-ref.test.ts             -75
add packages/workspace-store/src/schemas/v3.2/openapi/index.test.ts                   +92
```

140 lines of coverage out, 92 in. "Schema object injection" sounds like it was pinning behaviour that matters beyond the 3.1 generator, and `$dynamicRef` handling is actively being changed in #9625 and #9886. If the new `index.test.ts` covers the same cases against the 3.2 generator, saying so in the description would close it; if either was testing something the 3.2 file does not exercise, it should be ported rather than dropped.

**3. The editor now validates against 3.2, which is a user-visible behaviour change.**

`openapi-3.2-schema.json` is a rename with +331/-106, `configure-language-support.ts` is rewired, and there is a new dependency (`projects/scalar-app/package.json` +1, lockfile +3). That means the desktop app's OpenAPI editor validates **existing 3.1 documents** against a 3.2 schema. The new `openapi-editor-schema.ts` plus its test presumably handle the "permissive editor version behavior" your description mentions, but this is the part of the PR a user actually sees, and it is the one part with no screenshot. A before/after of a 3.1 document in the editor — showing no new spurious diagnostics — would make it reviewable.

Everything else is mechanical import path updates and reads fine. Once the changeset names the removed export and the deleted tests are accounted for, I am happy to approve.

---

amritk (commented):
Most of this is a mechanical `v3.1` → `v3.2` import sweep across consumers, stories and tests, and deleting the unused 3.1 loose-schema generator to stop schema drift is clearly right — a second generator nobody consumes is a liability.

One thing I do want to discuss before approving, because it is a deliberate product decision rather than a mechanical change.

**The editor offers 3.2 fields to documents declaring 3.1.** The changeset says this plainly: "The editor intentionally offers the 3.2 field set to documents declaring 3.1 as well; it does not flag 3.2-only fields solely because the declared version is 3.1."

That is in direct tension with the guidance being added in #10204, which says to "respect the version declared by the API description" and not to "apply newer-version behavior to older documents unless that version's specification supports it". The practical effect is that the editor will autocomplete and accept `itemSchema`, `additionalOperations`, `style: cookie` and the rest into a document declaring `openapi: 3.1.0`, producing a document that our tooling accepts and other OpenAPI tools reject — with no warning at the point of authoring.

I can see the argument for it (one schema to maintain, and the store coerces to 3.2 anyway), but silently permissive is the worst of the options. Could the editor at least surface these as informational rather than accepting them identically — or prompt to bump the declared version when a 3.2-only field is used? If the permissive behaviour stands, it deserves a note in the user-facing docs, not only in the changeset, since the resulting documents are not portable.

**Smaller:** bumping `@scalar/workspace-store` to `minor` is right given the type surface moves from `OpenAPIV3_1` to `OpenAPIV3_2` in generated namespaces — worth confirming no downstream integration pins the old namespace name.

---

cameronrohani (approved):

amritk (commented):
Not approving — there is a cross-PR dependency here that I think blocks this one, independently of its size.

**#10211 says this PR cannot adopt 3.2 until it handles that PR's aggregate error.** From #10211's description:

> upgrading to 3.2 remains strict. Inline XML body schemas without an inferable name still require an explicit `xml.name`; callers adopting 3.2 (including the workspace-store migration in #10177) must handle the aggregate error before changing their target version.

This PR is exactly that migration, and #10211 is currently `blocked`. So either this branch already contains the aggregate-error handling that #10211 expects (in which case it would be good to say so here, since the description does not mention it), or merging this first produces the failure mode #10211 is warning about: documents with inline XML body schemas lacking `xml.name` throwing during migration. That is a document-fails-to-load outcome, not a cosmetic one.

That same strictness concern is why I did not approve #10276 either — the 3.2 upgrade path is accumulating several places where a previously-loadable document now raises. Those are individually defensible and collectively worth looking at as one decision.

**On the change itself:** 686 additions against 1,583 deletions across 36 files is a healthy ratio for a migration, and "the editor still accepts 3.1 documents and offers 3.2 fields" is the right compatibility posture. The before/after screenshots showing the same 3.1 document producing no problems in JSON and YAML are the most convincing evidence in the PR — false warnings on valid paths are precisely the regression this class of change causes.

Worth confirming before merge: the description says full Knip is blocked by a missing Docusaurus build. With 1,583 deleted lines, unused-export detection is the check most likely to catch leftovers, so it would be good to get a clean run rather than relying on the scoped one.

---

PR #10178, opened 2026-09-14
feat(api-reference): support streaming item schemas

hanspagel:
API Reference and the API Client now display OpenAPI 3.2 `itemSchema` models and generate framed streaming examples. Previously an API description containing only `itemSchema` could leave the schema view and generated request body empty. A complete `schema` remains the primary model when both are present, with a separate Stream item view.

```yaml
openapi: 3.2.0
info:
  title: Event stream
  version: 1.0.0
paths:
  /events:
    post:
      requestBody:
        content:
          application/jsonl:
            itemSchema:
              type: object
              properties:
                message:
                  type: string
                  const: Hello
      responses:
        '200':
          description: Event records
          content:
            application/jsonl:
              itemSchema:
                type: object
                properties:
                  message:
                    type: string
                    const: Hello
```

This now shows the `message` stream item and generates `{"message":"Hello"}` followed by a newline in the request body, response example, and cURL `--data-binary` sample. JSON Sequence and SSE examples receive their corresponding record framing. Authored wire text is preserved; structured examples are framed. Generated non-streaming `0`, `false`, and empty-string request examples are also preserved. SSE examples with no valid fields are omitted with a warning reporting the omitted count.

Streaming classification extends the existing `@scalar/helpers/http/is-streaming-content-type` helper, including its existing multipart detection. Multipart framing is not added. The workspace-store serializer is shared with #10176; that PR should rebase onto this one and import the existing helper and serializer. Response example formatting keeps main's options-object signature, JSON default, composition selection, and `serializedValue`/`dataValue` precedence. The response variant picker also supports streaming item schemas. Stream item localization keys are present; non-English translations remain pending.

## Visual

Streaming schemas and examples in API Reference


amritk (commented):
The `schema ?? itemSchema` fallback threading is consistent and the "Stream item" panel shown alongside a complete `schema` (rather than instead of it) is the right UX call. A few things to fix.

### Worth addressing

1. **New user-facing strings bypass the localization engine.** `01c20d8` just landed the shared translation engine, and `ExampleResponses.vue` in this very diff uses `translate('response.showSchema')` — but the three new `Stream item` labels are hard-coded English:
   - `api-reference/src/features/Operation/components/ParameterListItem.vue` (the `<p>` and the `name="Stream item"` prop)
   - `api-reference/src/features/Operation/components/RequestBody.vue` (same two)
   - `api-reference/src/features/example-responses/ExampleResponses.vue` (the `<p>`)

   These are all in `api-reference`, which is localized, so they should go through `translate()`. (The `api-client` side of this PR has no `useLocalization` usage at all yet, so `RequestBody.vue` there is fine as-is.)

2. **`serializeStreamExample([], contentType, false) !== undefined` is being used as a "is this a streaming media type?" predicate** in `get-example-content.ts`:

   ```ts
   return typeof value === 'string' && serializeStreamExample([], contentType, false) !== undefined
     ? value
     : prettyPrintJson(value)
   ```

   That calls a serializer with a dummy empty array purely to observe whether it returns `''` or `undefined`. It works today but it is load-bearing on an implementation detail, and it is not obvious to a reader what is being tested. Export an `isStreamingMediaType(contentType)` from the same module and use it in both places.

3. **`getExampleFromBody`'s guard change is an unrelated behaviour change.** `if (!schemaExample)` → `if (schemaExample === undefined || schemaExample === null)` means a schema that generates `0`, `false`, or `''` now produces a request body example where it previously produced `null`. That is arguably the correct fix, but it affects every non-streaming body too and the changeset does not mention it. Either call it out or split it.

4. **The curl `--data-binary` branch does not do what its comment claims, and is incomplete.** The comment says "Preserve line framing", but for an inline string argument curl does not strip newlines from `--data` either — the stripping only happens with `--data @file`. So `--data-binary` is the more *explicit* choice rather than a fix. More importantly `application/json-seq` and `text/event-stream` are not in the list: `json-seq`'s subtype is not matched by `isJsonContentType` (I checked — `subtype === 'json' || subtype.endsWith('+json')`, and `json-seq` matches neither), so it falls through to the generic `else`. If the framing rationale holds, all four streaming types belong in this branch.

### Nits

5. `process-body.ts` drops the `getResolvedRef()` wrapper: `const contentSchema = mediaType?.schema ?? mediaType?.itemSchema`. The subsequent `typeof contentSchema !== 'undefined'` now passes for an unresolved `{ $ref }` node that previously resolved to `undefined` and skipped the block. `getResolvedRefDeep` probably absorbs it, but it is a silent edge-case change.

6. `ExampleSchema :id="currentResponseContent.schema ? \`${id}-item\` : id"` — good catch on the collision, but the conditional means the item schema's anchor id changes depending on whether a sibling `schema` exists. If anything deep-links to these, `${id}-item` unconditionally would be more stable.

### Cross-PR

`serialize-stream-example.ts` duplicates `packages/mock-server/src/utils/streaming-response.ts` from #10176 nearly line for line — same JSONL/json-seq/SSE framing, same `['event','id','retry','data']` field walk, same `retry` integer guard, differing only in how multi-line non-`data` fields are rejected (`lines.length === 1` here vs `/[\r\n]/.test(value)` there). `streaming-response.ts` already imports `@scalar/workspace-store`, so it should import this helper instead of keeping its own copy. Also note #10176's version has the non-string-`data` bug I flagged there; if these converge, fix it once.

Separately: this PR, #10176 and #10189 each introduce their own inline JSON-media-type regex (`/(?:\/|\+)json(?:;|$)/i` and a `$`-anchored variant), and `snippetz` already carries three private copies of `isJsonContentType` (curl, wget, restsharp). #10192 adds `packages/helpers/src/http/is-xml-media-type.ts` — a matching `isJsonMediaType` next to it would let all of these share one definition, which is what `AGENTS.md` asks for.

---

amritk (commented):
Two things here are straightforwardly good. Extracting `isJsonMediaType` into `@scalar/helpers/http` and deleting the three byte-identical `isJsonContentType` copies from curl, wget and RestSharp is exactly the right cleanup. And this is a real bug fix hiding in the diff:

```diff
-  if (!schemaExample) {
+  if (schemaExample === undefined || schemaExample === null) {
```

A generated example of `0`, `false` or `''` previously fell through to "no body". Worth its own line in the changeset — it changes request bodies for existing documents, independent of streaming.

I also confirmed that `packages/workspace-store/src/helpers/serialize-stream-example.ts` is **byte-identical** between this branch and #10176, so the add/add conflict between them resolves by taking either side. Good that it was kept in sync deliberately; worth saying so in both descriptions so whoever merges second does not try to reconcile it by hand.

**The thing I would fix: this PR deduplicates one media-type predicate and then inlines a fourth copy of another.**

```ts
} else if (
  ['application/jsonl','application/x-ndjson','application/json-lines','application/json-seq','text/event-stream']
    .includes(parseMimeType(normalizedRequest.postData.mimeType ?? '').essence) ||
  parseMimeType(normalizedRequest.postData.mimeType ?? '').subtype.endsWith('+json-seq')
) {
```

That list is precisely what `isStreamingMediaType` — added by this same PR, in `serialize-stream-example.ts` — already encodes, and `parseMimeType` is called twice on the same string to express it. Counting across the open PRs, the streaming media-type set now exists in four places:

- `serialize-stream-example.ts` → `getStreamFormat` (this PR and #10176)
- `response-stream.ts` → `getResponseStreamFormat` (#10207, plus `multipart`)
- this inline list in `curl.ts`
- and `#10102` needs it in `toPostmanResponse`, where it currently ships as `startsWith('text/event-stream')` and will silently miss every format added here

I appreciate `@scalar/snippetz` cannot reasonably depend on `@scalar/workspace-store`, which is presumably why it was inlined. That argues for **moving `isStreamingMediaType` next to `isJsonMediaType` in `@scalar/helpers/http`** — the same package this PR just established as the home for exactly this kind of predicate. Then curl, the mock server, the response viewer and the Postman adapter all import one definition, and this PR's own consolidation theme carries through.

**A concrete merge hazard with #10208 and #10203.** All three PRs add a third parameter to `getExampleContent`, and two of them use the *same name with a different default*:

```ts
// this PR
getExampleContent(response, example, contentType = '')
// #10208
// #10203
```

Git will not flag `''` versus `'application/json'` as anything more than a one-line conflict, but the defaults drive different branches — an empty default means `isStreamingMediaType('')` is false and the legacy `prettyPrintJson` path runs, while `'application/json'` means #10208's `getJsonExampleText` fires. Picking the wrong one on merge is a silent behaviour change in response rendering. These three want a deliberate order and an options-object signature rather than three stacked positionals; I have left the same note on the other two.

**One smaller note:** the eight locale files each gain one entry, but I assume `ar.ts`, `zh-cn.ts` and the rest received the English string. That is fine as a structural placeholder, but "All API Reference Stream item labels are localized" in the follow-up notes slightly oversells it — worth saying "keys added, translations pending" so they get picked up in the next translation pass rather than looking done.

---

amritk (commented):
The SSE serializer is the best part of this — folding multi-line `data` into repeated `data:` lines, validating `retry` as a non-negative integer, and dropping an `id` containing NUL are all real wire-format requirements that are easy to skip. Collapsing the three copies of `isJsonContentType` in curl/wget/restsharp into a shared `isJsonMediaType` is a good cleanup to ride along.

Two things I want to understand before approving, because both change output for documents that have nothing to do with streaming.

**The falsy guard in `getExampleFromBody`.** `if (!schemaExample)` became `if (schemaExample === undefined || schemaExample === null)`. That is the correct fix — a legitimately generated `''`, `0` or `false` body should not be treated as "no example" — but it applies to *every* request body, not only streaming ones, and it is not mentioned in the changeset. What does it do to the snippet snapshots for schemas that generate a falsy example today?

**Explicit examples get reframed.** The comment on `serializeStreamExample` says "explicit wire-format examples bypass this helper", but in `getExampleFromBody` only *string* examples bypass it — a non-string author-supplied example for a streaming media type is re-serialized into JSON-lines/SSE framing. I think that is what you want (an author writing a structured example does not want to hand-frame it), but the comment reads as if it does not happen. Worth aligning the two, and pinning the behaviour with a test for an authored array example on `application/jsonl`.

**Minor:** in the SSE branch, an item whose fields all get filtered out contributes `''`, so an example consisting only of such items produces an empty body with no indication anything was dropped. Probably unreachable in practice, but a silent empty body is an unhelpful failure mode.

**Also:** curl now emits `--data-binary` instead of `--data` for streaming media types. That is the right call for framed content, and I only mention it so it lands in the changeset as a user-visible snippet change.

---

cameronrohani (approved):

amritk (commented):
Not approving — `mergeable_state` is `dirty`, and there is a converging-refactor problem across four PRs that should be settled before any of them lands.

**The shared helper is claimed by multiple PRs.** This PR and #10176 both introduce `@scalar/helpers/http/is-streaming-media-type` (`getStreamFormat` / `isStreamingMediaType`) and the stream serializer, and the description says #10207 and #10102 should import it too, while #10189 uses "the identical helper" for JSON detection. So four or five open PRs are each carrying a copy of the same helper. Whichever lands first defines it; every other one then needs its duplicate deleted during conflict resolution rather than merged. That is a resolution people get wrong under time pressure — two slightly divergent copies of a media-type classifier is a genuinely bad outcome. Landing one of #10176/#10178 first and rebasing the rest onto it would remove the hazard entirely.

**There is also a signature the other PRs disagree about.** This branch changes the third argument of `getExampleContent` to an options object and says to "preserve this branch's default empty content type until explicitly integrating #10208's JSON example handling; #10203 can add composition selection as another option." Meanwhile #10208 says #10203 must rebase on *it* and use `getExampleContent(response, example, { contentType, compositionSelection })`. Three PRs are each describing a different intermediate state of the same function, in prose, on three separate pages. That needs to become one agreed signature — ideally by stacking the branches — before merges start.

**Also noted:** the description says workspace-store and API Reference type checks "remain blocked by missing or stale shared package declarations". CI should cover that, but combined with the conflict state it is another reason to rebase first and re-verify.

On substance the work reads well — explicit streaming detection rather than string sniffing, stable item anchors, the resolver guard retained for unresolved references, and cURL selecting binary mode for all stream formats (with the incorrect comment about inline `--data` removed) are all the right calls. The localization keys for stream item labels landing without translations is worth tracking so they do not ship as raw keys.

---

PR #10179, opened 2026-09-14
feat(api-client): support OAuth2 metadata URLs

hanspagel:
Merge this PR before #10191; then rebase and retarget #10191 onto main.

Add OAuth2 metadata discovery to the API client and reference. Serve metadata from the mock server and keep explicit OAuth settings.

Screenshot 2026-09-14 at 16 10 47

## Review follow-up

- Loopback HTTP discovery/endpoints are allowed; metadata coercion matches OIDC. Empty flow patches are omitted and declared scopes remain authoritative, including empty scopes.
- RFC 8414 consumption uses AuthorizationServerMetadata aliases for the shared field subset; this does not apply OIDC issuer URL semantics.
- Staleness is checked by scheme name and metadata URL. Replacing a reactive proxy for the same server no longer discards a response; switching server URLs still does. Both cases have regression coverage.
- Exact metadata paths preserve meaningful trailing slashes; absolute token URLs are normalized to route paths and documented in the changeset.
- Validation: 19 metadata/component tests, api-client type check, changed-file lint/format, scoped Knip, and changeset status. Merge this before #10191.

## Latest review follow-up

- Keep `oauth2MetadataUrl` in OpenAPI 3.2 schemas only and describe the HTTP loopback exception accurately.
- Warn when a metadata route collides with a declared API path; name the trailing-slash option at call sites.
- Validation: 109 focused tests pass, mock-server type check, changed-file Biome/Prettier, scoped Knip, and changeset status pass. Workspace-store validation uses local source aliases for available dependencies.


amritk (commented):
The staleness guard in `fetchConfiguration` (capturing `name`/`scheme`/`url` and re-checking after the await) is a nice touch, and the RFC 8414 default for `grant_types_supported` — `['authorization_code', 'implicit']` when absent — is correct. Two things need to change though.

### Blocking-ish

1. **The HTTPS check makes this untestable against the mock server this same PR adds.** `fetchOAuth2Metadata` rejects anything but `https:`:

   ```ts
   if (metadataUrl.protocol !== 'https:') {
     return [new Error('OAuth2 metadata URL must use HTTPS'), null]
   }
   ```

   The mock server serves its new metadata route over `http://localhost:…`, so "Fetch Configuration" against your own mock always errors out. Two further points: the existing sibling `fetchOpenIDConnectDiscovery` enforces no protocol check at all, so the two discovery paths now disagree; and standard OAuth tooling exempts loopback precisely because of local development. I would exempt `localhost` / `127.0.0.1` / `[::1]`, or drop the check and match the OIDC helper.

2. **`Value.Check` is stricter than the sibling's `coerceValue`, for no stated reason.** `fetchOpenIDConnectDiscovery` does `coerceValue(OpenIDConnectDiscoverySchema, await response.json())`; the new helper does `Value.Check(OpenIDConnectDiscoverySchema, data)` and bails on failure. A real authorization server that sends `scopes_supported` as a space-delimited string instead of an array — which happens — is accepted by the OIDC path and rejected outright by the new one, with the unhelpful message "missing or invalid endpoints". Use `coerceValue` for consistency.

### Worth addressing

3. **`oauth2MetadataToFlows` returns empty objects in the declared-flows branch.** When a declared flow already has both URLs, its entry maps to `{}`, so the payload is e.g. `{ implicit: {}, authorizationCode: {} }`. I checked `updateSecurityScheme` — it calls `mergeObjects(target, payload)`, so this is harmless rather than destructive. But it also means `scopes_supported` is never applied to declared flows even when their `scopes` are empty, which the comment ("Supplements declared flows") does not lead you to expect. Either filter the empties out or say why scopes are deliberately excluded.

4. **`tokenUrls.add(tokenRoute)` → `tokenUrls.add(getPathFromUrl(tokenRoute))` is an unrelated bug fix.** Previously an absolute `tokenUrl` (`https://auth.example.com/oauth/token`) was registered verbatim as a Hono route path. Good catch, but it changes mock-server routing for documents that have nothing to do with metadata discovery, and the changeset does not mention it. It deserves its own line.

5. **Why does `getPathFromUrl` need `preserveTrailingSlash`?** `/.well-known/oauth-authorization-server` has no trailing slash, so I cannot see what this flag is for. If it is defensive, a boolean parameter added to a shared helper for one caller's hypothetical is worth avoiding; if it is load-bearing, a comment would help.

### Nits

6. `OpenIDConnectDiscoverySchema` is now doing double duty for RFC 8414 authorization-server metadata, which is a different spec that merely overlaps. Renaming it (`AuthorizationServerMetadataSchema`) or re-exporting under an alias would stop the next reader assuming OIDC semantics apply.

7. `getOAuth2Metadata` is typed against `OpenAPIV3_1.OAuth2SecurityScheme['flows']`, so it cannot see `deviceAuthorization` — which #10191 adds. Given #10177 is moving consumers to 3.2 types, this should start there.

8. `scheme !== currentScheme` compares reactive-proxy identity. If an unrelated store mutation replaces the proxy mid-fetch, a successful response is silently discarded with no feedback. Comparing `name` plus `oauth2MetadataUrl` is probably sufficient.

### Cross-PR — please read before merging either

**#10191 is a strict superset of this PR.** I diffed the overlapping files on both branches: `OAuth2Metadata.vue` is byte-identical, and `fetch-oauth2-metadata.ts` differs by exactly one line (#10191 also validates `device_authorization_endpoint`). Both branch from the same `main`. So merging them independently guarantees a conflict, and everything I have flagged above needs fixing in whichever lands first or it ships twice. Either merge this first and rebase #10191 onto it, or close this and let #10191 carry the whole feature.

---

amritk (commented):
The security shape of this is right, and the part that matters most is the part most implementations skip:

```ts
for (const endpoint of [data.authorization_endpoint, data.token_endpoint]) {
  if (endpoint && !isAllowedMetadataUrl(new URL(endpoint))) {
    return [new Error('OAuth2 metadata endpoints must use HTTPS or HTTP on loopback'), null]
  }
}
```

Validating the **discovered** endpoints, not just the metadata URL, is what stops a metadata document from redirecting a token request to plain HTTP. The `{ token_endpoint: 'http://example.com/token' }` case is in the test matrix, which is the right thing to pin.

I also checked the proxy path, since `redirectToProxy(proxyUrl, metadataUrl.href)` on a loopback URL would otherwise be asking the proxy host to fetch its own localhost. `shouldUseProxy` already returns false for local URLs, so loopback metadata bypasses the proxy. No issue — noting it because the comment in `fetch-oauth2-metadata.ts` explains why `isLocalUrl` was *not* reused for the allowlist, and the interaction between those two different notions of "local" is worth someone confirming deliberately rather than by luck.

`oauth2MetadataToFlows` is careful in the right places: explicit endpoints are never overwritten, an explicitly empty scope set stays authoritative, discovery only invents grants when the Flows Object is empty, and the RFC 8414 default of `['authorization_code', 'implicit']` is applied when `grant_types_supported` is absent. The staleness check comparing both scheme name and metadata URL — with a test for "reactive proxy replaced but same URL" versus "URL actually changed" — is a genuinely subtle case to have covered.

**Four things before I approve:**

**1. The documented contract contradicts the implementation.** This comment appears in four places (`v3.1/openapi/index.ts`, `v3.1/strict/security-scheme.ts`, `v3.2/openapi/index.ts`, `v3.2/strict/security-scheme.ts`):

> URL to the OAuth2 authorization server metadata (RFC8414). **TLS is required.**

But `isAllowedMetadataUrl` deliberately permits `http:` on `localhost`, `127.0.0.1` and `[::1]`, and the mock server depends on that. On a security-relevant field, a doc comment that overstates the guarantee is worse than none — someone will cite it. Should read something like "HTTPS, or HTTP on loopback for local development".

**2. `oauth2MetadataUrl` is added to the 3.1 schemas as well as 3.2.** It is a 3.2 field, and this repo does not have a settled policy: #10173's follow-up says explicitly "QUERY was removed from the 3.1 schemas", the opposite decision for the same class of change landing the same week. Worth picking one and applying it to both, otherwise `v3.1` accumulates 3.2 fields case by case.

**3. The merge-order banner on this PR is wrong.** The body opens with "⚠️ Merge #10191 first", but its own Review follow-up closes with "Merge this before #10191", and #10191 says "Stacked on #10179. Merge the OAuth metadata PR first". The branch topology agrees with the latter — #10191's base *is* this branch. The banner should go before someone acts on it. (I have left the same note on #10191.)

**4. The mock server registers a route from a document-controlled path.**

```ts
app.get(getPathFromUrl(scheme.oauth2MetadataUrl, true), (c) => …)
```

Nothing checks that path against the document's own `paths`. A document declaring `oauth2MetadataUrl: '/pets'` silently shadows or is shadowed by the real mock route depending on registration order. Low severity for a mock server, but a one-line warning log on collision would save a confusing debugging session.

**One smaller note:** `getPathFromUrl(url, preserveTrailingSlash = false)` is a boolean trap — at the four call sites in `set-up-authentication-routes.ts` you cannot tell what `true` means without opening the function. An options object or a separate `getMetadataPathFromUrl` would read better, especially since the trailing-slash distinction is load-bearing here (your test covers both `/custom/metadata` and `/custom/metadata/`).

Also worth calling out as a genuine drive-by fix: normalising `tokenUrls.add(getPathFromUrl(tokenRoute))` in four places, where previously absolute token URLs were added raw and could never match a route. Worth a line in the changeset, since it changes behaviour for existing mock documents with absolute `tokenUrl`s.

---

amritk (approved):
Approving — the security posture is the part I looked hardest at and it holds up:

- The fetch is user-initiated ("Fetch Configuration" button), not automatic on document load, so an API description cannot make the client reach out to an arbitrary host on its own.
- Both the metadata URL *and* the endpoints it returns are validated, so a compromised metadata document cannot redirect the token exchange to a plaintext endpoint.
- The staleness guard (`name !== schemeName || scheme.oauth2MetadataUrl !== url`) correctly drops a response whose scheme or URL changed while in flight.
- `oauth2MetadataToFlows` treating a declared non-empty Flows Object as authoritative — only filling blank `authorizationUrl`/`tokenUrl`, never adding grants or overwriting scopes — is the right default. Discovering grants only from an empty Flows Object is a sensible line to draw, and the RFC 8414 fallback of `["authorization_code", "implicit"]` matches the spec.

Two notes, neither blocking:

- The JSDoc change from "TLS is required" to "HTTPS, or HTTP on loopback" is a deliberate departure from the 3.2 spec text for that field. I think it is the right call for local development and the allowlist is tight (`localhost`, `127.0.0.1`, `[::1]` only, explicitly not `isLocalUrl`), but it is worth saying so in the changeset rather than only in a comment, since it is a documented-behaviour deviation.
- `oauth2MetadataToFlows` fills `authorizationUrl` and `tokenUrl` but never `refreshUrl`, even though the metadata can imply one. Fine to leave, just noting it in case it was an oversight rather than a decision.

---

PR #10180, opened 2026-09-14
fix(proxy-scalar-com): allow QUERY requests through CORS

hanspagel:
Allows browser clients to send QUERY requests through the proxy. Adds a test for CORS preflight and forwarding a JSON body.


amritk (commented):
Correct and minimal. Two things I verified rather than assumed:

- **No method allowlist blocks QUERY.** `main.go` only branches on `r.Method == "OPTIONS"` and otherwise passes the method straight to `http.NewRequest(r.Method, target, r.Body)`, which accepts arbitrary method tokens. So relaxing the two `Access-Control-Allow-Methods` headers really is the whole fix — nothing else needed to gate on.
- **The new test exercises both halves.** It asserts the preflight does not reach upstream (`calls != 0`) and that the QUERY body survives forwarding, which is the part a header-only change could otherwise silently get wrong.

One pre-existing thing this PR inherits rather than introduces: the allow-methods list is duplicated between `executeProxyRequest` and `corsMiddleware`, so both had to be edited in lockstep and the test asserts the literal string in three places. Worth hoisting to a package-level `const allowedMethods = "POST, GET, OPTIONS, PUT, DELETE, PATCH, QUERY"` at some point so the next method addition is a one-line change — not a reason to hold this up.

---

amritk (approved):

PR #10182, opened 2026-09-14
fix(void-server): allow QUERY requests through CORS

hanspagel:
Allows QUERY in CORS preflight responses. Adds tests for preflight and echoing a QUERY request with a JSON body.


hanspagel:
Replaced by #10190, which updates Hono and keeps the regression tests.

PR #10185, opened 2026-09-14
docs(docs): document Content Signals preferences

hanspagel:
Document Content Signals in the site configuration reference and privacy guide. Explain the default of allowing search, AI input, and AI training; show configuration and generated robots.txt examples; distinguish declaring `no` from disabling the directive; and describe the dashboard controls and advisory nature of the preferences.


amritk (approved):
Documentation-only, two files, no code. The distinction the docs draw twice — that `enabled: false` omits the directive rather than declaring `no`, and that declining a use means keeping it enabled and setting that signal to `"no"` — is the thing people get wrong, so good that it is called out in both the reference and the privacy page. The custom-`robots.txt` caveat and the "advisory, not access control" note are the other two useful bits.

One thing I could not check from this repo: `siteConfig.contentSignals` is not defined anywhere in `scalar/scalar`, so the schema and the dashboard's **Settings → Content Signals** are presumably on the hosted side. Worth confirming the property names (`enabled`, `search`, `aiInput`, `aiTrain`) and the `Content-Signal: search=…, ai-input=…, ai-train=…` output match what ships before this goes live.

Approving on that basis.

---

PR #10186, opened 2026-09-14
feat(openapi): support additional operations

hanspagel:
Adds OpenAPI 3.2 `additionalOperations` support to the workspace store, API client, API reference, and mock server. Document-defined methods appear in navigation and the client method picker, and requests and code examples preserve their case.

```yaml
openapi: 3.2.0
info: { title: Custom method example, version: 1.0.0 }
paths:
  /cache:
    additionalOperations:
      PURGE:
        summary: Purge cached content
        responses:
          '204': { description: Cache cleared }
```

Previously the PURGE operation was omitted. It now appears in the reference and can be selected and sent from the client. Authoring arbitrary new methods in the picker is outside this change.

Custom HTTP methods are case-sensitive, so navigation anchors preserve their spelling. Fixed `get` and `additionalOperations.GET` receive distinct IDs and JSON pointers. The shared 3.1 working model preserves 3.2 operations for processing without claiming they are valid authored OpenAPI 3.1 fields.

The public `OperationMethod` type accepts custom strings while retaining completion for standard methods. The fixed-field `HttpMethod` union stays narrow; consumers handling arbitrary document-defined operations must handle unknown strings. Helpers and workspace-store changesets use minor releases and explain this widening.

The mock server preserves custom-method capitalization when routing. Deployments must configure their HTTP runtime and any reverse proxy to accept and forward those methods.

## Visual

Additional operations in the API client


amritk (commented):
The prototype-pollution handling here is careful in the right places — `preventPollution(method)` before writing into `additionalOperations`, `Object.hasOwn` on read, `Object.hasOwn(REQUEST_METHODS, …)` instead of a bare index, and `constructor`/`__proto__` in the `getHttpMethodInfo` tests. That is the class of bug a string-keyed method map invites and it is covered.

### Worth addressing

1. **`forEachPathItemOperation` silently drops `additionalOperations` keys, inconsistently.** The filter is:

   ```ts
   !isFixedOperationKey(method) && !(method === method.toUpperCase() && isHttpMethod(method))
   ```

   So `additionalOperations: { GET: {…} }` is skipped entirely, but `additionalOperations: { Get: {…} }` passes both clauses and *is* emitted. Both are the same authoring mistake (3.2 says standard methods belong in the fixed fields), and skipping `GET` without any signal means the operation just vanishes from navigation with no explanation. I would either normalise the check to `isHttpMethod(method)` regardless of case so both behave the same, or keep the skip and emit a `console.warn` so the author can find out why their operation disappeared.

2. **`canMethodHaveBody` now answers two different questions.** The new branch is:

   ```ts
   const isExtensionMethod =
     !isHttpMethod(method) && /^[!#$%&'*+.^_`|~0-9A-Za-z-]+$/.test(method) && ['connect', 'track'].includes(normalized)
   ```

   The logic is right (I walked the standard methods: `get`/`head`/`options`/`trace` all still return `false`), but `connect`/`track` are excluded because *browsers forbid the method*, not because it cannot carry a body — and the test name says as much (`rejects browser-forbidden method %s`). Overloading "can have a body" as a proxy for "fetch will accept this at all" will mislead the next caller. A separate `isForbiddenHttpMethod` / `isSendableHttpMethod` would keep both meanings honest. The RFC 7230 token regex is also worth naming and exporting rather than inlining, since #10189 and this PR both need method/token validation.

3. **`getHttpMethodInfo`'s fallback shape changed — that is a breaking change to a `@scalar/helpers` export.** The old fallback returned `{ short, color, backgroundColor }`; the new one returns `{ short, colorClass, colorVar, backgroundColor }`. Anything reading `.color` for an unknown method now gets `undefined`. This is clearly a bug fix (the old fallback did not satisfy `HttpInfo` at all — `color` was never a field on it), but both changesets mark `@scalar/helpers` as `patch`. Combined with the `canMethodHaveBody` widening, this should be `minor`.

4. **`TraversedOperation.method` going from `HttpMethod` to `Type.String()` / `string` is the single biggest decision in the PR** and is why 105 files moved. It removes compile-time checking everywhere a method is consumed, and it is silent — nothing will now catch a typo'd `'pots'` reaching a navigation entry. If you want custom methods plus autocomplete, `HttpMethod | (string & {})` preserves both. At minimum this deserves a sentence in the changeset, because it is the part consumers will feel.

### Nits

5. `setPathItemOperation`'s refactor to `const target = … ? (pathItem['$ref-value'] ?? pathItem) : pathItem` is equivalent to the old branching and reads better — no concern, just confirming I checked it rather than skimmed it.

6. `getPathItemOperationKey` returning `additionalOperations/${escapeJsonPointer(method)}` is the right shape, but it is the only place the pointer layout is encoded. Worth a test asserting a method containing `/` or `~` round-trips.

### Cross-PR

This overlaps #10173 (which also adds `query` to the v3.1 path item) and #10177 (which adds both `query` and `additionalOperations` to the v3.2 loose schema) on the same lines. All three branch from the same `main`, so whichever order they land in, the later two need a rebase rather than a merge-resolve — flagging so the conflict is expected rather than surprising.

---

amritk (commented):
I went looking for the three ways this kind of change usually breaks, and you have closed all of them. Worth recording, because 119 files is a lot to ask a reviewer to take on faith:

**Document-controlled keys cannot pollute.** `setPathItemOperation` calls `preventPollution(method)` before writing into `additionalOperations`, `getPathItemOperation` reads through `Object.hasOwn`, and `getHttpMethodInfo` was switched from `REQUEST_METHODS[normalizedMethod] ?? …` to an `Object.hasOwn` check — with `it.each(['customMethod', 'constructor', '__proto__'])` pinning it. A method literally named `constructor` was previously a live lookup into `Object.prototype`.

**Unknown methods no longer crash the address bar.** `AddressBar.vue` was indexing `REQUEST_METHODS[method].colorVar` directly, which would have thrown on any custom method the moment `method` widened to `string`. Switching it to `getHttpMethodInfo` is the fix, and fixing that function's fallback at the same time was necessary — the old fallback returned `{ short, color, backgroundColor }`, missing `colorClass` and `colorVar` entirely and inventing a `color` key that is not on `HttpInfo`. Adding the explicit `: HttpInfo` return type stops that drifting again. That is a real pre-existing bug fixed in passing and worth its own changeset line.

**Custom methods can still carry bodies.** I expected `canMethodHaveBody` to silently drop request bodies for extension methods, since `BODY_METHODS` only knows the fixed set — but the new `isExtensionMethod` branch handles it, gated on `HTTP_TOKEN` so the method name is a legal token, and `isForbiddenHttpMethod` excludes the ones browsers reject outright. That is exactly the right shape.

`getPathItemOperationKey` returning `additionalOperations/${escapeJsonPointer(method)}` and `deletePathItemOperation` removing the container when it empties are both the kind of detail that gets skipped.

**Two questions rather than objections:**

**Can a user select a custom method in the client?** `HttpMethod.vue` builds its options from `objectEntries(REQUEST_METHODS)`, which by construction cannot contain an extension method. So a `PURGE` operation defined in a document renders and sends correctly, but opening the method picker on it presumably shows a list that does not include the current value — and picking anything replaces it irrecoverably. If that is the intended scope ("view and send document-defined custom methods, do not author new ones"), worth a line in the description; if not, the picker needs the current method appended.

**What happens with `{ get: …, additionalOperations: { GET: … } }`?** `isFixedOperationKey` requires lowercase, so an uppercase `GET` under `additionalOperations` is preserved and surfaced as a separate operation — the comment says this is deliberate so validation can flag the placement rather than the document silently losing an operation. I agree with that choice. What I would want confirmed is that navigation IDs stay unique and the two entries are distinguishable in the sidebar, rather than two rows both reading `GET /pets`.

**On scope:** the `HttpMethod` → `string` widening is the right call and TypeScript is doing most of the verification work — a `Record<HttpMethod, …>` indexed with `string` fails to compile, which is why this touches 119 files. That gives me reasonable confidence the mechanical part is complete. What type checking cannot catch is a `as HttpMethod` cast papering over a site that then does a runtime lookup; if you have a count of casts added in this PR, that would be the useful thing to eyeball.

Good work. My hesitation is entirely about the size of the surface rather than anything I found in it — worth a second reviewer on the call-site sweep specifically.

---

amritk (commented):
`additionalOperations` is a genuinely awkward feature to plumb through, and two of the details here are handled better than I expected: hex-encoding custom method names in `getOperationChunkFilename` so they stay distinct on case-insensitive filesystems, and namespacing `additionalOperations/GET` separately from the fixed-field `GET` so a document declaring both does not collide.

Two questions before approving.

**`TraversedOperation.method` and `TraversedWebhook.method` widen from `HttpMethod` to `string`.** That is unavoidable given arbitrary methods, but it is a public type in `@scalar/workspace-store/schemas/navigation` and it drops the narrowing that every consumer currently relies on — which is most of the 124-file diff. Integrations and customers that switch on `method` lose exhaustiveness checking silently. Two things would soften it: a named type (`type OperationMethod = HttpMethod | (string & {})`) that keeps autocomplete while allowing anything, and a `minor` bump on `@scalar/workspace-store` since this narrows nothing but breaks inference downstream.

**Anchor stability for lowercase custom methods.** `getMethodId` returns the method unchanged when `isHttpMethod` is false, so `additionalOperations: { QUERY }` yields `QUERY/path` while `additionalOperations: { query }` yields `query/path` — two anchors for what an author would reasonably consider the same operation, and the `additionalOperations/` namespace is not applied to either. Is case-normalising non-standard methods for the *id* (while preserving wire case everywhere else) an option, or is exact-case-in-anchors deliberate?

**Worth noting, not blocking:** `additionalOperations` is added to the **3.1** strict `path-item.ts` with a "(OpenAPI 3.2)" comment. I assume that is because the server store upgrades to 3.1 as its working version, so the 3.1 schema is where the plumbing has to live rather than a deliberate relaxation of version rules — a one-line comment saying so would prevent the next reader reaching for the conclusion I first did.

---

cameronrohani (approved):

amritk (commented):
Not approving. 127 changed files with 1,192 additions and 350 deletions, touching workspace-store, API client, API reference and mock server, is beyond what I can meaningfully verify by reading — and there is one design decision here that outlasts the feature and deserves an explicit call.

**Widening `OperationMethod` to accept arbitrary strings is the part to think hard about.** The description puts it plainly: "The public `OperationMethod` type accepts custom strings while retaining completion for standard methods." That is almost certainly implemented as `HttpMethod | (string & {})`, which is the standard trick and keeps autocomplete — but it also means every `OperationMethod` in the codebase stops catching typos. `'gte'`, `'pot'`, a method that should have been normalized to lowercase: all compile now. With 127 files in play, that is a lot of call sites quietly losing a guarantee at once.

That may well be the right trade — `additionalOperations` exists precisely so documents can declare methods the spec does not know about, and you cannot type-check an open set. But it is the same class of decision as the index signature in #10273, and like that one it is worth being deliberate rather than incidental. Two things that would reduce the blast radius: keeping the narrow union as the type used for *fixed* path-item fields and introducing a separate wider type for additional operations, or adding a branded helper so a custom method has to be constructed explicitly rather than assigned from any string.

**The case-sensitivity handling looks right** and it is the detail most implementations get wrong. Custom methods genuinely are case-sensitive per RFC 9110, so preserving spelling in navigation anchors and giving fixed `get` and `additionalOperations.GET` distinct IDs and JSON pointers is correct. It is worth having a test that specifically asserts those two produce different anchors, since a later "normalize the slug" refactor would silently collide them.

Also worth confirming: the mock server registers routes for these methods via Hono. Hono's `app.on(method, …)` accepts arbitrary method strings, so that should work — but a custom method arriving over HTTP only reaches the app if the runtime forwards it, which is worth a note for anyone deploying behind a proxy that whitelists methods.

`minor` changesets with an explanation of the widening is the right call.

---

PR #10187, opened 2026-09-14
docs(dashboard): update API management navigation

hanspagel:
The upload and MCP guides still point readers to the retired Registry dashboard. Update them to APIs, the document editor, Settings > MCP tools, and Settings > Danger zone. Explain how recognized re-uploads can update an existing API, including version, access, and overwrite behavior.

Remove screenshots of the retired flow rather than presenting them as the current interface.


amritk (approved):
Documentation-only. Retiring the Registry dashboard screenshots rather than leaving stale images presented as the current UI is the right call, and the rewritten upload/re-upload/delete steps are clearer than what they replace.

Approving.

---

PR #10188, opened 2026-09-14
fix(mock-server): allow QUERY requests through CORS

hanspagel:
Allows QUERY in the OpenAPI mock server's CORS preflight responses. Updates the CORS test to request QUERY with a JSON content type. QUERY route registration is separate.


hanspagel:
Replaced by #10190, which updates Hono and keeps the regression tests.

PR #10189, opened 2026-09-14
feat(api-client): support OpenAPI querystring parameters

hanspagel:
Adds OpenAPI 3.2 `in: querystring` support across API Client, API Reference, code samples, and Mock Server. A whole-query parameter can now be edited, displayed, serialized, and validated using its media type and examples.

For this API description, the request editor exposes the whole-query value and sends `?status=available&limit=10`. Previously this parameter was omitted from the request editor and outgoing requests.

```yaml
openapi: 3.2.0
info:
  title: Querystring example
  version: 1.0.0
paths:
  /pets:
    get:
      parameters:
        - in: querystring
          name: filters
          content:
              schema:
                type: object
                properties:
                  status:
                    type: string
                  limit:
                    type: integer
              example:
                status: available
                limit: 10
      responses:
        '200':
          description: Matching pets
```

Media-type selection and JSON detection use shared helpers. Example provenance is explicit: parameter-level serialized examples remain URI-ready, while structured and opaque media-type examples follow their corresponding serialization. Editing an inline example replaces its external source. Schema validation preserves boolean schemas, including `false`.

Existing descriptions that combine named query parameters and a whole-query parameter retain both authored values and their combined outgoing query. The UI keeps existing rows editable and explains why adding named rows is disabled when a whole-query parameter is present. Whole-query values are emitted first, followed by named query values and query authentication. Duplicate keys are preserved: `status=available` in the whole-query value and named `status=sold` produce `?status=available&status=sold`; the receiving server determines which values apply.

## Visual

Querystring parameter editor


amritk (commented):
This is the most intricate of the 3.2 PRs and the hard parts are handled well — replacing the previous "treat `querystring` like a named `query` param" shim with real whole-query handling is correct, and keeping the value out of `URLSearchParams` in `resolveRequestFactoryUrl` (so it cannot gain `=` or have its percent-encoding rewritten) is exactly the trap I would have expected this to fall into.

### Worth addressing

1. **Media-type parsing and JSON detection are re-derived five times, with two different regexes.** The pattern

   ```ts
   const [contentType, media] = Object.entries(parameter?.content ?? {})[0] ?? []
   contentType?.split(';')[0]?.trim().toLowerCase()
   ```

   appears in `mock-server/src/utils/querystring-parameter.ts` (twice), `workspace-store/src/helpers/querystring-parameter.ts` (twice) and `validate-request.ts`. And JSON detection uses `/(?:\/|\+)json(?:;|$)/i` in three places and a differently-anchored `/(?:\/|\+)json$/` in two more. Each individual use is correct for its input, but this is precisely what `parseMimeType` in `@scalar/helpers/http/mime-type` exists for, and `AGENTS.md` asks us to reach for `@scalar/helpers` first. #10192 adds `is-xml-media-type.ts` right next to it; a sibling `isJsonMediaType` plus a small `getFirstMediaType(parameter)` would collapse all of this and remove the chance of the two regexes drifting.

2. **The `ParameterLocation` `typeComment` is now wrong.** Both parameter variants gain `literal('querystring')` but keep `'Possible values are "query", "header", "path" or "cookie".` — which flows into generated types. #10177 touches these exact lines and does update the comment, so this is also a guaranteed conflict; worth fixing here so whichever lands second is a clean resolve.

3. **`compileSchema`'s guard change makes it fragile.** `if (!schema)` → `if (schema === null)`. Every current caller passes `?? null` or an explicit `=== undefined ? null :`, so nothing breaks today — but the signature is `Record<string, unknown> | null` and the function now hands `undefined` straight to `ajv.compile`, which throws. Either keep `!schema` or widen the type to `| undefined`. As written the safety depends on all five call sites staying disciplined.

4. **`uriEncoded` / `serialized` encode a three-way state that is very hard to follow.** `getQuerystringParameter` decides between `value` / `dataValue` / `serializedValue` and then signals the result through two booleans, one of which is derived from a reactive-object identity comparison:

   ```ts
   uriEncoded: example?.serializedValue !== undefined && parameterExample === example,
   ```

   `upsertOperationParameter` then re-derives the same distinction to decide which of the three fields to write back. The JSDoc helps, but this is the part of the PR most likely to break silently later — a small discriminated union (`{ kind: 'uri-ready' | 'serialized' | 'data', value }`) would make the invariant checkable instead of conventional.

5. **`delete example.externalValue`** in the querystring branch discards a field the author may have set deliberately, alongside the three that are genuinely being replaced. Intentional?

6. **Mixing `in: querystring` with named `in: query` parameters.** `resolveRequestFactoryUrl` appends the named params after the whole-query value with `&`, and `operation-to-har.ts` folds `queryString` into the URL — so the behaviour is defined. But the UI shows them together under a `Query String` heading with the add-row disabled, which reads as if the named params are part of the whole-query value. Is that combination something you want to support, flag, or just leave as-is? Worth a decision rather than falling out of the implementation.

### Nits

7. `serializeQuerystringParameter`'s final `encodeURIComponent(...).replace(/[!'()*]/g, …)` percent-encodes `=`, `&`, `:`, `,`, `[`, `]` and friends for non-form content types. Valid, and arguably correct for an opaque whole-query value, but a `content: application/json` querystring comes out very hard to read in a code sample. Worth confirming that is intended.

8. `RequestBlock.vue`'s `'Query String'` / `'Query Parameters'` strings are hard-coded — fine here, since `api-client` has no `useLocalization` usage yet, unlike `api-reference`. Noting it only so it is a deliberate gap rather than an oversight.

### Cross-PR

`example.dataValue` / `example.serializedValue` and `parameter.in: querystring` are added to `v3.2/openapi/index.ts` here and again in #10177, on the same lines. Rebase rather than merge-resolve.

---

amritk (commented):
First, a non-finding worth recording so nobody else wastes the time: the two files named `querystring-parameter.ts` are **not** duplicates. `workspace-store/src/helpers` resolves and serializes the whole-query value (request side); `mock-server/src/utils` finds, parses and validates it (server side). Complementary halves, correctly placed. The only genuinely shared concept is "which parameter is the whole-query one", expressed as `getQuerystringParameter` in one and `findQuerystringParameter` in the other — small enough to live with, but worth a comment in each pointing at the other so the next reader does not deduplicate them by mistake.

Also good: `compileSchema` changing `if (!schema)` to an explicit `=== null || === undefined` check, so a falsy-but-valid schema is not treated as absent — the same class of bug #10178 fixes in `getExampleFromBody`.

**Three things I would want before approving.**

**1. Does a malformed query produce a 400 or a 500?**

`parseQuerystringParameter` is documented as *"Throws on malformed JSON or URI escaping"*, and unlike the body schema — which you deliberately wrapped ("resolving a malformed `$ref` should not crash setup") — I cannot see equivalent handling at request time. That parse runs on attacker-controlled input on every request to an operation with a whole-query parameter, so a client sending `?%ZZ` or `?{broken` decides whether the mock server returns a validation error or an unhandled exception. A test asserting a 400 for both a malformed percent escape and malformed JSON would settle it.

**2. Percent-encoding JSON delimiters is spec-correct and an interop hazard.**

From your follow-up:

> Opaque non-form query content is intentionally percent-encoded, including JSON delimiters: those bytes belong to the value rather than acting as query separators.

I agree with the reasoning — `{`, `}`, `"` and `,` are not query-safe under RFC 3986, and treating them as value bytes is right. But a great many real APIs document `?filter={"a":1}` and expect it literally, and a user whose generated cURL snippet suddenly emits `%7B%22a%22%3A1%7D` will read that as a Scalar bug rather than as spec compliance. This deserves more than a PR comment: a line in the changeset, and ideally a note in the querystring documentation explaining the rule and pointing at `serializedValue` as the escape hatch for URI-ready values.

**3. Mixed `in: query` and `in: querystring` needs its ordering pinned.**

OpenAPI 3.2 treats a whole-query parameter as exclusive with named query parameters. Tolerating both rather than dropping one is the right call for existing descriptions — "this PR does not silently delete either set of values" is the correct instinct. What I did not see is a test fixing the *order* of the combined output. Once both sets are emitted into one query string, whether the named parameters come before or after the opaque blob is observable to the server and is exactly the kind of thing that changes silently on a refactor. One test with both kinds present, asserting the full resulting query string, would lock it.

**One cross-PR note:** your follow-up says *"Hard-coded api-client labels remain consistent with that surface's existing localization policy."* #10175 changes that policy — it localizes the API Client across eight locales and converts roughly thirty components. Whichever of these lands second should pick up the other's convention, otherwise this PR's new UI is the one untranslated surface in the client. I have left the matching note on #10175.

**On scope:** 42 files spanning workspace-store, api-client, api-reference, snippetz and mock-server. The primitives I sampled are sound and the validation counts in your description are reassuring, but this is a lot of surface for one reviewer — worth a second pass focused specifically on the snippet plugins, since that is where the encoding decision in point 2 becomes user-visible.

---

amritk (commented):
`in: querystring` needs support and this covers the whole path — client, reference, HAR, snippets and mock-server validation. A few things to resolve first.

**This collides with #10178.** Both PRs add `packages/helpers/src/http/is-json-media-type.ts` and both rewrite `curl.ts`, `wget.ts` and `restsharp.ts` to use it. Neither conflicts with `main`, so CI is green on both, but whichever lands second will conflict. Worth landing the shared helper extraction as its own tiny PR first and rebasing both on it — otherwise the second merge is a manual resolution in three snippet generators, which is exactly where a silently wrong resolution produces bad snippets.

**The `kind` heuristic.** `'uri-ready' | 'serialized' | 'data'` is inferred from which example field the value came from, through a nested ternary that also depends on a second `getExample` call with `content: undefined`. Getting it wrong means either double-encoding a query string or emitting an unencoded one — both silent, both user-visible only as a broken request. The logic deserves to be a small named function with the cases spelled out, and a test per branch (parameter-level `serializedValue`, media `serializedValue`, `dataValue`, plain `value`, schema-generated).

**Two `querystring-parameter.ts` files.** One in `workspace-store` serializing outgoing, one in `mock-server` decoding incoming, each with a comment pointing at the other. They have to agree exactly or the mock server rejects requests the client generates. Is there a round-trip test — generate with one, parse with the other, assert the value survives — across styles, encodings and `allowReserved`? That is the test that keeps them honest as they drift.

**Smaller:** the change in `compileSchema` from `if (!schema)` to an explicit `null`/`undefined` check means an empty-object schema now compiles where it previously did not. Correct, but it applies to path/query/header/cookie/body validators too, not just the new querystring one — worth a line in the changeset.

---

cameronrohani (approved):

amritk (commented):
Not approving — 1,618 additions across 48 files spanning API Client, API Reference, code samples and Mock Server is more than I can verify by reading, and it shares a refactor with several other open PRs.

**Shared-helper collision.** #10178's description says "#10189 uses the identical helper" for JSON detection, and that the canonical home is `@scalar/helpers/http/is-streaming-media-type`. Between this PR, #10176, #10178, #10207 and #10102 there are several branches each carrying their own copy of the same media-type classification helpers. Whichever lands first owns them; the rest need their duplicates *deleted* during conflict resolution, not merged. Worth landing one of them and rebasing the others before any of this goes in.

**The UI decision I would want confirmed by a human.** The description says the editor "keeps existing rows visible and disables adding named rows when a whole-query parameter is present." That is a reasonable reading of 3.2 — `in: querystring` describes the entire query string, so mixing it with named `in: query` parameters is ambiguous by construction. But it means a document that legitimately declares both (which the description says is supported, with "both authored values and their combined outgoing query" retained) presents a request editor where some rows are editable and the add button is dead, with no explanation. Whatever the final behavior, it needs a hint in the UI saying *why*, otherwise it reads as a bug.

It is also worth being explicit about the precedence rule when both are present: if a named `status` parameter and a whole-query parameter that also sets `status` disagree, which wins? The spec does not settle it, so whatever this implements should be documented and tested rather than emergent.

**Good details in the description:** making example provenance explicit (parameter-level serialized examples stay URI-ready while media-type examples follow their own serialization), having an inline edit replace its external source, and preserving boolean schemas including `false` — that last one is the same falsy-value bug class #10275 is fixing elsewhere, so nice to see it handled here deliberately.

---

PR #10190, opened 2026-09-14
fix: update Hono for QUERY CORS support

hanspagel:
Updates Hono from 4.12.9 to 4.13.7 so default CORS responses allow QUERY. Keeps regression tests for void-server and mock-server.


amritk (commented):
Clean dependency bump. I checked the two things that usually go wrong with a catalog change:

- **The changeset scope is exactly right.** `hono` is a real `dependencies` entry in only `@scalar/void-server` and `@scalar/mock-server`; everywhere else (`api-reference`, `openapi-to-markdown`, `agent-chat`, `server-side-rendering`, `integrations/hono`, `galaxy-scalar-com`) it is a `devDependencies`. So patching just those two packages is correct, and leaving `integrations/hono`'s `peerDependencies: ^4.12.5` alone is also right — no reason to narrow a consumer's accepted range.
- **The lockfile resolved.** `pnpm-lock.yaml` pins `hono@4.13.7`, so the catalog bump actually took effect and the new CORS assertions are testing the upgraded middleware rather than passing by accident.

The two new regression tests are the right shape: asserting the sorted `Access-Control-Allow-Methods` set means a future Hono release that drops QUERY fails loudly instead of silently regressing browser clients.

One cosmetic note: the new `createVoidServer({ logger: false })` correctly has no `await` (the factory is sync), but the pre-existing test right below it does `await createVoidServer(...)`. Not a bug in either direction — just slightly odd to read side by side if you ever want to tidy it.

---

amritk (approved):

PR #10191, opened 2026-09-14
feat(api-client): add OAuth device authorization

hanspagel:
Add OpenAPI 3.2 OAuth device authorization to the API client. Selecting the device authorization flow displays a verification link and user code, then polls for an access token while handling approval, denial, expiry, cancellation, and server-requested backoff. Previously, this flow could not be used to obtain a token in the client.

For example, this API description exposes the device flow:

```yaml
openapi: 3.2.0
info:
  title: Device authorization example
  version: 1.0.0
servers:
  - url: https://api.example.com
security:
  - deviceOAuth: [read]
paths:
  /profile:
    get:
      responses:
        '200':
          description: The authenticated profile
components:
  securitySchemes:
    deviceOAuth:
      type: oauth2
      flows:
        deviceAuthorization:
          deviceAuthorizationUrl: https://auth.example.com/device
          tokenUrl: https://auth.example.com/token
          scopes:
            read: Read the profile
```

Device endpoints can also be discovered through OAuth and OpenID Connect metadata. OAuth Basic credentials use form encoding before Base64 encoding, and environment variables are resolved consistently across grant flows. Verification links require HTTPS, with HTTP allowed for local development URLs; polling delays remain bounded by device-code expiry.

The mock server registers approval and denial pages only when the API description contains a device authorization flow, allowing the interaction to be tried locally.

## Visual

OAuth device authorization prompt

https://github.com/user-attachments/assets/ca289b1d-3d41-4019-bbbb-16874784fe7e


amritk (commented):
I read `oauth-device-authorization.ts` against RFC 8628 fairly closely and the protocol handling is good: `slow_down` adds 5s (§3.5), `authorization_pending` continues without treating the 400 as fatal, the device code is never handed to the UI (only `user_code` / `verification_uri` / `verification_uri_complete`), polling is bounded by `expires_in`, and `waitForPoll` removes its abort listener on both paths so a long flow does not accumulate them. That is more careful than most implementations of this grant.

### Worth addressing

1. **The Basic-auth credential encoding differs from every other flow in this file's neighbourhood.** Here:

   ```ts
   headers.Authorization = `Basic ${encode(`${encodeURIComponent(clientId)}:${encodeURIComponent(clientSecret)}`)}`
   ```

   whereas `oauth.ts:476` and `oauth.ts:579` (authorizationCode, clientCredentials/password) do:

   ```ts
   headers.Authorization = `Basic ${encode(`${flow['x-scalar-secret-client-id']}:${flow['x-scalar-secret-client-secret']}`)}`
   ```

   RFC 6749 §2.3.1 is on your side — the values *should* be form-urlencoded before base64 — but the practical consequence is that the same client secret authenticates successfully under `authorizationCode` and fails under `deviceAuthorization` whenever it contains a character `encodeURIComponent` touches (a space, `~`, `+`, `/`). `encodeURIComponent` is also not quite form-urlencoding (space → `%20`, not `+`). Pick one behaviour and apply it to all four flows; my preference would be to fix `oauth.ts` too rather than leave device auth as the odd one out, but either way they should not disagree.

   Related and in the new code's favour: this version runs `resolveValue` (env-variable substitution) over the credentials before encoding, which `oauth.ts` does not. Another divergence worth reconciling in the same pass.

2. **Expiry during a poll wait surfaces a browser-internal message.** The `expiry` signal is merged into `signal`, and `waitForPoll` does `reject(signal.reason)`. For `AbortSignal.timeout` that reason is a `TimeoutError` DOMException, which passes `instanceof Error`, so the user sees "signal timed out" rather than the `Device authorization expired` message you wrote two lines below. The intended message is only reachable via the `Date.now() >= expiresAt` check after the wait resolves. Catching the expiry signal specifically, or comparing against `expiry.aborted`, would make the common timeout path readable.

3. **`interval <= 0` rejects the whole authorization.** A server sending `interval: 0` gets "Invalid device authorization response" and the flow never starts. RFC 8628 does not forbid it, and the defensive intent (don't hot-loop) is better served by clamping to a floor (`Math.max(interval, 1000)`) than by refusing to authenticate.

4. **Protocol posture is inconsistent within the PR.** `verificationUrl` accepts `http:` and `https:`, while `fetchOAuth2Metadata` in the same diff rejects everything but `https:`. The verification URI is only displayed, so allowing `http` there is defensible — but as I noted on #10179, the metadata check as written blocks fetching from the mock server this PR also extends, over `http://localhost`. Please exempt loopback.

### Nits

5. `AbortSignal.timeout(30000)` appears twice as a literal; a named `REQUEST_TIMEOUT_MS` would make it obvious both calls are meant to share the value.

6. `Math.min(data.expires_in * 1000, 2147483647)` — the `setTimeout` clamp is right, but the bare magic number reads better as a named constant or a short comment.

### Cross-PR — please decide before merging

**This PR contains all of #10179.** I diffed the shared files across both branches: `OAuth2Metadata.vue` is byte-identical, and `fetch-oauth2-metadata.ts` differs by one line (this branch also HTTPS-checks `device_authorization_endpoint`). `get-oauth2-metadata.ts`, `oauth2-metadata-to-flows.ts` and the `set-up-authentication-routes.ts` changes are shared too, and both branches target the same `main` commit.

So these two cannot merge independently without a conflict, and every review point on #10179 (the `Value.Check` vs `coerceValue` inconsistency, the empty-object flow updates, the drive-by `tokenUrls.add(getPathFromUrl(...))` fix) applies here verbatim. Cleanest path is to merge #10179 first and rebase this onto it; alternatively close #10179 and let this carry the whole feature, moving that PR's changeset lines over.

---

amritk (commented):
Blocked on mechanics before the code can be reviewed sensibly.

**The base branch is wrong for merging.** This PR targets `codex/oauth2-metadata-url` rather than `main`, so the 1,165 additions across 33 files shown here are the device-authorization diff only. That is fine for reading, but it cannot merge to `main` in this state.

**The stacking instructions in the two PRs contradict each other:**

- [#10179](https://github.com/scalar/scalar/pull/10179) opens with "⚠️ Merge #10191 first."
- This PR says "Stacked on #10179. Merge the OAuth metadata PR first, then rebase this PR onto main."

Both cannot be right, and the branch topology says #10179 is the parent — its head is this PR's base. I read the intended order as #10179 → then rebase this onto `main`, which matches the "Review follow-up" note here and the closing line of #10179 ("Merge this before #10191"). If that is correct, the banner on #10179 is a leftover and should be removed before someone acts on it.

Once #10179 lands and this is retargeted and rebased onto `main`, I will review the device-authorization flow properly. The parts I want to look at closely then, flagging now so they are on your radar:

- The polling loop — `interval` handling, the `slow_down` response, and the termination conditions. The description mentions interval zero being clamped and timeout limits being named, which is the right instinct; I want to confirm there is no path where a denied or expired authorization keeps polling.
- The shared form-urlencoded Basic credential handling, since that is touched for the existing OAuth flows too and is the part of this diff that can regress something already shipping.
- The mock approval and denial pages — worth confirming they are dev-only and cannot be served in a production build.

Not approving while it targets a feature branch.

---

amritk (commented):
`authorizeDevice` is a faithful RFC 8628 implementation — `authorization_pending` continuing silently, `slow_down` adding 5s, doubling the interval on connection timeout per §3.5, clamping `expires_in` to the browser timer maximum, validating every field of the device authorization response before using it, and never surfacing `device_code` to the UI (only `user_code` and the verification URI). The `AbortSignal.any([options.signal, expiry])` composition and the `expiry.aborted && !options.signal.aborted` disambiguation at the end are both correct and easy to get wrong.

Two things before approving.

**`verificationUrl` accepts plain `http:` on any host.** Your #10179, landing at the same time, restricts OAuth2 metadata URLs and their endpoints to HTTPS plus an explicit loopback allowlist. Here an authorization server can hand back `http://evil.example.com/activate` and we will show it to the user as the place to go enter their code — over plaintext, to an arbitrary host. The device flow is the one where we are explicitly instructing a human to visit a URL, so if anything it deserves the stricter check, not the looser one. Could the two PRs share one helper so this cannot drift?

**Size.** 47 files and ~1,900 lines covering the client flow, the mock-server device authorization endpoints (~244 lines), schema additions across 3.1/3.2, secret extraction and types. The client helper above is reviewable on its own; bundled with a mock-server implementation and schema changes it is much harder to give a careful reading. Splitting the mock-server side out would help, if that is still practical.

**Smaller:** `interval *= 2` on connection timeout has no ceiling of its own. The `expiresAt` loop bounds it in practice, but with a long `expires_in` and repeated timeouts the wait between polls can grow very large before expiry stops it. A `Math.min` cap would make the behaviour easier to reason about.

---

cameronrohani (approved):

Issue #10147, opened 2026-09-09
Form-body row key input loses focus on every keystroke, so a field name can only be typed one character per click


nick-lai:
### What happens?

In the Test Request panel, typing into a request-body row's key field loses focus after the first character. The next keystroke lands on `<body>` and is discarded, so a field name can only be entered one character per click. Adding a body field called `note` takes four clicks.

Query, header and cookie parameters behave differently. Their key field keeps focus for the whole name and remounts once at the end.

The cause is the row key. #9896 replaced `:key="index"` with a stable identity key so Vue would stop reusing a `RequestTableRow` instance for the appended placeholder row:

https://github.com/scalar/scalar/blob/ba0caf253feda3c5c5eac57e62bc40e62dbb579f/packages/api-client/src/v2/blocks/request-block/components/RequestTable.vue#L74-L91

Parameter rows carry `originalParameter`, so their key comes from the declared parameter identity and doesn't change while the row is edited. Body rows have no `originalParameter`, so they fall back to `row.name`, which is the value being typed. Every keystroke produces a different key, Vue unmounts the row and mounts a replacement, and the `contenteditable` the caret was in no longer exists.

The body side reaches that state on the first keystroke, because the placeholder row is committed to the row list immediately:

https://github.com/scalar/scalar/blob/ba0caf253feda3c5c5eac57e62bc40e62dbb579f/packages/api-client/src/v2/blocks/request-block/components/RequestBodyForm.vue#L62-L73

Each further keystroke takes the `map` branch, changes `name` again, and re-keys the row again.

Re-keying during a name edit is already known to happen. `RequestTableRow` guards one of its symptoms:

https://github.com/scalar/scalar/blob/ba0caf253feda3c5c5eac57e62bc40e62dbb579f/packages/api-client/src/v2/blocks/request-block/components/RequestTableRow.vue#L109-L117

That guard stops a valid local name from being clobbered by an empty incoming prop. It can't preserve focus once the element itself has been replaced.

### Reproduction

https://sandbox.scalar.com/e/alKkF

One operation with an optional query parameter and a form body, so both key fields can be tried in the same dialog.

Open the Test Request panel for `PATCH /items`. In the Query Parameters table, click the empty key cell of the trailing placeholder row and type `page`. All four characters land and the caret stays put; one `focusout` fires after the last keystroke.

Now do the same in the Request Body table. Only `p` lands. Recorded with a capture-phase listener on `document`:

```
focusin   DIV[undefined Key]
[...]

PR #10192, opened 2026-09-14
feat(workspace-store): generate proper XML examples

hanspagel:
Generate schema-aware XML examples consistently in requests, response examples, snippets, mocks, and Markdown. Mock-server XML response bytes now preserve attributes, namespaces, and root names rather than using `json2xml`; existing downstream response snapshots may change. Supplied serialized XML and edited request bodies remain unchanged.

For example, this complete API description now produces `<pet id="1"/>` instead of treating `id` as a child element:

```yaml
openapi: 3.2.1
info:
  title: Pets
  version: 1.0.0
paths:
  /pet:
    get:
      responses:
        '200':
          description: A pet
          content:
            application/xml:
              schema:
                type: object
                xml:
                  name: pet
                properties:
                  id:
                    type: integer
                    example: 1
                    xml:
                      attribute: true
```

The review has three boundaries: the bounded XML writer and schema mapper, opt-in provenance capture in example generation, and media-boundary consumers. The default JSON path avoids provenance allocation. The writer rejects malformed or incomplete documents; unsupported property patterns produce a diagnostic rather than silently omitting data.

Mapping permits 50 levels and 10,000 nodes; the writer permits 100 levels and 10,000 nodes. A response example that exceeds these bounds displays an explanation and asks for a serialized XML example. The public OpenAPI guide documents media-level `serializedValue`, which preserves a supplied complete payload without building a tree. Other consumers retain their documented fallback and diagnostic callback or console reporting.

## Integration

This branch preserves main's lazy external-example loading, shared example-value selection, edit provenance, lazy agent loading, document-origin handling, and prototype-safe component dictionaries. It remains independent of #10201 and #10203. When combining #10201, use this PR's strict `serializeXmlPart(value, schema?)` adapter: serialized strings bypass it, structured values honor `xml.name` or `root`, and invalid complete documents throw. Whichever branch merges second needs a rebase to select the shared adapter. Likewise, retain the `getExampleContent` options object and pending guard when combining #10203's composition-selection option.

## Visual

Schema-aware XML examples

## Ticket

See #10147


amritk (commented):
`write-xml.ts` is the strongest piece of code across this batch of 3.2 PRs. It gets the things that hand-rolled XML writers almost always get wrong: the real XML 1.0 NCName production for names, the `Char` production for content (correctly excluding `-`, `

`, `- `), `]]>` splitting inside CDATA, `\r` escaped as `&#13;` *outside* CDATA so line-ending normalisation cannot eat it, `\n`/`\t` encoded in attributes because parsers normalise literal whitespace there, pre-binding the reserved `xml` prefix, rejecting `xmlns` as a manual attribute, generated prefixes that avoid collisions with both inherited and explicit ones, and node/depth limits. I could not find a correctness hole in the serializer itself.

Also: thank you for putting `isXmlMediaType` in `@scalar/helpers/http/` rather than inlining a regex. That is the right home, and it is the pattern #10176/#10178/#10189 should follow for JSON media types — between them they add five inline `/(?:\/|\+)json…/` variants, and `snippetz` already carries three private copies of `isJsonContentType` (curl, wget, restsharp). A sibling `isJsonMediaType` next to this file would consolidate all of it.

### Worth addressing

1. **`writeXml` gates on `diagnostics.length` but `XmlDiagnostic.severity` is `'warning' | 'error'`.**

   ```ts
   return { xml: diagnostics.length ? undefined : declaration + body, diagnostics }
   ```

   `get-xml-example.ts:433` does the correct thing — `diagnostics.some(d => d.severity === 'error')` — and `xml-node.ts`'s own comment says "An error never carries a partial or malformed XML document", i.e. errors are meant to be the gate. Today `writeXml`'s `fail()` only ever emits `severity: 'error'`, so there is no live bug. But the moment anyone adds a warning to the writer, valid XML gets silently discarded and callers fall back to JSON with no indication why. One-line fix to match the other module, and worth doing now while the reason is obvious.

2. **`get-xml-body-example.ts` mixes schema versions in one file:**

   ```ts
   import type { SchemaObject } from '@/schemas/v3.1/strict/openapi-document'
   import type { ExampleObject } from '@/schemas/v3.2/strict/example'
   ```

   Given the file's whole purpose is reading `dataValue` / `serializedValue` (3.2-only fields), the `SchemaObject` import should be 3.2 too — and #10177 is moving consumers that way. Related: `schema ?? ({} as SchemaObject)` appears twice; `AGENTS.md` asks us to avoid casts, and an `Partial<SchemaObject>`-friendly signature or an early return would avoid both.

3. **A too-large example produces nothing rather than something truncated.** The `count > 10000 || depth > 100` guard calls `fail()`, which means `xml` comes back `undefined` for the entire document. That is a defensible choice, but it is indistinguishable to the caller from "this schema cannot be represented as XML", and the user just sees the JSON fallback. Worth confirming the diagnostics actually surface somewhere a user or developer can see them — if they are dropped at every call site, the careful `code`/`message`/`path` structure is doing no work.

### Nits

4. `indentChildren` is passed down as the child's `pretty`, so a single mixed-content node disables formatting for its whole subtree, including element-only branches further down. The comment explains why formatting is restricted to element-only content (it must not alter text values), and that reasoning is sound — but the propagation is broader than the stated rule requires.

5. `escape()` escapes `>` unconditionally. Harmless and arguably safer, just not required outside `]]>`.

6. The changeset marks six packages `patch` with `@scalar/workspace-store` as `minor`. Given `openapi-to-markdown`, `blocks` and `mock-server` all change what bytes they emit for XML content types, `patch` may undersell it for anyone snapshot-testing output downstream.

Nothing here is structural — (1) and (2) are both small, and I would take them before merge.

---

amritk (commented):
Not reviewable in its current state — GitHub reports this branch as conflicting with `main`, so it needs a merge from base before the diff means anything.

Beyond the conflict, this is 2,732 additions across 61 files and it is the largest of the open OpenAPI 3.2 PRs by some distance. The PR body has accumulated three rounds of review notes ("Review follow-up", "Two-axis review fixes") covering XML generation, diagnostics, body-edit provenance, Markdown rendering, component-name decoding and response-selection changes. That is a lot of independently-reviewable behaviour in one branch, and it reads like several of those rounds added surface rather than narrowing it.

Two concrete things before I spend a full pass on it:

1. **Resolve the conflict by merging `main` in**, not by rebasing — this branch has seven commits and other people may have it checked out. Regenerate any lockfile or generated output with the repo tooling rather than by hand.
2. **Consider splitting.** The XML builder itself (`writeXml`, the diagnostics channel, the escaping rules) is the substantive change and deserves its own review. The consumer updates — mock-server, openapi-to-markdown, blocks — are mostly mechanical follow-on and would land uncontroversially on their own. Right now a reviewer has to hold all of it at once, which is why this has been open since the 14th with no approvals.

One question I can raise without the diff applying, from the description: limit violations "deliberately reject the complete XML document rather than emit truncated or malformed XML". That is the right call, but it means a document that exceeds a limit renders *no* XML example where it previously rendered something. Worth confirming the limits are high enough that this cannot fire on a realistically large but legitimate schema, and that the failure is visible to the user rather than only in the developer console.

Happy to do a proper review once it merges cleanly — ping me.

---

amritk (commented):
The XML writer is excellent work — full XML 1.0 NCName validation, namespace binding with generated prefixes, `]]>` splitting inside CDATA, carriage returns pushed outside CDATA to survive line-ending normalisation, and explicit `&#10;`/`&#9;` in attributes because parsers normalise literal whitespace. Those are the details that separate a real serializer from `json2xml`, and they are all here. `matchXmlPropertyPattern` refusing to run arbitrary schema regexes on the UI thread and the mock-server event loop, rather than trusting them, is the right instinct.

My concerns are about scope and blast radius rather than correctness of the writer itself.

**61 files and ~2,800 added lines in one PR.** There are at least three separable changes here: the `xml/` subsystem in workspace-store, the provenance-capture mode added to the shared example generator, and the consumer rewiring (mock server, api-reference, blocks, snippets). The first is self-contained and could land and bake on its own; the third is where the regressions will be. Reviewing them as one unit means the risky part gets the same attention as the safe part. Is splitting feasible at this stage?

**The mock server's XML output changes for everyone.** `serializeResponseBody` moves from `json2xml` to `serializeXmlExample`. Anyone mocking an XML API gets different response bodies — better ones, presumably, but different, and mock responses are things people assert against in their own tests. That deserves top billing in the changeset rather than being folded into "generate proper XML examples".

**Provenance capture in the shared generator.** `get-example-from-schema.ts` picks up ~117 lines and an `EXAMPLE_EVALUATION` symbol option. The comment says the default JSON path stays free of provenance allocations, which is the right design — can you confirm that with a benchmark? This function runs for every example on every render, so even a small per-node regression is visible.

**`matchXmlPropertyPattern` returning `undefined`.** Three outcomes (matched / did not match / could not decide) collapse to a boolean in most call sites. What happens on `undefined` — is the property skipped, included, or reported as a diagnostic the user can see? A schema whose `patternProperties` silently stops contributing to the example, with no visible signal, is the failure mode I would want ruled out. Relatedly, the check permits one quantifier inside a character class with no groups or alternation, which I agree is backtracking-safe, but a comment stating that reasoning explicitly would help whoever relaxes these rules later.

---

amritk (commented):
Not approving. `mergeable_state` is `dirty`, and at 2,885 added lines across 62 files spanning workspace-store, API Client, API Reference, blocks, mock-server and Markdown, this is well past what I can responsibly sign off on by reading. It also has not been updated since 18 Sep while `main` has moved a long way, including several other 3.2 PRs in this same batch.

**The design question I would want answered before merge** is the failure mode. The description says mapping is bounded to 50 levels / 10,000 nodes and the writer to 100 levels / 10,000 nodes, that "sufficiently large legitimate examples can exceed these safety limits", and that "current UI consumers do not display a dedicated XML error message; they fall back to their data representation."

Put together, that means a large-but-valid schema silently renders as JSON where the author asked for XML, with a structured diagnostic that only reaches the console. Users will read that as "Scalar does not support my XML" rather than "you hit a node limit". Given the whole point of the PR is that object-to-XML conversion loses schema fidelity, quietly falling back to the lossy representation at the boundary undercuts it. Even a one-line note in the rendered output ("example truncated: schema exceeds N nodes") would make the difference between a bug report and a self-explaining limit. Worth deciding deliberately rather than by default.

Two smaller things:

- The `serializedValue` fast path that preserves a large existing XML payload without building a tree is a good escape hatch — it deserves to be prominent in the user-facing docs, not just the XML README, since it is the answer for anyone who trips the limits.
- The description notes type checks passed only "using temporary local-source aliases because this worktree has no freshly built package outputs". CI covers this, but combined with the conflict it is worth re-verifying after the rebase.

The three-part review structure in the description is genuinely helpful, and the boundary tests (root plus 9,999 children passing, one more failing the whole document) are the right way to pin a limit. This just needs a rebase and a proper read from someone who owns the request-example builder — @DemonHa, most likely.

---

cameronrohani (approved):

amritk (commented):
The earlier writer feedback is addressed. `writeXml` now gates on `severity === 'error'`, `get-xml-body-example.ts` is on 3.2 types with no casts, and the response panels explain generation failures. A few things before this can land:

1. **Merge conflicts with main** in `packages/mock-server/src/create-mock-server.ts`, `packages/mock-server/src/routes/mock-handler-response.ts` and `packages/openapi-to-markdown/src/get-markdown-examples.ts`. They collide with scalar/scalar#10329, scalar/scalar#10186 and scalar/scalar#10189. Please merge `main` in and let CI run again, since the last green run is from the 22nd.
2. **The mock server fails silently on XML** (`packages/mock-server/src/routes/mock-any-response.ts:145`). If generation hits a limit or an unresolved reference, the response is a 200 with an empty body, and the only trace is a `console.warn`. For a mock server, an empty success body is easy to mistake for real data. Consider returning an error status with a short explanation, or at least a response header or log line that names the diagnostic code.
3. Nit: `packages/mock-server/src/utils/serialize-response-body.ts:87` uses `(schema ?? {}) as SchemaObject`, while the workspace-store call sites use `coerceValue(SchemaObjectSchema, {})`. It would be good to use the same approach everywhere and drop the cast.

---

PR #10193, opened 2026-09-14
feat(aspnetcore): add .NET 11 preview support

hanspagel:
Add .NET 11 RC1 support to `Scalar.AspNetCore.Microsoft` alongside .NET 9 and .NET 10. Each target keeps the matching Microsoft generator and Microsoft.OpenApi dependency versions. ASP.NET Core 11 emits OpenAPI 3.2 by default, and applications can explicitly select OpenAPI 3.1.

For example, a .NET 11 endpoint can now produce this API description while retaining Scalar extensions:

```yaml
openapi: 3.2.0
info:
  title: Example
  version: 1.0.0
paths:
  /example:
    get:
      x-scalar-stability: experimental
      responses:
        '200':
          description: OK
```

Previously the Microsoft integration had no .NET 11 target; it now provides one using Microsoft.OpenApi 3.x, while .NET 9 and 10 retain their existing dependency families.

**Release impact:** `Scalar.AspNetCore.Microsoft` is published as `X.Y.Z-rc.1` for every target, including .NET 9 and .NET 10. Stable-only consumers remain on the last stable release unless they opt into prerelease updates. `Scalar.AspNetCore` and `Scalar.AspNetCore.Swashbuckle` continue to publish stable versions. The changeset and documentation now call out this consequence explicitly. The release policy still needs maintainer acceptance before merging this draft.

All .NET integrations require SDK `11.0.100-rc.1.26425.128`; contributor documentation explains the requirement and older runtime prerequisites. The playground intentionally builds with that RC SDK and retains its .NET 10 runtime image.

The Aspire JSON helpers now dispose parsed documents and clone returned elements so their lifetime is independent of the document. C# cast spacing and final-newline behavior follow the existing .NET `.editorconfig` (`csharp_space_after_cast = true`, `insert_final_newline = false`).

Browser verification of the .NET 11 path remains outstanding.

## Ticket

Related to [discussion #8230](https://github.com/scalar/scalar/discussions/8230).


amritk (commented):
Adding the `net11.0` TFM and the per-TFM `Microsoft.OpenApi` pin is the right mechanism, the `dotnet-11.md` table is genuinely useful (the "do not upgrade Microsoft.OpenApi to 3.x on the .NET 9/10 generator" warning especially), and rewriting `ExcludeFromApiReferenceTransformerTests` from a golden-string match to structural assertions is a real improvement — tag ordering varying between generator versions is exactly the kind of thing that should not fail a test.

Three things I would want resolved before this merges, in order of consequence.

**1. `VersionSuffix: rc.1` makes `Scalar.AspNetCore.Microsoft` prerelease-only for everyone, not just .NET 11 users.**

```xml
<!-- TODO: Remove the suffix when the ASP.NET Core 11 dependencies are stable. -->
<VersionSuffix>rc.1</VersionSuffix>
```

combined with the workflow switching that one package from `/p:Version=` to `/p:VersionPrefix=`, means every future publish of this package is `X.Y.Z-rc.1` — including its `net9.0` and `net10.0` assemblies. Consumers on stable .NET 10 who take `Scalar.AspNetCore.Microsoft` from NuGet stop receiving updates entirely unless they opt into prereleases (`--prerelease`, or a floating prerelease version). That is a significant distribution change for users who have nothing to do with .NET 11, and the changeset just says `minor`.

If the goal is "ship .NET 11 support without disrupting .NET 10 users", the usual shape is to keep the package stable and publish the RC-dependent TFM separately, or to hold the `net11.0` target on a prerelease branch until RTM. If you do want the whole package on `-rc.1` for now, that deserves a line in the changeset and probably a note in the README, because it will generate support questions.

**2. `global.json` now requires the RC SDK for every .NET integration in the repo.**

```json
"version": "11.0.100-rc.1.26425.128",
"allowPrerelease": true
```

This is repo-wide for `integrations/dotnet`, so aspire, azure-functions and swashbuckle contributors all need the .NET 11 RC1 SDK installed to build anything, even though only `Scalar.AspNetCore.Microsoft` targets net11.0. CI is handled (the workflows install both SDKs), but local development is not. `rollForward: latestFeature` will not roll *down* from 11 to 10, so a contributor with only the 10 SDK gets a hard failure. Worth either documenting this prominently in the .NET contributing docs or checking whether a lower `version` with `allowPrerelease: true` gets you what you need.

Related: the playground `Dockerfile` moves to `mcr.microsoft.com/dotnet/sdk:11.0.100-rc.1`, so the published playground image is built on an RC SDK too. Intentional?

**3. Unrelated churn in `ScalarResourceConfiguratorTests.cs`.**

The `(string)envVars[…]` → `(string) envVars[…]` cast-spacing changes across that file are pure formatting and make the diff harder to read. More importantly, two files now end **without a trailing newline**:

- `integrations/dotnet/aspire/tests/Scalar.Aspire.Tests/ScalarResourceConfiguratorTests.cs`
- `integrations/dotnet/aspnetcore/tests/Scalar.AspNetCore.Microsoft.Tests/ExcludeFromApiReferenceTransformerTests.cs`
- `integrations/dotnet/aspnetcore/tests/Scalar.AspNetCore.Microsoft.Tests/OpenApiVersionTests.cs` (new file)

That looks like an editor artifact rather than intent — worth restoring.

The `using var document = JsonDocument.Parse(…)` and `.Clone()` additions in the same file *are* genuine fixes (returning `JsonElement`s from a disposed `JsonDocument` is undefined), so no objection to those riding along — just call them out in the description so they are not mistaken for formatting.

Finally, the description notes "Browser testing is still pending". Worth completing that before merge, or saying explicitly that the .NET 11 path is untested end to end.

---

cameronrohani (approved):

hanspagel:
closing until 11 is officially out

PR #10194, opened 2026-09-14
fix(api-reference): allow editing configured server variables

### LINKED ISSUE #8106: can’t change variables passed in the configuration
when passing servers with variables in the configuration to createApiReference, I can’t change the value

> when passed in createApiReference - everything works, but attempting to change variable prints in console: Variable not found

hanspagel:
Fix editing server variables supplied in configuration. Keep the reference, code samples, and request modal in sync.

Build, type checks, scoped lint and Knip, and browser checks pass. All reference tests pass. Workspace-store has one existing Node filesystem cleanup test failure.

## Visual

| Before | After |
| --- | --- |
| Before | After |

## Ticket

Fixes #8106


amritk (approved):
The diagnosis is right: config servers were being substituted at read time (`normalized.config.servers ?? document.servers`) in two places, so there was no single object for an edit to land on. Putting them *into* the client document once and reading only from there gives edits somewhere to go, and `runtimeConfig` then feeds the edited values to the modal and sidebar so all three surfaces agree.

The test coverage is the reason I am comfortable with this — it covers the cases that actually break this kind of hook:

- Edits do not leak back into the caller's `configuration` object or the source store (`expect(servers[0]?.variables.env.default).toBe('prod')`), which the two `deepClone` calls exist for.
- An unrelated config change (`hideModels`) does not clobber an in-flight edit — that is what the `isObjectEqual` snapshot comparison buys.
- An *in-place* mutation of the configured default does propagate, which is the opposite case and easy to lose.
- Removing the override restores the source document's servers.
- `servers: []` is respected rather than treated as absent — the `!== undefined` checks rather than truthiness.
- Reload re-applies, and two documents keep independent values.

The `get-request-example-context.ts` change is consistent with that precedence: once `options.servers` is supplied, operation-level servers no longer win the meta, so variable edits target the document the user is actually looking at.

CI green, changeset covers both packages.

Approving. Two things I would like a comment for rather than a change:

**The `flush: 'sync'` is load-bearing and undocumented.** `ensureDocumentLoaded` calls `getServers(document.servers, …)` immediately after `addDocument`, and it only sees the configured servers because the sync watcher has already fired on the `documents[slug]` write. With the default `flush: 'pre'` that ordering breaks silently. Worth a line saying so, otherwise someone will "clean up" the flush option later.

**`deepClone(config.servers)` runs inside the watch getter**, so it re-clones every configured server on each reactive tick of anything the getter touches, synchronously. Fine for the handful of servers a config normally has, but it is the kind of thing that stops being fine in a workspace with many documents. Not worth restructuring now — just worth knowing it is there.

---

PR #10195, opened 2026-09-14
docs(sdks): add private registry publishing guide

amritk:
## Problem

Customers want to publish generated SDKs to internal enterprise registries — JFrog Artifactory, AWS CodeArtifact, Nexus, GitLab, Azure Artifacts. Our built-in publishing only covers the public registry for each language, so the question keeps coming up as a roadmap request.

It is already possible today. A workflow added under `.github/workflows/` is ordinary custom code: the generator has no delete path, and each build is a plain `git merge` of `scalar-generated` into `scalar-next`, which never removes a file that exists on only one side. A workflow a customer adds survives every regeneration indefinitely.

Nothing documented this, so nobody knew. Worse, the obvious first attempt fails silently — see below.

## Solution

A new **Private Registries** page under SDK publishing, with everything needed to get going and nothing that needs adapting from a different context.

**Complete copy-paste workflows**, not fragments:

- npm → Artifactory
- Python → Artifactory
- npm → AWS CodeArtifact, authenticating by assuming an AWS role with GitHub OIDC and minting a short-lived token, so nothing long-lived is stored in the repository
- The five-step shape to adapt for any other host

**The page leads with the trigger constraint**, because it is the failure people hit first and it gives no error. `release-please.yml` cuts the tag and the GitHub Release using the workflow's built-in `GITHUB_TOKEN`, and GitHub does not start workflow runs from events created by that token. So `on: release` and `on: push: tags:` — the two most natural things to reach for — never fire, with nothing in the Actions tab to explain why. Two triggers that do work are given instead: a `push` filtered to `.release-please-manifest.json` (rewritten only by the release commit, which a person merges), and `workflow_run` chained off the release workflow.

Also covered: which workflow paths are generator-owned versus yours, why the file belongs on `scalar-next` and not `scalar-generated`, testing via `workflow_dispatch` before a real release, an idempotency guard so re-runs skip an already-published version, whether this replaces or complements public publishing, and a troubleshooting table.

Cross-linked from the publishing overview, the package registries table, and the custom code guide, and registered in the sidebar after Package Registries.

### Verification

- Every YAML block in the page parses, and the three complete workflows each carry `name`, `on`, `permissions`, and `jobs`.
- Every internal link and anchor resolves against the files it targets.
- Confirmed against the generator that the npm example works unmodified: generated SDKs set no `publishConfig` and ship no `.npmrc`, so `setup-node`'s `registry-url` alone decides the destination. Node and Python versions, build commands, and action majors match what the generated repositories already use.
- `scalar.config.json` is valid, and its diff is only the six added lines.

## Checklist

- [ ] I added a changeset. 
- [ ] I added tests. 

https://claude.ai/code/session_01BHLgPXQpc3Zf4stFxeyPJ2

---
_Generated by [Claude Code](https://claude.ai/code/session_01BHLgPXQpc3Zf4stFxeyPJ2)_


hanspagel (approved):
nice

PR #10197, opened 2026-09-15
fix(api-reference): preserve host-app hash prefix when copying section links

AshishKumarPathak:
---
'@scalar/api-reference': patch
---

fix(api-reference): preserve host-app hash prefix when copying section links

When scalar is embedded inside a hash-routed SPA (e.g. the page URL is
`http://localhost:3000/#docs/api-spec`), copying a section anchor link was
producing a broken URL like:

```
http://localhost:3000/#tag/payer-list
```

instead of the correct:

```
http://localhost:3000/#docs/api-spec/tag/payer-list
```

This happened because `makeUrlFromId` with no `basePath` configured was
blindly overwriting the entire hash fragment, losing the host-app prefix.

The fix detects the existing host-app hash prefix and preserves it when
constructing the copied URL. Navigating between sections also correctly
replaces only the section portion without duplicating the prefix.

The recommended long-term solution for apps with hash-based routing is to
configure `pathRouting: { basePath: '#docs/api-spec' }`, which gives scalar
full ownership of its hash namespace.

### DIFF
--- a/.changeset/feat-openapi-32-nested-tags.md
+++ b/.changeset/feat-openapi-32-nested-tags.md
@@ -1,0 +1,33 @@
+---
+'@scalar/workspace-store': minor
+---
+
+feat(workspace-store): support OpenAPI 3.2 `parent` field for nested sidebar tags
+
+OpenAPI 3.2 introduced a `parent` field on `TagObject` that declares a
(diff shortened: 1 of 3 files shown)


hanspagel (approved):
thanks!

PR #10198, opened 2026-09-15
feat(workspace-store): support cookie serialization style

hanspagel:
Adds OpenAPI 3.2 `style: cookie` serialization without percent-encoding values that are already cookie-ready. Structured cookie values use semicolon-separated pairs and retain existing percent escapes when combined with legacy cookies, explicit Cookie headers, and authentication cookies.

```yaml
openapi: 3.2.0
info: { title: Cookie example, version: 1.0.0 }
paths:
  /items:
    get:
      parameters:
        - in: cookie
          name: preferences
          style: cookie
          schema:
            type: object
          example:
            language: en
            greeting: hello%20world
      responses:
        '200': { description: OK }
```

This sends `Cookie: language=en; greeting=hello%20world` instead of joining entries with `&` or encoding `%` again. An invalid `explode: false` value emits a developer-console warning once per parameter name and falls back to separate cookie pairs.

Browser snippets with cookies enable credentials so the browser can send its own cookie jar. Browsers still forbid setting a Cookie header directly, and cross-origin servers must allow credentialed CORS. Snippets with no cookies remain unchanged.


amritk (commented):
The 3.2 `style: cookie` work itself is well gated — `getParameterStyleAndExplode` and `buildRequestParameters` both check `'style' in param && param.style === 'cookie'`, so documents that do not opt in keep the existing `form` path. `serializeCookieStyle` expanding arrays and objects into separate entries regardless of `explode` is right, and the comment explaining why `explode: false` is invalid for cookies (RFC 6265) is the kind of note that saves the next reader a spec lookup.

What stops me approving is that two of the changes are **not** gated on the new style and will change output for documents that never use it:

**1. `browser-request.ts` changes every snippet that has any `Cookie` header**

```ts
const headers = (request.headers ?? []).filter(({ name }) => name.toLowerCase() !== 'cookie')
```

Previously a `Cookie` header — for example from a plain `{ name: 'Cookie', in: 'header' }` parameter — stayed in `headers` and the xhr/jquery snippets emitted `setRequestHeader("Cookie", …)`. Now it is stripped and re-emitted as `document.cookie = …` setup lines instead. That is arguably the more correct output, since browsers forbid setting `Cookie` via `setRequestHeader`, but it silently rewrites existing users' snippets and the changeset only mentions 3.2 cookie serialization.

The part I would want a second look at is this line:

```ts
withCredentials: Boolean(cookieValues.length)
```

`withCredentials` now flips to `true` for any request carrying a `Cookie` header, where before it required `request.cookies` to be non-empty. In a generated snippet that means credentialed cross-origin requests for a case that previously was not credentialed. Worth confirming that widening is intended.

**2. `operation-to-har.ts` empties `harRequest.cookies` whenever a `Cookie` header exists**

```ts
const cookieHeader = harRequest.headers.find((header) => header.name.toLowerCase() === 'cookie')
if (cookieHeader && harRequest.cookies.length) { … harRequest.cookies = [] }
```

The guard is the presence of a `Cookie` header, not the presence of cookie-style parameters. A document with an explicit `Cookie` header parameter plus global cookies or an `in: cookie` security scheme now gets a merged header and an empty `cookies` array, so any HAR consumer reading structured cookies loses them. Same for `processParameters` returning `cookies: cookieStyleEntries.length ? [] : harRequest.cookies`.

Both may well be what you want — but as written they are general HAR/snippet behaviour changes riding along with a 3.2 feature, and the tests only cover the cookie-style path. Two suggestions:

- Narrow the `operation-to-har.ts` merge to the cookie-style case, or add a test proving the explicit-`Cookie`-header-plus-global-cookies case is intentional.
- Add a `browser-request` test for a request with a `Cookie` header and no `request.cookies`, pinning the new `withCredentials: true`.

One smaller thing: the merged header mixes encodings — cookie-style entries are raw while the legacy ones go through `encodeURIComponent`. The comments say this is deliberate ("preserving legacy encoding for other cookies") and the `greeting=Hello%2C%20world!; global=a%20b; token=secret` assertion pins it, so this is just a note that anything parsing that header back out has to know it cannot uniformly decode.

Happy to approve once the two ungated changes are either narrowed or covered by tests that show they are intended.

---

amritk (commented):
Supporting 3.2 `style: cookie` is right, and keeping the un-encoded entries out of the HAR `cookies` array — where every snippet generator would percent-encode them — is the correct workaround for a HAR model that has no concept of "already serialised". Not approving because two of the changes reach well beyond documents that use cookie style.

**`browser-request.ts` changes snippets for everyone.** The `Cookie` header is now stripped and rewritten as `document.cookie` assignments, and `withCredentials` is set whenever any cookie value exists — including for requests with no cookie-style parameters, where previously a plain `Cookie` header was emitted. Anyone whose XHR/jQuery snippet currently shows a `Cookie` header gets a different snippet. Browsers do forbid setting `Cookie` on XHR, so the new output is more honest, but:
- `document.cookie = "...; path=/"` in the snippet only works when the reader runs it on the request origin. The comment says so, but for a cross-origin API the snippet now silently does nothing useful where before it at least showed the intended header.
- `withCredentials: true` changes what the request sends. For a snippet that previously worked cross-origin without credentials, flipping it on will fail CORS preflight unless the server sets `Access-Control-Allow-Credentials`. Is that trade-off intended for the non-cookie-style case, or should the old behaviour be kept when `hasCookieStyleEntries` is false?

**Collapsing `harRequest.cookies` into the header.** When cookie-style entries are present, all HAR cookies — auth cookies and global cookies included — are folded into the `Cookie` header and `harRequest.cookies` is emptied. Every snippet generator sees a different request shape for those documents. Worth checking what that does to the snippet output across the generator matrix, not just the browser ones.

**`serializeCookieStyle` ignores `explode`.** The doc comment says `explode: false` is invalid for cookie style, but the function does not take `explode` and always expands. So a document declaring `style: cookie, explode: false` is silently treated as exploded rather than rejected or warned about. Given the comment already identifies it as invalid, saying so somewhere the author can see it would be better than quiet coercion.

**Mixed encoding within one header.** The joined value interleaves percent-encoded legacy cookies with raw cookie-style ones. That is deliberate and documented in the comment, but it means one header can carry two encodings — worth a test with both kinds present to pin the exact output.

---

cameronrohani (approved):

amritk (approved):
Approving — the serialization work is spec-correct and unusually well covered. Two things I would fix before merging, neither of which is a correctness bug.

**1. The README contradicts the code this PR adds.** `packages/snippetz/README.md` says:

> Scalar tolerates that invalid combination by expanding arrays and objects into separate cookie entries; **it does not emit a validation warning**.

But `serializeCookieStyle` does exactly that:

```ts
console.warn(`Cookie parameter "${name}" uses invalid explode: false with style: cookie; serializing with explode: true.`)
```

and the changeset advertises the warning too. Since this README ships to npm, that sentence should say the opposite.

**2. The warning is unthrottled.** `serializeCookieStyle` runs during request building and snippet generation, so a document with `explode: false` on a cookie parameter will log on every re-render, not once. Worth hoisting to a warn-once per parameter name, or moving it to where the parameter is parsed rather than where it is serialized.

**What I checked and liked:**

- Not percent-encoding cookie-style values is the whole point, and the tests pin it precisely — `greeting=Hello%2C%20world!` survives intact while the legacy `a b+c` still becomes `a%20b%2Bc` in the same header, and `expect(snippet).not.toContain('Hello%252C%2520world')` guards against the double-encoding regression directly.
- Running that assertion across `snippetz().plugins()` means every generator is covered rather than just the two that were changed.
- `explode: false` genuinely is invalid for cookies — comma-separated values violate RFC 6265 — so expanding rather than rejecting is the right tolerance, and the three-way `absent`/`disabled`/`empty` test confirms structured HAR cookies are preserved when no cookie-style parameter is in play.
- `process-parameters.ts` copies the header array and the test asserts the input `harRequest.headers` is untouched, which is easy to get wrong when splicing.

**One thing to be deliberate about:** `withCredentials` now turns on for *any* Cookie header in XHR/jQuery, including existing documents with no cookie-style parameters. That changes generated snippets for current users and brings credentialed-CORS requirements with it. It is documented in both the changeset and the README, and browsers genuinely forbid setting `Cookie` directly, so I think it is right — but `minor` might describe it better than `patch`.

**Stray file:** `header.e2e.ts.snapshots/scalar-galaxy-header-mobile.png` is re-baselined here. That same binary also appears in #10274, #10275 and #10278 — four unrelated PRs, so it is a flaky baseline, not a real visual change. Whoever lands first will leave the other three conflicting on it; worth fixing the flake once on `main` and dropping the file from all four.

---

PR #10199, opened 2026-09-15
fix(workspace-store): honor discriminator `defaultMapping` in examples

hanspagel:
## Problem

Generated examples picked the first variant even when the discriminator defined a fallback.

## Solution

Use `defaultMapping` when the discriminator property is absent or unmapped. Keep explicit selections and matching mappings first. Includes tests and a patch changeset.


amritk (approved):
Low risk in the way that matters: `getDiscriminatorSelectionIndex` returns `undefined` immediately when `discriminator.defaultMapping === undefined`, so every existing document takes exactly the path it takes today. The blast radius is limited to documents that opt into the 3.2 field.

The resolution order is right and the tests pin each branch:

- Explicit `mapping` wins over the implicit component name (`{ Cat: 'Dog' }` with `kind: 'Cat'` selects `Dog`).
- An unmatched or absent tag falls through to `defaultMapping`.
- A *broken* explicit mapping returns `undefined` rather than silently selecting the fallback — the comment says why, and I agree that is the better failure: a mapping that points at a schema outside the union is an authoring error, and quietly substituting the fallback would hide it.
- `findTarget` trying `#/components/schemas/<escaped>` before treating the string as a raw reference is the right precedence, and routing the name through `escapeJsonPointer` means a component containing `/` or `~` still resolves. Covered by the `./other.json`, `https://example.com/other.json` and `Other~1Pet` cases.

Nice detail: inline `title`s are deliberately *not* treated as implicit mappings ("does not infer discriminator mappings from inline schema titles"). Only `$ref`'d variants have implicit names per spec, and getting that wrong would have silently changed examples for anyone using titled inline variants.

`Object.hasOwn` on the mapping rather than `in` also keeps a variant named `toString` from resolving through the prototype — the `{ tag: 'toString', result: 'OtherPet' }` case.

Changeset present. All 44 checks green.

Approving.

---

PR #10200, opened 2026-09-15
chore(api-client): update runtime dependencies

hanspagel:
Update js-base64, pretty-ms, set-cookie-parser, Zod, and PostHog. Shared dependencies are bumped in the catalog for all consumers.


hanspagel:
nah, those few are irrelevant

PR #10201, opened 2026-09-15
feat(workspace-store): support positional and nested multipart encoding

hanspagel:
Send positional and nested multipart bodies with `prefixEncoding` and `itemEncoding`, preserving part order and uploaded file bytes. Requests and generated snippets use the same MIME serializer; snippets show placeholders for uploaded files.

Previously, the example below did not produce an ordered JSON part followed by nested text parts. It now sends `{"id":1}` first, then a separate nested multipart body containing `hello` and `world`, with independent boundaries. Prefix encodings apply by index, and item encoding applies only after the prefix ends.

```yaml
openapi: 3.2.1
info:
  title: Nested multipart upload
  version: 1.0.0
paths:
  /upload:
    post:
      requestBody:
        content:
          multipart/mixed:
            schema:
              type: array
              prefixItems:
                - type: object
                  properties:
                    id:
                      type: integer
                - type: array
                  items:
                    type: string
            example:
              - id: 1
              - [hello, world]
            prefixEncoding:
              - contentType: application/json
              - contentType: multipart/mixed
                itemEncoding:
                  contentType: text/plain
      responses:
        '204':
          description: Uploaded
```

The serializer rejects ambiguous named positional items and nesting beyond eight levels. Equal-length random boundaries cannot be prefixes of one another, and candidates found in resolved text or nested delimiters are regenerated. Binary files remain opaque Blob chunks with 192-bit random boundary protection.

Content-type defaults follow the [OpenAPI 3.2.1 Encoding Object rules](https://raw.githubusercontent.com/OAI/OpenAPI-Specification/refs/heads/main/versions/3.2.1.md). Structured XML parts use `xml.name` or a stable `root` through a dedicated adapter; full schema-aware XML integration remains separate in #10192.


amritk (commented):
The header-injection handling in `encode-multipart-body.ts` is the part I went looking for, and it is done correctly:

```ts
const resolved = replace(value)
if (!/^[!#$%&'*+.^_`|~0-9A-Za-z-]+$/.test(name) || /[\r\n\0]/.test(resolved)) {
  throw new Error('Invalid multipart header')
}
```

Validating the header *name* as an RFC 7230 token and the *value* for CRLF/NUL is right, and crucially the check runs **after** `replace()`, so an environment variable expanding to `foo\r\nX-Evil: bar` cannot smuggle a header in. The same CRLF guard on both the outer `contentType` and each `partContentType` closes the other obvious route. Given that encoding `contentType`s and header examples both come straight from the API description, this is exactly where that validation belongs.

`selectContentType` is also nicer than it needed to be — honouring a range like `image/*` by substituting the concrete essence while keeping the range's parameters, and falling back to the first non-wildcard choice, is more thought than most implementations give it.

**Three things I would like addressed:**

**1. Named positional multipart silently drops all but the first key.**

```ts
if (named && item !== null && typeof item === 'object' && !Array.isArray(item)) {
  const entry = Object.entries(item)[0]
  return [entry?.[0], entry?.[1], itemEncoding, resolveLeafSchema(itemSchema, [entry?.[0] ?? ''])]
}
```

For a `multipart/form-data` body whose example is an array of objects, an item like `{ id: 1, name: 'x' }` becomes one part named `id` and `name` is gone — no error, no warning, and the generated snippet and the sent request both quietly omit it. If one-key-per-item is the intended contract (which the OpenAPI 3.2 positional encoding rules arguably imply), it should either fail loudly or be documented in a comment right there; as written it reads like an oversight. Worth a test either way.

**2. No recursion bound on nested multipart.** `buildMultipart` recurses into itself for a nested part, and `serializeMultipartBody` recurses in step, with no depth cap. #10207 caps its multipart *parser* at `MAX_NESTING = 8` for the same structure; the builder should be symmetric. The input here is a value rather than a network stream, so it is less exposed — but workspace values are proxy-backed and a self-referential example would recurse until the stack goes. A depth parameter mirroring #10207's would cost three lines.

**3. Dead branch in the encoder.**

```ts
const nested = part.type === 'multipart' ? serializeMultipartBody(...) : undefined
…
if (nested) { return [headers, ...nested.chunks, '\r\n'] }
if (part.type === 'multipart') { return [] }
```

`nested` is always truthy when `part.type === 'multipart'`, so the second check is unreachable. Harmless, but it implies a case that cannot happen and will make the next reader look for it.

**One question rather than a finding:** `serializePartValue` picks the XML root name as `schema?.xml?.name ?? (Object.keys(value).length === 1 ? undefined : 'root')`. So a single-key object is emitted unwrapped and a two-key object gains a literal `<root>`. That is a defensible heuristic, but it means adding a second property to a schema silently changes the XML shape of the part. Worth confirming against what #10192 does for XML generation — the two should agree on root naming, and right now they are separate implementations.

---

amritk (commented):
Keeping parts structured until send time so nested text can still resolve environment variables is the right architecture, and gating everything behind `needsMultipartEncoding` means ordinary `multipart/form-data` bodies keep their existing path — that containment is what makes this reviewable at all. `MAX_MULTIPART_NESTING` and rejecting `[\r\n\0]` in a content type before it reaches a header are both good.

Questions before approving.

**`selectContentType` wildcard resolution.** When an `encoding.contentType` is a range, the code picks a match, and if the match still contains a wildcard it rewrites only the essence via `match.replace(/^[^;]+/, preferred.essence)` while keeping the range's parameters. That is clever but subtle — `image/*; charset=utf-8` against an inferred `image/png` yields `image/png; charset=utf-8`, carrying a parameter the author attached to the range rather than to that type. Is that intended? And what is the behaviour when the list is all wildcards and none matches (the final `choices.find(...) ?? fallback`)? Each of these branches deserves a named test.

**`defaultContentType` inference.** A schema with no `type`, or `type: 'string'` with any `contentEncoding`, becomes `application/octet-stream`. The no-`type` case is the one I would double-check: an untyped schema is "anything", and defaulting it to binary rather than `text/plain`/`application/json` could mislabel ordinary values. What drove that choice?

**`serializePartValue` XML root naming.** Falling back to `schema?.xml?.name ?? 'root'` duplicates logic that #10192 is replacing wholesale with a real XML serializer. These two will disagree: a part serialized here uses `json2xml` while the same value in a response body uses `serializeXmlExample`. Worth deciding now which one owns XML, rather than shipping both and reconciling later.

**Scope:** ~480 non-test lines across the store, blocks and HAR generation. Not unreasonable, but combined with the other multipart/streaming PRs in flight it is a lot of concurrent change in the same request-building code.

---

cameronrohani (approved):

amritk (commented):
Not approving — 1,241 additions across 18 files implementing `prefixEncoding`, `itemEncoding` and nested multipart is more than I can verify by reading, and multipart serialization is unforgiving: a boundary bug produces a request that looks fine in the UI and is rejected or silently misparsed by the server.

**Specific things I would want checked**, all of which are the standard ways nested multipart goes wrong:

1. **Nested boundary uniqueness.** A nested `multipart/mixed` part needs a boundary that cannot appear in any enclosing or enclosed body. The dangerous case is an inner boundary that is a *prefix* of the outer one (or vice versa) — parsers that match on `--<boundary>` at line start will terminate the wrong part. Worth asserting generated boundaries are neither equal nor prefix-related at any nesting depth, and that a part whose *content* happens to contain the boundary string is handled.
2. **Per-part `Content-Type` from `prefixEncoding` / `itemEncoding`.** The example in the description has `prefixEncoding[0].contentType: application/json` and a nested part with `itemEncoding.contentType: text/plain`. The positional mapping is the subtle bit: `prefixEncoding` applies by index to leading items, `itemEncoding` to everything after the prefix runs out. Worth tests for prefix shorter than the array, longer than the array, and exactly equal.
3. **File bytes.** "Preserve file bytes" is claimed — the thing to verify is that binary parts are never routed through a string encode/decode step, which silently corrupts anything non-UTF-8. A round-trip test with a byte sequence that is invalid UTF-8 would prove it.
4. **Part order.** Positional encoding only means anything if order survives; since this flows through both the request builder and the snippet generators, both paths need the assertion.

**On the spec reading:** linking the 3.2.1 specification directly is the right instinct and matches this repo's guidance. One thing worth confirming against it is the interaction between `itemEncoding` and `contentType` defaults — 3.2 has default content-type rules based on the item schema's type, and an explicit `itemEncoding.contentType` should override them rather than merge.

`mergeable_state` is `clean`, which is more than most of this 3.2 batch can say. Given the overlap with #10207 and #10189 on request construction, it is worth thinking about merge order across those three as well.

---

PR #10202, opened 2026-09-15
fix(openapi-validator): explain invalid URI references

hanspagel:
Explain invalid URI references and how to fix OpenAPI component names. Return a normal parser error for malformed reference escapes.

## Examples

- `#/components/schemas/Användare`: explains that the reference contains non-ASCII characters. For OpenAPI components, rename `Användare` to `Anvandare` and update its references. Encoding the reference alone does not fix the component name.
- `#/properties/first name`: explains that the space must be removed or encoded as `%20`.
- `#/properties/discount%`: explains that `%` needs two hexadecimal digits. Use `%25` if the property name contains a literal percent sign.

Malformed escapes that reach the parser now return a reference error instead of an unexpected `URI malformed` exception.

## Ticket

Related to https://github.com/scalar/scalar/discussions/8566


amritk (approved):
Error-message work, and two of the changes are genuine bug fixes rather than just better wording:

**`pointerSegments` was dropping empty segments.** `path.split('/').filter(Boolean)` discards `""`, which is a legal JSON Pointer key, so a `$ref` nested under an empty-string property could never be read back out and the message silently fell through to the generic text. `.slice(1)` drops only the leading empty from the leading `/`. The `'a/b~c%': { '': { $ref: … } }` test pins exactly that.

**`resolveUri` was leaking `URI malformed`.** Moving `getSegmentsFromPath(path)` inside the `try` turns a `decodeURIComponent` throw on `#/components/schemas/Bad%ZZ` into a normal `INVALID_REFERENCE` result. Covered for both 3.1 and 3.2.

I checked the one change that could regress: `createSpecificationValidator` no longer forwards `options` into `getValidator(version)(...)` and throws at the outer layer instead. `runValidation` only reads `options.throwOnError`, in two places — the YAML-parse branch and the validation-failure branch — and the parse branch is unreachable here because `documentToValidate` is already an object by that point (version detection runs first). So the only delta is that the thrown message is the transformed one, which is the point of the PR. Message text is unchanged for anything `transformErrors` does not touch, and the "preserves the fallback for other formats and properties" test guards that.

`explainComponentReferences` is nicely conservative — it only appends guidance when it can actually see an offending component name, decodes the fragment first so `%55ser` and `User%2Fproperties` are correctly judged as fine, bails on malformed escapes rather than guessing, and is gated on `specification.openapi` so AsyncAPI and plain JSON Schema users never see OpenAPI-specific advice. Each of those is pinned by a test.

Changesets cover all three packages. CI green.

Approving. Only note: the messages are now asserted verbatim in several tests, so any future wording tweak means touching them — fine, just the cost of testing text.

---

PR #10203, opened 2026-09-15
fix(api-reference): select response schema variants

hanspagel:
Response examples with root `oneOf` or `anyOf` schemas now offer a variant picker. Selecting Phone in the example below shows and copies the phone response instead of always using Email. Explicit media-type and schema examples keep priority; external examples continue loading lazily.

```yaml
openapi: 3.1.0
info:
  title: Response variants
  version: '1'
paths:
  /contact:
    get:
      responses:
        '200':
          description: Contact
          content:
            application/json:
              schema:
                oneOf:
                  - title: Email
                    type: object
                    properties:
                      email:
                        type: string
                        const: hello@example.com
                  - title: Phone
                    type: object
                    properties:
                      phone:
                        type: string
                        const: '+49 123 456'
```

Selection is handled by the shared schema generator, including primitive and array branches. A selection applies only once at its schema path, leaving nested unions to choose their own default. The mock server preserves generated primitive and array response shapes.

## Related PRs

#10208 has merged, and this PR is rebased onto main with its explicit-example formatting preserved. The response panel passes `getExampleContent(response, example, { contentType, compositionSelection })`, retaining the external-example pending guard. Structured and serialized examples continue to take precedence over generated variants.

## Ticket

See [discussion #6157](https://github.com/scalar/scalar/discussions/6157).


amritk (commented):
The feature is a good idea and `get-response-variants.ts` is the well-behaved half of this PR: refusing to show a picker when the schema already carries an explicit `example`/`examples`/`default`/`const`/`enum`, preserving branch indexes when a `$ref` does not resolve (so key `'1'` still means `oneOf[1]`), and defaulting to the first non-`null` branch are all the right calls, each with a test.

My concern is `get-example-content.ts`, which does schema surgery in the presentation layer:

```ts
const { oneOf: _oneOf, anyOf: _anyOf, ...base } = schema
const selectedType = variantType ?? rootType ?? ('items' in schema && !('properties' in schema) ? 'array' : undefined)
const selectedSchema = variant && selectedType !== undefined && selectedType !== 'object' ? { ...base, ...variant } : schema
if (selectedSchema !== schema) {
  if ('properties' in selectedSchema) { delete selectedSchema.properties }
  if (selectedType !== 'array' && 'items' in selectedSchema) { delete selectedSchema.items }
}
```

The mutation itself is safe — `selectedSchema !== schema` is only true on the spread branch, so the `delete`s hit a fresh copy and never the source document. That part I checked and it is fine.

What bothers me is that the comment explains exactly why this exists:

> The generator handles an explicit primitive/array type before its root union.

So this is a workaround for a limitation in `getExampleFromSchema`, and the result is that **union branch selection now lives in two places**: `compositionSelection` inside `@scalar/workspace-store`, and this hand-rolled pre-merge in `@scalar/api-reference`. The two have to agree about type inference forever, and this copy re-derives things the generator already knows how to derive (`'items' in schema && !('properties' in schema)` is inferring "this is an array" by hand).

The better fix is in `getExampleFromSchema`: make `compositionSelection` apply to root unions with a non-object branch, and delete this block. That keeps one implementation of what selecting a branch means, and every other caller of the generator gets the fix too. #10199 has just been extending the same function's selection logic for discriminators, so this is the moment to consolidate rather than fork.

If that is too large for this PR, I would at least want:

1. A comment in `getExampleFromSchema` pointing at this workaround, so the next person changing root-union handling knows there is a second implementation to keep in sync.
2. A test for the nested case. `{ ...base, ...variant }` strips the root `oneOf`/`anyOf`, but if the *selected variant* itself has a `oneOf`, it survives the spread and the same `compositionSelection` index is then applied to that nested union. That looks unintended and is not covered.

Smaller notes:

- `const composition = schema.oneOf ? 'oneOf' : 'anyOf'` is duplicated between the two helpers. Same precedence in both today, but it is the kind of thing that drifts — worth exporting one of them.
- The `watch(..., { flush: 'sync' })` resetting `selectedVariantKey` depends on `currentResponse` and `currentResponseContent` object identity. The tests cover content-type switches and tab changes, but if those computeds ever return fresh objects on unrelated document updates, a user's variant selection resets mid-read. Worth confirming identity is stable through the workspace proxy.

Happy to approve once the duplication question is settled — either by moving it into the generator, or by a clear decision to keep it here with a pointer in both directions.

---

amritk (commented):
Following up on my review — I have since read #10208 and the two PRs collide directly, which is worth knowing before either merges.

Both rewrite the same computed in `ExampleResponses.vue` (`currentContentType` here, `currentResponseContentType` there), and both add a **different third parameter to `getExampleContent`**:

```ts
// this PR

// #10208
```

Whichever lands second conflicts, and resolving it by picking one side silently drops the other feature's argument at the call site. Suggest landing #10208 first (it decides *which example value is shown at all*, which this builds on), then rebasing this one and merging the signature into a single options object — `getExampleContent(response, example, { contentType, compositionSelection })` — rather than stacking positionals.

#10208 also rewrites the `if (response?.schema)` branch of `getExampleContent` that this PR adds the variant merging to, so that hunk in particular will need a careful manual resolution rather than whatever git produces.

---

amritk (commented):
The picker itself I have no concerns about — resetting `selectedVariantKey` when the response, content type or tab changes is the right scoping, `defaultKey` skipping a `null` variant is a nice touch, and deferring to explicit examples when they exist is the correct precedence.

What stops me approving is the second half of the changeset, which is a change to the shared generator and is much wider than "a picker for response examples".

**The new pre-dispatch union branch in `get-example-from-schema.ts` affects every caller.** The block that resolves `oneOf`/`anyOf` before type dispatch for non-object unions runs for *all* example generation — request body examples, code snippets via `operation-to-har`, the mock server, AsyncAPI payloads — not just the response panel. Today a root `oneOf` of primitives or arrays with sibling keywords produces whatever the current type dispatch yields; after this it produces the merged `{...base, ...selectedVariant}` with `properties` (and sometimes `items`) deleted. That is a plausible improvement, but it is a silent output change everywhere.

Concretely, what I would want before approving:
- Confirmation of what moved in the CDN and reference e2e snapshots. If nothing moved, does the corpus actually contain root-level primitive/array unions, or is the change simply untested against real documents?
- A check on the mock server, which serves generated examples as real response bodies — a changed example there is a changed API response for anyone mocking.

**The keyword-deletion heuristic.** `delete selectedSchema.properties` unconditionally and `delete selectedSchema.items` for non-array types is described as "these keywords only constrain their own types, but also drive our type inference". That is true of our inference, but it means a schema where the root declares `properties` and a `oneOf` branch declares `type: 'string'` now silently loses the root properties. Fine for the intended case; worth a test that pins it as deliberate, because the next reader will wonder.

**`consumeCompositionSelection` is the right idea** — a selection applying once at its path and not leaking into nested unions is exactly what you want — but it is now threaded through four call sites. A test that a selection at the root does *not* alter a nested union's default choice would lock that in.

**Minor:** `getResponseVariants` bails when `schema.enum?.length` is truthy, which also bails on `enum: []`. Harmless given 10232 may start producing empty enums from `allOf` narrowing, but the two PRs interact — worth confirming the intended behaviour when both land.

---

cameronrohani (approved):

amritk (commented):
Not approving yet, mainly on sequencing.

**`mergeable_state` is `dirty`** — needs `main` merged in.

**More importantly, this PR has an undocumented dependency on #10208.** That PR's description states that it must merge first, and that this one then has to rebase onto `main` and combine response content-type and composition selection into a single options argument:

```ts
getExampleContent(response, example, { contentType, compositionSelection })
```

retaining both at the call site. That requirement is currently written down only on #10208 — nothing here records it. If these land in the wrong order, or if someone picks this one up without reading the other, the result is a silent signature mismatch rather than a merge conflict that would catch it. Worth mirroring the note into this description, or restacking this branch on top of #10208 so the ordering is enforced by git rather than by memory.

On the feature itself: a picker for `anyOf`/`oneOf` response examples is a genuinely useful answer to [discussion #6157](https://github.com/scalar/scalar/discussions/6157), and keeping explicit examples at higher priority than generated variants is the right precedence. Having copy use the selected variant rather than the first one is the detail people would otherwise file a bug about.

At 684 lines across 10 files I would want to read it properly against its final, rebased form — the composition-selection plumbing is exactly what #10208 is going to move, so reviewing the current version risks commenting on code that is about to change shape. Ping me once it is rebased and I will go through it.

---

PR #10206, opened 2026-09-15
fix(workspace-store): resolve references using OpenAPI $self

hanspagel:
Honor OpenAPI 3.2 `$self` when bundling relative references, including external documents and partial bundles. Keep OpenAPI interpretation in a workspace-store plugin, with generic document identity hooks and an explicit root URI option in json-magic.

Enable the plugin in the workspace store, mock server, and agent-chat. Preserve schema reference resolution and restore the authored reference spelling, including `./` and fragments, when exporting editable documents after partial bundling. Keep the authored `$self` unchanged. Earlier OpenAPI versions keep their existing resolution rules, and the resolved references still pass through the configured file and network loader restrictions.

## Example

```yaml
openapi: 3.2.1
$self: https://example.com/api/openapi.yaml
info:
  title: Example API
  version: 1.0.0
components:
  schemas:
    Pet:
      $ref: ./models/pet.yaml
```

Even if this document is loaded from `https://mirror.example.com/download.yaml`, the reference resolves to `https://example.com/api/models/pet.yaml` using `$self`.


amritk (commented):
The layering is the right call and worth saying so: `json-magic` stays format-agnostic with a `DocumentResolver` hook, and the OpenAPI-specific interpretation of `$self` lives in a workspace-store plugin. Stripping the fragment for the base URI while keeping it on the declared identity ("RFC 3986 requires a fragment-free base") is correct and easy to get wrong.

Two genuine drive-by fixes in here too:

- `convert-to-local-ref.ts` previously produced the literal string `"undefined/path"` whenever `schemas.get(baseUrl)` missed. The `rootPath ? … : pathOrAnchor.slice(1)` guard fixes that.
- The round-trip check in `toRelativePath` — `new URL(relativeUri, base).href === inputUrl.href ? relativeUri : input` — is a nice invariant: only relativise when relativising is lossless. That is a better formulation than trying to enumerate the cases where it is not.

**My main ask: `preserveReference` is the crux of this change and is currently unreviewable.**

```ts
preserveReference:
  value === undefined ||
  (resource.schema && !prefix) ||
  (!resource.embedded &&
    resource.schema &&
    ((prefix === uri && resource.identifier === uri) || (!prefix && !fragment.startsWith('/')))) ||
  (!resource.embedded && !resource.schema && !prefix),
```

Five disjuncts over four flags, one lone comment, and the decision determines whether a reference gets inlined or left pointing outward — which is the whole behaviour of the bundler. I cannot convince myself this is right by reading it, and neither will the next person. Pulling each clause into a named local (`const isUnresolved = …`, `const isLocalSchemaResource = …`, `const isSelfIdentifiedRoot = …`) with a sentence each would make it reviewable without changing a line of logic. A truth table in the tests over `{embedded, schema, prefix, fragment}` would do even better.

**Second: `register` mutates the caller's document.**

```ts
if (isObject(document) && identity.metadata) {
  Object.assign(document, identity.metadata)
}
```

`resolveOpenApiDocument` returns `metadata: { openapi, $self: self }` where `self` is the *resolved absolute* URI. So a document authored with `$self: './openapi.yaml'` has that field rewritten in place to `https://…/openapi.yaml`. The PR description says references are restored after processing — worth confirming that covers `$self` too, because if not, a user who loads a document and then exports it gets a different `$self` than they wrote. A test asserting the input document is unchanged after bundling would settle it (#10172 does exactly this with `expect(document['x-ext'].external).toBe(false)`).

**Third, a small robustness gap:**

```ts
if (path.length > 0) {
  bundledResources.set(path[1], resource)
}
```

The guard checks `length > 0` but the index is `1`. In practice `path` is `[externalDocumentsKey, hash]`, so this is fine today — but a single-element path would insert an `undefined` key. `path.length > 1` costs nothing.

**Finally, the changeset says `patch` for all four packages, and I think that undersells it.** This does not only add `$self` support; it changes `toRelativePath` semantics for every `@scalar/json-magic` consumer (new early return for non-HTTP parseable URIs, `baseDir` derived from a trailing slash rather than always `path.dirname`, search/hash now preserved), and rewires how the bundler indexes and resolves every reference. `@scalar/json-magic` is published, and a consumer pinning `~` would pick this up as a bugfix release. `minor` for `json-magic` at least seems more honest, and the note about the `documentUri` proxy option being new API points the same way.

One thing worth a comment rather than a change: `visited` is a per-`register` `WeakSet`, so a node reachable by two paths is indexed only at the first, and the second path's `$id`/`$anchor` never gets registered. That matches the first-path-wins convention elsewhere in the proxy, but here it silently affects anchor resolution rather than just referential identity, which is a different kind of surprise.

---

amritk (commented):
Honouring `$self` is the right thing to implement, and separating document identity into a `DocumentResolver` so the bundler does not have to know about OpenAPI is a good boundary. This is core reference resolution though, so I want to understand the blast radius before approving — the two path helpers changed here are used by every bundling path, not just `$self`.

**`resolveReferencePath` no longer means "HTTP or filesystem".** Swapping `isHttpUrl` for a generic `/^[a-z][a-z0-9+.-]*:/i` scheme test means any scheme-bearing string — `urn:`, `file:`, `mailto:`, and anything else an author writes — is now returned verbatim as an absolute reference instead of being resolved as a path. It also switches the HTTP branch from manual `pathname` surgery to `new URL(relativePath, base)`, which correctly handles `..`, query and fragment but also resolves root-relative and protocol-relative references differently than before. Which of those is intentional, and are the previously-supported cases pinned by tests?

**`toRelativePath` short-circuits on any parseable non-HTTP URI.** The new first branch returns `input` unchanged for anything `URL.canParse` accepts that is not HTTP and not a Windows drive path. That is the correct handling for a `$self` URN, but it is a broad net. Was a narrower condition (only when the input matches a known document identity) considered?

Both functions also gained trailing-slash handling, so a base ending in `/` is now treated as a directory rather than a file. That is a behaviour change for existing callers that happened to pass a trailing slash.

**`preserveReference` is four heuristics.** `isUnresolved || isLocalSchemaReference || isAbsoluteRootSchemaReference || isLocalRootDocumentReference` decides whether a `$ref` is rewritten or left alone, and getting it wrong in either direction produces a dangling reference or a lost external link — the failure mode described in the `preserveBundledExternals` comment. Each clause has a one-line rationale, which helps, but I would want a test per clause showing both the case it catches and the neighbouring case it must *not* catch.

**Practical ask:** could you run this against the openapi-parser comparison corpus and the CDN snapshots and report whether anything moved? Given how many entry points now pass `openApiDocument()` into `bundle` — `addDocument`, the export path, the lazy-resolve path and `client.ts`'s rehydration — "no snapshot moved on a few thousand real documents" is the evidence that would let me approve this quickly.

---

cameronrohani (approved):

amritk (commented):
Not approving. `mergeable_state` is `dirty`, and at 1,357 added lines across 29 files touching reference resolution in `json-magic`, `workspace-store`, the mock server and agent-chat, this needs a proper read from someone who owns the bundler rather than a sign-off from me.

The design instinct looks right, for what it is worth. `$self` exists in 3.2 precisely so a document can declare its own identity independently of where it was fetched from, and the example in the description — a document mirrored at `mirror.example.com` still resolving `./models/pet.yaml` against `https://example.com/api/` — is the case that matters. Keeping the OpenAPI interpretation of `$self` in a workspace-store plugin and exposing only generic "document identity" hooks plus an explicit root-URI option in `json-magic` is the correct layering; `json-magic` should not know what OpenAPI is.

Two things I would want a reviewer to press on:

- **Security.** `$self` lets a document redirect its own relative-reference base to an arbitrary origin. The existing loaders have deliberate guards — `blockPrivateNetworks` in `fetchUrls`, and `readFiles({ basePath })` confining local `$ref`s to the document's own directory. It is worth confirming a hostile `$self` cannot widen either of those, in particular that it cannot turn a local-file document into one that fetches, or point a file-backed base path somewhere outside its directory.
- **Partial bundles.** The description mentions these explicitly, and "restore original references after processing" is the part most likely to have edge cases — worth checking that a document with `$self` still round-trips to the references the author wrote when it is exported.

Happy to take another look once it merges cleanly, though given the size I would still defer to @DemonHa on the resolution semantics.

---

PR #10207, opened 2026-09-15
feat(api-client): parse live JSON and multipart response streams

hanspagel:
Show JSON Lines, JSON sequences, and multipart response parts as they arrive. Keep HTTP status visible, handle split UTF-8 characters, and cancel and release the reader with a visible error when a record/part exceeds 8 MiB or displayed text exceeds 16 MiB. Limits count UTF-8 bytes for JSON records and displayed text, and raw bytes including headers for multipart parts. Multipart delimiter framing is excluded from the completed part limit.

Existing finite JSON Lines, JSON sequence, `multipart/mixed`, and `multipart/x-mixed-replace` responses also use this streaming viewer. Copy text and Download text export the displayed transcript, including formatted JSON and multipart labels/base64; previews and virtualized raw-body rendering remain exclusive to the buffered viewer. Other multipart formats keep that buffered viewer.

Classification uses the shared `@scalar/helpers/http/is-streaming-content-type` helper already on main, shared with post-response script handling. The 8 MiB/16 MiB thresholds remain product tradeoffs.

file-d36cc1d57a778f3c6508910361bfcfa0

## Related

See #10178, #10176, and #10201.


amritk (commented):
A lot of this is excellent. The RFC 7464 handling in particular — requiring trailing whitespace for non-self-delimiting values, with `null` special-cased because `typeof null === 'object'` — is the kind of detail most implementations skip, and the "every possible byte split" tests for both UTF-8 and multipart framing are exactly the right way to test a push parser. Rewriting `ResponseBodyStreaming` to a plain `let currentReader` with `releaseLock()` in `finally` is also a genuine improvement over the previous reactive-ref race.

**The blocker for me is how `createMultipartParser` buffers, which I think is quadratic on realistic input.**

```ts
push: (chunk) => {
  if (closed) return
  for (const byte of chunk) {
    buffer += String.fromCharCode(byte)
  }
  drain(false)
}
```

and inside `drain`:

```ts
let offset = 0
while (!closed) {
  const index = buffer.indexOf(marker, offset)
```

`offset` resets to `0` on every `drain` call, so each chunk rescans the **entire accumulated buffer** from the start. A part delivered as 8,192 × 1 KiB chunks scans roughly 32 GiB of string in total before the closing boundary arrives — on the main thread, while the user watches. The `MAX_PART_SIZE` of 8 MiB bounds memory but not this.

Two fixes, both small:

1. **Do not rescan from zero.** Record `buffer.length` before appending and start the search at `max(0, previousLength - marker.length - 2)`, since a delimiter can only straddle the join by that much.
2. **Do not build the buffer one `String.fromCharCode` at a time.** Byte-by-byte `+=` for 8 MiB is 8 million rope nodes, and the latin1-per-byte representation costs 2 bytes of UTF-16 per input byte. Either accumulate `Uint8Array` chunks and search for the boundary over bytes, or at minimum build each chunk's string with `String.fromCharCode.apply` over slices before concatenating once.

I would want (1) regardless; (2) is what makes the 8 MiB limit actually reachable without jank.

**Two smaller performance notes:**

`ResponseBodyStreaming.vue` measures every emitted record by allocating:

```ts
displayedBytes += encoder.encode(text).byteLength
```

That is a fresh `Uint8Array` per record, discarded immediately, purely to get a length. On a high-rate JSON Lines stream that is a lot of garbage for a limit check. A cheap code-point-based length (the same arithmetic `createResponseStreamParser` already does inline for `recordBytes`) avoids it entirely — and would let both places share one helper rather than having two different byte-counting strategies in the same feature.

`btoa(body)` on a binary part inflates an 8 MiB part to ~11 MiB of string on top of the buffer copy. Within the 16 MiB display budget, but worth knowing it stacks.

**Two behaviour questions:**

**Nested parts restart their numbering.** Your own test shows the output as `Part 1 … Part 1 … hello` for a nested multipart — two `Part 1` lines at different nesting levels with nothing distinguishing them. Passing a prefix down (`Part 1.1`) or indenting nested output would make that readable.

**`type === 'multipart'` captures every multipart subtype,** including `multipart/form-data`. Any multipart response now takes the reader path and the streaming viewer instead of the buffered body viewer, which loses whatever that view offers (virtualised text for large bodies, download, preview). Same question for `application/jsonl`. That is probably an improvement, but it is a change for existing users with finite multipart responses and it is not mentioned in the changeset.

**One cross-PR note:** #10102 adds a guard in `toPostmanResponse` that skips `await response.arrayBuffer()` for `text/event-stream` so post-response scripts cannot hang on an endless body. The formats this PR adds have exactly the same property, and that guard will not match them. `getResponseStreamFormat` is the right predicate for it to use — worth coordinating so whichever lands second wires them together.

Happy to approve once the multipart buffering is addressed; the rest are notes rather than blockers.

---

amritk (commented):
The multipart parser is the most careful piece of code in this batch. Testing every possible byte split of a fixture, delivering a body one byte at a time, boundary-like text inside a part body, padded delimiter lines, nesting, and a multi-megabyte part in 1 KiB chunks — that is the right test suite for an incremental MIME parser, and it is the suite people usually do not write. The explicit `MAX_PART_SIZE`/`MAX_NESTING` limits and the boundary-syntax validation are good defensive choices too.

Two things before approving.

**The streaming path now catches much more than SSE.** `send-request.ts` changes from `contentType?.startsWith('text/event-stream')` to any `getResponseStreamFormat(contentType)`, so `application/jsonl`, `application/x-ndjson`, `application/json-seq` and `multipart/mixed` responses stop being buffered and start going through the reader. Any API already returning those types gets a different response viewer than before. That is the point of the PR, but it means the blast radius is "every existing ndjson endpoint", not "new streaming feature" — worth being explicit about in the changeset, and worth confirming the non-streaming affordances people rely on (copy, download, size display) still behave for those responses.

**Error handling.** `createMultipartParser` throws from the constructor, from `push` and from `finish` — invalid boundary, oversized part, excess nesting, missing header separator, truncated body. Where are those caught? A server that closes a multipart response early is not exotic, and the difference between "the response panel shows a readable error" and "an exception escapes the read loop" matters a lot here. If there is a catch around the reader I could not see it in this diff; if not, it needs one, with a test for the truncated case end to end rather than just at the parser.

**Smaller**

- `decodeBytes`/`toByteString` round-trip bodies through latin-1 byte strings, so an 8 MiB part allocates several full copies. Bounded by the limit, but the peak is a few times the part size — fine if intentional, worth a comment saying so.
- `multipart/form-data` and `multipart/related` are deliberately excluded from `getResponseStreamFormat` (and tested), which I assume is because they are request/aggregate formats rather than progressive ones. A one-line comment on that decision would save the next reader the trip through the spec.

---

cameronrohani (approved):

amritk (commented):
Not approving — 822 additions across 13 files implementing incremental stream parsing is squarely in the category where a careful line-by-line read by someone who owns the client beats a structural review from me. Incremental parsers are where off-by-one and buffer-boundary bugs live.

**It also sits inside the shared-helper convergence.** #10176's description names `@scalar/helpers/http/is-streaming-media-type` as the canonical classifier and says "#10207 and #10102 should import it for client and Postman classification", while #10176 and #10178 each ship their own copy. Whichever of those lands first owns the helper; this PR then needs to import it rather than carrying its own classification. Worth settling that ordering before any of the four merge.

**The specific things I would want a reviewer to verify**, since they are the classic failure modes and the description says they are handled:

- *Split UTF-8 across chunk boundaries.* The right implementation is a single `TextDecoder` with `{ stream: true }` held across reads; anything that decodes each chunk independently corrupts multi-byte characters at the seam. Worth a test that feeds a 4-byte emoji split across two chunks at every one of the three possible split points.
- *Split framing delimiters.* JSON Lines separated by `\n` and JSON Sequences framed by `\x1e` both break if a delimiter lands on a chunk boundary, and multipart boundaries can split mid-token. A test that replays one response body at every possible chunk size — 1 byte, 2, 3, … — against a known expected output catches all of these at once and is cheap to write.
- *The 8 MiB / 16 MiB limits.* Two questions: is the limit measured on decoded characters or raw bytes (they differ for non-ASCII, and the answer determines whether a legitimate CJK response trips it early), and does hitting it abort the reader and release the underlying connection rather than just stopping the UI update?

**On the limits generally:** "stop with an error if a record exceeds 8 MiB or displayed text exceeds 16 MiB" is the right instinct — an unbounded stream into a DOM buffer is a real hazard. Whether those particular numbers are right is a product call, but the error needs to be legible to the user ("response truncated at 16 MB") rather than a console message, for the same reason I raised on #10192.

Keeping the HTTP status visible while the body is still arriving is a nice touch and exactly what people want from a streaming view.

---

PR #10208, opened 2026-09-15
feat(workspace-store): share OpenAPI 3.2 example value selection

hanspagel:
Support OpenAPI 3.2 `dataValue` and `serializedValue` examples in request bodies, code snippets, and response previews. Raw editors and outgoing payloads preserve serialized text exactly; form editors prefer structured data. Explicit edits replace the authored example source, including `externalValue`, while preserving summary, description, and extensions in the exported API description.

For example, this response now displays `"hello"` as a JSON string rather than dropping its example:

```yaml
openapi: 3.2.0
info:
  title: Example fields
  version: 1.0.0
paths:
  /message:
    get:
      responses:
        '200':
          description: A message
          content:
            application/json:
              examples:
                greeting:
                  dataValue: hello
```

When `dataValue` and `serializedValue` coexist, response previews and raw editors use the serialized form while form editors retain the structured data. Downloaded `externalValue` text follows the same distinction without modifying the authored document. Legacy response `value: null` continues to display `null`.

## Visual

Response examples using the new fields

## Related PRs and merge coordination

#10192, #10189, #10178, #10177; coordinate with #10203.

Both #10208 and #10203 remain independent branches based on main. Prefer merging #10208 first, then rebase #10203. If #10203 merges first, rebase #10208 instead before merging. In either order, preserve both behaviors with `getExampleContent(response, example, { contentType, compositionSelection })`: this PR supplies `{ contentType?: string }` (default `application/json`), and #10203 adds `compositionSelection?: Record<string, number>`. The final response computed must forward both options and retain the external-example pending guard. Do not resolve the overlapping helper or computed by selecting only one side.


amritk (commented):
`get-example-value.ts` is the right abstraction and I like it a lot. One place decides the `serializedValue` → `dataValue` → `value` precedence, the `!== undefined` checks mean `false`/`0`/`null`/`''` are never mistaken for absence, and returning `{ source, value }` rather than a bare value is what lets every consumer avoid double-serializing wire text. The `externalValue` comment ("fetched by the bundler; until then it is not an inline payload") is the correct call and matches what `getExampleFromBody` already assumed.

Two things before this can merge, one mechanical and one a behaviour question.

**1. This collides head-on with #10203.**

Both PRs rewrite the same computed in `ExampleResponses.vue` — this one calls it `currentResponseContentType`, #10203 calls it `currentContentType` — and, more seriously, both add a **different third parameter to the same function**:

```ts
// #10208 (this PR)

// #10203
```

Whichever lands second will conflict, and a careless resolution could silently drop one feature's argument at the call site. These two need to be sequenced deliberately with the second rebased on the first, and the merged signature probably wants an options object (`{ contentType, compositionSelection }`) rather than a third and fourth positional. Worth deciding now which goes first — I would land this one first, since #10203's variant selection builds on top of "which example value are we even showing".

**2. `replaceExampleValue` deletes `externalValue`, and that is destructive.**

```ts
const replaceExampleValue = (example: ExampleObject, value: unknown): void => {
  delete example.serializedValue
  delete example.dataValue
  delete example.externalValue
  example.value = value
}
```

Replacing the authored source on an explicit edit is defensible — keeping a stale `serializedValue` that then wins over the user's edit would be worse. But this mutates the workspace document, so an authored `externalValue: '/examples/big-payload.json'` is gone permanently once anything calls `updateOperationRequestBodyFormValue`, and it will be gone from the document the user exports.

The `RequestBody` test asserting `expect(wrapper.emitted('update:value')).toBeUndefined()` on mount shows you already thought about incidental triggers, which is reassuring. What I would still want:

- A test that the form path (`updateOperationRequestBodyFormValue`, not just `updateOperationRequestBodyExample`) does not fire from a render or a focus, only from an actual edit. The current test covers the explicit-clear case through the other mutator.
- A line in the changeset saying an edit discards `externalValue`, since that is user-visible data loss rather than an internal detail.

Smaller notes:

- `getJsonExampleText` returns `serializedValue` before it looks at the content type at all, so serialized XML comes back verbatim for `application/xml`. That is right — it is wire text — and the test pins it. Just noting the function's name undersells what it does; `getExplicitExampleText` would read truer.
- `getExampleFromBody` now returns `{ ...example, value: selected.value }` for the data/serialized sources, allocating a fresh object per call and spreading what may be a workspace proxy. Fine at current call frequency, but worth knowing if this ends up in a render path.
- Adding `dataValue: optional(any())` to the generated schema means no validation on that field at all. Unavoidable given it is arbitrary JSON — just flagging that anything downstream has to treat it as untrusted shape.

Happy to approve once the #10203 sequencing is settled and the `externalValue` deletion is covered/documented.

---

amritk (commented):
`getExampleValue` is a good consolidation — the explicit "do not confuse `false`, `0`, `null` or `''` with absence" framing is exactly the bug class this kind of helper usually introduces, and carrying the `source` through so consumers do not re-serialize wire text is the right design. `replaceExampleValue` clearing `serializedValue`/`dataValue`/`externalValue` on edit is also correct; without it a stale serialized value would silently win over the user's edit.

Holding off on approval on one question.

**Precedence when both `dataValue` and `serializedValue` are present.** 3.2 allows an example to carry both — they describe the same example at different stages. `getExampleValue` returns `serializedValue` first, so in `RequestBody.vue` the editor shows the wire text. That is right for the raw editor, but `parsedBody` then only short-circuits on `source === 'data'`; with a serialized value it falls through to parsing the string. For `application/json` that round-trips fine, but for an XML or form-encoded body with both fields set, the form view gets a parse failure where `dataValue` was sitting right there ready to use. Should `parsedBody` prefer `dataValue` (structured, for the form) while the raw editor prefers `serializedValue` (wire text, for the textarea), rather than both going through one precedence order?

**Two smaller things**

- `getJsonExampleText` returns `serializedValue` for *any* content type, including non-JSON, despite the "Serialize only structured JSON data" comment. That is probably intended — serialized text is already in wire form whatever the media type — but the name and the comment suggest otherwise. Worth renaming or adjusting the comment.
- This changes which example wins for every consumer of `getExampleFromBody`, including code snippets and the mock server, not just the request body editor. A document that sets `dataValue` now produces a different snippet than before. Worth confirming that is intended and covered, and calling it out in the changeset.

---

cameronrohani (approved):

amritk (commented):
Not approving — three things need resolving first, none of which are about the code being wrong.

1. **`mergeable_state` is `dirty`.** This needs `main` merged in before it can go anywhere.
2. **Type checks were not run.** The description says package type checks "are blocked by missing/stale workspace dependency declarations (including workspace-store/client and oas-utils response hook types)". For a 444-line change that threads a new example-selection shape through request bodies, code samples, and response previews, an unverified `types` run is exactly the gap I cannot close by reading. Worth getting a clean local `pnpm --filter @scalar/workspace-store types:check` (and the same for the api-client side) before merge, since the reported blockage sounds like a workspace setup issue rather than a real error.
3. **The sequencing with #10203 is load-bearing and currently implicit.** This PR's description says it must merge before #10203, and that #10203 then has to rebase and fold response content-type and composition selection into a single `getExampleContent(response, example, { contentType, compositionSelection })`. #10203 is also `dirty` right now. That ordering lives only in prose on two PRs — it would be much safer as either a stacked branch or an explicit note in #10203's description too, otherwise whoever merges them in the wrong order gets a silent API mismatch at the call site rather than a conflict.

On substance, the direction is right: OpenAPI 3.2 replaces the overloaded `value` with `dataValue` and `serializedValue` precisely so tooling stops guessing whether an example is a value or its serialization, and "keep serialized text unchanged, preserve JSON value types" is the correct reading. The added coverage for explicit form edits discarding `externalValue`/`dataValue`/`serializedValue` while retaining example metadata is the subtle case worth having a test for, and documenting the exported-document effect in the changeset is the right instinct.

Happy to look again once it merges cleanly and the type check is green.

---

PR #10211, opened 2026-09-15
fix(openapi-upgrader): preserve semantics when upgrading to OpenAPI 3.2

hanspagel:
Preserve literal examples, caller-owned input, XML metadata, tag-group information, and parameter serialization when upgrading to OpenAPI 3.2. Schema migration follows OpenAPI object contexts and local references, including bundled schemas, without treating arbitrary extension or example data as schemas. References only read owned data properties, and upgrades work on an independent copy.

Explicit 3.2 migration remains strict by default: incompatibilities requiring an author's decision are reported together as `UpgradeIncompatibilityError` (an `AggregateError`) with JSON pointers. This includes unnamed inline XML elements, conflicting XML flags, repeated template variables, and optional discriminator properties without a default mapping. Analysis truncation is explicit. Malformed 3.1 versions and unsafe cyclic or disproportionate alias expansion report separate errors. The upgrader receives a minor release with migration guidance.

For example, this complete description has no inferable XML root name:

```yaml
openapi: 3.1.2
info:
  title: Pets
  version: 1.0.0
paths:
  /pets:
    get:
      responses:
        '200':
          description: A pet
          content:
            application/xml:
              schema:
                type: object
                properties:
                  id:
                    type: integer
```

Previously, the upgrader relabeled it as 3.2 without addressing the missing XML name. An explicit `upgrade(input, '3.2')` now reports the affected schema pointer, leaving input unchanged. Authors can name the XML root deliberately before migrating.

Callers that need to keep rendering can use `upgrade(input, '3.2', { onIncompatible: 'collect' })`. It returns `{ document, diagnostics }`: a 3.2 document with no diagnostics on success, or a complete 3.1 document plus all detected compatibility errors on failure. The fallback contains no partial 3.2 changes; malformed-version and clone-safety errors still throw. Swagger 2.0 and OpenAPI 3.0 inputs first convert to 3.1.

The Markdown converter and mock server use collect mode and retain the complete OpenAPI 3.1 description with its declared version, so existing descriptions continue loading without invented names, defaults, or partial transformations. Compatible descriptions still upgrade to 3.2. Malformed-version and clone-safety errors continue to surface. The workspace store still targets 3.1; a future change to 3.2 can use collect mode to preserve existing descriptions while reporting migration diagnostics.

Ambiguous tag groups retain their extension information without throwing; unambiguous groups gain native navigation parents. The extension remains to preserve renderer-specific ordering and visibility.


amritk (commented):
The craft here is high. A few things I want to call out as good before the concern, because they are easy to miss in a 400-line file:

- `resolve()` reads through `Object.getOwnPropertyDescriptor(value, key)?.value`, so a pointer cannot walk into `__proto__` or a getter, and it refuses to cross a schema resource boundary (`$id`/`$schema`) where the containing document's dialect no longer applies. That is the correct JSON Schema reading.
- `requires()` handles cycles properly: re-entry returns `undefined` rather than a guess, and `cyclicRequired` prevents caching a result that depended on the active ancestor chain. Treating `if`/`not`/`const`/`enum`/`minProperties`/`dependentRequired` as "cannot determine" rather than "not required" is the conservative direction, which is right when the consequence is rejecting a document.
- The `oneOf`/`anyOf` unanimity rule (all-true → true, all-false → false, otherwise unknown) is the only sound reading.

**The concern is that this changes `upgrade()` from a best-effort transform into a validator that throws, and one of the new checks fires on a very ordinary document shape.**

Your own test states it plainly:

```js
for (const schema of [{ type: 'object' }, { type: 'array', items: { type: 'string' }, xml: { wrapped: true } }]) {
  …
  expect(() => upgrade(input)).toThrow('An inline XML element needs an explicit xml.name.')
}
```

So this:

```yaml
content:
  application/xml:
    schema:
      type: object
      properties:
        id: { type: integer }
```

now throws. An inline object schema under an XML media type is not an exotic construct — it is what most hand-written XML responses look like. The escape hatches (`xml: { name: … }`, or `$ref` to a component) are both edits to the user's document.

I can follow the spec argument: an inline schema that is neither a component root nor a property has no name to infer, so 3.2 genuinely cannot serialize it unambiguously. What I am questioning is the *response*. An exception means the caller gets nothing — no document, no partial render, no list of what to fix. For a library whose job is "make my old document work with the new tooling", refusing the whole document because one media type is underspecified is a harsh default.

What I would want instead, roughly in order of preference:

1. An options argument — `upgrade(doc, '3.2', { onIncompatible: 'throw' | 'collect' })` — where `collect` returns `{ document, diagnostics }` and picks a defensible fallback (skip the XML rewrite for that schema, leave it as authored). Callers that want strictness opt in.
2. Failing that, at minimum aggregate: walk the whole document, collect *every* incompatibility, and throw once with all of them. Right now `fail()` throws on the first, so a user with five problems fixes them one round-trip at a time.

**This is latent today but will not stay that way.** Nothing in the repo calls `upgrade(doc, '3.2')` — `workspace-store/src/server.ts:543`, `workspace-store/src/client.ts:1065` and `mock-server/src/utils/process-openapi-document.ts:66` all target `'3.1'`. But #10177 is explicitly moving the store to 3.2 schemas, and when those call sites flip, every document with an inline XML body schema stops loading in the reference. That is the sequencing risk worth naming in this PR's description.

**Two smaller things:**

**The work budget degrades validation silently.** `requiredWork.evaluations > 100_000` is a single counter for the whole document, and once it is exhausted every subsequent `requires()` returns `undefined`, which makes the discriminator check stop firing. So whether a document is rejected depends on its size and traversal order — two documents with the same defect can get different answers. It does not poison the cache (good), but a large document silently gets weaker validation than a small one. Worth either raising it well above anything realistic, or surfacing "analysis was truncated" in the result so the behaviour is explicable.

**Dead component in the `xmlCache` key.** `validateXmlNames` returns early unless `(schema.$schema ?? dialect) === undefined`, so by the time the context key is built:

```ts
const context = [inferredName, propertyName, dialect, changedBase].join(':')
```

`dialect` is always `undefined` and contributes an empty segment. Harmless, but it reads as though the cache is dialect-aware when it cannot be.

The refactor into `clone-document` / `migrate-objects` / `migrate-tag-groups` is a clear improvement over the single file, and the documentation addition is good. My only real ask is the throw-versus-diagnostics decision — that is a product call more than a code one, so worth getting a second opinion rather than treating it as a review nit.

---

amritk (commented):
The old implementation deserved this — mutating the caller's document in place, and the `x-tagGroups` → `kind` mapping guessing from substrings like `'nav'` with a comment admitting it was simplified, were both real problems. Not approving yet because the rewrite introduces new ways for `upgrade()` to fail on documents that previously upgraded fine.

**The version test is now stricter.** `/^3\.1\.\d+$/` replaces `startsWith('3.1')`. A document declaring `openapi: "3.1"` — no patch component — used to be upgraded and is now silently returned untouched, so it flows onward still claiming 3.1 while the caller asked for 3.2. Given the store dereferences against 3.2 schemas, is silent passthrough the behaviour you want, or should a malformed version be an error? Same question for `3.1.0-rc1`-style prereleases.

**`upgrade()` can now throw where it did not.** `cloneDocument` raises on cyclic objects and on "excessive YAML alias expansion" (the `100_000` / 10× heuristic). Both are defensible guards, but they turn a previously total function into a partial one, and the thresholds are magic numbers with no stated derivation. What happens at the call sites — does the server store surface this as a failed document load, and is there a test for a document that trips the alias guard? A document that renders today and stops rendering after this lands is the regression I would want ruled out.

**The new `AggregateError`.** "Collects incompatibilities requiring an author's decision" means `upgradeFromThreeOneToThreeTwo` now rejects documents it previously converted (the old code only threw on the `wrapped` + `attribute` conflict). That is the single biggest behavioural change in the PR and it is one line of the doc comment. Which incompatibilities are in that set, how many of the corpus documents trip it, and do callers handle an `AggregateError` or does it escape as an unhandled failure?

**Double clone.** `upgrade()` clones when `targetVersion === '3.2'` and `upgradeFromThreeOneToThreeTwo` clones again, so a 3.2 upgrade walks and copies the whole document twice. Also worth noting `cloneDocument` is exported from the `3.1-to-3.2` folder but imported by the top-level `upgrade.ts`, which reads like it wants to live one level up.

**Positive:** dropping the substring-matching `kind` inference and threading `operationTags` through `migrateTagGroups` is a clear improvement — please make sure the changeset calls out that tag `kind` values may differ from before, since that is user-visible in the sidebar.

---

cameronrohani (approved):

amritk (commented):
Not approving. `mergeable_state` is `blocked`, it is 1,550 additions against 679 deletions across 15 files, and the description says the standard commands could not be run ("blocked by unavailable pnpm 12.2.1", whole-repo Knip blocked by an unrelated Vite config error). But my main concern is a design one.

**Strictness is becoming a pattern across the 3.2 upgrade path, and it deserves one decision rather than three.** This PR keeps the upgrade strict — inline XML body schemas without an inferable name still require an explicit `xml.name`, and incompatibilities are collected into an `AggregateError`. Meanwhile #10276 adds throwing for tags in multiple groups, conflicting parents, and even duplicate tag names. Individually each is defensible; together they mean a document that renders in Scalar today can stop rendering after the 3.2 migration, with an exception rather than a degraded view.

That matters more here than in most places because `upgrade()` sits on the path that arbitrary user documents take into the reference. An `AggregateError` with precise JSON pointers is a great developer experience for someone deliberately migrating their description. It is a blank page for a reader who just opened someone else's docs site. Those are different audiences hitting the same code.

The PR is candid that this is deliberate — "upgrading to 3.2 remains strict… This PR does not introduce a best-effort collect/return-document mode" — and that #10177 must handle the aggregate error before changing its target version. I think that best-effort mode is precisely what is needed before the strict path goes live for rendering, even if strict stays the default for tooling. At minimum, the sequencing constraint on #10177 should be enforced by stacking rather than by a sentence on this page.

**Genuinely good calls in here:** replacing silent weakening with an explicit truncation error when discriminator analysis is exhausted is right — silently weakening validation is the worst option of the three. Aggregating across the whole document with individual pointers, rather than failing on the first problem, is what makes a migration actually completable. And dropping the unreachable dialect component from the XML cache key is a nice bit of housekeeping.

`patch` feels light for a change that can make a previously-upgradable document throw; `minor` with a migration note would serve users better.

---

PR #10212, opened 2026-09-15
refactor(types): remove unnecessary type error suppressions

hanspagel:
## Problem

Type error suppressions hide useful compiler checks.

## Solution

Remove 20 suppressions using typed test fixtures, display metadata, configuration fields, and Vite declarations. Postman is unchanged.


amritk (commented):
The refactor itself reads well — I checked the two spots where removing a suppression could have changed behaviour and both are equivalent:

- `merge-all-of-schemas.ts`: restructuring into an early return is a no-op. The old code built `{...existing, ...incoming}` and then overwrote `.properties` with `{...existing.properties}` before merging `incoming.properties` in; the new object literal produces the identical final shape, just without needing the two suppressions.
- `api-reference-configuration.ts` (both `@scalar/schemas` and `@scalar/types`): widening the local to `Omit<..., 'showToolbar'> & Partial<Pick<..., 'showToolbar'>>` makes `delete` legal without changing that it still mutates the same parsed object. Same migration semantics.

The `DisplaySchema = SchemaObject & { nullable?: boolean }` alias is a nicer way to say what the four `@ts-expect-error We use nullable` comments were saying, and the typed cyclic fixtures in `coerce.test.ts` are a real improvement over `@ts-expect-error` on every back-reference.

Not approving yet because CI is red on the head commit (`0a4d970`) — `required-jobs-pull-requests` failed, with six jobs down:

- `deploy-previews / deploy-components`
- `deploy-previews / deploy-examples`
- `test-component-snapshots`
- `test-cdn-snapshots`
- `test-e2e-api-reference`
- `test-e2e-scalar-app`

`build`, `types`, `knip`, `lint`, `format` and all three `test-js / test-packages` shards passed, which argues against this being the type changes. The two preview deploys failing alongside the four Playwright jobs that consume them looks like the deploys took the e2e jobs down with them, so a re-run may well clear it. Worth confirming that rather than assuming — `merge-all-of-schemas.ts` and `optimize-value-for-display.ts` are both rendering-path files, so red snapshot jobs are exactly where a real regression would show up.

Happy to approve once CI is green.

---

cameronrohani (approved):

amritk (approved):
Approving. I checked each suppression removal for behavior drift and did not find any — every one is either a type-level annotation or a rearrangement that computes the same value.

The one worth calling out because it actually moves code is `mergeItemsInner`. Building `{ ...existing, ...incoming, properties: { ...existing.properties } }` inside the branch and returning early, instead of constructing `merged` up front and conditionally overwriting `.properties`, produces an identical object in both paths — but it lets TypeScript keep the narrowing from `'properties' in existing && 'properties' in incoming`, which is what the two `@ts-expect-error`s were papering over. Good trade.

The rest:

- `DisplaySchema = SchemaObject & { nullable?: boolean }` is the honest way to say what this module was already doing — the renderer tracks a null branch that is not part of the stored OpenAPI schema. It also lets `rootProperties` be destructured without `as any`, and the intersection stays assignable to `SchemaObject` for callers.
- The `Omit<typeof parsed, 'showToolbar'> & Partial<Pick<…>>` local in both configuration schemas is a neat way to make `delete` legal without a suppression, and the comment explains why the field is being dropped.
- The typed `NamedNode` / `KindA` / `KindB` fixtures in `coerce.test.ts` are a real improvement — those cyclic assignments now check rather than being waved through, which matters most in exactly the test file that is about cycles.
- `vite-env.d.ts` for `?raw` imports is the standard fix.

Nit, non-blocking: the changeset bumps `@scalar/code-highlight`, but the only change there is in `playground/`, which is not shipped. Not worth a rewrite, just noise in that package's changelog.

---

PR #10213, opened 2026-09-15
feat(api-reference): add opt-in Mermaid Markdown plugin

hanspagel:
Adds an explicitly loaded `@scalar/mermaid-plugin` package that renders fenced Mermaid diagrams throughout API Reference Markdown: introductions, operations, parameters, responses, models, summaries, and the embedded client.

The shared plugin API now exposes a browser-only Markdown rendering hook with per-render cancellation and cleanup. Mermaid loads only when an enabled plugin encounters a diagram; ordinary references retain code fences without loading it. Diagrams support zoom controls, pointer panning and keyboard panning. Invalid diagrams retain readable source, and rendering uses Mermaid’s strict security mode.

## Usage

```ts
import { mermaidPlugin } from '@scalar/mermaid-plugin'

createApiReference('#app', {
  url: '/openapi.json',
  plugins: [mermaidPlugin()],
})
```

A minimal API description demonstrating the opt-in behavior:

````yaml
openapi: 3.1.0
info:
  title: Order API
  version: 1.0.0
  description: |
    ```mermaid
    flowchart LR
      Request --> Authentication --> Response
    ```
paths: {}
````

Without the plugin, this displays the diagram source as a code block. With the plugin, it displays a diagram with pan and zoom controls. Viewer controls follow Scalar's theme; the drawing surface intentionally stays light for Mermaid's default palette and can be customized with `--scalar-mermaid-canvas-background`.

The plugin inserts SVG after Markdown sanitization and relies on Mermaid's strict-mode sanitizer for diagram output. Enabling the plugin explicitly accepts that additional sanitization dependency. Markdown hooks are provided before the embedded client modal mounts so its initial render receives the configured hooks.

The package is prepared for a separate npm release; it has not been published.

## Visual

image

## Ticket

Replaces #9962


amritk (commented):
The plumbing here is genuinely nice. `MarkdownRenderHook` is a small, well-shaped extension point, the lazy `import('mermaid')` keeps a heavy dependency out of the default bundle, and the cancellation handling is more careful than most — registering the abort listener *before* the render loop so a cancellation during a later diagram still tears down earlier ones, and re-checking `element.contains(pre)` after the await, are exactly the races this pattern usually loses. The test for "a diagram that resolves after cancellation is not inserted" is the right test to have written.

Three things before I approve.

**Hardcoded `background:#fff` on the canvas.** Everything else in `createDiagram` uses `var(--scalar-*, fallback)`, but `canvas.style.cssText` sets a bare `background:#fff`. In dark mode that is a white rectangle inside a dark panel. If the intent is that Mermaid's default theme needs a light backdrop to stay legible, that is worth a comment saying so — and ideally Mermaid's own dark theme wired to the color mode instead. Right now it reads like an oversight.

**Inline `cssText` rather than the design system.** All the styling is inline strings with hardcoded fallbacks, which sidesteps `@scalar/themes` and `@scalar/components` entirely. I understand why — the hook receives a raw `HTMLElement`, not a Vue tree — but it means these controls will not track theme changes and cannot be restyled by consumers. Is a small Vue component mounted into the container an option, or is the DOM-level API deliberate?

**`canvas.innerHTML = svg`.** The comment notes Mermaid runs with `securityLevel: 'strict'`, and that is the right mitigation. But this writes raw HTML into the DOM *after* Scalar's own Markdown sanitizer has run, so the security boundary for diagram content is now Mermaid's sanitizer rather than ours — for a hosted reference, the Markdown source is the customer's document, which is not always trusted. Opt-in makes this acceptable, but the opt-in nature and that trust assumption belong in the plugin's README and the docs, not only in a code comment.

**Minor:** moving the modal to `mountOnInitialize: false` plus an explicit `mount()` so hooks can be provided first is the right mechanism, but it changes modal init ordering for every consumer, not just Mermaid users. Worth a line in the changeset.

---

cameronrohani (approved):

amritk (commented):
Not approving — a brand-new published package plus a new public plugin hook, at 1,710 additions across 37 files, is a scope that needs an owner's review rather than mine. A few observations that I hope are useful to whoever takes it.

**The opt-in design is right, and the important part is keeping it that way.** Mermaid is large, and the value of this PR rests entirely on "ordinary references retain code fences without loading it". That claim is worth a test that asserts it mechanically — something that renders a document containing a ```mermaid fence with the plugin *absent* and fails if the Mermaid chunk is requested. Otherwise a future refactor that moves an import to the top level silently adds a very large dependency to everyone's bundle, and nothing catches it. This matters more than usual given #10222 in the same batch is already having a bundle-size argument, and #10284/#10287 are working to cut startup weight.

**`securityLevel: 'strict'` is the right choice and worth a comment saying why.** Mermaid renders user-controlled text into SVG, and API descriptions are untrusted input — a Markdown description in someone's OpenAPI document is an injection vector. Strict mode is what stops that. Since it is a one-word config that looks innocuous to delete, it deserves a note in the source explaining that it is a security control, not a preference. It would also be worth confirming the rendered SVG is inserted in a way that cannot smuggle script through Mermaid's own escaping, and that the plugin's output is sanitized on the same path as other Markdown.

**The new plugin hook is the most permanent thing here.** A "browser-only Markdown rendering hook with per-render cancellation and cleanup" becomes public API the moment it ships, and third-party plugins will depend on its exact cancellation semantics. That deserves the closest read of anything in this PR — particularly what happens when a render is cancelled mid-flight and the component unmounts, which is where these APIs leak.

The feature itself looks nicely finished: zoom, pointer and keyboard panning, invalid diagrams falling back to readable source rather than an error box, and coverage across introductions, operations, parameters, responses, models and the embedded client. Keeping it unpublished until a separate release is the right call.

---

amritk (commented):
Thanks for working through the earlier feedback. The security and opt-in docs, the themeable class-based styles and the e2e check that Mermaid is not requested without the plugin all look good. The hook's abort and cleanup semantics are well tested. A few things before this merges:

1. **Local e2e runs.** `playwright/test/mermaid.e2e.ts` is now in the default `testMatch` (`packages/api-reference/playwright.config.ts:49`). The Vite playground server is only started in CI, though (`webServer: CI ? playgroundServer : getDockerServer()`, line 61), while every other spec serves itself via `serveExample()`. So a local `pnpm test:e2e` will fail these tests unless Vite happens to be running on 5173, bound to 0.0.0.0 for the Docker browser. Could we start the playground server locally as well, or keep the spec only in `playwright.mermaid.config.ts`? Right now it lives in both configs.
2. **First publish of `@scalar/mermaid-plugin`.** The package is public and has a `minor` changeset, so the next release will bump it from 0.1.0 to 0.2.0 and `pnpm -r publish --provenance` will try to publish it. `release.yml` relies on OIDC trusted publishing with no `NPM_TOKEN`. As far as I know, that requires the package to already exist on npm with a trusted publisher configured. Can we confirm the package is bootstrapped on npm before this lands? Otherwise, marking it `"private": true` for now would keep it from failing the shared publish job. Starting at `0.0.0` would also make the first published version 0.1.0.
3. **Nit:** `pnpm-lock.yaml` has some peer-resolution churn that looks unrelated to Mermaid. Storybook entries lose their `prettier@3.8.0` peer suffix, and the nuxt/`@nuxt/kit` entries lose `magicast`. It is worth checking that this is intended, or regenerating against current main. The new `lodash-es: 4.18.1` override would also benefit from a short comment explaining why it is pinned.

---

PR #10217, opened 2026-09-16
feat(api-reference): generate AsyncAPI payload examples

hanspagel:
Generate a payload example from an AsyncAPI JSON Schema when no explicit payload example is provided. Keep authored examples and header-only examples available, with a copy button for generated payloads.


amritk (commented):
The AsyncAPI side is nicely contained — gating on "no example already supplies a payload", restricting to JSON Schema `schemaFormat` values so Avro falls through, and labelling the result **Generated example** so readers know it is synthesised. Computing it lazily behind `isExpanded` is a good call too.

My hesitation is the `deep-clone.ts` change, which is not an AsyncAPI change at all.

**`deepClone` is on the hot ingestion path.** Switching the per-key assignment to `Object.defineProperty` is correct — it stops `__proto__` and inherited accessors from being invoked — but `deepClone` is called on whole documents during client ingestion, and `Object.defineProperty` is substantially slower than plain assignment in every engine I know of. On a large document that is a per-key cost across the entire graph. Is there a benchmark? `packages/api-reference` has `test:benchmark`, and the store has `withMeasurementSync('coerceValue', ...)` wrapping a `deepClone` call, so there is somewhere to measure. If it is measurable, `key === '__proto__'` as the only special case would get the same safety with none of the cost.

Separately, this is a security-relevant fix to a shared helper riding along in an AsyncAPI feature PR. It would land faster and be easier to revert independently as its own patch.

**`hideVirtualReferenceValues`.** Walking the clone in lockstep with the proxied source to mark synthetic `$ref-value` keys non-enumerable works, but it is doing something fairly deep to hide an implementation detail of the magic proxy from the generator. Is there no way to ask the proxy for the raw shape directly — `unpackProxyObject` is already imported, so is the issue that it is shallow? Also: the recursion guards on `snapshot` identity via `seen`, but it reads `Reflect.get(source, key)` on the proxy for every key, which can materialise `$ref-value` targets. For a document with heavily shared refs, is that bounded?

**Small one:** the doc addition says "boolean payload schemas do not generate examples", but I do not see that case in `get-generated-payload-example.ts` — `getAsyncApiMessagePayloadSchema` presumably returns undefined for it. Worth a test so the documented behaviour is pinned.

---

cameronrohani (approved):

amritk (approved):
Approving. This is purely additive — generation only runs when the document provides no payload example — and the test coverage is the most careful I have seen in this batch.

Things I specifically checked:

- **The "already has an example" guard is falsy-safe.** `examples?.some((example) => getResolvedRef(example)?.payload !== undefined)` means an authored `null`, `false`, `0` or `''` payload correctly suppresses generation, and there is an `it.each` proving it. That is the exact bug #10275 is fixing elsewhere, avoided here by construction.
- **`schemaFormat` is an allowlist, not a denylist.** Avro and protobuf fall through to `undefined` rather than being fed to a JSON Schema generator that would produce nonsense. Stripping the `;version=…` suffix and lowercasing before comparison covers the real-world spellings.
- **`snapshotSchema` is doing something subtle and doing it right.** The `WeakMap` keyed on `unpackProxyObject(source)` gives cycle termination *and* shared-target identity, and the test that counts property reads on a shared schema referenced by 100 properties, asserting exactly `1`, is the right way to prove it. Cloning via `Object.defineProperty` rather than assignment preserves literal `__proto__` keys — and the test that installs a throwing setter on `Object.prototype` and asserts it is never invoked is a genuinely good way to pin that.
- **`enumerable: key !== '$ref-value' || Object.hasOwn(raw, key)`** is a neat trick: virtual reference targets stay readable by the generator but disappear from `JSON.stringify`, so literal payload data containing a `$ref` key round-trips as authored. The `it.each(['const', 'example', 'default', 'examples', 'enum'])` block covers all five keyword positions.
- **The laziness is real.** Gating on `isExpanded` in the template means the computed is never evaluated for a collapsed message, and the test asserts `MessageExamples` does not exist before the accordion is clicked. The reactivity test mutating `payload.properties.id.const` twice and checking both the rendered `<pre>` and the copy button's `content` confirms the snapshot does not stale-cache.

**Two nits, neither blocking:**

1. **`'Generated example'` is not localized.** `MessageExamples.vue` already destructures `translate` from `useLocalization()`, but the new entry is a hardcoded English literal. Worth a translation key — #10178 is dealing with the same gap for stream item labels, so it may be worth doing both together.
2. The changeset is `patch`, but this puts a new panel entry in front of every reader of an AsyncAPI document that lacks examples. `minor` would describe it better.

The documentation addition is good, particularly the honesty that generated values "illustrate the payload structure; they are not guaranteed to satisfy every JSON Schema constraint" and that Avro and boolean payload schemas are explicitly out of scope.

---

PR #10218, opened 2026-09-16
test(use-hooks): migrate to Vitest browser mode

hanspagel:
Run `@scalar/use-hooks` browser tests in Chromium instead of jsdom. Keep pure logic and server-rendering tests in Node, and install Chromium in CI.

All 65 tests, type-check, lint, and Prettier pass. Knip and the package-format hook were blocked by missing workspace dependencies.


hanspagel:
> pnpm exec playwright install --with-deps chromium

this would make CI slower :(

PR #10220, opened 2026-09-16
fix(ci): purge ESM bundles from jsDelivr cache

hanspagel:
Purge the jsDelivr cache for `esm.js` and `dist/browser/standalone.esm.js` after publishing so users get the latest release.


amritk (approved):
Both new purge targets exist in the published package — `esm.js` is in `files`, and `dist/browser/standalone.esm.js` is produced by `build:standalone:esm` — so these are real gaps being closed. Workflow-only, gated on `published == 'true'` like the existing steps.

Non-blocking nit: the existing standalone purge uses `@latest/dist/browser/standalone.js` while the two new ones omit `@latest`. jsDelivr caches the versionless and `@latest` URLs under separate keys, so whichever form consumers actually use is the one that needs purging. Worth purging both forms for all three entry points rather than leaving the set mixed.

---

PR #10221, opened 2026-09-16
chore: remove stuck proxy-scalar-com changeset

hanspagel:
`.changeset/proxy-query-method.md` targets only `proxy-scalar-com`, which is in the changesets `ignore` list. `changeset version` never consumes it, so it sits in `.changeset/` forever — `hasChangesets` stays true, and the release action keeps opening a "chore: release" PR (#10121 → #10219 → …) instead of ever running `pnpm -r publish`. That's why `main` is bumped to 1.69.0 but npm still shows 1.68.0.

Deleting the file lets the next release run find no changesets and actually publish the backlog. #10219 can be closed after this merges.

The changeset was pointless anyway: `proxy-scalar-com` is a private, ignored, separately-deployed project.


hanspagel:
bypassing the requirements to unblock the release

Issue #9789, opened 2026-07-29
perf(api-reference): cold start remains network-bound with read-only configuration


kayx23:
### What happens?

A read-only API reference has a substantial cold-start cost even with a small public OpenAPI document and the interactive features disabled.

This is distinct from the large-document rendering work tracked in #6910/#3929: the reproduction below uses Swagger Petstore, and most of the wait is downloading and evaluating the base API Reference bundle before the first operation becomes visible.

With `@scalar/api-reference@1.63.0`, the standalone script transfers about **1.06 MB compressed** from jsDelivr in this reproduction. On a fresh Chrome profile with cache disabled, 1 Mbps download throughput, 40 ms latency, and 4x CPU throttling, `Find pet by ID.` became visible after **12.6–13.1 seconds** in two runs. The Scalar script alone took 9.6–9.8 seconds to transfer.

This matters for documentation sites whose API references are mostly read-only. Disabling the client, agent, models, test requests, developer tools, telemetry, and default fonts does not materially reduce the downloaded JavaScript.

### How can we reproduce the issue?

Save this as `index.html`, serve it from any local HTTP server, then load it in a fresh browser profile with the network and CPU throttling above:

```html
<!doctype html>
<html lang="en">
  <head>
    <meta charset="utf-8">
    <meta name="viewport" content="width=device-width, initial-scale=1">
    <title>Scalar cold-start reproduction</title>
  </head>
  <body>
    <div id="app"></div>
    <script src="https://cdn.jsdelivr.net/npm/@scalar/api-reference@1.63.0"></script>
    <script>
      Scalar.createApiReference('#app', {
        agent: { disabled: true },
        hideClientButton: true,
        hideDarkModeToggle: true,
        hideModels: true,
        hideTestRequestButton: true,
        showDeveloperTools: 'never',
        telemetry: false,
        url: 'https://petstore3.swagger.io/api/v3/openapi.json',
        withDefaultFonts: false,
      })
    </script>
  </body>
</html>
```

The same base cost is observable in a Docusaurus integration, but the standalone reproduction is included to remove the host framework and any private specification from the test.

### What did you expect to happen?

A read-only configuration should have a supported delivery path with a materially smaller initial payload and faster first-operation visibility.

Possible solutions could include:

- a documented read-only/lite entry point;
- code splitting or lazy loading for API client, agent/AI, editor, and other optional features;
- build-time feature entry points that allow bundlers to exclude disabled features;
[...]

PR #10222, opened 2026-09-16
feat(api-reference): copy operations as Markdown

hanspagel:
Add **Copy as Markdown** to operations and webhooks in both layouts. The button copies the selected operation or webhook, including its schemas and security details, using the new browser entry in `@scalar/openapi-to-markdown`. Scalar-org can use it after updating API Reference.

For example, this API description previously had no operation Markdown-copy action; it now offers **Copy as Markdown** on `GET /pets` in both Modern and Classic layouts:

```yaml
openapi: 3.1.0
info:
  title: Pets
  version: 1.0.0
paths:
  /pets:
    get:
      summary: List pets
      responses:
        '200':
          description: A list of pets
```

## Bundle impact

Compared with main `546e37e922006f616b99bc9367bcd63177e07fbd`, using identical production settings and Node 24.8.0 `gzipSync` defaults (decimal kB; source maps excluded):

| Shipped JavaScript | Main gzip | This PR gzip | Delta |
| --- | ---: | ---: | ---: |
| ESM startup: entry plus all recursively imported static chunks | 737.09 kB | 738.30 kB | **+1.21 kB** |
| Single-file UMD standalone | 1,101.09 kB | 1,212.02 kB | **+110.93 kB** |

The ESM measurement sums gzip sizes per requested file, including shared chunks, rather than comparing only the entry file. The emitted static import graph and source maps confirm that the Markdown renderer is outside startup. Clicking the button dynamically loads the browser renderer chunk (32.61 kB gzip); descriptions containing HTML or GitHub alerts additionally load the HTML converter chunk (68.76 kB gzip). Existing shared dependencies remain shared. The npm ESM module also preserves `import('@scalar/openapi-to-markdown/browser')` inside the click handler; downstream bundlers determine their final chunk sizes.

UMD inlines dynamic imports into its single downloadable script, so it pays the converter cost upfront even when copying is never used. The previous approximately 102 kB estimate was against an older baseline; the rebased comparison above is the current tradeoff. Avoiding this cost for UMD would require changing its single-file delivery model or making the feature unavailable there.

The `relative-ci` size review still requires an explicit maintainer decision on accepting the UMD increase, as requested by the reviewer (@DemonHa or @hwkr). This PR does not approve or bypass that gate.

## Visual

| Layout | Before | After |
| --- | --- | --- |
| Modern | Modern before | Modern after |
| Classic | Classic before | Classic after |


cameronrohani (approved):

amritk (commented):
Not approving yet — there is an unresolved check and a bundle-size question that I think needs an explicit decision rather than my sign-off.

**CI:** `relative-ci` is sitting at `pending` / "Check failed — waiting for review", which is why `mergeable_state` is `unstable`. That is the bundle-size gate reacting to the growth this PR describes, and someone has to actively accept it before this can go in.

**The bundle-size question.** The description says the UMD bundle grows by ~102 kB gzip. That is worth weighing deliberately, because it lands in the middle of a batch of PRs pulling hard in the other direction — #10284 removes ~292 kB gzip from startup and #10287 removes ~56.5 kB, both explicitly in service of #9789. Adding 102 kB back for a copy button gives up roughly a third of what #10284 buys.

I suspect the answer is "the ESM build loads `@scalar/openapi-to-markdown` on demand, and only the UMD standalone build inlines it because that is what UMD does" — in which case the ESM number is the one that matters and the UMD growth is unavoidable. If that is right, it is worth saying so directly on the PR, along with the ESM startup delta, so the relative-ci reviewer has the number they actually need. If it is *not* right, and the markdown renderer ends up in the initial chunk for ESM consumers too, that should be fixed before this merges.

Given the `priority` label, I would rather flag this than hold it up over process — but the size gate needs a human decision either way, and @DemonHa or @hwkr are better placed than me to make the call on whether the feature earns the bytes.

The feature itself looks well scoped (both layouts, operations and webhooks, dedicated browser entry) and the before/after screenshots are appreciated.

---

PR #10223, opened 2026-09-16
fix(ci): record bundled frontend versions in integration changelogs

### LINKED ISSUE #10069: Scalar.AspNetCore changelog no longer documents bundled frontend dependency updates
### What happens?

I'm using:

```<PackageReference Include="Scalar.AspNetCore" Version="2.16.16" />```

and recently noticed a change in authentication behavior related to multiple security requirements, apparently introduced by #8888.

Previously, when multiple authentication schemes were selected, Scalar applied all of them to the request. Now, even with Multiple selected, only the scheme from the currently active authentication tab is sent.

I understand this was likely changed to be more compliant with the OpenAPI specification, where separate security requirement objects represent OR, but this behavior used to work differently.

The problem is that it's difficult to determine which frontend/API Reference version is bundled with a given Scalar.AspNetCore release.

Recent changelog entries such as:

2.16.16
2.16.15
2.16.14

contain no dependency information, while older releases included sections like:

Updated Dependencies

- @scalar/api-reference@1.49.0

I expected behavioral changes in the bundled frontend to be traceable from the Scalar.AspNetCore changelog.

Please include bundled frontend dependency versions in the changelog, or provide another documented way to determine which frontend version is included in each Scalar.AspNetCore release.

This would make changes like #8888 much easier to diagnose from the .NET consumer side.

### OpenAPI Document

_No response_

hanspagel:
Record the bundled API Reference version in .NET, Java, Docker, and Rust changelogs. Keep dependency changes out of other package changelogs.

Also track the Rust bundle dependency so frontend updates trigger Rust releases.

Validation: 11 tests, build, type check, lint, and formatting pass.

## Ticket

Fixes #10069


amritk (approved):
I verified the two things this hinges on:

- The `bundledIntegrations` filter (`@scalar/api-reference` dependency **and** a `copy:standalone` script referencing `packages/api-reference/dist/browser/standalone.js`) matches exactly docker, java, rust and the four dotnet packages. Rust was the only one missing the dependency, which is what the manifest change fixes — so the filter and the dependency addition agree.
- `dist/version.js` exists at release time: the Build step already builds `./{packages,integrations,tooling}/**` before `changesets/action` runs `pnpm release:version`, and the changelog generator was already required from `dist` before this PR. So no new build-order requirement.

One sharp edge worth knowing about rather than fixing here: `versionBundledIntegrations` runs `changeset version` first and only then checks for the release heading, so the `Missing release heading` throw leaves the tree half-versioned. That is recoverable on a bot branch, but it means a changelog-format change would surface as a failed release run rather than a clean skip.

---

PR #10226, opened 2026-09-16
perf(openapi-to-markdown): reduce large-document conversion memory

amritk:
## Summary

Large API descriptions can exhaust memory when conversion builds whole-document HTML and Markdown syntax trees. Render and convert each operation or schema sequentially, compile Vue templates for SSR, and skip redundant HTML minification on the Markdown path.

Bundle a targeted `rehype-minify-whitespace` fix that compiles constant element predicates once. The build checks that the fix is applied and includes bundled dependency licenses.

Use the existing workspace store loader unchanged from the parent PR. There is no custom reference adapter or new package dependency. The plain-loader and @amritk parser experiments are deferred.

## Measurements

Cloudflare API description, 25.49 MB, Node 24.18.0, 6,144 MiB old-space limit. Fresh sequential processes using the same fixture and optimized rendering pipeline:

| Loader | Peak RSS | Time | Output size |
| --- | ---: | ---: | ---: |
| Previous plain loader | 1.17 GB | 32.29 s | 36.95 MB |
| Restored workspace store (this PR) | 2.97 GB | 81.15 s | 34.81 MB |

Restoring the workspace store added about 1.80 GB peak RSS in this single-run comparison. Output hashes differ between loaders; this is not an output-equivalence result. Fixture and output hashes are recorded in `packages/openapi-to-markdown/benchmarks/README.md`.

The current implementation remains above the 1 GB target. Concurrent conversions require separate runner-side concurrency limits.

## Validation and remaining work

- All 10 focused reusable-renderer and large-document tests pass.
- Package build, type-check, scoped lint/format checks, and changeset status pass.
- `pnpm knip` cannot complete because the unrelated Docusaurus playground configuration requires the missing `@scalar/docusaurus/dist/index.js` build artifact.
- Full correctness testing and production validation remain deferred; keep this PR in draft.


packages/openapi-to-markdown/vite.config.ts
hanspagel, line 1: gnah, I’d prefer to keep the Vite config vanilla/simple
let me check whether we can find a less hacky way to optimize the perf

also my plan was to offer a browser version, too (as an optional plugin for the self-hosted version)

hanspagel (changes requested):

hanspagel:
I think I’d prefer this approach: #10233
wdyt?

PR #10229, opened 2026-09-17
fix(api-reference): Test Request modal shows no operation

kou029w:
## Summary

Opening the API client modal for a document's first operation, right after `changeSelectedDocument` ran again for any reason, showed "Select an operation to view details" instead of the real request.

`changeSelectedDocument` re-syncs the active document by calling `apiClient.value.route({ documentSlug: slug, method: 'get', path: '/' })` on every navigation. `path: '/'` and `method: 'get'` were placeholders. The comment next to the call even said they did not matter, because something else would set the real operation before the modal opened. That assumption broke for one specific operation: the modal marks a document's first operation as selected the moment it mounts, before any click happens. Reopening an already-selected operation skips re-routing on purpose, so reopening a visible operation does not wipe out an edited request body. The placeholder route from `changeSelectedDocument` never got replaced for that operation, so the modal kept showing it.

The fix drops the placeholder. `route()` already treats a missing path or method as "resolve to the document's first operation," which is a real route instead of one that does not exist. `RoutePayload.path` and `RoutePayload.method` are now optional, so `ApiReference.vue` can pass just `documentSlug`.

## Test plan

- Added a test in `ApiReference.test.ts` that reproduces the exact trigger: mount with two document sources, wait for the modal to load, reselect the active document through the document selector (this is what calls `changeSelectedDocument` a second time, now that the modal exists), then open the client modal for the pre-selected first operation. Verified by hand that this test fails against the unfixed code, rendering "Select an operation to view details", and passes after the fix.
- `pnpm vitest` for `@scalar/api-client` (1851 tests) and `@scalar/api-reference` (2046 tests): all pass.
- `pnpm --filter @scalar/api-client types:check` and `pnpm --filter @scalar/api-reference types:check`: clean.
- `pnpm biome check` / `pnpm prettier` on the changed files: clean.
- `pnpm knip`: no new findings.


hanspagel (approved):
nice

PR #10231, opened 2026-09-17
fix(api-client): trim OAuth2 credentials and bearer tokens

hanspagel:
Trim surrounding whitespace from OAuth2 client IDs and secrets before authorization and token refresh. Also trim bearer tokens after resolving environment variables. This prevents authentication failures caused by copied spaces.

Adds regression tests and patch changesets. Basic auth credentials and API keys are unchanged.

## Ticket

See https://github.com/scalar/scalar/discussions/10227


amritk (approved):
Trimming happens at the point of use and stored secrets are left alone, which is the right split — the tests assert both. I like that the scope is deliberately narrow: bearer tokens and OAuth2 client ID/secret only, with Basic credentials and query API keys explicitly left untouched and covered by tests, since whitespace can be significant there.

Worth calling out as intended behaviour rather than a bug: a whitespace-only client secret now makes `hasClientSecret` false, so the client silently switches to the public-client path (`client_id` in the body, no Basic header) instead of sending a blank secret. That is the better outcome and the test pins it, just noting it is a behaviour change, not only a cosmetic trim.

---

PR #10232, opened 2026-09-17
fix(api-reference): intersect `allOf` enum constraints

### LINKED ISSUE #9969: allOf enum narrowing is not reflected in schema rendering
### What happens?

When an object schema uses `allOf` to narrow an inherited property enum, Scalar does not clearly display the effective narrowed constraint.

In this example, `PrincipalResourceIdentifier.type` can only be `"principals"` because both `allOf` branches apply. However, Scalar presents the array item as a generic resource identifier and does not clearly surface that only `"principals"` is accepted.

Scalar should resolve the intersection of the two `type` constraints and clearly show that `PrincipalResourceIdentifier.type` only accepts `"principals"`.

The rendered schema should not suggest that the broader `ResourceIdentifier.type` enum values are valid for this derived schema.

### OpenAPI Document

https://sandbox.scalar.com/files/p0aEu/openapi.json

hanspagel:
Intersect enum values across `allOf` branches so inherited properties show only allowed values. Preserves nested constraints and keeps enum names and descriptions matched to their values.

Checks: 743 schema tests, scoped type check, lint, formatting, and Knip pass. Full package checks hit local dependency and Node storage errors; package type errors also occur without this fix.

## Ticket

Fixes #9969


amritk (commented):
Intersecting rather than unioning is the correct reading of `allOf`, and the annotation realignment (keeping `x-enum-varnames` / `x-enum-descriptions` attached to the values that survive) is the part I would have expected to be missed — good that it is covered in both positional and value-keyed forms.

Holding off on approval for one question about the disjoint case.

**Disjoint enums now render nothing.** The updated expectation in `merge-all-of-schemas.test.ts` goes from six values to `[]`, and `{ first: ['resources'], second: ['principals'] }` gives `[]`. `SchemaEnums.vue` is gated on `v-if="enumValues.length > 0"`, so those schemas lose their enum section entirely — where today the reader at least sees the values, they will now see nothing at all.

That is *technically* right, since such a schema is unsatisfiable. But plenty of real documents write `allOf` branches with overlapping-but-not-nested enums and do not mean "no valid value", and the failure is silent: no warning, no hint that the values were dropped rather than never declared. Have you checked what this does to the CDN snapshot corpus, or to a few large public documents? If disjoint enums turn out to be common in the wild, rendering nothing is a worse outcome than the old over-permissive union, and it would be worth falling back to the un-narrowed list (or surfacing the contradiction) instead of an empty section.

**Smaller things**

- `mergePropertiesIntoResult` changing `{ allOf: [existing, ...schema.allOf] }` to `{ allOf: [existing, schema] }` is a real fix for sibling keywords, but it is a separate behavioural change from the enum work and is easy to miss in the diff. Worth a line in the changeset.
- In `mergeEnums`, when a retained value is absent from `source.enum` the annotation lookup is `annotation[-1] ?? ''`, which yields an empty varname/description rather than dropping the entry. Reachable only on mismatched input, but an empty label renders as `value = ` in `SchemaEnumPropertyItem`, which looks like a bug to a reader.

---

cameronrohani (approved):

cameronrohani (approved):

PR #10233, opened 2026-09-17
perf(openapi-to-markdown): generate Markdown directly from syntax trees

hanspagel:
Generate Markdown directly from a syntax tree to avoid Vue rendering and full-document HTML conversion. Keep the Markdown API and page selectors, and remove the HTML API.

Local benchmarks across six workloads showed 

* 1.4–9.5x faster document loading and first rendering, 
* with 26–67% lower peak memory use.


hanspagel:
still uses too much memory

amritk (approved):
This should be merged as a stack, speed improvements in this one, and memory improvements in the other

PR #10237, opened 2026-09-17
fix(api-reference): apply document environment defaults

hanspagel:
Use the API description’s default environment in the reference and Test Request client. Variable tooltips now show their values. Keep user selections, including no environment.

Added eight regression cases and checked the tooltip in a browser. All 40 targeted tests pass, including selection changes across three documents. Full checks remain limited by missing build files, network access, stale dependency types, and a Knip config error.

## Ticket

See https://github.com/scalar/scalar/discussions/10234


amritk (approved):
The override tracking is the subtle part and it is done well: the sync-flush watcher on `x-scalar-active-environment` records user intent, and `applyingDefault` keeps the hook's own writes from being mistaken for one. The test that a user choice survives an intervening document which happens to carry the same default is exactly the case a naive "compare to previous value" implementation would get wrong.

Applying it to both `workspaceStore` and `clientStore` keeps the reference and the Test Request client on the same environment, which is the actual bug. Two lines in `ApiReference.vue`, everything else in a new file.

---

PR #10238, opened 2026-09-17
fix(api-client): display vendor XML responses as text

hanspagel:
Vendor `+xml` responses showed “Binary file.” This displays them as XML text and downloads them as `.xml`, while preserving SVG and XHTML handling.

image

## Ticket

See https://github.com/scalar/scalar/discussions/10225. Complements #10192 for actual responses.


amritk (approved):
Clean, contained fix. The `+xml` fallback sits after the exact-match lookup, so `image/svg+xml` and `application/xhtml+xml` keep their specialised configs, and it mirrors the existing `+json` fallback exactly. Tests cover the preserved cases and the negative case (`+xmlish` is not treated as XML).

---

PR #10239, opened 2026-09-17
perf(openapi-to-markdown): render large documents with bounded memory

amritk:
## Summary

Large OpenAPI descriptions exceeded the production runner's memory budget during Markdown conversion. This combines the direct Markdown syntax-tree renderer from #10233 with the plain-document loading lesson from #10226, then serializes operations and schemas in bounded sections. Cloudflare's 25.49 MB API description now completes at **659 MiB peak RSS with a 6 GiB V8 heap limit** (8.8 s), below the 1 GiB target. The same output completes at **559 MiB peak RSS with a 512 MiB heap limit**.

The Markdown path no longer creates a workspace store or proxies. It applies the existing TypeBox OpenAPI coercion to plain documents, links references to shared objects, bundles external references only when needed, and caps the description cache. It also removes the whole-document `nodes.push(...operations)` call that fails on Cloudflare's operation count. The package uses the ordinary TypeScript build; there is no Vite transform or dependency patch. Node-only file loading is imported lazily, leaving the core renderer bundleable for a future browser entry. Existing HTML exports remain available through on-demand Markdown-to-HTML conversion.

## Validation

- Cloudflare fixture SHA-256: `b46b170fce0e663e2fc0cd8c12e3d64bd39ecd6d587542e972ddaf43a2fb97d8`; both heap-limit runs produced the same 31,625,730-byte Markdown output hash (`1cb8a6455ea2bddd7dcd399995e841d3c60420a7ae625f0c2156544099f49748`).
- 118 package tests pass; package build, type check, changed-file lint/format, frozen lockfile install, scoped Knip, and changeset status pass.
- A vanilla Vite library-mode browser bundle smoke build succeeds. Browser runtime behavior and the full production runner remain to be validated before merging.
- Full-repository Knip cannot load the unrelated Docusaurus playground because its `@scalar/docusaurus/dist/index.js` build artifact is missing.

## Notes

The direct renderer changes Markdown whitespace/escaping in some cases. Coercion also normalizes fields and changes the Cloudflare output size relative to the previous plain-loader draft. Plain reference links also populate recursive example fields that the legacy workspace-store output left `null`; the compatibility test preserves the surrounding document structure while checking this change explicitly. HTML is now generated from Markdown on demand, so its markup can differ from the former Vue-rendered HTML.


amritk:
going to put em up as a staack

PR #10241, opened 2026-09-17
perf(openapi-to-markdown): review lean memory changes on direct renderer

amritk:
This draft PR is a focused comparison of [#10239](https://github.com/scalar/scalar/pull/10239) against the direct Markdown renderer in [#10233](https://github.com/scalar/scalar/pull/10233). Its base is `codex/openapi-markdown-ast`, so the Files changed tab shows only the later memory work.

The added loader keeps OpenAPI coercion and external-reference bundling while avoiding workspace-store instances and proxies. Rendering serializes operations and schemas in bounded sections, caps the description cache, and avoids the large `nodes.push(...operations)` argument-limit failure. The existing HTML API remains available through on-demand Markdown conversion. This also removes the large legacy Markdown golden fixture; its replacement test covers recursive schemas and examples.

On the 25.49 MB Cloudflare description, the resulting branch measured 659 MiB peak RSS with a 6 GiB V8 heap limit and 559 MiB with a 512 MiB limit. The isolated stack passes 118 package tests, package build, and type check.

This is a review-only stack. #10239 remains the draft PR against `main`.

PR #10251, opened 2026-09-18
refactor(openapi-to-markdown): remove HTML output and preserve chained operations

hanspagel:
Chained Path Item references can lose their final operations when an intermediate reference link is non-enumerable. Preserve those own reference links while merging siblings, so operations remain present in generated Markdown. Inherited reference links are not promoted into the resolved path item.

For example, this API description now includes `GET /pets` in the output:

```yaml
openapi: 3.1.0
info:
  title: Pets
  version: 1.0.0
paths:
  /pets:
    $ref: '#/components/pathItems/Alias'
components:
  pathItems:
    Alias:
      $ref: '#/components/pathItems/Target'
    Target:
      get:
        responses:
          '200':
            description: OK
```

Remove HTML output from the package, including the public `createHtmlFromOpenApi` and `renderer.renderHtml` APIs and their output-only dependencies. Callers needing HTML must use `createMarkdownFromOpenApi` or `renderer.render` and convert the resulting Markdown in their application. Embedded HTML in descriptions still converts to Markdown.

This removes public APIs and is marked for a minor release (1.1.0).

## Ticket

See #10241


cameronrohani (approved):

amritk (approved):
Approving. Two small, well-targeted fixes with regression tests that pin the actual mechanism rather than just the symptom.

**The chained-`$ref` fix is the interesting one.** `hasUnfollowedRef` detects a pending hop by checking for an own `$ref-value` key, but `mergeSiblingReferences` built its result with object spread — and spread only copies *enumerable* own properties. For plain (non-proxy) document loaders that define `$ref-value` non-enumerably, the link was silently dropped at the first merge, so a chain like `/pets` → `Alias` → `Target` stopped after one hop and `Target`'s operations disappeared. Re-establishing the link with `Object.defineProperty(..., { enumerable: false })` fixes it without making the internal key serializable, which is the right constraint.

The tests are good:

- The in-memory case asserts the full chain resolves *and* that sibling overrides win at every hop — `Alias get` beats `Original get`, `post` comes through from `Target`, and the outer `parameters` override the inner ones.
- It explicitly re-asserts `enumerable === false` on both links afterwards, so a future refactor back to spread fails loudly.
- The cycle test checks termination, that `$ref-value` does not leak into the resolved object, and that exactly one warning is emitted.
- The external-file case covers the bundled path separately, which is where the non-enumerable form actually comes from.

**The GFM fix** is a one-liner and `remark-gfm` is already a declared dependency of `@scalar/openapi-to-markdown`, so nothing is missing from `package.json`. Asserting `createHtmlFromOpenApi(...) === renderer.renderHtml(...)` is a nice way to keep the two entry points from drifting apart again.

Nit: `'$ref-value' in target` walks the prototype chain, while the very next clause uses `Object.hasOwn(merged, …)`. Using `Object.hasOwn` on both sides would be consistent and marginally safer. Not blocking.

---

PR #10253, opened 2026-09-18
fix(api-reference): account for sticky headers when scrolling to anchors

AshishKumarPathak:
# fix(api-reference): account for sticky headers when scrolling to anchors

## Problem

When a host app embeds Scalar inside a layout that has a sticky or fixed navigation bar (e.g. a top nav + tab bar), deep-linking to a section anchor via a copied URL would scroll the target heading behind the sticky bar. The heading appeared visually truncated or completely hidden, even though the scroll had technically completed.

This affected every scroll-to-anchor path:
- Initial deep link on page load
- The retry loop while lazy content renders
- The freeze loop that holds position during lazy rendering

## Solution

Added two focused helpers to `lazy-bus.ts`:

- `getStickyHeaderOffset()` — queries all `sticky`/`fixed` elements whose bounding rect top is within 4px of the viewport top, and returns the tallest height among them
- `scrollToElement(el)` — uses `window.scrollTo({ top, behavior: 'instant' })` with the computed offset subtracted when a sticky header is present, falling back to `el.scrollIntoView({ block: 'start' })` when there is none

Both `tryScroll` and `freeze` now call `scrollToElement` instead of the bare `scrollIntoView({ block: 'start' })`, so the fix covers all three scroll paths consistently.

Alternative considered: passing the offset in as a configuration option. Rejected because it would require every caller to know about the host app's layout — querying the DOM directly keeps the fix self-contained and zero-config.

## Checklist

- [x] I added a changeset.
- [x] I added tests.


hanspagel (approved):
cool!

PR #10254, opened 2026-09-18
feat(api-reference): add per-surface typography CSS custom properties

AshishKumarPathak:
## Problem

Scalar's typography is controlled by a single set of global `--scalar-*` design tokens. There is no way for a host application to independently adjust text size, weight, colour, or letter-spacing for individual surfaces (sidebar badges, operation titles, server selector, tag headings, etc.) without overriding internal implementation classes — which is fragile and breaks across upgrades.

## Solution

Introduces a new `packages/api-reference/src/styles/surface-typography.css` file that adds scoped CSS custom property tokens for each distinct surface:

| Token namespace | Surface |
|---|---|
| `--refs-badge-*` | HTTP method badges (sidebar + body pane) |
| `--refs-body-*` | Operation title, path, section headings |
| `--refs-server-*` | Server selector label and URL text |
| `--refs-sidebar-*` | Sidebar item labels |
| `--refs-tag-*` | Tag section headings |

All tokens fall back to the existing `--scalar-*` design tokens, so nothing changes visually unless a consumer explicitly sets a token. The change is purely additive.

**Example usage in a host app:**

```css
.scalar-app {
  --refs-badge-font-size:      11px;
  --refs-badge-color-post:     #00b050;
  --refs-body-title-font-size: 28px;
  --refs-tag-font-size:        22px;
  --refs-sidebar-font-size:    13px;
}


hanspagel:
can you show me a screenshot of what you're trying to fix?

PR #10255, opened 2026-09-18
fix(api-client): auto-enable optional params with pre-populated values

AshishKumarPathak:
# fix(api-client): auto-enable optional params with pre-populated values

## Problem

Optional parameters (headers, query params, cookies) are disabled by default in the Test Request panel. When the OpenAPI spec provides a pre-populated value for such a parameter — for example `x-scenario-id` with an enum default like `200_createEnrollment_success` — the row renders with its checkbox **unchecked** even though a value is already selected. The parameter is silently dropped from every request until the user manually checks it.

## Solution

In `createParameterRows`, compute an `effectiveIsDisabled` flag: if a parameter is only disabled by default (no explicit `x-disabled: true`) **and** already carries a non-empty value, start it enabled.

```ts
const hasValue = example?.value !== undefined && example.value !== '' && example.value !== null
const isDisabledByDefault = isDisabled && example?.['x-disabled'] === undefined
const effectiveIsDisabled = isDisabledByDefault && hasValue ? false : isDisabled
```

This mirrors the existing behaviour in `RequestTableRow.vue` where typing a value into an empty optional row auto-checks it — the fix applies the same logic at row construction time for spec-provided values.

**Invariants preserved:**
- `isDisabledByDefault` is still passed as `true` so the user can manually uncheck the row and the type-to-enable logic keeps working
- Explicit `x-disabled: true` is fully respected — those rows stay disabled regardless of value
- Required parameters and path parameters are unaffected

**Files changed:**
- `packages/api-client/src/v2/blocks/request-block/helpers/create-parameter-rows.ts` — fix
- `packages/api-client/src/v2/blocks/request-block/helpers/create-parameter-rows.test.ts` — regression test for the `x-scenario-id` case
- `.changeset/fix-auto-enable-header-with-value.md` — patch changeset

## Checklist

- [x] I added a changeset.
- [x] I added tests.


hanspagel (approved):
👍 👍 👍 👍

PR #10256, opened 2026-09-18
docs(sdks): document pagination configuration and generated helpers

amritk:
## Problem

A customer asked how to get pagination helpers for an endpoint that takes `offset` and `size` query parameters. That question could not be answered from the SDK docs.

Pagination had one section in `configuration.md`: a single `cursor` example and a list of type names. Missing were an offset example, the field-role vocabulary (7 request roles, 11 response roles, 4 response locations), how `paginated` binds a method, when paging stops, and what any target actually generates. `fakePage` was a supported type the docs never mentioned.

The overview page also overstated support: it claimed auto-paginating iterators across ten schemes including `Link` header and header token. Generated pages read their fields from the response body only, so a header-located cursor never advances a page.

## Solution

New `documentation/guides/sdks/pagination.md`:

- offset/limit quickstart with TypeScript, Python and Go usage
- declaring schemes in config or via `x-scalar-pagination`
- the `paginated` binding table, including that `true` is a no-op outside an imported config transform
- full scheme, request-field and response-field reference
- one section per strategy with how the next request is built and when paging stops
- the generated page surface per target
- the three `Pagination/*` diagnostics

It also states the uneven parts rather than implying uniform support: `cursorUrl` only follows the URL in Rust, `cursorPath`/`itemCursor` are read by a subset of targets, a `header`-located paging parameter is advanced as a query parameter, and Swift emits no pagination.

Supporting edits: `configuration.md` gains the `paginated` example and `fakePage` and links out; `openapi-extensions.md`, `diagnostics.md` and `getting-started.md` gain cross-links; the overview's networking bullet now lists the strategies that do auto-page.

Every claim was checked against the generator rather than inferred from the existing docs.

### Two things worth a reviewer's attention

**The nav entries in `scalar.config.json`.** The navigation is an explicit tree, not a directory scan, so the new page needed registering or it would have shipped unreachable. While adding it I found that `documentation/guides/sdks/openapi-extensions.md` has never had an entry either — `configuration.md` already links to it and now the pagination guide does too, and both links resolve to nothing. I registered both pages. Drop the second entry if that page is meant to stay hidden.

**The `index.md` bullet** is the one change here that alters positioning copy rather than adding reference material. Reword it however you prefer; the point is only that `Link` header and header token do not auto-page today.

## Checklist

- [ ] I added a changeset. — not applicable, no package changes
- [ ] I added tests. — not applicable, documentation only


amritk:
`required-jobs-pull-requests` failed on `ae0a9e7`. That is not a failure in this diff, and no fix is needed — recording why so nobody re-investigates it.

`ae0a9e7`'s PR run ([2492](https://github.com/scalar/scalar/actions/runs/35379614348)) has conclusion `cancelled`. Pushing `d8ac801` a few minutes later tripped the workflow's `cancel-in-progress`, which cancelled `build`, `knip`, `types`, `test-integrations`, `test-server-side-rendering`, `test-proxy-server` and `test-e2e-scalar-app` mid-flight. The aggregator gate fails on `contains(needs.*.result, 'cancelled')` as well as on `failure`, so it exited 1 for the abandoned commit. Everything that finished before the cancellation was green: `format`, `check-changes`, `lint`, `dependency-review`, `Verify generated docs PR`.

The current head `d8ac801` is running as [2493](https://github.com/scalar/scalar/actions/runs/35380161055) and will report its own `required-jobs-pull-requests`.

I am deliberately not re-running 2492: it gates a superseded commit, so a re-run would spend CI on a tree that is no longer the PR head without changing any status here. I am also not pushing anything to clear it, because a third push would cancel 2493 and reproduce exactly this failure.


---

amritk:
`test-cdn-snapshots` reports 1 failed / 25 passed on `d8ac801` — `Diff with CDN - Valtown`. Not this PR's, and not something I can fix from here.

This branch changes seven files: six Markdown pages under `documentation/guides/sdks/` and `scalar.config.json`. No source, no package manifests, no lockfile. There is no path from a docs page or the docs-site navigation tree to the `@scalar/api-reference` bundle, so nothing here can change how that bundle renders the Val Town document.

The check compares this PR's build against the **latest CDN release**, which its own footer notes makes it sensitive to changes already merged to `main` but not yet published. That is the likely cause, and it sits upstream of this branch.

It also does not gate the PR: `test-cdn-snapshots` is not among the ten jobs in `required-jobs-pull-requests`'s `needs`, matching the "non-blocking" note the report carries.

Not re-running it. A re-run would spend an 8-vCPU Playwright job to re-render the same unrelated drift, and with no source files in the diff there is nothing a second run could isolate. If someone recognises the Valtown diff as real and unreleased, it wants its own PR rather than this one.


---

amritk:
Final CI state on `d8ac801`, and a correction to my comment above.

**`required-jobs-pull-requests` passed.** All ten required jobs are green: `build`, `format`, `knip`, `types`, `test-js` (4 shards), `test-integrations`, `test-server-side-rendering`, `test-proxy-server`, `test-e2e-scalar-app`, `verify-generated-docs`.

**Correction:** `test-cdn-snapshots` finished **green**. Its `Run snapshot tests` step succeeded — the Valtown test failed on a first attempt and passed on retry, and the report comment reflects that first attempt. So there is no failing CDN check to reconcile; my note above overstated it.

**Why the run reads `cancelled`.** [Run 2493](https://github.com/scalar/scalar/actions/runs/35380161055) is marked cancelled because of `test-e2e-api-reference`, and specifically its `Deploy Playwright Report to Netlify` step, which hung for ~12 minutes and was cancelled. Every test step in that job passed first: storybook build, storybook snapshots, and the e2e tests. Report publishing, not a test failure — and that job is not in the required gate either.

So: gate green, no merge conflict, `mergeable_state: blocked` only for want of review. Nothing further from me unless CI, the base, or a review changes.


---

marclave (approved):

PR #10260, opened 2026-09-19
perf(api-reference): reuse schema walks across rows and pages

geoffgscott:
Rendering an operation whose schemas share a deep component graph re-merges and re-partitions the same schemas for every row and every page. On a 100-operation document over 150 shared schemas (allOf/oneOf, 7 levels), a harness that walks each operation the way `Schema`, `SchemaProperty` and `SchemaComposition` do went from 429 ms to 108 ms per operation.

A plain WeakMap on `mergeAllOfSchemas` hit zero times in 7,400 lookups, because `resolve.schema` and `optimizeValueForDisplay` hand back a new object on every call, so no two rows ever present the same node. The change therefore makes identity stable first (`resolveSchemaNode`, and `optimizeValueForDisplay` sharing its result), then caches the merge, the allOf partition and the display value on it.

These components also render inside the client, where documents are edited in place and `unwrap-for-read` strips reactivity from schema subtrees, so every cache entry is validated against a signature read from the live document on each lookup, and only the outermost call of a walk is cached. Mutation tests cover an edit inside an allOf member, behind a `$ref`, to a reference's siblings, and a replaced target.

Also: `rehypeHighlight` built a lowlight registry per instantiation (0.84 ms for the standard 51 grammars, 1.9 ms for all 192); it now keeps one per `languages` object.

Draft because `geoff/workspace-store-render-perf` is looking at giving `$ref` resolution stable identity at the source, which would make `resolveSchemaNode` mostly unnecessary. Not run locally: the Playwright component and storybook snapshot suites. `optimizeValueForDisplay` now returns the same object to every caller; all callers were checked to be read-only.


geoffgscott:
Closing for now. This sits on top of scalar/scalar#10261, which changes what resolving a reference costs and adds `getDocumentRevision`. Once that is released this gets rebased, re-measured and reopened; the branch `geoff/reference-render-perf` stays.

PR #10261, opened 2026-09-19
perf(workspace-store): cheaper schema walks and chained reference resolution

geoffgscott:
Four commits, each with its own changeset.

**Cheaper reads.** The detect-changes proxy allocated `{ ...args, path: [...path, prop] }` on every property read. It now returns primitives and cached child proxies first and keeps the path as a parent-link chain, materialized only when a `set` or `deleteProperty` hook fires; both hooks still get the same array instance, which `client.ts` relies on. `getResolvedRefDeep` strips proxies without the per-property read and write, `getExampleFromSchema` computes its options key once per call tree, and the typebox schema `resolve.schema` cast against is built once instead of per call. On a 100-operation document over 150 shared schemas, walking each operation the way a render does went from 188 ms to 53 ms per operation on a client store, 14.8 ms to 9.9 ms on a plain magic-proxied document. The proxy `get` trap fell from 43% of the profile to 7%.

**Reference chains.** After `resolve()` loads a chunk, a component stub stays in the document as `{ $ref: '#/x-ext/<hash>', $global: true }`, so a property pointing at `#/components/schemas/User` resolves in two hops and `getResolvedRef` followed one: consumers received a stub with no `type` or `properties`. `getResolvedRef` (both transforms) and `getResolvedRefDeep` now follow a chain through pass-through references, a `$ref` with only `$ref-value`, `$global` and `$status` beside it, with a cycle guard. A reference carrying `$id`, `$defs` or a `$dynamicAnchor` is a schema resource and is not collapsed, which keeps `$dynamicRef` scoping intact. The static round-trip test in `client.test.ts` fails on main with exactly that stub and passes here; the `multiple ref depth` todo block is enabled.

**Document revision.** `getDocumentRevision(document)` returns a number the store bumps on every mutation, keyed on the raw document so it reads the same through the store's proxies and through a view with them stripped, and 0 for a document no store tracks. It lets a consumer cache validate in O(1).

**Read-only resolved schema.** `resolve.schema` returns a `Readonly` schema; writes belong to the store's mutators. `vue-tsc` across api-reference and api-client is clean with it. `getResolvedRef` is left writable because the store's own mutators write through it.

Not included: returning the bare target from `mergeSiblingReferences` for identity stability. The `$ref` on a resolved value is load-bearing (cycle guards in `merge-all-of-schemas` and `partition-all-of-compositions` key on it, so does display naming and `pathItemIsEmpty`); dropping it failed 16 tests across two packages. Memoizing the merged view per reference node, validated by the revision above, gets stable identity without that and is a follow-up.

Found while testing, present on main and not fixed here: `getExampleFromSchema`'s module-level result cache serves a stale example after a schema is edited in place.


amritk (approved):
Yolo fridays

PR #10268, opened 2026-09-21
fix(workspace-store): save content parameter edits in media-type examples

### LINKED ISSUE #10267: JSON-content query parameter is omitted after enabling it when a media-type example is present
### What happens?

An optional `in: query` parameter using `content: application/json` and a media-type `example` remains absent from the actual request after checking **Include filter in request**. The checkbox is checked and the JSON value is visible, but the echo server receives no `filter` parameter.

Reproduced with the standalone CDN build of **`@scalar/api-reference@1.69.2`** in Chromium 153. The ordinary optional integer parameter in the same request works when enabled.

Steps:

1. Save the self-contained reproduction below as `repro.mjs` and run `node repro.mjs`. It needs no npm install.
2. Open the localhost URL printed by the script.
3. Click **Test Request** for `GET /search`.
4. Check **Include filter in request** and **Include limit in request**. Leave their displayed values unchanged.
5. Send the request and inspect the echoed URL or the terminal output.

Actual request received:

```text
/search?term=test&limit=5
```

The UI has both optional parameters checked, but only `limit` is sent. Editing JSON is not necessary to trigger this reproduction.

An explicitly enabled `filter` should be sent as one JSON-encoded query parameter. After URL-decoding, its value should be `{"status":null,"labels":[]}`. Disabling the parameter should omit it.

### OpenAPI Document

The complete OpenAPI 3.1 description is embedded in this small local server and served at `/openapi.json`. `/search` only echoes what it receives. The description also passes independent validation with Swagger Parser.

<details>
<summary>Self-contained reproduction: repro.mjs</summary>

```js
import { createServer } from 'node:http';

const description = {
  openapi: '3.1.0',
  info: { title: 'JSON query parameter reproduction', version: '1.0.0' },
  paths: {
    '/search': {
      get: {
        summary: 'Echo query parameters',
        operationId: 'search',
        parameters: [
          { name: 'term', in: 'query', required: true, schema: { type: 'string' }, example: 'test' },
          { name: 'limit', in: 'query', schema: { type: 'integer' }, example: 5 },
          {
            name: 'filter', in: 'query', required: false,
            content: {
              'application/json': {
                schema: {
                  type: 'object',
                  properties: {
                    status: { type: ['string', 'null'] },
                    labels: { type: 'array', items: { type: 'string' } },
                  },
                },
                example: { status: null, labels: [] },
              },
            },
          },
        ],
        responses: { '200': { description: 'Received query parameters' } },
      },
    },
  },
};

const requests = [];
const server = createServer((req, res) => {
  const url = new URL(req.url, `http://${req.headers.host}`);
  res.setHeader('Cache-Control', 'no-store');
  const json = (data) => {
    res.setHeader('Content-Type', 'application/json');
    res.end(JSON.stringify(data));
  };
  if (url.pathname === '/openapi.json') {
    return json({ ...description, servers: [{ url: url.origin }] });
  }
  if (url.pathname === '/search') {
    const received = { url: req.url, query: Object.fromEntries(url.searchParams) };
    requests.push(received);
    console.log(JSON.stringify(received));
    return json(received);
  }
  if (url.pathname === '/requests') return json(requests);
  res.setHeader('Content-Type', 'text/html');
  res.end(`<!doctype html>
<html><head><title>JSON query parameter reproduction</title></head><body>
<div id="app"></div>
<script src="https://cdn.jsdelivr.net/npm/@scalar/api-reference@1.69.2"></script>
<script>Scalar.createApiReference('#app', { url: '/openapi.json' });</script>
</body></html>`);
});
server.listen(0, '127.0.0.1', () => {
  console.log(`Open http://127.0.0.1:${server.address().port}`);
});
```

</details>

### Source investigation

The current source at `641d2792a774b2e9045bd4b9e5140fe50207685a` appears to explain the behavior:

- [`upsertOperationParameter`](https://github.com/scalar/scalar/blob/641d2792a774b2e9045bd4b9e5140fe50207685a/packages/workspace-store/src/mutators/operation/parameters.ts#L57-L87) writes the edited value and `x-disabled` to `parameter.examples[exampleKey]`, including for content-based parameters.
- [`getExample`](https://github.com/scalar/scalar/blob/641d2792a774b2e9045bd4b9e5140fe50207685a/packages/workspace-store/src/request-example/builder/helpers/get-example.ts#L45-L65) returns the media-type example before looking at those parameter-level edits.
- [`isParamDisabled`](https://github.com/scalar/scalar/blob/641d2792a774b2e9045bd4b9e5140fe50207685a/packages/workspace-store/src/request-example/builder/header/is-param-disabled.ts#L14-L33) then sees an example without the edited `x-disabled: false` and defaults the optional parameter to disabled.

Related older reports: #7863 and #7864 are closed. This reproduction focuses specifically on enabling an optional content-based parameter with an authored media-type example in 1.69.2.

lazerg:
## Problem

An optional JSON-content query parameter can remain absent from the request after checking **Include filter in request**. The editor saves the value and enabled state in parameter-level `examples`, while the request builder reads the authored media-type example instead.

## Solution

Save content-based parameter edits in `content[mediaType].examples`, where both the editor and request builder can read the value and enabled state. Schema-based parameters continue to use parameter-level examples.

Preserve singular authored examples as a `default` entry when converting to an examples map, retaining other named examples and their metadata. Previously saved parameter-level edits remain readable and move into the media type the next time the parameter is edited. Header-default filtering now reads examples through the same helper, so enabled content-based header edits still override defaults.

## Example

Load this OpenAPI document, open **Test Request**, enable `filter`, and send the request:

```yaml
openapi: 3.1.0
info:
  title: JSON query parameter
  version: 1.0.0
servers:
  - url: http://localhost:3000
paths:
  /search:
    get:
      parameters:
        - name: filter
          in: query
          required: false
          content:
            application/json:
              schema:
                type: object
                properties:
                  status:
                    type: [string, 'null']
                  labels:
                    type: array
                    items:
                      type: string
              example:
                status: null
                labels: []
      responses:
        '200':
          description: Search results
```

Before: the request is `/search` even though `filter` is checked.

After: the request includes one JSON-encoded `filter` query parameter. Its decoded value is `{"status":null,"labels":[]}`. Editing the value changes the sent JSON; disabling the parameter omits it.

## Ticket

Fixes #10267


hanspagel (approved):
great!

PR #10272, opened 2026-09-21
fix(workspace-store): preserve __proto__ component names

hanspagel:
Preserves components named `__proto__` in sparse server documents. Previously the name changed the schema dictionary’s prototype and disappeared from JSON output; it now remains an ordinary own property.

```yaml
openapi: 3.1.0
info: { title: Example, version: 1.0.0 }
components:
  schemas:
    __proto__: { type: string }
```

The component externalization helper also creates component-type keys as own properties, so an unrecognized literal `components.__proto__` key cannot change the output prototype. This is defensive handling of malformed input; `__proto__` is valid as a component name, but is not a standard component type.


cameronrohani (approved):

amritk (approved):
Approving — correct fix, tightly scoped. `result[type][name] = …` with `name === '__proto__'` goes through `[[Set]]`, which hits the `Object.prototype.__proto__` accessor and reassigns the dictionary's prototype instead of creating a property, so the component vanished from `JSON.stringify`. `Object.fromEntries` uses `CreateDataPropertyOrThrow`, so the key lands as an ordinary own property. That is the right primitive.

The test is thoughtful in a way that is easy to get wrong here: comparing `Object.entries(...)` instead of the objects directly, because a component legitimately named `constructor` shadows what `toStrictEqual` leans on, and asserting `Object.getPrototypeOf(result.schemas) === Object.prototype` to prove the prototype was not touched. Covering `__proto__`, `constructor` and `toString` across both `ssr` and `static` modes is the right matrix, and the end-to-end case through `createServerWorkspaceStore` confirms the chunk is actually retrievable at `#/api/components/schemas/__proto__`.

**Worth folding into this PR — the same bug exists one level up.** In `externalizeComponentReferences`:

```ts
Object.entries(document.components).forEach(([type, component]) => {
  …
  result[type] = Object.fromEntries(…)   // ← `type` is unguarded
})
```

`result[type] = …` is still a plain assignment, so a document with a literal `components.__proto__` key hits exactly the failure this PR is fixing, just at the component-*type* level rather than the component-*name* level. Same one-line shape of fix (build the outer object with `Object.fromEntries` too, or `Object.defineProperty`). Not blocking, since this PR is a strict improvement and that case is rarer, but it would be a shame to leave half of it.

---

PR #10274, opened 2026-09-21
feat(json-magic): add configurable join utility

hanspagel:
Move shared join logic to `@scalar/json-magic/join`. Keep OpenAPI rules, version upgrades, and component prefixes in the parser.

For example, joining these API descriptions still returns both paths:

```yaml
openapi: 3.1.0
info: { title: Users, version: 1.0.0 }
paths:
  /users:
    get:
      responses:
        '200': { description: OK }
---
openapi: 3.1.0
info: { title: Orders, version: 1.0.0 }
paths:
  /orders:
    get:
      responses:
        '200': { description: OK }
```


cameronrohani (approved):

amritk (commented):
Commenting rather than approving. The refactor looks correct to me, but it rewrites prototype-pollution-sensitive merge code in a published package *and* carries unrelated changes, which together put it past what I am willing to rubber-stamp.

**Unrelated changes in the diff.** Alongside the json-magic work this PR also contains Playwright flake fixes and snapshot re-baselines:

- `client.e2e.ts` — scoping the screenshot to the dialog and asserting the Close button
- `pattern.e2e.ts` — `document.fonts.ready` and `scrollIntoViewIfNeeded`
- three re-baselined PNGs

Notably `scalar-galaxy-header-mobile.png` is re-baselined here **and** in #10278, which suggests it is a flaky snapshot rather than a real visual change in either PR. Whichever lands second will conflict on the binary. These e2e fixes look worth having on their own — they would be much easier to evaluate as their own PR. (Also worth noting `scalar-galaxy-classic.png` and `scalar-galaxy.png` now have identical content, which follows from scoping the shot to the dialog, but means the two cases no longer distinguish anything.)

**On the refactor itself — I checked the parts that worried me and they hold up:**

- *Pollution handling changed meaningfully and, I think, for the better.* The old `mergeObjects` **dropped** `__proto__` / `constructor` / `prototype` keys outright. The new code preserves them as literal data via `Object.defineProperty` and `Object.fromEntries`, both of which use define semantics rather than `[[Set]]`, so the prototype is never touched. The tests assert exactly this: `Object.getPrototypeOf(result.document) === Object.prototype`, `Object.hasOwn(result.document, key)`, and `Reflect.get({}, 'polluted') === undefined`, across all three strategies. Preserving a legitimately named schema property called `constructor` is a real improvement over silently deleting it.
- *The narrowed `isPollutionKey` guard is correct.* It only fires where the old code fired: at `components.<type>.<name>` and at every depth under `info` (which uses `merge-by-index` and so recurses). Deeper component content is reached through `replace`, which hands off to `copy()` without consulting the strategy, which is why literal property names inside a schema survive.
- *Precedence is preserved.* Inputs are still reversed before joining, and `uniqueBy` keeps the first occurrence of the accumulated list, so tags and servers still resolve to the last original document — matching the old `mergeTags` / `mergeServers`. The updated doc comment ("retain the last input occurrence") is easy to misread as a behavior change; wording it as "the last input document wins" would be clearer.
- *No aliasing.* Every `merge` branch starts from an output-owned object and every terminal value goes through `copy()`, and there is a test asserting nested outputs are not the same references as the inputs.

The `join` module's README section and the strategy table are genuinely good documentation.

---

PR #10275, opened 2026-09-21
fix(openapi-to-markdown): preserve examples and schema details

hanspagel:
Preserve supplied examples, response headers and links, encoding settings, and schema details in Markdown.

For example, this response previously had no example in the output. It now shows `false`:

```yaml
openapi: 3.1.1
info: { title: Example, version: '1' }
paths:
  /example:
    get:
      responses:
        '200':
          description: OK
          content:
            application/json:
              example: false
```

Preserve `array` when no `items` schema is supplied, and retain nullable type unions alongside object properties and array items.

The compatibility output stops recursion at the first repeated ancestor. Its five removed 28-line blocks duplicated the same fields already rendered in their parent schema; no unique fields or descriptions were removed.

Replaces #10162 using the current renderer.


cameronrohani (approved):

amritk (commented):
Commenting rather than approving — this is a ~1,200 line rewrite across the `openapi-to-markdown` renderer (five new modules, a substantially rewritten `render-schema.ts`, and a regenerated `compatibility.md` fixture). I reviewed the structure rather than every rendering path, and a change of that size to output that downstream consumers diff against deserves a proper line-by-line read from someone who owns this package.

**One concrete thing to fix regardless: the unrelated e2e files.** This PR carries

- `packages/api-reference/test/snapshots/client.e2e.ts`
- `client.e2e.ts.snapshots/scalar-galaxy-classic.png`
- `client.e2e.ts.snapshots/scalar-galaxy.png`
- `header.e2e.ts.snapshots/scalar-galaxy-header-mobile.png`

The exact same `scalar-galaxy-header-mobile.png` re-baseline also appears in #10274 and #10278, and the `client.e2e.ts` changes also appear in #10274. Three open PRs re-baselining one binary snapshot means it is flaky, not that three unrelated changes each altered the Galaxy mobile header — and whichever two land later will conflict on the binary. Those Playwright stabilizations (scoping the screenshot to the dialog, `document.fonts.ready`, `scrollIntoViewIfNeeded`) look worth having, but as their own PR where they can actually be judged.

**On the substance**, the split into `get-markdown-examples`, `render-examples`, `render-operation-details`, and `restore-boolean-schemas` reads like a sensible decomposition, and each new module lands with its own test file, which is the right instinct. The bug in the description is real and worth fixing — `example: false` being dropped is the classic falsy-check mistake, and I can see `restore-boolean-schemas` exists to handle the `true`/`false` schema forms that a naive object walk loses.

The thing I would most want a reviewer to check closely is the renamed `compatibility.md` fixture: 16 added against 146 deleted lines means the reference output shrank a lot, and it is worth confirming every removed line was genuinely redundant rather than a regression that the regenerated fixture now blesses.

Happy to take another look if this gets split into the renderer change and the e2e stabilization.

---

PR #10276, opened 2026-09-21
fix(openapi-upgrader): migrate tag groups to OpenAPI 3.2 hierarchy

hanspagel:
The OpenAPI 3.2 upgrader previously discarded group names when converting `x-tagGroups` into tag kinds. Migrate each group into a navigation tag, connect its members with `parent`, and remove `x-tagGroups` after conversion. Preserve group/member order and existing tag metadata, and declare members that previously appeared only on operations.

When a group shares a name with an existing tag, give the group a unique name such as `Pets-group` and retain its display label in `summary`. If necessary, append `-2`, `-3`, and so on. Reserve declared tags, operation-only tags (including callbacks and webhooks), and other group names before allocating names. Operation tag references stay unchanged.

For example, this complete API description has a group and a tag both named `Pets`:

```yaml
openapi: 3.1.0
info: { title: Pets, version: 1.0.0 }
paths:
  /pets:
    get:
      tags: [Pets]
      responses:
        '200': { description: OK }
tags:
  - name: Pets
x-tagGroups:
  - name: Pets
    tags: [Pets]
```

After upgrading, `openapi` is `3.2.0`, `x-tagGroups` is absent, operation tags remain `[Pets]`, and the top-level tags are:

```yaml
tags:
  - name: Pets-group
    summary: Pets
    kind: nav
  - name: Pets
    parent: Pets-group
```

Tags belonging to multiple distinct groups, conflicting existing parents, duplicate tag declarations, and malformed groups produce a warning and retain the original tags and `x-tagGroups`. The version upgrade continues, so an unsupported grouping does not prevent the API description from rendering. A single OpenAPI 3.2 `parent` cannot represent multiple memberships without rewriting operation tags.

The upgrader copies the document root before migration, so a document reached through a prototype property is not modified in place. Tag names such as `__proto__` and `constructor` remain ordinary names.


cameronrohani (approved):

amritk (commented):
The migration itself is a clear improvement — the old code threw away group names entirely and guessed `kind` from substring matches on the group name, which is not a migration so much as a lossy relabel. Mapping groups to real `nav` tags with `parent` is the right 3.2 shape, and the collision handling (`Pets-group`, then `-2`, `-3`, reserving declared tags, operation-only tags, and other group names first) is carefully done. Idempotency being asserted — `upgrade(upgrade(x)) === upgrade(x)` — is a good sign.

I also checked whether dropping the `audience`/`badge` inference matters: nothing in `api-reference` or `workspace-store` reads `tag.kind === 'audience'` or `'badge'`, so those values were effectively write-only. That de-risks the change a lot.

**What I am not comfortable approving is the new throwing.** This function now raises on several documents that previously upgraded without complaint:

- a tag listed in two different groups
- a tag that already has a `parent` different from its group
- a non-string group member
- `x-tagGroups` present but not an array
- **duplicate names in the `tags` array** (`expected uniquely named tags`)

That last one worries me most. Duplicate tag entries are a common real-world authoring mistake, and they are completely unrelated to tag groups — but now any document that has both duplicate tags *and* an `x-tagGroups` block fails to upgrade at all. Since `upgrade()` runs on arbitrary user documents on the way into the reference, a throw here is not a helpful validation message to an author, it is a blank page for their readers. The same goes for multi-group membership: the PR is right that 3.2's single `parent` cannot represent it, but "cannot represent this perfectly" and "refuse to render the document" are different calls.

I would rather see these degrade — pick the first group and keep going, leave a tag ungrouped, skip a malformed group — with a `console.warn`, and reserve throwing for genuinely unparseable input. If throwing is the deliberate product decision, it deserves a `minor` changeset and a release note rather than a `patch`, because it can take a previously working document offline.

CI is fully green on `4819b00`, for what it is worth. `mergeable_state` currently reads `blocked`.

---

PR #10278, opened 2026-09-21
fix(openapi-upgrader): preserve example data during XML migration

hanspagel:
The 3.2 upgrader treated `xml` fields inside example data as schema metadata. This could change the data or throw an error. Only visit schema fields during XML migration.

This API description previously threw an error. It now upgrades and keeps the example unchanged:

```yaml
openapi: 3.1.0
info: { title: Example, version: 1.0.0 }
paths: {}
components:
  examples:
    sample:
      value:
        xml: { wrapped: true, attribute: true }
```


cameronrohani (approved):

amritk (commented):
The fix itself is right and I would happily approve it — but there is an unexplained file in the diff I cannot account for, so I am commenting instead.

**The thing blocking my approval:** `packages/api-reference/test/snapshots/header.e2e.ts.snapshots/scalar-galaxy-header-mobile.png` is re-baselined in this PR. Nothing about scoping XML migration to schema positions should change how the Galaxy mobile header renders. Either it is an unrelated flaky snapshot that got swept in — in which case it does not belong here — or the XML change is altering rendered output somewhere, which would mean the fix is not as contained as the description says. Worth confirming which before merge.

**On the actual change — it is a good one.** The old walker recursed over every object in the document and rewrote any `.xml` it found, so a perfectly valid example payload like `{ xml: { wrapped: true, attribute: true } }` hit the "wrapped and attribute cannot both be true" guard and threw on a document that has nothing wrong with it. Replacing blind recursion with an explicit `ObjectKind` → children map is the correct shape of fix: XML metadata only has meaning in schema position, so only schema position should be rewritten.

I walked the child map against where schemas can actually appear and did not find a hole:

- Parameter and Header share the `parameter` kind, which is right since both carry `schema` and `content`.
- `encoding.headers` is covered, which is easy to forget.
- Callbacks get their own handling for the two named levels, on both `operation` and `components`.
- The `x-` skip is correctly narrowed to `paths` and *operation-level* `responses` — the two places the spec allows extensions in an otherwise-named map — so a component schema legitimately named `x-data` still gets visited.

The tests are thorough in the way that matters here: the first `it.each` block enumerates fourteen data-bearing positions and asserts the document round-trips byte-identical apart from the version bump, and the `properties` / `patternProperties` / `$defs` / `definitions` / `dependentSchemas` case specifically checks that a *member* named `xml` or `examples` is treated as a schema rather than a keyword. The `JSON.parse(JSON.stringify(...))` trick to de-alias shared objects so every route has to be visited independently is a nice touch.

---

amritk (approved):
Following up on my earlier comment — the one thing I could not account for is now explained, so I am approving.

I said the re-baselined `scalar-galaxy-header-mobile.png` was the reason I was holding off, because a schema-scoping change in the upgrader should not alter how the Galaxy mobile header renders. Having now gone through the rest of the open PRs in this batch, that same binary is re-baselined with an **identical** before/after blob pair (`ede18f1ef7a..2e6ebb42c8f`) in #10274 and #10198 as well, and it appears again in #10275. Four unrelated PRs producing the same new baseline means it is a flaky snapshot, not a visual regression caused by any of them. So it is not evidence of a hidden problem here, and my concern was unfounded.

It is still worth cleaning up: whichever of the four lands first will leave the other three conflicting on that binary. Fixing the flake once on `main` and dropping the file from all four would save some pain. That is a batch-level issue rather than something this PR needs to solve.

The substance of my earlier review stands — scoping XML migration to schema positions via the explicit `ObjectKind` child map is the right fix, the fourteen-position round-trip test is strong coverage, and the `properties` / `patternProperties` / `$defs` / `definitions` / `dependentSchemas` cases correctly prove a member named `xml` or `examples` is treated as a schema rather than a keyword.

---

PR #10279, opened 2026-09-21
fix(workspace-store): support absolute output paths

hanspagel:
Fix workspace chunk output paths such as `/tmp/chunks` and `D:\chunks`. Absolute paths now point to the requested directory instead of being added to the working directory.


cameronrohani (approved):

amritk (approved):
Approving — a one-line fix that is obviously right. `resolve(dir)` already returns `dir` when it is absolute and joins it onto `process.cwd()` when it is not, so it subsumes the old `${cwd()}/${dir}` for the relative case while fixing the absolute one. It also drops the hardcoded forward slash, which is what broke `D:\chunks` and UNC paths.

The Windows test is a nice piece of work — stubbing `process.cwd()` to `C:\project`, swapping `path.resolve` for `path.win32.resolve`, and using a rejected `fs.mkdir` as a tripwire to assert the resolved root without actually writing anything. Covering `D:\nested output\chunks`, `\\server\share\chunks`, and plain `chunks` in one `it.each` gets absolute, UNC, and relative in a single pass, and the `syncBuiltinESMExports()` calls on both sides of the try/finally are the right way to make the builtin re-export stick and then come back.

`mergeable_state` is `clean` and the changeset is correctly scoped to `@scalar/workspace-store` as a patch.

---

PR #10280, opened 2026-09-21
perf(api-reference): use non-reactive stores during server rendering

hanspagel:
Server rendering does not need reactive stores. Use non-reactive stores automatically during SSR, while keeping browser stores reactive. Configured servers and environment defaults register with the workspace store. Loading, replacing, selecting, deleting, or importing documents automatically runs the registered synchronizers; `settle()` handles batches of direct writes. Browser stores retain synchronous Vue watchers. Repeated synchronization preserves edited server variables and explicit environment selections.

Original benchmark before the synchronization centralization: median full-page rendering of the 11 MB GitHub API description on Apple M4 with Node 24, after two warmups and across seven samples:

| Before | After | Improvement |
| ---: | ---: | ---: |
| 3,253 ms | 3,039 ms | 6.6% |

HTML matches after normalizing generated Vue style IDs. Timing ranges overlap, so the measured gain is modest. No new API or caller changes are needed. Benchmark files are excluded. This timing has not been remeasured after centralizing synchronization.

Builds on #10263.

For example, render this API description with configuration `servers: [{ url: 'https://configured.example.com' }]`:

```yaml
openapi: 3.1.0
info:
  title: Example API
  version: '1.0'
servers:
  - url: https://original.example.com
paths:
  /pets:
    get:
      responses:
        '200':
          description: Pet response
```

The rendered request URL continues to use `https://configured.example.com`. Both embedded stores avoid deep Vue reactivity during server rendering and remain reactive during browser hydration.


cameronrohani (approved):

amritk (commented):
Commenting rather than approving. Two reasons, one mechanical and one design.

**Mechanical:** `mergeable_state` is `dirty` — this needs `main` merged in. It also says it builds on #10263, so it presumably wants that to land first.

**Design — the part I would want a second opinion on.** Making the stores non-reactive under SSR means the `watch(…, { flush: 'sync' })` calls in `use-document-environment.ts` and `use-configured-servers.ts` silently stop firing on the server. The PR compensates by extracting the watcher bodies into `sync()` functions and calling them manually at two sites, each guarded by `if (isServerRendering)`:

- `addDocument` → `syncConfiguredServers()`
- `changeSelectedDocument` → `syncWorkspaceEnvironment()` / `syncClientEnvironment()`

That works for the paths the new tests cover, but it establishes an invariant that is invisible at the point where it matters: from now on, *every* future code path that loads or swaps a document on the server has to remember to call the right synchronizer, or configured servers and environment defaults quietly go missing from the rendered HTML. Nothing enforces it and nothing fails loudly — the SSR output just silently loses a server URL. That is a nasty class of regression to debug months later.

Given the measured win is 6.6% with, as the description says, overlapping timing ranges, I am not sure the ongoing hazard is worth it as written. Some options that would change my mind: having `createWorkspaceStore({ reactive: false })` expose an explicit "settle"/"flush" call that runs all registered synchronizers in one place, or a dev-mode assertion that catches a server store mutated without a following sync.

**What I did like:**

- `inject(ssrContextKey, null) !== null` is a neat detection trick, and the second test specifically proves it returns `false` during browser hydration — which is the case a naive `typeof window === 'undefined'` check would get wrong. Worth noting `ssrContextKey` is a lightly documented export, so a comment pointing at why it beats the obvious alternatives would age well.
- `ApiReference.reactivity.test.ts` asserting `isReactive(store.workspace) === false` on *both* stores, plus `getDocumentRevision(document) === 0`, is a real assertion rather than a smoke test. And the `ssr.test.ts` addition checking `html` contains `configured.example.com` but *not* `original.example.com` is exactly the regression the manual-sync approach is at risk of.

---

hanspagel:
> Given the measured win is 6.6% with, as the description says, overlapping timing ranges, I am not sure the ongoing hazard is worth it as written. 

Agree

PR #10281, opened 2026-09-21
fix(api-client): prevent duplicate parameter rows while typing

### LINKED ISSUE #10258: Typing a key in the request parameter tables duplicates the row and moves focus to the duplicate
### What happens?

In the request panel, typing a key into the empty **Key** cell under Cookies, Headers or Query Parameters immediately creates a second row with the same key. Focus lands in the duplicate, so continued typing edits the copy rather than the row that is enabled.

Steps:

1. Open any operation → **Test Request**
2. Click the empty **Key** cell under Headers
3. Type `x-demo-header`

Result — two rows, before clicking anything else:

```
[x]  x-demo-header     Value     (trash)
[ ]  x-demo-header|    Value               <- focus is here
```

One row named `x-demo-header`, with a fresh empty row beneath it.

---

`@scalar/api-reference` 1.69.1, loaded from `cdn.jsdelivr.net/npm/@scalar/api-reference/esm.js`. Chrome.

### LINKED ISSUE #10259: Trash icon on a disabled parameter row toggles the checkbox instead of removing the row
### What happens?

The delete action on a **disabled** (unchecked) row in the request parameter tables does not remove the row. The click toggles the row's enabled state instead.

Steps:

1. Open any operation → **Test Request**
2. Under Headers, add a row with key `x-demo-header`
3. Uncheck it
4. Hover the row and click the trash icon

Result: the row is still there, and its checkbox has flipped back to enabled.

Related: the trash icon only renders on the enabled row, so a disabled row often has no delete affordance at all.

The row is removed, whether or not it is enabled.

---

`@scalar/api-reference` 1.69.1, loaded from `cdn.jsdelivr.net/npm/@scalar/api-reference/esm.js`. Chrome.

hanspagel:
Typing a header, cookie, or query parameter could create duplicate rows. This keeps focus on the saved row and leaves a blank row below it. Unchecking and deleting a newly added parameter now leaves only the empty entry row. If another row arrives while a rename is pending, the editor stays attached to its original parameter.

To reproduce, open **Test Request** and type `x-demo-header` under Headers:

```json
{"openapi":"3.2.0","info":{"title":"Demo","version":"1"},"paths":{"/demo":{"get":{"responses":{"200":{"description":"OK"}}}}}}
```

## Visual

After: one saved header and a blank row

## Overlap with #10282

This is the recommended survivor of the two alternative row-identity implementations. It covers progressive header, cookie, and query renames without losing focus, value-before-key entry, delayed saves across unrelated refreshes, and the unchecked-row deletion behavior from #10282. The alternative implementation in #10282 loses focus during progressive renames. Merge only one implementation; maintainer selection and closure of the other PR remain pending.

## Ticket

Fixes #10258
Fixes #10259


cameronrohani (approved):

amritk (commented):
Holding off on approval because of a collision rather than anything wrong with the code.

**This and #10282 are two different rewrites of the same thing.** Both replace `getRowKey` with a `keyedRows` computed in `RequestTable.vue`, both rewire the same `v-for` and `@upsertRow` binding, and both add cases to `RequestTable.test.ts`. They are not complementary — they are competing designs for identical machinery:

- Here: `Symbol()` keys plus a `pendingUpdates: Map<symbol, TableRowUpsertPayload>`, matched on `name` + `value`, with explicit cleanup for rows that disappear.
- #10282: `nanoid()` keys plus a single `pendingRow` shallowRef, matched on the placeholder key plus `name`.

They also fix two symptoms of one root cause (#10258 duplicates while typing, #10259 the unchecked row surviving delete). Whichever merges first, the second will conflict hard, and resolving that conflict by hand risks landing a hybrid that neither set of tests actually covers. I would rather see one of them pick up the other's test cases and the other be closed.

**On this implementation specifically** (assuming it is the one that wins):

- The test matrix is strong — three parameter types, progressive renames (`x` → `x-demo` → `x-demo-header`) each asserting `document.activeElement` survives, the value-before-key ordering, and the "unrelated refresh lands before the save arrives" case that is exactly what the `pendingUpdates` map exists for.
- `matchesPendingUpdate` keys off `name` + `value` only. Two rows that transiently share both (two blank rows, say) could hand the key to the wrong one. The `!entry.data.sourceParameterValuePath` guard narrows it but does not close it. Probably unreachable in practice; worth a thought if you keep this version.
- `[...available].find(…)` runs twice per row inside the `map`, so it is O(n²) with a fresh array allocation each time. Irrelevant at realistic table sizes, just noting it since #10282's `Map` lookup avoids it.

Happy to approve once it is clear which of the two is the survivor.

---

PR #10282, opened 2026-09-21
fix(api-client): prevent duplicate parameter rows

### LINKED ISSUE #10259: Trash icon on a disabled parameter row toggles the checkbox instead of removing the row

hanspagel:
Adding a header could create a duplicate when you unchecked it. Deleting the unchecked row then left the enabled duplicate behind.

Keep the saved row focused and create a fresh empty row, so deleting an unchecked header removes it completely.

<details>
<summary>Reproduce</summary>

Open Test Request, add `x-demo-header`, uncheck it, then click trash.

```json
{"openapi":"3.1.0","info":{"title":"Demo","version":"1"},"paths":{"/demo":{"get":{"responses":{"200":{"description":"OK"}}}}}}
```

</details>

## Visual

After clicking trash on the unchecked header:

| Before | After |
| --- | --- |
| Before: an enabled duplicate remains | After: the header is removed |

## Related implementation

This overlaps with #10281. Prefer #10281 as the merge candidate: it also preserves focus while renaming saved parameters. The unchecked-delete regression here should accompany that implementation. Merge one implementation to avoid competing row-identity changes.

## Ticket

Fixes #10259


cameronrohani (approved):

amritk (commented):
Holding off on approval because of a collision rather than anything wrong with the code.

**This and #10281 are two different rewrites of the same thing.** Both replace `getRowKey` with a `keyedRows` computed in `RequestTable.vue`, both rewire the same `v-for` and `@upsertRow` binding, and both add cases to `RequestTable.test.ts`. They are not complementary — they are competing designs for identical machinery:

- Here: `nanoid()` keys plus a single `pendingRow` shallowRef, matched on the placeholder key plus `name`, with `getRowKey` renamed to `getRowIdentity`.
- #10281: `Symbol()` keys plus a `pendingUpdates: Map<symbol, TableRowUpsertPayload>` matched on `name` + `value`.

They are also two symptoms of one root cause (#10259 the unchecked row surviving delete, #10258 duplicates while typing). Whichever merges first, the second conflicts hard, and hand-resolving that risks a hybrid that neither test suite really covers. Better for one to absorb the other's test cases and the other to close.

**On this implementation specifically** (assuming it is the one that wins):

- The test is the sharper of the two for the delete bug — asserting `upsertRow` fires exactly twice with the same payload, then that `deleteRow` fires for index 0, then that the table falls back to a single blank row. The inline comment about the checkbox click blurring the key first is the kind of thing that stops someone "simplifying" the fix later.
- The self-limiting behaviour of `pendingRow` holds up: it is never explicitly cleared, but the `placeholder` lookup requires the previous row at that key to still have empty `name` *and* `value`, so once the key transfers to a saved row the branch cannot fire again. Worth a comment saying so, since "a ref that is set but never reset" reads like a leak at a glance.
- Map-based `previous` lookup is cheaper than #10281's repeated array spread, for whatever that is worth at these table sizes.

Happy to approve once it is clear which of the two is the survivor.

---

hanspagel:
Superseded by #10281, which covers the unchecked-delete regression and also preserves focus during progressive parameter renames. Keeping #10281 as the merge candidate.

PR #10283, opened 2026-09-21
feat(api-reference): add `expandAllParameters` option

### LINKED ISSUE #10265: Expose parameter list collapse (collapsableItems) as a public config option
### Summary

Expose the existing internal `collapsableItems` behavior as a public API-reference configuration option, so operations with many parameters can render each parameter as a collapsed one-line row (expand on click) instead of a long list of fully-expanded parameter cards.

### Problem

Some OpenAPI operations carry a very large number of query parameters. For example, the OpenSearch `GET /{index}/_search` endpoint has ~50 query parameters. Scalar renders every parameter fully expanded, so the **Query Parameters** section becomes extremely long and pushes the request body and responses far down the page, hurting scannability.

There is currently no public configuration to change this. The existing expand/collapse options only cover tags, responses, model sections, and schema *properties* — none control the parameter list itself:

- `defaultOpenAllTags`
- `expandAllResponses`
- `expandAllModelSections`
- `expandAllSchemaProperties`

### The capability already exists internally

`ParameterListItem.vue` already implements a collapsible, single-line rendering of each parameter gated on a `collapsableItems` prop (renders a `<Disclosure>` trigger per parameter). `ParameterList.vue` also declares `collapsableItems?: boolean` and forwards it to each item.

However, `OperationParameters.vue` never passes `collapsableItems` when it instantiates `ParameterList` for the query/path/header/cookie sections, and the prop is not wired to any public config, so consumers cannot turn it on.

Relevant files:
- `packages/api-reference/src/features/Operation/components/ParameterListItem.vue` — implements the `collapsableItems` disclosure rendering
- `packages/api-reference/src/features/Operation/components/ParameterList.vue` — declares and forwards `collapsableItems`
- `packages/api-reference/src/features/Operation/components/OperationParameters.vue` — does not pass `collapsableItems`, and there is no config option feeding it

### Proposed solution

Add a public configuration option (e.g. `collapseParameters` / `expandAllParameters`, matching the naming of the existing `expandAll*` options) that threads through `OperationParameters.vue` → `ParameterList.vue` → `ParameterListItem.vue` as `collapsableItems`. Default preserves current behavior (parameters expanded).

This reuses the existing rendering path rather than adding a new mechanism.

### Alternatives considered

- **DOM manipulation from the host page** (MutationObserver or a plugin `onInit` hook that toggles the parameter section). This works but relies on internal class names / heading labels and breaks on CDN updates — not maintainable.
- The plugin API (`plugins` / `pluginUrls`) only renders components for custom `x-` spec extensions and exposes lifecycle hooks; it cannot re-render or collapse a built-in section.

### Additional context

Reproducible on any spec with a high parameter count; OpenSearch `_search` is a convenient real-world example.

hanspagel:
Long parameter lists push the rest of an operation down the page.

Adds `expandAllParameters`. It defaults to `true`, keeping the current view. Set it to `false` for compact rows that open on click. Names, types, and required markers stay visible in both layouts. Parameters without details stay static instead of opening an empty panel.

<details>
<summary>Example API description</summary>

With this document, `limit` shows its details by default. With `expandAllParameters: false`, click the row to show them.

```yaml
openapi: 3.2.0
info:
  title: Search API
  version: 1.0.0
paths:
  /search:
    get:
      parameters:
        - in: query
          name: limit
          description: Maximum number of results.
          schema:
            type: integer
            minimum: 1
            maximum: 100
      responses:
        '200':
          description: Matching results
```

</details>

## Visual

Screenshot 2026-09-21 at 16 09 44

## Ticket

Fixes #10265


cameronrohani (approved):

amritk (approved):
Approving. `default: true` in both `@scalar/schemas` and `@scalar/types` means existing references render identically, so the risk is contained to people who explicitly opt in.

I traced the plumbing to make sure the option is not dead: `runtimeConfig` is the fully coerced `ApiReferenceConfiguration` and is passed wholesale as `:options` to `Content`, so `expandAllParameters` reaches `OperationParameters` even though the prop types are declared as narrow `Pick<>`s. Good.

The tests cover the parts I would have worried about — `aria-expanded` toggling across all four parameter locations, the deep-link case where `scrollTargetId` forces a compact row open, and the response-row guard (`'in' in parameter`) so `204` response rows keep rendering statically.

**Nits, none blocking:**

1. **The `{ expandAllParameters?: boolean } &` intersection is not needed.** This PR adds `expandAllParameters: boolean` to `ExtendedConfiguration` in `packages/types/src/api-reference/types.ts`, and `ApiReferenceConfigurationRaw` is `Omit<BaseConfiguration & ExtendedConfiguration, …>` — so the field is already reachable. Adding `| 'expandAllParameters'` to the six `Pick<>` lists would match how `expandAllResponses` is declared right next to it, instead of introducing a parallel optional field that types it more weakly than its siblings.

2. **`configuration.md` is missing the `**Type:** `boolean`` line** that `expandAllResponses` and the other neighbours all carry.

3. **A compact parameter with nothing to reveal still renders as a disclosure.** `shouldCollapse` now short-circuits on `isCompactParameter`, so `{ in: 'query', name: 'limit' }` with no schema and no description gets a clickable row that expands to empty content. Cheap guard would be to keep requiring some detail to exist.

---

PR #10284, opened 2026-09-21
perf(api-reference): load the client modal on demand

hanspagel:
Load the Test Request modal only when a request is opened, from either the reference or the agent. Opening the agent alone does not load the modal.

While the request editor downloads, the reference and agent show a loading message. If it fails, an error asks visitors to refresh the page and try again; browsers can cache failed module imports. The public lazy-client composable also works outside a Vue effect scope.

This saves about 292 KB gzip (28.3%) on startup.

| Startup JavaScript (gzip) | Before | After |
| --- | ---: | ---: |
| Read-only Petstore configuration | 1,033.5 KB | 741.2 KB |

Local ESM measurements. Exact sizes may vary slightly.

## Ticket

See #9789


cameronrohani (approved):

amritk (commented):
I went through this carefully and did not find a correctness bug — the mechanics are solid. I am commenting rather than approving because deferring 292 KB behind the primary CTA is a product tradeoff more than a mechanical refactor, and there is one UX gap I think is worth deciding on before merge.

**What I verified**

- The event-handler handoff is safe. `initializeWorkspaceEventHandlers` now returns a disposer, and the factory calls `stopReferenceClientEvents()` *before* `createApiClientModal`, so the modal's own registration never overlaps with the eager one. `app.mount()` is synchronous, so `initializeModalEvents` runs inside that call — no window where both are live. The new `workspace-events.test.ts` pins the "stop, re-init, mutation applied exactly once" invariant, and `stop()` is idempotent.
- The eager `hooks: {}` loses nothing. The only hook the modal passes is `operation:create:draft-example` → `onAfterExecute`, which just selects the new example in the modal's sidebar. Meaningless before the modal exists.
- No cross-bus interference between the reference and the agent: `agent-chat`'s `state.ts` calls `createWorkspaceEventBus()` for itself, so the two eager registrations are on different buses. The "loads independently for reference and agent event buses" test matches the real topology.
- The pending/loading state machine holds up under interleaving. open → close → open while `load()` is in flight correctly mounts and replays the *second* payload, because `pending` is overwritten and only cleared in `finally` after the replay emit. And the replay happens after `unsubscribeOpen()`, so there is no self-feedback loop.
- `onScopeDispose` covers the unmount that `Chat.vue` and `ApiReference.vue` each dropped from their own `onBeforeUnmount`.

The test file is genuinely good — the null-element retry, the rejected-import retry, and the "disposed mid-load" case are all the ones I would have asked for.

**The thing I would like a decision on**

There is no loading or error affordance for the visitor. On first click of Test Request, the button now triggers a ~292 KB download and nothing visible happens until it lands; on a slow connection that reads as a dead button. If the import rejects, `useLazyApiClient` logs to `console.error` and the button stays dead permanently from the visitor's point of view (the retry works, but only if they think to click again). Before this PR the chunk was already in flight from mount, so the click was effectively instant.

Given this lands on the reference's main interaction, some spinner-on-the-button or a toast on failure seems worth having. Happy to be told that is a follow-up.

**One nit**

`client.value = scope?.run(createClient) ?? null` silently yields `null` when `getCurrentScope()` is undefined, so calling `useLazyApiClient` outside a setup/effect scope produces a modal that never opens — plus a Vue warning from `onScopeDispose`. Both call sites are in `<script setup>` so this is fine today, but the composable is now a public entry point (`@scalar/api-client/modal/use-lazy-api-client`), so `scope ? scope.run(createClient) : createClient()` would be kinder to outside consumers.

Flagging @DemonHa since you are on the review list for the client side of this.

---

PR #10285, opened 2026-09-21
perf(api-reference): load highlighting and snippet generators on demand

hanspagel:
Load syntax-highlighting languages and code-snippet generators when the selected sample needs them. The default Petstore page fetches cURL and JSON support; selecting Python Requests downloads its generator and grammar afterward. Hiding samples avoids their generator and grammar requests.

Client-picker metadata is separate from generator implementations. Pending imports are shared, and stale asynchronous results cannot replace the current selection. Code escaping and Markdown sanitization stay synchronous, with readable plain content while highlighting loads. Unlabelled Markdown fences and HTTP-body auto-detection retain the full-registry fallback. Existing synchronous library APIs remain available through their current entry points.

Includes `tooling/esm-size`, a configuration-specific production ESM harness with frozen document support, requested-file hashes, gzip/Brotli sizes, startup budgets, and optional client-selection/modal interactions.

| Startup JavaScript, #9789 configuration | Before | Highlighting/generators only |
| --- | ---: | ---: |
| Gzip bytes | 1,033,513 | 969,179 |
| Requested files | 7 | 15 |

This saves approximately **64 KB gzip (6.2%)** independently of modal deferral. Smaller chunks increase request count. The local research comparison uses identical dependency builds and Node 24, but some workspace outputs remain cached; these are payload estimates, not a release CDN budget. The harness records eventual interaction bytes separately from startup.

The client modal still loads on mount in this PR. Modal deferral is a separate, independently mergeable change.

## Ticket

See #9789


hanspagel:
not enough improvement

PR #10286, opened 2026-09-21
fix(mock-server): support document origins and exports

hanspagel:
Fix JSON and YAML exports for file and URL inputs. Add `origin` so already loaded documents can resolve relative references without fetching the root again.

For a file containing:

```json
{"openapi":"3.1.0","info":{"title":"Pets","version":"1.0.0"},"paths":{}}
```

The export routes now return the loaded API description instead of treating the file path as document content.


cameronrohani (approved):

amritk (approved):
Approving. This is a real bug fix with tests that actually pin the behavior rather than just exercising it.

The export bug was clear: `/openapi.json` was handed the raw `configuration.document`, so for a file path or URL input it serialized the *path string* instead of the document. Routing those cases through `getRaw(schema)` fixes it and also means relative `$ref`s come out bundled, which is what an exported document needs to stay usable off-server.

Things I checked:

- The `exportDocument` condition reads correctly — `||` binds tighter than `?:`, so it is `(origin || (string input that does not normalize to an object)) ? getRaw(schema) : input`. Inline JSON/YAML strings normalize to objects and so keep the caller's original bytes; file paths and URLs do not (a colon not followed by a space stays a plain YAML scalar), so they take the bundled branch. The three-way `object`/`json`/`yaml` inline test locks that in.
- `getRaw(schema)` is safe to serialize here even though the fixture has a self-recursive `Pet.friend` → `#/components/schemas/Pet`, because the raw document keeps `$ref` strings rather than resolved cycles. The `not.toContain('$ref-value')` assertion is a nice guard against magic-proxy internals leaking into the export.
- `expect(dispatch).toHaveBeenCalledTimes(preloaded ? 1 : 2)` is the assertion that actually earns the "without fetching the root again" claim, and `assertNoPendingInterceptors()` confirms the relative `./pets.json` still resolves against the URL origin rather than falling back to `basePath`.

**Two nits, neither blocking:**

1. The changeset is `patch`, but this adds a new public `origin` option to `MockServerOptions`. `minor` would describe it better.
2. The new doc line says the export routes "return the loaded, bundled OpenAPI 3.1 document". #10288 is open right now and moves `processOpenApiDocument` to upgrade to 3.2, so whichever of these merges second will leave that sentence stale. Worth fixing up in whichever lands later.

---

PR #10287, opened 2026-09-21
perf(api-reference): load the agent when first opened

hanspagel:
Load the agent drawer and chat when the visitor first opens it. Keep the conversation when closing and reopening.

This saves about 56.5 KB gzip on startup when the agent is enabled but unopened.

| Startup JavaScript (gzip) | Before | After |
| --- | ---: | ---: |
| Agent enabled, unopened | 1,089.7 KB | 1,033.2 KB |

Local ESM measurements from before the final scope cleanup. Exact sizes may vary slightly.

## Ticket

See #9789


cameronrohani (approved):

amritk (approved):
Approving — the shape of this is right. Hoisting the async boundary from the chat interface up to the whole drawer, and latching `hasOpenedAgent` so the drawer never unmounts again, gets the 56.5 KB off startup without throwing the conversation away on close. The `v-if` keeps `agent.agentEnabled.value` as the outer condition, so a disabled agent still renders nothing.

The test is the best part: mounting the real `ApiReference` and comparing the `AgentScalarChatInterface` `vm` identity across two close/reopen cycles is exactly what proves the "keep the conversation" claim, and doing it without stubs means a lazy-loading failure surfaces instead of hiding.

Also confirmed dropping `AgentScalarDrawer` from `components/AgentScalar/index.ts` is safe — it is not re-exported from `src/components/index.ts` and nothing outside `ApiReference.vue` referenced it, so it was never reachable through the published `@scalar/api-reference/components` entry.

**One nit, non-blocking:** `AgentScalarDrawer.vue` animates with `v-show` inside two `<Transition>` wrappers, and Vue skips the enter transition on a component's initial render unless the transition is marked `appear`. Previously the drawer was already mounted by the time the visitor clicked, so the fade and slide-in played on the very first open. Now the component mounts with `showAgent` already `true`, so the first open snaps in with no animation, and only the second and later opens animate. Adding `appear` to both `<Transition>` elements should restore it.

---

PR #10288, opened 2026-09-21
refactor(mock-server): upgrade documents to OpenAPI 3.2

hanspagel:
Upgrade bundled documents to OpenAPI 3.2 and use 3.2 for empty inputs. Allow authentication to accept the upgraded document type.

For this input:

```json
{"openapi":"3.1.0","info":{"title":"Pets","version":"1.0.0"},"paths":{}}
```

The mock server now processes an OpenAPI 3.2.0 document instead of 3.1.0.


cameronrohani (approved):

amritk (approved):
Low risk — approving. I traced the blast radius of moving the upgrade target from `3.1` to `3.2` and it is narrower than it looks:

- `3.1-to-3.2` only migrates XML objects (`xml.attribute`/`xml.wrapped` → `xml.nodeType`) and `x-tagGroups` → tag hierarchy, then bumps the version string.
- The mock server does not read schema-level `xml` metadata at all — `serialize-response-body.ts` derives XML from the response body via `json2xml`, so the XML migration cannot change a mock response.
- It does not read `tags` for routing either, so the tag-group migration is inert here.
- Example selection (`select-response-example.ts`, `collect-sse-events.ts`, `mock-handler-response.ts`) reads `mediaType.example` / `examples[name].value`, and the 3.1→3.2 step leaves both untouched. The `example` → `examples` rewrite that does exist lives in `3.0-to-3.1`, which was already in the path before this PR.

So the real user-visible change is that `GET /openapi.json` and `/openapi.yaml` now serve a `3.2.0` document, which is what the `minor` changeset is for.

The new tests are good coverage — parametrizing across 2.0/3.0.4/3.1.2/3.2.0, asserting 3.2-native fields (`paths./search.query`, `tags[].kind`) survive untouched, and checking the bundled-plus-recursive case still serializes via `getRaw` without blowing up on the cycle.

Small nit, non-blocking: `handleAuthentication` now takes `OpenAPIV3_2.Document` but still types `operation` as `OpenAPIV3_1.OperationObject`, and `resolveSchemes` still returns `OpenAPIV3_1.SecuritySchemeObject[]`. Harmless today since security schemes did not change shape in 3.2, but the mixed generations will read oddly next time someone touches this file.

---

PR #10290, opened 2026-09-21
chore(deps): update hono and migrate websocket support

hanspagel:
Update Hono and its dependencies. Move mock-server WebSockets to Node server v2, removing `@hono/node-ws` and the remaining Node server 1.x dependency.

AsyncAPI callers now pass `websocket` to `serve({ fetch: app.fetch, websocket })` instead of calling `injectWebSocket(server)`. Includes updated Docker setup, docs, and a minor changeset for the API change.


cameronrohani (approved):

amritk (commented):
Not approving this one — not because I found a bug, but because it is a major dependency bump plus a breaking change to a published API, and it currently does not merge cleanly.

**Blocking**

- `mergeable_state` is `dirty` — needs `main` merged in and the lockfile regenerated with pnpm before this can land.

**What I checked and liked**

- The new end-to-end test in `create-asyncapi-mock-server.test.ts` that opens a real `ws` client against `serve({ …, websocket })` is exactly the right way to prove the v2 migration works. It asserts the pushed message, the echo reply, the `onMessage` ordering, a clean 1000 close, and that plain HTTP routing still 404s. That plus a green `types` job is good evidence that the module-level `upgradeWebSocket` from `@hono/node-server` v2 correctly binds to the `WebSocketServer` handed to `serve()`.
- Cleaning up the docker `startMockServer` so it no longer has to hold the server handle just to call `injectWebSocket(server)` is a nice simplification.
- The changeset spells out the migration (`injectWebSocket(server)` → `serve({ fetch, port, websocket })`), which is what consumers will need.

**Worth a second opinion before merge**

- `@hono/node-server` goes `^1.19.10` → `^2.1.1` across the whole catalog, so this also moves `void-server`, `proxy-scalar-com`, `galaxy-scalar-com` and the examples onto v2, not just the mock server. Those all pass CI here, but the blast radius is wider than the title suggests.
- `createAsyncApiMockServer()` returning `websocket` instead of `injectWebSocket` is a breaking change for anyone calling it directly, and it also requires them to be on `@hono/node-server` v2 (v1's `serve()` ignores the `websocket` option, so WebSocket channels would silently stop accepting connections rather than fail loudly). The changeset is `minor` per repo convention, which is fine, but that silent-failure mode is worth a line in the release notes.
- Since @DemonHa is on the review list and owns this area, I would rather they sign off on the AsyncAPI transport semantics than have me rubber-stamp it.

Nothing here needs changing beyond the conflict as far as I can tell.

---

PR #10291, opened 2026-09-21
perf(api-reference): defer disabled developer tools

hanspagel:
Evaluate developer-tools visibility before importing the toolbar. With `showDeveloperTools: 'never'`, or the default setting on a non-local URL, the configuration, share, and deploy UI is no longer loaded just to render nothing.

Keep the public `DeveloperTools` component as a lightweight visibility wrapper. Its async toolbar receives the same configuration, workspace, external URLs, attributes, and two-way overrides. The `always`, `never`, localhost, and runtime visibility behavior stays the same.

Controlled local production ESM measurements for the frozen Petstore document:

| Startup JavaScript | Before | Disabled tools after | Visible tools after |
| --- | ---: | ---: | ---: |
| Raw bytes | 3,542,883 | 3,496,436 | 3,545,528 |
| Gzip bytes | 1,034,491 | 1,026,725 | 1,041,185 |
| Brotli bytes | 854,684 | 854,423 | 866,783 |
| Requested files | 7 | 11 | 13 |

Disabled tools save **46,447 raw bytes / 7,766 gzip bytes**. The toolbar chunk is absent from disabled-page requests and arrives when visible. Code splitting increases shared-chunk request count and compression overhead: visible tools cost an additional 6,694 gzip bytes, and disabled-page Brotli savings are only 261 bytes. Shared styles remain in the initial stylesheet.

Measurements sum individually compressed requested files after a two-second observation window, with identical dependency builds and Node 24. Some workspace outputs remain cached; these are local payload comparisons, not a release CDN budget or a timing claim.

Independent of the modal, highlighting/generators, and agent PRs (#10284, #10285, #10287).

## Ticket

See #9789


hanspagel:
not enough saved

PR #10300, opened 2026-09-21
fix(api-reference): pick badge text color from the background's light…

eightdeekay:
## Problem

A colored `x-badges` entry derives its text color from its own background, by mixing that
background with black:

```js
color: `color-mix(in srgb, ${color}, black 40%)`
```

That only reads well when the background is pale. On a mid-tone color the text is muddy, and
on a dark one it is barely there — a badge with `color: black` renders black text on black,
so the label cannot be read at all.

Every color in the "before" row below is a legal `x-badges` color today.

## Visual

badge-contrast

"Before" is the published `@scalar/api-reference@1.69.0` browser bundle; "after" is this
branch. Same document, same viewport.

## Solution

Pick the text color from the background's lightness instead — black on a light background,
white on a dark one — and do it in CSS, so the browser reads the lightness and no color
parsing happens in JavaScript:

```css
color: lch(from var(--badge-color) calc((50 - l) * infinity) 0 0);
```

`l` is the background's lightness in LCH (0–100). Multiplying its distance from the midpoint
by `infinity` and letting the channel clamp turns that into 0 or 100 — black or white — for
any color format a user can write, including `color(...)` and `oklch(...)`.

The rule sits behind `@supports`, so browsers without relative color syntax (Chrome < 119,
Safari < 16.4, Firefox < 128) keep exactly today's rendering rather than losing the color.

`Badge.vue` now passes the background in as a `--badge-color` custom property rather than an
inline `background-color`, which is what lets the CSS derive from it. Badges with no `color`
are untouched and keep the themed `text-*` variants.

### Alternative considered

Adding a `textColor` field to `x-badges`, so a document could state the text color outright.
It fixes the symptom but pushes the problem onto everyone writing a document, needs the field
threaded through the schema packages, and leaves the default just as unreadable for anyone who
does not set it. Deriving a sensible default seemed like the better fix; an explicit override
could still be added later if there is demand.

## Checklist

- [x] I added a changeset.
- [x] I added tests.


hanspagel (approved):
quality PR

PR #10303, opened 2026-09-21
fix(api-client): preserve parameter edits across example overlays

### LINKED ISSUE #10299: Query parameter edits disappear after editing a sibling parameter (1.70.0)
### What happens?

In the standalone CDN build of `@scalar/api-reference@1.70.0`, editing a JSON-content query parameter can discard an earlier edit to a neighboring schema-based parameter. The earlier value remains visible in the editor, but the actual next request omits it or reverts to its authored example.

This is distinct from #10267: I retested #10268 in 1.70.0, and enabling, editing, disabling and re-enabling the JSON-content parameter itself works. The remaining problem is loss of the neighboring parameter's edit. Chromium 153; OpenAPI 3.1; no application framework, request interceptor, plugin or custom serializer.

### Reproduction

Run the self-contained Node script below, open its `?case=external` page, then:

1. Open **Test Request** for `GET /search/external`.
2. The required `term` dropdown is initially empty, although the cURL snippet contains `term=test`. Select `test` from that dropdown and let the UI settle.
3. Send. The echo server receives both `term=test` and the initial JSON `filter`.
4. Replace the filter with `{"owner":"null","tags":[]}`, leave `term` unchanged, and let the UI settle again.
5. Send. The editor still shows `term=test`, but the server receives only `filter`.

Actual consecutive received requests:

```text
/search/external?term=test&filter=%7B%22owner%22%3Anull%2C%22tags%22%3A%5B%5D%7D
/search/external?filter=%7B%22owner%22%3A%22null%22%2C%22tags%22%3A%5B%5D%7D
```

This persists after waiting and sending again; it is not just an immediate-click/debounce result. Entering a custom term, then editing the filter again, repeats the loss.

The `?case=inline` page is a useful control: the authored `test` fallback survives a filter edit. To expose lost edits there, enter a *different* custom term. A published-package reproduction using the real reader, overlay, mutator and request builder shows that custom term reverting to `test` after the sibling edit. A pre-existing parameter-level `examples.default` map preserves the custom edit.

### Expected behavior

A user-entered query value should survive an edit to a different parameter. The editor, snippet and sent request should agree. Defaults/examples on a resolved schema reference should be available just as they are for an inline schema.

### Source diagnosis

Examined release/main commit `e2f3b5c64c3e96813db2d19aa8832d2d9826a2ae`, including published `@scalar/workspace-store@0.64.0`. Seven Node reproduction checks exercise the native published functions with Vue reactive/computed state, with no debounce emitter or copied implementation.

Two issues interact:

1. [`getExample`](https://github.com/scalar/scalar/blob/e2f3b5c64c3e96813db2d19aa8832d2d9826a2ae/packages/workspace-store/src/request-example/builder/helpers/get-example.ts#L69-L93) resolves the parameter, but does not resolve `parameter.schema` before reading `default`, `enum`, `examples` or `example`. A resolved schema `$ref` therefore gives no fallback value to the request builder. The widget/schema and snippet paths resolve that reference separately, explaining the initial discrepancy.
2. [`resolveOperationExamples`](https://github.com/scalar/scalar/blob/e2f3b5c64c3e96813db2d19aa8832d2d9826a2ae/packages/workspace-store/src/helpers/operation-examples.ts#L45-L54) shallow-copies every schema-based parameter even when there is no fetched example overlay. The row passes that copy as `originalParameter` to the mutator. When the source has no parameter-level `examples` map, [`upsertOperationParameter`](https://github.com/scalar/scalar/blob/e2f3b5c64c3e96813db2d19aa8832d2d9826a2ae/packages/workspace-store/src/mutators/operation/parameters.ts#L60-L105) creates the edit only on the detached copy. Editing the sibling content parameter mutates its retained reactive media object, invalidates the computed operation, and recreates the first parameter from the unedited source.

The native-function control confirms that writing the same edit to the source parameter preserves it across the sibling change. Resolving the schema fallback alone would mask omission with an authored value, but would not preserve a custom edit. Fetched external example payloads should remain read-only overlays rather than being persisted as document changes.

### OpenAPI / self-contained browser script

Save as `repro.mjs`, run `node repro.mjs`, and use `?case=external` or `?case=inline`. `/requests` retains the actual wire requests.

```js
import { createServer } from 'node:http';
const requests = [];
const textSchema = { type: 'string', minLength: 2, examples: ['test'] };
const server = createServer((req, res) => {
  const url = new URL(req.url, `http://${req.headers.host}`);
  const json = (data) => {
    res.setHeader('Content-Type', 'application/json');
    res.end(JSON.stringify(data));
  };
  if (url.pathname === '/query.json') return json({
    $schema: 'https://json-schema.org/draft/2020-12/schema',
    type: 'object', properties: { term: textSchema }, required: ['term'],
  });
  if (url.pathname === '/requests') return json(requests);
  if (url.pathname.startsWith('/search/')) {
    const received = { url: req.url, query: Object.fromEntries(url.searchParams) };
    requests.push(received);
    console.log(JSON.stringify(received));
    return json(received);
  }
  if (url.pathname.startsWith('/spec/')) {
    const variant = url.pathname.split('/').at(-1);
    return json({
      openapi: '3.1.0', info: { title: 'Query example reproduction', version: '1' },
      servers: [{ url: url.origin }],
      paths: { [`/search/${variant}`]: { get: {
        operationId: 'search', summary: 'Echo query', parameters: [
          { name: 'term', in: 'query', required: true, schema: variant === 'inline'
            ? textSchema : { $ref: '/query.json#/properties/term' } },
          { name: 'filter', in: 'query', content: { 'application/json': {
            schema: { type: 'object', properties: {
              owner: { type: ['string', 'null'] },
              tags: { type: 'array', items: { type: 'string' } },
            } },
            example: { owner: null, tags: [] },
          } } },
        ], responses: { 200: { description: 'Received query parameters' } },
      } } },
    });
  }
  const variant = url.searchParams.get('case') === 'inline' ? 'inline' : 'external';
  res.setHeader('Content-Type', 'text/html');
  res.end(`<!doctype html><html><head><title>Query example reproduction</title></head>
<body><div id="app"></div>
<script src="https://cdn.jsdelivr.net/npm/@scalar/api-reference@1.70.0"></script>
<script>Scalar.createApiReference('#app', {url: ${JSON.stringify(url.origin + '/spec/' + variant)}});</script>
</body></html>`);
});
server.listen(0, '127.0.0.1', () => console.log(`http://127.0.0.1:${server.address().port}/?case=external`));
```

fitchmultz:
## Problem

Editing one query parameter can silently discard an edit to another parameter. In API Reference 1.70.0, enter `entered` for `term`, then edit the JSON-valued `filter`: the editor still displays `entered`, but the next request omits `term`. With an inline schema example, the same edit can instead revert to the authored value.

Parameters whose schema uses `$ref` also start with an empty request value even when the referenced schema supplies an example and the documentation snippet displays it.

## Solution

Keep the original parameter object when no downloaded example is overlaid. When an overlay is needed, route edits to the existing authored parameter through `originalParameter`, keeping downloaded examples read-only until explicitly edited. Let the flushed edit propagate through Vue before constructing a request on Send.

Read schema fallback values through the existing reference resolver and sibling-merging transform. This preserves annotations beside `$ref` and the existing example/default precedence. Request serialization is unchanged.

## Reproduction

Serve these two files from the same origin, with `GET /search` returning the query parameters it receives. Adjust the server URL to your local echo server.

`openapi.yaml`:

```yaml
openapi: 3.1.0
info:
  title: Query edit persistence
  version: '1'
servers:
  - url: http://127.0.0.1:3000
paths:
  /search:
    get:
      summary: Echo query parameters
      operationId: search
      parameters:
        - name: term
          in: query
          required: true
          schema:
            $ref: /query.json#/properties/term
        - name: filter
          in: query
          content:
            application/json:
              schema:
                type: object
                properties:
                  owner:
                    type: [string, 'null']
                  tags:
                    type: array
                    items:
                      type: string
              example:
                owner: null
                tags: []
      responses:
        '200':
          description: Actual received query parameters
```

`query.json`:

```json
{
  "$schema": "https://json-schema.org/draft/2020-12/schema",
  "type": "object",
  "properties": {
    "term": { "type": "string", "examples": ["authored"] }
  },
  "required": ["term"]
}
```

1. Open **Test Request**. Add the custom value `entered` for `term` and send the request.
2. Replace `filter` with `{"owner":"null","tags":[]}` and send again.
3. Repeat the edit with `{"owner":"","tags":["kept"]}`, then disable and re-enable `filter`.

Before: step 2 drops `term` from the actual request, despite the editor still displaying it. After: every request retains `term=entered`; disabling `filter` omits only `filter`, and re-enabling it restores its last edited value. The referenced `authored` example is also available immediately when opening the request.

## Visual

Both screenshots show actual responses from a local echo server after step 2. The before image uses the published 1.70.0 standalone asset; the after image uses this branch's locally built standalone asset at `e0b5e64`.

**Before — the editor shows `entered`, but the received query contains only `filter`.**

Before: term is displayed but omitted from the request

**After — the received query retains `term=entered`.**

After: the request retains the entered term alongside the edited filter

## Ticket

Fixes #10299


fitchmultz:
Verification for `e0b5e64be4818ad42ae15ae019fd4361957e850f`, based on current `main` at `e2f3b5c64c3e96813db2d19aa8832d2d9826a2ae`:

- 192 scoped workspace-store tests and 106 scoped api-client tests pass. The regressions exercise source identity, downloaded-example edits, schema-reference fallback/sibling annotations, and pressing Send while an edit is still pending.
- Both affected-package typechecks, workspace-store/api-client builds, and the API-reference standalone build pass.
- Changed-file Biome/Prettier/ESLint, scoped Knip, changeset status, and `git diff --check` pass. ESLint retains an unchanged Vue event-name warning; changeset status emitted a nonfatal `/dev/tty` diagnostic.
- Independent source review against current main is complete; both findings from that review were corrected and rechecked.
- The real-browser before/after run used the published 1.70.0 asset and this branch's local standalone build. Actual HTTP requests confirm that repeated edits and disabling/re-enabling `filter` retain `term=entered`. The screenshots in the description are GitHub-uploaded assets, not committed files.

The local browser build is not a published release. The PR-title check passes. GitHub's PR and Dependency Review workflows currently require maintainer approval for this fork contribution.

hanspagel (approved):
🙏

PR #10304, opened 2026-09-22
fix(astro): accept Astro 7 as a peer dependency

haydenbleasel:
## Problem

`@scalar/astro` declares `peerDependencies.astro: "^4.0.0 || ^5.0.0 || ^6.0.0"`, but Astro 7 has been out for a while and the official integrations (`@astrojs/mdx`, `@astrojs/node`, `@astrojs/vercel`, `@astrojs/starlight` 0.42) all require it now. Any project on Astro 7 that installs `@scalar/astro` gets a peer warning, and npm goes further: it silently installs a second, nested `astro@6.x` under `@scalar/astro` to satisfy the range, which adds ~120 packages to a fresh install.

This surfaced downstream in [Blume](https://github.com/haydenbleasel/blume/issues/295), which ships `@scalar/astro` for its `scalar()` reference adapter on Astro 7:

```text
bun install v1.4.2
warn: incorrect peer dependency "astro@7.3.3"
```

## Solution

Add `^7.0.0` to the peer range. Nothing in the integration imports Astro's runtime, and the two `.astro` components use no version-specific APIs, so this is a metadata fix.

To check that claim I ran this package's own `vitest` suite and `astro check` against a clean install of `astro@7.3.3` (and again against `astro@6.4.8`). `astro check` was clean on both. One test needed a tweak: Astro 7 escapes double quotes inside attribute values as `&quot;` where Astro 6 emitted `&#34;`, so the client-render assertion now normalizes the entity before matching. All three tests pass on both majors.

I left the `devDependencies` range and the lockfile alone so `pnpm install --frozen-lockfile` keeps working; CI still exercises the version the lockfile resolves. Widening that range and the pnpm catalog is a separate choice when you want CI on Astro 7. `@scalar/starlight` still declares `^5.0.0 || ^6.0.0`; I did not touch it because the catalog pins Starlight 0.37, which itself needs Astro 5/6.

## Checklist

- [x] I added a changeset.
- [x] I added tests.


hanspagel (approved):
thanks!

PR #10305, opened 2026-09-22
fix(snippetz): preserve nested arrays in JavaScript snippets

### LINKED ISSUE #10301: Nested number[][] displayed incorrectly for JavaScript example
### What happens?

https://sandbox.scalar.com/e/wZso5

Given schema, that has nested number array (number[][]) - 
In some language previews - it is generated broken and incorrect.
I have noticed on multiple JS options - like Fetch.

e.g.

in JS Fetch - it tells me to use it like this:

```
fetch('/test', {
  method: 'POST',
  headers: {
    'Content-Type': 'application/json'
  },
  body: JSON.stringify({
    type: 'Example',
    coordinates: [
      {
        '0': [1]
      }
    ]
  })
})
```
For C# HttpClient - it looks all right.
```
using var client = new HttpClient();

var request = new HttpRequestMessage(HttpMethod.Post, "/test");
request.Content = new StringContent(
"""
{
  "type": "Example",
  "coordinates": [
    [
      [
        1
      ]
    ]
  ]
}
""",
System.Text.Encoding.UTF8, "application/json");

using var response = await client.SendAsync(request);
```
I think - somewhere where the schema is parsed for JS - it's arrays are treated as objects, so it looks like that..

If I click try the requests - it is all rights there.

I expect him to not display broken example

### OpenAPI Document

https://sandbox.scalar.com/e/wZso5

hanspagel:
Fix JavaScript and Node.js snippets turning nested arrays into objects with numeric keys. For example, `[[[1]]]` now stays `[[[1]]]` instead of becoming `[{ '0': [1] }]`.

Reproduce with this API description and select JavaScript Fetch:

```yaml
openapi: 3.2.1
info: { title: Nested arrays, version: '1.0' }
paths:
  /test:
    post:
      requestBody:
        content:
          application/json:
            schema:
              type: array
              items:
                type: array
                items: { type: number }
            example: [[1, 2], [3, 4]]
      responses:
        '200': { description: OK }
```

## Ticket

Fixes #10301


cameronrohani (approved):

amritk (approved):
Clean fix. With arrays handled at the top of `objectToString`, top-level and nested arrays now take one route instead of being turned into objects by `Object.entries` (before, `[[1,2],[3,4]]` rendered as `{ '0': [1, 2], '1': [3, 4] }`).

Output for inputs without nested arrays (headers, objects inside arrays, repeated query params, deep objects) is identical to `main`. Indentation for arrays inside objects matches the removed branch. Every generator that shares the helper (js/node fetch, ofetch, undici, axios) gets the fix, and the PHP generators are separate and untouched.

Two things I noticed. Neither blocks this PR, since both existed before it:
- String items in arrays (and string values in objects) are wrapped as `'${item}'` without escaping, so a value like `it's` still produces invalid JS. That could be a follow-up using `escapeJsString`.
- Optional: a node/undici or axios test with a top-level array body would guard against the generators drifting apart.

LGTM 🚀

---

PR #10306, opened 2026-09-22
fix(openapi-upgrader): preserve Swagger x-nullable

### LINKED ISSUE #10296: upgrade() drops x-nullable when upgrading Swagger 2.0, losing nullability
### What happens?

`upgrade()` leaves `x-nullable` untouched when upgrading a Swagger 2.0 document, so the resulting 3.1 document has no null union

Swagger 2.0 has no `nullable` keyword, so `x-nullable` is the convention tooling settled on to express it. The 3.0 → 3.1 step already knows how to turn `nullable: true` into a `['<type>', 'null']` union, but the 2.0 → 3.0 step never produces a `nullable` for it to act on, so the nullability is lost one hop earlier.

The asymmetry is easy to see side by side:

```js
import { upgrade } from '@scalar/openapi-upgrader'

// Swagger 2.0 — x-nullable
upgrade({
  swagger: '2.0', info: { title: 't', version: '1' }, paths: {},
  definitions: { T: { type: 'object', properties: { v: { type: 'number', 'x-nullable': true } } } },
}, '3.1')
// -> { "type": "number", "x-nullable": true }      ❌ nullability lost

// OAS 3.0 — nullable
upgrade({
  openapi: '3.0.0', info: { title: 't', version: '1' }, paths: {},
  components: { schemas: { T: { type: 'object', properties: { v: { type: 'number', nullable: true } } } } },
}, '3.1')
// -> { "type": ["number", "null"] }                 ✅ handled
```

Reproduced on `@scalar/openapi-upgrader@0.2.17`.

### Why this matters downstream

This surfaced as a silent codegen regression in [orval](https://github.com/orval-labs/orval). orval v8.0.0 ([`9479e54b4`](https://github.com/orval-labs/orval/commit/9479e54b4), "Upgrade openapi to 3.1") replaced `swagger2openapi` with `@scalar/openapi-parser`. The previous converter translated the extension explicitly ([`swagger2openapi/index.js:173-175`](https://github.com/Mermade/oas-kit/blob/b1bba3fc5007e96a991bf2a015cf0534ac36b88b/packages/swagger2openapi/index.js#L173-L176)):

```js
if (typeof schema["x-nullable"] === 'boolean') {
    schema.nullable = schema["x-nullable"];
    delete schema["x-nullable"];
}
```

So the same Swagger 2.0 document generates different types depending on orval version:

| orval | generated TypeScript |
| --- | --- |
| 7.21.0 (`swagger2openapi`) | `myField?: number \| null` ✅ |
| 8.31.0 (`@scalar/openapi-parser`) | `myField?: number` ❌ |
| 8.35.0 (latest) | `myField?: number` ❌ |

`x-nullable: true` on a Swagger 2.0 schema should produce the same 3.1 output as `nullable: true` on a 3.0 schema:

```json
{ "type": ["number", "null"] }
```

Mapping `x-nullable` onto the 3.0 `nullable` keyword during the 2.0 → 3.0 hop is enough. The existing 3.0 → 3.1 logic ([`upgrade-from-three-to-three-one.ts:95`](https://github.com/scalar/scalar/blob/e2f3b5c64c3e96813db2d19aa8832d2d9826a2ae/packages/openapi-upgrader/src/3.0-to-3.1/upgrade-from-three-to-three-one.ts#L95)) then handles it like any other nullable schema, including the `$ref` and combinator cases.

### OpenAPI Document

https://sandbox.scalar.com/e/ejdnm

hanspagel:
Swagger 2.0 upgrades left `x-nullable` untranslated, so nullable fields lost their null type. Convert it to `nullable` in OpenAPI 3.0 so later upgrades preserve null values, including in parameters and headers.

```yaml
swagger: '2.0'
info: { title: Example, version: '1' }
paths: {}
definitions:
  Value:
    type: number
    x-nullable: true
```

Previously, upgrading this to 3.1 kept `type: number`. It now produces `type: [number, 'null']`.

## Ticket

Fixes #10296


cameronrohani (approved):

amritk (approved):
This mirrors swagger2openapi's `x-nullable` handling and fills the gap before the 3.0→3.1 step, which already knows how to build null unions. The traversal only visits subschemas (`properties`, `items`, `additionalProperties`, `allOf`), so `example`/`default` data and other extensions stay untouched. The tests cover that well, including a property literally named `x-nullable`. Coverage of parameters, headers, form data, body parameters and global responses looks complete.

A couple of small, non-blocking thoughts:

- **`x-nullable: false` ends up as `nullable: false` in 3.1/3.2 output.** The 3.0→3.1 step only removes `nullable` when it is `true`, so `nullable: false` survives into documents where `nullable` is not a keyword. The test at `upgrade.test.ts:275` asserts this for every target. Before this change it stayed as a harmless `x-nullable: false` extension. `false` is the default in 3.0 anyway, so we could write `nullable` only when it is `true` and just drop `x-nullable: false`. Alternatively, the 3.0→3.1 step could delete `nullable: false`.
- Nice-to-have: the request/response/parameter/header test only asserts the 3.1 output. A quick 3.0 case (`nullable: true`) would also pin down the intermediate step.
- Possible follow-up: NSwag also emits `x-nullable` on body parameters and on Response Objects themselves. Those are left as they are today, which matches swagger2openapi, so this is not a regression.

LGTM 👍

---

PR #10308, opened 2026-09-22
fix(galaxy): always use the ESM build

hanspagel:
Always use ESM for Galaxy. Production loads the published ESM build from jsDelivr. Staging and PR previews load the branch-built ESM version and its chunks.


cameronrohani (approved):

amritk (approved):
Looks good 👍

I checked the PR preview end to end. `/` renders the module import for `/scalar/standalone.esm.js`, and all chunks reachable from the entry, including the lazy ones, resolve through the `/scalar/*` → `ASSETS` route. Production passes `bundle: true`, which renders the `@scalar/api-reference/esm.js` import from jsDelivr, so production and previews now use the same build format.

One small, non-blocking nit: the build script copies `standalone.esm.js` but not `standalone.esm.js.map`. The entry still has a `sourceMappingURL` comment, so DevTools gets a 404 for it on previews. The chunk maps work because the whole `chunks/` folder is copied. Something like `cp node_modules/@scalar/api-reference/dist/browser/standalone.esm.js* dist/scalar/` would bring the map along.

The tests for the rendered module import in both modes, plus the assets passthrough, are a nice addition.

---

PR #10310, opened 2026-09-22
feat(blocks): connect SDK samples to the example switcher

hanspagel:
SDK samples now follow the request example switcher while keeping the selected language. Samples sharing a language and label appear once in the language menu. Static samples without an example link hide the switcher; a linked sample missing for the selected example shows an unavailable message.

With this API description, switching from `simple` to `detailed` updates the Python sample from `Basic` to `Premium`. Previously, the two samples appeared as separate language options and changing the request example did not update the SDK sample.

```yaml
openapi: 3.1.0
info: { title: Widgets API, version: '1.0' }
paths:
  /widgets:
    post:
      requestBody:
        content:
          application/json:
            schema: { type: object }
            examples:
              simple: { value: { name: Basic } }
              detailed: { value: { name: Premium } }
      responses:
        '200': { description: Created }
      x-codeSamples:
        - lang: Python
          example: simple
          contentType: application/json
          source: client.widgets.create(name="Basic")
        - lang: Python
          example: detailed
          contentType: application/json
          source: client.widgets.create(name="Premium")
```

See [the complete example](https://github.com/scalar/scalar/blob/codex/linked-sdk-examples/documentation/openapi.md#link-code-samples-to-request-examples).


cameronrohani (approved):

amritk (commented):
This is a clean way to link SDK samples to request body examples. Only samples that set `example` get grouped, so existing `custom/<lang>` ids and saved selections stay stable. The tests cover the tricky cases well: matching by key plus media type, the fallback for samples without `contentType`, keeping the selected language while switching examples, and webhooks.

I'm leaving a comment rather than approving because of one behaviour change I'd like confirmed first:

- **Picker hidden for static SDK samples** (`CodeExample.vue`, `showExamplePicker`). Custom samples are selected by default, and none of today's samples have `example` links. So every existing page with Stainless or `x-scalar-examples` samples will lose the body example picker on first load. That picker also decides which example "Test Request" opens with. Users can get it back by switching to a built-in client like cURL, but that is easy to miss. If this is what we want for existing SDK users, 👍. Otherwise, maybe keep the picker visible for static samples.

Smaller things, none blocking:

- **API client follow-up:** `RequestCodeSnippet` in `packages/api-client/src/v2/blocks/request-block/RequestBlock.vue` is never given `selectedExample`. So in the API client a linked SDK sample always shows the first example's source. Passing `:selectedExample="exampleKey"` should be enough.
- **ReadMe (optional):** ReadMe samples already carry `correspondingExample`. Mapping it to `example` in `fromReadmeSamples` would give ReadMe documents this feature for free.
- **Nit:** `contentType` matching is exact, so `application/json; charset=utf-8` will not match `application/json`. That is probably fine, but worth a sentence in the docs.
- **Nit:** the "No code sample available for this example." message is hard-coded English and is rendered and copyable as code (same as the old "Custom example not found").
- **Nit:** the "complete example" link in the description points at the feature branch and will break after merge.

---

PR #10311, opened 2026-09-22
fix(localization): preserve dollar sequences in interpolation

hanspagel:
This fix is now included in `main` through #10175. The standalone branch was originally prepared separately in response to its review feedback.

Interpolated translation values currently treat dollar sequences as replacement instructions. For example, translating `Save {name}` with the name `$&` produces `Save {name}`, while `$$` becomes a single dollar sign.

Insert interpolation values literally so dollar sequences and repeated placeholders retain their intended text. This fixes existing API Reference translations and is released as a patch to `@scalar/localization`.

## Ticket

See #10175


cameronrohani (approved):

amritk (commented):
The fix is right: passing a replacer function to `replaceAll` makes the value literal, and the `it.each` over `$&`, `$'`, `` $` ``, `$1` and `$$` with a repeated placeholder is a nice, focused test.

This has already landed on `main` through #10175, though. The same `() => String(paramValue)` line is at `packages/localization/src/create-localization.ts:157` and the same test is in `create-localization.test.ts`. It also shipped in `@scalar/localization@0.2.1` (#10307), whose changelog already has this entry.

Merged into current `main`, the only change left would be `.changeset/tidy-dollars-translate.md`. That would cut an unneeded patch release of `@scalar/localization` and its downstream packages, with a duplicate changelog line. I think we can close this as superseded by #10175.

Small follow-up idea, not caused by this PR: interpolation fills parameters one after another. So a value that contains another `{placeholder}` gets filled in again by a later step. A single-pass `template.replace(/\{(\w+)\}/g, …)` would close that off if user-controlled values ever reach `translate`.

---

